PR #2986, opened 2025-03-23
chore: Refactor return of first gen token in PD

Shunkangz:
In the current implementation, we will return the first and second generated token together from generation worker. I refactor this logic and return the first generated token from context worker as soon as it finishes computation.

PR #3241, opened 2025-04-02
fix: fix return double first token

Shunkangz:
In PD, we have different behaviors for overlap and non-overlap scheduler. With non-overlap scheduler, we always return the first two generated tokens togethers. With overlap scheduler, the request might return the response without calculating the second generated token. This MR fix the change of https://github.com/NVIDIA/TensorRT-LLM/pull/2986 in overlap scheduler case.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/decoder.py
+++ b/tensorrt_llm/_torch/pyexecutor/decoder.py
@@ -281,6 +281,7 @@ class TorchDecoder(Decoder):
             if request.state != LlmRequestState.GENERATION_COMPLETE:
                 new_token = new_tokens_list[idx]
                 num_tokens = request.add_new_token(new_token, beam_idx)
+                request.decoding_iter += 1
                 self._handle_stop_criteria(request, new_token, num_tokens,
                                            beam_idx)
                 request.py_decoding_iter += 1
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -1401,7 +1401,7 @@ class PyExecutor:
             if req.is_disagg_generation_transmission_complete:
                 req.state = LlmRequestState.GENERATION_IN_PROGRESS
                 req.context_current_position = req.prompt_len
-                req.decoding_iter = 1
+                req.py_decoding_iter = 1
                 first_gen_tokens = req.context_phase_params.first_gen_tokens
                 req.py_draft_tokens = req.context_phase_params.draft_tokens
                 beam_width = req.sampling_config.beam_width
@@ -1820,6 +1820,15 @@ class PyExecutor:
             if request.is_dummy == True:
                 requests_to_terminate.append(request)
                 continue
+
+            # Unify the behavior of overlapping and non-overlapping cases
+            # return the response after we have second generated token
+            if self.kv_cache_transceiver is not None:
+                if request.state == LlmRequestState.GENERATION_IN_PROGRESS and \
+                    request.is_generation_only_request:
+                    if request.py_decoding_iter == 1:
+                        new_active_requests.append(request)
+                        continue
 
             request.draft_tokens = request.py_draft_tokens
             request.decoding_iter = request.py_decoding_iter
--- a/tensorrt_llm/executor/result.py
+++ b/tensorrt_llm/executor/result.py
@@ -179,7 +179,7 @@ class GenerationResultBase:
         # Skip output the first generated token in generation response
         # TODO: We should have a better way to handle this when enable
         # beam search with PD.
-        if not self.sampling_params.use_beam_search and \
+        if self.disaggregated_params is not None and \
             len(response_tensors.output_token_ids[src_idx]) == 2:
             output._last_token_ids_len = 1
 
@@ -352,10 +352,12 @@ class GenerationResult(GenerationResultBase):
         executor (GenerationExecutor, optional): The executor that created this result. Defaults to None.
     '''
 
-    def __init__(self,
-                 generation_request: "GenerationRequest",
-                 background_error_handler: Optional[Callable] = None,
-                 executor: Optional["GenerationExecutor"] = None) -> None:
+    def __init__(
+            self,
+            generation_request: "GenerationRequest",
+            background_error_handler: Optional[Callable] = None,
+            executor: Optional["GenerationExecutor"] = None,
+            disaggregated_params: Optional[DisaggregatedParams] = None) -> None:
         super().__init__(
             generation_request.id,
             generation_request.sampling_params,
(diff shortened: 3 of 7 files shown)


tensorrt_llm/_torch/pyexecutor/decoder.py
```diff
@@ -280,6 +280,7 @@ def update_requests(self, scheduled_requests: ScheduledRequests,
             if request.state != LlmRequestState.GENERATION_COMPLETE:
                 new_token = new_tokens_list[idx]
                 num_tokens = request.add_new_token(new_token, beam_idx)
+                request.decoding_iter += 1
```
kaiyux, line 284: FYI - Seems that this is also being addressed here: https://github.com/NVIDIA/TensorRT-LLM/pull/3240, I'll take a closer look tomorrow.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1816,6 +1816,13 @@ def _handle_responses(self):
                 requests_to_terminate.append(request)
                 continue
 
+            # Unify the behavior of overlapping and non-overlapping cases
+            # return the response after we have second generated token
```
Tabrizian, line 1825:
```
import argparse

from tensorrt_llm import SamplingParams, DisaggregatedParams
from tensorrt_llm._torch import LLM
from tensorrt_llm._torch.pyexecutor.config import PyTorchConfig


def parse_arguments():
    parser = argparse.ArgumentParser()
    parser.add_argument('--model_dir',
                        type=str,
                        default='/code/llm-models/llama-3.1-model/Llama-3.1-8B-Instruct/')
    parser.add_argument('--tp_size', type=int, default=1)
    parser.add_argument('--enable_overlap_scheduler',
                        default=False,
                        action='store_true')
    parser.add_argument('--enable_chunked_prefill',
                        default=False,
                        action='store_true')
    parser.add_argument('--kv_cache_dtype', type=str, default='auto')
    args = parser.parse_args()
    return args


def main():
    args = parse_arguments()

    pytorch_config = PyTorchConfig(
        enable_overlap_scheduler=args.enable_overlap_scheduler,
        kv_cache_dtype=args.kv_cache_dtype)
    llm = LLM(model=args.model_dir,
              tensor_parallel_size=args.tp_size,
              enable_chunked_prefill=args.enable_chunked_prefill,
              pytorch_backend_config=pytorch_config)

    prompts = [
        "Hello, my name is",
        "The president of the United States is",
        "The capital of France is",
        "The future of AI is",
    ]
    sampling_params = SamplingParams(max_tokens=1)
    disaggregated_params = DisaggregatedParams(request_type="context_only")

    outputs = llm.generate(prompts, sampling_params, disaggregated_params= disaggregated_params)

    # Print the outputs.
    for output in outputs:
        prompt = output.prompt
        generated_text = output.outputs[0].text
        print(f"Prompt: {prompt!r}, Generated text: {generated_text!r}")


if __name__ == '__main__':
    main()
```
Why do we need the second generated token? I tried the snippet above but see only one token generated for context.

Tabrizian: Synced offline with @Shunkangz. I now better understand the issue. Is it possible to add a unit test for this?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1816,6 +1816,13 @@ def _handle_responses(self):
...
+            if self.kv_cache_transceiver is not None:
```
Tabrizian, line 1826: Instead of this condition can we skip it if disagg context transfer is in progress and the request is generation only?

tensorrt_llm/_torch/pyexecutor/decoder.py
kaiyux: Nvm - I see the difference now.

tensorrt_llm/_torch/pyexecutor/py_executor.py
Shunkangz: Thank you Iman. I will add a test to check the double first token issue.

tensorrt_llm/_torch/pyexecutor/py_executor.py
Shunkangz: I think that there is still a gap between the two cases. The case above is that we already finish the cache transfer and forward calculation. However, due to the async decode logic, we will add the new generated token in the next iteration which is not included in the current _handle_response

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1800,6 +1800,14 @@ def _handle_responses(self):
...
+                if request.state == LlmRequestState.GENERATION_IN_PROGRESS:
+                    if request.py_decoding_iter == 1:
+                        new_active_requests.append(request)
+                        continue
```
Tabrizian: Can we modify the condition [here](https://github.com/NVIDIA/TensorRT-LLM/blob/f3237e52ed5359f4e8ce9d396f183329f6460913/cpp/tensorrt_llm/batch_manager/llmRequest.cpp#L28) to return a nullptr when the request is generation only but has created only one token? I think that might be cleaner.

pcastonguay: Iman, I also think this would be cleaner, but we might need to change C++ disaggregated examples, so that we extract tokens from context response. It doesn't look like we're doing that right now:

https://github.com/NVIDIA/TensorRT-LLM/blob/main/examples/cpp/executor/executorExampleDisaggregated.cpp#L259-L266

https://github.com/NVIDIA/TensorRT-LLM/blob/main/benchmarks/cpp/disaggServerBenchmark.cpp

Shunkangz: Thanks for the suggestion. I will modify this.

Tabrizian (approved):

tensorrt_llm/executor/result.py
```diff
@@ -179,7 +179,7 @@ def _handle_sequence(self, finish_reasons, response_tensors,
         # Skip output the first generated token in generation response
         # TODO: We should have a better way to handle this when enable
         # beam search with PD.
-        if not self.sampling_params.use_beam_search and \
+        if self.disaggregated_params is not None and \
             len(response_tensors.output_token_ids[src_idx]) == 2:
             output._last_token_ids_len = 1
```
pcastonguay, line 184: I don't think we should rely on the len of `output_token_ids` since for spec decoding, we could have 2 tokens even after the first gen token. Can you have a look at: https://github.com/NVIDIA/TensorRT-LLM/pull/3427. I think it's a more general fix.

Shunkangz: Thank you. I also think that should be a good solution. I will close this PR.

PR #3243, opened 2025-04-02
feat: Add option to run disaggregated serving without ctx servers,…

pcastonguay:
Adds an option to run disaggregated serving without ctx servers. This can be done by setting environment variable `TRTLLM_DISAGG_BENCHMARK_GEN_ONLY` to 1. In that case, the KV cache for the prefill tokens is left uninitialized for the generation phase. This can be used to benchmark generation only latency/throughput.

### DIFF
--- a/tensorrt_llm/serve/openai_disagg_server.py
+++ b/tensorrt_llm/serve/openai_disagg_server.py
@@ -3,6 +3,7 @@ import asyncio
 import copy
 import json
 import logging
+import os
 import signal
 from contextlib import asynccontextmanager
 from http import HTTPStatus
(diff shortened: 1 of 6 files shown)


tensorrt_llm/serve/openai_disagg_server.py
```diff
@@ -39,8 +40,11 @@ def __init__(self,
         self.ctx_server_idx = 0
         self.gen_server_idx = 0
 
-        if (len(self.ctx_servers) == 0) or (len(self.gen_servers) == 0):
-            raise ValueError("At least one context server and one generation server must be provided")
+        if (len(self.gen_servers) == 0):
+            raise ValueError("At least one generation server must be provided")
+
+        if os.getenv("TRTLLM_DISAGG_BENCHMARK_GEN_ONLY") != "1" and len(ctx_servers) == 0:
```
Tabrizian, line 46: I think we need to provide a default value here to make sure the code doesn't break when the env is not set.

Tabrizian (approved):

qiaoxj07 (approved):

PR #3132, opened 2025-03-27
chore: cleanup py_executor code

achartier:
* Add common loop cleanup function
* Remove checks for attention DP if nothing to queue
* Remove extra return statements
* Remove extra variables
* Remove commented debug print

PR #3244, opened 2025-04-02
chore: Raise error for PP + MTP

amukkara:
1. Raise error for PP + MTP config that is not supported yet
2. Revert the assertion check for attention DP + PP added in https://github.com/NVIDIA/TensorRT-LLM/pull/3132

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -513,11 +513,12 @@ class PyExecutor:
                     can_queue = 0 not in tp_batch_sizes
                 else:
                     can_queue = scheduled_batch.batch_size > 0
+                    if not can_queue:
+                        assert len(self.inflight_req_ids) > 0, (
+                            "fail to schedule any pending request, probably run out of resource"
+                        )
 
                 if not can_queue:
-                    assert len(self.inflight_req_ids) > 0, (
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -513,11 +513,12 @@ def _executor_loop_pp(self):
                     can_queue = 0 not in tp_batch_sizes
                 else:
                     can_queue = scheduled_batch.batch_size > 0
+                    if not can_queue:
```
achartier, line 516: Why are we disabling the check for attention DP?

amukkara: consider TP2PP2 case with attention DP at execution start with 2 requests arriving, one per iteration.

iter=0, microbatch=0: rank=2 schedules req_id=0, rank=3 schedules one dummy request.
iter=1, microbatch=1: rank=2 does not schedule a dummy request since req_id=0 is inflight, rank=3 schedules req_id=1. can_queue = False and this microbatch is skipped for rank 2, 3. But there are no inflight requests for rank 3, assertion fails.
iter=2, microbatch=0: rank=2 schedules req_id=0, rank=3 schedules req_id=1

achartier: Makes sense, thanks for the explanation!

achartier (approved):

PR #3245, opened 2025-04-02
feat: Enable serialization and deserialization of AutoTuner cache on disk.

hyukn:
Add a simple dump option to achieve a deterministic tuning cache.


SimengLiu-nv:
Is the autotuner aware of MNMG?

PR #3246, opened 2025-04-02
chore: Reenabling test_llm get_stats_async test, which was fixed by another change to test_llm.py

pcastonguay:
Reenable get_stats_async test with overlap enabled. Doesn't fail locally anymore. Seems to have been fixed by adding a sleep here: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/test_llm.py#L1814-L1815


hchings (approved):
Thanks Patrice. If you've made sure this is not an intermittent issue, then re-enabling this LGTM.

schetlur-nv (approved):

PR #3247, opened 2025-04-02
feat: Support gemma-3-1b-it

brb-nv:
This MR adds model support for `gemma-3-1b-it`.

Details:
* This model has attn layers of two types - one with sliding-window attn with 10k rope base and another with regular attn with 1 million rope base.
* Most models in TRTLLM seem to be working with attention layers with attn layers of same type. Hence, there's only one set of RoPE params under `AttentionParams` that is passed to model's forward pass.
* However, in this case, we need two sets of RoPE params (one for each layer type mentioned above).
* Changes in this MR are adding one more set of RoPE fields under `AttentionParams` with a `_local` suffix to handle the additional layer type.
* Alternative is to copy quite some functionality from `modeling_utils.py` (`DecoderLayerList` and `DecoderModelForCausalLM`) and carefully orchestrate forward pass by passing a different set of `AttentionParams` to each layer type. I felt that's quite some code duplication and maintenance overhead.
* Sliding window requirement is mentioned to the model at runtime using `max_attention_window_size` with `run.py` or whatever API is being used.
[512, 512, 512, 512, 512, 2048, 512, 512, 512, 512, 512, 2048, 512, 512, 512, 512, 512, 2048, 512, 512, 512, 512, 512, 2048, 512, 512]
```
$ rm -rf gemma3_1b_ckpt/ && python3 ./examples/gemma/convert_checkpoint.py --ckpt-type hf --model-dir ../random/hf_models/gemma-3-1b-it/ --dtype float16 --output-model-dir gemma3_1b_ckpt/
$ rm -rf gemma3_1b_eng/ && trtllm-build --checkpoint_dir gemma3_1b_ckpt/ --output_dir gemma3_1b_eng/trtllm_engine --max_batch_size 256 --max_seq_len 32768 --max_num_tokens 32768 --workers 1 --use_paged_context_fmha disable --gpus_per_node 1 --nccl_plugin auto
$ rm -rf tllm_debug/ && python3 examples/run.py --max_output_len 512 --max_input_length 2048 --input_text 'The main cities in Italy are (Write a blog post)' --engine_dir gemma3_1b_eng/trtllm_engine --tokenizer_dir ../random/hf_models/gemma-3-1b-it/ --end_id 106
```

### DIFF
--- a/tensorrt_llm/layers/attention.py
+++ b/tensorrt_llm/layers/attention.py
@@ -175,6 +175,13 @@ class AttentionParams(object):
         self.embed_positions = None
         self.rotary_inv_freq = None
         self.embed_positions_for_gpt_attention = None
+
+        # auxiliary params to support models with non-homegeneous attn layers requiring
+        # a different set of rope params. e.g. Gemma3.
+        self.embed_positions_local = None
+        self.rotary_inv_freq_local = None
+        self.embed_positions_for_gpt_attention_local = None
+
         # long rope const parameters
         self.long_rope_embed_positions = None
         self.long_rope_rotary_inv_freq = None
@@ -186,10 +193,16 @@ class AttentionParams(object):
             self,
             embed_positions: Tensor = None,
             rotary_inv_freq: Tensor = None,
-            embed_positions_for_gpt_attention: Tensor = None):
+            embed_positions_for_gpt_attention: Tensor = None,
+            embed_positions_local: Tensor = None,
+            rotary_inv_freq_local: Tensor = None,
+            embed_positions_for_gpt_attention_local: Tensor = None):
         self.embed_positions = embed_positions
         self.rotary_inv_freq = rotary_inv_freq
         self.embed_positions_for_gpt_attention = embed_positions_for_gpt_attention
+        self.embed_positions_local = embed_positions_local
+        self.rotary_inv_freq_local = rotary_inv_freq_local
+        self.embed_positions_for_gpt_attention_local = embed_positions_for_gpt_attention_local
         return self
 
     def fill_attention_const_params_for_long_rope(
@@ -359,6 +372,7 @@ class Attention(Module):
                  dtype=None,
                  position_embedding_type=PositionEmbeddingType.learned_absolute,
                  rotary_embedding_base=10000.0,
+                 rotary_embedding_base_local=1.0,
                  rotary_embedding_scaling=None,
                  rotary_embedding_percentage=1.0,
                  rope_scaling_short_factors=None,
@@ -388,7 +402,8 @@ class Attention(Module):
                  cp_size=1,
                  cp_rank=0,
                  max_seqlen_for_logn_scaling=8192,
-                 use_logn_scaling=False):
+                 use_logn_scaling=False,
+                 is_local=False):
         super().__init__()
 
         self.local_layer_idx = local_layer_idx
@@ -417,6 +432,7 @@ class Attention(Module):
         self.cp_group = cp_group
         self.cp_size = cp_size
         self.cp_rank = cp_rank
+        self.is_local = is_local
 
         self.num_layers = num_layers
         self.apply_query_key_layer_scaling = apply_query_key_layer_scaling
@@ -437,6 +453,7 @@ class Attention(Module):
         self.max_distance = max_distance
         self.num_buckets = num_buckets
         self.rotary_embedding_base = rotary_embedding_base
+        self.rotary_embedding_base_local = rotary_embedding_base_local
         self.rotary_embedding_scaling = rotary_embedding_scaling
         self.rotary_embedding_scale_type = RotaryScalingType.none
         self.rotary_embedding_scale = 1.0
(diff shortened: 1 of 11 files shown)


examples/utils.py
```diff
@@ -337,7 +337,11 @@ def add_common_args(parser):
     parser.add_argument(
         '--max_attention_window_size',
         type=int,
-        default=None,
+        default=[
+            512, 512, 512, 512, 512, 2048, 512, 512, 512, 512, 512, 2048, 512,
+            512, 512, 512, 512, 2048, 512, 512, 512, 512, 512, 2048, 512, 512
+        ],
+        # default=None,
```
amukkara: why set the default to gemma-3 attention window values?

tensorrt_llm/layers/attention.py
```diff
@@ -677,6 +692,33 @@ def create_attention_const_params(model_cls, config):
                           dtype='float32',
                           is_buffer=True))
 
+            rotary_embedding_base_local = getattr(config,
```
amukkara: this code block is exactly the same as above but with a different rope base frequency. is it possible to put this block in a class function or lambda and invoke it with two different frequencies?

tensorrt_llm/models/gemma/config.py
```diff
@@ -85,17 +90,26 @@ def __init__(
...
             self.final_logit_softcapping = final_logit_softcapping
             self.attn_logit_softcapping = attn_logit_softcapping
+        elif self.is_gemma_3:
+            self.inter_layernorms = True
+            assert query_pre_attn_scalar is not None, "Gemma3 models must configure `query_pre_attn_scalar`"
+            self.query_pre_attn_scalar = query_pre_attn_scalar
+            self.final_logit_softcapping = final_logit_softcapping
+            self.sliding_window_pattern = sliding_window_pattern
+            self.rope_local_base_freq = rope_local_base_freq
+            self.sliding_window = sliding_window
```
amukkara: avoid repeating the assignments common for Gemma2 and Gemma3. can refactor to:

```
self.inter_layernorms = self.is_gemma2 or self.is_gemma3
if self.is_gemma2 or self.is_gemma3:
    query_pre_attn_scalar = ...
    final_logit_softcapping = ...
if self.is_gemma2:
    ....
elif self.is_gemma3:
   ....
```

tensorrt_llm/layers/attention.py
```diff
@@ -175,6 +175,11 @@ def __init__(self,
         self.embed_positions = None
         self.rotary_inv_freq = None
         self.embed_positions_for_gpt_attention = None
+
+        self.embed_positions_local = None
+        self.rotary_inv_freq_local = None
+        self.embed_positions_for_gpt_attention_local = None
+
```
amukkara, line 184: why `_local` is used as suffix specifically? does it come from Gemma naming convention?

tensorrt_llm/models/modeling_utils.py
```diff
@@ -1027,6 +1040,8 @@ def forward(self,
             else:
                 assert False, "Context parallelism with non-remove-padding is not supported yet."
 
+        is_gemma_3_cg = self.config.has_config_group(Gemma3ConfigGroup)
+
```
amukkara, line 1045: you could define `is_gemma_2_cg` similarly and use it below.

tensorrt_llm/quantization/quantize_by_modelopt.py
```diff
@@ -905,9 +906,9 @@ def quantize_and_export(*,
                 with open(f"{export_path}/config.json", "r") as f:
                     tensorrt_llm_config = json.load(f)
                 tensorrt_llm_config["mapping"]["cp_size"] = cp_size
+                tensorrt_llm_config["mapping"]["world_size"] *= cp_size
                 tensorrt_llm_config["mapping"]["attn_tp_size"] = -1
                 tensorrt_llm_config["mapping"]["attn_cp_size"] = -1
-                tensorrt_llm_config["mapping"]["world_size"] *= cp_size
```
amukkara: revert unnecessary reordering.

tensorrt_llm/quantization/quantize_by_modelopt.py
```diff
@@ -142,6 +142,7 @@ def model_type_is_enc_dec(model_type):
     "QWen": "qwen",
     "Qwen2VLForConditionalGeneration": "qwen2_vl",
     "RecurrentGemma": "recurrentgemma",
+    "Gemma3": "gemma3",
```
amukkara, line 145: keeping names sorted in preferred, add gemma3 after gemma2.

bebilli:
Does it support Gemma-3-27B?

tensorrt_llm/layers/attention.py
brb-nv: That's right: https://huggingface.co/google/gemma-3-1b-it/blob/main/config.json#L28
I'm open to making it more generic (like auxiliary) so that any model which needs a mix of 2 types could use the same fields without confusion.

tensorrt_llm/quantization/quantize_by_modelopt.py
brb-nv: I'm avoiding moving Gemma up because of how early return is being used here, as you might know:
https://github.com/NVIDIA/TensorRT-LLM/blob/41be90e1fa002d84b1a3d541c4b5b537e917f5e3/tensorrt_llm/quantization/quantize_by_modelopt.py#L364-L366

tensorrt_llm/models/modeling_utils.py
brb-nv: Done. https://github.com/NVIDIA/TensorRT-LLM/pull/3247/commits/2ff1414eb346d0f23732beffc08e8fce9630ddac

tensorrt_llm/layers/attention.py
amukkara: that's fine. we can it as `_local` in this PR and reconsider later.

tensorrt_llm/quantization/quantize_by_modelopt.py
amukkara: my initial suggestion was for `RecurrentGemma`, 'Gemma2', 'Gemma3', 'Gemma'. but that is as badly sorted as your version :) we can skip this.

amukkara:
@brb-nv can you update `examples/gemma/README.md` with gemma3 instructions. specifically, clearly show instructions to set up the correct attention window values.

ideally, the attention window values should be part of the model definition so that users do not have to provide these values. we can add that in a follow-up PR.

examples/utils.py
brb-nv: Removed.

tensorrt_llm/layers/attention.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/3247/commits/d8e2ef42dab59c75de7e6064135304e1f63cb198

tensorrt_llm/quantization/quantize_by_modelopt.py
brb-nv: Yeah, that ordering is intentional so that `RecurrentGemma` model doesn't get recognized as `Gemma` model if we go with ideal alphabetical ordering:
'Gemma', 'Gemma2', 'Gemma3', 'RecurrentGemma'.

I admit my adjustment makes it look worse. Will undo. Thank you!

brb-nv: Done. https://github.com/NVIDIA/TensorRT-LLM/pull/3247/commits/ad0e30ef325cc1b5f053cb8a15a1e2d40fc5f6c8

tensorrt_llm/quantization/quantize_by_modelopt.py
brb-nv: Done. https://github.com/NVIDIA/TensorRT-LLM/pull/3247/commits/2ff1414eb346d0f23732beffc08e8fce9630ddac

tensorrt_llm/layers/attention.py
brb-nv: Sounds good. Leaving it as is.

tensorrt_llm/models/gemma/config.py
brb-nv: Done. https://github.com/NVIDIA/TensorRT-LLM/pull/3247/commits/2ff1414eb346d0f23732beffc08e8fce9630ddac

brb-nv:
> Does it support Gemma-3-27B?

Hi, goal is to add support for text generation model first. We'll get to the multimodal models in a follow-up MR.

amukkara (approved):

brb-nv:
> @brb-nv can you update `examples/gemma/README.md` with gemma3 instructions. specifically, clearly show instructions to set up the correct attention window values.
> 
> ideally, the attention window values should be part of the model definition so that users do not have to provide these values. we can add that in a follow-up PR.

Updated here, Anurag. https://github.com/NVIDIA/TensorRT-LLM/pull/3247/commits/1ad80eaac010c6b751695d46ecb94072e24a7d18

chzblych (approved):
Approving this MR for merging.

PR #3250, opened 2025-04-03
feat: add trtllm stress test

dominicshanshan:


tests/integration/test_lists/test-db/l0_a10.yml
```diff
@@ -22,14 +27,14 @@ l0_a10:
   - test_cpp.py::test_model[recurrentgemma-86]
   - test_cpp.py::test_model[eagle-86]
   # ------------- TRT tests ---------------
-  - unittest/trt/attention/test_gpt_attention.py -k "partition0"
```
tongyuantongyu: There shouldn't be changes to this file other than addition of `stress_test/stress_test.py` tests.

dominicshanshan: Let me update

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+
+
+@dataclass(frozen=True)
+class PerformanceParams:
+    """Dataclass to store test parameters for genai-perf"""
+    input_len_mean: int = 64
+    input_len_std: int = 16
+    output_len_mean: int = 128
+    output_len_std: int = 32
+    test_timeout: int = 3600  # 1 hour
```
litaotju: whats the differences of this test_timeout value with the other 2 timeout values in the StressTestConfig config?

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+        yield process
+    finally:
+        if process:
+            print_info(f"Stopping process with PID: {process.pid}")
+            if process.poll() is None:
+                # Send termination signal
+                process.terminate()
+
+                try:
+                    process.wait(timeout=GRACEFUL_TERMINATION_TIMEOUT)
```
litaotju, line 237: This terminate will just exit the process with finishing to on-the-fly tokens or not?
If all the token generation has to be finished, 4 seconds may not be enough for long osl cases.

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+    # Start server
+    print_info("Starting trtllm-serve server...")
+    print_info(f"Model path: {model_path}")
+
+    # Verify that model path exists
+    if not os.path.exists(model_path):
+        raise RuntimeError(f"Model path does not exist: {model_path}")
+
+    # For tp_size > 1, use multiple GPUs
+    gpu_list = ",".join(str(i) for i in range(config.tp_size))
```
litaotju: I suggest to not set this inside the test case.
And let outside world to set it.

Setting it to 0~n-1 here equals to not set it actually.

Sometimes QA or other dev want to set GPUs if their server has multiple different GPUs, like both A100 and H100.

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+    # backend = None means trt backend
+    # backend = pytorch means pytorch backend
+    if config.backend:
+        server_cmd.extend(["--backend", config.backend])
+
+    # Log the command we're about to run
+    print_info(f"Running command: {' '.join(server_cmd)}")
+
+    # Start server with the launch_process context manager and filtered output
+    with altered_env(CUDA_VISIBLE_DEVICES=gpu_list):
```
litaotju: If this can be avoided, lets not set this inside the case.

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+
+
+@dataclass(frozen=True)
+class ServerConfig:
+    """Dataclass to store server configuration for trtllm-serve"""
+    port: int = 8000
+    host: str = "0.0.0.0"
+    wait_interval: int = 10  # seconds
+    max_wait_seconds: int = 360  # 6 mins <- Larger model need longer wait time
+    health_check_timeout: float = 8  # seconds <- Make it smaller than wait_interval
```
litaotju, line 76: 8 seconds seems very optimistic. Considering load models from NFS, can be really slow in the first time. 

Maybe not set this, have a env to control this?

tests/integration/test_lists/test-db/l0_a10.yml
```diff
@@ -9,19 +9,52 @@ l0_a10:
...
+      system_gpu_count:
+        gte: 1
+        lte: 1
+    wildcards:
+      gpu:
+      - '*a10*'
+      linux_distribution_name: ubuntu*
+    terms:
+      stage: pre_merge
+      backend: cpp
```
litaotju: I am not sure. Should this be "tensorrt"? or "cpp"? is "backend: cpp" is valid backend here in the yml file context?

tests/integration/defs/stress_test/stress_test.py
dominicshanshan: I will add each timeout variable meaning in code as comment. 

The `stress_timeout` and `customized_stress_timeout` is maximum time allowed for the stress test/customized stress test to complete, it used to prevent stress test hang situations, include process initialization, warmup and all kinds of delays, so to set those values greater than `stress_time` and `customized_stress_time` accordingly. 


The `test_timeout` is maximum time allowed for the entire performance test phase to complete, it prevents each concurrency test took longer time.

tests/integration/defs/stress_test/stress_test.py
dominicshanshan: It first wait for process kill signal, e.g., trtllm-serve or genai-perf, then wait 4 seconds for the process to exit cleanly, if not then will be forced to kill the process and its subprocess cleanly (use `cleanup_process_tree`), I know your point, 4 seconds is good enough for tinyllama case, for larger model like v3 we can adjust this value based some pre-run experiments.

tests/integration/defs/stress_test/stress_test.py
dominicshanshan: Agree, let me figure out this

tests/integration/defs/stress_test/stress_test.py
dominicshanshan: Agree, will change

tests/integration/defs/stress_test/stress_test.py
dominicshanshan: This is purely periodically check if serve is ready and could let us know exactly how long time model will be ready after loading from NFS, etc. And can be changed ad-hoc after pre-run.

tests/integration/test_lists/test-db/l0_a10.yml
dominicshanshan: cpp backend  is not used in our stress test case.

tests/integration/defs/stress_test/stress_test.py
tongyuantongyu: > with finishing to on-the-fly tokens

This piece of code is to stop `trtllm-serve` after all test finished or something went wrong, which means there's either no on-the-fly requests or we merely want to clean up all processes in whatever manner.

> may not be enough for long osl cases

`genai-perf` will wait all tokens to be generated before quit, so when reaching this line, `trtllm-serve` has completed generating all tokens for all requests, and is idling.

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+
+# Define a constant for process termination timeouts
+GRACEFUL_TERMINATION_TIMEOUT = 4  # seconds
+
+
+@dataclass(frozen=True)
+class ServerConfig:
+    """Dataclass to store server configuration for trtllm-serve"""
+    port: int = 8000
+    host: str = "0.0.0.0"
```
joyang-nv: Can you try to use `localhost` instead? Reason is I am not sure running env, if it shared the host interface, listening all interfaces might have higher chance for conflict.

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+                except subprocess.TimeoutExpired:
+                    # Process didn't exit within timeout, force kill
+                    print_warning(
+                        "Process did not terminate gracefully, force killing..."
+                    )
+                    cleanup_process_tree(process, has_session=True)
+                    print_info("Process killed forcefully")
+
+            # Wait for reader threads to finish
+            if stdout_reader and stdout_reader.is_alive():
```
joyang-nv: Please use `stdout_reader is None` to avoid ambiguity.  Similar to `stderr_reader`

joyang-nv: You won't need to wait for threads.
As you had implemented inside threading function, broken pipe should appear after subprocess is done. At that time, thread function will exit themselves.
It is pretty save to remove checking and join.

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+
+    # Get a list of directory names in the artifacts directory
+    directories = [
+        d for d in os.listdir(artifacts_dir)
+        if os.path.isdir(os.path.join(artifacts_dir, d))
+    ]
+
+    # Extract model names from directory names (before "-openai-completions-")
+    model_name_map = {}
+    for directory in directories:
```
joyang-nv, line 860: Considering local test case, I think this could be refined.
1. Look at all possible matches.
2. Sort by created time despondingly. 
3. If all matches count is 1, go ahead with additional info. Otherwise, pick up latest one and log info about other folders.

tests/integration/defs/stress_test/stress_test.py
```diff
@@ -0,0 +1,941 @@
...
+    # Create a DataFrame to store the metrics
+    metrics_df = pd.DataFrame({
+        "Model": model_name,
+        "Concurrency": concurrency,
+        "RequestThroughput": request_throughput,
+        "OutputTokenThroughput": output_token_throughput
+    })
+
+    # Sort by Model, Concurrency, and RequestThroughput
+    metrics_df = metrics_df.sort_values(
```
joyang-nv, line 943: I would like to propose another idea as P1.
https://stackoverflow.com/questions/37288421/how-to-plot-a-chart-in-the-terminal Look at this terminal plot repo, in our future change, I could like draw the plot which I think most important facts about stress testing in terminal.

joyang-nv (changes requested):
Overall looks good. Just a few comments.
@dominicshanshan nice job.

tests/integration/defs/stress_test/stress_test.py
dominicshanshan: Agree, will change

tests/integration/defs/stress_test/stress_test.py
dominicshanshan: Agree, will remove this check

joyang-nv (approved):

PR #3257, opened 2025-04-03
feat: Introduce UB allocator for pytorch flow

liji-nv:
* Instead of allocating UserBuffers at beginning of runtime, UB buffers are now managed with global allocator. The allocator will dynamically assign free UB buffer or allocate new buffer for torch tensor. It makes userbuffers easier to use.

* In common usecase, the Userbuffers will be allocated correctly during warm up stage. There is no dynamic allocation during inference.

* UB fusion pattern is rewroten using the new UB Allocator. It contains following passes:

1. Fuse Quant with allreduce, replace with UB impl, and insert a copy_to_userbuffers. Currently the normal allreduce still does not support FP8 quant. So this need to be done in UB pass
2. Convert all supported allreduce with UB and insert copy_to_userbuffers.
3. Fuse op before ar with the copy_to_userbuffers. So the op directly writes to the userbuffer
4. Remove userbuffers finalize if the output is connect to another UB allreduce.

### DIFF
--- a/cpp/tensorrt_llm/kernels/userbuffers/userbuffersManager.cpp
+++ b/cpp/tensorrt_llm/kernels/userbuffers/userbuffersManager.cpp
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "userbuffersManager.h"
+
+namespace tensorrt_llm::runtime::ub
+{
+
+void UserBufferDeleter::operator()(void* ptr)
+{
+    UserBuffersManager::get_instance().release_buffer(ptr);
+}
+
+UserBuffersManager& UserBuffersManager::get_instance()
+{
+    static UserBuffersManager allocator;
+    return allocator;
+}
+
+void UserBuffersManager::initialize(
+    int64_t tp_size, int64_t pp_size, int64_t cp_size, int64_t rank, int64_t gpus_per_node, int64_t buffer_size)
+{
+    std::lock_guard<std::mutex> lock(mutex_);
+    tensorrt_llm::runtime::WorldConfig world_config(tp_size, pp_size, cp_size, rank, gpus_per_node);
+    tensorrt_llm::runtime::ub::ub_initialize(world_config);
+    TLLM_CHECK(tensorrt_llm::runtime::ub::ub_is_initialized());
+    buffer_size_ = buffer_size;
+}
+
+std::pair<UBBufferPtr, UBBuffer> UserBuffersManager::allocate_userbuffers(int64_t buffer_size)
+{
+    std::lock_guard<std::mutex> lock(mutex_);
+    TLLM_CHECK(buffer_size <= buffer_size_);
+
+    // Check for all unused buffers
+    int i = 0;
+    for (auto& buffer : buffers_)
+    {
+        if (buffer.second)
+        {
+            i++;
+            continue;
+        }
+        buffer.second = true;
+        TLLM_LOG_DEBUG("Reusing buffer %d", i);
+        return std::make_pair(std::unique_ptr<void, UserBufferDeleter>(buffer.first.addr), buffer.first);
+    }
+
+    auto new_ub = tensorrt_llm::runtime::ub::ub_allocate(buffer_size_);
+    TLLM_CHECK(!new_ub.invalid());
+    buffers_.push_back({new_ub, true});
+    TLLM_LOG_DEBUG("Creating new buffer %d", static_cast<int>(buffers_.size() - 1));
+    return std::make_pair(std::unique_ptr<void, UserBufferDeleter>(new_ub.addr), new_ub);
+}
+
+void UserBuffersManager::release_buffer(void* addr)
+{
+    std::lock_guard<std::mutex> lock(mutex_);
+    auto buffer_iter = std::find_if(
+        buffers_.begin(), buffers_.end(), [addr](auto const& buffer) { return buffer.first.addr == addr; });
+    // The UB should be assigned to a tensor
+    TLLM_CHECK(buffer_iter != buffers_.end());
+    TLLM_CHECK(buffer_iter->second);
+    TLLM_CHECK(!buffer_iter->first.invalid());
+    TLLM_LOG_DEBUG("Releasing buffer %d", static_cast<int>(std::distance(buffers_.begin(), buffer_iter)));
+    buffer_iter->second = false;
+}
+
+tensorrt_llm::runtime::ub::UBBuffer UserBuffersManager::search_buffer(void* addr)
+{
+    std::lock_guard<std::mutex> lock(mutex_);
+    auto buffer_iter = std::find_if(
+        buffers_.begin(), buffers_.end(), [addr](auto const& buffer) { return buffer.first.addr == addr; });
+    if (buffer_iter == buffers_.end())
+    {
+        return tensorrt_llm::runtime::ub::UBBuffer();
+    }
+    return buffer_iter->first;
+}
+
+tensorrt_llm::runtime::ub::communicator* UserBuffersManager::comm()
+{
+    return tensorrt_llm::runtime::ub::ub_comm();
+}
+
+void initialize_userbuffers_manager(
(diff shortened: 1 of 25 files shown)


tests/unittest/_torch/multi_gpu/test_user_buffers.py
```diff
@@ -701,3 +680,132 @@ def test_user_buffers_ar_rms_norm_fp4(mnk):
...
+                                     mm1_input_1, residual_0, residual_1)
+            torch.cuda.synchronize()
+            del output_fused
+            # Run twice
+            output_fused = model_opt(mm0_input_0, mm0_input_1, mm1_input_0,
+                                     mm1_input_1, residual_0, residual_1)
+            torch.cuda.synchronize()
+            output_ref = model(mm0_input_0, mm0_input_1, mm1_input_0,
+                               mm1_input_1, residual_0, residual_1)
+        # 3 AR_NORM fusion happens first
```
liji-nv, line 774: Comment is outdated.

tensorrt_llm/_torch/compilation/patterns/ub_allreduce.py
```diff
@@ -15,195 +15,381 @@
...
+                    input, None, [residual_in, gamma, scale], mapping.tp_group,
+                    int(AllReduceStrategy.UB), 0,
+                    int(AllReduceFusionOp.RESIDUAL_RMS_NORM_QUANT_FP8), eps,
+                    True, False, True)
+                finalize_output = torch.ops.trtllm.userbuffers_allreduce_finalize(
+                    all_reduce_output[1], False)
+                return all_reduce_output[0], scale, finalize_output
+
+            def extra_check_fp8_quant_pattern(match: Match) -> bool:
+                # Residual should be a tensor
```
liji-nv:
```suggestion
```

tensorrt_llm/_torch/custom_ops/userbuffers_custom_ops.py
```diff
@@ -1,80 +1,44 @@
...
+    ub_tensor = torch.ops.trtllm.create_userbuffers_tensor(a.shape, a.dtype)
+    return ub_tensor.copy_(a, non_blocking=True)
+
+
+@copy_to_userbuffers.register_fake
+def _(a) -> torch.Tensor:
+    return torch.empty_like(a)
+
+
+# Custom ops below are used to avoid in-place operation to torch fx graph.
```
liji-nv:
```suggestion
# Custom ops below are used to avoid in-place operation added torch fx graph.
```

cpp/tensorrt_llm/kernels/userbuffers/userbuffersManager.cpp
```diff
@@ -0,0 +1,105 @@
...
+        }
+        buffer.second = true;
+        std::cout << "Reusing buffer " << i << std::endl;
+        return std::make_pair(std::unique_ptr<void, UserBufferDeleter>(buffer.first.addr), buffer.first);
+    }
+
+    auto new_ub = tensorrt_llm::runtime::ub::ub_allocate(buffer_size_);
+    TLLM_CHECK(!new_ub.invalid());
+    buffers_.push_back({new_ub, true});
+    std::cout << "Creating new buffer " << buffers_.size() - 1 << std::endl;
```
liji-nv: Turn to use TRTLLM_LOG_DEBUG

tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
```diff
@@ -231,3 +231,7 @@ def _(
 
         return (input.new_empty(output_shape, dtype=torch.uint8),
                 global_scale.new_empty(scale_shape, dtype=torch.uint8))
+
+    @torch.library.register_fake("trtllm::copy_to_userbuffers")
```
liji-nv: Need to remove it

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -5,3 +5,9 @@
 @torch.library.custom_op("trtllm::bmm_out", mutates_args=("out", ))
 def bmm_out(a: torch.Tensor, b: torch.Tensor, out: torch.Tensor) -> None:
     torch.bmm(a, b, out=out)
+
+
+# Used to WAR an issue that torch.add not allowed to use out in target pattern
```
liji-nv: Need to remove it

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -208,10 +208,11 @@ class AllreduceOp
                 || mOp == AllReduceFusionOp::RESIDUAL_RMS_NORM_QUANT_NVFP4);
             TLLM_CHECK_WITH_INFO(
                 tensorrt_llm::runtime::ub::ub_is_initialized(), "UserBuffer has not been initialized!");
-            auto ub_buffer0 = tensorrt_llm::runtime::ub::ub_get(0);
-            TLLM_CHECK(input.data_ptr() == ub_buffer0.addr);
-            auto ub_buffer1 = tensorrt_llm::runtime::ub::ub_get(1);
-            auto ub_comm = tensorrt_llm::runtime::ub::ub_comm();
+            auto& ub_allocator = tensorrt_llm::runtime::ub::UserBuffersManager::get_instance();
```
yizhang-nv: The name is quite confusing. I think ub allocator means the component under ub_allocator which is only responsible for allocating the user buffer. Here should be something like ub_manager.

cpp/tensorrt_llm/kernels/userbuffers/userbuffersManager.cpp
```diff
@@ -0,0 +1,105 @@
...
+    TLLM_CHECK(buffer_iter->second);
+    TLLM_CHECK(!buffer_iter->first.invalid());
+    TLLM_LOG_DEBUG("Releasing buffer %d", static_cast<int>(std::distance(buffers_.begin(), buffer_iter)));
+    buffer_iter->second = false;
+}
+
+tensorrt_llm::runtime::ub::UBBuffer UserBuffersManager::search_buffer(void* addr)
+{
+    std::lock_guard<std::mutex> lock(mutex_);
+    auto buffer_iter = std::find_if(
```
yizhang-nv, line 84: Should we consider using an unordered map here to store the user buffer to speed up the look-up process? (e.g. `[addr, std::pair<tensorrt_llm::runtime::ub::UBBuffer, bool>]`) So that we may save some time in looking up the address in the custom op & find a reusable buffer. 

If we do not allocate many user buffers, then I think it's fine to leave it like this.

cpp/tensorrt_llm/thop/userbuffersTensor.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "userbuffersTensor.h"
+
+namespace torch_ext
+{
+
+std::pair<torch::Tensor, tensorrt_llm::runtime::ub::UBBuffer> create_ub_tensor(
```
yizhang-nv: These two functions' names are quite confusing. Better rename them to emphasize the different values.

cpp/tensorrt_llm/kernels/userbuffers/userbuffersManager.cpp
liji-nv: Usually only 3 buffers are in use at the same time. Unless the graph contains complex sub-branch.

liji-nv: Using a vector generates a 1 to 1 map to the real allocator. And a vector can make sure after allocating, the index never change. It can make UB buffer usage analyzation easier.

cpp/tensorrt_llm/thop/userbuffersTensor.cpp
liji-nv: I will rename it to create_userbuffers_tensor and rename the one below create_userbuffers_tensor_op. The one below is only used to create a pytorch custom op so the buffer can be created on python side.

cpp/tensorrt_llm/kernels/userbuffers/userbuffersManager.cpp
yizhang-nv: I saw the searching method only uses addr as a key to find the user buffer, so the index generated by the vector seems to be equivalent to the address itself. But it seems that we only use few user buffers so it's fine to leave it like this.

cpp/tensorrt_llm/thop/allreduceOp.cpp
liji-nv: Use ub_manager now.

yizhang-nv (approved):

yilin-void (approved):
LGTM, but one question: After modifying the allocation and search mechanism of the ub buffer, will it affect the correctness of the old implementation that obtains the ub buffer through a fixed index using ub::get() in the trt flow?

liji-nv:
> LGTM, but one question: After modifying the allocation and search mechanism of the ub buffer, will it affect the correctness of the old implementation that obtains the ub buffer through a fixed index using ub::get() in the trt flow?

No. The old way still works when using the old ub interfaces. Just when allocate, need to care about the order to init the buffers.

PR #3270, opened 2025-04-03
fix: Add thread leak check and fix thread/memory leak issues

yuxianq:
This PR add thread leak check for pytest and fix known thread leak issues, which helps the CI to reduce OOM errors and raise OOM in the correct place. This PR also fixes some memory leak issues.

### DIFF
--- a/tensorrt_llm/llmapi/mpi_session.py
+++ b/tensorrt_llm/llmapi/mpi_session.py
@@ -122,7 +122,7 @@ class MpiPoolSession(MpiSession):
 
     def shutdown(self):
         if self.mpi_pool is not None:
-            self.mpi_pool.shutdown(wait=False)
+            self.mpi_pool.shutdown(wait=True)
             self.mpi_pool = None
 
     def _start_mpi_pool(self):
@@ -200,8 +200,11 @@ class MpiCommSession(MpiSession):
 
     def shutdown(self):
         if self.mpi_pool is not None:
-            self.mpi_pool.shutdown(wait=False)
+            self.mpi_pool.shutdown(wait=True)
             self.mpi_pool = None
+        if self.thread_pool is not None:
+            self.thread_pool.shutdown(wait=True)
+            self.thread_pool = None
 
     def _start_mpi_pool(self):
         assert not self.mpi_pool, 'MPI session already started'
--- a/tests/unittest/_torch/test_attention.py
+++ b/tests/unittest/_torch/test_attention.py
@@ -516,6 +516,10 @@ def test_attention_backend(s: Scenario):
         },
     )
 
+    del flashinfer_kv_cache
+    del ref_kv_cache
+    torch.cuda.empty_cache()
+
 
 def generate_causal_mask(seq_lens, qo_lens, batch_size, dtype):
     causal_masks = []
@@ -696,6 +700,11 @@ def test_attention_backend_ifb(s: PagedScenario):
                               vanilla_outputs[i],
                               atol=fp8_atol if is_fp8 else atol,
                               rtol=rtol)
+
+    del flashinfer_kv_cache
+    del ref_kv_cache
+    del vanilla_kv_cache
+    torch.cuda.empty_cache()
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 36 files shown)


ZhanruiSunCh:
Sorry for this error: [Bot-Command / Bot command check (pull_request)]
This issue has be fixed. After rebase with TOT main, it will auto passed.

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -122,7 +122,7 @@ def submit_sync(self, task: Callable[..., T], *args, **kwargs) -> List[T]:
 
     def shutdown(self):
         if self.mpi_pool is not None:
-            self.mpi_pool.shutdown(wait=False)
+            self.mpi_pool.shutdown(wait=True)
```
QiJune, line 125: cc @Superjomn to review this change

QiJune (approved):
LGTM

tensorrt_llm/llmapi/mpi_session.py
Superjomn: Yeah, I think it is fine for blocking shutdown here.

tests/unittest/_torch/multi_gpu_modeling/test_deepseek.py
```diff
@@ -332,3 +332,4 @@ async def test():
                            1.0), f"Expected '{expected}' but get '{result}'"
 
     asyncio.run(test())
+    llm.shutdown()
```
Superjomn: Why explicit shutdown needed here? Ideally, it should be triggered when the llm instance is destroyed, which occurs when it runs out of the function body. For users, we don't want them to call the shutdown(), just as our [quickstart](https://nvidia.github.io/TensorRT-LLM/quick-start-guide.html). Can we remove the shutdown calling and make the test consistent with users' usage?  So is in other places. I am a bit worried that after some time, we may merge some bugs that make the LLM quit depending on explicit shutdown() calling, but the tests cannot catch that.

tests/unittest/_torch/test_attention.py
```diff
@@ -516,6 +516,10 @@ def produce(Attention: type[AttentionBackend], kv_cache: torch.Tensor):
         },
     )
 
+    del flashinfer_kv_cache
+    del ref_kv_cache
+    torch.cuda.empty_cache()
```
Superjomn, line 521: The release_gc method could be reused here and in other places.

yuxianq: Only `flashinfer_kv_cache` and `ref_kv_cache` is leaked here, so `torch.cuda.empty_cache()` is enough. `release_gc` will trigger full GC, I avoid to use it everywhere to speedup tests and make the intention clearer.

tests/unittest/_torch/multi_gpu_modeling/test_deepseek.py
yuxianq: I add `llm.shutdown()` to every test function that does not use with block for `llm` to ensure that they can pass the CI. Let me try to remove the unnecessary ones.

yuxianq: Updated.

tests/unittest/_torch/test_attention.py
Superjomn: That makes sense. Thanks for the explanation.

tests/unittest/_torch/multi_gpu_modeling/test_deepseek.py
Superjomn: Can you remove this `llm.shutdown` here? It should be invoked automatically once out of the function block, or there may be some issue that prevents it from shutting down.

yuxianq: Yes, I have removed it and other 15 unnecessary `llm.shutdown()` in the latest update, feel free to review them: https://github.com/NVIDIA/TensorRT-LLM/pull/3270/commits/34023403720468b340fb6683da2bdee50e12cc7f.

Superjomn: I see, thanks.

Issue #3137, opened 2025-03-28
Lookahead decoding and multimodal input support


maxilevi:
Hi,

I get the following error when:
* Lookahead decoding is enabled
* Request has multimodal input (e.g.  just custom prompt table with fake vocabulary) 
* batch size > 1
* Inflight fused batching is enabled 
 
Model is Llama 8B.

```
[TensorRT-LLM][ERROR] IExecutionContext::inferShapes: Error Code 7: Internal Error (LLaMAForCausalLM/transformer/vocab_embedding/__add___L322/elementwise_binary_L2901/ELEMENTWISE_SUM_0: dimensions not compatible for elementwise. Broadcast has incompatible dimensions: 2 != 18 && 2 != 1 && 18 != 1. Instruction: CHECK_BROADCAST 2 18.)
[TensorRT-LLM][ERROR] Encountered an error in forwardAsync function: Invalid input shape (/home/jenkins/agent/workspace/LLM/release-0.17/L0_Test-x86_64/tensorrt_llm/cpp/tensorrt_llm/runtime/tllmRuntime.cpp:574)
1       0x7f4097fd7277 /home/maximilianolevi/.cache/pypoetry/virtualenvs/tensorrt-inference-8MUMp6os-py3.10/lib/python3.10/site-packages/tensorrt_llm/libs/libtensorrt_llm.so(+0x6e3277) [0x7f4097fd7277]
2       0x7f4098cadc88 tensorrt_llm::batch_manager::TrtGptModelInflightBatching::prepareBuffers(std::vector<std::shared_ptr<tensorrt_llm::batch_manager::LlmRequest>, std::allocator<std::shared_ptr<tensorrt_llm::batch_manager::LlmRequest> > > const&, std::vector<std::shared_ptr<tensorrt_llm::batch_manager::LlmRequest>, std::allocator<std::shared_ptr<tensorrt_llm::batch_manager::LlmRequest> > > const&, int) + 184
[...]

PR #3276, opened 2025-04-03
fix: #3137 speculative decoding and multimodal input support

maxilevi:
Currently lookahead decoding does not work for batch_size > 1 when multimodal input is enabled.

This is due to the PromptTuningEmbedding layer assuming prompt_tokens are always `[batch_size, seq_len]` but when speculative decoding is enabled shape will be `[batch_size, 1]` thus causing an incompatible shape error as discussed [here](https://github.com/NVIDIA/TensorRT-LLM/issues/3137)

This PR fixes the issue by broadcasting the tasks tensor to the correct shape

### DIFF
--- a/tensorrt_llm/layers/embedding.py
+++ b/tensorrt_llm/layers/embedding.py
@@ -20,8 +20,8 @@ import torch
 
 from .._utils import set_obj_attrs, str_dtype_to_torch, trt_dtype_to_np
 from ..functional import (ACT2FN, Tensor, arange, concat, constant, cos, div,
-                          embedding, exp, identity, meshgrid2d, outer, pad,
-                          shape, sin, slice, unsqueeze, where)
+                          embedding, exp, expand, identity, meshgrid2d, outer,
+                          pad, shape, sin, slice, unsqueeze, where)
 from ..mapping import Mapping
 from ..module import Module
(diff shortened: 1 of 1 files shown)


tensorrt_llm/layers/embedding.py
```diff
@@ -177,6 +177,9 @@ def forward(self, tokens, prompt_embedding_table, tasks, task_vocab_size):
 
         # tasks: [batch_size, seq_len]
         # prompt_tokens: [batch_size, seq_len]
+        # if speculative decoding is enabled tasks is [batch_size, seq_len + max_draft_len],
```
byshiue: A little confused the explanation here. Should it be
>         # if speculative decoding is enabled, the shape of prompt_tokens is [batch_size, seq_len + max_draft_len]

Otherwise, the shape of task is already `[batch_size, seq_len + max_draft_len]`, why do we need to expand it?

maxilevi: Yes, you are right

maxilevi: Pushed a fix

byshiue (approved):

lfr-0531 (approved):

PR #3282, opened 2025-04-03
infra: move nvrtc_wrapper to conan

tburt-nv:
This MR removes the nvrtc_wrapper git-lfs objects and replaces them with Conan packages, which are fetched at build-time.

Conan requires the sqlite3 python module, which is missing from the rocky8 container image, so we cannot install Conan with pip yet. Instead, we install the standalone Conan executable from the official Conan GitHub releases.


nv-guomingz (approved):
LGTM

PR #3283, opened 2025-04-04
feat:[AutoDeploy] Update MoE pattern matcher to drop expert selection logic

Fridah-nv:
- [X] Step 0: update moe pattern matcher: expect selection part is removed from detection and the algo only matches expert computation and results aggregation part
- [X] Support 2D and 3D input for trtllm fused moe op

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe.py
+++ b/tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe.py
@@ -151,6 +151,9 @@ def trtllm_fused_moe(
     w3_w1_stacked_weight: torch.Tensor,
     w2_stacked_weight: torch.Tensor,
 ) -> torch.Tensor:
+    hidden_dim = x.shape[-1]
+    x = x.view(-1, hidden_dim)
+
     routing_weights = routing_weights.to(torch.float32)
     selected_experts = selected_experts.to(torch.int32)
     quant_scales = []
--- a/tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe.py
+++ b/tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe.py
@@ -35,7 +35,9 @@ def torch_moe(
         torch.Tensor: Output tensor with the same shape as the input x.
     """
 
-    hidden_dim = x.shape[-1]
+    x_shape = x.shape
+    hidden_dim = x_shape[-1]
+    x = x.view(-1, hidden_dim)
     num_experts = len(w1_weight)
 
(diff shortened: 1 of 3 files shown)


lucaslie (commented):
I think the DeepSeek MoE patch needed extra logic because the moe ops didn't consistently support 2d/3d input. Let's update the patch as well now that both format should work.

tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe.py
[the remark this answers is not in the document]
Fridah-nv: Thank you for the suggestions, made the update.

tensorrt_llm/_torch/auto_deploy/transformations/library/fused_moe.py
```diff
@@ -19,71 +19,71 @@ def match_moe_pattern(gm: GraphModule) -> GraphModule:
...
+            attr_next="all_input_nodes",
+            boundary=start_boundary,
+        ).args[0]
+        if not selected_experts:
+            continue
+
+        # Step 3: Trace upwards to find input node:
+        hidden_states = _find_lowest_common_ancessor(pattern_input_nodes)
+        if not hidden_states:
+            continue
```
lucaslie, line 51: It seems to be for steps 2 + 3 we rely on some type of "unchecked" graph traversal. Could this go wrong? 

In other words, do you think it would be better to trace the exact pattern to identify `normalized_routing_weights`, `selected_experts` and `hidden_states` just like you to for `_find_final_hidden_state_node`?

Fridah-nv: This is a good question. When I wrote this, I hope it to also be able to capture DeepSeek MoE pattern, which is mathematically equivalent to Mitral style MoE but wrote differently and contains different ops. So I only checked the subgraph structure in step2 and 3 and ignored the ops along the branch. 
I agree that it's saver to check exact patterns for finding `normalized_routing_weights ` and `selected_experts `.  In the long run, maybe it's good to break the MoE patterns into some smaller patterns, subpatterns can capture different implementations and the structure shared by different implementations can be kept.

lucaslie: TODO: please summarize this discussion and add an issue for this to the backlog of our project board

Fridah-nv: Created a ticket for this: https://github.com/NVIDIA/TensorRT-LLM/issues/4333

lucaslie (approved):

PR #3284, opened 2025-04-04
[Draft]: Use pydantic json serialization for ZMQ socket IPC communication

yibinl-nvidia:


tensorrt_llm/executor/utils.py
```diff
@@ -124,6 +143,28 @@ class ExecutorResponse(NamedTuple):
     disaggregated_params: Optional[DisaggregatedParams] = None
 
 
+class ExecutorResponseSafe(BaseModel):
```
Superjomn, line 146: Are you trying to switch the IPC serialization methods to JSON based? That could be a larger task than the modification; AFAK, there are many DS used in IPC, ExecutorResponse, and ResponseTensors are just two of them. The remaining I know includes at least the [Response sent here](https://github.com/superjomn/TensorRT-LLM/blob/c7548ad72c5015f7d8522486f45c0a4f8124061e/tensorrt_llm/executor/worker.py#L759) which is a pybinding-based, [send request](https://github.com/superjomn/TensorRT-LLM/blob/017361c26c9d3c0425322f5fc0e04862ee3f8b4f/tensorrt_llm/executor/proxy.py#L372) which contains several complex nested DS, and some raw ZMQ usage such as [here](https://github.com/superjomn/TensorRT-LLM/blob/017361c26c9d3c0425322f5fc0e04862ee3f8b4f/tensorrt_llm/llmapi/mpi_session.py#L259), which I will help unify to ZeroMqQueue.  To get a full set of DS used in IPC, you can collect them by adding some code in the [ZeroMqQueue.put](https://github.com/superjomn/TensorRT-LLM/blob/017361c26c9d3c0425322f5fc0e04862ee3f8b4f/tensorrt_llm/executor/ipc.py#L91) method.

PR #3286, opened 2025-04-04
chore: [MLA] Deallocate tensors after use

hlu1:
There are a few tensors in MLA that were no deallocated even though they're no longer used. In pytorch model code, we should deallocate the tensors that are no longer in use to release the memory ASAP. However, this does not necessarily always result in reduction in peak memory.


QiJune (approved):
LGTM

PR #3294, opened 2025-04-04
[None][feat] CUTLASS MoE FC2+Finalize fusion

sklevtsov-nvidia:
This MR adds a new EVT-based implementation of GEMM2+Finalize fusion that supports both Blackwell and Hopper.

Using `trtllm-bench --backend pytorch --model mistralai/Mixtral-8x7B-Instruct-v0.1` with `--input-mean 1024 --output-mean 512 --num-requests 2048` on a single B200,
  * before (no fusion):
```
Request Throughput (req/sec):                     24.9259
Total Output Throughput (tokens/sec):             12762.0376
Per User Output Throughput (tokens/sec/user):     8.5910
Per GPU Output Throughput (tokens/sec/gpu):       12762.0376
Total Latency (ms):                               82163.6824
Average request latency (ms):                     62291.9844
```
  * after (with fusion):
```
Request Throughput (req/sec):                     25.9745
Total Output Throughput (tokens/sec):             13298.9388
Per User Output Throughput (tokens/sec/user):     8.9563
Per GPU Output Throughput (tokens/sec/gpu):       13298.9388
Total Latency (ms):                               78846.5916
Average request latency (ms):                     59856.6157
```

## Summary by CodeRabbit

* **New Features**
  * Added configuration options in Python and C++ APIs to enable or disable fused finalize optimization in Mixture-of-Experts (MoE) layers.
  * Introduced new environment variable and model configuration fields for controlling finalize fusion behavior.
  * Added support for routing scales and row permutation mappings in MoE kernel operations.

* **Enhancements**
  * Improved flexibility in MoE kernel fusion, allowing users to recover deterministic numerical behavior when needed.
  * Updated internal logic and APIs to support additional routing and permutation parameters in MoE operations.
  * Modernized epilogue fusion handling and updated launcher code for better architecture detection and scheduling.
  * Simplified and unified stride and pointer management in fused finalize epilogue structures.
  * Refined supported tile shapes for SM100 architecture and optimized kernel dispatch logic.

* **Bug Fixes**
  * Adjusted test tolerances for NVFP4 quantized fused MoE to improve reliability.
  * Corrected spelling in test variable names related to deterministic reduction.

* **Documentation**
  * Added descriptions for new configuration fields related to MoE fusion options.

* **Chores**
  * Updated internal library version hashes and checksums for consistency.


djns99 (approved):
LGTM thanks! See my comments on the internal MR

cpp/tensorrt_llm/cutlass_extensions/include/cutlass_extensions/gemm_configs.h
```diff
@@ -133,11 +133,6 @@ enum class CutlassTileConfigSM100
     CtaShape128x256x128B,
     CtaShape128x128x256B,
     CtaShape128x256x256B,
-
-    // M=256
-    CtaShape256x64x128B,
```
litaotju, line 138: why these are deleted?

cpp/tensorrt_llm/kernels/cutlass_kernels/python/generate_kernels.py
```diff
@@ -544,7 +539,7 @@ def generate_sm100_grouped_gemm_operations(is_arch_enabled):
 
     epi_fusions = [
```
litaotju, line 681: So we are keeping the path of non-fusion right?

How a user/developer can disable the fusion path? 
Or the fusion/non-fusion is automatically profiled or selected?

litaotju (dismissed):

cpp/tensorrt_llm/kernels/cutlass_kernels/python/generate_kernels.py
djns99: The fusion can introduce non-determinism for DeepSeek and other models with topk>2. The flag is not exposed in pytorch path yet afaik. 

In the TRT path this was controlled by the plugin flag (hardcoded to false at the moment)
```
    p_force_determinism = trt.PluginField(
        "force_determinism", np.array([int(False)], dtype=np.int32),
        trt.PluginFieldType.INT32)
```


Note there is an unrelated variable `FORCE_MOE_KERNEL_DETERMINISTIC` which causes a significant perf loss, and should probably be reconfigured to set this flag (and use hardcoded heuristics).


As a wider issue @litaotju, MOE does an autotuning step on start up. This can give different results for every run of the model. In TRT this was not an issue as it was fixed at engine build time, but on pytorch, every time the model is loaded we rerun and may get different results run to run.

sklevtsov-nvidia: I can add a way to disable fusion via an env var (unless TRT-LLM has another preferred way of setting perf options/hints). My thinking was along the lines of a more direct/explicit name like `TRTLLM_DISABLE_MOE_FINALIZE_FUSION` that addresses potential perf regression concerns rather than determinism concerns (i.e. applies even to `top_k < 3` models) - but happy to use whatever we settle on.

Somewhat related to the second issue, Murat asked me today whether the auto-tuning step also searches over fusion on/off configs (it does not; if fusion is not disabled for the model, it always stays on for all cutlass 3.x kernels in the profiler). It would take some config/profiler refactoring to add this capability, and my perspective is that fusion affects all configs in roughly similar ways - although technically it can consume extra registers and make some larger tile configs spill. Would it be an issue if we increased the profiler search space by 2x-4x (if we take into account the Cartesian product of `activation x finalize` fusions)?

cpp/tensorrt_llm/cutlass_extensions/include/cutlass_extensions/gemm_configs.h
sklevtsov-nvidia: I refactored the way tile configs are handled. Previously, for Blackwell only, this would actually be a product of `CtaTile x CgaTile`, with the division happening the in the kernel launcher to recover `CtaShape`. This change makes it consistent with Hopper kernels: `CtaShape` is now just per-CTA tile shape (which never goes above 128), so we can remove the extra tile configs that were only used for the 2SM MMA kernels (these are now controlled just by `CgaShape`).

Previously:
```
CtaShape64x64x128B  x CgaShape1x1x1 -> 1SM with per-CTA tile shape 64x64x128B
CtaShape128x64x128B x CgaShape1x1x1 -> 1SM with per-CTA tile shape 128x64x128B
CtaShape128x64x128B x CgaShape2x1x1 -> 2SM with per-CTA tile shape 64x64x128B
CtaShape256x64x128B x CgaShape2x1x1 -> 2SM with per-CTA tile shape 128x64x128B
CtaShape256x64x128B x CgaShape1x1x1 -> invalid
```
This PR:
```
CtaShape64x64x128B  x CgaShape1x1x1 -> 1SM with per-CTA tile shape 64x64x128B
CtaShape128x64x128B x CgaShape1x1x1 -> 1SM with per-CTA tile shape 128x64x128B
CtaShape64x64x128B  x CgaShape2x1x1 -> 2SM with per-CTA tile shape 64x64x128B
CtaShape128x64x128B x CgaShape2x1x1 -> 2SM with per-CTA tile shape 128x64x128B
```

cpp/tensorrt_llm/kernels/cutlass_kernels/python/generate_kernels.py
djns99: My preference would be a pytorch flag in the moeOp so that this can be controlled by model config files. The environment variable was added as a hack to unblock deepseek fires, I don't think this is a good solution for end users.

There should be no product between activation and finalize as GEMM 1 & 2 are profiled separately. Increasing profiler search space is not a problem I don't think. Though, I don't know how much of a priority fast start up times are (@litaotju any suggestions).
I am more concerned about increasing the template instantiations though, are all the relevant cases already instantiated for GEMM1 or will this add more?

cpp/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
```diff
@@ -377,72 +377,62 @@ std::vector<CutlassGemmConfig> get_candidate_configs_sm100(CutlassGemmConfig::Ca
...
-                        }
-                    }
-
-                    std::vector onesm{CutlassTileConfigSM100::CtaShape64x64x128B,
-                        CutlassTileConfigSM100::CtaShape64x128x128B, CutlassTileConfigSM100::CtaShape64x256x128B,
-                        CutlassTileConfigSM100::CtaShape128x64x128B};
-                    std::copy(onesm.begin(), onesm.end(), std::back_inserter(base));
-                }
+            tile_configs.push_back({CutlassTileConfigSM100::CtaShape128x16x128B, ClusterShape::ClusterShape_1x1x1});
+            // TODO(sklevtsov): re-enable when handled by the MoE GEMM dispatch
```
djns99, line 428: I think we are supposed to avoid todos with names? I think they will fail the guardwords check unless this has changed recently. Make sure you have set up pre-commit for the automated checks (though guardwords is only on linux)

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -4459,15 +4478,12 @@ void GemmProfilerBackend::prepareTmaWsInputs(
 
             bool apply_bias = true;
             bool use_w4afp8 = (mDType == nvinfer1::DataType::kFP8 && mWType == nvinfer1::DataType::kINT4);
-            bool using_fused_finalize
-                = !mInterface->use_deterministic_hopper_reduce_ && mSM == 90 && !mMinLatencyMode && !use_w4afp8;
+            bool using_fused_finalize = mInterface->use_fused_finalize_ && mSM >= 90 && !mMinLatencyMode && !use_w4afp8;
```
djns99, line 4481: Does it work on SM120?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1788,6 +1788,10 @@ class TorchLlmArgs(BaseLlmArgs):
     moe_backend: str = Field(default='CUTLASS',
                              description="MoE backend to use.")
 
+    moe_disable_finalize_fusion: bool = Field(
+        default=False,
+        description="Disable FC2+finalize kernel fusion in CUTLASS MoE backend")
```
djns99, line 1793: Can you clarify why someone would want to do this, there's no documentation anywhere I can see that tells someone that this will resolve non-determinism

djns99 (approved):

syuoni (approved):
Do we have an ETA of this PR landing to main? Thanks @sklevtsov-nvidia

cpp/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
sklevtsov-nvidia: Good catch, removed username. I am running pre-commit on linux, but I don't see it running any guardwords script, nor does it error in CI, perhaps something changed indeed.

sklevtsov-nvidia:
> do we have an ETA of this PR? It is good if we can take advantage of it in Mlperf V5.1 (Before 7/24), thanks

> Do we have an ETA of this PR landing to main?

I just had to do another rebase to account for LLM args changes in https://github.com/NVIDIA/TensorRT-LLM/pull/6003. If it passes CI without any new issues, it should be good to go.

Kefeng-Duan:
could we use this feature directly for Mlperf without all CI pass?

cpp/tensorrt_llm/cutlass_extensions/include/cutlass_extensions/epilogue/fusion/sm90_visitor_scatter.hpp
```diff
@@ -0,0 +1,547 @@
...
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ **************************************************************************************************/
+
+/*! \file
+  \brief Visitor tree store operations for the sm90 TMA warp-specialized (ws) epilogue
+*/
+
+#pragma once
```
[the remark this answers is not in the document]
sklevtsov-nvidia: This is a CUTLASS extension file and it's written using the coding style of CUTLASS which employs `#pragma once` instead of header guards.

cpp/tensorrt_llm/kernels/internal_cutlass_kernels/include/moe_gemm_kernels.h
```diff
@@ -147,37 +142,26 @@ struct TmaWarpSpecializedGroupedGemmInput
...
+        void const** ptr_bias = nullptr;
+        float const** ptr_router_scales = nullptr;
 
-        int64_t const* ptr_expert_first_token_offset = nullptr;
-        int const* ptr_source_token_index = nullptr;
+        int const** ptr_source_token_index = nullptr;
+        int num_rows_in_final_output = 0;
 
-        size_t num_rows_in_final_output = 0;
+        bool use_reduction = true;
```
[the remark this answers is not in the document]
sklevtsov-nvidia: These changes are already applied in respective places.

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -1196,7 +1197,18 @@ __device__ void computeTmaWarpSpecializedInputPointers(TmaWarpSpecializedGrouped
...
+        layout_info.ptr_d[out_idx] = safe_inc_ptr(output, num_tokens_before_expert * gemm_n);
+    }
+    if (layout_info.fusion == TmaWarpSpecializedGroupedGemmInput::EpilogueFusion::FINALIZE)
+    {
+
+        layout_info.fused_finalize_epilogue.ptr_source_token_index[expert]
+            = permuted_row_to_unpermuted_row + num_tokens_before_expert;
+        layout_info.fused_finalize_epilogue.ptr_router_scales[expert] = router_scales + num_tokens_before_expert;
+        if (layout_info.fused_finalize_epilogue.ptr_bias != nullptr)
+        {
```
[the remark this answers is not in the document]
sklevtsov-nvidia: There is an impliciit assumption that `router_scales != nullptr` when `layout_info.fusion` is set to `FINALIZE`. The fused finalize kernel epilogue does not have a `nullptr` check (by design) so it wouldn't be able to deal with nullptr values in `ptr_router_scales` anyway.

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_gemm_tma_warp_specialized_input.cu
```diff
@@ -103,28 +106,24 @@ void TmaWarpSpecializedGroupedGemmInput::configureWorkspace(int8_t* start_ptr, i
...
-    fused_finalize_epilogue.stride_bias
-        = transpose_stride(cute::make_stride(cute::Int<0>{}, cute::Int<1>{}, hidden_size));
-    fused_finalize_epilogue.stride_router_scales = {};
+
+    fused_finalize_epilogue.stride_final_output = cutlass::make_cute_packed_stride(
+        FusedFinalizeEpilogue::StrideFinalOutput{}, cute::make_shape(hidden_size, num_output_tokens, 1));
 
     fused_finalize_epilogue.num_rows_in_final_output = num_output_tokens;
+    fused_finalize_epilogue.use_reduction = use_reduction;
 }
```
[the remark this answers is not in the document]
sklevtsov-nvidia: This is correct because we do swap+transpose, so rows of output tensor are mapped to `hidden_size` and columns to output tokens.

sklevtsov-nvidia:
Passing CI now. Need codeowners approval @juney-nvidia @nv-guomingz @litaotju @yizhang-nv @Naveassaf @pcastonguay @lucaslie

Superjomn (approved):
LGTM on the llmapi change.

QiJune (approved):
LGTM

litaotju:
Could you help help to measure DeepSeek R1 performance with this one?

And what about WIDEEP moe backend? can it support  this optimization?

litaotju:
The feature is enabled by default. Can it support all the precisions we support in MOE?

BF16, FP8, FP4, and W4A8?

litaotju:
This optimization is done using the public cutlass only right?

Why do we need to update the cutlass binary integration .a?

sklevtsov-nvidia:
> The feature is enabled by default. Can it support all the precisions we support in MOE?
> 
> BF16, FP8, FP4, and W4A8?

It can only support BF16/FP16/FP32 outputs (due to use of HW atomics), but these are the only types for MoE plugin output [we currently support](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu#L4790-L4822) on CUTLASS path. It is agnostic w.r.t. input or intermediate precision/quantization.

sklevtsov-nvidia:
> This optimization is done using the public cutlass only right?
> 
> Why do we need to update the cutlass binary integration .a?

Correct. The only reason internal library is updated is to avoid code divergence (and because this was originally implemented in internal library before the kernels were OSS'ed, so I just kept the changes in both places).

sklevtsov-nvidia:
> Could you help help to measure DeepSeek R1 performance with this one?
> 
> And what about WIDEEP moe backend? can it support this optimization?

Based on my earlier measurements (~2-3 months ago), there was no observable change in DSR1 perf due to a very small fraction of runtime taken by finalize kernel. I will repeat the measurement on current version of the code if time permits.

IIUC, whenever WIDEEP uses CUTLASS MoE backend, this fusion is enabled. I have not measured runtime impact in that case, but it's likely to be minimal.

zongfeijing (approved):
LGTM

litaotju (approved):

Issue #3034, opened 2025-03-24
[RFC] [PyTorch Flow] Re-implement LlmRequest and Scheduler in pure Python


QiJune:
### Motivation

Currently, we have a pure Python based [PyExecutor](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/py_executor.py) class, which handles the main event loop. It provides good flexibility to support features like overlap scheduler and attention data parallelism quickly.

Inside it, we still use lots of pybind classes, including [LlmRequest](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/llm_request.py), [KVCacheManager ](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/resource_manager.py#L84) and [Scheduler](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/scheduler.py).

To improve flexibility further, we want to migrate more components from C++ to Python.

### Analysis

`LlmRequest`, `KVCacheManager` and `Scheduler` are coupled tightly. There are many state tensors maintained in `LlmRequest`, including output_tokens/chunk_size/state/etc. Both `KVCacheManager` and `Scheduler` read and write members of `LlmRequest` internally.

We tried to implement a pure Python [CapacityScheduler ](https://github.com/NVIDIA/TensorRT-LLM/blob/5b4a5014d1b6b72ebae03ae7ceb88642286c80d1/tensorrt_llm/_torch/pyexecutor/scheduler.py#L94-L145) before, but it introduces too much pybind calls of LlmRequest. We observed pybind calls are about 2X-3X slower than pure Python calls.  So, we don’t enable this pure Python CapacitySchedule due to its big host overhead.

Considering the complexity of `KVCacheManager`, we decide to re-implement `LlmRequest` and `Scheduler` in pure Python as the first step. We will remove `LlmRequest` from the `KVCacheManager` interface.

### Proposed Solution

1. Introduce a new flag `enable_pure_python_scheduler` in `PyTorchConfig` to enable pure Python based scheduler

Considering it takes some time to migrate all the components and do performance tuning, pure Python based scheduler will be hidden from users at the begining.

2. Refactor `LlmRequest` to support maintaining all state tensors in Python side

[...]

PR #3297, opened 2025-04-05
fix: fix the py_decoding_iter update in decoder

lfr-0531:
fix the py_decoding_iter update in decoder


Shunkangz:
May I know why we cannot reuse the request.decoding_iter instead of creating a new one? Because I also modify this in this [MR](https://github.com/NVIDIA/TensorRT-LLM/pull/3241). I want to ensure consistency. Thank you!

QiJune:
> May I know why we cannot reuse the request.decoding_iter instead of creating a new one? Because I also modify this in this [MR](https://github.com/NVIDIA/TensorRT-LLM/pull/3241). I want to ensure consistency. Thank you!

@Shunkangz , `request.decoding_iter` is a pybind call, which is 2X-3X slower than a pure Python call, `request.py_decoding_iter`. So, we will maintain states of `LlmRequest` in Python side as many as we can. More background: https://github.com/NVIDIA/TensorRT-LLM/issues/3034

QiJune (approved):
LGTM

Shunkangz:
> > May I know why we cannot reuse the request.decoding_iter instead of creating a new one? Because I also modify this in this [MR](https://github.com/NVIDIA/TensorRT-LLM/pull/3241). I want to ensure consistency. Thank you!
> 
> @Shunkangz , `request.decoding_iter` is a pybind call, which is 2X-3X slower than a pure Python call, `request.py_decoding_iter`. So, we will maintain states of `LlmRequest` in Python side as many as we can. More background: #3034

Thank you! I will also refactor my code.

PR #3299, opened 2025-04-05
fix: fix attentionDP padding request type

Shunkangz:
The previous implement only targets IFB case where we can only fetch the context request from the request queue. In disaggregated serving, we can fetch both context request and generation request from it. In this case, we need to additional check condition to determine whether this request is context or not.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -18,6 +18,7 @@ import numpy as np
 import torch
 
 from tensorrt_llm._utils import global_mpi_rank, nvtx_range
+from tensorrt_llm.bindings import executor as tllme
 from tensorrt_llm.bindings.executor import (FinishReason, InflightBatchingStats,
                                             IterationStats, KvCacheStats)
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -18,6 +18,7 @@
 import torch
 
 from tensorrt_llm._utils import global_mpi_rank, nvtx_range
+from tensorrt_llm.bindings import executor as tllme
 from tensorrt_llm.bindings.executor import (FinishReason, InflightBatchingStats,
```
jinyangyuan-nvidia, line 21: Since only `RequestType` is used in `tllme`, maybe we can just additionally import `RequestType` from `tensorrt_llm.bindings.executor`.

jinyangyuan-nvidia (approved):

tensorrt_llm/_torch/pyexecutor/py_executor.py
Shunkangz: Fixed. Thank you!

PR #3300, opened 2025-04-05
refactor: batch slot management in decoder classes

Funatiq:
Highlights:
* Set up batch slots in `MakeDecodingBatchInputOutput` (outside of the decoder).
* Set up logits vector in `MakeDecodingBatchInputOutput` (outside of the decoder).
* Remove `active` vector from `decoder_batch::Input`.
* Unify the creation of decoder batch inputs in algorithm and tests.

See commit messages for details.

### DIFF
--- a/cpp/include/tensorrt_llm/batch_manager/decoderBuffers.h
+++ b/cpp/include/tensorrt_llm/batch_manager/decoderBuffers.h
@@ -37,7 +37,7 @@ public:
     using TensorPtr = runtime::ITensor::SharedPtr;
 
     explicit DecoderInputBuffers(
-        SizeType32 maxBatchSize, SizeType32 maxTokensPerEngineStep, runtime::BufferManager const& manager);
+        SizeType32 maxBatchSize, SizeType32 maxDecoderSteps, runtime::BufferManager const& manager);
 
     // buffers for setup
     TensorPtr setupBatchSlots;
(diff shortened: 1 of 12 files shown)


Funatiq:
/bot run --stage-list "H100_PCIe-TensorRT-1, H100_PCIe-TensorRT-3"

cpp/include/tensorrt_llm/batch_manager/decoderBuffers.h
```diff
@@ -48,7 +48,7 @@ class DecoderInputBuffers
     TensorPtr forwardBatchSlotsRequestOrderDevice;
     TensorPtr fillValues;
     TensorPtr fillValuesDevice;
-    TensorPtr forwardBatchSlots;
+    std::vector<TensorPtr> forwardBatchSlots;
```
dcampora, line 51: Why this change?

dcampora (approved):

cpp/include/tensorrt_llm/batch_manager/decoderBuffers.h
Funatiq: So each tensor can be resized to the actual batch size. Previously it was a sparse 2-dim tensor. Also see [here](https://github.com/NVIDIA/TensorRT-LLM/pull/3300/files#diff-a86768fcf5826f655591d71de3615f66d0f5b55583c7329b6007cfa21c6f9e1fR73).

PR #3302, opened 2025-04-05
feat: Add Llama 4

mikeiovine:
Add LLaMA4. Support BF16 Maverick (128e) and Scout (16e) currently; FP8 support to be added in followups.

This MR squashes our work from the internal branch and rebases onto the latest main. A few important notes:
* I've commented out all topk/softmax fusion op stuff for now. These needs to be merged into `internal_cutlass_kernels` before it can be used in the external GitHub.
* Vision support is not included. We should figure out a better way to structure things for multimodality in a followup.
* **I have updated transformers** to avoid the previous hacks to detect the model architecture type.
* For testing: I'm adding one end to end sanity test for both Scout and Maverick. **It cannot be enabled yet because these models have not been added to `llm-models`**. We need to enable these ASAP, the only reason I'm not waiting is because this PR needs to land to unblock others.


tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+            dtype: Optional[torch.dtype] = None,
+            tune_max_num_tokens: int = 8192,
+            model_config: ModelConfig = ModelConfig(),
+    ):
+        from tensorrt_llm._torch.distributed import AllReduce
+
+        super().__init__()
+        config = model_config.pretrained_config
+        self.top_k = top_k
+        self.use_dp = model_config.mapping.enable_attention_dp
```
hlu1: Remove all `enable_attention_dp` code. That's deepseek only.

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        all_rank_num_tokens=None,
+        final_all_reduce_params: Optional[AllReduceParams] = None,
+        min_latency_mode: Optional[bool] = False,
+    ) -> torch.Tensor:
+        if min_latency_mode:
+            assert not self.use_dp
```
hlu1: Remove

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+        self.top_k = top_k
+        self.use_dp = model_config.mapping.enable_attention_dp
+        self.experts = FusedMoE(
+            routing_method=RenormalizeMoeRoutingMethod(top_k),
+            num_experts=num_experts,
+            hidden_size=hidden_size,
+            intermediate_size=intermediate_size,
+            dtype=dtype,
+            reduce_results=
+            False,  # In both low latency and attention dp scenarios, FusedMoE needs not to do allreduce inside op.
```
hlu1:
```suggestion
            False,  # In both low latency and max-throughput scenarios, FusedMoE needs not to do allreduce inside op.
```

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+            weight_loading_mode=MoEWeightLoadingMode.FUSED_GATE_UP_PROJ,
+            model_config=model_config)
+
+        self.shared_expert = GatedMLP(
+            hidden_size=hidden_size,
+            intermediate_size=shared_expert_intermediate_size,
+            bias=False,
+            dtype=dtype,
+            config=model_config,
+            is_expert=not self.use_dp)
```
hlu1:
```suggestion
            is_expert=True)
```

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+            return input_tensor
+        dst_tensor = input_tensor
+        outputs = reducescatter(dst_tensor, self.parallel_config, scatter_dim=0)
+        depad_tensors = outputs[:all_rank_num_tokens[rank]]
+        return depad_tensors
+
+    def compute_routed_output(self, hidden_states, all_rank_num_tokens,
+                              min_latency_mode):
+        if self.use_dp:
+            hidden_states = self.all_gather(hidden_states, all_rank_num_tokens)
```
hlu1: Remove

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+    def all_gather(self, input_tensor, all_rank_num_tokens):
+        world_size = self.parallel_config.tensor_parallel_size
+        max_num_token = max(all_rank_num_tokens)
+        if world_size == 1:
+            return input_tensor
+
+        pad_tensor = torch.nn.functional.pad(
+            input_tensor, (0, 0, 0, max_num_token - input_tensor.shape[0]))
+        outputs = allgather(pad_tensor, self.parallel_config, gather_dim=0)
+        return outputs
```
hlu1: Remove

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+    def compute_routed_output(self, hidden_states, all_rank_num_tokens,
+                              min_latency_mode):
+        if self.use_dp:
+            hidden_states = self.all_gather(hidden_states, all_rank_num_tokens)
+        router_logits = self.router(hidden_states)
+        routed_output = self.experts(hidden_states, router_logits,
+                                     min_latency_mode)
+        if self.use_dp:
+            routed_output = self.reduce_scatter(routed_output,
+                                                all_rank_num_tokens)
```
hlu1: Remove

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+    def reduce_scatter(self, input_tensor, all_rank_num_tokens):
+        world_size = self.parallel_config.tensor_parallel_size
+        rank = self.parallel_config.tensor_parallel_rank
+        max(all_rank_num_tokens)
+        if world_size == 1:
+            return input_tensor
+        dst_tensor = input_tensor
+        outputs = reducescatter(dst_tensor, self.parallel_config, scatter_dim=0)
+        depad_tensors = outputs[:all_rank_num_tokens[rank]]
+        return depad_tensors
```
hlu1: Remove

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,194 @@ def __init__(
...
+            routed_output = self.compute_routed_output(hidden_states,
+                                                       all_rank_num_tokens,
+                                                       min_latency_mode)
+        if min_latency_mode:
+            return [shared_output, *routed_output]
+
+        assert shared_output.size() == routed_output.size(
+        ), f'unmatched tensor shape'
+        final_hidden_states = shared_output + routed_output
+        if not self.use_dp and self.parallel_config.tensor_parallel_size > 1:
```
hlu1:
```suggestion
        if self.parallel_config.tensor_parallel_size > 1:
```

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -107,28 +297,85 @@ def forward(
...
+        ) <= 128 and self.fusion_config.POST_MOE_FUSION and is_blackwell and self.is_quanted
 
         # Self Attention
         hidden_states = self.self_attn(
             position_ids=position_ids,
             hidden_states=hidden_states,
             attn_metadata=attn_metadata,
+            all_reduce_params=AllReduceParams(enable_allreduce=not (
+                self.fusion_config.PRE_MOE_FUSION or self.parallel_config.
+                tensor_parallel_size == 1 or self.enable_attention_dp)),
```
hlu1:
```suggestion
                tensor_parallel_size == 1)),
```

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -107,28 +297,85 @@ def forward(
...
+            hidden_states, residual = self.post_attention_layernorm(
+                hidden_states, residual)
+
+        hidden_states = self.feed_forward(
+            hidden_states,
+            all_rank_num_tokens=attn_metadata.all_rank_num_tokens,
+            final_all_reduce_params=AllReduceParams(enable_allreduce=not (
+                self.fusion_config.POST_MOE_FUSION
+                or self.fusion_config.POST_MLP_FUSION or self.parallel_config.
+                tensor_parallel_size == 1 or self.enable_attention_dp)),
```
hlu1:
```suggestion
                tensor_parallel_size == 1)),
```

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -15,6 +15,24 @@
...
+
+    def _norm(self, x):
+        flash_infer_kernel = RMSNorm(hidden_size=x.shape[-1],
+                                     eps=self.eps,
+                                     dtype=self.dtype,
+                                     device=x.device)
+        return flash_infer_kernel(x)
+
+    def forward(self, x):
+        return self._norm(x.contiguous()).type_as(x)
```
hlu1:
```suggestion
        return self._norm(x)
```
flash_infer norm now supports strided inputs.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -126,24 +157,48 @@ def forward(
...
+                    self.ln_events[0].wait()
+                    q = self.qk_norm(q)
+                    self.ln_events[1].record()
+                self.ln_events[1].wait()
+            else:
+                q = self.qk_norm(q)
+                k = self.qk_norm(k)
+            qkv = torch.concat(
+                [q.contiguous(), k.contiguous(),
+                 v.contiguous()], dim=-1)
```
hlu1:
```suggestion
            qkv = torch.concat(
                [q, k,
                 v], dim=-1)
```
concat supports strided inputs

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -126,24 +157,48 @@ def forward(
...
+                if self.pos_embd_params is None and position_ids is not None:
+                    q, k, v = qkv.split(
+                        [self.q_size, self.kv_size, self.kv_size], dim=-1)
+                    q, k = self.rotary_emb(
+                        position_ids,
+                        [q.contiguous(), k.contiguous()], attn_metadata)
+                    qkv = torch.concat(
+                        [q.contiguous(),
+                         k.contiguous(),
+                         v.contiguous()],
```
hlu1: Same

tests/unittest/_torch/multi_gpu_modeling/test_llama4.py
```diff
@@ -0,0 +1,51 @@
...
+from tensorrt_llm import SamplingParams
+from tensorrt_llm._torch import LLM
+from tensorrt_llm._torch.pyexecutor.config import PyTorchConfig
+
+
+@pytest.mark.parametrize(
+    "model_name",
+    ["Llama-4-Maverick-17B-128E-Instruct", "Llama-4-Scout-17B-16E-Instruct"],
+    ids=['maverick', 'scout'])
+@pytest.mark.parametrize("backend", ["TRTLLM", "FLASHINFER"],
```
hlu1, line 16: Does FLASHINFER actually work?

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -126,24 +157,48 @@ def forward(
...
-                q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size],
-                                    dim=-1)
-                q, k = self.rotary_emb(
-                    position_ids,
-                    [q.contiguous(), k.contiguous()], attn_metadata)
-                qkv = torch.concat(
-                    [q.contiguous(),
-                     k.contiguous(),
-                     v.contiguous()], dim=-1)
+            if self.use_rope:
```
QiJune: If attention backend is TRTLLM, then,  `fuse_pos_embd` is True, `is_fused_qkv` is True. Thus, `use_rope` is always equal to `pos_embd_params is not None`.

If attention backend is FLASHINFER, then, `fuse_pos_embd` is False, `is_fused_qkv` is False, `pos_embd_params` is always None. `use_rope` is used to decide whether to call `self.rotary_emb` or not.

QiJune: I know @yuxianq plans to refactor position embedding part of attention layer.

tests/unittest/_torch/multi_gpu_modeling/test_llama4.py
mikeiovine: Yeah, it works. I used it a lot for debugging. It's helpful because it does rope externally

yechank-nvidia:
Can you add llama4 on supported-models on [examples/pytorch/REAME.md](https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples/pytorch#supported-models)?

tensorrt_llm/_torch/modules/attention.py
mikeiovine: I see, could definitely use a refactor. I didn't know that flashinfer always set fused qkv to false! Removing `use_rope` flag in attention for now.

hlu1 (approved):

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -126,24 +155,41 @@ def forward(
...
+                k = self.qk_norm(k)
+            qkv = torch.concat([q, k, v], dim=-1)
+
         if is_fused_qkv:
             if self.pos_embd_params is None and position_ids is not None:
                 q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size],
                                     dim=-1)
                 q, k = self.rotary_emb(
                     position_ids,
                     [q.contiguous(), k.contiguous()], attn_metadata)
```
hlu1: Is this part no-op then if we're invoking the trtllm attention? If so, remove the `.contiguous()` calls.

hlu1: We should add a comment on which attention backend corresponds to which code path to make the code easier to follow.

cpp/tensorrt_llm/kernels/unfusedAttentionKernels/unfusedAttentionKernels_2_template.h
```diff
@@ -750,6 +751,10 @@ __global__ void applyBiasRopeUpdateKVCacheV2(QKVPreprocessingParams<T, KVCacheBu
     //  1. Contiguous QKV output.
     //  2. Contiguous Q output + Paged KV output (needed by Paged KV FMHA kernels).
 
+#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
```
Yang-YiFan: This `griddepcontrol.wait;` can be pushed further at the end of the prolog.

Yang-YiFan (approved):
The PDL part looks good to me. Though there is one comment I made that could potentially improve perf. We can leave it as is. I can take a pass later in the day to tune all kernels for PDL in a separate MR.

tensorrt_llm/_torch/modules/attention.py
mikeiovine: Leaving the commenting to @yuxianq when he does his refactor

cpp/tensorrt_llm/kernels/unfusedAttentionKernels/unfusedAttentionKernels_2_template.h
mikeiovine: ~~Leaving to someone who is more familiar with PDL~~ just gonna do it here since it's simple enough

cpp/tensorrt_llm/kernels/allReduceFusionKernels.cu
```diff
@@ -605,24 +596,24 @@ bool use_fp32_acc()
...
+        return allreduce_fusion_kernel_launcher<TEMP_0, TEMP_1, TEMP_2, TEMP_3, TEMP_4, false>(params);                \
     }
 
-#define DISPATCH0(NRanks, ResidualOut, NormOut, QuantOut)                                                              \
-    if (params.nranks == NRanks && params.dtype == nvinfer1::DataType::kHALF)                                          \
+#define DISPATCH0(TEMP_1, TEMP_2, TEMP_3, TEMP_4)                                                                      \
+    if (params.nranks == TEMP_1 && params.dtype == nvinfer1::DataType::kHALF)                                          \
     {                                                                                                                  \
-        DISPATCH1(half, NRanks, ResidualOut, NormOut, QuantOut);                                                       \
+        DISPATCH1(half, TEMP_1, TEMP_2, TEMP_3, TEMP_4);                                                               \
```
litaotju: Nit. 
But is this some mis-rebase? I dont see reasons to change these for llama4

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -107,28 +271,85 @@ def forward(
         spec_metadata: Optional[SpecMetadata] = None,
         **kwargs,
     ) -> torch.Tensor:
-        if residual is None:
-            residual = hidden_states
```
litaotju, line 111: Will this breaks previous llama with pre layer norm?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -132,18 +132,12 @@ def load_weights(checkpoint_dir: str):
         for file in weight_files:
             # try mmap first, if failed, turn off mmap
             try:
-                part_weights = torch.load(file,
-                                          weights_only=True,
```
litaotju: why weights_only=True removed?

tensorrt_llm/_torch/models/modeling_llama.py
mikeiovine: That's what the unit tests are for. We'll see

cpp/tensorrt_llm/kernels/allReduceFusionKernels.cu
mikeiovine: ~~This change is authored by someone else. Can we check the blame on the internal branch for the full context?~~ After actually looking at it: yeah, let's change this back.

tensorrt_llm/_torch/pyexecutor/model_engine.py
mikeiovine: I'll put it back

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,168 @@ def __init__(
...
+            tune_max_num_tokens: int = 8192,
+            model_config: ModelConfig = ModelConfig(),
+    ):
+        from tensorrt_llm._torch.distributed import AllReduce
+
+        super().__init__()
+        config = model_config.pretrained_config
+        self.top_k = top_k
+        self.experts = FusedMoE(
+            routing_method=RenormalizeMoeRoutingMethod(top_k),
```
syuoni: Hi @mikeiovine ,

I'm comparing the LLaMA 4 implementations of TRT-LLM against SGLang, vLLM, and HF. I see in HF the router logits are normalized by `sigmoid` (instead of `softmax`) and also multiplied to MoE inputs (see [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama4/modeling_llama4.py#L174-L183)). Are these two points reflected in TRT-LLM?

mikeiovine: That could explain why the MMLU is slightly regressed compared to the baseline. Let me investigate briefly, thanks for pointing this out

requirements.txt
```diff
@@ -23,10 +23,10 @@ tensorrt~=10.8.0
 # https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/rel-25-01.html#rel-25-01 uses 2.6.0a0.
 torch>=2.6.0a0,<=2.6.0
 torchvision
-nvidia-modelopt[torch]~=0.25.0
+nvidia-modelopt[torch]~=0.27.0
```
syuoni, line 26: Why should we upgrade modelopt in this PR?

mikeiovine: Required. Older versions of model opt don't work with the new version of transformers.

tensorrt_llm/_torch/models/modeling_llama.py
mikeiovine: Switching to sigmoid indeed significantly boosts MMLU. We have another team investigating accuracy, so I'll leave the rest to them. Checking in the sigmoid change for now.

cpp/tensorrt_llm/kernels/unfusedAttentionKernels/unfusedAttentionKernels_2_template.h
```diff
@@ -800,6 +801,10 @@ __global__ void applyBiasRopeUpdateKVCacheV2(QKVPreprocessingParams<T, KVCacheBu
     // Make sure there are multiple of tokens_per_block otherwise syncthreads will lead to deadlocks.
     int const tokens_loop_end = int((params.token_num + TOKENS_PER_BLOCK - 1) / TOKENS_PER_BLOCK) * TOKENS_PER_BLOCK;
 
+#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
```
Yang-YiFan: Which member of `params` is produced by the previous kernel? Looks like potentially `tokens_info`, `cu_seq_lens ` and  `token_num`? If so, this would affect correctness. Please put `griddepcontrol.wait` at line 782 then (after `int const head_idx = blockIdx.y;`).

tests/integration/test_lists/waives.txt
```diff
@@ -462,3 +462,4 @@ examples/test_qwen.py::test_llm_qwen_moe_single_gpu_summary[qwen1.5_moe_a2.7b_ch
 disaggregated/test_disaggregated.py::test_disaggregated_overlap_dp[DeepSeek-V3-Lite-fp8] SKIP (https://nvbugs/5166600)
 disaggregated/test_disaggregated.py::test_disaggregated_deepseek_v3_lite_fp8_ucx[DeepSeek-V3-Lite-fp8] SKIP (https://nvbugs/5201168)
 unittest/_torch/multi_gpu_modeling -k "deepseek and tp2_pp2_ep1_nextn0_enable_dp" SKIP (https://nvbugspro.nvidia.com/bug/5206873)
+accuracy/test_cli_flow.py::TestLlama3_8BInstruct::test_nvfp4_gemm_plugin[disable_norm_quant_fusion-enable_fused_quant] SKIP (bug in modelopt)
```
mikeiovine: Waiving due to confirmed bug in latest modelopt version. Bug will be fixed in the next release.

mikeiovine:
/bot skip --comment "Manually confirmed that the one test failure is happening on main as well"

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,180 @@ def __init__(
...
+        self.fusion_config.POST_MLP_FUSION = False
+
+        self.is_llama4 = config.model_type == "llama4_text"
         self.self_attn = LlamaAttention(
             model_config,
             layer_idx=layer_idx,
+            use_qk_norm=getattr(config, "use_qk_norm", False),
+            aux_stream=aux_stream,
+            use_gptj_style_rope=self.is_llama4,
         )
```
syuoni, line 215: Hi @mikeiovine ,

LLaMA 4 has [attention temperature scaling](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama4/modeling_llama4.py#L331-L337), but I cannot find the corresponding code in this PR. Did I miss something?

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -64,32 +83,180 @@ def __init__(
...
+        )
+        self.fusion_config = EagerFusionConfig()
+        # self.fusion_config.PRE_MOE_FUSION = model_config.mapping.has_tp(
+        # )
+        # TODO: re-enable these fusions
+        self.fusion_config.PRE_MOE_FUSION = False
+        self.fusion_config.POST_MLP_FUSION = False
+
+        self.is_llama4 = config.model_type == "llama4_text"
         self.self_attn = LlamaAttention(
```
syuoni, line 209: Hi @mikeiovine ,

Sorry I need to ask one more question. LLaMA 4 has [chunked attention](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama4/modeling_llama4.py#L828-L850), did we implement that in this PR?

tensorrt_llm/_torch/modules/rotary_embedding.py
```diff
@@ -177,6 +179,9 @@ def forward(
                     k = k.view(seq_len, -1)
                     return [q, k]
 
+        if use_gptj_style_rope:
```
nvzhihanj, line 182: @mikeiovine turns out this line will be hit after the `if IS_FLASHINFER_AVAILABLE` clause - I will attempt to fix it in https://github.com/NVIDIA/TensorRT-LLM/pull/3382
(I wonder why unit test didn't catch this)

tensorrt_llm/_torch/models/modeling_llama.py
mikeiovine: No. It's being handled separately.

tensorrt_llm/_torch/models/modeling_llama.py
mikeiovine: That is being handled separately.

tensorrt_llm/_torch/modules/rotary_embedding.py
mikeiovine: @nvzhihanj This is the intended behavior though. If you can't run `flashinfer_apply_rope_inplace`, you need to bail out. Because the other path is doing the wrong computation

tensorrt_llm/_torch/modules/attention.py
nvzhihanj: > If attention backend is TRTLLM, then, fuse_pos_embd is True, is_fused_qkv is True. Thus, use_rope is always equal to pos_embd_params is not None.

@QiJune this assumption is broken because Llama4 NOPE layer uses `is_fused_qkv=True` but has `pos_embd_params=None` because the rope is skipped. If we don't have a flag to skip rope I am not sure how to handle this. I will add a fix and tag you in the new PR: https://github.com/NVIDIA/TensorRT-LLM/pull/3382/files#

nvzhihanj: I see 2 ways to resolve this - one is we should not have `position_ids` passed down to attention module in nope layers; second is to manually override, using Michael's old approach.

PR #3303, opened 2025-04-06
feat: use cudaMalloc to allocate kvCache

chuangz0:
use cudaMalloc to allocate kvCache , so kvCache pool can be register by transfer enigne


Shixiaowei02 (approved):

schetlur-nv:
@chuangz0 this change seems to be pretty general, and is not limited to disaggregated serving. Did we assess that using `cudaMalloc` does not affect other use-cases?

PR #3304, opened 2025-04-06
feat: Draft: add scheduler module to disaggregated serving

Shunkangz:
Add a schedule module in disaggregated serving.


tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+        num_tokens = self._get_request_num_tokens(request, server_type)
+
+        with self._lock:
+            if server_type == "context":
+                server = heapq.heappop(self._ctx_server_load_heap)[1]
+                self._ctx_server_load[server] += num_tokens
+                heapq.heappush(self._ctx_server_load_heap,
+                               (self._ctx_server_load[server], server))
+                # TODO: Can we have a unique id for each request?
+                self._ctx_req_routing_table[request.prompt] = server
```
zhengd-nv: Does [python object id](https://docs.python.org/3.12/library/functions.html#id) applicable in this case? Prompts of different requests can be the same.

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+            else:
+                server = heapq.heappop(self._gen_server_load_heap)[1]
+                self._gen_server_load[server] += num_tokens
+                heapq.heappush(self._gen_server_load_heap,
+                               (self._gen_server_load[server], server))
+                # TODO: Can we have a unique id for each request?
+                self._gen_req_routing_table[request.prompt] = server
+        return server
+
+    def finish_request(self, request: CompletionRequest, server_type: str):
```
zhengd-nv: Can you add it as base method in `Scheduler` and make it no-op in `RoundRobinScheduler`.
Also, I think `finish_request(self, request, server)` is better, because the disagg server already knows which server it selected previously, doing so would save a server lookup.

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+
+    scheduler_class = scheduler_map.get(scheduler_type.lower())
+    if scheduler_class is None:
+        raise ValueError(f"Unsupported scheduler type: {scheduler_type}. "
+                         f"Supported types are: {list(scheduler_map.keys())}")
+
+    return scheduler_class(ctx_servers, gen_servers)
+
+
+if __name__ == "__main__":
```
zhengd-nv: A standalone unit test would be better.

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+        else:
+            num_tokens = 1
+        return num_tokens
+
+    def schedule(self, request: CompletionRequest, server_type: str) -> str:
+        num_tokens = self._get_request_num_tokens(request, server_type)
+
+        with self._lock:
+            if server_type == "context":
+                server = heapq.heappop(self._ctx_server_load_heap)[1]
```
zhengd-nv: We can use something like `ServerStatus` and `add/remove_request()` to avoid copying codes between context and generation servers. The status class can also be exposed to the upper level for status tracking and manual updating (e.g. KV cache status update).

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+                del self._ctx_req_routing_table[request.prompt]
+            else:
+                server = self._gen_req_routing_table[request.prompt]
+                self._gen_server_load[server] -= num_tokens
+                heapq.heappush(self._gen_server_load_heap,
+                               (self._gen_server_load[server], server))
+                del self._gen_req_routing_table[request.prompt]
+
+
+def creat_scheduler(scheduler_type: str,
```
zhengd-nv: `create_scheduler`

tensorrt_llm/serve/scheduler.py
Yuening-wa: Agree. I am also thinking about using a class like `ServerInfo` to track the the status of the server, including whether it is healthy. Here is an reference example.
```
class ServerInfo:
    def __init__(self, url: str):
        self.url = url
        self.active_requests = 0
        self.is_healthy = True
        self._lock = asyncio.Lock()

    async def increment_load(self):
        async with self._lock:
            self.active_requests += 1

    async def decrement_load(self):
        async with self._lock:
            self.active_requests = max(0, self.active_requests - 1)
```

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
+import heapq
+from abc import ABC, abstractmethod
+from threading import Lock
+from typing import List
+
+from tensorrt_llm.serve.openai_protocol import CompletionRequest
+
+
+class Scheduler(ABC):
```
pcastonguay: I wonder if `Router` would be a better name for the class, as it routes requests to the best server.

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+    @abstractmethod
+    def schedule(self, request: CompletionRequest, server_type: str):
+        pass
+
+
+class RoundRobinScheduler(Scheduler):
+
+    def __init__(self,
+                 ctx_servers: List[str] = None,
+                 gen_servers: List[str] = None):
```
pcastonguay: I think we should have a router for ctx servers and a separate router for gen servers. This way you don't have to store both lists. 
I think we should have one router for ctx servers and one router for gen servers to simplify the code and make it more flexible (could you different routing strategies for ctx/gen servers.

```suggestion
    def __init__(self,
                 servers: List[str] = None,
```

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+        super().__init__(ctx_servers, gen_servers)
+        self._ctx_server_idx = 0
+        self._gen_server_idx = 0
+
+    def schedule(self, request: CompletionRequest, server_type: str):
```
pcastonguay:
```suggestion
    def get_next_server(self, request: CompletionRequest):
```

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+        if server_type == "context":
+            server = self._ctx_servers[self._ctx_server_idx]
+            self._ctx_server_idx = (self._ctx_server_idx + 1) % len(
+                self._ctx_servers)
+        else:
+            server = self._gen_servers[self._gen_server_idx]
+            self._gen_server_idx = (self._gen_server_idx + 1) % len(
+                self._gen_servers)
```
pcastonguay:
```suggestion
            server = self._servers[self._server_idx]
            self._server_idx = (self._server_idx + 1) % len(
                self._servers)
```

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+        return server
+
+
+class LoadBalancingScheduler(Scheduler):
+
+    def __init__(self,
+                 ctx_servers: List[str] = None,
+                 gen_servers: List[str] = None):
```
pcastonguay: Similar to `RoundRobin` only deal with one list of `servers`

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+    def remove_server(self, server: str, server_type: str):
+        raise NotImplementedError(
+            "Currently not support dynamic removing of servers")
+
+    @abstractmethod
+    def schedule(self, request: CompletionRequest, server_type: str):
+        pass
+
+
+class RoundRobinScheduler(Scheduler):
```
pcastonguay:
```suggestion
class RoundRobinRouter(Router):
```

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+        raise NotImplementedError(
+            "Currently not support dynamic adding of servers")
+
+    def remove_server(self, server: str, server_type: str):
+        raise NotImplementedError(
+            "Currently not support dynamic removing of servers")
+
+    @abstractmethod
+    def schedule(self, request: CompletionRequest, server_type: str):
+        pass
```
pcastonguay:
```suggestion
    @abstractmethod
    def get_next_server(self, request: CompletionRequest):
        pass
```

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+        self._gen_servers = gen_servers
+        self._lock = Lock()
+
+    def add_server(self, server: str, server_type: str):
+        raise NotImplementedError(
+            "Currently not support dynamic adding of servers")
+
+    def remove_server(self, server: str, server_type: str):
+        raise NotImplementedError(
+            "Currently not support dynamic removing of servers")
```
pcastonguay:
```suggestion
    def add_server(self, server: str):
        raise NotImplementedError(
            "Currently not support dynamic adding of servers")

    def remove_server(self, server: str):
        raise NotImplementedError(
            "Currently not support dynamic removing of servers")
```

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+            self._ctx_server_idx = (self._ctx_server_idx + 1) % len(
+                self._ctx_servers)
+        else:
+            server = self._gen_servers[self._gen_server_idx]
+            self._gen_server_idx = (self._gen_server_idx + 1) % len(
+                self._gen_servers)
+        return server
+
+
+class LoadBalancingScheduler(Scheduler):
```
pcastonguay: Right now, load balanding is done using number of tokens. Should we have an option to load balance based on number of active requests?

tensorrt_llm/serve/scheduler.py
pcastonguay: Zhen, if we have a separate router for ctx and gen servers, this would simplify the code. 

I agree that `ServerInfo` would be useful to track state of each server. Maybe call it `ServerState`? `increment_load` should probably take the `request` as input so that we can keep track of number of tokens too for example.

tensorrt_llm/serve/scheduler.py
pcastonguay: `server` could be optional.

tensorrt_llm/serve/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+                self._gen_req_routing_table[request.prompt] = server
+        return server
+
+    def finish_request(self, request: CompletionRequest, server_type: str):
+        num_tokens = self._get_request_num_tokens(request, server_type)
+
+        with self._lock:
+            if server_type == "context":
+                server = self._ctx_req_routing_table[request.prompt]
+                self._ctx_server_load[server] -= num_tokens
```
pcastonguay: As suggested earlier, we should have a `decrement_load` function for the `ServerState`, and store a map between `server`s and `ServerState`s

pcastonguay (commented):
Thanks for the PR

tensorrt_llm/serve/scheduler.py
jgangani: I agree. "Router" is a better word, also avoids confusion with IFB scheduler

tensorrt_llm/serve/router.py
```diff
@@ -0,0 +1,202 @@
...
+    async def finish_request(self, request: CompletionRequest):
+        async with self._lock:
+            server = self._req_routing_table[id(request)]
+            await self._server_state[server].decrement_load(request)
+            heapq.heappush(self._server_load_heap,
+                           (self._get_server_load(server), server))
+            del self._req_routing_table[id(request)]
+
+
+def create_router(router_type: str, servers: List[str]) -> Router:
```
pcastonguay: Should we also pass the load_balancing criteria (tokens or active_requests)?

tensorrt_llm/serve/router.py
```diff
@@ -0,0 +1,202 @@
...
+    async def _monitor_servers(self, poll_interval: int = 10):
+        """Monitor servers update from metadata service"""
+        while True:
+            new_servers = await self.fetch_live_servers()
+
+            async with self._lock:
+                if new_servers != self._servers:
+                    self._servers = new_servers
+
+            await asyncio.sleep(poll_interval)
```
pcastonguay, line 100: Should we use the `health` endpoint here to remove servers that are not responding?

tensorrt_llm/serve/router.py
```diff
@@ -0,0 +1,202 @@
...
+        pass
+
+
+class LoadBalancingRouter(Router):
+
+    def __init__(self, servers: List[str] = None, use_tokens: bool = True):
+        super().__init__(servers)
+        # Load map between servers and their number of tokens processed
+        self._server_state = {}
+        self._server_load = {}
```
pcastonguay: Doesn't seem to be used.

tensorrt_llm/serve/router.py
Shunkangz: Yes, we can add this check logic here. But we still cannot guarantee that the sever is still alive when we send the requests. My thinking is that we can simply regard the servers are healthy if they still appear in etcd. Otherwise, the servers will be removed as long as its lease outdated. Besides, we can additional logic if any request is timeout we will reschedule it.

pcastonguay: Yes, makes sense to use `etcd` here.

pcastonguay: Do we want to get this merged before adding support for `etcd`? If so, we could remove the health check for now, and add it back later when networking team adds etcd support.

pcastonguay:
Closing. Superseded by https://github.com/NVIDIA/TensorRT-LLM/pull/3584. Dynamic addition/removal of servers will be added in separate MR.

PR #3320, opened 2025-04-07
chore: exchange connection id with tagSend/tagRecv

chuangz0:
Since Stream Send/Recv is unstable , we use tagSend/tagRecv and lock to replace.


Shixiaowei02 (approved):

chuangz0:
/bot skip --comment "the ci has passed while the reuse-pipeline fail"

PR #3322, opened 2025-04-07
perf: Optimize overhead due to get_mla_metadata before each FlashMLA kernel.

bobboli:
get_mla_metadata is slow and currently it is called before each FlashMLA kernel. 
Call it once per step rather than per attention layer, and pass in the metadata into AttentionOp.


cpp/tensorrt_llm/thop/flashMlaOp.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    return std::make_tuple(tile_scheduler_metadata, num_splits);
+}
+
+TORCH_LIBRARY_FRAGMENT(trtllm, m)
+{
+    m.def("get_mla_metadata(Tensor seqlens_k, int num_heads_per_head_k, int num_heads_k) -> (Tensor, Tensor)");
+}
+
+TORCH_LIBRARY_IMPL(trtllm, CUDA, m)
```
dongjiyingdjy, line 101: Can this op define move to `attentionOp.cpp?` Since it is the preprocess of attenton.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -24,6 +25,26 @@ class AttentionInputType(IntEnum):
...
+        Cached wrapper for torch.ops.trtllm.get_mla_metadata that ensures the function
+        is only called once for the same parameters.
+        """
+        # tensor is not directly hashable, so we use its ID as the cache key
+        # and pass the actual tensor to the function
+        return torch.ops.trtllm.get_mla_metadata(cache_seqlens,
+                                                 num_heads_per_head_k,
+                                                 num_heads_k)
+
+    return _get_mla_metadata_cached(id(cache_seqlens), num_heads_per_head_k,
```
yuxianq: Does this cache work? cache_seqlens is `metadata.kv_lens_cuda[:num_seqs]` in `TrtllmAttention.forward`, so `id(cache_seqlens)` is different for each layer. Thus we always cache miss in this case. Can you validate it?

yuxianq: How about use `(num_heads_per_head_k, num_heads_k)` as cache key and clear cache when `layer_idx == 0`? Thus this cache can reset for each new step.

bobboli: yes ur right, this doesn't work. Using a cache class and reset the cache at layer 0 is a better way, and we don't need hashing for caching at all.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -267,6 +299,18 @@ def plan(
                                                    dtype=torch.float32,
                                                    device="cuda")
 
+            if self.use_flash_mla:
+                # Clear cache at the start of a new step (layer_idx == 0)
+                if self.layer_idx == 0:
+                    flash_mla_cache.clear_cache()
```
dongjiyingdjy: Also needs to be called when MTP is enabled.

cpp/tensorrt_llm/common/attentionOp.h
```diff
@@ -379,6 +375,7 @@ class AttentionOp
     int32_t mSpecDecodingMaxGenerationLength = 1;
     bool mIsMLAEnabled = false;
     bool mUseFlashMLA = false;
+    int32_t mFlashMlaNumSmParts = 0;
```
yuxianq, line 380: Please help to add `mFP8GenerationMLA` and `mFlashMlaNumSmParts` to `data()` function. Every field in `AttentionOp` should be added to `data()` for correct hash value.

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -466,6 +471,12 @@ torch::Tensor attention(torch::Tensor q, torch::optional<torch::Tensor> k, torch
         // mNumKVHeads/mHeadSize are overwritten in common/attentionOp.cpp.
         op->mNumKVHeads = 1;
         op->mHeadSize = op->mMLAParams.kv_lora_rank + op->mMLAParams.qk_rope_head_dim;
+        op->mUseFlashMLA = use_flash_mla;
+        if (op->mUseFlashMLA)
+        {
+            TLLM_CHECK(tile_scheduler_metadata.has_value());
+            op->mFlashMlaNumSmParts = tile_scheduler_metadata.value().size(0);
```
yuxianq: Is it necessary to set `mFlashMlaNumSmParts` here? It will be assigned in op's `initialize`.

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -639,6 +650,9 @@ TORCH_LIBRARY_FRAGMENT(trtllm, m)
         ", int? v_head_dim"
         ", Tensor? mrope_rotary_cos_sin"
         ", Tensor? mrope_position_deltas"
+        ", bool use_flash_mla"
+        ", Tensor? tile_scheduler_metadata"
+        ", Tensor? num_splits"
```
yuxianq: Can you prefix `tile_scheduler_metadata` and `num_splits` with `mla_` to make it clear that they are for MLA. Just like `mrope_rotary_cos_sin` and `mrope_position_deltas`.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
```diff
@@ -0,0 +1,117 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * reference: https://github.com/deepseek-ai/FlashMLA
+ */
+
+// Adapted from https://github.com/Dao-AILab/flash-attention/blob/main/csrc/flash_attn/flash_api.cpp
```
yuxianq: Is this license correct? What is the relationship between our code and `flash_api.cpp`? Can we use a normal Apache license like `cpp/tensorrt_llm/thop/attentionOp.cpp`?

cpp/tensorrt_llm/thop/flashMlaOp.cpp
```diff
@@ -0,0 +1,117 @@
...
+ ******************************************************************************/
+
+#include <ATen/cuda/CUDAContext.h>
+#include <c10/cuda/CUDAGuard.h>
+#include <torch/nn/functional.h>
+#include <torch/python.h>
+
+#include <cutlass/fast_math.h>
+
+#include "../kernels/flashMLA/flash_mla.h"
```
yuxianq: Use `#include "tensorrt_llm/kernels/flashMLA/flash_mla.h"` instead.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+#define CHECK_DEVICE(x) TORCH_CHECK(x.is_cuda(), #x " must be on CUDA")
+#define CHECK_SHAPE(x, ...)                                                                                            \
+    TORCH_CHECK(x.sizes() == torch::IntArrayRef({__VA_ARGS__}), #x " must have shape (" #__VA_ARGS__ ")")
+#define CHECK_CONTIGUOUS(x) TORCH_CHECK(x.is_contiguous(), #x " must be contiguous")
```
yuxianq: There are already `CHECK_TH_CUDA/CHECK_CONTIGUOUS` in `cpp/tensorrt_llm/thop/thUtils.h`. And `CHECK_SHAPE` is unused here.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
```diff
@@ -0,0 +1,117 @@
...
+    at::Tensor& seqlens_k, int64_t const num_heads_per_head_k, int64_t const num_heads_k)
+{
+    // This should match the logic in the MLA kernel.
+    static constexpr int block_size_m = 64;
+    static constexpr int block_size_n = 64;
+    static constexpr int fixed_overhead_num_blocks = 5;
+
+    CHECK_DEVICE(seqlens_k);
+    TORCH_CHECK(seqlens_k.is_contiguous());
+    TORCH_CHECK(seqlens_k.dtype() == torch::kInt32);
```
yuxianq: Can use `CHECK_INPUT` instead of these three lines.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include "../kernels/flashMLA/flash_mla.h"
+
+#define CHECK_DEVICE(x) TORCH_CHECK(x.is_cuda(), #x " must be on CUDA")
+#define CHECK_SHAPE(x, ...)                                                                                            \
+    TORCH_CHECK(x.sizes() == torch::IntArrayRef({__VA_ARGS__}), #x " must have shape (" #__VA_ARGS__ ")")
+#define CHECK_CONTIGUOUS(x) TORCH_CHECK(x.is_contiguous(), #x " must be contiguous")
+
+/**
+ * @brief Get metadata for MLA computation
+ * @param cache_seqlens Tensor of shape (batch_size) containing sequence lengths, dtype int32
```
yuxianq: `cache_seqlens` should be `seqlens_k`

cpp/tensorrt_llm/thop/flashMlaOp.cpp
```diff
@@ -0,0 +1,117 @@
...
+ * @param num_heads_per_head_k Equal to seq_len_q * num_heads_q / num_heads_k
+ * @param num_heads_k Number of key heads
+ * @return Tuple containing:
+ *         - tile_scheduler_metadata: Tensor of shape (num_sm_parts, TileSchedulerMetaDataSize), dtype int32
+ *         - num_splits: Tensor of shape (batch_size + 1), dtype int32
+ */
+std::tuple<at::Tensor, at::Tensor> get_mla_metadata(
+    at::Tensor& seqlens_k, int64_t const num_heads_per_head_k, int64_t const num_heads_k)
+{
+    // This should match the logic in the MLA kernel.
```
yuxianq: Please point to the exact file or class or even fields.

jenkins/license_cpp.json
```diff
@@ -46,6 +46,7 @@
         "tensorrt_llm/kernels/flashMLA/static_switch.h": "dual license",
         "tensorrt_llm/kernels/flashMLA/utils.h": "dual license",
         "tensorrt_llm/kernels/flashMLA/fp8_transpose_v.h": "dual license",
+        "tensorrt_llm/thop/flashMlaOp.cpp": "dual license",
```
yuxianq, line 49: `flashMlaOp.cpp` is new code you write, why we need to use dual license for it?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -24,6 +24,34 @@ class AttentionInputType(IntEnum):
...
+
+    def clear_cache(self):
+        """
+        Clear the cached metadata.
+        """
+        self._cache = None
+
+
+# Create a single instance of the cache
+flash_mla_cache = FlashMLAMetadataCache()
```
yuxianq: If cache key is unnecessary, can we use `_mla_metadata = get_mla_metadata(...)` and reset it using `_mla_metadata = None` instead of creating a `FlashMLAMetadataCache` class?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -247,7 +277,9 @@ def plan(
             'mrope_position_deltas') if mrope_config is not None else None
         self.kwargs.update(kwargs)
         self.block_ids_per_seq = block_ids_per_seq
-
+        self.use_flash_mla = use_flash_mla and self.attention_input_type == AttentionInputType.generation_only
```
yuxianq: Can we move the `nd self.attention_input_type == AttentionInputType.generation_only` to the forward function when we call `plan`?

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: This cache should be thread-local. Please see `get_metadata` in `tensorrt_llm/_torch/attention_backend/flashinfer.py` for how to cache thread-local info.

cpp/tensorrt_llm/thop/attentionOp.cpp
bobboli: The one in `initialize` is unnecessary. It should be calculated in `get_mla_metadata`

yuxianq: We cache `op` for reusing, if we update `mFlashMlaNumSmParts` for every step, we may re-initialize `op` for different steps, which is unexpected. Will `tile_scheduler_metadata.value().size(0)` change for each step? If it is unchanged, calculate `mFlashMlaNumSmParts` without depending on per-step tensor `tile_scheduler_metadata` is better.

bobboli: Where is it cached? Isn't a new AttentionOp created for each torch attention() call? https://github.com/NVIDIA/TensorRT-LLM/blob/d0671494cdb9a73cb38bb4198fd5d4b94b3edf47/cpp/tensorrt_llm/thop/attentionOp.cpp#L423

yuxianq: It is cached here: https://github.com/NVIDIA/TensorRT-LLM/blob/515dd0d78fe8f5f86f8a655fc7bac36a6f8da7ac/cpp/tensorrt_llm/thop/attentionOp.cpp#L474

We create the un-initialized op to calculate the cache key by `op->data()` and `runner->data()`. If the cache misses, we initialize the op; if the cache hits, we use the cached op. So we can avoid the repeated op initialization. In most case (the attention config is same for all layers), we only initialize the op once.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
bobboli: It basically extracts `get_mla_metadata` from `flash_api.cpp` since we need to use this function on the Python side. The `mha_fwd_kvcache_mla` does not need to be exposed as an Op since its used in C++ `AttentionOp::mlaGeneration`.
Since it partially uses the code from `flash_api.cpp` I think dual license is still necessary.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
bobboli: These macros are removed.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
bobboli: Well this is a comment from the original repo. I marked the path to the kernel.

jenkins/license_cpp.json
bobboli: As mentioned, `get_mla_metadata` is from FlashMLA repo even though it's a very simple function.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
yuxianq: I cannot find `get_mla_metadata` in https://github.com/Dao-AILab/flash-attention/blob/main/csrc/flash_attn/flash_api.cpp, is it updated? Can you point to the original version with commit ID in the URL?

cpp/tensorrt_llm/thop/flashMlaOp.cpp
yuxianq: Got it.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
bobboli: ohoh I mean in FLashMLA repo https://github.com/deepseek-ai/FlashMLA/blob/b31bfe72a83ea205467b3271a5845440a03ed7cb/csrc/flash_api.cpp#L21

yuxianq: Good, please update the URL.

cpp/tensorrt_llm/thop/attentionOp.cpp
bobboli: > Will tile_scheduler_metadata.value().size(0) change for each step?

No, `tile_scheduler_metadata.value().size(0)` is `num_sm_parts = sm_count / num_heads_k / cutlass::ceil_div(num_heads_per_head_k, block_size_m);` which only depends on model structure (and number of predicted tokens in MTP). This calculation is done in `get_mla_metadata` so we do not need to do again in initialize().  Thus, making `mFlashMlaNumSmParts` as part of AttentionOp hash is okay.

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Okay, will do 👍. Just curious are there possible race conditions if non thread-local? I thought we were using processes for different ranks.

yuxianq: Different ranks have not data race, but we have unit tests that may run simultaneously in the same process. Use thread-local variable makes it safe by default.

cpp/tensorrt_llm/thop/attentionOp.cpp
yuxianq: Good, then only one of the `initialize` part and the `tile_scheduler_metadata` part is needed. We can safely remove the `intialize` part.

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -639,6 +652,9 @@ TORCH_LIBRARY_FRAGMENT(trtllm, m)
         ", int? v_head_dim"
         ", Tensor? mrope_rotary_cos_sin"
         ", Tensor? mrope_position_deltas"
+        ", bool use_flash_mla"
+        ", Tensor? flash_mla_tile_scheduler_metadata"
+        ", Tensor? flash_mla_num_splits"
```
yuxianq, line 676: Please also update the `@torch.library.register_fake("trtllm::attention")` part in `tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py`.

cpp/tensorrt_llm/thop/flashMlaOp.cpp
```diff
@@ -0,0 +1,108 @@
...
+    params.fixed_overhead_num_blocks = fixed_overhead_num_blocks;
+    params.num_sm_parts = num_sm_parts;
+    get_mla_metadata_func(params, stream);
+
+    return std::make_tuple(tile_scheduler_metadata, num_splits);
+}
+
+TORCH_LIBRARY_FRAGMENT(trtllm, m)
+{
+    m.def("get_mla_metadata(Tensor seqlens_k, int num_heads_per_head_k, int num_heads_k) -> (Tensor, Tensor)");
```
yuxianq: Please add a `@torch.library.register_fake("trtllm::get_mla_metadata")` to `tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py`.

yuxianq (approved):

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -468,9 +520,26 @@ def __post_init__(self) -> None:
...
+            # TODO(lbo): Management of block ids should not be coupled with flash_mla so tightly.
             self.block_ids_per_seq = None
             self.kv_block_ids_per_seq = None
+
+            self.flash_mla_metadata = FlashMlaMetadata()
             if self.enable_flash_mla:
+                tile_scheduler_metadata_size = 8  # Note: must be aligned with TileSchedulerMetaDataSize in flash_mla.h
+                block_size_m = 64
+                sm_count = get_sm_count()
+                # TODO begin: get from extra_attrs
```
bobboli: How should we get required info from extra_attrs here? @yuxianq

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -14,6 +16,14 @@
 from .vanilla import VanillaAttention
 
 
+# The type of requests in qkv passed to attention
+# Please keep sync with AttentionInputType in cpp/tensorrt_llm/thop/attentionOp.cpp
+class AttentionInputType(IntEnum):
```
yuxianq: `AttentionInputType` has moved to `tensorrt_llm/_torch/attention_backend/interface.py`, we don't need to re-define it here.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -219,12 +235,36 @@ def plan(
...
             self.rope_params.max_positions = max_sequence_length
             self.rotary_inv_freq, self.rotary_cos_sin = self.rope_params.create_rope_const_params(
             )
 
+        self.use_flash_mla = use_flash_mla
+        self.flash_mla_tile_scheduler_metadata = None
+        self.flash_mla_num_splits = None
+        if self.is_mla_enable:
+            # max_context_length will increment 1 when overlap scheduler enabled
+            if self.max_context_length > (self.rotary_cos_sin.shape[1] /
```
yuxianq: This part is unnecessary, we have used `self.rope_params.create_rope_const_params()` instead in L240.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -219,12 +235,36 @@ def plan(
...
+                    self.mla_rope_params.beta_fast,
+                    self.mla_rope_params.beta_slow,
+                    self.mla_rope_params.mscale,
+                    self.mla_rope_params.mscale_all_dim,
+                )
+                self.rotary_cos_sin = torch.tensor(rope_cos_sin,
+                                                   dtype=torch.float32,
+                                                   device="cuda")
+            if self.use_flash_mla:
+                self.flash_mla_tile_scheduler_metadata = flash_mla_tile_scheduler_metadata
```
yuxianq: Is it necessary to guard `self.flash_mla_tile_scheduler_metadata = flash_mla_tile_scheduler_metadata` with `if self.use_flash_mla` and `if self.is_mla_enable`? The `flash_mla_tile_scheduler_metadata` argument is None by default. So as `flash_mla_num_splits`.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -432,6 +481,9 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
 
     def __post_init__(self) -> None:
         super().__post_init__()
+
+        # get_model_extra_attrs()
```
yuxianq: Remove this line if unused.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -504,7 +582,7 @@ def prepare(self) -> None:
             # set params that are used in wrapper.plan()
             self.kv_cache_block_offsets = None
             self.host_kv_cache_block_offsets = None
-            self.block_ids_per_seq = None
+            self.block_ids_per_seq = None  # TODO(lbo): this seems to be conflicting with block_ids_per_seq managed by flash_mla path
```
yuxianq: There is no conflict. `self.block_ids_per_seq = None` only happens when `self.kv_cache_manager is None`, but `self.kv_cache_manager is not None` for FlashMLA path.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -468,9 +499,26 @@ def __post_init__(self) -> None:
...
                 pin_memory=True,
             )
+
+            # TODO: Management of block ids should not be coupled with flash_mla so tightly.
             self.block_ids_per_seq = None
             self.kv_block_ids_per_seq = None
+
+            self.flash_mla_metadata = FlashMlaMetadata()
             if self.enable_flash_mla:
+                tile_scheduler_metadata_size = 8  # Note: must be aligned with TileSchedulerMetaDataSize in flash_mla.h
```
dongjiyingdjy, line 500: flashmla meadata should be updated in MTP

cpp/tensorrt_llm/thop/flashMlaOp.cpp
yuxianq: `flashMlaOp.cpp` is a dual license file coming from FlashMLA, moving to `attentionOp.cpp` may infect the license of  `attentionOp.cpp`.

dongjiyingdjy: yes, you are right.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: Add a line in `PyTorchModelEngine.__init__` after https://github.com/NVIDIA/TensorRT-LLM/blob/d4c0423cdb69b709f481c06c68487859a3cff0fa/tensorrt_llm/_torch/pyexecutor/model_engine.py#L268
```
self.model.extra_attrs['model_config'] = self.model.model_config
```
Then you can access `model_config` here:
```
model_config = get_model_extra_attrs()['model_config']
predicted_tokens_per_seq = model_config.spec_config.num_nextn_predict_layers + 1 if model_config.spec_config is not None else 1
num_heads_q = model_config.pretrained_config.num_attention_heads
num_heads_k = model_config.pretrained_config.num_key_value_heads
```

Since we can get `predicted_tokens_per_seq` here, can we remove `predicted_tokens_per_seq` from MLAParams/create_attention/DeepseekV3Attention and so on?

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: ok, this is mis-introduced during merge.

tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -953,6 +953,10 @@ def forward(
             req_types = attn_metadata.host_request_types[:batch_size].clone()
             if hasattr(attn_metadata, 'kv_lens_cuda'):
                 kv_lens_cuda = attn_metadata.kv_lens_cuda[:batch_size].clone()
+            if hasattr(attn_metadata,
+                       'flash_mla_metadata') and attn_metadata.enable_flash_mla:
+                # tile_scheduler_metadata_eagle has been pre-allocated. Swap to use it.
+                attn_metadata.flash_mla_metadata.tile_scheduler_metadata, attn_metadata.flash_mla_metadata.tile_scheduler_metadata_eagle = attn_metadata.flash_mla_metadata.tile_scheduler_metadata_eagle, attn_metadata.flash_mla_metadata.tile_scheduler_metadata
```
yuxianq, line 1015: Don't need to store `tile_scheduler_metadata_eagle` inside `attn_metadata`, we can use `tile_scheduler_metadata = attn_metadata.flash_mla_metadata.tile_scheduler_metadata.clone()` here and use `attn_metadata.flash_mla_metadata.tile_scheduler_metadata.copy_(tile_scheduler_metadata)` to restore it.

bobboli: The problem is that the shape of tile_scheduler_metadata_eagle and tile_scheduler_metadata might be different. See [here](https://github.com/bobboli/TensorRT-LLM/blob/70d95ec3b052925ceba303f02517ca0e06585a4d/tensorrt_llm/_torch/attention_backend/trtllm.py#L539).

yuxianq: Got it. Since `tile_scheduler_metadata_eagle` is unrelated to the base model, should we put it in MTP's `spec_metadata` instead of the base model's `attn_metadata`?

bobboli: Let's refactor in another PR. This one has been too long.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -479,17 +524,38 @@ def __post_init__(self) -> None:
...
                     device='cuda',
                 )
+                self.flash_mla_metadata.tile_scheduler_metadata = torch.empty(
+                    [num_sm_parts, tile_scheduler_metadata_size],
+                    dtype=torch.int32,
+                    device='cuda',
+                )
 
-    def prepare(self) -> None:
+                # This is specific to MTP Eagle, where the query seqlen is always 1.
```
lfr-0531, line 542: If this tensor is only for MTP, could you move it to `MTPSpecMetadata.__post_init__()`? 

We want all of the MTP-related codes to only appear in `mtp.py`, so that we can decouple the different speculative decoding methods with other parts.

tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -986,19 +990,28 @@ def forward(
...
                                                  num_contexts].fill_(1)
                 attn_metadata.num_contexts = 0
-                if i == 0 and num_contexts > 0 and attn_metadata.enable_flash_mla:
-                    reorder_block_ids_per_seq = torch.cat([
-                        attn_metadata.
-                        kv_block_ids_per_seq[num_contexts:batch_size],
-                        attn_metadata.kv_block_ids_per_seq[:num_contexts]
-                    ])
+
+                if attn_metadata.enable_flash_mla:
```
lfr-0531, line 998: Put together the two parts related to flashmla:
```
            if hasattr(attn_metadata, 'kv_lens_cuda'):
                attn_metadata.kv_lens_cuda[:batch_size] += 1
            # cannot run generation if their is no kv cache
            if inputs["attn_metadata"].kv_cache_manager is not None:
                # Turn context request to generation request
                attn_metadata.host_request_types[:attn_metadata.
                                                 num_contexts].fill_(1)
                attn_metadata.num_contexts = 0

                if attn_metadata.enable_flash_mla:
                    # block_ids_per_seq is used by FlashMLA and needs to be updated.
                    attn_metadata.block_ids_per_seq[:batch_size, :].copy_(
                        attn_metadata.kv_block_ids_per_seq[:batch_size, :],
                        non_blocking=True)
                    # Since KV cache length is increasing, we need to update the MLA metadata.
                    torch.ops.trtllm.get_mla_metadata(
                        attn_metadata.kv_lens_cuda[:batch_size], attn_metadata.
                        flash_mla_metadata.tile_scheduler_metadata,
                        attn_metadata.flash_mla_metadata.
                        num_splits[:batch_size + 1])
```

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Let's refactor in a separate PR. Lots of things need to be refactored.

PR #3326, opened 2025-04-07
fix: Fix p-tuning test bug

amirkl94:
A change in the vocab_size calculation for T5Tokenizer, introduced in transformers version 4.34, caused addition of incorrect vtokens for ptuning. In general, instead of adding tokens which are outside the vocabulary, tokens inside the vocabulary were added.


Naveassaf (approved):
LGTM

PR #3339, opened 2025-04-07
chore: Unwaive DS + overlap disagg test

pcastonguay:
The disaggregated serving test with DS + overlap was waived because it was failing. I re-enabled it but disabled MTP because it looks like the combination DS+overlap+MTP is not giving the correct transcripts. Filed a bug to fix that but at least adding coverage for DS with overlap.

### DIFF
--- a/tests/integration/defs/disaggregated/test_configs/disagg_config_ctxtp2_gentp2_deepseek_v3_lite_attention_dp_overlap.yaml
+++ b/tests/integration/defs/disaggregated/test_configs/disagg_config_ctxtp2_gentp2_deepseek_v3_lite_attention_dp_overlap.yaml
@@ -1,19 +1,13 @@
-model: DeepSeek-V3-Lite/fp8
 hostname: localhost
 port: 8000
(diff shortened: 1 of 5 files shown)


Tabrizian (approved):

Shixiaowei02 (approved):

tests/integration/defs/disaggregated/test_configs/disagg_config_ctxtp2_gentp2_deepseek_v3_lite_attention_dp_overlap.yaml
```diff
@@ -1,19 +1,13 @@
-model: DeepSeek-V3-Lite/fp8
 hostname: localhost
 port: 8000
+model: DeepSeek-V3-Lite/fp8
+free_gpu_memory_fraction: 0.25
 backend: "pytorch"
-free_gpu_memory_fraction: 0.2
-speculative_config:
-  decoding_type: MTP
```
Shixiaowei02, line 7: May I ask the reason for removing MTP here?

PR #3346, opened 2025-04-07
feat: Add numNodes to ParallelConfig

achartier:
If not provided, attempt to find the number of nodes by adding the number of local ranks 0

Update device IDs check accordingly

### DIFF
--- a/tests/unittest/bindings/test_executor_bindings.py
+++ b/tests/unittest/bindings/test_executor_bindings.py
@@ -1633,6 +1633,25 @@ def test_parallel_config():
     assert parallel_config.orchestrator_config.spawn_processes == True
 
 
+def test_parallel_config_pickle():
+    comm_type = trtllm.CommunicationType.MPI
+    comm_mode = trtllm.CommunicationMode.LEADER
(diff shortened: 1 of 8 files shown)


tests/unittest/bindings/test_executor_bindings.py
```diff
@@ -1595,12 +1595,17 @@ def test_parallel_config():
...
+    parallel_config = trtllm.ParallelConfig(comm_type,
+                                            comm_mode,
+                                            device_ids,
+                                            participant_ids,
+                                            num_nodes=num_nodes)
     assert parallel_config.communication_type == comm_type
     assert parallel_config.communication_mode == comm_mode
     assert parallel_config.device_ids == device_ids
     assert parallel_config.participant_ids == participant_ids
+    assert parallel_config.num_nodes == num_nodes
```
Funatiq, line 1619: Could you please add a `test_parallel_config_pickle` similar to the other tests?

achartier: Added test.

Funatiq (approved):

PR #3239, opened 2025-04-02
feat: use NVRTC for DeepGEMM JIT compilation

lucifer1004:

PR #3351, opened 2025-04-08
fix: revert extra cmake var

lucifer1004:
I added a local cmake var in test_cpp.py but incorrectly committed it in #3239 . This PR reverts the extra cmake var.


QiJune (approved):
LGTM

PR #3362, opened 2025-04-08
chore : fix estimated kv cache memory when cuda graph is enabled

peaceh-nv:
More non-torch extra memory would be introduced if cuda_graph is enabled, make the estimated kv cache memory not precise. After this fix, the `kv_cache_free_gpu_mem_fraction` can be as large as the one without cuda graph.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -132,6 +132,8 @@ def estimate_max_kv_cache_tokens(model_engine: PyTorchModelEngine,
         # Get the torch-managed peak memory
         torch_peak_memory = torch.cuda.memory_stats(
         )["allocated_bytes.all.peak"]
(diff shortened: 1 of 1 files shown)


jiahanc (approved):
LGTM

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -142,6 +144,18 @@ def estimate_max_kv_cache_tokens(model_engine: PyTorchModelEngine,
         total_used_bytes = total_gpu_memory - end
         # Get the non-torch memory which comes from NCCL and some other components
         extra_cost = max(total_used_bytes - torch_used_bytes, 0)
+        if model_engine._run_cuda_graphs:
+            extra_cost_cuda_graph = 0
+            # Each graph needs 60MB memory
```
kaiyux, line 149: Are we sure that the size will always be around 60MB regardless of model and workload?

PR #3364, opened 2025-04-08
test: add cuda visible device constraint for phi_1gpu test

crazydemo:
add cuda visible device constraint for phi_1gpu test.
related nvbug: https://nvbugspro.nvidia.com/bug/5206572

### DIFF
--- a/tests/integration/defs/examples/test_phi.py
+++ b/tests/integration/defs/examples/test_phi.py
@@ -515,6 +515,7 @@ def test_llm_phi_quantization_1gpu(data_type, llm_phi_model_root, llm_venv,
                                    cmodel_dir, engine_dir, phi_example_root,
                                    llm_datasets_root, llm_rouge_root, qformat):
     "Run phi quantization tests"
+    # Workaround for Modelopt can't convert Phi-3 on multi GPUs.
     gpu_constraint = {"CUDA_VISIBLE_DEVICES": "0"}
 
     print("Convert checkpoint by modelopt...")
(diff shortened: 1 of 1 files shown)


crazydemo:
/bot run --stage-list "A10-PackageSanityCheck-manylinux-py310"

tests/integration/defs/examples/test_phi.py
```diff
@@ -515,6 +515,8 @@ def test_llm_phi_quantization_1gpu(data_type, llm_phi_model_root, llm_venv,
                                    cmodel_dir, engine_dir, phi_example_root,
                                    llm_datasets_root, llm_rouge_root, qformat):
     "Run phi quantization tests"
+    gpu_constraint = {"CUDA_VISIBLE_DEVICES": "0"}
```
kaiyux, line 519: Sorry I'm a bit confused - why do we need such constraint?

crazydemo: this constraint is related to a [flaky issue](https://nvbugspro.nvidia.com/bug/5206572). we will encounter cuda error when this case is running on a multi-gpu machine.

kaiyux: @brb-nv Any plan to remove the WAR?

brb-nv: Hi Kaiyu, tbh I don't have bandwidth to look into this issue right now. I think it's low impact because the tests are meant to be run on a single GPU anyway. So, mentioning CUDA_VISIBLE_DEVICES=0 is not too bad a WAR imo. Please let me know if you think otherwise.

kaiyux: Sure, I agree to unblock this. I guess if we need more and more such WARs in test code then it could be a problem. Thanks!

kaiyux, line 519: @crazydemo Sorry, can you add a comment in the code saying that this is a WAR? I'm concerned others might introduce the same behavior in other tests without knowing the context.

We don't need to rerun the pipeline for only comments, I can merge afterwards.

kaiyux (approved):

tests/integration/defs/examples/test_phi.py
crazydemo: comment added.

PR #3369, opened 2025-04-08
feat: add qwen2 moe to torch flow; fix wrong imported KvCacheConfig in gpqa…

wm2012011492:
1. add qwen2 moe to torch flow; 
2. fix wrong imported KvCacheConfig in gpqa_llmapi.py

Local accuracy test for [Qwen2-57B-A14B](https://huggingface.co/Qwen/Qwen2-57B-A14B):
GPQA: Average accuracy: 0.354
MMLU weighted average accuracy: 75.72 (14042)

TODO：
Add accuracy unittest after new version is released

### DIFF
--- a/examples/gpqa_llmapi.py
+++ b/examples/gpqa_llmapi.py
@@ -61,9 +61,8 @@ from transformers import AutoTokenizer
 from tensorrt_llm import SamplingParams
 from tensorrt_llm._torch.llm import LLM as PyTorchLLM
 from tensorrt_llm._torch.pyexecutor.config import PyTorchConfig
-from tensorrt_llm.bindings.executor import KvCacheConfig
 from tensorrt_llm.builder import BuildConfig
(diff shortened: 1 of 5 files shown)


byshiue:
You should require add something like https://github.com/NVIDIA/TensorRT-LLM/blob/215fb205677433b116b2a9278ebfeb78607cb5fc/tests/integration/test_lists/test-db/l0_h100.yml#L21

If it is not necessary to run the model on h100, please consider a30 and a10 first, then a100, then h100.

wm2012011492:
> You should require add something like
> 
> https://github.com/NVIDIA/TensorRT-LLM/blob/215fb205677433b116b2a9278ebfeb78607cb5fc/tests/integration/test_lists/test-db/l0_h100.yml#L21
> 
> If it is not necessary to run the model on h100, please consider a30 and a10 first, then a100, then h100.

Done

byshiue (approved):

PR #3370, opened 2025-04-08
refactor: remove ParallelConfig in tensorrt_llm._torch.distributed module

QiJune:
To avoid unnecessary cyclical conversion "Mapping -> ParallelConfig -> Mapping"


tensorrt_llm/_torch/modules/gated_mlp.py
```diff
@@ -75,13 +74,12 @@ def __init__(self,
...
             bias=bias,
             dtype=dtype,
-            parallel_config=ParallelConfig(
-                tensor_parallel_rank=tp_rank,
-                tensor_parallel_size=tp_size,
-                tensor_parallel_mode=TensorParallelMode.ROW,
-                gpus_per_node=gpus_per_node,
-                pipeline_parallel_size=pp_size,
-                parallel_rank=config.mapping.rank),
+            mapping=Mapping(world_size=tp_size * pp_size,
```
hlu1: This mapping is the same as the one on line 60. You can create one Mapping object and use it in both places.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -79,12 +81,15 @@ def __init__(
...
         config = config or ModelConfig()
         tp_size = config.mapping.tp_size
-        tp_rank = config.mapping.tp_rank
+        pp_size = config.mapping.pp_size
+        rank = config.mapping.rank
         gpus_per_node = config.mapping.gpus_per_node
         if config.mapping.enable_attention_dp:
             tp_size = 1
-            tp_rank = 0
+            pp_size = 1
```
hlu1: Why is `pp_size` set to 1 here? You can have pp > 1 with attn_dp.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -119,13 +125,14 @@ def __init__(
             self.hidden_size,
             bias=self.dense_bias,
             dtype=dtype,
-            parallel_config=ParallelConfig(
-                tensor_parallel_size=tp_size,
-                tensor_parallel_rank=tp_rank,
-                tensor_parallel_mode=TensorParallelMode.ROW,
+            mapping=Mapping(
```
hlu1: Use the same Mapping object created on line 110

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -279,27 +286,20 @@ def __init__(
...
-            tensor_parallel_mode=TensorParallelMode.ROW,
-            gpus_per_node=gpus_per_node,
-            pipeline_parallel_size=config.mapping.pp_size,
-            parallel_rank=config.mapping.rank,
-        )
-        col_parallel_config = ParallelConfig(
-            tensor_parallel_rank=tp_rank,
-            tensor_parallel_size=tp_size,
-            tensor_parallel_mode=TensorParallelMode.COLUMN,
+            pp_size = 1
```
hlu1: Ditto

tensorrt_llm/_torch/modules/gated_mlp.py
QiJune: Done

tensorrt_llm/_torch/modules/attention.py
QiJune: Done

tensorrt_llm/_torch/modules/attention.py
QiJune: Done, revert this change.

tensorrt_llm/_torch/modules/attention.py
QiJune: Done, revert this change

tensorrt_llm/mapping.py
```diff
@@ -262,10 +262,12 @@ def rank(self):
...
-        if not isinstance(rank, int) or rank < 0 or rank >= self.world_size:
-            raise ValueError(
-                f"Rank should be an integer between 0 and {self.world_size-1}, but got {rank}."
-            )
+        if not isinstance(rank, int) or rank < 0:
+            # TODO(qijun): support attention_dp_size
+            if not self.enable_attention_dp and rank >= self.world_size:
+                raise ValueError(
+                    f"Rank should be an integer between 0 and {self.world_size-1}, but got {rank}."
+                )
```
amukkara, line 271: this change seems erroneous to me. can you split the check for attention dp on/off cleanly:

```
if enable_attention_dp:
    if ..:
       raise()
else:
   if ..:
       raise()
```

tensorrt_llm/_torch/distributed.py
```diff
@@ -242,34 +183,26 @@ def reducescatter(input: torch.Tensor,
...
         super().__init__()
 
-        self.parallel_config = parallel_config
-        self.tp_size = self.parallel_config.tensor_parallel_size
-        self.rank = self.parallel_config.parallel_rank
-        self.gpus_per_node = self.parallel_config.gpus_per_node
+        self.mapping = mapping
+        self.tp_size = self.mapping.tp_size
+        self.rank = self.mapping.rank
+        self.gpus_per_node = self.mapping.gpus_per_node
```
amukkara: why make a copy (for `tp_size`, `rank`, `gpus_per_node`) instead of directly accessing from `self.mapping`?

tensorrt_llm/_torch/distributed.py
```diff
@@ -281,30 +214,23 @@ def forward(
...
-        self.parallel_config = parallel_config
-        self.tp_size = self.parallel_config.tensor_parallel_size
-        self.tp_rank = self.parallel_config.tensor_parallel_rank
-        self.gpus_per_node = self.parallel_config.gpus_per_node
-        self.rank = self.parallel_config.parallel_rank
+        self.mapping = mapping
+        self.tp_size = self.mapping.tp_size
+        self.tp_rank = self.mapping.tp_rank
+        self.gpus_per_node = self.mapping.gpus_per_node
+        self.rank = self.mapping.rank
```
amukkara: same comment on making copies.

tensorrt_llm/_torch/modules/mamba.py
```diff
@@ -166,9 +165,10 @@ def __init__(
         super().__init__()
 
         config = config or ModelConfig()
+        self.mapping = config.mapping
         tp_rank = config.mapping.tp_rank
         tp_size = config.mapping.tp_size
-        gpus_per_node = config.mapping.gpus_per_node
+        config.mapping.gpus_per_node
```
amukkara: delete unused line.

tensorrt_llm/mapping.py
```diff
@@ -191,9 +191,9 @@ def __init__(
         self.attn_cp_size = attn_cp_size
         self.auto_parallel = auto_parallel
         self.world_size = world_size
+        self.enable_attention_dp = enable_attention_dp
         self.rank = rank
         self.gpus_per_node = gpus_per_node
-        self.enable_attention_dp = enable_attention_dp
```
amukkara, line 196: why move `enable_attention_dp` before `rank`?

tensorrt_llm/_torch/modules/linear.py
```diff
@@ -134,7 +143,9 @@ def __init__(self,
                  out_features: int,
                  bias: bool = True,
                  dtype: torch.dtype = None,
-                 parallel_config: Optional[ParallelConfig] = None,
+                 mapping: Optional[Mapping] = None,
+                 gather_output: bool = False,
+                 tensor_parallel_mode: Optional[TensorParallelMode] = None,
```
amukkara: since only `mapping` and `tensor_parallel_mode` are used in most cases, does it make sense to change the order
to `mapping`, `tensor_parallel_mode`, `gather_output`? We should keep class definition and instantiation order same whenever possible.

Please check the order is consistent across all definitions in `_torch/modules` (e.g `_torch/modules/embedding.py`) and their instantiations.

tests/unittest/_torch/multi_gpu/test_deepseek_allreduce.py
```diff
@@ -283,10 +283,11 @@ def moe_residual_norm_fusion_forward(
...
         dtype=dtype,
-        parallel_config=ParallelConfig(
-            tensor_parallel_size=tensor_parallel_size,
-            tensor_parallel_rank=tensor_parallel_rank,
-            tensor_parallel_mode=TensorParallelMode.ROW,
+        tensor_parallel_mode=TensorParallelMode.ROW,
+        mapping=Mapping(
+            world_size=tensor_parallel_size,
+            tp_size=tensor_parallel_size,
+            rank=tensor_parallel_rank,
```
amukkara, line 289: nit: `tensor_parallel_mode` after `mapping`

SimengLiu-nv (approved):

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -279,27 +277,17 @@ def __init__(
         # tensor parallel
         config = config or ModelConfig()
         tp_size = config.mapping.tp_size
-        tp_rank = config.mapping.tp_rank
-        gpus_per_node = config.mapping.gpus_per_node
+        pp_size = config.mapping.pp_size
         if config.mapping.enable_attention_dp:
             tp_size = 1
-            tp_rank = 0
```
hlu1, line 317: Is the rank in Mapping correct for attn_dp = True? Mapping doesn't modify rank based on enable_attention_dp.

hlu1: As a follow-up, should we move the attention_dp logics to inside Mapping? It currently accepts a `enable_attention_dp` flag but does nothing.

https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/mapping.py#L129

QiJune: In our current implementation, we only have a bool `enable_attention_dp` flag.  Actually, `enable_attention_dp=True` means `attention_tp_size=1`, and `attention_dp_size=tp_size`. Thus, `world_size` should be `pp_size * attention_tp_size(1) * attention_dp_size(tp_size)`, which keeps unchanged. In the future, we should remove `enable_attention_dp` field in Mapping, and replace it with  `attention_dp_size`  and `attention_tp_size` fields.


> Is the rank in Mapping correct for attn_dp = True? Mapping doesn't modify rank based on enable_attention_dp.

The rank is correct, but world_size is not correct now, since we don't have a `attention_dp_size` yet.


> should we move the attention_dp logics to inside Mapping?

No,  `attention_dp_size` and `attention_tp_size` still need to be decided first, and set to Mapping.

tensorrt_llm/mapping.py
QiJune: Please see here: https://github.com/NVIDIA/TensorRT-LLM/pull/3370#discussion_r2036589711

We don't have an int `attention_dp_size` now, it's a bool `enable_attention_dp`. So, the `world_size` is not correct when `enable_attention_dp=True`. We have some check logic in

```
    @rank.setter
    def rank(self, rank: int):
         xxx
```

So, I move `enable_attention_dp` before `rank` to ensure `enable_attention_dp` can be passed into the check logic.

This will be refactored once we support `attention_dp_size` in the future.

tensorrt_llm/mapping.py
QiJune: Done

tensorrt_llm/_torch/distributed.py
QiJune: removed

tensorrt_llm/_torch/modules/mamba.py
QiJune: Done

tensorrt_llm/_torch/modules/linear.py
QiJune: Done

tests/unittest/_torch/multi_gpu/test_deepseek_allreduce.py
QiJune: Done

tensorrt_llm/mapping.py
amukkara: makes sense. thanks.

hlu1 (approved):

PR #3371, opened 2025-04-08
feat: Add NVFP4 UB pattern optimization pass in torch compile

liji-nv:
* Add an additional flag for UB fp4 pattern to avoid inverse the scale
* Add NVFP4 related UB patterns

### DIFF
--- a/cpp/tensorrt_llm/kernels/userbuffers/userbuffers.cu
+++ b/cpp/tensorrt_llm/kernels/userbuffers/userbuffers.cu
@@ -568,7 +568,7 @@ __global__ void __launch_bounds__(MAX_THREADS)
 #if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
     using PackedVec = PackedVec<DType>;
     cudaTriggerProgrammaticLaunchCompletion();
-    float const sf = 1.f / *scale;
+    float sf = *scale;
     __shared__ float s_variance;
     int hidden_dim = blockDim.x * UNROLL_NLINES * sizeof(int4) / sizeof(DType);
 
@@ -683,7 +683,7 @@ __global__ void __launch_bounds__(MAX_THREADS)
 #if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
     using PackedVec = PackedVec<DType>;
     cudaTriggerProgrammaticLaunchCompletion();
-    float const sf = 1.f / *scale;
+    float sf = *scale;
     __shared__ float s_variance;
     int hidden_dim = blockDim.x * UNROLL_NLINES * sizeof(int4) / sizeof(DType);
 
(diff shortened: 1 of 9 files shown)


tensorrt_llm/_torch/compilation/patterns/ub_allreduce.py
```diff
@@ -224,6 +304,52 @@ def target_scaled_mm_prologue_pattern(
...
+                alpha: torch.Tensor,
+                sf_use_ue8m0: bool,
+                output_dtype: torch.dtype,
+            ):
+                nvfp4_gemm_output = torch.ops.trtllm.nvfp4_gemm(
+                    act_fp4, weight, act_sf, weight_scale, alpha, sf_use_ue8m0,
+                    output_dtype, True)
+                return nvfp4_gemm_output
+
+            # No extra check needed as the output dtype of scaled_mm has been verified when
```
liji-nv:
```suggestion
            # No extra check needed as the output dtype of nvfp4_gemm has been verified when
```

cpp/tensorrt_llm/kernels/userbuffers/userbuffers.cu
```diff
@@ -568,7 +568,12 @@ __global__ void __launch_bounds__(MAX_THREADS)
 #if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
     using PackedVec = PackedVec<DType>;
     cudaTriggerProgrammaticLaunchCompletion();
-    float const sf = 1.f / *scale;
+    float sf = *scale;
+    if constexpr (INVERSE_SCALE)
+    {
+        sf = 1.f / sf;
+    }
+
```
yilin-void: I wonder if the inverse logic is necessary, because during actual quantization, we always multiply by `1.f / scale_factor`. However, in FP4 quantization, this calculation of the reciprocal is placed in the `cvt_warp_fp16_to_fp4` function, making it appear as if the handling of FP4 and FP8 is reversed in the kernel.

yilin-void (approved):

tests/unittest/_torch/multi_gpu/test_user_buffers.py
```diff
@@ -812,3 +812,282 @@ def test_user_buffers_mm_add_prologue(hidden, tokens, dtype):
...
+        # 3 AR_NORM fusion happens first
+        # 2 AR_NORM fused with Quant
+        # 1 AR_NORM replacement
+        # 3 Scaled MM Prologue
+        # 2 UB Finalize Removal
+        assert backend.match_count == [3, 0, 2, 0, 1, 0, 3, 0, 2, 0]
+        torch.cuda.synchronize()
+        torch.testing.assert_close(output_fused,
+                                   output_ref,
+                                   atol=1.5,
```
liji-nv:
```suggestion
                                   atol=1.6,
```

PR #3373, opened 2025-04-08
Add and Remove server dynamically in PD

Shunkangz:


tensorrt_llm/serve/openai_disagg_server.py
```diff
@@ -81,6 +85,12 @@ def register_routes(self):
         self.app.add_api_route("/v1/completions",
                                self.openai_completion,
                                methods=["POST"])
+        self.app.add_api_route("/addserver",
```
Shixiaowei02, line 88: I suggest being cautious with interface changes. It seems that adding nodes (via launch workers) and removing them (by exiting with ctrl-c) can be accomplished without adding these two interfaces. Is my understanding correct?

Shunkangz: Yes. But we should let the proxy server be aware of the change. With the above two interfaces, I can let the proxy server be aware of the change immediately and avoid inconsistent state. Do you have any suggestion to deal with this?

Shixiaowei02: I am not very familiar with the common practices for the OpenAI API. Could we look for other references?

kaiyux: +1 on Xiaowei's concern - this file is also named as "openai_disagg_server", we should be careful when introducing anything that is not defined as OpenAI API.

Shunkangz: Let me think how to solve this problem without introducing new router function.

PR #3374, opened 2025-04-08
feat: Allow individual gatherContext for each additional output

Funatiq:
Summary: 
* Add support for a gatherContext flag for each additional output in ExecutorConfig.
* Enhance copyAdditionalOutputs function to work with AdditionalModelOutput.
* Add validation for final indices of additional output tensors.
* Update unit tests to verify the changes.
  * Cover the new gatherContext parameter.
  * Add lastTokenIds input tensor to test engines.
  * Make logits output dependent on gatherContextLogits parameter.

### DIFF
--- a/cpp/tensorrt_llm/batch_manager/runtimeBuffers.cpp
+++ b/cpp/tensorrt_llm/batch_manager/runtimeBuffers.cpp
@@ -53,12 +53,12 @@ RuntimeBuffers::RuntimeBuffers(SizeType32 maxBatchSize, SizeType32 maxBeamWidth,
     std::vector<SizeType32> const& maxAttentionWindowVec, SizeType32 maxAttentionWindow, SizeType32 sinkTokenLen,
     TllmRuntime const& runtime, ModelConfig const& modelConfig, WorldConfig const& worldConfig,
     executor::DecodingConfig const& decodingConfig, bool gatherGenerationLogits, std::optional<SizeType32> maxNumTokens,
-    std::optional<std::vector<std::string>> const& additionalOutputNames)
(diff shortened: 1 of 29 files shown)


cpp/tensorrt_llm/batch_manager/runtimeBuffers.cpp
jdebache, line 1: Unused headers: `"tensorrt_llm/runtime/runtimeKernels.h"` and `<valarray>`

cpp/tensorrt_llm/pybind/bindings.cpp
```diff
@@ -523,8 +523,7 @@ PYBIND11_MODULE(TRTLLM_PYBIND_MODULE, m)
         .def_readwrite("gpu_weights_percent", &tb::TrtGptModelOptionalParams::gpuWeightsPercent)
         .def_readwrite("max_beam_width", &tb::TrtGptModelOptionalParams::maxBeamWidth)
         .def_readwrite("scheduler_config", &tb::TrtGptModelOptionalParams::schedulerConfig)
-        .def(py::pickle(gptModelParamsGetState, gptModelParamsSetState))
-        .def("__eq__", &tb::TrtGptModelOptionalParams::operator==);
```
jdebache, line 527: Wasn't there a place where we actually use this equality operator? It sounds vaguely familiar.

Funatiq: It was used in the pickle test, but I got too annoyed to update it every time, so I changed the test too. IMO `TrtGptModelOptionalParams` shouldn't be exposed at all.

jdebache (approved):

PR #3375, opened 2025-04-08
Fix: Beam Search Diversity

wili-65535:
Fix the bug in "Beam Search + IFB" discussed in [NVBUG](https://nvbugspro.nvidia.com/bug/5077106).


byshiue:
@wili-65535  Can you help add a unit test to protect this function?

wili-65535:
> @wili-65535 Can you help add a unit test to protect this function?

OK but I have no idea to create such a unit test, do you have some suggestion?

The trigger condition for this bug is, some requests of the batch finish and exit, and new requests join (replace some places of the slot).
The code before this fix might use wrong slot index for beam search.

In addition, considering the deadline of release-0.19, could we merge the fix firstly and add the test later?

byshiue (approved):

PR #3378, opened 2025-04-08
feat: Cache sin cos in model instead of global LRU cache.

yuxianq:


tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -354,57 +357,74 @@ def from_config(config) -> "RopeParams":
             rope_params.beta_slow = rope_scaling.get("beta_slow", 1)
             rope_params.mscale = rope_scaling.get("mscale", 1.0)
             rope_params.mscale_all_dim = rope_scaling.get("mscale_all_dim", 0.0)
+        # Workaround for DeepSeek V3 Lite since its rope_scaling is null in config.json.
+        if config.model_type == "deepseek_v3":
```
kaiyux: How about this, so that non-lite case will not be overwritten?
```suggestion
        elif config.model_type == "deepseek_v3":
```

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -354,57 +357,74 @@ def from_config(config) -> "RopeParams":
...
+                self.dim,
+                self.theta,
+                self.scale,
+                self.original_max_positions,
+                self.beta_fast,
+                self.beta_slow,
+                self.mscale,
+                self.mscale_all_dim,
+            )
+        else:
```
kaiyux, line 398: Nit:
```suggestion
        elif self.scale_type == RotaryScalingType.longrope:
            raise NotImplementedError("Long RoPE is not supported.")
        else:
```

tensorrt_llm/_torch/utils.py
```diff
@@ -33,6 +35,43 @@ def is_torch_compiling() -> bool:
...
+def model_extra_attrs(attrs: Dict):
+    old_attrs = getattr(_model_extra_attrs, 'attrs', None)
+    _model_extra_attrs.attrs = attrs
+    try:
+        yield
+    finally:
+        _model_extra_attrs.attrs = old_attrs
+
+
+def with_model_extra_attrs(get_attrs):
```
kaiyux, line 62: I felt a bit confused when both `with_model_extra_attrs` and `model_extra_attrs` are used in the code, can we merge those two functions?

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -15,6 +16,7 @@
 
 from ..metadata import KVCacheParams
 from ..pyexecutor.resource_manager import KVCacheManager
+from ..utils import get_model_extra_attrs
```
QiJune, line 19: May I know the motivation of this PR? It seems that we want to share the same `rope_cos_sin` tensors among different layers.  The current LRU cache approach provides a gobal Dict, and this PR provides a per-model Dict. Both approaches work.

yuxianq: Global dict always leaks the cached tensors. The per-model dict can release these cached tensors when the model is released. It is hard to decide the size of the global dict. The original implementation uses cache size = 1, which can only support the case that all layers use the same rope parameters (true in most cases). If the model uses different rope parameters for interleaved layers, the cache will always miss. Another case is running CI tests in parallel. Different models in different tests will share the same global dict and evict each other's rope tensors. If we use cache size = N (or even unlimited), all N rope tensors will be leaked.

tensorrt_llm/_torch/utils.py
yuxianq: `with_model_extra_attrs` wraps functions like `PyTorchModelEngine.forward`, `model_extra_attrs` wraps code blocks. If we use `model_extra_attrs` for `PyTorchModelEngine.forward`, we have to indent the whole function body, so I add `with_model_extra_attrs` to keep `PyTorchModelEngine.forward` unchanged like:

```
    @with_model_extra_attrs(lambda self: self.model.extra_attrs)
    def forward(self, ...):
        ...
```

tensorrt_llm/_torch/attention_backend/interface.py
yuxianq: Updated.

tensorrt_llm/_torch/attention_backend/interface.py
yuxianq: Updated.

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -354,57 +357,75 @@ def from_config(config) -> "RopeParams":
             rope_params.beta_slow = rope_scaling.get("beta_slow", 1)
             rope_params.mscale = rope_scaling.get("mscale", 1.0)
             rope_params.mscale_all_dim = rope_scaling.get("mscale_all_dim", 0.0)
+        # Workaround for DeepSeek V3 Lite since its rope_scaling is null in config.json.
```
hlu1, line 360: Should we modify the model `config.json` directly instead of adding a hack in code?

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -354,57 +357,75 @@ def from_config(config) -> "RopeParams":
...
+                device='cuda',
+            )
         rope_cos_sin = torch.torch.tensor(
             rope_cos_sin,
             dtype=torch.float32,
             device='cuda',
         )
-        rope_inv_freq = None
+        if extra_attrs is not None:
+            cache[self] = RopeConstParams(
```
hlu1, line 424: Why do the rope param tensors have to be cached in the first place? Why can't they be saved in the model and passed as inputs, as in https://github.com/meta-llama/llama-models/blob/main/models/llama4/model.py#L354

tensorrt_llm/_torch/attention_backend/interface.py
yuxianq: The config.json is provided by DeepSeek, we cannot modify it by ourselves. We are not the model provider.

tensorrt_llm/_torch/attention_backend/interface.py
yuxianq: By caching rope param tensors, using cos_sin_cache is the implementation details of our attention module. Creating the rope param tensors once and threading it from top-level module to attention layers requires the modification of modeling files, which leaks implementation details. I think it is worth to simplify the modeling files to lower the threshold to contribute new models by third-party developers.

yuxianq:
This PR is blocked by https://github.com/NVIDIA/TensorRT-LLM/pull/3449 since the multi-gpu tests cannot pass without it.

yuxianq:
/bot run --disable-fail-fast --stage-list "DGX_H100-4_GPUs-TensorRT-2"

kaiyux (approved):

PR #3380, opened 2025-04-08
feat: Offloading Multimodal embedding table to CPU in Chunked Prefill Mode

katec846:
# Optimization for Large Embedding Tables in Multimodal Models

## Problem Statement
In multimodal models, large embedding tables generated by the multimodal encoder consume significant GPU memory. Currently, these tables are stored entirely in GPU memory during processing, which can be inefficient and limiting, especially for models processing large batches or long sequences.

## Solution
This PR implements a memory optimization strategy that offloads the multimodal model embedding table (mm_embedding_table) to CPU memory and uses a chunked prefetching mechanism during processing. This approach is only available when operating in context chunk mode, allowing for efficient memory usage without compromising performance through strategic prefetching.

### Core Implementation
- Added new buffer management system in `promptTuningBuffers.h` for handling prompt tuning ping-pong buffers.
- Implemented chunk-based prefetching and processing in `trtGptModelInflightBatching.cpp`.
- Added support for mm_embedding_table offloading in Python interface layers.

### File Changes
1. **C++ Changes**
   - `promptTuningBuffers.h`: Added new buffer management system with ping-pong buffers for efficient chunk handling.
   - `trtGptModelInflightBatching.cpp`: Implemented asynchronous prefetching and chunk processing mechanisms.
   - Various other C++ files: Added supporting functionality for buffer management and data transfer.

2. **Python Changes**
   - `model_runner_cpp.py`: Added multimodal model embedding table offloading support with pinned memory allocation.
   - `multimodal_model_runner.py`: Updated model generation process and move the embedding table to pinned memory.
   - `utils.py`: Added new `--mm_embedding_offloading` argument with automatic default configuration.

## Usage
The feature can be enabled using the new `--mm_embedding_offloading` argument. When not specified, it defaults to True if using a multimodal model with chunked context. This optimization is particularly useful for large-scale multimodal applications where GPU memory is a constraint.


cpp/include/tensorrt_llm/batch_manager/promptTuningBuffers.h
```diff
@@ -0,0 +1,177 @@
...
+    PromptTuningBuffers(SizeType32 maxBatchSize, runtime::BufferManager const& manager,
+        runtime::ModelConfig const& modelConfig, runtime::WorldConfig const& worldConfig);
+
+    void validate(std::optional<TensorPtr> const& optReqPromptEmbeddingTable,
+        std::optional<SizeType32> const& optReqPromptVocabSize);
+
+    void fill(RequestVector const& contextRequests, RequestVector const& genRequests,
+        runtime::BufferManager const& manager, bool packed);
+
+    void initializeChunkPtableBuffers(
```
symphonylyh: can we do the standard .h/.cpp isolation like only put method signature in .h and move all implementations to .cpp?

cpp/include/tensorrt_llm/batch_manager/promptTuningBuffers.h
```diff
@@ -0,0 +1,177 @@
...
+
+    runtime::PromptTuningParams mPromptTuningParams;
+    SizeType32 mMaxPromptVocabSize;
+
+    bool runtimeIsChunkedContext;
+
+    bool mChunkPtableInitialized{false};
+    std::optional<std::array<TensorPtr, 2>> mChunkPtableBuffers;
+    std::optional<std::vector<std::vector<SizeType32>>> mChunkPtableBufferStartPositions;
+    size_t mChunkPtableCurrentIndex{0};
```
symphonylyh: it would be great to add some explanation to these new fields and the ping-pong buffer idea somewhere, and mark the section as "Specific for Chunked Prefill mode"

cpp/include/tensorrt_llm/batch_manager/promptTuningBuffers.h
```diff
@@ -0,0 +1,177 @@
...
+        std::array<TensorPtr, 2> buffers;
+        std::vector<std::vector<SizeType32>> startPositions(2);
+
+        for (int i = 0; i < 2; i++)
+        {
+            // Initialize each buffer's positions with 0
+            startPositions[i].emplace_back(0);
+
+            buffers[i] = manager.gpu(runtime::ITensor::makeShape({contextChunkSize, modelConfig.getHiddenSize()}),
+                nvinfer1::DataType::kHALF); // TODO: change to embedding table data type
```
symphonylyh: remind of TODO

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.h
```diff
@@ -352,6 +352,15 @@ class TrtGptModelInflightBatching : public TrtGptModel
     /// @brief Change the speculative decoding mode.
     void changeSpecDecMode(ScheduledRequests const& scheduledRequests);
 
+    // void prefetchPromptTableChunk(ScheduledRequests const& currRequests);
```
symphonylyh: remind cleanup

cpp/include/tensorrt_llm/batch_manager/runtimeBuffers.h
```diff
@@ -135,6 +135,12 @@ class RuntimeBuffers
 
 public:
     TensorPtr sequenceLengthsDevice;
+    bool runtimeIsChunkedContext;
+
+    std::unique_ptr<PromptTuningBuffers>& getPromptTuningBuffers()
+    {
+        return promptTuningBuffers;
+    }
```
symphonylyh: I actually don't understand why `promptTuningBuffers` has to be a private member. how about we put it to public along with Medusa/Lookahead etc and directly access without a get?

cpp/include/tensorrt_llm/batch_manager/llmRequest.h
```diff
@@ -1772,6 +1780,9 @@ class GenericLlmRequest
 
     LlmRequestState mState{LlmRequestState::kCONTEXT_INIT};
 
+    // Position of current ptable index
```
symphonylyh: outside prompttuningbuffer context, better call it prompt tuning table so people can understand. ptable in variable name is fine to be simple
is it better to be "// current position of the prompt tuning table"

cpp/include/tensorrt_llm/batch_manager/llmRequest.h
```diff
@@ -1016,7 +1024,7 @@ class GenericLlmRequest
     void setPrepopulatedPromptLen(SizeType32 prepopulatedPromptLen, SizeType32 kvTokensPerBlock)
     {
         TLLM_LOG_DEBUG("Setting pre-populated prompt length for request %lu to %i.", mRequestId, prepopulatedPromptLen);
-
+        printf("setPrepopulatedPromptLen: %u\n", prepopulatedPromptLen);
```
symphonylyh: remind cleanup

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
```diff
@@ -2601,4 +2615,169 @@ SizeType32 TrtGptModelInflightBatching::getMaxCapacityBatchSize(SizeType32 input
...
+
+    // Bounds check
+    if (beginPos + processChunkSize > inputTokensMutable.size())
+    {
+        TLLM_THROW("Invalid chunk access: beginPos(%zu) + processChunkSize(%zu) > totalSize(%zu)", beginPos,
+            processChunkSize, inputTokensMutable.size());
+        return;
+    }
+
+    // Process tokens
```
symphonylyh: would be great to add a high-level explanation on the code logic here, or at the beginning of this function

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
```diff
@@ -2601,4 +2615,169 @@ SizeType32 TrtGptModelInflightBatching::getMaxCapacityBatchSize(SizeType32 input
...
+    auto inputTokensChunk = inputTokensMutable.begin() + beginPos;
+    std::vector<SizeType32> outOfVocabTokens;
+    SizeType32 chunkedPTableOffset = 0;
+    for (size_t i = 0; i < processChunkSize; i++)
+    {
+        if (inputTokensChunk[i] >= vocabSize)
+        {
+            outOfVocabTokens.push_back(inputTokensChunk[i]);
+            inputTokensChunk[i] = vocabSize + chunkedPTableOffset;
+            chunkedPTableOffset++;
```
symphonylyh: can acutally do
```suggestion
SizeType32 ptableTokenId = vocabSize;
...
            inputTokensChunk[i] = ptableTokenId++;
```

tensorrt_llm/runtime/model_runner_cpp.py
```diff
@@ -821,14 +825,19 @@ def _prepare_names_list(self, names_list: List[str], batch_size: int):
...
         if input_token_extra_ids:
             assert len(batch_input_ids_list) == len(input_token_extra_ids), \
                 f"Batch size of input_token_extra_ids ({len(input_token_extra_ids)}) must be the same as input batch size ({len(batch_input_ids_list)})"
         prompt_tuning_configs = len(batch_input_ids_list) * [None]
         if prompt_table is not None:
-            prompt_table_data = self._prepare_embedding_table(
-                prompt_table).cuda()
+            if ptable_offloading:
+                prompt_table_data = self._prepare_embedding_table(
+                    prompt_table).pin_memory()
```
symphonylyh, line 843: better to explain for cpu offloading & prefetching we need pinned memory

examples/multimodal/utils.py
```diff
@@ -72,6 +72,13 @@ def add_common_args(parser):
         action='store_true',
         help='Enables chunked context (only available with cpp session).',
     )
+    parser.add_argument(
+        '--ptable_offloading',
```
symphonylyh: nitpick: bettern name it --mm_embedding_offloading to because ptable is underlying implementation that multimodal users may not need to know
also ptable is "prompt tuning table" instead of "position table"

symphonylyh: may add a section in multimodal/README.md to briefly introduce the idea -- you can reuse the PR description in the README

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
```diff
@@ -2601,4 +2615,169 @@ SizeType32 TrtGptModelInflightBatching::getMaxCapacityBatchSize(SizeType32 input
     return mKvCacheManager->getMaxCapacityBatchSize(inputLength, outputLength);
 }
 
+void TrtGptModelInflightBatching::prefetchPromptTableChunk(
```
symphonylyh: will do more careful review in 2nd round, but firstly can one of prefetchPromptTableChunk and prefetchPromptTableChunkToGpu be renamed to be less confusing? right now they're overlapping in naming

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
```diff
@@ -1553,6 +1555,7 @@ TrtGptModelInflightBatching::prepareBuffers(
     NVTX3_SCOPED_RANGE(prepareBuffers);
 
     auto& runtimeBuffers = *mBuffers[bufferId];
+    runtimeBuffers.runtimeIsChunkedContext = mIsChunkedContext;
```
symphonylyh: how about we set this field when ctor the runtimeBuffers? std::make_shared<RuntimeBuffers>(..., mIsChunkedContext)

btw, you can reuse the field name, like `runtimeBuffers.isChunkedContext` seems equally good

cpp/tensorrt_llm/batch_manager/runtimeBuffers.cpp
```diff
@@ -792,6 +792,7 @@ void RuntimeBuffers::setFromInputs(RequestVector const& contextRequests, Request
     }
     if (modelConfig.usePromptTuning())
     {
+        promptTuningBuffers->runtimeIsChunkedContext = runtimeIsChunkedContext;
```
symphonylyh: same here. initialize during ctor because this is more like a static attribute throughout the lifetime

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
```diff
@@ -1597,6 +1600,11 @@ void TrtGptModelInflightBatching::executeStep(
         "executeStep: " + std::to_string(contextRequests.size()) + " ctx reqs, "
             + std::to_string(generationRequests.size()) + " gen reqs");
 
+    if (mPromptTableOffloading)
+    {
+        prefetchNextPromptTableChunk(contextRequests, true, bufferId);
+    }
```
katec846, line 1624: For this if statement, do you think it's better to write it here or check it inside prefetchNextPromptTableChunk?

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
katec846: I removed this flag and changed it to promptTableOffloading

cpp/include/tensorrt_llm/batch_manager/llmRequest.h
```diff
@@ -1772,6 +1780,9 @@ class GenericLlmRequest
 
     LlmRequestState mState{LlmRequestState::kCONTEXT_INIT};
 
+    // current position of the prompt tuning table
```
symphonylyh: feels maybe we should still add "(only used in chunked prefill mode)"

cpp/include/tensorrt_llm/batch_manager/promptTuningBuffers.h
```diff
@@ -0,0 +1,105 @@
...
+     *    - Maintains the starting position of each batch's data within each buffer
+     *    - Maintained separately for each ping-pong buffer
+     *
+     * Memory Optimization:
+     * - Only two GPU buffers are maintained regardless of total embedding table size
+     * - Each buffer size is limited to contextChunkSize * hiddenSize
+     * - Efficient memory usage through chunk-based processing
+     */
+    void initializeChunkPtableBuffers(runtime::BufferManager const& manager, runtime::ModelConfig const& modelConfig,
+        SizeType32 contextChunkSize, std::shared_ptr<LlmRequest> const&);
```
symphonylyh: missing param name `llmReq`?

cpp/include/tensorrt_llm/batch_manager/promptTuningBuffers.h
```diff
@@ -0,0 +1,105 @@
...
+
+    runtime::PromptTuningParams mPromptTuningParams;
+    SizeType32 mMaxPromptVocabSize;
+
+    bool mPromptTableOffloading;
+
+    bool mChunkPtableInitialized{false};
+    std::optional<std::array<TensorPtr, 2>> mChunkPtableBuffers;
+    std::optional<std::vector<std::vector<SizeType32>>> mChunkPtableBufferStartPositions;
+    size_t mChunkPtableCurrentIndex{0};
```
symphonylyh: nit: you can move this part below the comment block?

cpp/include/tensorrt_llm/batch_manager/trtGptModelOptionalParams.h
```diff
@@ -145,6 +147,8 @@ class TrtGptModelOptionalParams
     bool isLeaderInOrchMode;
     std::optional<std::vector<std::string>> additionalOutputNames;
     bool gatherGenerationLogits;
+    // Whether to offload the prompt table to the GPU
```
symphonylyh: "to the CPU and prefetching to GPU"

cpp/tensorrt_llm/batch_manager/promptTuningBuffers.cpp
```diff
@@ -148,11 +202,123 @@ void PromptTuningBuffers::fill(RequestVector const& contextRequests, RequestVect
...
+    }
+
+    mChunkPtableBuffers = std::move(buffers);
+    mChunkPtableBufferStartPositions = std::move(startPositions);
+
+    mChunkPtableCurrentIndex = 0;
+    mChunkPtableInitialized = true;
+}
+
+void PromptTuningBuffers::moveToNextChunkPtableBuffer()
```
symphonylyh: better call it `switchChunkPtableBuffer` because "next chunk" could be mistaken as the next chunk of data instead of the pingpong buffer?

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
symphonylyh: put it here is fine, as I think it can make the code path clearer. recommend to put a `, /* isFirstChunk */ true/false` to be self-explanatory

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
```diff
@@ -2601,4 +2615,188 @@ SizeType32 TrtGptModelInflightBatching::getMaxCapacityBatchSize(SizeType32 input
...
+ * 2. remapInputTokensForPromptTable:
+ *    - Identifies tokens that need prompt table embeddings (tokens that are greater than vocabSize)
+ *    - Remaps IDs to match chunked prompt table layout
+ *
+ * 3. copyPromptTableToGpuInChunk:
+ *    - Handles actual transfer from CPU pinned memory to GPU
+ *    - Uses appropriate buffer manager based on isBeforePrepareBuffers
+ */
+void TrtGptModelInflightBatching::prefetchNextPromptTableChunk(
+    RequestVector const& contextRequests, bool isBeforePrepareBuffers, SizeType32 bufferId)
```
symphonylyh: `isFirstChunk` you used in .h is a good name, can replace here

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.h
```diff
@@ -489,6 +499,8 @@ class TrtGptModelInflightBatching : public TrtGptModel
     std::optional<SizeType32> mMaxNumTokensRuntime;
     // Controls if generation logits should be gathered, so that returnGenerationLogits can be requested.
     bool mGatherGenerationLogits{false};
+    // Whether the context is chunked
```
symphonylyh: the description can made more clearer like `offloading and prefetching the prompt tuning table (only effective in chunked prefill mode)`

examples/multimodal/utils.py
```diff
@@ -72,6 +72,13 @@ def add_common_args(parser):
         action='store_true',
         help='Enables chunked context (only available with cpp session).',
     )
+    parser.add_argument(
+        '--mm_embedding_offloading',
+        type=lambda s: s.lower() in ("yes", "true", "t", "1"),
```
symphonylyh: is there a specific reason why this is yes/true/t/1 instead of store_true i.e. once you specify it, it's True, otherwise default False?

symphonylyh (approved):
Great work!

examples/multimodal/README.md
```diff
@@ -1176,3 +1177,16 @@ The full set of commands to enable 2-way tensor parallelism for LLaVA is:
         --hf_model_dir tmp/hf_models/${MODEL_NAME} \
         --engine_dir tmp/trt_engines/${MODEL_NAME}/fp16/2-gpu \
     ```
+## Enabling Embedding Table Offloading
+
+Embedding Table Offloading is a memory optimization technique that helps manage large embedding tables  more efficiently. It offloads the embedding table to CPU memory and uses a chunked prefetching mechanism during processing. This approach is only available when operating in context chunk mode.
```
symphonylyh: "tables^^more" there is an extra space 😆

examples/multimodal/utils.py
katec846: This flag would be automatically enabled if this is a multimodal model and with chunk context enabled, so I think this flag would only be specified if we want to turn it off

symphonylyh: I was mainly saying the yes/true/t/1 is strange and is not alike other args
in this case, can't we just do store_true? so (1) if users don't want it, doesn't need to do anything, it's default False (2) if users specifically want to enable it, they can specify `<command> --mm_embedding_offloading`, and your README already explained this usage (3) if users forget to turn on it under long context chunked prefill mode, no worry because you have the auto-on logic

katec846: So, do we want to default this flag to false? I thought we discussed before that we want to minimize the number of flags, so this flag would be automatically enabled if and only if this is a multimodal model with context chunking enabled.

If there's an auto-on logic (which we have now) and change the vairable to `store_true`, the user wouldn't be able to set it to `false` under a multimodal model with context chunking enabled.

I took a reference for the way to set this flag from [here](https://github.com/NVIDIA/TensorRT-LLM/blob/7d823eafdfd8bb70d1a331c31d3eae483e3ef030/examples/utils.py#L348)

symphonylyh: ah I see. you're right, I completely missed that
how to turn it off btw? is `--mm_embedding_offloading false` a valid usage? does it mean the only allowed options are yes/true/t/1?

katec846: No, it means
```
def convert_to_bool(s):
    lowercase_input = s.lower()
    return lowercase_input in ("yes", "true", "t", "1")
```
So if pass in anything other than yes, true, t, 1 it would be false

symphonylyh: ok sounds good

cpp/tensorrt_llm/pybind/executor/executorConfig.cpp
```diff
@@ -562,6 +565,8 @@ void initConfigBindings(pybind11::module_& m)
             &tle::ExecutorConfig::setGatherGenerationLogits)
         .def_property("use_variable_beam_width_search", &tle::ExecutorConfig::getUseVariableBeamWidthSearch,
             &tle::ExecutorConfig::setUseVariableBeamWidthSearch)
+        .def_property("mm_embedding_offloading", &tle::ExecutorConfig::getPromptTableOffloading,
+            &tle::ExecutorConfig::setPromptTableOffloading)
```
Funatiq, line 569: Why does the Python parameter have a different name than the C++ parameter?

katec846: From the user's perspective, it's offloading the multimodal model embedding table, but in terms of implementation details, this embedding table is stored in the prompt table (the user doesn't know this). Therefore, there are two different names. If we are concerned about unifying the names, I would change the flag to `prompt_table_offloading`.

Funatiq: Sounds reasonable. I just got a little confused when I added this to the `test_executor_config` (which was missed in this PR).

Issue #1644, opened 2024-05-21
[Model Requests] Add support for CogVLM2


harry-stark:
Following up on Cogvlm, CogVlm2 is here: https://github.com/THUDM/CogVLM2

Easily one of the best open-source multimodal model, that is competitive to GPT-4 and Gemini. 
https://github.com/THUDM/CogVLM2?tab=readme-ov-file#benchmark

The community would be grateful for having a Tensort-LLM engine for such a good model,since the support for Cogvlm #1554  has been recently added, it can be built upon it. 

cc @ncomly-nvidia

isidentical:
this'd be huge @AdamzNV

MultiChen:
Very much looking forward to Tensorrt-LLM supporting CogVLM2

sxyu:
the existing CogVLM code already works for CogVLM2 with some minor changes (comment out tensor parallel, hardcode the vision length/kv heads)

sxyu:
however the cogvlm implementation does not support inflight batching, which is rather unfortunate

Issue #1229, opened 2024-03-04
When will FP8 be available for Mixtral?


Pernekhan:
Could you guys share rough timeline on the support of FP8 quantization for Mixtral (MoE) model?

cc: @Tracin

Tracin:
Mixtral with FP8 is almost done for now, will be released in version 0.9.0

vip-china:
Does Mixtra support AWQ 4-bit?

Marlinski:
> Mixtral with FP8 is almost done for now, will be released in version 0.9.0

Hi @Tracin , it seems that I still can't export an fp8 quantized version of Mixtral 8x7B in version 0.9.0
Is there something I missed in the doc or is this still not supported ? 

When using the python script quantized.py like this:

```
python /app/tensorrt_llm/examples/quantization/quantize.py \
   --model_dir mistralai/Mixtral-8x7B-Instruct-v0.1  \
   --output_dir /workspace/checkpoint/mistralai_Mixtral-8x7B-Instruct-v0.1/tllm_checkpoint_fp8 \
   --qformat fp8 \
   --tp_size 1  \
   --pp_size 1  \
   --kv_cache_dtype fp8
```

it exit with an error:
```
Warning: this is an old NPZ format and will be deprecated soon.
Traceback (most recent call last):
  File "/app/tensorrt_llm/examples/quantization/quantize.py", line 52, in <module>
    quantize_and_export(model_dir=args.model_dir,
  File "/usr/local/lib/python3.10/dist-packages/tensorrt_llm/quantization/quantize_by_ammo.py", line 360, in quantize_and_export
    with safetensors.safe_open(f"{export_path}/rank0.safetensors",
FileNotFoundError: No such file or directory: "/workspace/checkpoint/mistralai_Mixtral-8x7B-Instruct-v0.1/tllm_checkpoint_fp8/rank0.safetensors"
```

It did generate two files though, but I can't use them an input for trtllm-build:

```
$ ls /workspace/checkpoint/mistralai_Mixtral-8x7B-Instruct-v0.1/tllm_checkpoint_fp8/
unknown:MixtralForCausalLM_tp1.json  unknown:MixtralForCausalLM_tp1_rank0.npz
```

The generated json file does not look like the ones usually generated with convert_checkpoint. Please advise ! :)

hawkeoni:
Hi!
I'm sorry, I couldn't understand it from release notes, but is it released in 0.9.0?
[...]

PR #3382, opened 2025-04-08
feat: support llama4 nope layers; support FP8 checkpoint loading;

nvzhihanj:


tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -55,8 +55,11 @@ def __init__(
     ):
         config = model_config.pretrained_config
 
-        # TODO: do we need to use NoPE in any versions of LLaMA4?
-        nope_layer_interval = None
+        # TODO: There has been inconsistency in the use of params "nope_layer_interval" and "no_rope_layers"
+        # in llama4 HF models and Transformer Autoconfig fields
+        # Hardcoding the use of nope_layer_interval for now, and will remove once the inconsistency is resolved.
+        self.is_llama4 = config.model_type == "llama4_text"
+        nope_layer_interval = 4 if self.is_llama4 else None
```
mikeiovine: One small issue - I'm not sure NoPE layers work with the current setup.I previously added a `use_rope` flag to attention, but I removed it after code review to keep attention API changes minimal in the first MR. Now, it seems like we have a problem as there is no way to tell the attention layer to skip RoPE.

For example, in the TRTLLM backend, we'll have:
* `fused_qkv == True`
* `pos_embd_params is None` for NoPE layers
* Then, we'll (incorrectly) apply RoPE here: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/modules/attention.py#L187

There are a few ways to fix, what do we think is the cleanest? Does it make sense to always skip RoPE when `pos_embd_params is None`, even for FlashInfer? This way, we don't have to deal with 2 flags.
cc @QiJune @hlu1

nvzhihanj: I will follow what vLLM does for now: https://github.com/vllm-project/vllm/blob/db104221845e44e0b85336a54e44eb8850b8a5bc/vllm/model_executor/models/llama4.py#L129
where `no_rope_layer == 0` means nope_layer is activated

QiJune: @mikeiovine for NoPE layer, I suggest we make `pos_embd_params` and `rotary_emb` both `None` when initilizing `LlamaAttention`.

```python
        super().__init__(
            xxx
            rotary_emb=None if is_nope_layer else LlamaRotaryEmbedding(config),
            pos_embd_params=pos_embd_params,
            layer_idx=layer_idx,
            dtype=config.torch_dtype,
            config=model_config,
            use_qk_norm=use_qk_norm and not is_nope_layer,
            aux_stream=aux_stream,
        )
```

Then, in the forward of Attention:

```python
        if is_fused_qkv:
            if self.pos_embd_params is None and position_ids is not None and self.rotary_embd is not None:
                q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size],
                                    dim=-1)
                q, k = self.rotary_emb(position_ids, [q, k], attn_metadata)
                qkv = torch.concat([q, k, v], dim=-1)
                xxx
         else:
             if self.pos_embd_params is None and position_ids is not None and self.rotary_embd is not None:
                q, k = self.rotary_emb(position_ids, [q, k], attn_metadata)
```

tensorrt_llm/_torch/modules/rotary_embedding.py
```diff
@@ -179,7 +179,7 @@ def forward(
                     k = k.view(seq_len, -1)
                     return [q, k]
 
-        if use_gptj_style_rope:
+        elif use_gptj_style_rope:
```
mikeiovine: This does not look right to me. If you can't run `flashinfer_apply_rope_inplace` with `interleave=True`, the computation is gonna be wrong.

mikeiovine (changes requested):

tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -812,9 +812,16 @@ def load_expert_fc2_input_scale_fp8_qdq(
...
+            if f"{expert_id}.w1.input_scale" in weights:
+                w1_input_scale = weights[f"{expert_id}.w1.input_scale"]
+                w3_input_scale = weights[f"{expert_id}.w3.input_scale"]
+                w2_input_scale = weights[f"{expert_id}.w2.input_scale"]
+            else:
+                # Hack to load non standard checkpoint format
+                # TODO: Remove it after we switch to standard formart
+                w1_input_scale = weights[f"gate_up_proj_input_scale"]
+                w3_input_scale = weights[f"gate_up_proj_input_scale"]
+                w2_input_scale = weights[f"down_proj_input_scale"]
```
mikeiovine: I added an `MoEWeightLoadingMode` in the previous PR to specify the fused `gate_up_proj` format, can we use that? Seems like it's here to stay

tensorrt_llm/_torch/modules/rotary_embedding.py
nvzhihanj: @mikeiovine Thanks, my change was following the error message but seems like it's misleading - just to confirm even if flashinfer is available, this line could still fail under other conditions (when the `return` is not hit above) right?

Let me check with you offline to understand the right behavior

nvzhihanj: Resolved.

nvzhihanj:
> [PR_Github #1644](https://nv/trt-llm-cicd/job/helpers/job/PR_Github/1644/) [ run ] completed with state `FAILURE` [/LLM/main/L0_MergeRequest_PR pipeline #1229](https://nv/trt-llm-cicd/job/main/job/L0_MergeRequest_PR/1229/) completed with status: 'FAILURE'

Seems to be a `self_attn` mismatch for DeepSeek due to the change. Will try to repro it locally.

PR #3384, opened 2025-04-08
fix: Use hmac authentication for pickle encryption

yibinl-nvidia:


tensorrt_llm/executor/postproc_worker.py
```diff
@@ -70,16 +70,16 @@ class Output(NamedTuple):
 
     def __init__(
         self,
-        pull_pipe_addr: str,
-        push_pipe_addr: str,
+        pull_pipe_addr: tuple[str, bytes],
```
Superjomn: Should it be `tuple[str, Optional[bytes]]` to be aligned with the address-type of ZeroMqQueue? So are all the other places.

tensorrt_llm/executor/ipc.py
```diff
@@ -26,40 +30,68 @@ class ZeroMqQueue:
...
     def __init__(self,
-                 address: Optional[str] = None,
+                 address: Optional[tuple[str, Optional[bytes]]] = None,
                  *,
                  socket_type: int = zmq.PAIR,
                  is_server: bool,
                  is_async: bool = False,
-                 name: Optional[str] = None):
+                 name: Optional[str] = None,
+                 use_hmac_encryption: bool = True):
```
Superjomn, line 36: Do we need to measure the performance gap for this? Or what about creating an environment variable in `utils.py` for `use_hmac_encryption`'s default value so that we can switch it on and off globally?

Superjomn (approved):
LGTM overally

tensorrt_llm/executor/postproc_worker.py
yibinl-nvidia: Good point! Fixed.

tensorrt_llm/executor/ipc.py
yibinl-nvidia: Using env variable is an excellent suggestion. Will decide whether we need this after the QA result out.

yibinl-nvidia: resolved offline.

kaiyux (approved):

tensorrt_llm/executor/ipc.py
```diff
@@ -1,3 +1,7 @@
+import hashlib
```
yibinl-nvidia, line 1: block to wait after 0.18.2 release

chuangz0:
I get error 
```
0: ValueError: Server should not receive HMAC key when encryption is enabled
```
0: Traceback (most recent call last):
0:   File "<frozen runpy>", line 198, in _run_module_as_main
0:   File "<frozen runpy>", line 88, in _run_code
0:   File "/root/scratch/projects/tekit/tensorrt_llm/llmapi/mgmn_leader_node.py", line 29, in <module>
0:     launch_server_main()
0:   File "/root/scratch/projects/tekit/tensorrt_llm/llmapi/mgmn_leader_node.py", line 16, in launch_server_main
0:     server = RemoteMpiCommSessionServer(
0:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
0:   File "/root/scratch/projects/tekit/tensorrt_llm/llmapi/mpi_session.py", line 319, in __init__
0:     self.queue = ZeroMqQueue(addr, is_server=True)
0:                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
0:   File "/root/scratch/projects/tekit/tensorrt_llm/executor/ipc.py", line 62, in __init__
0:     raise ValueError(
0: ValueError: Server should not receive HMAC key when encryption is enabled

yibinl-nvidia:
@chuangz0 should be fixed by https://github.com/NVIDIA/TensorRT-LLM/pull/3649

PR #3221, opened 2025-04-02
feat: Return logits in PyTorch flow

tongyuantongyu:

PR #3388, opened 2025-04-09
feat: Support Top-K logprobs and prompt_logprobs in LLMAPI

hchings:
This MR only handles TRT path. PyT flow will be handling in [this MR](https://github.com/NVIDIA/TensorRT-LLM/pull/3221/).

## Usage
```python

llm = LLM(
    model="TinyLlama/TinyLlama-1.1B-Chat-v1.0",
    build_config=BuildConfig(gather_context_logits=True),
    gather_generation_logits=True)
        
sampling_params = SamplingParams(
        temperature=0.8,
        top_p=0.95,
        logprobs=5,   # new
        prompt_logprobs=2 # new
        ) 
        
# output.outputs[0].logprobs (list of dict, where the dict is token_id to Logprob): 
# [{11785: Logprob(logprob=-2.128943920135498, rank=1), 1244: Logprob(logprob=-2.878943920135498, rank=2), 297: Logprob(logprob=-3.003943920135498, rank=3), 451: Logprob(logprob=-3.316443920135498, rank=4), 263: Logprob(logprob=-3.503943920135498, rank=5), 2086: Logprob(logprob=-6.722693920135498, rank=113)}, {17999: Logprob(logprob=-1.5640053749084473, rank=1), 11785: Logprob(logprob=-1.7515053749084473, rank=2), 4280: Logprob(logprob=-2.5015053749084473, rank=3), 4100: Logprob(logprob=-3.0640053749084473, rank=4), 5566: Logprob(logprob=-3.1890053749084473, rank=5)}, {304: Logprob(logprob=-0.16267399489879608, rank=1), 363: Logprob(logprob=-2.6626739501953125, rank=2), 29892: Logprob(logprob=-4.2876739501953125, rank=3), 29889: Logprob(logprob=-4.3501739501953125, rank=4), 322: Logprob(logprob=-4.6001739501953125, rank=5)}...]       


# output.outputs[0].prompt_logprobs
#     same shape as above


 
```

### DIFF
--- a/tensorrt_llm/executor/executor.py
+++ b/tensorrt_llm/executor/executor.py
@@ -26,7 +26,8 @@ from ..llmapi.mpi_session import (MpiSession, external_mpi_comm_available,
 from ..llmapi.utils import (AsyncQueue, enable_llm_debug,
                             enable_worker_single_process_for_tp1, print_colored,
                             print_colored_debug)
-from ..sampling_params import BatchedLogitsProcessor, SamplingParams
+from ..sampling_params import (BatchedLogitsProcessor, LogprobParams,
+                               SamplingParams)
 from .ipc import FusedIpcQueue
 from .postproc_worker import PostprocParams, PostprocWorkerConfig
 from .request import GenerationRequest, LoRARequest, PromptAdapterRequest
@@ -203,6 +204,24 @@ class GenerationExecutor(ABC):
         # (self._last_client_id + 1) % UINT64_MAX
         self._last_client_id = (self._last_client_id + 1) & ((1 << 64) - 1)
         return self._last_client_id
+
+    def _get_logprob_params(
+            self, request: GenerationRequest) -> Optional[LogprobParams]:
+        """Store logprobs-related fields from request for the later logprob calculation."""
+        logprob_params = None
+        if request.sampling_params.logprobs or request.sampling_params.prompt_logprobs:
+            logprob_params = LogprobParams(
+                logprobs=request.sampling_params.logprobs,
+                prompt_logprobs=request.sampling_params.prompt_logprobs,
+                # drop logits if users didn't explicitly ask for it, or if it's using PostProcess flow
+                drop_context_logits=(
+                    not request.sampling_params._need_return_context_logits)
+                or self.postproc_config.num_postprocess_workers > 0,
+                drop_generation_logits=(
+                    not request.sampling_params._need_return_generation_logits)
+                or self.postproc_config.num_postprocess_workers > 0)
+
+        return logprob_params
 
     def _maybe_initialize_iteration_results(self):
         if self._is_llm_executor:
--- a/tensorrt_llm/executor/result.py
+++ b/tensorrt_llm/executor/result.py
@@ -3,17 +3,19 @@ import json
 import weakref
 from dataclasses import dataclass, field
 from queue import Empty, Queue
-from typing import TYPE_CHECKING, Any, Callable, List, Literal, Optional, Union
+from typing import (TYPE_CHECKING, Any, Callable, List, Literal, NamedTuple,
+                    Optional, TypeAlias, Union)
 from weakref import WeakMethod
 
 import torch
+import torch.nn.functional as F
 
 from .._utils import nvtx_range_debug
 from ..bindings import executor as tllm
 from ..disaggregated_params import DisaggregatedParams
 from ..llmapi.tracer import global_tracer
 from ..llmapi.utils import AsyncQueue
-from ..sampling_params import SamplingParams
+from ..sampling_params import LogprobParams, SamplingParams
 from .utils import ErrorResponse, has_event_loop, is_llm_response
 
 if TYPE_CHECKING:
@@ -31,15 +33,48 @@ __all__ = [
 
 
 @dataclass(slots=True)
+class Logprob:
+    """Holds logprob and vocab rank for a token."""
+    logprob: float
+    rank: Optional[int] = None
+
+
+# List of token_id_to_Logprob dict for prompt or generation texts
+TokenLogprobs: TypeAlias = list[dict[int, Logprob]]
+
+
+class LogProbsResult(NamedTuple):
+    """Optional log probability outputs computed post runtime."""
+    prompt: Optional[TokenLogprobs] = None
+    generation: Optional[TokenLogprobs] = None
+
+
+class ResponseWrapper:
+    """Wrapper of runtime response with optional outputs computed post runtime.
+    """
+
+    def __init__(self,
+                 response: Union["PostprocWorker.Output", tllm.Response],
+                 logprobs: Optional[LogProbsResult] = None):
+        self._response = response
+        self.logprobs = logprobs
+
+    def __getattr__(self, name):
+        response = object.__getattribute__(self, '_response')
+        return getattr(response, name)
+
+
+@dataclass(slots=True)
 class CompletionOutput:
     """The output data of one completion output of a request.
 
     Args:
         index (int): The index of the output in the request.
         text (str): The generated output text. Defaults to "".
-        token_ids (List[int], optional): The token ids of the generated output text. Defaults to None.
+        token_ids (List[int], optional): The token ids of the generated output text. Defaults to [].
         cumulative_logprob (float, optional): The cumulative log probability of the generated output text. Defaults to None.
-        logprobs (List[float], optional): The log probabilities of the top probability words at each position if the logprobs are requested. Defaults to None.
+        logprobs (TokenLogprobs, optional): The log probabilities of the top probability words at each position if the logprobs are requested. Defaults to None.
+        prompt_logprobs (TokenLogprobs, optional): The log probabilities per prompt token. Defaults to None.
         finish_reason (Literal['stop', 'length', 'timeout', 'cancelled'], optional): The reason why the sequence is finished. Defaults to None.
         stop_reason (int, str, optional): The stop string or token id that caused the completion to stop, None if the completion finished for some other reason. Defaults to None.
         generation_logits (torch.Tensor, optional): The logits on the generated output token ids. Defaults to None.
@@ -53,9 +88,10 @@ class CompletionOutput:
     """
     index: int
     text: str = ""
-    token_ids: Optional[List[int]] = None
+    token_ids: Optional[List[int]] = field(default_factory=list)
     cumulative_logprob: Optional[float] = None
-    logprobs: Optional[List[float]] = None
+    logprobs: Optional[TokenLogprobs] = field(default_factory=list)
+    prompt_logprobs: Optional[TokenLogprobs] = field(default_factory=list)
     finish_reason: Optional[Literal['stop', 'length', 'timeout',
                                     'cancelled']] = None
     stop_reason: Optional[Union[int, str]] = None
@@ -71,12 +107,6 @@ class CompletionOutput:
                                                 repr=False)
     # the result of result_handler passed to postprocess workers
     _postprocess_result: Any = None
-
-    def __post_init__(self):
-        if self.token_ids is None:
-            self.token_ids = []
-        if self.logprobs is None:
-            self.logprobs = []
 
     @property
     def length(self) -> int:
@@ -160,8 +190,11 @@ class GenerationResultBase:
     def context_logits(self) -> Optional[torch.Tensor]:
         return self._context_logits
 
-    def _handle_sequence(self, finish_reasons, response_tensors,
-                         sequence_index):
+    def _handle_sequence(self,
+                         finish_reasons,
+                         response_tensors,
+                         sequence_index,
+                         logprobs_result=None):
         """ Handle a single sequence in the response. """
 
         seq_idx = sequence_index
@@ -182,6 +215,11 @@ class GenerationResultBase:
 
         if response_tensors.cum_log_probs is not None:
             output.cumulative_logprob = response_tensors.cum_log_probs[src_idx]
+
+        if logprobs_result:
+            output.prompt_logprobs = logprobs_result.prompt
+            output.logprobs = logprobs_result.generation
+
         if response_tensors.log_probs is not None:
             output._last_logprobs_len = len(output.logprobs)
             output.logprobs = response_tensors.log_probs[src_idx]
@@ -223,8 +261,14 @@ class GenerationResultBase:
     @nvtx_range_debug("handle_response",
                       color="red",
                       category="GenerationResultBase")
-    def _handle_response(self, response: Union["PostprocWorker.Output",
-                                               tllm.Response, ErrorResponse]):
+    def _handle_response(self,
+                         response: Union["PostprocWorker.Output", tllm.Response,
+                                         ResponseWrapper, ErrorResponse]):
+        if isinstance(response, ResponseWrapper):
+            logprobs_result = response.logprobs
+            response = response._response
+        else:
+            logprobs_result = None
 
         if isinstance(response, PostprocWorker.Output):
             self._done = response.is_final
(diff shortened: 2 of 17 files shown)


tensorrt_llm/llmapi/llm.py
```diff
@@ -447,6 +447,12 @@ def _prepare_sampling_params(
                         "tokenizer is required to reset end_id if it is None, or you can explicitly specify the end_id for sampling_params"
                     )
                 sampling_params._setup(self.tokenizer)
+            if sampling_params.prompt_logprobs:
+                # Context logits are required to compute prompt_logprobs
+                sampling_params.return_context_logits = True
+                # TODO: need a way to know whether to drop context logits at worker later
```
hchings: @Superjomn  - for me to dump contexts (or generation logits) before sending it via IPC, I need a flag to know whether users actually set `return_context_logits`  or it was enabled here.  

It could be achieved by adding an additional internal flag in `SamplingParams`, lmk if you suggest against that.

tensorrt_llm/executor/worker.py
```diff
@@ -747,16 +747,27 @@ def handle_for_worker(self, responses: List[tllm.Response]) -> None:
             assert response is not None
             queue = self.worker.return_queue(response.client_id)
 
+            # TODO: pass in k
```
hchings: @Superjomn  - I need to pass in `k` (aka `logprobs` or `prompt_logprobs`) from `request.sampling_params` to here.  Haven't found a clean way given how worker is currently structured.

tensorrt_llm/executor/result.py
```diff
@@ -532,3 +565,58 @@ async def __anext__(self):
         except asyncio.TimeoutError:
             self._done = True
             raise StopAsyncIteration
+
+
+def compute_logprobs(
```
hchings, line 594: I put all logprob stuff in `result.py`, we can move to a single file if needed.

tensorrt_llm/sampling_params.py
```diff
@@ -280,13 +282,6 @@ def __post_init__(self):
 
         self.best_of = self.best_of or self.n
 
-        if (not self.use_beam_search and self.n < self.best_of
```
hchings, line 283: I'm not exactly sure what this is, still confirming w/ Xiaowei this is no longer needed.

tensorrt_llm/executor/result.py
```diff
@@ -29,6 +31,17 @@
 ]
 
 
+@dataclass
+class Logprob:
```
Superjomn, line 36: `class Logprob(slots=True)` is better for performance.

tensorrt_llm/executor/result.py
```diff
@@ -76,6 +91,8 @@ def __post_init__(self):
             self.token_ids = []
         if self.logprobs is None:
             self.logprobs = []
+        if self.prompt_logprobs is None:
+            self.prompt_logprobs = []
```
Superjomn: For setting default values with containers, we can declare the parameters in dataclass as
```python
logprobs: LogprobResults = field(default_factory=list)
```
This eliminates the need for the code here.

So are other parameters here.

tensorrt_llm/executor/result.py
```diff
@@ -90,7 +107,7 @@ def token_ids_diff(self) -> List[int]:
         return self.token_ids[self._last_token_ids_len:]
 
     @property
-    def logprobs_diff(self) -> List[float]:
+    def logprobs_diff(self) -> List[dict[int, Logprob]]:
```
Superjomn: We can unify all the `List/Dict` to `list/dict` as they are suggested from [python 3.9](https://docs.python.org/3.9/library/typing.html#typing.List)

tensorrt_llm/executor/result.py
```diff
@@ -224,8 +244,10 @@ def _handle_sequence(self, finish_reasons, response_tensors,
                     f"Unknown finish reason: {finish_reasons[src_idx]}")
 
     @nvtx_range("handle_response", color="red", category="GenerationResultBase")
-    def _handle_response(self, response: Union["PostprocWorker.Output",
-                                               tllm.Response, ErrorResponse]):
+    def _handle_response(self,
+                         response: Union["PostprocWorker.Output", tllm.Response,
+                                         ErrorResponse],
+                         additional_outputs: Optional[Dict[str, Any]] = None):
```
Superjomn: It seems that the addtional_outputs contains multiple members, and Dict is an opaque compared to a concrete type. How about making it a namedtuple:
```python
class AdditonalOutputs(Namedtuple):
    ''' some comments '''
    prompt_logprobs: Optional[...] = None
    generation_logprobs: Optional[...] = None
```

This can make this argument more intuitive and good for jumping in IDE as well.

tensorrt_llm/executor/result.py
```diff
@@ -532,3 +565,58 @@ async def __anext__(self):
...
             raise StopAsyncIteration
+
+
+def compute_logprobs(
+    context_logits: Optional[torch.Tensor],
+    generation_logits: Optional[torch.Tensor],
+    output_token_ids: Optional[list[int]],
+    k_prompt_logprobs: int = 5,
+    k_logprobs: int = 5,
+) -> Tuple[Optional[LogprobResults], Optional[LogprobResults]]:
```
Superjomn: replace Tuple => tuple.

tensorrt_llm/executor/result.py
```diff
@@ -532,3 +565,58 @@ async def __anext__(self):
...
+
+    Returns:
+        Tuple of:
+            - prompt_logprobs: list of dicts {token_id -> Logprob}
+            - generation_logprobs: list of dicts {token_id -> Logprob}
+    """
+
+    def _topk_logprobs(logits: torch.Tensor, top_k: int,
+                       tokens: Optional[list[int]]) -> LogprobResults:
+        if logits.dim() == 3 and logits.size(0) == 1:
```
Superjomn: Where do these magic numbers come from? Can we eliminate them by deducing from some shape-like variable, or maybe some comment is helpful for understanding?

tensorrt_llm/executor/result.py
```diff
@@ -532,3 +565,58 @@ async def __anext__(self):
...
+            logits = logits.squeeze(0)  # shape: [T, vocab]
+
+        logprobs = F.log_softmax(logits.to("cuda", dtype=torch.float32), dim=-1)
```
Superjomn, line 616: Please add a `NOTE` here to mark the potential performance drop, as we may reuse this part and enhance in PyT sampler.

tensorrt_llm/executor/result.py
```diff
@@ -532,3 +565,58 @@ async def __anext__(self):
...
+        topk_vals, topk_indices = torch.topk(logprobs, k=top_k, dim=-1)
+
+        results: Logprobs = []
```
Superjomn: It seems the Logprobs is not bound; should it be `list[Logprob]`?

tensorrt_llm/executor/worker.py
Superjomn: Can you elaborate on this?

tensorrt_llm/executor/worker.py
```diff
@@ -747,16 +747,27 @@ def handle_for_worker(self, responses: List[tllm.Response]) -> None:
...
+                "prompt_logprobs": prompt_logprobs,
+                "generation_logprobs": logprobs,
+            }
+
             # For AsyncQueue.sync_q, we will batch the events to avoid too many
             # event notifications, thus put without wait here.
             if isinstance(queue, _SyncQueue):
                 global_tracer().log_instant("worker-rsp.put")
-                queue.put_nowait(response)
+                queue.put_nowait((response, additional_outputs))
```
Superjomn: How about making a Namedtuple for this so that we can add the typing for related variables? This could be better for reading and IDE jumping.

tensorrt_llm/llmapi/llm.py
Superjomn: Yeah, that makes sense. How about introducing it in SamplingParams as a private property? As SamplingParams is a 1.0 committed API and currently, this implementation is just a WAR.
```python
@property
def _need_return_context_logits(self) -> bool:
    return self.prompt_logprobs
```
This shouldn't break the API stability check.

tensorrt_llm/executor/result.py
hchings: Yes, ignore this if clause now, I'm updating this part.

tensorrt_llm/executor/result.py
hchings: Do you mean perf drop due to a few more additional operation?

hchings: Do you mean perf drop due to a few more additional operation?

tensorrt_llm/executor/worker.py
hchings: Please see changes here: https://github.com/NVIDIA/TensorRT-LLM/pull/3388/files#diff-1b6c3b1327b4cb6f697a78ad27820f044fd5ee2b2738c8851d680b8b8a0d8a7fR431

Basically need to pass some logprob-related sampling_params to `handle_response()`. Not sure whether there's a cleaner method than above link.

tensorrt_llm/executor/worker.py
```diff
@@ -747,16 +764,35 @@ def handle_for_worker(self, responses: List[tllm.Response]) -> None:
...
+                response.request_id)
+            if logprob_params:
+                additional_outputs = compute_logprobs(
+                    logprob_params.get("prompt_logprobs"),
+                    logprob_params.get("logprobs"),
+                    response.result.context_logits,
+                    response.result.generation_logits,
+                    response.result.output_token_ids[0])  # TODO
+
+                if logprob_params.get("drop_context_logits"):
```
hchings: Hi @Superjomn  - to drop logits as we discussed, I'll need to make the `Result` pybind ( `.def_property_readonly("result", &tle::Response::getResult)` in `request.cpp` ) mutable. 

Can you confirm that there's no concerns of doing so?

tensorrt_llm/executor/result.py
Superjomn: I mean that the copy the logits tensor from host to cuda, we can record the reason here.

tensorrt_llm/executor/worker.py
Superjomn: I think it is reasonable, but we can leave some comments there to describe the context. In the long run,  we will have some equivalent Python DS, and will unify the DS throughout PyExecutor and LLM-API, so it is a normal need to post-process the Response in each stage.

cc @MartinMarciniszyn @QiJune @kaiyux for viz.

tensorrt_llm/executor/worker.py
```diff
@@ -426,6 +426,19 @@ def submit(self, request: GenerationRequest) -> GenerationResult:
         request_id = self._enqueue_request(request)
         # request_id returned from backend is necessary for the abort_request method.
         self._client_id_to_request_id[client_id] = request_id
+        # Store lobprobs-related fields from request for the later logprob calculation.
+        if request.sampling_params.logprobs or request.sampling_params.prompt_logprobs:
+            self._await_response_helper._request_id_to_logprob_params[
```
Superjomn: There is one mapping from request_id (actually the client_id) to some DS in worker and proxy, [Worker._results](https://github.com/nvidia/TensorRT-LLM/blob/510c5952cb373d1f0ae21b0fe19190770e2ad1d1/tensorrt_llm/executor/worker.py#L72) and [Proxy.results](https://github.com/nvidia/TensorRT-LLM/blob/510c5952cb373d1f0ae21b0fe19190770e2ad1d1/tensorrt_llm/executor/proxy.py#L79), maybe we can extend the GenerationResult for these new fields.

tensorrt_llm/executor/worker.py
hchings: updated, kept `Result` readonly.

MartinMarciniszyn:
@Funatiq , could you please also take a look at these changes?

tensorrt_llm/executor/result.py
Funatiq: I'm wondering if we shouldn't do the calculations while the logits are still on the GPU in the model instead of copying them to host and back.

tensorrt_llm/executor/worker.py
```diff
@@ -746,16 +763,33 @@ def handle_for_worker(self, responses: List[tllm.Response]) -> None:
...
+
+                if logprob_params.get("drop_generation_logits"):
+                    response.clear_generation_logits()
+
             # For AsyncQueue.sync_q, we will batch the events to avoid too many
             # event notifications, thus put without wait here.
             if isinstance(queue, _SyncQueue):
                 global_tracer().log_instant("worker-rsp.put")
-                queue.put_nowait(response)
+                queue.put_nowait((response, additional_outputs))
```
Funatiq: Is it possible to add the `additional_outputs` to `tllm.Result` instead of keeping them separate? There is already a member for it.

Funatiq:
In general, do we consider how users can easily add new outputs to a model? In the TRT workflow we enable additional outputs via the [ExecutorConfig](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/include/tensorrt_llm/executor/executor.h#L1538) and [OutputConfig](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/include/tensorrt_llm/executor/executor.h#L237) and the runtime handles all the rest. Can we implement something similar for the PyTorch workflow?

tensorrt_llm/executor/worker.py
Superjomn: @Funatiq @hchings   Since we are going to have a pure Python Response to replace the pybinding one, keep some new fields in Python and hesitate to introduce new fields in `tllm.Response` if not exist seems acceptable as an intermediate status. But we hope the code to be intuitive enough. One alternative is to make the tuple here a Namedtuple Response with `response` and fields in `additioal_outputs` expanded there with some comments; this way will be better.

Funatiq: Fine with me.

hchings: I'll prefer we wait for a pure Python Response, instead of introducing a new Namedtuple (which doesn't differ much than the current impl) and removing it again later.

tensorrt_llm/executor/worker.py
hchings: updated to use `GenerationResult` instead.

tensorrt_llm/executor/result.py
hchings: updated

tensorrt_llm/executor/executor.py
```diff
@@ -204,6 +204,23 @@ def _get_next_client_id(self):
         self._last_client_id = (self._last_client_id + 1) & ((1 << 64) - 1)
         return self._last_client_id
 
+    def _get_logprob_params(self, request: GenerationRequest) -> Optional[dict]:
```
Superjomn, line 207: A dedicated type for the logprob_params is preferred to help track this DS used across the codebase:

```python
class LogprobParams(Namedtuple):
    logprobs: Logprobs
    prompt_probs: Logprobs
    # Drop the logits once the logprobs are computed from the Worker
    drop_context_logits: bool = False
    # Drop the geneation_logits once the logprobs are computed from the Worker
    drop_generation_logits: bool = False
```

tensorrt_llm/sampling_params.py
```diff
@@ -224,13 +225,16 @@ class SamplingParams:
...
-    return_log_probs: bool = False
+    logprobs: Optional[int] = None
+    prompt_logprobs: Optional[int] = None
     return_context_logits: bool = False
     return_generation_logits: bool = False
     exclude_input_from_output: bool = True
     return_encoder_output: bool = False
     return_perf_metrics: bool = False
     additional_model_outputs: Optional[List[AdditionalModelOutput]] = None
+    _context_logits_auto_enabled: bool = False
```
Superjomn, line 236: Can you add some comments here to record the background for the WAR? Since this is TRT-only and SamplingParams is shared between both TRT and PyT, we may deprecate it in the future.

hchings: updated

tensorrt_llm/executor/executor.py
hchings: updated

Superjomn (approved):
LGTM

hchings:
/bot skip --comment "Latest full pipeline has passed, but CI failed to report back status. The rebase after CI passed is only for sign-off".

PR #3392, opened 2025-04-09
test: fix memory leak of tests

xinhe-nv:
fix memory leak issue of test_openai_multi_chat_example


tests/unittest/llmapi/apps/_test_openai_multi_chat.py
```diff
@@ -59,6 +59,8 @@ def engine_from_fp8_quantization(model_name):
 
     yield engine_dir.name
 
+    llm.shutdown()
```
Superjomn, line 62: This should not be needed, since the `with llm` above will invoke [LLM.__exit__](https://github.com/superjomn/TensorRT-LLM/blob/deb876ecdb938b930a2a0b0c9331ecf13bb80b9e/tensorrt_llm/llmapi/llm.py#L643-L646), and it will shutdown there.

kaiyux, line 62: @Superjomn Can you help clarify if this is necessary? I feel like if it does, it seems to be a usability question because we can't expect users to always explicitly call `shutdown`?

kaiyux: Just saw your comments, thanks.

xinhe-nv: I added this step because the test is failure. When I added this step, the case can pass.

```shell
@pytest.hookimpl(hookwrapper=True)
    def pytest_runtest_call(item):
        start_threads = None
        exclude_regex = item.config.getini("threadleak_exclude")
        if is_enabled(item):
            start_threads = current_threads(exclude_regex)
        yield
        if start_threads:
            end_threads = current_threads(exclude_regex)
            leaked_threads = end_threads - start_threads
            if leaked_threads:
>               pytest.fail("Test leaked %s" % sorted_by_name(leaked_threads))
E               Failed: Test leaked [<Thread(asyncio_0, started 139763220883136)>]
```

Superjomn: It is odd. But we can merge this fix first, and I will investigate the shutdown method later.

xinhe-nv: maybe the code should like this.
with LLM() as llm:
    llm.save()

Superjomn: I thought they had a similar effect:
```
with llm:
    llm.save(...)
or 
with LLM() as llm:
   llm.save(...)
```
Both should invoke the `llm.__exit__`.

But you can have a try.

Superjomn (approved):
LGTM, I will investigate the shutdown issue later.

PR #3397, opened 2025-04-09
fix: mllama e2e pytorch flow fix

yechank-nvidia:
This PR is the bugfix for the e2e pytorch flow of mllama.

python3 quickstart_advanced.py --model_dir meta-llama/Llama-3.2-11B-Vision --enable_chunked_prefill --enable_overlap_scheduler
=>
AttributeError: 'MllamaForConditionalGeneration' object has no attribute 'infer_max_seq_len'

`MllamaForCausalLM` is not based on `DecoderModelForCausalLM` which causes AttributeError of `infer_max_seq_len`. Copied `infer_max_seq_len` from `DecoderModelForCausalLM`.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_mllama.py
+++ b/tensorrt_llm/_torch/models/modeling_mllama.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 
 import copy
+import math
 from typing import Dict, Optional, Tuple
 
 import torch
@@ -23,6 +24,7 @@ import transformers.models.mllama.configuration_mllama as config_mllama
 from torch import nn
 from tqdm import tqdm
 
+from ...logger import logger
 from ..attention_backend.interface import AttentionMetadata
 from ..distributed import ParallelConfig, TensorParallelMode
 from ..model_config import ModelConfig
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/models/modeling_mllama.py
```diff
@@ -259,6 +262,32 @@ def forward(
         )
         return hidden_states
 
+    def infer_max_seq_len(self) -> int:
```
kaiyux, line 265: Should we add tests for the changes?

yechank-nvidia: I think there is no e2e test for torch flow of Mllama. There is a unittest for modeling_mllama but this problem only occurs on e2e with quickstart_advanced.py

We do have [test_e2e.py](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/defs/test_e2e.py#L1483)for other models(e.g. llama, nemotron)

We can add the test mllama here. But, personally I think we don't need to add test case for this one.

kaiyux (approved):

PR #3398, opened 2025-04-09
feat: Add tool_call support for serving that can handle agentic framework function/tool call

sarattha:
**Sorry for closing the previous PR, I really messed up that PR.
For this PR, I already resolve all comments from previous PR.**

**For the comment of this feature usecase**: `I'm currently working on a project that deploys a multi-agent system using the OpenAI Agent SDK framework with local LLM models (e.g., Llama3.3 and Qwen). In our framework, an agent is able to call tools to execute more complex tasks. The mechanism involves embedding a <tool_call></tool_call> tag within the generated message along with a json_schema response type.
Since I also use TensorRT-LLM as the inference engine for our LLM models, it's essential that the TensorRT-LLM serving handles tool calls in a manner similar to the OpenAI API.
` 

**For the comment on unimplemented section (i.e. `pass`)**: `Yes, I intentionally left the implementation incomplete because I wasn’t certain about the optimal way to handle explicit function calls at that time. I used a placeholder (pass) to indicate where a proper implementation should go and wanted to open the discussion for suggestions—perhaps something along the lines of an explicit tool call like those in the OpenAI API (e.g., RAG or the Web search tool). If this functionality turns out to be unnecessary, I’m happy to remove it. For now, I’ve included my proposed implementation for explicit tool calls in this commit and look forward to your feedback.`

This pull request introduces several changes to the `tensorrt_llm/serve` module, focusing on enhancing function call detection and parsing, and updating response formats. Below are the most important changes:

### Function Call Detection and Parsing:

* Added `detect_and_parse_function_calls` and `validate_function_arguments` functions to `function_calling_utils.py` to detect and parse function calls from model output and validate function arguments against JSON schema.

### Response Format Updates:

* Updated `ResponseFormat` class in `openai_protocol.py` to allow a new type `json_schema` and include additional optional fields such as `strict`, `name`, `description`, and `schema`. [[1]](diffhunk://#diff-4b6ece0a77c7a186737d731752536269c760e4aa3e9ebc896f8023a37280d10aL48-R55) [[2]](diffhunk://#diff-4b6ece0a77c7a186737d731752536269c760e4aa3e9ebc896f8023a37280d10aL189-R196)


LinPoly:
Thanks for contribution!! And don't worry to mess up PR, you can force push when necessary, all commits in this PR will be squash into single commit when merged onto main, the title of PR will be the commit message. Here are some general suggestions and questions for this PR:

1. After the implementation of vLLM and SGlang, it seems like we'd better narrow the scope of this PR, the parser function in this PR maybe applicable for non-streaming scenarios, but streaming case need more processing than simple regex matching. How about we focusing on non-streaming tool call in this PR? Feel free to correct me if I am wrong.
2. We need a test to ensure correctness, vLLM's test can be a reference https://github.com/vllm-project/vllm/blob/1575c1701a80befec8efe274b338cb26bc199275/tests/entrypoints/openai/test_chat.py#L655, but we may need to adapt it accordingly. The test can be added to https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/apps/_test_openai_chat.py
3. Your regex seems to extract content quoted by json, function and tool_call, can you explain the coverage of these three formats, like what model can be used with this regex?

sarattha:
> Thanks for contribution!! And don't worry to mess up PR, you can force push when necessary, all commits in this PR will be squash into single commit when merged onto main, the title of PR will be the commit message. Here are some general suggestions and questions for this PR:
> 
> 1. After the implementation of vLLM and SGlang, it seems like we'd better narrow the scope of this PR, the parser function in this PR maybe applicable for non-streaming scenarios, but streaming case need more processing than simple regex matching. How about we focusing on non-streaming tool call in this PR? Feel free to correct me if I am wrong.
> 2. We need a test to ensure correctness, vLLM's test can be a reference https://github.com/vllm-project/vllm/blob/1575c1701a80befec8efe274b338cb26bc199275/tests/entrypoints/openai/test_chat.py#L655, but we may need to adapt it accordingly. The test can be added to https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/apps/_test_openai_chat.py
> 3. Your regex seems to extract content quoted by json, function and tool_call, can you explain the coverage of these three formats, like what model can be used with this regex?

1. I totally agree.
2. Ok
3. These features can support models that utilize tool calls or function calls, such as the Llama family and Qwen 2.5++.

Thank you for your response and suggestion

kaiyux:
@LinPoly Can you follow up this PR?

PR #3399, opened 2025-04-09
vibe coded bmm sharder

suyoggupta:


lucaslie (commented):
What do you think it would take to write a pattern matcher for

```
bmm1 --> bmm2 --> all_reduce
```

?

suyoggupta:
> What do you think it would take to write a pattern matcher for
> 
> ```
> bmm1 --> bmm2 --> all_reduce
> ```
> 
> ?

We'd need to add all_reduce if we're sharding a dimension that is reduced. The current sharding strategy is to shard the 0th dimension, which happens to be of size num_experts. So the strategy of bmm1->all_gather->bmm2->all_gather works. The first all_gather can be optimized away if all ops between bmm1 and bmm2 are elementwise ops. Does that make sense?

suyoggupta:
open this one instead. https://github.com/nv-auto-deploy/TensorRT-LLM/pull/1.

PR #3407, opened 2025-04-09
doc: genai-perf benchmark & slurm multi-node for trtllm-serve doc

LinPoly:
1. genai-perf example and doc for benchmarking trtllm-serve perf
2. doc for launching trtllm-serve in multiple nodes with slurm (todo)

### DIFF
--- a/docs/source/commands/trtllm-serve.rst
+++ b/docs/source/commands/trtllm-serve.rst
@@ -34,7 +34,7 @@ For the full syntax and argument descriptions, refer to :ref:`syntax`.
 Inference Endpoints
 -------------------
 
-After you start the server, you can send inference requests through completions API and Chat API, which are compatible with corresponding OpenAI APIs.
+After you start the server, you can send inference requests through completions API and Chat API, which are compatible with corresponding OpenAI APIs. We use ``TinyLlama-1.1B-Chat-v1.0`` for examples in the following sections.
 
 Chat API
 ~~~~~~~~
@@ -86,6 +86,24 @@ Finally, run performance benchmark with the following command:
     :linenos:
 
 Refer to `README <https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/README.md>`_ of ``genai-perf`` for more guidance.
+
+Multi-node Serving with Slurm
+-----------------------------
+
+You can deploy ``DeepSeek-V3`` in two nodes with Slurm and ``trtllm-serve``
+
+.. code-block:: bash
+
+    echo -e "enable_attention_dp: true\npytorch_backend_config:\n  enable_overlap_scheduler: true" > extra-llm-api-config.yml
+
+    srun -N 2 -w [NODES] \
+        --output=benchmark_2node.log \
+        --ntasks 16 --ntasks-per-node=8 \
+        --mpi=pmix --gres=gpu:8 \
+        --container-image=<CONTAINER_IMG> \
+        --container-mounts=/workspace:/workspace \
+        --container-workdir /workspace \
+        bash -c "trtllm-llmapi-launch trtllm-serve deepseek-ai/DeepSeek-V3 --backend pytorch --max_batch_size 161 --max_num_tokens 1160 --tp_size 16 --ep_size 4 --kv_cache_free_gpu_memory_fraction 0.95 --extra_llm_api_options ./extra-llm-api-config.yml"
 
 Metrics Endpoint
 ----------------
--- a/docs/source/commands/trtllm-serve.rst
+++ b/docs/source/commands/trtllm-serve.rst
@@ -34,7 +34,7 @@ For the full syntax and argument descriptions, refer to :ref:`syntax`.
 Inference Endpoints
 -------------------
 
-After you start the server, you can send inference requests through completions API and Chat API, which are compatible with corresponding OpenAI APIs. We use ``TinyLlama-1.1B-Chat-v1.0`` for examples in the following sections.
+After you start the server, you can send inference requests through completions API and Chat API, which are compatible with corresponding OpenAI APIs. We use `TinyLlama-1.1B-Chat-v1.0 <https://huggingface.co/TinyLlama/TinyLlama-1.1B-Chat-v1.0>`_ for examples in the following sections.
 
 Chat API
 ~~~~~~~~
@@ -69,7 +69,7 @@ Another example uses ``curl``:
 Benchmark
 ---------
 
-You can use any benchmark clients to test the serving performance of ``trtllm_serve``, in the following example, we use ``genai-perf`` as an example.
+You can use any benchmark clients compatible with OpenAI API to test serving performance of ``trtllm_serve``, we recommend ``genai-perf`` and here is a benchmarking recipe.
 
 First, install ``genai-perf`` with ``pip``:
 
@@ -77,9 +77,9 @@ First, install ``genai-perf`` with ``pip``:
 
    pip install genai-perf
 
-Then, :ref:`start a server<Starting a Server>` with ``trtllm-serve``.
+Then, :ref:`start a server<Starting a Server>` with ``trtllm-serve`` and ``TinyLlama-1.1B-Chat-v1.0``.
 
-Finally, run performance benchmark with the following command:
+Finally, test performance with the following command:
 
 .. literalinclude:: ../../../examples/serve/genai_perf_client.sh
     :language: bash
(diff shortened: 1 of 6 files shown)


examples/serve/genai_perf_client.sh
```diff
@@ -0,0 +1,18 @@
...
+    --service-kind openai \
+    --endpoint-type chat \
+    --num-prompts 1 \
+    --random-seed 123 \
+    --synthetic-input-tokens-mean 128 \
+    --synthetic-input-tokens-stddev 0 \
+    --output-tokens-mean 128 \
+    --output-tokens-stddev 0 \
+    --concurrency 1 \
+    --measurement-interval 4000 \
```
ganeshku1: The guidance from genai-perf team is to move away from measurement-interval option and user --request-rate.
Can you please use --request-rate instead of measurement-interval?

ganeshku1: Reference slack thread on reasoning why too use --request-rate: https://nvidia.slack.com/archives/C06SUT79XTL/p1741888549651459

LinPoly: Thanks. Updated. Replace concurrency and measurement-interval with request-rate=1

examples/serve/genai_perf_client.sh
```diff
@@ -0,0 +1,17 @@
+#! /usr/bin/env bash
+
+genai-perf profile \
+    -m TinyLlama-1.1B-Chat-v1.0 \
+    --tokenizer TinyLlama/TinyLlama-1.1B-Chat-v1.0 \
+    --service-kind openai \
+    --endpoint-type chat \
+    --num-prompts 1 \
```
kaiyux: Let's use a bit more, like 100.
```suggestion
    --num-prompts 100 \
```

examples/serve/genai_perf_client.sh
```diff
@@ -0,0 +1,17 @@
...
+    --random-seed 123 \
+    --synthetic-input-tokens-mean 128 \
+    --synthetic-input-tokens-stddev 0 \
+    --output-tokens-mean 128 \
+    --output-tokens-stddev 0 \
+    --request-rate 1 \
```
kaiyux:
```suggestion
    --request-rate 10 \  # send 10 requests per second
```

docs/source/commands/trtllm-serve.rst
```diff
@@ -66,6 +66,45 @@ Another example uses ``curl``:
     :language: bash
     :linenos:
 
+Benchmark
+---------
+
+You can use any benchmark clients to test the serving performance of ``trtllm_serve``, in the following example, we use ``genai-perf`` as an example.
```
kaiyux:
```suggestion
You can use any benchmark clients to test the serving performance of ``trtllm-serve``, in the following example, we recommend using ``genai-perf``.
```

kaiyux: Besides, why is "``" used, instead of "`"?

docs/source/commands/trtllm-serve.rst
```diff
@@ -66,6 +66,45 @@ Another example uses ``curl``:
...
+.. literalinclude:: ../../../examples/serve/genai_perf_client.sh
+    :language: bash
+    :linenos:
+
+Refer to `README <https://github.com/triton-inference-server/perf_analyzer/blob/main/genai-perf/README.md>`_ of ``genai-perf`` for more guidance.
+
+Multi-node Serving with Slurm
+-----------------------------
+
+You can deploy ``DeepSeek-V3`` in two nodes with Slurm and ``trtllm-serve``
```
kaiyux: Also link documents to `trtllm-llmapi-launch` here.

docs/source/commands/trtllm-serve.rst
```diff
@@ -34,7 +34,7 @@ For the full syntax and argument descriptions, refer to :ref:`syntax`.
 Inference Endpoints
 -------------------
 
-After you start the server, you can send inference requests through completions API and Chat API, which are compatible with corresponding OpenAI APIs.
+After you start the server, you can send inference requests through completions API and Chat API, which are compatible with corresponding OpenAI APIs. We use ``TinyLlama-1.1B-Chat-v1.0`` for examples in the following sections.
```
kaiyux:
```suggestion
After you start the server, you can send inference requests through completions API and Chat API, which are compatible with corresponding OpenAI APIs. We use [``TinyLlama-1.1B-Chat-v1.0``](https://huggingface.co/TinyLlama/TinyLlama-1.1B-Chat-v1.0) for examples in the following sections.
```

examples/serve/genai_perf_client.sh
LinPoly: Would that be too heavy for CI workloads?

LinPoly: Ah NVM

docs/source/commands/trtllm-serve.rst
LinPoly: > Besides, why is "``" used, instead of "`"?

No special reason, just follow the previous notation in this doc

LinPoly: Would be weird to say "in the example we recommend", will change the statement though.

examples/serve/genai_perf_client.sh
LinPoly: This option doesn't seem to stand for the number of requests that will be run in this benchmark, I just find that it can be removed without affecting benchmark, how about we just remove it.

examples/serve/genai_perf_client.sh
LinPoly: This will increase workload by 10 times, in local test genai-perf will send out 350 requests with isl=osl=128, would it be too heavy for our CI?

docs/source/commands/trtllm-serve.rst
LinPoly: We cannot do this since this is a rst file, it works for markdown though.

kaiyux: Ok, is there a way to link into rst file?

examples/serve/genai_perf_client.sh
kaiyux: We can keep it as an example usage, users will need to select one of the following anyway: offline/concurrency/req rate.

examples/serve/genai_perf_client.sh
kaiyux: I'm not expecting it to be too long, shouldn't be more than 5mins, no?

ganeshku1 (approved):
The GenAI-Perf changes looks good to me.

docs/source/commands/trtllm-serve.rst
LinPoly: Yes, but I think it is kind of hard to show a piece of text as both code and link, how about using link for the first occurrence of TinyLlama-1.1B-Chat-v1.0 and code for others? I preview locally and the doc looks acceptable.

examples/serve/genai_perf_client.sh
LinPoly: Maybe longer, this test is running on A10.

examples/serve/genai_perf_client.sh
LinPoly: But this may cause some confusion when using together with request-rate, user can see how many requests genai-perf sends in the report (see the following screenshot), and this number is not as same as num-prompts.
image

LinPoly: I think we'd better remove it if we don't use any offline dataset in this example.

docs/source/commands/trtllm-serve.rst
kaiyux: Sure, sounds good to me, thanks.

docs/source/commands/trtllm-serve.rst
kaiyux: Sure, thanks.

examples/serve/genai_perf_client.sh
kaiyux: It should be fine, let's try it. Or maybe you can have a quick run locally?

examples/serve/genai_perf_client.sh
kaiyux: @LinPoly Users are going to use it anyway, if you think this is going to confuse the users, we should report to genai-perf.

@ganeshku1 Can you help clarify?

lkomali: @LinPoly You’re right. `--num-prompts` doesn’t control the request count.
`--num-prompts` tells GAP how many unique prompts it should prepare.

For example, if `--request-count` is set to 70 and `--num-prompts` is 100, GAP will generate **100 unique prompts**, but only create **70 input payloads** — since request count limits how many are actually sent. It will just sample from those 100 prompts.

This is an alternative name to the CLI option [--num-dataset-entries](https://github.com/triton-inference-server/perf_analyzer/tree/main/genai-perf#:~:text=%2D%2Dnum%2Ddataset%2Dentries%20%3Cint%3E%3A%20The%20number%20of%20unique%20payloads%20to%20sample%20from.%20These%20will%20be%20reused%20until%20benchmarking%20is%20complete.)
To reduce confusion for users, especially in examples or docs, you could consider using `--num-dataset-entries` instead of `--num-prompts`.

LinPoly: Thanks, then we can use request-count to control it.

kaiyux (approved):

PR #3410, opened 2025-04-09
test: torch-flow conditional disagg test

zhengd-nv:
Add a mock disagg server (worker tester) to test conditional disaggregation in torch workflow.

### DIFF
--- a/tests/integration/defs/disaggregated/test_configs/disagg_config_cache_reuse.yaml
+++ b/tests/integration/defs/disaggregated/test_configs/disagg_config_cache_reuse.yaml
@@ -1,0 +1,26 @@
+hostname: localhost
+port: 8000
+model: TinyLlama/TinyLlama-1.1B-Chat-v1.0
+free_gpu_memory_fraction: 0.15
+backend: "pytorch"
+pytorch_backend_config:
+  use_cuda_graph: False
+  enable_overlap_scheduler: False
+context_servers:
+  num_instances: 1
+  tensor_parallel_size: 1
+  pipeline_parallel_size: 1
+  kv_cache_config:
+    enable_block_reuse: True
+    enable_partial_reuse: True
+  urls:
+      - "localhost:8001"
+generation_servers:
+  num_instances: 1
+  tensor_parallel_size: 1
+  pipeline_parallel_size: 1
+  kv_cache_config:
+    enable_block_reuse: True
+    enable_partial_reuse: True
+  urls:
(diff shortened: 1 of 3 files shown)


Shunkangz:
Thank you for proposing this. May I ask why we create a mock proxy server here? It looks like that most of code is similar to the existing proxy server. This test might not cover the case when we change the logic of proxy server.

zhengd-nv:
> Thank you for proposing this. May I ask why we create a mock proxy server here? It looks like that most of code is similar to the existing proxy server. This test might not cover the case when we change the logic of proxy server.

The mock proxy server is not an actual server. It is intended to test the underlying workers instead of the existing proxy server, so it will mimic some logics in an actual proxy server to test if these logics applicable to the workers. It's role is similar to `cpp/tests/executor/disaggExecutor.h`

Shunkangz:
> > Thank you for proposing this. May I ask why we create a mock proxy server here? It looks like that most of code is similar to the existing proxy server. This test might not cover the case when we change the logic of proxy server.
> 
> The mock proxy server is not an actual server. It is intended to test the underlying workers instead of the existing proxy server, so it will mimic some logics in an actual proxy server to test if these logics applicable to the workers. It's role is similar to `cpp/tests/executor/disaggExecutor.h`

@zhengd-nv Thank you for explanation. But why not directly add a test with real proxy server? In this case, we can test both building blocks and upper level interface.

zhengd-nv:
> Thank you for explanation. But why not directly add a test with real proxy server? In this case, we can test both building blocks and upper level interface.

The proxy server doesn't support direct routing (i.e. explicitly specify ctx&gen server for a request) and mixing aggregated requests. Besides, we only have `test_disaggregated.py` as upper level interface test, but no tests about workers (building blocks) for potential upper level extensions.

Shunkangz:
> > Thank you for explanation. But why not directly add a test with real proxy server? In this case, we can test both building blocks and upper level interface.
> 
> The proxy server doesn't support direct routing (i.e. explicitly specify ctx&gen server for a request) and mixing aggregated requests. Besides, we only have `test_disaggregated.py` as upper level interface test, but no tests about workers (building blocks) for potential upper level extensions.

Okey. So ideally, we should add another test with the real proxy server to test both.

pcastonguay (approved):
Thanks for adding this test. Lgtm.

Shixiaowei02 (approved):

PR #3412, opened 2025-04-09
chore : Split GptExecutor tests out of gpt tests to reduce single tes…

peaceh-nv:
Split GptExecutor tests out of gpt tests to reduce single test time, from 48 min for total test time to 32min test time. The e2e time of this test also includes build google test time and build engine time, which can not to be split out in this MR.


DomBrown (approved):

DomBrown:
Building the google tests is required and done in a pytest fixture with session scope, so should only happen once per CI stage
Building the engine in this way is on my to-do list when I get time

PR #3413, opened 2025-04-09
fix: fix partialMatch

pansicheng:
If other.uniqueTokens.end() is not passed, the second range will have std::distance(first1, last1) elements, potentially causing undefined behavior if other.uniqueTokens is smaller than this.uniqueTokens.

https://en.cppreference.com/w/cpp/algorithm/mismatch


Funatiq (approved):
Good catch!

akhoroshev:
I have version 0.19.0.dev2025032500 in prod and sometimes while stress testing the server gives an error
`TLLM_CHECK(prepopulatedPromptLen < promptLen);`

After debugging I found this error and fixed it, now the server is working fine.

Issue #3408, opened 2025-04-09
Make scaffolding Controller more generic


WeiHaocheng:
We need to do some optimizations to make the Controller more generic, which will help with better modularity.
These changes will not affect the existing API.

- We hope the Controller such as MajorityVoteController can be constructed by any Generation Controller. For a example, the Generation Controller may be a dynasor cot Controller which will yield for different and uncertainly turns between different sample. We develop "parallel_process_helper" to support it. The code looks like:
`sample_num = kwargs.get("sample_num", self.default_sample_num)
        controllers = [self.generation_controller.clone() for _ in range(sample_num)]
        tasks_list = [[copy.deepcopy(tasks[0])] for _ in range(sample_num)]
        kwargs_list = [kwargs for _ in range(sample_num)]

        yield from parallel_process_helper(controllers, tasks_list, kwargs_list)`

- We make the reward method in BestOfNController more generic as we observe that reward methods become various in recently paper and works.

- We remove most of type check for Task as users may override them.

PR #3416, opened 2025-04-09
feat: Make scaffolding Controller more generic #3408

WeiHaocheng:
We need to do some optimizations to make the Controller more generic, which will help with better modularity.
These changes will not affect the existing API.

- We hope the Controller such as MajorityVoteController can be constructed by any Generation Controller. For a example, the Generation Controller may be a dynasor cot Controller which will yield for different and uncertainly turns between different sample. To resolve that, we support yield "ParallelProcess" to "ScaffoldingLlm". The code of majority vote controller looks like:

`

    controllers = [self.generation_controller.clone() for _ in range(sample_num)]

    tasks_list = [[copy.deepcopy(tasks[0])] for _ in range(sample_num)]

    kwargs_list = [kwargs for _ in range(sample_num)]

    yield ParallelProcess(generation_controllers, tasks_list,
                              generation_kwargs_list)`

Then the "ScaffoldingLlm" will run the "process" of each controller with tasks and kwargs parallel. And the generation controller here can be someone which customized by user. In other words, users can pass any class of generation controller into majority vote controller without changing the code of majority vote controller.

- We make the reward method in BestOfNController more generic as we observe that reward methods become various in recently paper and works.

- We remove most of type check for Task as users may override them.

### DIFF
--- a/tensorrt_llm/scaffolding/controller.py
+++ b/tensorrt_llm/scaffolding/controller.py
@@ -1,7 +1,8 @@
 import copy
 from abc import ABC
+from dataclasses import dataclass
 from enum import Enum
-from typing import List, Tuple
+from typing import Any, List, Mapping, Tuple
 
 from tensorrt_llm.scaffolding.math_utils import get_digit_majority_vote_result
 from tensorrt_llm.scaffolding.task import (GenerationTask, RewardTask,
@@ -32,6 +33,13 @@ class Controller(ABC):
         raise NotImplementedError
 
 
+@dataclass(frozen=True)
+class ParallelProcess:
+    controllers: List[Controller]
+    tasks_list: List[List[Task]]
+    kwargs_list: List[Mapping[str, Any]]
+
+
 # Controller runs multiple generation tasks.
 class NativeGenerationController(Controller):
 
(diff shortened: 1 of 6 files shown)


tensorrt_llm/scaffolding/controller.py
```diff
@@ -67,17 +61,97 @@ class WorkerTag(Enum):
...
-                raise ValueError(
-                    "NativeRewardController requires exactly one RewardTask")
-
         for task in tasks:
             task.worker_tag = self.WorkerTag.REWARD
 
         yield tasks
 
 
+class ParallelTaskBuffer(ABC):
```
dongxuy04: Seems no abstract method for `ParallelTaskBuffer`, do we need `ABC` here?

tensorrt_llm/scaffolding/controller.py
```diff
@@ -67,17 +61,97 @@ class WorkerTag(Enum):
...
+
+            try:
+                obj = next(gen)
+                if isinstance(obj, ParallelTaskBuffer):
+                    paraller_tasks_buffer = obj
+                    for task in paraller_tasks_buffer.pending_tasks:
+                        task_controller_map[id(
+                            task)] = i, True  # True means from buffer
+                        buffer.pending_tasks.append(task)
+                    sub_buffer_map[i] = paraller_tasks_buffer
```
dongxuy04:
```suggestion
                    parallel_tasks_buffer = obj
                    for task in parallel_tasks_buffer.pending_tasks:
                        task_controller_map[id(
                            task)] = i, True  # True means from buffer
                        buffer.pending_tasks.append(task)
                    sub_buffer_map[i] = parallel_tasks_buffer
```

dongxuy04 (approved):

WeiHaocheng:
bot run

PR #3419, opened 2025-04-09
feat: output-format arg for synthetic data generation script and improved support for Mistral models in the Pytorch workflow

jdebache:
- Added an `output-format` argument to `prepare_dataset.py` allowing the generation of datasets compatible with `tllm-bench` without having to go through `stdout`
- Added explicit support for Mistral models in the Pytorch modelling code, as their configurations are slightly different from the Llama models (even though the architecture is mostly similar)
- Added an example script on how to quantize a HuggingFace checkpoint to a HuggingFace checkpoint for usage in the Pytorch workflow


jdebache:
@kaiyux @FrankD412  I would like to reopen the discussion on the output format of the synthetic dataset generation scripts.

Personally, I think it makes sense for the script to be able to generate different output formats. This set of changes should add support for the tllm-bench format without breaking the current behaviour as the default value for the new `output-format` argument is `gptManagerBenchmark`, which is what the script is generating currently.

jdebache:
That being said, of course, it would be preferrable if, in the future, tools that ingest datasets could conform to an existing format.

FrankD412:
> That being said, of course, it would be preferrable if, in the future, tools that ingest datasets could conform to an existing format.

I agree, the whole reason I introduced the second format was because `gptManagerBenchmark` was limited to input ids because it was in C++ (and didn't have access to the `transformers` library in Python). When I started `trtllm-bench`, an initial goal was to allow it to take either a prompt or the input ids and be able to take in requests from stdin. The existing JSON didn't lend itself to the latter, but the latter ended up becoming less important (though it might be coming back up again).

Also, maybe this is personal preference, but this PR now addresses two unrelated things which makes it harder to roll back or track changes in `trtllm-bench`. Mind if we split these sorts of things in the future?

jdebache:
I will split the Mistral stuff out.

jdebache:
Here is the Mistral changes: https://github.com/NVIDIA/TensorRT-LLM/pull/3843. I'll rebase this one on top of it once merged, which should leave only the dataset changes.

jdebache:
Moved the dataset generation stuff to https://github.com/NVIDIA/TensorRT-LLM/pull/3866, and expanded it. Closing this one as all changes are now represented on other branches.

PR #3423, opened 2025-04-09
feat: trtllm-gen fp4 GEMM for pytorch workflow

DomBrown:
Adds FP4 GEMM from TRTLLM-Gen using the new exported interface.


cpp/tensorrt_llm/thop/fp4GemmTrtllmGen.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+template <trtllm::gen::Dtype out_dtype>
+void runGemm(at::Tensor& out, at::Tensor const& mat1, at::Tensor const& mat2, at::Tensor const& mat1Scale,
+    at::Tensor const& mat2Scale, at::Tensor const& globalScale, int64_t m, int64_t n, int64_t k)
+{
+    auto eltType = trtllm::gen::Dtype::E2m1;
+
+    tensorrt_llm::kernels::TrtllmGenGemmRunner runner(eltType, out_dtype);
+
+    int64_t const numBytesWorkspace = runner.getWorkspaceSizeInBytes(m, n, k);
```
nekorobov: nit: for the generality, the dtype should also be passed to `getWorkspaceSizeInBytes` even if it is not used in this particular impl.

nekorobov (approved):

cpp/tensorrt_llm/thop/fp4GemmTrtllmGen.cpp
DomBrown: Constructing the runner chooses a GemmConfig inside the class, so do I actually need to pass it through here? The runner instance should already know what the types are right?

Should I be passing through eltType, out_type or both?

DomBrown: Agreed on fix in slack discussion. Marking resolved.

PR #3424, opened 2025-04-09
feat: Replace cyclic kv cache with sliding kv cache

tomeras91:
Currently on main, the KV cache in case of sliding window attention is implemented using a cyclic KV cache - when the sequence length exceeds the attention window, the KV activations of subsequent tokens override those of previous tokens in a cyclic manner. This causes difficulties with KV cache reuse (AKA prefix caching), as it corrupts KV cache blocks, making them not valid for reuse.

This PR changes this cyclic behavior -  now, the KV cache "slides":
* When a new block is needed, it is allocated (regardless of if the sequence length is longer or shorter than the attention window)
* An old block is detached from the sequence if the sequence length is large enough such that after we detach it we're left with enough tokens to fill the attention window. This way, we still utilize the memory savings made possible by sliding window attention.

"Detaching" a block means:
* If reuse is not enabled, the block is released (marked as free)
* If reuse is enabled:
    * If the block is not used by any other sequence and there's a free block in secondary memory, it is offloaded
    * If not, it is kept in primary memory
* Anyway, the block is removed from the allocated blocks and from the sequence.

RE sliding window attention + reuse - reuse is now supported for requests with prompt length smaller than the attention window, even if they exceeded the attention window during generation. Reuse for prompts longer than the attention window is still not supported.


tomeras91:
Currently, this PR only deals with the necessary changes in the `KVCacheManager`. It does **not** make the complementary changes in the attention kernels. Therefore, it is not ready to merge until the changes in the kernels are ready as well.

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
```diff
@@ -555,17 +555,18 @@ std::shared_ptr<kv_cache_manager::KVCacheManager> TrtGptModelInflightBatching::c
...
         if (maxAttenWin != getMaxSequenceLen())
         {
-            enableCyclicKvCache = true;
+            enableSlidingWindowKvCache = true;
             break;
         }
     }
-    TLLM_CHECK_WITH_INFO(
-        getMaxBeamWidth() == 1 || !enableCyclicKvCache, "Can't support cyclic kv cache with beam search.");
+    // TODO (tasida): Should this be removed? Beam search + sliding window kv cache should be supported.
```
tomeras91: I'm pretty sure beam search + sliding window attention is supported now, but wanted to validate with the reviewers before removing this assertion

Funatiq: I also think we should be able to remove this check. We definitely need to verify it with a test.

cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h
```diff
@@ -396,12 +402,14 @@ class GenerationRequest
...
-    [[nodiscard]] bool isCyclic() const
+    [[nodiscard]] bool getContextRequiresSlidingWindowKvCache() const
     {
-        return mNumTokens >= mCyclicThreshold;
+        return mContextRequiresSlidingWindowKvCache;
+    }
+
+    void setContextRequiresSlidingWindowKvCache(bool contextRequiresSlindigWindowKvCache)
+    {
+        mContextRequiresSlidingWindowKvCache = contextRequiresSlindigWindowKvCache;
```
Funatiq: I think we should make this a c-tor parameter and remove the setter.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1253,7 +1248,7 @@ KVCacheManager::KVCacheManager(std::vector<SizeType32> const& numKvHeadsPerLayer
     mMaxTokenNum = mMaxAttentionWindow + mSinkBubbleLength;
 
     // If maxBeamWidth > 1, use one more block for each sequence in the paged kv cache to avoid dropping the needed
-    // tokens, when enabling cyclic kv cache.
+    // tokens, when enabling sliding window kv cache.
     mUseOneMoreBlock
         = maxSequenceLength.has_value() && maxSequenceLength.value() > mMaxAttentionWindow && maxBeamWidth > 1;
```
Funatiq: I think we don't need `mUseOneMoreBlock` anymore.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1904,10 +1949,16 @@ SizeType32 KVCacheManager::getMaxCapacityBatchSize(SizeType32 inputLength, SizeT
 SizeType32 KVCacheManager::calculateMaxBlockRequirementsPerBeam(
     SizeType32 sequenceLength, SizeType32 sinkTokenLength, SizeType32 maxAttentionWindow, SizeType32 tokensPerBlock)
 {
+    // TODO (tomer): this should probably be changed
```
Funatiq: Please don't add usernames. We should use other means of tracking todos.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -430,14 +430,17 @@ bool BlockManager::verifyQueueIntegrity()
 
 void BlockManager::storeContextBlocks(GenerationRequest& sequence, LlmRequest const& llmRequest)
 {
-    constexpr int beamIdx = 0; // no need to consider more than one beam for input tokens
-    auto const& cacheBlockIds = sequence.getCacheBlockIds();
-    auto const& uniqueTokens = llmRequest.getUniqueTokens(beamIdx);
+    if (!sequence.getContextRequiresSlidingWindowKvCache())
```
Funatiq: Seems like we do this check twice now. Can we remove it here?

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1136,39 +1093,77 @@ std::deque<tle::KVCacheEvent> BlockManager::getLatestEvents(std::optional<std::c
...
         auto const usableSize = static_cast<runtime::SizeType32>(uniqueTokens.size()) - 1;
         auto blockedUniqueTokens = chopVectorIntoBlocks<UniqueToken>(uniqueTokens, usableSize, mTokensPerBlock, true);
         auto blockKeys = buildBlockKeys(blockedUniqueTokens, *llmRequest);
-        storeBlocks(std::move(blockKeys), cacheBlockIds[beamIdx]);
+
+        // In case of sliding window attention, the number of blocks in the entire sequence can be larger than the
+        // number of blocks in the cacheBlockIds vector (which holds only the blocks in the current attention window).
+        // In this case, get all blocks in the sequence by traversing back until nullptr.
+        if (blockKeys.size() > cacheBlockIds.size())
+        {
```
Funatiq, line 1439: Please refactor this part into a separate function.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1400,20 +1398,23 @@ SizeType32 KVCacheManager::getNeededBlocksOneStep(LlmRequest const& req, bool tw
...
+        auto const numNewBlocks = (numNextTokens > mMaxTokenNum) ? std::min(numNextBlocks - numPastBlocks, 1)
+                                                                 : (numNextBlocks - numPastBlocks);
+        numRequiredBlocks = numNewBlocks * req.mSamplingConfig.beamWidth;
     }
     return numRequiredBlocks;
 }
 
 SizeType32 KVCacheManager::getRemainingBlocksToCompletion(LlmRequest const& req) const
 {
+    if (req.isFinished())
```
Funatiq: Is this supposed to occur here? Should we use `TLLM_CHECK` instead?

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1445,12 +1446,29 @@ SizeType32 KVCacheManager::getRemainingBlocksToCompletion(LlmRequest const& req)
         }
     }
 
-    if (numAllocBlocksPerBeam < numContextBlocks)
+    // In case of sliding window attention, a new block is allocated when the window slides (and then the out-of-window
+    // block is detached) So we need an extra block for generation if the diff between the max sequence length and the
+    // current sequence length crosses a block boundary
+    auto const isSlidingWindow
+        = req.mPromptLen + req.mMaxNewTokens > mNumNonSinkTokensInWindow + mTemporaryAttentionWindow;
+    // req.getNumTokens() does NOT include sink tokens, so don't consider them in maxSeqlenInBlocks as well
```
Funatiq: Yes, it does. The sink tokens are just the first few tokens of a request.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1591,21 +1630,27 @@ void KVCacheManager::addSequence(
...
 
+    // Enable sliding window kv cache when inputLength exceeds maxAttentionWindow.
+    // Note that currently sliding window kv cache doesn't work with shared kv cache of different beams.
+    sequence.setContextRequiresSlidingWindowKvCache(inputLength > mMinAttentionWindow
+            + mSinkBubbleLength); // We decide at the outset whether a request uses sliding window kv-cache or not.
+
     // Get the final token index in kv cache
     SizeType32 const finalTokenKVIdx
-        = mSinkBlockTokenLength + ((inputLength - 1 - mSinkBlockTokenLength) % (mMaxTokenNum - mSinkBlockTokenLength));
+        = mSinkBlockTokenLength + ((inputLength - 1 - mSinkBlockTokenLength) % mNumNonSinkTokensInWindow);
```
Funatiq: The calculation for `finalTokenKVIdx` and `unsharedBlockIdx` also assume cyclic cache (hence the modulo). I don't think we need it anymore. We just have to make sure that the block where we split into multiple beams (can contain part of the context) is not `shareAmongBeams`, i.e. we need a separate block for each beam. It should be fine to allocate the additional blocks once we reach that point.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: This was leaked from my internal working process. Indeed shouldn't be here

cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h
tomeras91: Done

tomeras91: Done

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: Done

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: Done

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: This is indeed not needed. Removed to reduce diff from main

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: removed

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: I specifically checked this.
Check out the last testcase in the RemainingBlocksToCompletionCorrectlyEstimated test (--gtest_filter=RemainingBlocksToCompletionCorrectlyEstimated/RemainingBlocksToCompletionTest.RemainingBlocksToCompletionCorrectlyEstimated/5)

It has a request with 5000 prompt tokens and 4 sink tokens. req.getNumTokens(0) returns 5000.
This makes sense - sink tokens are defined at the KvCacheManager level, and not at the request level

tomeras91: Or do you mean that in this case, the 5000 prompt tokens are actually 4 sink tokens + 4996 "real" input tokens?

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: Actually, was `mUseOneMoreBlock` even ever `true`? AFAIU, it is used if cyclic kv cache + beam search are enabled. But there's a specific assertion (which we're now trying to remove) that makes sure cyclic kv cache cannot be used with beam seach [here](https://github.com/NVIDIA/TensorRT-LLM/blob/3e6264c95b68a5700e9ea3fa844a4c75ee39aa6b/cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp#L568)

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: OK got it now. There are no "special" sink tokens - they are just the first few tokens of the request. They are "real" tokens in the sense that they represent real text. 
I thought sink tokens were some special tokens that are prepended to the request, like the BOS token.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
Funatiq: Yeah, I think at some point cyclic kv cache + beam search was allowed or at least we tried to make it work by introducing `mUseOneMoreBlock`.

ming-wei: SGTM. Let's remove `useOneMoreBlock` in a separate PR (if it turns out beneficial after all).

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1403,52 +1374,66 @@ std::deque<tle::KVCacheEvent> BlockManager::getLatestEvents(std::optional<std::c
...
     auto const requestId = sequence.getRequestId();
-
     auto node = mAllocatedBlocksPerSeq.extract(requestId);
     TLLM_CHECK(node);
-    auto& allocatedBlocks = node.mapped();
-    for (auto it = allocatedBlocks.rbegin(); it != allocatedBlocks.rend(); ++it)
+    auto& sequenceBlocks = node.mapped();
+
+    // TODO: refactor this method in two: store blocks for reuse and just 'release blocks'. Only the caller
+    // can know which blocks to store for reuse and which to just release.
```
Funatiq, line 1391: Is this planned as a followup?

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1721,18 +1704,36 @@ SizeType32 KVCacheManager::getRemainingBlocksToCompletion(LlmRequest const& req,
...
 
-void KVCacheManager::cacheBlockOffsets(GenerationRequest& sequence, SizeType32 windowSize)
+void BlockManager::cacheBlockOffsets(GenerationRequest& sequence, SizeType32 windowSize)
 {
-    auto const& cacheBlocks = sequence.getCacheBlockIds(windowSize);
-    auto& cacheBlocksTensor = sequence.getCacheBlockIndices(windowSize);
+    mWindowBlockManagers.at(windowSize).cacheBlockOffsets(sequence);
+}
+
+void WindowBlockManager::cacheBlockOffsets(GenerationRequest& sequence)
```
Funatiq, line 1733: Method 'cacheBlockOffsets' can be made const (clang-tidy: readability-make-member-function-const)

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1744,15 +1745,15 @@ void KVCacheManager::cacheBlockOffsets(GenerationRequest& sequence, SizeType32 w
...
         {
             auto const blockId = beamCacheBlock.at(blockIdx);
-            mBlockManager.setOffsets(offsetsPtr, offsetsShape, beamIdx, blockIdx, blockId, windowSize);
+            setOffsets(offsetsPtr, offsetsShape, beamIdx, blockIdx, blockId);
         }
     }
 }
 
-void KVCacheManager::cacheNewBlockOffsets(GenerationRequest& sequence, SizeType32 windowSize)
+void WindowBlockManager::cacheNewBlockOffsets(GenerationRequest& sequence)
```
Funatiq, line 1753: Method 'cacheNewBlockOffsets' can be made const (clang-tidy: readability-make-member-function-const)

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1763,87 +1764,121 @@ void KVCacheManager::cacheNewBlockOffsets(GenerationRequest& sequence, SizeType3
...
     if (addToken)
     {
-        sequence.addNewTokens(1);
-    }
-    else
-    {
-        sequence.removeTokens(1);
-    }
+        auto const isBlockBoundary = (prevNumTokens % getTokensPerBlock() == 0);
+        auto const shouldAllocateBlock = isBlockBoundary && (!isCrossKv || newNumTokens <= metadata.maxTokenNum);
```
Funatiq, line 1791: I'm not exactly sure how cross kv works, but the comment
```
// cross KV cache doesn't grow after the initial context phase
```
suggests that we shouldn't even call `updateToken` for `isCrossKv`. Does that make sense?

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1763,87 +1764,121 @@ void KVCacheManager::cacheNewBlockOffsets(GenerationRequest& sequence, SizeType3
...
+            && ((numTokensWithoutSink - minTokensForBlockDetach) % getTokensPerBlock() == 0) && !isCrossKv;
 
-    for (auto const [windowSize, metadata] : mBlockManager.getWindowSizesMetadata())
-    {
-        auto const maxTokenNum = metadata.maxTokenNum;
-        SizeType32 const cyclicTokenNum = maxTokenNum - mSinkBlockTokenLength;
-        SizeType32 const nextTokenIdxInCycle = (currNumTokens - mSinkBlockTokenLength) % cyclicTokenNum;
-        SizeType32 const nextTokenIdxInCache = mSinkBlockTokenLength + nextTokenIdxInCycle;
+        if (canDetachBlock)
+        {
```
Funatiq, line 1799: Can we refactor this block into a separate function?

cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp
```diff
@@ -131,7 +131,7 @@ TEST_F(KVCacheManagerTest, BlockManagerTest)
 
     auto constexpr requestId = 42;
     GenerationRequest seq0{requestId, numTokens, beamWidth, blockManager.getWindowSizesMetadata()};
-    blockManager.addSequence(seq0, numBlocksPerBeam, numBlocksPerBeam - 1, maxAttentionWindow);
+    blockManager.addSequence(seq0, numBlocksPerBeam, false, maxAttentionWindow);
```
Funatiq, line 134: I think we should be more explicit here. Either add a comment that we add a request with/without `shareLastContextBlock` or use a separate variable for it.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1919,7 +1940,8 @@ void KVCacheManager::addSequence(
                     "have no effect.",
                     llmRequest->mRequestId);
             }
-            mBlockManager.addSequence(sequence, numContextBlocks, unsharedBlockIdx, windowSize);
+            auto const shareLastContextBlock = isCrossKv() || effectiveInputLength % getTokensPerBlock() == 0;
```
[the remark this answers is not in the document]
tomeras91: I don't think that's true..

Consider the following scenario:
* `inputLength = 9`
* `maxTokenNum = 5`, `temporaryAttentionWindow = 0`
* `getTokensPerBlock() = 4`

In that case, 
* `effectiveInputLength = 5`
* `sharedLength = 8`
=> `effectiveInputLength <= sharedLength` is `true`

But in this case, we don't want to share the last context block.. In this case, the context has 2 blocks, one full and another one with only a single token, so we can't share it.

The condition I had before, `effectiveInputLength % getTokensPerBlock() == 0`, evaluates to `false` which is what we want in this case.

Am I missing something?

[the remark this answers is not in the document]
Funatiq: But we do want to share the second block. It will only contain context tokens. Until the context phase is finished there are no beams which need separate blocks.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: Done

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: Done

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: Done

cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp
tomeras91: Done

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: I thought so too, but the capcity scheduler unittests specifically call `addToken` for a cross KV cache manager - see [here](https://github.com/NVIDIA/TensorRT-LLM/blob/e534bf09ccb7e2167beb59b220c1b948cece854e/cpp/tests/unit_tests/batch_manager/capacitySchedulerTest.cpp#L410). I guess there's some good reason for that.
Without this extra check, the test fails.

cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp
Funatiq: What I meant was to make it explicit for every call of `addSequence` because there is one other call that sets it to true. So basically set the variable explicitly before each call. The reason is that it should be easy to tell what's going on in the test by looking at a small, localized piece of code.

tomeras91: missed the call with `true`. Fixed it now

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1779,69 +1809,78 @@ void KVCacheManager::updateNewBlockPointer(GenerationRequest& sequence, SizeType
...
+
+void WindowBlockManager::updateSequenceBlocks(GenerationRequest& sequence, bool const addToken,
+    SizeType32 const sinkBlockTokenLength, bool const isCrossKv, bool const enableBlockReuse,
+    WindowSizeMetadata const& metadata)
 {
-    auto currNumTokens = sequence.getNumTokens();
+    auto const newNumTokens = sequence.getNumTokens();
+    auto const prevNumTokens = newNumTokens + (addToken ? (-1) : 1);
 
     if (addToken)
```
Funatiq, line 1835: Since this function has two completely different branches now, I think it would make sense to split it in separate functions again. There doesn't seem to be a reason to have common `updateToken` and `updateSequenceBlocks` functions for `add` and `remove`. What do you think?

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
tomeras91: After an internal conversation, removed call to `addToken` on a cross kv cache manager from unittest, and removed all crossKV related logic from `updateToken`. It now assumes (and asserts) that it is never called on a cross kv cache

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1744,33 +1743,64 @@ void KVCacheManager::cacheBlockOffsets(GenerationRequest& sequence, SizeType32 w
...
+        {
+            if (storeBlocksForReuse)
+            {
+                // If possible, offload the out-of-window block to secondary cache to save primary memory.
+                // Block can be released only after it is added to radix tree for reuse search, which is done
+                // after sequence is complete. Otherwise, we will create ophaned blocks and/or blocks that can't
+                // be freed since they are not leaf blocks.
+                if (getNumFreeBlocks(kSecondaryLevel) > 0)
+                {
+                    offloadBlock(outOfWindowBlock);
```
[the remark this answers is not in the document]
symphonylyh: Thanks @tomeras91, for the explanation. Basically it means the eviction process starts from the leaf and can iteratively trace to the root (if the entire path is clean with ref count = 0)
This brings some confusion around Robin's comment "we don't have a mechanism to automatically offload blocks that are not leaf blocks when blocks from the GPU pool are required for other requests. IMO it's a flaw in the current implementation" - if leaf block will strictly have a <= ref count than its parent blocks, the eviction does naturally happen from the leaf, right? I couldn't follow what is a scenario that we need to directly offload a non-leaf block?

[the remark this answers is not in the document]
symphonylyh: And @tomeras91 @schetlur-nv , should be further divide the KV reuse into two common scenarios:

1. **Reuse of the out-of-window (OOW) KV**. This is the KV reuse we have been mainly talking about. Like request0 is 8k with 7k OOW KV + 1k SWA KV. Now if request1 is 4k and request2 is 2k and both matches the prefix, we want to them to be able to reuse the KV from the OOW KV. This could be handled by either the current enforced CPU offloading (and onboarding again when cache hit) or more preferably the auto priority-based offloading (so they likely still stay on GPU when hit happens)
2. **Reuse of the sliding-window KV**. Discussion with the customers reminds me of this case today. This is mainly for multi-turn conversation scenario: request0 is 8k with 7k OOW KV + 1k SWA KV, and then request1 comes in as 9k (first 8k exactly matches request0, plus 1k new prompt in the current chat turn) -- in this case, actually we don't need to onboarding any of the 7k OOW KV. We just need to check the 7k OOW token match --> check the 1k SWA token match --> reuse the 1k SWA KV. Have we considered such scenario in the current design?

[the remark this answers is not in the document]
schetlur-nv: Right, the decision to offload should be based on whether there is a need to repurpose that block at a given time. If we don't need it (say if the load on the server is low), we should just leave it on-device and favor reuse opportunity.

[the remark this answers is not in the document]
tomeras91: Just to clarify - offloading a block does not mean it can't be reused. It just means that in case of a reuse opportunity, it needs to be onboarded first. 
So I wouldn't say immediate offload is detrimental to reuse. It does hurt it a bit because of the onboarding overhead, but reuse is still possible.
I do agree though that offloading when actually required is better. I just think it's a possible improvement, not a necessity.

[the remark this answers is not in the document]
schetlur-nv: Agreed/Understood.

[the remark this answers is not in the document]
symphonylyh: @tomeras91 thanks for the clarification. Yeah my "against reuse" is inaccurate, it's just not as efficient.
One more thing I want to double check: is this onboarding cost one-time?
e.g., if request0 immediately offloads 7k OOW KV to CPU & keeps 1k SWA KV on GPU. When request1 hits the 7k OOW tokens, the blocks get onboarded -- and they will stay in GPU, so if there is a request2 hits the 7k OOW tokens again, it's fast GPU reuse? or the same offloading logic will apply to these request2 OOW blocks again and push them to CPU?
If the latter, I think the improvement design could be things like: (1) have a property of priority/polularity that tracks how often a block is reused, including the reuse history from finished requests (2) evict not based on `RefCount=0`, but based on this popularity/priority. Do we already have things like this? it's like a dynamic priority

thorjohnsen:
I see some confusion regarding how offloading of blocks work. All blocks are represented by KVCacheBlock instances, which holds meta-data for the block, such as reference count, pointers to parent- and child-nodes in radix search tree and also poolBlockIndex.  poolBlockIndex can be thought of as a raw pointer into the memories we allocate to hold KV state for all blocks. When a block is offloaded, we acquire a free block in secondary memory and copy KV state from the primary block to the secondary block. The two KVCacheBlock instances then swap poolBlockIndex so that the offloaded block now points to location in secondary memory KV state was copied to. The location in primary memory that previously held KV state is released since it is no longer in use, this is achieved by placing the former offloadBlock on the eviction policy primary free queue.

PR #3430, opened 2025-04-09
feat: Nemotron-H model support

vegaluisjose:
The following adds support to Nemotron-H models.


juney-nvidia:
@QiJune  Hi Freddy, can you help review this MR?

Thanks
June

tensorrt_llm/_torch/models/modeling_nemotron_h.py
```diff
@@ -1,19 +1,18 @@
-from typing import Dict, Optional, Tuple
+from typing import Dict, Optional
 
 import torch
 from torch import nn
 from torch.nn import functional as F
+from transformer_engine.pytorch import RMSNorm
```
QiJune: I think it should be 

```python
from ..modules.mamba.layernorm_gated import RMSNorm
```

tests/unittest/_torch/modeling/test_modeling_mamba_hybrid.py
```diff
@@ -1,166 +0,0 @@
...
-    "transformers_version": "4.48.0.dev0",
-    "use_bias": False,
-    "use_cache": True,
-    "use_conv_bias": True,
-    "use_mamba_kernels": True,
-    "vocab_size": 131072,
-}
-
-
-class TestMambaHybrid(unittest.TestCase):
```
QiJune, line 68: I think we can keep it as a sanity check test for Nemotron-H model.

tensorrt_llm/_torch/modules/mamba/__init__.py
lfr-0531, line 1: All of the codes in mamba are copied from the 3rd party repo, right?

tensorrt_llm/_torch/models/modeling_nemotron_h.py
```diff
@@ -197,70 +190,44 @@ def __init__(self, model_config: ModelConfig[MambaHybridConfig]):
...
-        self.mamba_cache = None
-
     def forward(
         self,
         attn_metadata: AttentionMetadata,
         input_ids: Optional[torch.LongTensor] = None,
         position_ids: Optional[torch.LongTensor] = None,
         inputs_embeds: Optional[torch.FloatTensor] = None,
-        **kwargs,
+        spec_metadata: Optional[SpecMetadata] = None,
```
lfr-0531: If there is no speculative decoding support for this model, I think we can keep using `**kwargs` here.

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -70,6 +66,34 @@ def _create_kv_cache_manager(model_engine: PyTorchModelEngine, mapping: Mapping,
...
             if spec_config is None else spec_config.num_extra_kv_tokens,
         )
+    elif is_nemotron_hybrid(config):
+        config = model_engine.model.model_config.pretrained_config
+        num_layers = config.hybrid_override_pattern.count("*")
+        mamba_num_layers = num_mamba_layers = config.hybrid_override_pattern.count(
+            "M")
+        return MambaHybridCacheManager(
+            # mamba cache parameters
+            config.hidden_size,
```
lfr-0531: Can we add a new mamba_config? We can use this config to create the Mamba model and also here.

tensorrt_llm/_torch/models/modeling_nemotron_h.py
vegaluisjose: The transformer engine `RMSNorm` was the one used for the `NemotronHLayer` in Megatron and the `RMSNorm` from `layernorm_gated` is only used within the `MambaMixer` module.

tensorrt_llm/_torch/modules/mamba/__init__.py
vegaluisjose: yes

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
vegaluisjose: There is a config already defined in `modeling_nemotron_h.py`, link [here](https://github.com/NVIDIA/TensorRT-LLM/blob/eb74c0ff289e86cafe64da6c5486c9314cb5a27c/tensorrt_llm/_torch/models/modeling_nemotron_h.py#L35-L36).

Are you saying to add the fields used in here in the config just to make sure it would work once it reaches this point? say adding `hidden_size`, etc. Right now, we only have `model_type` in `NemotronHConfig`.

tensorrt_llm/_torch/models/modeling_nemotron_h.py
vegaluisjose: Done.

tests/unittest/_torch/modeling/test_modeling_mamba_hybrid.py
vegaluisjose: Done.

tensorrt_llm/_torch/models/modeling_nemotron_h.py
QiJune: `transformer_engine` is a new dependency. It's installed in NGC PyTorch container by default. However, bare metal Ubuntu container does not contain this library. cc @chzblych 

Can we port it via source codes? Or use another `RMSNorm` implementation?

vegaluisjose: I used try/except to do the import and later assert that `RMSNorm is not None` with a message saying model needs transformer engine and you should pip install it with this.

QiJune (approved):
LGTM

PR #3432, opened 2025-04-10
chore: code cleanup for error logging and SharedMemory in proxy.py

Superjomn:
Changes:

1. Changed several logging.error to logging.debug as there are unnecessary error logs when LLM shutdown.
2. Deprecating SharedMemory-related code in proxy.py as the ZMQ sends the tllm.Response with logit tensors directly without the need for SharedMemory.


tensorrt_llm/executor/ipc.py
```diff
@@ -1,19 +1,15 @@
...
 
-import torch
 import zmq
 import zmq.asyncio
 
 from tensorrt_llm.logger import logger
 
 from ..llmapi.utils import (ManagedThread, enable_llm_debug, nvtx_mark,
                             nvtx_range, print_colored, print_colored_debug)
-from .utils import ExecutorResponse, ExecutorResponseTensors
```
kaiyux, line 16: I believe this is the last part we are using `ExecutorResponse` - if we remove the usage here, I guess we should also remove those data structures?

Superjomn: Good catch, removed.

kaiyux (approved):

PR #3435, opened 2025-04-10
fix: Fix the issues related to fused moe path.

hyukn:
* One of the tactics is not supported during fused_moe kernel dispatch. Remove it for now.
* `final_hidden_states` should be unpacked if it is not in `min_latency_mode`.


litaotju (approved):

cpp/tensorrt_llm/kernels/cutlass_kernels/cutlass_heuristic.cpp
```diff
@@ -376,7 +376,9 @@ std::vector<CutlassGemmConfig> get_candidate_configs_sm100(CutlassGemmConfig::Ca
 
                     if (cluster_n == 1 && cluster_m == 1 && ((config & CutlassGemmConfig::FP8_ONLY) != 0))
                     {
-                        base.push_back(CutlassTileConfigSM100::CtaShape128x8x256B);
```
zongfeijing, line 379: I think we can directly remove this config, as fp8 does not support the current CTA shape.

hyukn: Thanks for the confirmation. Do we need to keep the corresponding dispatch logic, which has been commented out currently?

litaotju: Aligned with @zongfeijing offline, this one should be indeeded removed, and the comments inside the internal cutlass kernels shall be removed too, since its not supported.


@zongfeijing will help to remove the internal cutlass one in another repo.

@hyukn pls remove this one here. Thx

hyukn: Done.

PR #3436, opened 2025-04-10
feat: replace MLA context (192x128 packed) fmha from ampere-style to hopper-style

zhou-yuxin:


poweiw:
@zhou-yuxin is this PR still relevant?

PR #3447, opened 2025-04-10
chore: Rename nvsmall to nemotron nas

amitz-nv:
Note that `tests/unittest/trt/model/test_nemotron_nas.py` accesses `llm_models_root / "nvsmall/tests"`, so those paths are not updated in this PR, as a new hardlink with `nemotron_nas` should be created first.


Naveassaf (approved):
LGTM :raised_hands:

PR #3448, opened 2025-04-10
fix: nvbugs/5187237: fix deterministic mode crash

VALLIS-NERIA:
The accuracy issue still persists.

### DIFF
--- a/tensorrt_llm/llmapi/llm.py
+++ b/tensorrt_llm/llmapi/llm.py
@@ -517,8 +517,12 @@ class LLM:
         if self.args.kv_cache_config is not None:
             executor_config.kv_cache_config = PybindMirror.maybe_to_pybind(
                 self.args.kv_cache_config)
-        if os.getenv("FORCE_DETERMINISTIC", "0") == "1":
+        if os.getenv("FORCE_DETERMINISTIC",
+                     "0") == "1" and self.args.backend == "pytorch":
             # Disable KV cache reuse for deterministic mode
+            logger.info(
+                "FORCE_DETERMINISTIC is incompatible with KV cache reuse in PyTorch backend. KV cache reuse is disabled."
+            )
             executor_config.kv_cache_config.enable_block_reuse = False
             executor_config.kv_cache_config.enable_partial_reuse = False
         if self.args.peft_cache_config is not None:
--- a/tensorrt_llm/llmapi/llm.py
+++ b/tensorrt_llm/llmapi/llm.py
@@ -517,12 +517,8 @@ class LLM:
         if self.args.kv_cache_config is not None:
             executor_config.kv_cache_config = PybindMirror.maybe_to_pybind(
                 self.args.kv_cache_config)
-        if os.getenv("FORCE_DETERMINISTIC",
-                     "0") == "1" and self.args.backend == "pytorch":
+        if os.getenv("FORCE_DETERMINISTIC", "0") == "1":
             # Disable KV cache reuse for deterministic mode
-            logger.info(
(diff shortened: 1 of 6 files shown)


tensorrt_llm/plugin/plugin.py
```diff
@@ -706,8 +706,9 @@ def set_workspace_tensor(self,
     @staticmethod
     def max_workspace_size_auto(tp_size: int) -> int:
         if force_all_reduce_deterministic():
-            workspace_size = os.getenv("FORCE_ALLREDUCE_KERNEL_WORKSPACE_SIZE",
-                                       "1000000000")
+            workspace_size = os.getenv(
+                "FORCE_ALLREDUCE_KERNEL_WORKSPACE_SIZE",
+                "16000000")  # TODO: choose a reasonable default
```
VALLIS-NERIA: Need to explain how the number is selected

tensorrt_llm/llmapi/llm.py
```diff
@@ -516,6 +516,10 @@ def _build_model(self):
         if self.args.kv_cache_config is not None:
             executor_config.kv_cache_config = PybindMirror.maybe_to_pybind(
                 self.args.kv_cache_config)
+        if os.getenv("FORCE_DETERMINISTIC", "0") == "1":
```
VALLIS-NERIA, line 520: Verify whether kv cache reuse is incompatible with FORCE_DETERMINISTIC, or it is just a bug of kv cache reuse function.

dongxuy04 (approved):

tensorrt_llm/plugin/plugin.py
```diff
@@ -767,12 +767,13 @@ def allocate_allreduce_fusion_workspace(
...
         import torch
+        force_deterministic = force_all_reduce_deterministic()
         is_p2p_supported = can_access_peer(mapping)
-        ipc_buffers_size = size * mapping.tp_size
+        ipc_buffers_size = size if force_deterministic else size * mapping.tp_size
         ipc_buffers = IpcMemory(mapping, ipc_buffers_size, is_p2p_supported)
         ipc_barriers = IpcMemory(mapping, 256 * mapping.tp_size,
                                  is_p2p_supported)
-        lamport_buffers_size = size * mapping.tp_size
+        lamport_buffers_size = 1 if force_deterministic else size * mapping.tp_size
```
yilin-void: There is no support for `force_deterministic` in this path yet, so do not modify these buffer size here, as it will cause illegal memory access at runtime.

yizhang-nv: Do allreduce fusion kernels support deterministic? I though for lamport kernels, we've already supported it. Maybe the only thing we should do is to disable the two shot kernels. So that the workspace part could leave untouched.

yilin-void: I think the output of the new kernel's Lamport and twoshot kernels should be consistent, as the computation process for the same token is identical in both kernels. But the selection of NCCL/Custom kernels is done within the torch op, so we need modify the torch op to achieve determinism.

yizhang-nv: I get it. So the no support for `force_deterministic` means the op doesn't support it but the kernel does. Since this part is only related to the kernel itself, we don't need to modify this part.

yilin-void: Yes.

VALLIS-NERIA: > There is no support for force_deterministic in this path

Does this mean the argument `size` should be same as normal path, 16MB or 8MB, instead of 1GB of force deterministic path?

Issue #1775, opened 2024-06-13
ChatGLM3 6B Multi-batch Failed with Error


RobinJYM:
### System Info

- CPU: INTEL RPL
- GPU Name: NVIDIA GTX 4090
- TensorRT-LLM: tensorrt_llm==0.11.0.dev2024060400
- Container Used: Yes and reproduced in Conda as well
- Driver Version: 555.42.02
- CUDA Version: 12.5
- OS: Ubuntu 24.04
- Docker Img: nvidia/cuda:12.5.0-devel-ubuntu22.04

### Who can help?

@hijkzzz 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

python3 benchmarks/python/benchmark.py --engine_dir /trt_engines/chatglm3_6b/float16/1-gpu/ --dtype float16 --batch_size 16 --input_output_len "1024,512"

### Expected behavior

 pass with output

### actual behavior

```
[TensorRT-LLM] TensorRT-LLM version: 0.11.0.dev2024061100
Allocated 770.13 MiB for execution context memory.
/usr/local/lib/python3.10/dist-packages/torch/nested/__init__.py:166: UserWarning: The PyTorch API of nested tensors is in prototype stage and will change in the near future. (Triggered internally at ../aten/src/ATen/NestedTensorImpl.cpp:178.)
  return _nested.nested_tensor(
[06/13/2024-03:26:19] [TRT] [E] 3: [executionContext.cpp::setInputShape::2068] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::setInputShape::2068, condition: satisfyProfile Runtime dimension does not satisfy any optimization profile.)
[06/13/2024-03:26:19] [TRT] [E] 3: [executionContext.cpp::setInputShape::2068] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::setInputShape::2068, condition: satisfyProfile Runtime dimension does not satisfy any optimization profile.)
[06/13/2024-03:26:19] [TRT] [E] 3: [executionContext.cpp::resolveSlots::2842] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::resolveSlots::2842, condition: allInputDimensionsSpecified(routine) )
[...]

PR #3450, opened 2025-04-10
chore: Unify Python NVTX call

kaiyux:
* Move and unify NVTX implementation to `tensorrt_llm/_utils.py`
* Replace the original LLM API NVTX with `nvtx_range_debug` call

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -49,4 +49,4 @@ opencv-python-headless
 xgrammar==0.1.16
 backoff
 nvtx
-matplotlib
+matplotlib # FIXME: this is added to make nvtx happy
--- a/requirements.txt
+++ b/requirements.txt
@@ -49,3 +49,4 @@ opencv-python-headless
 xgrammar==0.1.16
 backoff
 nvtx
+matplotlib
--- a/requirements.txt
+++ b/requirements.txt
@@ -48,3 +48,4 @@ flashinfer-python~=0.2.3
 opencv-python-headless
 xgrammar==0.1.16
 backoff
+nvtx
--- a/tensorrt_llm/_utils.py
+++ b/tensorrt_llm/_utils.py
@@ -17,10 +17,10 @@ import gc
 import inspect
 import json
 import math
+import os
 import struct
 import trace
 import weakref
-from contextlib import contextmanager
 from dataclasses import asdict
 from enum import EnumMeta
 from functools import partial, wraps
@@ -28,6 +28,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, Union
 
 import numpy as np
+import nvtx
 from cuda import cuda
 from mpi4py import MPI
 from packaging import version
@@ -733,13 +734,18 @@ class QuantModeWrapper:
         return self.objs[index]
 
 
-@contextmanager
-def nvtx_range(msg):
-    torch.cuda.nvtx.range_push(msg)
-    try:
-        yield
-    finally:
-        torch.cuda.nvtx.range_pop()
+def nvtx_range(msg, color="grey", domain="TensorRT-LLM", category=None):
+    return nvtx.annotate(msg, color=color, domain=domain, category=category)
+
+
+def nvtx_range_debug(msg, color="grey", domain="TensorRT-LLM", category=None):
+    if os.getenv("TLLM_LLMAPI_ENABLE_NVTX", "0") == "1" or \
+            os.getenv("TLLM_NVTX_DEBUG", "0") == "1":
+        return nvtx_range(msg, color=color, domain=domain, category=category)
+
+
+def nvtx_mark(msg, color="grey", domain="TensorRT-LLM", category=None):
+    nvtx.mark(msg, color=color, category=category, domain=domain)
 
 
 def volume(d: Sequence[int]):
--- a/tensorrt_llm/_utils.py
+++ b/tensorrt_llm/_utils.py
@@ -740,11 +740,48 @@ def _null_context_manager():
     yield
 
 
-def nvtx_range(msg, color="grey", domain="TensorRT-LLM", category=None):
+def nvtx_range(msg: str,
+               color: str = "grey",
+               domain: str = "TensorRT-LLM",
+               category: Optional[str] = None):
+    """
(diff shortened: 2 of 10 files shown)


Superjomn:
We mentioned having nvtx levels in an earlier discussion for tradeoff between details and performance with nsys, I noticed this MR introduces two levels, `nvtx_range` and `nvtx_range_debug`, are these two levels sufficient? Or will we introduce more or per-module switching in the future? @kaiyux @FrankD412 

And I noticed this PR move nvtx_range to `._utils.py`, will you unify the torch.cuda.nvtx_range usages in other places like pyexecutor? @kaiyux

kaiyux:
@Superjomn Thanks for the comment

> are these two levels sufficient? Or will we introduce more or per-module switching in the future?

I can't think of a scenario that we will need more levels - we already have layerwise nvtx markers as a feature. 

> And I noticed this PR move nvtx_range to `._utils.py`, will you unify the torch.cuda.nvtx_range usages in other places like pyexecutor? 

It should already be unified in this PR.

Superjomn:
> @Superjomn Thanks for the comment
> 
> > are these two levels sufficient? Or will we introduce more or per-module switching in the future?
> 
> I can't think of a scenario that we will need more levels - we already have layerwise nvtx markers as a feature.
> 
> > And I noticed this PR move nvtx_range to `._utils.py`, will you unify the torch.cuda.nvtx_range usages in other places like pyexecutor?
> 
> It should already be unified in this PR.

Got it; that is clear. Thanks for your reply.

requirements.txt
```diff
@@ -48,3 +48,5 @@ flashinfer-python~=0.2.3
 opencv-python-headless
 xgrammar==0.1.16
 backoff
+nvtx
+matplotlib
```
Superjomn: Why was this added? I think it is not used, right?

tensorrt_llm/_utils.py
```diff
@@ -734,12 +736,24 @@ def __getitem__(self, index):
 
 
 @contextmanager
-def nvtx_range(msg):
-    torch.cuda.nvtx.range_push(msg)
-    try:
-        yield
-    finally:
-        torch.cuda.nvtx.range_pop()
+def _null_context_manager():
```
Superjomn, line 739: If nvtx is added to the requirement, is this necessary? Or should we make nvtx an item only in `requirement-dev.txt`?

requirements.txt
kaiyux: > [PR_Github #1775](https://nv/trt-llm-cicd/job/helpers/job/PR_Github/1775/) [ run ] completed with state SUCCESS
[/LLM/main/L0_MergeRequest_PR pipeline #1316](https://nv/trt-llm-cicd/job/main/job/L0_MergeRequest_PR/1316/) completed with status: 'FAILURE'

The CI pipeline was failed because `nvtx` depends on `matplotlib`.

tensorrt_llm/_utils.py
kaiyux: There are basic `nvtx` markers always enabled in `py_executor.py`.

kaiyux: Oh I see what you meant - we need `_null_context_manager` to handle the case when `TLLM_LLMAPI_ENABLE_NVTX` and `TLLM_NVTX_DEBUG` are not specified.

requirements.txt
Superjomn: Ideally, pip install nvtx should install its dependencies; it is unnecessary to add matplotlib explicitly. Is there anything wrong there?

tensorrt_llm/_utils.py
Superjomn: Got it, that makes sense.

tensorrt_llm/executor/result.py
```diff
@@ -223,7 +224,9 @@ def _handle_sequence(self, finish_reasons, response_tensors,
                 raise ValueError(
                     f"Unknown finish reason: {finish_reasons[src_idx]}")
 
-    @nvtx_range("handle_response", color="red", category="GenerationResultBase")
+    @nvtx_range_debug("handle_response",
```
FrankD412, line 223: I was confused for a second why this was being used as a decorator, turns out that it's because the `@contextmanager` decorator works with functions that return a generator. Bit of a nit pick, but is there 1. A plan to put type hinting here? 2. Can we document the usage so that it's clear?

That said, I learned something and it's a nuance but still neat!

requirements.txt
FrankD412: I agree, this should be fixed upstream.

FrankD412:
> We mentioned having nvtx levels in an earlier discussion for tradeoff between details and performance with nsys, I noticed this MR introduces two levels, `nvtx_range` and `nvtx_range_debug`, are these two levels sufficient? Or will we introduce more or per-module switching in the future? @kaiyux @FrankD412 
> 
> 
> 
> And I noticed this PR move nvtx_range to `._utils.py`, will you unify the torch.cuda.nvtx_range usages in other places like pyexecutor? @kaiyux 

I think two levels is fine -- when we profile we want as much information as possible so as long as it's documented how to do that I think we're happy on our end.

tensorrt_llm/executor/result.py
kaiyux: @FrankD412 Thanks for the comments, can you please check if my latest [commit](https://github.com/NVIDIA/TensorRT-LLM/pull/3450/commits/46d67fa03c65965c1723bd46c46d9371434f106b) address your questions?

requirements.txt
kaiyux: Not sure if it's because `nvtx` does not add `matplotlib` to its dependency. I can double check.

kaiyux: It does not hurt if it's really depended though, right? We will need to install it anyway.

Superjomn: Can you add a comment to describe the context? Since it is just for amending the nvtx's dependency, not the direct dependency of TRTLLM.

Superjomn (approved):
LGTM

requirements.txt
kaiyux: Added, thanks.

FrankD412 (approved):

PR #3455, opened 2025-04-10
feat/loraOp 

danielafrimi:
1. A new C++ file (`cpp/tensorrt_llm/thop/loraOp.cpp`) was added.
This file defines a PyTorch custom operator named `trtllm::lora_grouped_gemm.`

2. The forward method of the LoraLayer class was modified. it now calls the newly created custom operator `torch.ops.trtllm.lora_grouped_gemm` (the file above).

3. Tests: 

- MLP with lora (loraOp)
- Attention with lora (loraOp) - as the pytorch flow assume the input is not padded, we compare both of the attention layers with no padding (loraOp changed  a little bit).
- Comparison of `loraPlugin` vs `loraOp`


cpp/tensorrt_llm/thop/loraOp.cpp
```diff
@@ -0,0 +1,192 @@
...
+    auto stream = at::cuda::getCurrentCUDAStream().stream();
+    auto const numReqs = lora_ranks[0].sizes()[0];
+    auto const out_shape = input.sizes();
+    int const numLoraModules = lora_ranks.size();
+    TLLM_CHECK_WITH_INFO(lora_ranks.size() == lora_weights_pointers.size(), "both should be numLoraModules");
+    std::vector<th::Tensor> output_torch;
+    for (int i = 0; i < numLoraModules; i++)
+    {
+        std::vector<int64_t> output_shape = {out_shape[0], out_shape[1]};
+        if (!isRemoveInputPadding)
```
danielafrimi, line 68: @byshiue 
those lines are the only changes i did, where output shape determine by the `output_hidden_sizes` that we get as input. in addition, when we get ragged tensor (no padded), output shape is like the input shape without bs dim.

danielafrimi:
@Naveassaf @shaharmor98 @byshiue 

PR is open for feedback

cpp/tensorrt_llm/thop/loraOp.cpp
byshiue: For the case that isRemoveInputPadding == True, shouldn't the output_shape be `{output_shape[0], output_hidden_size[i]}`?

tests/unittest/_torch/modules/tests_lora_modules/test_lora_mlp_pytorch_flow_vs_trt.py
```diff
@@ -0,0 +1,283 @@
+import unittest
```
byshiue: I don't think we need to use trt as reference all the time. We can use torch as reference directly to prevent the overhead of building engine, the complicated codes of initializing the TRT engine and inputs.

cpp/tensorrt_llm/thop/loraOp.cpp
danielafrimi: yes, you're right. thanks for notice

Naveassaf:
Spoke about this PR over slack. It adds incomplete functionality along with passing, none-flaky tests.

The updates to this are already WIP and about to be placed in a separate PR.

Naveassaf (approved):

PR #3457, opened 2025-04-10
test:add fp8_kv_cache functionality test case.

nv-guomingz:
Don't merge this PR until modelopt 0.27.1 release.

### DIFF
--- a/tests/integration/defs/examples/test_llama.py
+++ b/tests/integration/defs/examples/test_llama.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 import copy
 import csv
+import json
 import os
 import re
 import shutil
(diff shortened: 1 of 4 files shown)


tests/integration/defs/examples/test_llama.py
```diff
@@ -620,6 +621,55 @@ def test_llm_llama_1gpu(run_type, data_type, fp8_cache, llama_example_root,
         venv_check_call(llm_venv, summary_cmd)
 
 
+@skip_pre_ada
+@pytest.mark.parametrize("data_type", ['bfloat16', 'float16'])
```
litaotju: @nv-guomingz since we care most about FP8 kv cache implementation/workflow here. I think its fine to only use one of bf16 or fp16 to save time. How do you think?

These 2 paths won't affect the fp8 kv cache.

litaotju (approved):

tests/integration/defs/examples/test_llama.py
nv-guomingz: make sense. Code updated.

PR #3459, opened 2025-04-10
feat: [Deepseek] Redesign multi-stream API

hlu1:
This PR applied the new API to Deepseek only. Llama will be updated later.

Applied multi-stream to overlap bmm and concat in `MLA.forward_generation`.

### DIFF
--- a/tensorrt_llm/_torch/modules/attention.py
+++ b/tensorrt_llm/_torch/modules/attention.py
@@ -15,6 +15,7 @@ from ..distributed import AllReduceParams
 from ..model_config import ModelConfig
 from ..peft.lora.layer import LoraLayer, LoraModuleType
 from .linear import Linear, TensorParallelMode, WeightMode, WeightsLoadingConfig
+from .multi_stream_utils import maybe_execute_in_parallel
 from .rms_norm import RMSNorm
 from .rotary_embedding import RotaryEmbedding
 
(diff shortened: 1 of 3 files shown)


tensorrt_llm/_torch/modules/attention.py
SimengLiu-nv, line 1: Attention Op has multi-stream part as well here: https://github.com/NVIDIA/TensorRT-LLM/blob/c0ab93938a4881ac0213034c80e18289f98a4125/tensorrt_llm/_torch/modules/attention.py#L187-L99

SimengLiu-nv (approved):

QiJune (approved):
LGTM

tensorrt_llm/_torch/modules/multi_stream_utils.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any, Callable, Optional
+
+import torch
+
+
+def maybe_execute_in_parallel(
+        fn0: Callable, fn1: Callable, event0: torch.cuda.Event,
+        event1: torch.cuda.Event,
+        aux_stream: Optional[torch.cuda.Stream]) -> tuple[Any, Any]:
```
QiJune:
```suggestion
        aux_stream: Optional[torch.cuda.Stream]=None) -> tuple[Any, Any]:
```

tensorrt_llm/_torch/modules/attention.py
hlu1: That's used by llama. I'll leave the llama changes to another PR because I need to test llama4.

tensorrt_llm/_torch/modules/multi_stream_utils.py
```diff
@@ -0,0 +1,46 @@
...
+        event0 (torch.cuda.Event): cuda event for fn0
+        event1 (torch.cuda.Event): cuda event for fn1
+        aux_stream (Optional[torch.cuda.Stream]): the second cuda stream for fn1.
+            Mutil-stream is disabled when aux_stream is None.
+
+    Returns:
+        tuple[Any, Any]: the return values of fn0() and fn1()
+    """
+
+    do_multi_stream = torch.cuda.is_current_stream_capturing(
```
liji-nv: Use is_graph_capturing() utility in pyexecutor.cuda_graph_runner to avoid graph break.

PR #3462, opened 2025-04-11
fix: switch ZMQ from file socket to tcp socket in RemoteMpiCommSession

Superjomn:
For zmq.PAIR based on file socket requires the `bind` side to set up first and then the `connect` side, or there may be unknown issues. This PR switched the file socket to TCP socket to eliminate such requirement and enhance stability.

### DIFF
--- a/tensorrt_llm/llmapi/mpi_session.py
+++ b/tensorrt_llm/llmapi/mpi_session.py
@@ -232,12 +232,11 @@ class RemoteMpiCommSessionClient():
     '''
 
     def __init__(self, addr: str):
+        from tensorrt_llm.executor.ipc import ZeroMqQueue
+        self.addr = addr
         print_colored_debug(
             f"RemoteMpiCommSessionClient connecting to {addr}\n", "yellow")
-        self.addr = addr
-        self.context = zmq.Context()
-        self.socket = self.context.socket(zmq.PAIR)
-        self.socket.connect(addr)
+        self.queue = ZeroMqQueue(addr, is_server=False)
         self._is_shutdown = False
 
     def submit(self, task: Callable[..., T], *args, **kwargs) -> list:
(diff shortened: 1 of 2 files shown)


syuoni (approved):
LGTM

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -232,12 +232,11 @@ class RemoteMpiCommSessionClient():
     '''
 
     def __init__(self, addr: str):
+        from tensorrt_llm.executor.ipc import ZeroMqQueue
```
kaiyux, line 236: Any special reason we need to import here instead of on the top of this file?

kaiyux (dismissed):

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -281,10 +277,9 @@ def __init__(self,
                  addr: str = f'tcp://127.0.0.1:*',
                  comm=None,
                  is_comm: bool = False):
-        self.context = zmq.Context()
-        self.socket = self.context.socket(zmq.PAIR)
-        self.socket.bind(addr)
-        self.addr = self.socket.getsockopt(zmq.LAST_ENDPOINT).decode()
+        from tensorrt_llm.executor.ipc import ZeroMqQueue
```
kaiyux, line 282: I'm still concerned those duplicated imports could be easily leading to some issues, unless there are specific reason I guess we should avoid this.

Superjomn: This is used to avoid cyclic importing.

tensorrt_llm/llmapi/mpi_session.py
Superjomn: Same reason as below.

tensorrt_llm/llmapi/mpi_session.py
kaiyux: If we start seeing cyclic importing, then it means there are already issues on the module design, isn't it? I feel uncomfortable resolving cyclic importing this way because it's only bypassed on this case.

kaiyux: @syuoni WDYT?

Superjomn: Yeah, will resolve it in the following PRs. This is a quick fix maybe we can merge it first. Does this work for you?

Superjomn: @kaiyux  Just add two FIXME there, I will resolve them in bulk in a subsequent PR.

kaiyux: Thanks!

kaiyux (approved):

PR #3468, opened 2025-04-11
test: Automatically clean checkpoints and engines

syuoni:


tests/integration/defs/accuracy/accuracy_core.py
```diff
@@ -303,24 +304,22 @@ def setup_class(cls, request):
...
+    def setup_method(self):
+        model_workspace = f"{self.llm_venv.get_working_directory()}/{self.MODEL_NAME}"
+        if not exists(model_workspace):
+            makedirs(model_workspace)
+        with tempfile.TemporaryDirectory(dir=model_workspace) as workspace:
+            self.ckpt_dir = f"{workspace}/cmodels"
+            makedirs(self.ckpt_dir)
+            self.engine_dir = f"{workspace}/engines"
+            makedirs(self.engine_dir)
+            yield
```
syuoni, line 314: Hi @xinhe-nv ,

I use `TemporaryDirectory` so that the engines/checkpoints can be automatically cleaned when a test case finishes. 

We don't need to do this for LLM API, because LLM also manages engines in a `TemporaryDirectory`, which is automatically cleaned when the LLM instance is deconstructed. cc @Superjomn

xinhe-nv (approved):

tests/integration/defs/accuracy/accuracy_core.py
Superjomn: Yes, for the TRT path, LLM-API will clean up the engines and checkpoints automatically if users don't call `LLM.save` explicitly. The default path to hold those temporary files is `/tmp`, and users can specify a directory via `LLM(workspace=...)`.

PR #3476, opened 2025-04-11
chore: add llama mlp_bias reading from hf_config

WhatGhost:
When i want to convert and build a llama model with attn_bias and mlp_bias. I noticed that in the config of trtllm's ckpt, the mlp bias is always false even i add mlp_bias in hf_model's config.

So i check the code of trtllm, I find the llama's config.py doesn't read mlp_bias param like the attn_bias.
So i add it.And the model can be converted and built correctly.


juney-nvidia:
@WhatGhost Hi, thanks for the contribution.  Pls do the sign-off of your own information to pass the DCO check firstly.

Thanks
June

WhatGhost:
@juney-nvidia hi,i have sign-off follow the guide.

WhatGhost:
@juney-nvidia Hi,Is there any work I need to do

PR #3479, opened 2025-04-11
refactor: Clean up CMakeLists.txt

tongyuantongyu:
- Use Modern CMake style to find dependencies.
- Split out CUDA compiler and architecture config logic.
- Split out Sanitizers config logic.


cpp/CMakeLists.txt
```diff
@@ -139,158 +137,21 @@ configure_file(
...
-endforeach()
-set(CMAKE_CUDA_ARCHITECTURES ${CMAKE_CUDA_ARCHITECTURES_NORMALIZED})
+setup_cuda_compiler()
+setup_cuda_architectures()
 
 enable_language(C CXX CUDA)
 
-find_package(CUDAToolkit REQUIRED)
+find_package(CUDAToolkit 11.2 REQUIRED COMPONENTS cudart_static cuda_driver
+                                                  cublas cublasLt curand)
```
WilliamTambellini: nvml seems to also be a must have, right?

cpp/CMakeLists.txt
```diff
@@ -303,58 +164,19 @@ message(STATUS "    include path: ${CUDAToolkit_INCLUDE_DIRS}")
...
   # NCCL dependencies
-  set_ifndef(NCCL_LIB_DIR /usr/lib/${CMAKE_SYSTEM_PROCESSOR}-linux-gnu/)
-  set_ifndef(NCCL_INCLUDE_DIR /usr/include/)
-  find_library(NCCL_LIB nccl HINTS ${NCCL_LIB_DIR})
+  find_package(NCCL 2 REQUIRED)
+  set(NCCL_LIB NCCL::nccl)
 endif()
 
+# TRT dependencies
+find_package(TensorRT 10 REQUIRED COMPONENTS OnnxParser)
```
WilliamTambellini, line 179: Apparently, 10.7 not supported anymore, any way here to specify the minor version too: eg "10.8" ?

WilliamTambellini (commented):
cool. Tks.

cpp/CMakeLists.txt
tongyuantongyu: Added. Thanks.

cpp/CMakeLists.txt
tongyuantongyu: We are specifying verified and hence recommended version of dependencies through our official docker image and its build script:
https://github.com/NVIDIA/TensorRT-LLM/blob/fe6f14b2b1c801906202d387cd5b3b6038aa069b/docker/common/install_tensorrt.sh#L5-L18

The version in `CMakeLists.txt` will not be regularly updated and is only for basic sanity check (such that unsatisfied version likely won't even compile).

MartinMarciniszyn (approved):

tongyuantongyu:
Do to #3282, "Support building internal cutlass kernels in the same CMake build" need some non-trivial changes, so I'm going to split that part into another PR and let other parts merged first.

tongyuantongyu:
/bot run --stage-list "A30-CPP-2, A30-TensorRT-3, L40S-PyTorch-1"

cpp/cmake/modules/FindNCCL.cmake
```diff
@@ -0,0 +1,74 @@
...
+      "${NCCL_VERSION_MAJOR}.${NCCL_VERSION_MINOR}.${NCCL_VERSION_PATCH}"
+      PARENT_SCOPE)
+endfunction()
+
+_nccl_get_version()
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(
+  NCCL
+  FOUND_VAR NCCL_FOUND
```
zeroepoch, line 58: What happens when NCCL is not found?

cpp/cmake/modules/FindTensorRT.cmake
```diff
@@ -0,0 +1,177 @@
...
+        TensorRT_Plugin_DLL "nvinfer_plugin_${TensorRT_VERSION_MAJOR}.dll"
+        nvinfer_plugin.dll)
+    endif()
+  endif()
+endif()
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(
+  TensorRT
+  FOUND_VAR TensorRT_FOUND
```
zeroepoch, line 124: Similar comment here. Do we handle the cases when TensorRT is not found?

examples/cpp/executor/CMakeLists.txt
```diff
@@ -13,11 +13,13 @@
...
 
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.27)
 
 set(TRTLLM_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../..")
-include(${TRTLLM_DIR}/cpp/cmake/modules/set_ifndef.cmake)
-include(${TRTLLM_DIR}/cpp/cmake/modules/find_library_create_target.cmake)
+list(APPEND CMAKE_MODULE_PATH "${TRTLLM_DIR}/cpp/cmake/modules")
+
+include(set_ifndef)
```
zeroepoch: I don't think `set_ifndef` is used in this file.

zeroepoch (approved):

examples/cpp/executor/CMakeLists.txt
tongyuantongyu: Nice catch. I realized that we no longer using this macro anywhere, so we can get rid of it.

cpp/cmake/modules/FindNCCL.cmake
tongyuantongyu: `find_package` will recognize the situation. And since we are passing `REQUIRED` to it:
https://github.com/NVIDIA/TensorRT-LLM/blob/974a2549a332678370dce6b6946bf1037a9ce1b1/cpp/CMakeLists.txt#L174

CMake configuration will fail at this line, and telling user that it is unable to find NCCL due to unable to determine `NCCL_LIBRARIES` and/or `NCCL_INCLUDE_DIR` variable, and hint the user to provide the path to NCCL installation via `-DNCCL_ROOT`.

PR #3481, opened 2025-04-11
fix: build engine for TRT path in slurm

Superjomn:
Add submit_sync to MpiCommSessionClient to support trtllm-bench build.

### DIFF
--- a/tensorrt_llm/llmapi/llm_utils.py
+++ b/tensorrt_llm/llmapi/llm_utils.py
@@ -40,7 +40,7 @@ from .tokenizer import TransformersTokenizer, load_hf_tokenizer
 # TODO[chunweiy]: move the following symbols back to utils scope, and remove the following import
 from .utils import (download_hf_model, download_hf_pretrained_config,
                     enable_llm_debug, get_directory_size_in_gb, print_colored,
-                    print_traceback_on_error)
+                    print_colored_debug, print_traceback_on_error)
 
 
 @dataclass
@@ -488,6 +488,7 @@ class ModelLoader:
         self._model_info = _ModelInfo.from_pretrained_config(
             self.pretrained_config)
 
+    @print_traceback_on_error
     def _load_model_from_ckpt(self):
         ''' Load a TRT-LLM model from checkpoint. '''
         self.pretrained_config = PretrainedConfig.from_json_file(
@@ -515,10 +516,14 @@ class ModelLoader:
         assert isinstance(self.llm_args.model, Module)
         self._model_info = _ModelInfo.from_module(self.model)
 
+    @print_traceback_on_error
     def _build_engine(self):
         assert isinstance(
             self.build_config,
             BuildConfig), f"build_config is not set yet: {self.build_config}"
+
+        print_colored_debug(f"rank{mpi_rank()} begin to build engine...\n",
+                            "green")
 
         # avoid the original build_config is modified, avoid the side effect
         copied_build_config = copy.deepcopy(self.build_config)
@@ -535,6 +540,7 @@ class ModelLoader:
 
         # delete the model explicitly to free all the build-time resources
         self.model = None
+        print_colored_debug(f"rank{mpi_rank()} build engine done\n", "green")
 
     def _save_engine_for_runtime(self):
         '''
--- a/tensorrt_llm/llmapi/mpi_session.py
+++ b/tensorrt_llm/llmapi/mpi_session.py
@@ -4,6 +4,8 @@ import os
 import socket
 import sys
 import threading
+import time
+import traceback
 from collections.abc import Callable
 from concurrent.futures import Future, ThreadPoolExecutor
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple, TypeVar
@@ -13,8 +15,8 @@ import zmq
 from tensorrt_llm.bindings.BuildInfo import ENABLE_MULTI_DEVICE
 from tensorrt_llm.logger import logger
 
-from .._utils import global_mpi_rank
-from .utils import print_colored_debug
+from .._utils import global_mpi_rank, mpi_barrier, mpi_rank
+from .utils import print_colored, print_colored_debug
 
 if ENABLE_MULTI_DEVICE:
     import mpi4py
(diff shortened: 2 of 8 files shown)


tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -809,11 +815,19 @@ def _node_build_task(
...
-            model_loader(engine_dir=engine_dir)
-            return model_loader.llm_build_stats.build_steps_info
+        try:
+            with ModelLoader(llm_args,
+                             workspace=workspace,
+                             llm_build_stats=llm_build_stats) as model_loader:
+                model_loader(engine_dir=engine_dir)
+                return model_loader.llm_build_stats.build_steps_info
+        except Exception as e:
+            import traceback
```
kaiyux: Should we move this to the top of this file?

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -272,19 +275,51 @@ def __init__(self, addr: str):
         self.queue = ZeroMqQueue(addr, is_server=False)
         self._is_shutdown = False
 
-    def submit(self, task: Callable[..., T], *args, **kwargs) -> list:
+    def submit(self,
+               task: Callable[..., T],
+               sync: bool = False,
+               *args,
+               **kwargs) -> None:
```
kaiyux: Should still be the following?
```suggestion
               **kwargs) -> []:
```
Or you want to return `None`?

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -272,19 +275,51 @@ def __init__(self, addr: str):
...
-    def submit_sync(self, task, *args, **kwargs):
-        return self.submit(task, *args, **kwargs)
+    def submit_sync(self, task, *args, **kwargs) -> List[T]:
+        ''' Submit a task to the remote MPI pool and wait for task completion. '''
+        self.submit(task, sync=True, *args, **kwargs)
+
+        while not (res := self.poll() or self._is_shutdown):
+            print_colored_debug(f"Waiting for task completion... {res}\n",
+                                "grey")
+            time.sleep(8)
```
kaiyux: I'm a little bit concerned on the hard-coded `8` here, maybe we should at least make something like this?
```
# top of this file
TASK_COMPLETION_IDLE=8

time.sleep(TASK_COMPLETION_IDLE)
```

kaiyux (commented):
Sorry - I have to admit that I'm not 100% clear on how the changes can fix the issue, I can follow up with you offline to clear my understanding.

Superjomn:
Tested ten times, and all passed.

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -272,19 +278,53 @@ def __init__(self, addr: str):
...
+                                "grey")
+            time.sleep(self.SYNC_IDLE_INTERVAL)
+
+        print_colored_debug(
+            f"rank{global_mpi_rank()} RemoteMpiCommSessionClient.send_sync received results: {res}\n",
+            "green")
+
+        if not res:
+            raise RuntimeError(
+                "RemoteMpiCommSessionClient received unexpected response")
```
syuoni, line 316: IIUC, `res` must be `True` according to LIne 303. 

Or, Line 303 should be the following?

```python
while not ((res := self.poll()) or self._is_shutdown):
```

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -272,19 +278,53 @@ def __init__(self, addr: str):
...
+        return res
+
+    def poll(self) -> bool:
+        ''' Poll the queue for a response.
+        Returns:
+            True if a response is received, False otherwise.
+        '''
+        if self._is_shutdown:
+            return False
```
syuoni, line 325: If `self._is_shutdown` is `True`, is Line 303 an endless loop?

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -302,6 +342,9 @@ def shutdown(self, wait=True):
         finally:
             self._is_shutdown = True
 
+    def abort(self):
+        self.shutdown()
```
syuoni, line 348: Why do we need `abort` given we already have `shutdown`? Is there any difference from the usage perspective?

tensorrt_llm/llmapi/mpi_session.py
Superjomn: Good catch! thanks.

tensorrt_llm/llmapi/mpi_session.py
Superjomn: Updated 303 as above, it can break the endless loop now.

tensorrt_llm/llmapi/mpi_session.py
Superjomn: I think it is just an [abstractmethod](https://github.com/superjomn/TensorRT-LLM/blob/b9b1c1368ce19bc6aec931710839f3fcdde4017b/tensorrt_llm/llmapi/mpi_session.py#L98) of the parent MpiSession, and the different child has different implementation, such as [here](https://github.com/superjomn/TensorRT-LLM/blob/b9b1c1368ce19bc6aec931710839f3fcdde4017b/tensorrt_llm/llmapi/mpi_session.py#L158-L159). This is a bit mess, and need a refine later.

tensorrt_llm/llmapi/llm_utils.py
Superjomn: Removed the import.

tensorrt_llm/llmapi/mpi_session.py
Superjomn: Updated, thanks.

tensorrt_llm/llmapi/mpi_session.py
Superjomn: Good suggestion, updated the code.

syuoni (approved):
LGTM, thanks!

Superjomn:
Switch to a new [PR](https://github.com/NVIDIA/TensorRT-LLM/pull/3825) targetting the main branch.

PR #3485, opened 2025-04-11
fix: don't perform memory estimation for star_attention

HuiGao-NV:
Mem estimation for star attention is not supportted now. So need to skip memory estimation for start_attention.


yuxianq (approved):

yuxianq:
@HuiGao-NV Please unwaive the tests in https://github.com/NVIDIA/TensorRT-LLM/pull/3464.

PR #3489, opened 2025-04-11
fix: Allow context_and_generation request type in disagg overlap

Tabrizian:


tensorrt_llm/executor/worker.py
```diff
@@ -348,9 +348,7 @@ def _enqueue_request(self, request: GenerationRequest) -> int:
         if is_overlap_enabled:
             is_disaggregated = self.engine.kv_cache_transceiver is not None
             if is_disaggregated and (
-                    request_type == tllm.RequestType.REQUEST_TYPE_CONTEXT_ONLY
-                    or request_type
-                    == tllm.RequestType.REQUEST_TYPE_CONTEXT_AND_GENERATION):
+                    request_type == tllm.RequestType.REQUEST_TYPE_CONTEXT_ONLY):
```
pcastonguay, line 351: We should add a test where we send a request to the disaggregated server, and then send a request to the generation worker directly and verify we get meaningful output.

pcastonguay (approved):
The test can come in another MR if Tanmay needs this fix ASAP.

tensorrt_llm/executor/worker.py
Tabrizian: I'll add testing in follow up PR.

PR #3491, opened 2025-04-11
fix: llama4: address couple of issues in llama4 attention module

chang-l:
This PR resolves: (1) adding attention temperature tuning, (2) unfused and reordered rope with qk_norm, and (3) normalize qk_norm against `head_dim`.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_llama.py
+++ b/tensorrt_llm/_torch/models/modeling_llama.py
@@ -24,7 +24,7 @@ from ..modules.gated_mlp import GatedMLP
 from ..modules.linear import (Linear, TensorParallelMode, WeightMode,
                               WeightsLoadingConfig)
 from ..modules.multi_stream_utils import maybe_execute_in_parallel
-from ..modules.rms_norm import L2Norm, RMSNorm
+from ..modules.rms_norm import RMSNorm
 from ..modules.rotary_embedding import RotaryEmbedding
 from ..speculative import Eagle3SpecMetadata, SpecMetadata
 from .modeling_utils import (DecoderModel, DecoderModelForCausalLM,
@@ -80,8 +80,10 @@ class Llama4Attention(Attention):
 
         if self.use_qk_norm:
             self.head_dim = config.hidden_size // config.num_attention_heads
-            self.qk_norm = L2Norm(hidden_size=self.head_dim,
-                                  dtype=config.torch_dtype)
+            self.qk_norm = RMSNorm(hidden_size=self.head_dim,
+                                   eps=1e-6,
+                                   dtype=config.torch_dtype,
+                                   has_weights=False)
         else:
             self.qk_norm = None
 
--- a/tensorrt_llm/_torch/models/modeling_llama.py
+++ b/tensorrt_llm/_torch/models/modeling_llama.py
@@ -10,8 +10,10 @@ from tensorrt_llm._torch.distributed import (AllReduce, AllReduceFusionOp,
 from tensorrt_llm._torch.pipeline_interface import PipelineInterface
 from tensorrt_llm.functional import PositionEmbeddingType
 
-from ..attention_backend import AttentionMetadata
-from ..attention_backend.interface import PositionalEmbeddingParams, RopeParams
+# For Llama4 attention backend
+from ..attention_backend import AttentionMetadata, TrtllmAttention
+from ..attention_backend.interface import (PositionalEmbeddingParams,
+                                           PredefinedAttentionMask, RopeParams)
 from ..model_config import ModelConfig
 from ..modules.attention import Attention
 from ..modules.decoder_layer import DecoderLayer
@@ -21,7 +23,7 @@ from ..modules.fused_moe import (FusedMoE, Llama4RenormalizeMoeRoutingMethod,
 from ..modules.gated_mlp import GatedMLP
 from ..modules.linear import (Linear, TensorParallelMode, WeightMode,
                               WeightsLoadingConfig)
-from ..modules.rms_norm import RMSNorm
+from ..modules.rms_norm import L2Norm, RMSNorm
 from ..speculative import Eagle3SpecMetadata, SpecMetadata
 from .modeling_utils import (DecoderModel, DecoderModelForCausalLM,
                              EagerFusionConfig, MissingLayer,
@@ -36,18 +38,15 @@ class Llama4Attention(Attention):
         model_config: ModelConfig[LlamaConfig],
         layer_idx: Optional[int] = None,
         use_qk_norm: bool = False,
+        nope_layer: bool = False,
+        attn_temperature_tuning: bool = True,
         aux_stream: Optional[torch.cuda.Stream] = None,
     ):
         config = model_config.pretrained_config
 
-        # Note the convention no_rope_layers[layer_idx] == 0 means nope_layer
-        is_nope_layer = config.no_rope_layers[layer_idx] == 0
-
-        use_rope = not is_nope_layer
-        attn_temperature_tuning = is_nope_layer and config.attn_temperature_tuning > 0
-        use_qk_norm = use_qk_norm and not is_nope_layer
-
-        if use_rope and not use_qk_norm:
+        self.use_rope = not nope_layer
+        self.use_qk_norm = use_qk_norm and not nope_layer
+        if self.use_rope and not self.use_qk_norm:
             # We can fuse pos embed only when: is_rope=True, is_qk_norm=False
             pos_embd_params = PositionalEmbeddingParams(
                 type=PositionEmbeddingType.rope_gptj,
@@ -56,7 +55,6 @@ class Llama4Attention(Attention):
             )
         else:
             pos_embd_params = None
-
         super().__init__(
             hidden_size=config.hidden_size,
             num_attention_heads=config.num_attention_heads,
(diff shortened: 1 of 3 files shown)


tensorrt_llm/_torch/modules/attention.py
```diff
@@ -19,13 +19,17 @@
 
 class L2Norm(torch.nn.Module):
```
hlu1: Move to `rms_norm.py` file

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -53,6 +57,7 @@ def __init__(
         config: Optional[ModelConfig] = None,
         use_qk_norm: bool = False,
         aux_stream: Optional[torch.cuda.Stream] = None,
+        attn_temperature_tuning: bool = False,
```
hlu1: Please create a new class `Llama4Attention` in `modeling_llama.py` and move all the new changes there. Remove the changes from the current `Attention` module. We should remove all the llama4 specific logic from this class in follow-up PR.

cc @yuxianq , what do you think?

hlu1 (changes requested):

tensorrt_llm/_torch/modules/attention.py
yuxianq: I agree. There is already a `LlamaAttention` (subclass of `Attention`) in `tensorrt_llm/_torch/models/modeling_llama.py`, we can remove attn_temperature_tuning and qk_norm from `Attention` and override the forward function in `LlamaAttention`. If llama4 changes the `Attention` module too much, using a new class is better (like `MLA`).

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -300,7 +304,7 @@ def forward(
 
         # Self Attention
         # For NOPE layers (like in Llama4), the position_ids needs to be set to None, so the rotary embedding will not be applied.
-        if self.is_nope_layer:
+        if self.is_nope_layer and not self.self_attn.attn_temperature_tuning:
```
syuoni: I remember No-RoPE layers should be consistent with enabling attention temperature scaling. See [here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama4/modeling_llama4.py#L334). This line looks like they are not consistent. Am I missing something?

chang-l: Right, but to apply the attn scaling, we need to pass `position_ids` even if it's a no-rope layer.

tensorrt_llm/_torch/modules/attention.py
chang-l: Updated.

tensorrt_llm/_torch/modules/attention.py
chang-l: Thanks for the suggestion. I separated Llama4Attention forward method out from Attention module.

chang-l:
I've updated this PR to address this round of comments. With this PR, Scout achieves MMLU score of `80.53`, and Maverick reaches `86.0`.
@hlu1 @yuxianq @syuoni

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -54,7 +36,6 @@ def __init__(
         dtype: torch.dtype = None,
         dense_bias: Optional[bool] = None,
         config: Optional[ModelConfig] = None,
-        use_qk_norm: bool = False,
         aux_stream: Optional[torch.cuda.Stream] = None,
```
yuxianq: `aux_stream` can also be removed.

tensorrt_llm/_torch/modules/rms_norm.py
```diff
@@ -47,3 +47,25 @@ def forward(
...
+                 hidden_size: int,
+                 dtype: Optional[torch.dtype] = None,
+                 eps: float = 1e-6):
+        super().__init__()
+        self.eps = eps
+        self.dtype = dtype
+        self.hidden_size = hidden_size
+
+    def _norm(self, x):
+        flash_infer_kernel = RMSNorm(hidden_size=self.hidden_size,
```
yuxianq: Can we move this line to `__init__` as `self.norm = RMSNorm(...)`? Where does `RMSNorm`'s weight come from if it is not in `__init__`?

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -53,36 +55,163 @@ def __init__(
...
+        attention_mask: PredefinedAttentionMask = PredefinedAttentionMask.
+        CAUSAL,
+        mrope_config: Optional[dict] = None,
+        all_reduce_params: Optional[AllReduceParams] = None,
+        lora_params: Optional[dict] = None,
+        **kwargs,
+    ) -> torch.Tensor:
+        qkv = self.qkv_proj(hidden_states)
+        is_fused_qkv = False
+        if isinstance(self.attn, TrtllmAttention):
```
yuxianq: We can check `self.attn_backend` instead of `self.attn` to avoid importing `TrtllmAttention`.

tensorrt_llm/_torch/modules/rms_norm.py
chang-l: It has no weights and does pure L2 norm against hidden_dim.

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -53,36 +55,163 @@ def __init__(
...
+        if self.attn_temperature_tuning:
+            # this must be a nope layer
+            assert position_ids is not None, "attn_temperature_tuning requires position_ids"
+            q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size],
+                                dim=-1)
+            attn_scale = self._get_attn_scale(position_ids)
+            q = (q * attn_scale).to(q.dtype)
+            qkv = torch.concat(
+                [q.contiguous(), k.contiguous(),
+                 v.contiguous()], dim=-1)
```
hlu1: Remove all `contiguous` call. `concat` supports strided inputs

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -53,36 +55,163 @@ def __init__(
...
+        if lora_params is not None:
+            qkv_lora = self.splitted_qkv_lora(hidden_states, lora_params,
+                                              self.layer_idx)
+            if qkv_lora is not None:
+                qkv = qkv + qkv_lora
+
+            qkv_lora = self.fused_qkv_lora(hidden_states, lora_params,
+                                           self.layer_idx)
+            if qkv_lora is not None:
+                qkv = qkv + qkv_lora
```
hlu1: Let's remove these for now. I'm not even sure if they're inserted at the right place.

tensorrt_llm/_torch/modules/rms_norm.py
yuxianq: We call `flash_infer_kernel(x)`, which will call `RMSNorm`'s forward. `RMSNorm`'s forward uses its weight, so I think the weight should be initialized. How do we perform L2 norm with `RMSNorm`'s forward?

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -53,36 +55,163 @@ def __init__(
...
+                                            attention_mask=attention_mask,
+                                            mrope_config=mrope_config)
+
+        attn_output = self.o_proj(attn_output,
+                                  all_reduce_params=all_reduce_params)
+        if lora_params is not None:
+            attn_lora_output = self.o_lora(attn_output, lora_params,
+                                           self.layer_idx)
+            if attn_lora_output is not None:
+                attn_output = attn_output + attn_lora_output
```
hlu1: Remove

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -53,36 +55,163 @@ def __init__(
...
+                                              self.layer_idx)
+            if qkv_lora is not None:
+                qkv = qkv + qkv_lora
+
+            qkv_lora = self.fused_qkv_lora(hidden_states, lora_params,
+                                           self.layer_idx)
+            if qkv_lora is not None:
+                qkv = qkv + qkv_lora
+
+        if is_fused_qkv:
```
hlu1: This code path is not hit right? I think you should have two forward function, one for is_fused_qkv = True, and one for is_fused_qkv = False, and clean up the several concat/split. The code is way too complicated right now.

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -53,36 +55,163 @@ def __init__(
...
+
+                q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size],
+                                    dim=-1)
+                q, k = self.rotary_emb(position_ids, [q, k], attn_metadata)
+                qkv = torch.concat([q, k, v], dim=-1)
+
+            # TODO: make this more efficient.
+            q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size],
+                                dim=-1)
+            do_multi_stream = torch.cuda.is_current_stream_capturing(
```
hlu1: Please apply the multi-stream API introduced in https://github.com/NVIDIA/TensorRT-LLM/pull/3459 here

hlu1 (changes requested):

tensorrt_llm/_torch/modules/rms_norm.py
chang-l: Pure L2Norm is performed as if RMSNorm's weights are all ones (basically disable scaling at the end) (see [here](https://github.com/NVIDIA/TensorRT-LLM/blob/efabf6b44374ccad29d8de923c351870d87a0c78/tensorrt_llm/_torch/modules/rms_norm.py#L19)), unlike `input_layernorm`/`post_attention_layernorm`, whose weights are loaded from ckpt.

chang-l: I can see this is a WAR and not ideal (we re-use RMSNorm module for L2Norm)... Maybe we should have a similar flag `has_weight=False` for RMSNorm, as vllm does. I can add this in another PR.

tensorrt_llm/_torch/models/modeling_llama.py
chang-l: You mean `is_fused_qkv`? I think we will hit this branch when we use attn_backend=TRTLLM and qk_norm is disabled (e.g., Maverick).

chang-l: There are many possible paths for a llama4 attention layer depending on configs/backends — e.g., `qk_norm enabled/disabled`, `rope layer` or `nope layer`, and `fused_qkv` vs. `unfused_qkv`. I’ll work on refactoring the forward call to make it more clear

tensorrt_llm/_torch/modules/rms_norm.py
yuxianq: I see. How about making L2Norm to inherit from RMSNorm and override its self.weight to a constant `torch.ones(...)` instead of `nn.Parameter` to avoid accidently loading weight from checkpoint, like:

```
class L2Norm(RMSNorm):

    def __init__(self,
                 *,
                 hidden_size: int,
                 dtype: Optional[torch.dtype] = None,
                 eps: float = 1e-6):
        super().__init__(hidden_size=hidden_size, dtype=dtype, eps=eps)
        # L2Norm is performed as if RMSNorm's weights are all ones
        # Override weight since it is constant now
        self.weight = torch.ones(hidden_size, dtype=dtype, device=device)

    def forward(self, x):
        return super().forward(x).type_as(x)

```

tensorrt_llm/_torch/models/modeling_llama.py
yuxianq: The attention module's forward has been simplified, see https://github.com/NVIDIA/TensorRT-LLM/pull/3517/files#diff-94d88f20ca4a510ea0879428279cc50121223458ccae6090ceba391d1e32a371R180. We create several flags `enable_rope_fusion/support_fused_qkv/support_unfused_qkv` in `__init__` to help conditionally execution in forward. There will be a follow-up PR to make these flags a part of the attention backend, so that we don't need to detect attention backend manually in the attention module.

tensorrt_llm/_torch/modules/attention.py
chang-l: Removed

tensorrt_llm/_torch/models/modeling_llama.py
chang-l: Updated

tensorrt_llm/_torch/models/modeling_llama.py
chang-l: @hlu1 @yuxianq  I have updated the llama4 attention part to accommodate the latest attention module refactor.

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -52,20 +59,148 @@ def __init__(
...
+        attention_mask: PredefinedAttentionMask = PredefinedAttentionMask.
+        CAUSAL,
+        mrope_config: Optional[dict] = None,
+        all_reduce_params: Optional[AllReduceParams] = None,
+    ):
+        if self.use_qk_norm:
+            qkv = self.qkv_proj(hidden_states)
+            q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size],
+                                dim=-1)
+            assert self.rotary_emb is not None and not self.enable_rope_fusion, "qk_norm requires rope fusion to be disabled"
```
hlu1:
```suggestion
            assert self.rotary_emb is not None and not self.enable_rope_fusion, "qk_norm requires attention rope fusion disabled"
```

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -52,20 +59,148 @@ def __init__(
...
+                                  dtype=config.torch_dtype)
+        else:
+            self.qk_norm = None
+
+        self.attn_temperature_tuning = attn_temperature_tuning
+        self.floor_scale = getattr(config, "floor_scale", 8192.0)
+        self.attn_scale = getattr(config, "attn_scale", 0.1)
+
+    def _attn_qkv(self, q, k, v, attn_metadata, attention_mask, mrope_config,
+                  all_reduce_params):
```
hlu1: Add typing informations to inputs

hlu1 (approved):
Looks much better now. Thanks for the great work!

tensorrt_llm/_torch/modules/rms_norm.py
chang-l: This actually is not allowed (assign torch.Tensor to nn.Parameters)...
```TypeError: cannot assign 'torch.BFloat16Tensor' as parameter 'weight' (torch.nn.Parameter or None expected)```

I have to update to add `has_weights` in RMSNorm for pure L2Norm calculation

File: tensorrt_llm/_torch/models/modeling_llama.py (hunk 13 above)
Comment (chang-l) [reply tensorrt_llm/_torch/models/modeling_llama.py]: Thanks. 
I added the typing info there.

File: tensorrt_llm/_torch/modules/rms_norm.py (hunk 5 above)
Comment (yuxianq) [reply tensorrt_llm/_torch/modules/rms_norm.py]: Sure.

File: tensorrt_llm/_torch/models/modeling_llama.py [hunk 14]
```diff
@@ -10,8 +10,10 @@
 from tensorrt_llm._torch.pipeline_interface import PipelineInterface
 from tensorrt_llm.functional import PositionEmbeddingType
 
+# For Llama4 attention backend
```
yuxianq: Can remove this line.

yuxianq (approved):

PR #3492, opened 2025-04-11
fix: llama4: add an option `apply_router_weight_on_input` for in FusedMoE

chang-l:
This PR is to adjust MoE for LLaMA4 to handle applying routing weights upfront.


tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -565,6 +569,12 @@ def forward_chunk(
         assert token_final_scales.dtype == torch.float32
         assert token_selected_experts.dtype == torch.int32
 
+        if self.apply_router_weight_on_input:
+            assert self.routing_method.top_k == 1, "Current walkaround only supports top-1 routing"
+            x = x * token_final_scales.to(x.dtype)
+            # TODO: remove this once we have correct fusedmoe kernel ready
+            token_final_scales = torch.ones_like(token_final_scales)
```
hlu1:
```suggestion
            token_final_scales = None
```
Can you do a quick mmlu run to verify this?

tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -565,6 +569,12 @@ def forward_chunk(
         assert token_final_scales.dtype == torch.float32
         assert token_selected_experts.dtype == torch.int32
 
+        if self.apply_router_weight_on_input:
+            assert self.routing_method.top_k == 1, "Current walkaround only supports top-1 routing"
+            x = x * token_final_scales.to(x.dtype)
```
hlu1, line 567:
```suggestion
            x = x * token_final_scales
```

tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -307,6 +308,9 @@ def __init__(
         if not model_config.skip_create_weights:
             self.create_weights()
 
+        # If True, the router weight will be multiplied on the input rather than at the end of FC2
+        self.apply_router_weight_on_input = apply_router_weight_on_input
```
syuoni, line 307: Just out of curiosity, what is the difference between applying routing scales to inputs and outputs? They should be mathematically equivalent. Is there any numerical or perf difference?

chang-l: It's actually not mathematically equivalent due to GatedMLP.

syuoni: Oh I see. Thanks.

tensorrt_llm/_torch/modules/fused_moe.py
chang-l: It looks like type promotion would promote the input (`x`) to float, matching the dtype of `token_final_scales`, which is undesirable for our moe runner.

hlu1 (approved):

PR #3494, opened 2025-04-12
fix: Correct reporting of text dtype for Llama 4.

FrankD412:
This is a quick fix that corrects a configuration file format change for Llama 4.

*NOTE:* This fix is only meant to be an unblocking fix. I will submit a PR in the future to correct this reporting.

### DIFF
--- a/tensorrt_llm/bench/dataclasses/reporting.py
+++ b/tensorrt_llm/bench/dataclasses/reporting.py
@@ -264,7 +264,9 @@ class ReportUtility:
                 "backend":
                 "Pytorch",
                 "dtype":
-                torch_dtype_to_str(model_config.pretrained_config.torch_dtype),
+                torch_dtype_to_str(
+                    model_config.pretrained_config.torch_dtype
+                    or model_config.pretrained_config.text_config.torch_dtype),
                 "kv_cache_dtype":
                 model_config.quant_config.kv_cache_quant_algo,
                 "quantization":


kaiyux (approved):
LGTM, thanks.

tensorrt_llm/bench/dataclasses/reporting.py
```diff
@@ -264,7 +264,9 @@ def get_statistics_dict(self) -> Dict[str, Any]:
                 "backend":
                 "Pytorch",
                 "dtype":
-                torch_dtype_to_str(model_config.pretrained_config.torch_dtype),
+                torch_dtype_to_str(
+                    model_config.pretrained_config.torch_dtype
```
kaiyux, line 268: Sorry, had a question after double checking the format - on which checkpoint did you see that `model_config.pretrained_config.torch_dtype` is `None`?

FrankD412: It was Llama 4 Scout/Maverick. See [here](https://huggingface.co/meta-llama/Llama-4-Scout-17B-16E-Instruct/blob/main/config.json#L9)

But the general point is that they have a vision and text component so the config reflects a `text_config` and `vision_config`. Because the Pytorch `ModelConfig` class carries the raw pretrained config that's reflected in the file, you have to know exactly what sub-key the values are in.

PR #3502, opened 2025-04-13
refactor: Remove enforced sorted order of batch slots

Funatiq:
Change the order of decoder `batchSlots` by removing the sort [here](https://github.com/NVIDIA/TensorRT-LLM/pull/3502/files#r2041689369). 
This means that the order passed to the decoder is exactly the same as `seqSlots`. The only difference is that `seqSlots` contain all requests and `batchSlots` contain only finished context and generation requests.

This change simplifies the code and removes some extra steps and buffers. However, it introduces a sync after the decoder setup in the Eagle/ReDrafter/Lookahead case for correctness [here](https://github.com/NVIDIA/TensorRT-LLM/pull/3502#discussion_r2192403292). Without this the Eagle and ReDrafter tests might fail spuriously (so there is likely a race condition).

### DIFF
--- a/cpp/tensorrt_llm/batch_manager/makeDecodingBatchInputOutput.cpp
+++ b/cpp/tensorrt_llm/batch_manager/makeDecodingBatchInputOutput.cpp
@@ -92,27 +92,18 @@ namespace
 std::pair<std::vector<SizeType32>, std::vector<SizeType32>> getActiveSlots(
     RequestVector const& contextRequests, RequestVector const& generationRequests)
 {
-    std::vector<std::pair<SizeType32, SizeType32>> slots;
+    std::vector<SizeType32> activeSlots;
+    std::vector<SizeType32> generationSteps;
     for (auto const& requests : {contextRequests, generationRequests})
     {
         for (auto const& llmReq : requests)
         {
             if (llmReq->isGenerationInProgressState() || llmReq->isLastContextChunk())
             {
-                slots.push_back({llmReq->mSeqSlot.value(), llmReq->getDecodingIter()});
+                activeSlots.push_back(llmReq->mSeqSlot.value());
+                generationSteps.push_back(llmReq->getDecodingIter());
             }
         }
-    }
-
-    std::sort(slots.begin(), slots.end(),
-        [](std::pair<SizeType32, SizeType32> const& a, std::pair<SizeType32, SizeType32> const& b)
-        { return a.first < b.first; });
-
-    std::vector<SizeType32> activeSlots, generationSteps;
-    for (auto const& slot : slots)
-    {
-        activeSlots.push_back(slot.first);
-        generationSteps.push_back(slot.second);
     }
 
     return {activeSlots, generationSteps};
--- a/cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
+++ b/cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
@@ -983,8 +983,10 @@ void TrtGptModelInflightBatching::forwardAsync(RequestList const& activeRequests
         if (fittingRequests.empty() && fittingDisaggGenInitRequests.empty())
         {
             TLLM_LOG_WARNING(
-                "CapacityScheduler didn't schedule any requests, probably because of insufficient resources such as KV "
-                "cache, will try wait for KV cache transfer to complete");
+                "CapacityScheduler didn't schedule any requests in iteration %lu, "
+                "probably because of insufficient resources such as KV cache, "
+                "will try wait for KV cache transfer to complete",
+                mIterCounter);
             if (mCacheTransceiver)
             {
                 mCacheTransceiver->checkContextTransferStatus(1);
--- a/cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
+++ b/cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
@@ -2433,7 +2433,7 @@ void TrtGptModelInflightBatching::rewindKVCacheBlocks(SizeType32 numSequences)
         *mDecoderState->getAcceptedLengthsCumSum(), *mDecoderState->getAcceptedPackedPaths(),
         *runtimeBuffers.sequenceLengthsDevice, pointerArrayPtr, offsetArrayPtr, localNbLayers, numSequences,
         mRewindInputs.numKvHeads, sizeInBytesPerKVHead, commonRewindLen, rewindLens,
-        *runtimeBuffers.seqSlotRemappingDevice, *runtimeBuffers.sortedSeqSlots, getMaxAttentionWindow(),
+        *runtimeBuffers.seqSlotRemappingDevice, *runtimeBuffers.seqSlots, getMaxAttentionWindow(),
         mRewindInputs.maxBlocksPerSeq, tokensPerBlock, mRewindInputs.isUseOneMoreBlock,
         mRuntime->getStreamPtr()->get());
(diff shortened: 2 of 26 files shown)


cpp/tensorrt_llm/runtime/runtimeKernels.cu
```diff
@@ -1246,16 +1246,14 @@ void invokeUpdateKVBlockArrayDraftTokenLocation(ITensor const& seqAcceptedDraftT
...
+    SizeType32 maxBlocksPerSeq, SizeType32 tokensPerBlock, bool canUseOneMoreBlock, cudaStream_t stream)
 {
     tensorrt_llm::kernels::speculative_decoding::updateKVBlockArrayDraftTokenLocation(
         bufferCast<SizeType32>(seqAcceptedDraftTokenOffsets), bufferCast<SizeType32>(packedAcceptedDraftTokensIndices),
         bufferCast<SizeType32>(pastKeyValueLengths), pointerArray, offsetArray, layerCount, seqCount, numKVHeads,
-        sizeInBytesPerKVHead, rewindDraftTokenCommonCount, rewindDraftTokenSeparateAdjustments,
-        bufferCast<SizeType32>(seqSlotRemapping), bufferCast<SizeType32>(batchSlots), maxKVCacheLen, maxBlocksPerSeq,
-        tokensPerBlock, canUseOneMoreBlock, stream);
+        sizeInBytesPerKVHead, rewindDraftTokenCommonCount, rewindDraftTokenSeparateAdjustments, nullptr,
+        bufferCast<SizeType32>(batchSlots), maxKVCacheLen, maxBlocksPerSeq, tokensPerBlock, canUseOneMoreBlock, stream);
```
Funatiq, line 482: It would be possible to remove the `seqSlotRemapping` parameter in the entire call stack.

cpp/tensorrt_llm/kernels/speculativeDecoding/common.cu
```diff
@@ -81,22 +81,7 @@ __global__ void packAcceptedPaths(SizeType32* acceptedLengthsCumSum, SizeType32*
...
-            {
-                auto const seqSlot = seqSlots[bi];
-                if (batchSlot == seqSlot)
-                {
-                    pathBatchIdx = bi;
-                    break;
-                }
-            }
-        }
+        auto const pathBatchIdx = isPathsSeqSlotIdx ? bi : batchSlot;
```
nekorobov, line 84: I think this is wrong. Let me think out loud and correct me if I'm wrong.

Let's say `batchSlots = [0, 2, 4]` -- the order passed to the decoder
`seqSlots = [4, 2, 0]` -- the order passed to the engine

then for `batchSlot = 0` and `bi = 0`, the path, produced by the engine is located at index `2`.

cpp/tensorrt_llm/runtime/runtimeKernels.cu
nekorobov: I think if we delete this, the logic with [this code ](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/kernels/speculativeDecoding/kvCacheUpdateKernels.cu#L58) becomes wrong. We must access `pastKeyValueLengths`, which comes from `runtimeBuffers.sequenceLengthsDevice` according to the sorted `seqSlot` to sync the accesses between `runtimeBuffer` structs and `decoderBuffer` structs.

cpp/tensorrt_llm/kernels/speculativeDecoding/common.cu
Funatiq: The change in this PR is that the order passed to the decoder is exactly the same as `seqSlots`. The only difference is that `seqSlots` contain *all* requests and `batchSlots` contain only finished context and generation requests.

cpp/tensorrt_llm/layers/eagleDecodingLayer.cpp
```diff
@@ -278,27 +276,24 @@ void EagleDecodingLayer<T>::packAcceptedPaths(EagleOutputs const& outputs, Eagle
...
-    auto numNewTokensCumSum = bufferCast<SizeType32>(*outputs.numNewTokensCumSum);
-    auto pathsOffsets = bufferCast<SizeType32>(*outputs.pathsOffsets);
-    auto batchSlots = augmentedBatchSlots;
-    auto seqSlots = augmentedSeqSlots;
-    auto bestPathIndicesSlotsPtr = bufferCast<SizeType32>(*inputs.acceptedPathIds);
-    auto lastDraftPathsSlotsPtr = bufferCast<SizeType32>(*inputs.lastDraftPaths);
+    auto const* numNewTokens = bufferCast<SizeType32>(*outputs.numNewTokens.value());
+    auto* numNewTokensCumSum = bufferCast<SizeType32>(*outputs.numNewTokensCumSum);
+    auto* pathsOffsets = bufferCast<SizeType32>(*outputs.pathsOffsets);
+    auto const* batchSlots = augmentedSeqSlots;
```
Funatiq, line 288: We pass `augmentedSeqSlots` which means `seqSlots` with `-1` for skipped chunked requests.

cpp/tensorrt_llm/kernels/speculativeDecoding/common.cu
nekorobov: Hm, sorry, I do not see the change related to that. Where do you change the batch slots order in `gptDecoderBatched`?

Funatiq: In this kernel we use `augmentedSeqSlots` which means `seqSlots` with -1 for skipped chunked requests.

cpp/tensorrt_llm/batch_manager/makeDecodingBatchInputOutput.cpp
```diff
@@ -107,9 +107,6 @@ std::vector<SizeType32> getActiveSlots(RequestVector const& contextRequests, Req
             }
         }
     }
-
-    std::sort(activeSlots.begin(), activeSlots.end());
```
Funatiq: Remove the sorting of decoder `batchSlots` here.

cpp/tensorrt_llm/kernels/speculativeDecoding/common.cu
Funatiq: Sorry, the decoder code has changed quite a lot. The change is subtle here: https://github.com/NVIDIA/TensorRT-LLM/pull/3502/files#r2041689369

nekorobov: Understood now, thanks!

nekorobov (approved):

cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
```diff
@@ -1019,6 +1021,10 @@ void TrtGptModelInflightBatching::forwardAsync(RequestList const& activeRequests
                 auto const contextBufferId = mCtxGenFusion ? getFusedBufferId() : getContextBufferId();
                 setupDecoderStep(currRequests.contextRequests, *mBuffers.at(contextBufferId),
                     mDecoderInputBuffers.at(getFusedBufferId()));
+                // WAR: Sync to ensure that the decoder setup is complete before the context phase starts.
+                // Without this, there may be a race condition between the decoder setup and the context phase
+                // which also leads to spurious test failure in trtGptModelRealDecoderTest.
+                mRuntime->getStream().synchronize();
```
Funatiq, line 1027: Not this WAR.

PR #3506, opened 2025-04-13
refactor: Introduce DecoderOutputBuffers per batch

Funatiq:
- Move communication logic from buffer classes to communication classes.
  - This allows to send various buffers instead of being limited to the buffer class.
- Move (host) decoder outputs from DecoderBuffers to new DecoderOutputBuffers class


dcampora:
Looks good to me. Great news that overlap scheduling finally works with the C++ decoders!

dcampora (approved):

PR #3517, opened 2025-04-14
feat: Support cos_sin_cache in all cases.

yuxianq:
This MR contains the following updates:

1. Handle `fuse_pos_embd=True/False` and create `RotaryEmbedding` inside attention module, so that the users don't need to handle it in the modeling files.
2. Cache `cos_sin` for unfused rope implementation. If flashinfer is available, use `apply_rope_with_cos_sin_cache_inplace` instead of `apply_rope_inplace`. Otherwise, we fallback to pure pytorch implementation, which can support any rope now.


litaotju (approved):

tensorrt_llm/_torch/modules/rotary_embedding.py
```diff
@@ -1,213 +1,62 @@
...
-                    rope_theta = self.config.rope_theta
-                    if self.rope_type in ['default', 'llama3']:
-                        flashinfer_apply_rope_inplace(
-                            q,
-                            k,
-                            attn_metadata.qo_indptr,
-                            attn_metadata.cached_token_lens,
-                            rope_theta=rope_theta,
-                            rotary_dim=self.rotary_dim,
-                            interleave=use_gptj_style_rope)
```
chang-l, line 159: Just to clarify — here when `interleave=True` (for llama4), can we still use `flashinfer_apply_rope_with_cos_sin_cache_inplace`?
It looks like in L25, you enforced `interleave=False`. Are these referring to the same interleave definition? @yuxianq

yuxianq: llama4 uses `is_neox=False` while llama3 uses `is_neox=True`. `interleave=True/False` is for `rope_cos_sin` layout. For fused rope in trtllm attention backend, we always use `interleave=True`. For unfused rope implemented by flashinfer or native PyTorch ops, we always use `interleave=False`.

chang-l: Thanks @yuxianq.
Can I understand it as: whether interleave is True or not would not impact the attention output itself — as long as we follow the expected format per backend (i.e., use interleaved format for TRTLLM and non-interleaved for FlashInfer)?

Also, for my own education, what does `is_neox` represent for?

vanshilshah97: Hi @yuxianq 
Thanks for the commit. I think i was looking for this for our use case.
I was discussing this with @chang-l today afternoon regarding this.

@chang-l I think `is_neox` refers to the type of RoPE embeddings that TRT-LLM supports that were found in the respective  [model implementation](https://nvidia.github.io/TensorRT-LLM/advanced/gpt-attention.html#rotary-positional-embedding-rope).
@yuxianq one question
 For our use-case to support trt-llm attention for llama4 scout . We want to apply qk_norm on Q and K after RoPE is applied.
I was planning to use the pytorch implementation for generating unfused rope then apply qk norm on pytorch level and pass these things to trtllm attention kernels thorugh [here](https://github.com/NVIDIA/TensorRT-LLM/blob/b2fb0fe843ca6ed09be691b9219747da4167d3c2/tensorrt_llm/_torch/attention_backend/trtllm.py#L315) with `pos_embed_params=None`, according to my understanding as we have already done RoPE. Do you forsee any issue for that ? Accuracy wise or format wise ?

chang-l: @vanshilshah97  It seems w/ this PR, unfused rope (using flashinfer kernel) can work outside together with trtllm attention backend. I will update the code path in this PR: https://github.com/NVIDIA/TensorRT-LLM/pull/3491 later tonight, then you can cherry-pick it to your llama4 branch.

yuxianq: @vanshilshah97 The main issue is that the current MLA kernel cannot skip rope, see https://github.com/NVIDIA/TensorRT-LLM/blob/515dd0d78fe8f5f86f8a655fc7bac36a6f8da7ac/cpp/tensorrt_llm/kernels/mlaKernels.cu#L451.
@hlu1 Can you help to estimate the effort needed to skip rope in our mla kernels?

yuxianq: @chang-l `interleave=True` means the `rope_cos_sin` layout is [cos0, sin0, cos1, sin1, ...], cos and sin are interleaved; `interleave=False` uses [cos0, cos1, ..., sin0, sin1, ...] layout.
`is_neox` is about q/k layout. When calculating rope, we need to split q/k into two tensors and use them to multiple with cos/sin. `is_neox=True` means we split q/k into the first half part (`q[..., :head_dim//2]`) and the second half part (`q[..., head_dim//2:]`), while `is_neox=False` means we split q/k into the even part (`q[..., ::2]`) and the odd part (`q[..., 1::2]`).

chang-l: @yuxianq  Got it. Thanks Yuxian.  
So `interleave=True` must come with `is_neox=False`; and `interleave=False` must come with `is_neox=True`

If so, like you mentioned:
> For unfused rope implemented by flashinfer or native PyTorch ops, we always use interleave=False

then we cannot use the existing unfused rope from flashinfer or native pytorch op for llama4 due to format issue (interleave = True/is_neox=False for llama4)? Am i missing something?

yuxianq: No, `is_neox` is related to algorithm, it changes the rope output; `interleave` is related to implementation, it does not change the rope output. I list all the combinations:
- `is_neox=False` and `interleave=False`: for rope_gptj + flashinfer/native op
- `is_neox=False` and `interleave=True`: for rope_gptj + trtllm fused rope
- `is_neox=True` and `interleave=False`: for rope_gpt_neox + flashinfer/native op
- `is_neox=True` and `interleave=True`: for rope_gpt_neox + trtllm fused rope

yuxianq: @hlu1 Sorry for misunderstanding, this issue is related to llama4, not deepseek, so MLA without rope fusion is unnecessary here.

chang-l: @yuxianq  Thanks! Thats very helpful

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -208,36 +221,25 @@ def forward(
...
+            out_scale = self.o_proj.inv_input_scale
 
-            attn_output = self.attn.forward(q.contiguous(),
-                                            k.contiguous(),
-                                            v.contiguous(),
-                                            attn_metadata,
-                                            attention_mask=attention_mask,
-                                            mrope_config=mrope_config)
+        q, k, v = self.convert_qkv(q, k, v)
+        attn_output = self.attn.forward(q,
```
chang-l, line 236: @yuxianq  
Another question: do we need to include `.contiguous()` after q, k, v when passing to `attn.forward`  after applying rope

cc. @hlu1

hlu1: It depends on if the attention kernel requires contiguous inputs or strided inputs. A common practice is for the attention kernel to call `.contiguous()` inside if contiguous inputs are required.

chang-l: I see.. do you know whether flashinfer and trtllm attn backend support strided input? AFAIK, there seems no `contiguous()` call inside both attn kernels? (I may miss something)

PR #3520, opened 2025-04-14
test: Add llama 4 to ci

dongfengy:
(1) llama4 models are uploaded to llm_models_root()
(2) maverick and scout tests will be enabled on 8x h200 post merge

### DIFF
--- a/tests/integration/test_lists/test-db/l0_dgx_h200.yml
+++ b/tests/integration/test_lists/test-db/l0_dgx_h200.yml
@@ -15,5 +15,5 @@ l0_dgx_h200:
   tests:
   # ------------- PyTorch tests ---------------
   - accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_bfloat16_4gpus[ep4-attention_dp-cuda_graph-overlap_scheduler]
-  - unittest/_torch/multi_gpu_modeling/test_llama4.py::test_llama4[tp8-trtllm-scout]
-  - unittest/_torch/multi_gpu_modeling/test_llama4.py::test_llama4[tp8-trtllm-maverick]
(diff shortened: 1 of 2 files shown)


mikeiovine (approved):
LGTM assuming tests pass

dongfengy:
according to @yiqingy0  pipeline got time out because the test is running for more than 6 hours
this doesn't make sense as it's  just a single prompt. I will try to run it locally +CC @mikeiovine

tests/integration/test_lists/test-db/l0_dgx_h200.yml
```diff
@@ -15,3 +15,4 @@ l0_dgx_h200:
   tests:
   # ------------- PyTorch tests ---------------
   - accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_bfloat16_4gpus[ep4-attention_dp-cuda_graph-overlap_scheduler]
+  - unittest/_torch/multi_gpu_modeling -k "llama4"
```
litaotju: whats the runtime of these?
If its slow, lets only check TRT-LLM backend not flashinfer one.

litaotju (approved):

tests/integration/test_lists/test-db/l0_dgx_h200.yml
dongfengy: scout is 1h
maverick is 3h

Yes we will need to only check trtllm ones

dongfengy: as per discussion with Xiwen Yu , we need to add only scout first. Otherwise the entire stage will timeout. (Since it need to run other tests as well). +CC @mikeiovine

PR #3524, opened 2025-04-14
chore : Split more tests out of gpt tests

peaceh-nv:
Before this change on A30:
test_cpp.py::test_model[gpt-80] : 6951.72s
After this change on A30:
test_cpp.py::test_model[gpt-80] : 2343.66s
test_cpp.py::test_model[gpt_tests-80] : 1622.67s
test_cpp.py::test_model[gpt_session-80] : 1136.56s
Reduce test time on each single test but will add 30min in total test time on A30 test. 

And do not parallelize gpt* test to avoid OOM issues.

### DIFF
--- a/tests/integration/defs/cpp_common.py
+++ b/tests/integration/defs/cpp_common.py
@@ -72,6 +72,16 @@ def generate_result_file_name(test_list: List[str],
         yield "fp8"
 
 
+def generate_excluded_test_list(test_list):
+    if "gpt" in test_list:
+        if "gpt_session" not in test_list:
+            yield "gpt_session"
+        if "gpt_executor" not in test_list:
+            yield "gpt_executor"
+        if "gpt_tests" not in test_list:
+            yield "gpt_tests"
+
+
 def find_dir_containing(files: Sequence[str],
                         start_dir: Optional[_pl.Path] = None) -> _pl.Path:
     if start_dir is None:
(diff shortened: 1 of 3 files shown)


QiJune (approved):
LGTM

MartinMarciniszyn:
@DomBrown , could you please also take a quick look at this?

tests/integration/defs/cpp_common.py
```diff
@@ -719,6 +721,12 @@ def run_single_gpu_tests(build_dir: _pl.Path,
     if "gpt" in test_list and "gpt_executor" not in test_list:
         excluded_tests.append("GptExecutor")
 
+    if "gpt" in test_list and "gpt_session" not in test_list:
```
DomBrown: I would really prefer if we could do this using a generator function to build the lists as is done above, if possible. Otherwise I think we will build up a lot of "if x then append y" over time. I overlooked this last time for Gpt executor because it was just one test case, but I can see how this could grow now.

Try a generator, if that doesn't work maybe extract building this list into a function.

Otherwise, looks good to me. Thanks for increasing the granularity :)

DomBrown (changes requested):
See comment below. Generally looks good to me. Thanks :)

DomBrown (approved):

tests/integration/defs/cpp_common.py
```diff
@@ -726,7 +737,13 @@ def run_single_gpu_tests(build_dir: _pl.Path,
         if excluded_tests:
             ctest.extend(["-E", "|".join(excluded_tests)])
 
-        parallel = default_test_parallel
+        gpt_tests = {"gpt", "gpt_session", "gpt_tests", "gpt_executor"}
+
+        # gpt* tests are not parallelized as it would cause OOM because kv cache memory allocations
+        # exist in multiple running tests
+        if gpt_tests.intersection(test_list):
+            parallel = 1
```
DomBrown, line 745: Looks like you need an:
`
else:
    parallel = default_test_parallel
`
here to fix the pipeline

peaceh-nv: Fixed in the latests commit. Thx!

PR #3525, opened 2025-04-14
fix: Intercept the error of multi-ranks bound to a single device

Shixiaowei02:


cpp/tensorrt_llm/common/opUtils.cpp
```diff
@@ -68,11 +68,30 @@ ncclUniqueId getUniqueId(std::set<int> const& group) noexcept
...
+    return std::unordered_set<int>(deviceIds.begin(), deviceIds.end()).size() != deviceIds.size();
+}
+
 } // namespace
 
 std::shared_ptr<ncclComm_t> getComm(std::set<int> const& group)
 {
     auto const rank = COMM_SESSION.getRank();
+    TLLM_CHECK_WITH_INFO(
+        isDevicesDuplicate(COMM_SESSION), "Binding multiple ranks to a single CUDA device is not supported.");
```
yuxianq: We should check `LOCAL_COMM_SESSION` instead of `COMM_SESSION`, device IDs are only unique in the same node.

yuxianq: Can you also run a local test to validate that this fix works for multi-node cases?

yuxianq (approved):

Superjomn (approved):
LGTM

cpp/tensorrt_llm/common/opUtils.cpp
```diff
@@ -68,11 +68,24 @@ ncclUniqueId getUniqueId(std::set<int> const& group) noexcept
...
 }
+
+bool isDevicesUnique(tensorrt_llm::mpi::MpiComm const& comm)
+{
+    int deviceId;
+    TLLM_CUDA_CHECK(cudaGetDevice(&deviceId));
+    auto const size = comm.getSize();
+    std::vector<int> deviceIds(size);
+    comm.allgather(&deviceId, deviceIds.data(), 1, tensorrt_llm::mpi::MpiType::kINT32);
+    return std::unordered_set<int>(deviceIds.begin(), deviceIds.end()).size() == deviceIds.size();
```
litaotju, line 79: I think this is not right when we running multi nodes.
Example: 
- node0: 0,1,2,3
- node1: 0,1,2,3.

How do you think?

litaotju (changes requested):

cpp/tensorrt_llm/common/opUtils.cpp
yuxianq: This PR runs allgather on `LOCAL_COMM_SESSION`, so the device IDs are still unique.

litaotju (approved):

PR #3531, opened 2025-04-14
CI: Performance regression tests update

amirkl94:
This PR updates the performance regression tests running in CI.

The changes include:
1. `l0_perf` tests update.
2. Update `sanity_perf_check.py` script to generate patch file for differing metrics.
3. A tool for generating comparison reports between 2 performance test results.
4. Updating `base_perf.csv`.

Example plot of report (pytorch inference time):
image


jenkins/L0_Test.groovy
```diff
@@ -1256,7 +1264,6 @@ def launchTestJobs(pipeline, testFilter, dockerNode=null)
         "H100_PCIe-TensorRT-[Post-Merge]-2": ["h100-cr", "l0_h100", 2, 2],
         "DGX_H100-4_GPUs-PyTorch-[Post-Merge]": ["dgx-h100-x4", "l0_dgx_h100", 1, 1, 4],
         "DGX_H100-4_GPUs-TensorRT-[Post-Merge]": ["dgx-h100-x4", "l0_dgx_h100", 1, 1, 4],
-        "A100_80GB_PCIE-TensorRT-Perf": ["a100-80gb-pcie", "l0_perf", 1, 1],
```
venkywonka, line 1259: what's the reason for this removal? is ampere outdated in this context?

tests/integration/defs/perf/sanity_perf_check.py
```diff
@@ -12,121 +12,58 @@
...
-                            cleaned_row.append(v)
-
-                    print(",".join(['\"' + row + '\"' for row in cleaned_row]))
-                    print('=' * 40)
-                    break
+    def report_diff(self, full_diff: pd.DataFrame) -> None:
+        print("=" * 40)
+        for diff in full_diff.itertuples():
+            print(
+                f"perf_case_name: {diff.Index}, base->target: {diff.perf_metric_base}->{diff.perf_metric_target}"
```
venkywonka, line 34: since it only prints `diff.perf_metric_*` without printing `diff.metric_type`, it might not be straightforward what the metric is right? I understand that we can probably deduce what that is from `perf_case_name: <...>`. 
So I imagine it would look something like :
```bash
========================================
perf_case_name: H100_PCIe-TensorRT-Perf/perf/test_perf.py::test_perf_metric_seq_throughput[llama_v3.1_8b_instruct-bench-pytorch-float16-maxbs:512-maxnt:2048-input_output_len:128,128-reqs:8192], base->target: 82.63618->yy.xxxx
...
========================================
```
 Would you think its worth adding code to help make it obvious if this is a perf regression or improvement or if higher is better etc. for the metric etc.
 
 I see some removals for code that does this kind - hence.

jenkins/L0_Test.groovy
amirkl94: I think it is outdated (@kaiyux correct me if I'm wrong).
Either way, as a first step we discussed only re-enabling for H100 the regression checks for H100.

tests/integration/defs/perf/sanity_perf_check.py
amirkl94: We also create a report file that compares all the metrics between base and current, so I think it will be obvious there if need be.
I did think it'll be clear enough from the `perf_case_name` but you are correct, it might not be obvious enough. I'll think about it for a bit and will update.

tests/integration/defs/perf/base_perf.csv
```diff
@@ -0,0 +1,8 @@
+network_name,perf_case_name,test_name,threshold,absolute_threshold,metric_type,perf_metric
+"llama_v3.1_8b_instruct-bench-float16-maxbs:512-maxnt:2048-input_output_len:128,128-reqs:8192","H100_PCIe-TensorRT-Perf/perf/test_perf.py::test_perf_metric_build_time[llama_v3.1_8b_instruct-bench-float16-maxbs:512-maxnt:2048-input_output_len:128,128-reqs:8192]","test_perf_metric_build_time[llama_v3.1_8b_instruct-bench-float16-maxbs:512-maxnt:2048-input_output_len:128,128-reqs:8192]",0.1,30,BUILD_TIME,143.5976
```
kaiyux, line 2: Should we move `BUILD_TIME`, `INFERENCE_TIME`, etc as columns?

kaiyux:
@amirkl94 FYI - you can use `bot run --stage-list "xxx"` to run specific pipeline stages.

tests/integration/defs/perf/base_perf.csv
amirkl94: They are values in the `metric_type` column

jenkins/L0_Test.groovy
```diff
@@ -1272,7 +1280,6 @@ def launchTestJobs(pipeline, testFilter, dockerNode=null)
...
         "H100_PCIe-TensorRT-Perf": ["h100-cr", "l0_perf", 1, 1],
```
kaiyux, line 1283: Can we also add `H100_PCIe-PyTorch-Perf`?

tests/integration/test_lists/test-db/l0_perf.yml
```diff
@@ -7,26 +7,11 @@ l0_perf:
...
-      - perf/test_perf.py::test_perf[gpt_350m-cppmanager-plugin_ifb-float16-bs:32-input_output_len:60,20]
-      - perf/test_perf.py::test_perf[gpt_350m-cppmanager-plugin_ifb-float16-gwp:0.0-bs:32-input_output_len:60,20]
-      - perf/test_perf.py::test_perf[gpt_350m-cppmanager-static_batching-plugin_ifb-float16-bs:32-input_output_len:60,20]
-      - perf/test_perf.py::test_perf[gpt_350m-cppmanager-plugin-float16-bs:32-input_output_len:60,20]
-      - perf/test_perf.py::test_perf[gpt_350m-cppmanager-static_batching-plugin-float16-bs:32-input_output_len:60,20]
-      - perf/test_perf.py::test_perf[t5_base-plugin-float16-bs:8-input_output_len:60,20]
-      - perf/test_perf.py::test_perf[flan_t5_base-plugin-float16-bs:8-input_output_len:60,20]
-      - perf/test_perf.py::test_perf[bart_large_cnn-plugin-float16-bs:8-input_output_len:60,20]
+      - perf/test_perf.py::test_perf[llama_v3.1_8b_instruct-bench-float16-input_output_len:128,128-reqs:8192]
+      - perf/test_perf.py::test_perf[llama_v3.1_8b_instruct-bench-pytorch-float16-input_output_len:128,128-reqs:8192]
```
kaiyux, line 17: Seems that this is using the pytorch workflow, but it will be included in `H100_PCIe-TensorRT-Perf`?

tests/integration/defs/perf/base_perf.csv
kaiyux: Yes, I was wondering if it's clearer to make each of the metric one column, what do you think?

tests/integration/defs/perf/base_perf.csv
```diff
@@ -0,0 +1,8 @@
+network_name,perf_case_name,test_name,threshold,absolute_threshold,metric_type,perf_metric
```
chzblych, line 1: Are the perf baselines/results stable enough for CI testing?

chzblych (approved):

kaiyux (approved):

tests/integration/defs/perf/base_perf.csv
amirkl94: I think it is. Either way, currently we won't fail CI on perf regressions and I'll make sure we don't get any false-alarms in the next couple of weeks, after we see it's stable I'll enable it.

tests/integration/defs/perf/diff_tools.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import annotations
```
hypdeb, line 1: Which feature is this required for?

tests/integration/defs/perf/create_perf_comparison_report.py
```diff
@@ -0,0 +1,154 @@
...
+
+    return name_mapping, merged, suffixes
+
+
+def generate_perf_compare_report(perf_files: list[str], output_path: str):
+    name_mapping, merged, suffixes = parse_perf_data(perf_files)
+    generate_plots(Path(output_path), name_mapping, merged, suffixes)
+
+
+def parse_arguments():
```
hypdeb, line 136: Type hints

tests/integration/defs/perf/create_perf_comparison_report.py
```diff
@@ -0,0 +1,154 @@
...
+
+    merged[MEAN_COL] = merged[[
+        f'{METRIC_VALUE}_{suffix}' for suffix in suffixes
+    ]].mean(axis=1)
+    merged, name_mapping = shorten_names(merged)
+
+    return name_mapping, merged, suffixes
+
+
+def generate_perf_compare_report(perf_files: list[str], output_path: str):
```
hypdeb, line 131: return type hint

jdebache (changes requested):
The Python utilities could use some documentation on the shapes and types of the dataframes and numpy arrays they take / return. I wouldn't be against a few unit tests for those either.

tests/integration/defs/perf/create_perf_comparison_report.py
```diff
@@ -0,0 +1,155 @@
...
+    plt.title("Long name to short name mapping")
+    table = ax.table(cellText=table_data, cellLoc='left', loc='center')
+    table.auto_set_font_size(False)
+    table.set_fontsize(10)
+    table.scale(1, 1.5)
+    pdf.savefig(fig)
+    plt.close(fig)
+
+
+def plot_metric(merged: pd.DataFrame, metric: str,
```
zeroepoch, line 46: Can you attach an example image from a resulting plot to help with the review?

tests/integration/defs/perf/README.md
```diff
@@ -1,11 +1,62 @@
...
+2. The current HEAD introduces a new setup/metric in `l0_perf.yml` or removes some of them. This will result in new metrics collected by `test_perf.py` which will **fail** `sanity_perf_check.py`. This requires an update for `base_perf.csv`.
+3. The current HEAD improves performance for at least one metric by more than the accepted threshold, which will **fail** `sanity_perf_check.py`. This requires an update for `base_perf.csv`
+4. The current HEAD introduces a regression to one of the metrics that is over the accepted threshold, which will **fail** `sanity_perf_check.py`. This will require to fix the current branch and rerun the pipeline.
+
+### Updating `base_perf.csv`
+If a CI run fails `sanity_perf_check.py`, it will upload a patch file as an artifact. This file can be applied to current branch using `git apply <patch_file>`.
+
+This patch will only update the metrics that had a difference which was over the accepted threshold. The patch will also remove/add metrics according to the removed or added tests.
+
+## Running locally
```
zeroepoch, line 35: We should include a `requirements.txt` or `pyproject.toml` file to help with installing dependencies.

zeroepoch (approved):

tests/integration/defs/perf/README.md
amirkl94: I think the `requirements.txt` file is better since this is a short script. Should I just place it in the `perf` dir?

tests/integration/defs/perf/create_perf_comparison_report.py
zeroepoch: Thanks. This looks good.

zeroepoch (approved):

jdebache (approved):

PR #3541, opened 2025-04-14
feat: Adding FP8 BMM from Codegen

evezhier:
Adds a standalone FP8 BMM from Codegen and the corresponding pytorch op. 
TRT-LLM Gen commit: `2c51b3636b939a8d9985f05e64f7f50abb18a611`


tests/unittest/_torch/test_fp8_batched_gemm.py
```diff
@@ -0,0 +1,380 @@
...
+
+
+@dataclass(frozen=True)
+class Fp8BatchedGemmTestCase:
+    b: int
+    m: int
+    n: int
+    k: int
+    dtypeC: torch.dtype
+    ds: bool
```
DomBrown, line 33: I presume this toggles use_deepseek_fp8? I'd rename this variable so that this is clear.

cpp/tensorrt_llm/thop/fp8BatchedGemmTrtllmGen.cpp
```diff
@@ -0,0 +1,231 @@
...
+    out = batchM ? at::narrow(out, 1, 0, m) : at::narrow(out, 1, 0, n);
+
+    return out;
+}
+} // namespace
+
+namespace torch_ext
+{
+
+extern at::Tensor fp8_batched_gemm(at::Tensor& inputBatchA, int64_t m, at::Tensor const& dsPerInputAScalingFactors,
```
DomBrown, line 201: Nitpick: I don't think we need `extern` here right? As I understand it, functions have external linkage unless made static.

cpp/tensorrt_llm/kernels/trtllmGenKernels/batchedGemm/kernelRunner.cpp
```diff
@@ -0,0 +1,128 @@
...
+    TLLM_CU_CHECK(mDriver->cuFuncSetAttribute(
+        mFunction, CU_FUNC_ATTRIBUTE_MAX_DYNAMIC_SHARED_SIZE_BYTES, mKernelInfo->sharedMemSize));
+}
+
+void TrtllmGenBatchedGemmRunner::run(int32_t m, int32_t n, int32_t k, void* a, void* b, void* c, float const* cScale,
+    float* dDqSfsA, float* dDqSfsB, float* dDqSfsC, std::vector<int32_t> const& batchedM,
+    std::vector<int32_t> const& batchedN, CUstream stream)
+{
+    // Shuffle the A matrix (if needed). TODO: on-device function
+    if (mKernelInfo->useShuffledMatrixA)
```
DomBrown, line 79: Will we start doing this on the device once the functionality is available in main? We don't want do be doing this on the host in production. I don't think this should block the PR, though, as long as there is some timeline to fix it properly.

DomBrown (approved):
Approving as suggested changes are very minor. I don't need to re-review this afterwards :)

cpp/tensorrt_llm/kernels/trtllmGenKernels/batchedGemm/kernelRunner.cpp
evezhier: Yes, we aligned with @nekorobov that weight shuffling should be on device and called upon startup, but this requires extensive modifications out of the scope of this PR, so we're okay to merge this as an intermediate result.

tests/unittest/_torch/test_fp8_batched_gemm.py
evezhier: This will be refactored to adopt the new integration workflow, we can pick the cosmetic stuff then

cpp/tensorrt_llm/thop/fp8BatchedGemmTrtllmGen.cpp
evezhier: I'm not sure if torch has any additional requirements. Blockscale GEMM uses `extern` here. I suppose, we can leave this for refactoring when migrating to the new integration workflow

DomBrown: Fine with me.

hlu1:
Can you add the trtllm-gen commit for the cubins to the summary for bookkeeping?

PR #3550, opened 2025-04-14
fix: remove extraneous local docker user command

tburt-nv:
These lines were added accidentally, and break `make -C docker run LOCAL_USER=1`.


tburt-nv:
/bot skip --comment "change affects only local dev and not L0"

vegaluisjose (approved):
LGTM

Funatiq:
/bot skip --comment "change affects only local dev and not L0"

PR #3552, opened 2025-04-15
test: Unwaive test for nvbug_5150466 

hchings:
Using this MR to verify whether a flaky test is no longer an issue, since I cannot repro the issue locally after 50+ continuous runs and there had been a related fix.


Superjomn (approved):
LGTM

But let's try to run the CI at least twice to make sure there are no flaky issues.

PR #3553, opened 2025-04-15
Clean up linear.py, mlp.py, gated_mlp.py

hlu1:
- Rename `is_expert` to `reduce_output` and the relationship between the two is `reduce_output = not is_expert`.
- Rename `has_nv_fp4` to `has_nvfp4`.
- Change local func `copy` to `_copy`
- Refactor out `_maybe_fuse_bias_into_allreduce` to make the code cleaner.
- Add `forward_lora` to mlp/gated_mlp so the normal forward is not crowded with lora code.


QiJune (approved):
LGTM

PR #3558, opened 2025-04-15
chore: add assertion for devices to avoid underlying errors

Superjomn:


Shixiaowei02 (approved):

tensorrt_llm/executor/proxy.py
```diff
@@ -298,6 +300,12 @@ def mpi_done_callback(future: concurrent.futures.Future):
             self.mpi_session.shutdown()
             raise ready_signal
 
+    def _check_enough_devices(self):
+        if torch.cuda.device_count() < self.model_world_size:
```
litaotju: This is not right for multi node right?

local device count can be smaller than world size.

litaotju (approved):

tensorrt_llm/executor/proxy.py
Superjomn: Updated.

PR #3562, opened 2025-04-15
fix: LLM API _hf_model_dir for non-cached case

syuoni:


tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -642,17 +641,18 @@ def __call__(self) -> Tuple[Path, Union[Path, None]]:
...
 
+        if self.model_loader.model_obj.is_hub_model:
+            # This will download the config.json from HF model hub, this helps to create a PretrainedConfig for
+            # cache key.
+            self._hf_model_dir = download_hf_pretrained_config(
+                self.model_loader.model_obj.model_name,
+                revision=self.llm_args.revision)
+
+        elif self.model_loader.model_obj.is_local_model:
+            self._hf_model_dir = self.model_loader.model_obj.model_dir if self.llm_args.model_format is _ModelFormatKind.HF else None
```
syuoni, line 652: Hi @Superjomn ,

Users may found it convenient to access the HF model path via `llm._hf_model_dir`. Currently, this field is valid for 
* PyTorch flow; and
* TRT flow with build cache enabled.

This PR fix `llm._hf_model_dir` for the TRT flow without build cache enabled.

yechank-nvidia: Thx for the fix @syuoni !, have one question, what if user gives model_path as TensorRT engine path and passing Path for the tokenizer, do we expect this `self._hf_model_dir` to have the value?

syuoni: No. In that case, `llm._hf_model_dir` should be `None`.

Users are allowed to provide a loaded tokenizer to `LLM(tokenizer=...)`. Even if it's a path, it's not guaranteed that the path points to a folder that contains complete files of a HF model checkpoint.

yechank-nvidia: I see. Then is there a way of getting the user's given tokenizer path from LLM's class attributes? 

=> This can be retrieved by `self.tokenizer.tokenizer.name_or_path`.

yechank-nvidia (approved):

tensorrt_llm/llmapi/llm_utils.py
Superjomn: I wonder if a test can cover this specific case.

syuoni: I think this case is a bit trivial. We can add a test if it's raised to be a formal API.

Superjomn (approved):
LGTM

PR #3565, opened 2025-04-15
Chore: Remove test_fp4_quantize_gemm_torch_profiling test.

hyukn:
Because it is duplicated with test_fp4_linear. Also, the original CPP profiler has been unified with the new AutoTuner already.


hlu1 (approved):

hlu1:
Remove dead code. No CI run is needed.

PR #3567, opened 2025-04-15
fix: FP8 quantized lm_head (NvBug 5214229)

syuoni:


tensorrt_llm/quantization/quantize.py
```diff
@@ -244,9 +246,9 @@ def fp8_rowwise_quantize(model, quant_config: QuantConfig):
         Attention: Fp8RowwiseAttention,
     }
 
-    exclude_modules = list(
-        set((quant_config.exclude_modules or []) +
-            ['*ln_f', '*ln_embed', '*lm_head']))
```
nv-guomingz, line 249: I think user_defined exclude_modules should be the superset of `['*ln_f', '*ln_embed', '*lm_head']`.

exclude_modules = list(
        set((quant_config.exclude_modules or []) +
            ['*ln_f', '*ln_embed', '*lm_head']))

should be exclude_modules = list(
        set(quant_config.exclude_modules or []+ ['*ln_f', '*ln_embed', '*lm_head'])

tensorrt_llm/quantization/quantize.py
```diff
@@ -28,14 +28,16 @@ def quantize_layers(
...
 ):
-    exclude_modules = quant_config.exclude_modules or [
-        '*lm_head',
-        '*router',
-        '*vocab_embedding',
-        '*position_embedding',
-        '*block_embedding',
-        '*shared_expert_gate',
-    ]
+    exclude_modules = quant_config.exclude_modules
```
nv-guomingz, line 31: what's the difference between new implementation and the original one?

tensorrt_llm/quantization/quantize.py
syuoni: Oh this is unexpected. Doesn't that mean users can never quantize `lm_head`? Even if they provide an empty `exclude_modules`.

tensorrt_llm/quantization/quantize.py
syuoni: The case is, users want to quantize `lm_head` by providing `exclude_modules = []`.

The original code isn't type-safe, and overrides the empty `exclude_modules` with default ones.

tensorrt_llm/quantization/quantize.py
nv-guomingz: Yes, let me ping u for details

nv-guomingz (approved):
LTGM

tensorrt_llm/quantization/quantize.py
syuoni: Reverted to the original logic. I made it more clear and add a comment here so that other developers can understand the intention. Thanks!

Barry-Delaney (approved):

PR #3571, opened 2025-04-15
feat: support kv cache reuse for MLA

zhhuang-nv:
Workflow:
1. Load compressed_kv and k_pe and do up-projection.
2. Apply RoPE for current q and k, then set current compressed_kv and k_pe into paged KV Cache.
3. Build full KV and its offsets, then use 192/128 head size MLA context kernel with q_paged_kv input type to compute.

Performance Test with DeepSeek-R1-FP4 on 8 * B200:
1. Use trtllm-bench throughput (ref: https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/blogs/Best_perf_practice_on_DeepSeek-R1_in_TensorRT-LLM.md#b200-max-throughput)

w/o reuse:
```
===========================================================
= PERFORMANCE OVERVIEW 
===========================================================
Request Throughput (req/sec):                     17.1784
Total Output Throughput (tokens/sec):             35181.2906
Total Token Throughput (tokens/sec):              52771.9358
Total Latency (ms):                               2861273.5462
Average request latency (ms):                     178135.6332
Per User Output Throughput [w/ ctx] (tps/user):   11.5321
Per GPU Output Throughput (tps/gpu):              4397.6613
```

w/ reuse:
```
===========================================================
= PERFORMANCE OVERVIEW 
===========================================================
Request Throughput (req/sec):                     17.0529
Total Output Throughput (tokens/sec):             34924.3690
Total Token Throughput (tokens/sec):              52386.5534
Total Latency (ms):                               2882322.5442
Average request latency (ms):                     179450.9751
Per User Output Throughput [w/ ctx] (tps/user):   11.4620
Per GPU Output Throughput (tps/gpu):              4365.5461
```

Total token throughput drops about 1%, probably because some requests could reuse KV Cache but only could reuse few tokens.

2. trtllm-serve, submit requests of 5000 samples from Orca_long_system_prompt dataset

w/o reuse:
```
============ Serving Benchmark Result ============
Backend:                                 trtllm-serve
Traffic request rate:                    inf       
Max reqeuest concurrency:                not set   
Successful requests:                     5000      
Benchmark duration (s):                  96.94     
Total input tokens:                      3531042   
Total generated tokens:                  743596    
Total generated tokens (retokenized):    656550    
Request throughput (req/s):              51.58     
Input token throughput (tok/s):          36425.13  
Output token throughput (tok/s):         7670.71   
Total token throughput (tok/s):          44095.84  
Concurrency:                             3049.57   
```

w/ reuse:
```
============ Serving Benchmark Result ============
Backend:                                 trtllm-serve
Traffic request rate:                    inf       
Max reqeuest concurrency:                not set   
Successful requests:                     5000      
Benchmark duration (s):                  74.61     
Total input tokens:                      3531042   
Total generated tokens:                  743596    
Total generated tokens (retokenized):    667215    
Request throughput (req/s):              67.02     
Input token throughput (tok/s):          47329.66  
Output token throughput (tok/s):         9967.07   
Total token throughput (tok/s):          57296.73  
Concurrency:                             2943.31   
```

Total token throughput increases about 30%.

Note:
1. Only support SM90 and SM100 now.
2. Performance is not good enough, which will be improved in later PRs.


tests/integration/defs/accuracy/test_llm_api_pytorch.py
```diff
@@ -297,7 +298,8 @@ def test_bfloat16(self, mtp_nextn, attention_dp, cuda_graph,
     def test_bfloat16_4gpus(self, tp_size, pp_size, ep_size, mtp_nextn,
                             attention_dp, cuda_graph, overlap_scheduler):
         # OOM on H100 with default free_gpu_memory_fraction=0.9
-        kv_cache_config = KvCacheConfig(free_gpu_memory_fraction=0.7)
+        kv_cache_config = KvCacheConfig(free_gpu_memory_fraction=0.7,
+                                        enable_block_reuse=False)
```
litaotju: Curious, whats the reason to disable the block reuse here?

tests/integration/defs/accuracy/test_llm_api_pytorch.py
```diff
@@ -430,6 +440,34 @@ def test_nvfp4_4gpus(self, tp_size, pp_size, ep_size, attention_dp,
...
+    def test_kv_cache_reuse(self, quant_dtype):
+        model_path = self.MODEL_PATH
+        if quant_dtype == "fp8":
+            model_path = f"{llm_models_root()}/DeepSeek-V3-Lite/fp8"
+        elif quant_dtype == "nvfp4":
+            model_path = f"{llm_models_root()}/DeepSeek-V3-Lite/nvfp4_moe_only"
+
+        kv_cache_config = KvCacheConfig(free_gpu_memory_fraction=0.6,
+                                        enable_block_reuse=True)
+        pytorch_config = PyTorchConfig(enable_overlap_scheduler=False,
```
litaotju: whats the reason to turn these off?

Is KV cache reuse not working with overlap scheduler and cuda graph now? Thx

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -131,7 +131,16 @@ def create_py_executor(executor_config: ExecutorConfig,
...
             )
-        executor_config.kv_cache_config.enable_block_reuse = False
+        if executor_config.kv_cache_config.enable_block_reuse:
+            if is_hopper() or is_blackwell():
+                logger.info(
+                    f"Enable block reuse for MLA, performance may have regression in some cases"
+                )
+            else:
+                logger.warning(
+                    f"Block reuse for MLA only can be enabled on Hopper/Blackwell, change enable_block_reuse to False"
```
litaotju: Nit. 

Are we calling it "block reuse" or "kv-cache reuse" in the user document about the feature?

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -131,7 +131,16 @@ def create_py_executor(executor_config: ExecutorConfig,
             logger.info(
                 f"Change tokens_per_block to: {executor_config.tokens_per_block} for using FlashMLA"
             )
-        executor_config.kv_cache_config.enable_block_reuse = False
+        if executor_config.kv_cache_config.enable_block_reuse:
+            if is_hopper() or is_blackwell():
+                logger.info(
+                    f"Enable block reuse for MLA, performance may have regression in some cases"
```
litaotju: Shall we add this into the DeepSeek example doc instead of the log msg here? 
I don't think one will look into this info level msg about the perf impact.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -36,6 +37,22 @@ def is_mla(config):
...
+    except Exception:
+        return None
+
+
+def is_hopper():
+    return get_cuda_arch() == (9, 0)
+
+
+def is_blackwell():
+    return get_cuda_arch() == (10, 0)
```
litaotju: blackwell is not only about sm100, also sm120, and some others.

tensorrt_llm/_torch/modules/rotary_embedding.py
```diff
@@ -88,6 +87,100 @@ def compute_llama3_parameters(
     return inv_freq_llama, attention_factor
 
 
+def compute_deepseek_parameters(
+    config: PretrainedConfig,
+    device: Optional[torch.device] = None,
+) -> "Tuple[torch.Tensor, float]":
```
litaotju: https://github.com/NVIDIA/TensorRT-LLM/pull/3517/files#diff-0ab1c25929efb031ef9db9ad5ae8f325848cb281ab198e9bb1469628d8ce40fe

Also toches this. lets make sure these two commits aligns. Thx

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -585,12 +586,13 @@ def forward(
...
         return attn_output
 
-    def forward_context(
+    def forward_context_default(
         self,
         q: torch.Tensor,
         compressed_kv: torch.Tensor,
         k_pe: torch.Tensor,
         attn_metadata: AttentionMetadata,
+        position_ids: Optional[torch.LongTensor] = None,
```
litaotju: This params is never used?

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -629,6 +631,144 @@ def forward_context(
 
         return attn_output
 
+    def forward_paged_context_fmha_for_mla(
```
litaotju: is forward_context_kv_reuse is better name?

Using "paged_context_fmha" is an implementation detail for kv cache reuse, I feel that "forward_context_kv_reuse" can specify why we need a new function in the first place.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -629,6 +631,144 @@ def forward_context(
...
+        self,
+        q: torch.Tensor,
+        compressed_kv: torch.Tensor,
+        k_pe: torch.Tensor,
+        attn_metadata: AttentionMetadata,
+        position_ids: Optional[torch.LongTensor] = None,
+    ) -> torch.Tensor:
+        trtllm_attention = cast(TrtllmAttention, self.mha)
```
litaotju, line 686: Just for my own education and not a issue. The cast here is not required right? Since python has duck typing.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -629,6 +631,144 @@ def forward_context(
...
+                self.num_heads * self.v_head_dim
+            ],
+            -1,
+        )
+        past_k_nope = past_k_nope.view(-1, self.num_heads,
+                                       self.qk_nope_head_dim)
+        past_v = past_v.view(-1, self.num_heads, self.v_head_dim)
+
+        # compute current k_nope and v from compressed_kv
+        kv = self.kv_b_proj(compressed_kv)
```
litaotju, line 712: Shall we concat and then b_proj, and then split?

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -629,6 +631,144 @@ def forward_context(
...
+               self.qk_nope_head_dim:] = q_pe.view(-1, self.num_heads,
+                                                   self.qk_rope_head_dim)
+        q = q_view.view(
+            -1,
+            self.num_heads * (self.qk_nope_head_dim + self.qk_rope_head_dim))
+        assert q.is_contiguous()
+
+        # set compressed paged kv cache for mla
+        # we may finish it inside the attention op by passing latent_cache
+        trtllm_attention.set_compressed_paged_kv_cache_for_mla(
```
litaotju: "append_paged_kv_cache*", we are only appending not set/override it right?

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -629,6 +631,144 @@ def forward_context(
...
+        full_kv = torch.zeros([
+            attn_metadata.num_contexts, 2,
+            (attn_metadata.max_ctx_full_seq_len + tokens_per_block - 1) //
+            tokens_per_block, self.num_heads, tokens_per_block,
+            max(self.qk_nope_head_dim + self.qk_rope_head_dim, self.v_head_dim)
+        ],
+                              dtype=q.dtype,
+                              device=q.device)
+
+        mla_context_kv_cache_block_offsets = trtllm_attention.set_paged_kv_cache_v2_for_mla(
```
litaotju, line 772: I do not fully understand the "v2" set here.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -726,13 +816,173 @@ def forward(
...
+            self.mla_params.qk_nope_head_dim,
+            self.mla_params.qk_rope_head_dim,
+            tokens_per_block,
+        )
+
+        max_block_num = (max_seq_len + tokens_per_block - 1) // tokens_per_block
+        assert paged_kv_offsets.shape == (num_contexts, 2, max_block_num)
+        return paged_kv_offsets
+
+    def set_paged_kv_cache_v2_for_mla(
```
litaotju, line 917: whats the diff of v2 and non-v2 here?

cpp/tests/unit_tests/kernels/mlaPreprocessTest.cu
```diff
@@ -0,0 +1,708 @@
...
+                    static_cast<float>(expected[i]), static_cast<float>(output[i]));
+                return false;
+            }
+        }
+        return true;
+    }
+};
+
+using MLATypes = ::testing::Types<half,
+#ifdef ENABLE_BF16
```
litaotju: This is not needed anymore. 
We shall always ENBLE_BF16. Volta support had been removed.

cpp/tensorrt_llm/thop/mlaPreprocessOp.cpp
```diff
@@ -0,0 +1,496 @@
...
+    {
+        loadPagedKVCacheForMLAHelper<half>(
+            output, kv_cache_buffer, num_contexts, cu_ctx_cached_kv_lens, max_ctx_cached_kv_len, head_size);
+    }
+    else if (out_dtype == torch::kFloat32)
+    {
+        loadPagedKVCacheForMLAHelper<float>(
+            output, kv_cache_buffer, num_contexts, cu_ctx_cached_kv_lens, max_ctx_cached_kv_len, head_size);
+    }
+#ifdef ENABLE_BF16
```
litaotju: Not needed anymore

tests/integration/defs/accuracy/test_llm_api_pytorch.py
zhhuang-nv: I want to keep the behavior of these tests not changed, e.g., output results, total duration, activation memory usage, and etc. But since we enable kv cache reuse by default, it also makes sense that we run these tests with block reuse enabled. Which way is better in your option?

tests/integration/defs/accuracy/test_llm_api_pytorch.py
zhhuang-nv: Just to reduce the number of test case combinations. KV Cache reuse works with these two features. Maybe I can add the options and just run the test with these features both enabled.

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
zhhuang-nv: Thanks, let me change it to "KV Cache reuse" to align with other documents.

tensorrt_llm/_torch/modules/rotary_embedding.py
zhhuang-nv: I will rebase these part after https://github.com/NVIDIA/TensorRT-LLM/pull/3610 has been merged

tensorrt_llm/_torch/modules/attention.py
zhhuang-nv: Chunked context will also go into this function. Maybe we can change the name to "forward_context_with_cached_kv". Basically KV Cache reuse and chunked context are very similar since they both have cached KV in context phase.

tensorrt_llm/_torch/modules/attention.py
zhhuang-nv: I think we cannot simply concat past_compressed_kv and compressed_kv together. The token order cannot be kept by simple concatenation.
For example, we have batch_size=2, cached_token_num=[10, 20], uncached_token_num=[30,40]. Then past_compressed_kv.shape[0] would be 10+20=30, compressed_kv.shape[0] would be 30+40=70.
What we want for the order of up-projected tokens' kv is [10,30,20,40] (total tokens of req0, then total tokens of req1). If we use simple concatenation, the order will be [10,20,30,40], and we have to use another kernel to adjust the order, which will bring more complexity here.
So I tend to keep serveral kv_b_proj calls and merge the results with a single kernel.

tensorrt_llm/_torch/modules/attention.py
zhhuang-nv: V1 requires all the KV have correct order as mentioned before. V2 take the past tensors and current tensors separately and merge them in a single kernel.

tensorrt_llm/_torch/attention_backend/trtllm.py
zhhuang-nv: V1 requires all the KV have correct order as mentioned before. V2 take the past tensors and current tensors separately and merge them in a single kernel.

tensorrt_llm/_torch/modules/attention.py
zhhuang-nv: Yes it is not required for correctness. I add it here to make IDE happy, e.g., we can jump to the implementation smoothly.

tests/integration/defs/accuracy/test_llm_api_pytorch.py
litaotju: Lets just use the "default" config (enable kv cache reuse) for these accuracy tests, which is more important to users and memic the real use cases.

I don't think the original test want to emphasize on the not-enabling-kv-cache-reuse, it just used the default and w/o knowing the underlying choices.

tests/integration/defs/accuracy/test_llm_api_pytorch.py
litaotju: Yes, lets test with both of them enabled, which is more important I think.
When local debugging, we can turn it on/off to isolate issue.

tensorrt_llm/_torch/modules/attention.py
litaotju: Looks good.

tensorrt_llm/_torch/modules/attention.py
litaotju: I see, I should not mention it as "concat", a better naming is "batched concat" or 2 scatters

zhhuang-nv:
Some CI tests failed which depends on https://github.com/NVIDIA/TensorRT-LLM/pull/4084, will update this PR once PR 4084 has been merged.

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -140,7 +140,12 @@ def create_py_executor(executor_config: ExecutorConfig,
             logger.info(
                 f"Change tokens_per_block to: {executor_config.tokens_per_block} for using FlashMLA"
             )
-        executor_config.kv_cache_config.enable_block_reuse = False
+        if executor_config.kv_cache_config.enable_block_reuse and not (
+                is_sm90() or is_sm100()):
```
litaotju, line 144: what about sm120?

For RTX6000, does it work?

cpp/tensorrt_llm/kernels/contextFusedMultiHeadAttention/fmhaRunner.cpp
```diff
@@ -394,11 +394,14 @@ void FusedMHARunnerV2::setupLaunchParams(MHARunnerParams runnerParams)
...
         mLaunchParams.kernel_s = 0;
 
-        // Now we have SM90 generation MLA kernels. These treatments are only for context MLA and non SM90 generation
-        // MLA.
-        bool isFP8GenerationMLAOnHopper = mFixedParams.dataType == DATA_TYPE_E4M3
-            && (mFixedParams.headSize == 576 && mFixedParams.headSizeV == 512) && isSm90;
-        if (!isFP8GenerationMLAOnHopper)
+        // Now we have SM90 FP8 generation and BF16 context MLA kernels
+        bool isHopperFP8GenerationMLA
+            = isSm90 && mFixedParams.dataType == DATA_TYPE_E4M3 && mFixedParams.headSizeV == 512;
```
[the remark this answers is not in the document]
zhhuang-nv: We have this strong assumption in many places, e.g., FMHA kernels, MLA related preprocess kernels, runtime workflow and etc. We cannot make it general at this time, maybe we can do that when there are real requirements.

cpp/tensorrt_llm/kernels/mlaKernels.cu
```diff
@@ -61,50 +61,71 @@ template <typename T>
 struct VecType
 {
     using Type = T;
+    using GPTJEltType = T;
```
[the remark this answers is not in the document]
zhhuang-nv: It is used for GTPJ style RoPE calculation, which is composed of two basic elements.

examples/models/core/deepseek_v3/README.md
```diff
@@ -566,6 +567,9 @@ pytorch_backend_config:
   # ...
 ```
 
+### KV Cache Reuse
+KV cache reuse is supported for MLA on SM90 and SM100. It is enabled by default and does not support FP8 KV cache right now. Due to extra operations like memcpy and GEMMs, GPU memory consumption may be higher and the E2E performance may have regression in some cases. Users could pass `KvCacheConfig(enable_block_reuse=False)` to LLM API to disable it.
```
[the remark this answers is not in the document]
zhhuang-nv: Currently we would pad all context requests in the same batch to `max_context_seq_len_in_this_batch`, which will consume much more activation memory when there is huge variance in the context seq lens. For the perf regression number, I have tested DeepSeek-V3-Lite/nvfp4_moe_only with 5000 samples from Orca_long_system_prompt dataset, and the total token throughput droped about <10%. Numbers will change depends on models and datasets. I have tested DeepSeek-R1-FP4 on 8 * B200 with the same dataset and the total token throughput increased about 30%.

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
zhhuang-nv: I think we don't have sm120 paged context MLA kernels now. So it doesn't work for SM120 like RTX 6000. PR description changed.

litaotju (approved):

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -486,6 +498,33 @@ def __post_init__(self) -> None:
...
+                    device='cuda',
+                    dtype=torch.int64,
+                )
+                self.host_cu_ctx_full_seq_lens = torch.empty_like(
+                    self.cu_ctx_full_seq_lens,
+                    device='cpu',
+                    pin_memory=True,
+                )
+                self.max_ctx_full_seq_len = 0
+                self.max_ctx_uncached_seq_len = 0
```
yuxianq, line 527: Don't need to init `num_ctx_cached_tokens/max_ctx_cached_kv_len/max_ctx_full_seq_len/max_ctx_uncached_seq_len` here, preparing pre-allocated tensors is enough.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -568,6 +611,45 @@ def prepare_flash_mla(self) -> None:
...
+            self.max_ctx_cached_kv_len = 0
+            self.max_ctx_full_seq_len = 0
+            self.max_ctx_uncached_seq_len = 0
+        cu_ctx_cached_kv_lens = torch.cumsum(
+            cached_token_lens[:self.num_contexts], dim=0, dtype=torch.int64)
+        cu_ctx_cached_kv_lens = torch.cat(
+            [torch.zeros(1, dtype=torch.int64), cu_ctx_cached_kv_lens])
+        self.host_cu_ctx_cached_kv_lens[:self.num_contexts +
+                                        1].copy_(cu_ctx_cached_kv_lens)
+        self.cu_ctx_cached_kv_lens[:self.num_contexts + 1].copy_(
```
yuxianq, line 638: Can we copy directly from `cu_ctx_cached_kv_lens` instead of `host_cu_ctx_cached_kv_lens` on CPU?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -568,6 +611,45 @@ def prepare_flash_mla(self) -> None:
...
+            non_blocking=True)
+
+        cu_ctx_full_seq_lens = torch.cumsum(kv_lens[:self.num_contexts],
+                                            dim=0,
+                                            dtype=torch.int64)
+        cu_ctx_full_seq_lens = torch.cat(
+            [torch.zeros(1, dtype=torch.int64), cu_ctx_full_seq_lens])
+        self.host_cu_ctx_full_seq_lens[:self.num_contexts +
+                                       1].copy_(cu_ctx_full_seq_lens)
+        self.cu_ctx_full_seq_lens[:self.num_contexts + 1].copy_(
```
yuxianq, line 649: Can we directly copy from `cu_ctx_full_seq_lens` instead of `host_cu_ctx_full_seq_lens` on CPU?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -735,3 +826,155 @@ def support_fused_qkv(cls) -> bool:
...
+            metadata.kv_cache_manager.kv_cache_pool_pointers,
+            metadata.kv_cache_manager.kv_cache_pool_mapping,
+            self.kv_scale_orig_quant,
+            self.kv_scale_quant_orig,
+            self.layer_idx,
+            metadata.kv_cache_manager.head_dim,
+            metadata.kv_cache_manager.tokens_per_block,
+            metadata.kv_cache_manager.max_seq_len,
+            0,
+            1,
```
yuxianq, line 868: please assign a variable for 0 and 1 here so that we know what they mean.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -735,3 +826,155 @@ def support_fused_qkv(cls) -> bool:
...
+            metadata.kv_cache_manager.kv_cache_pool_pointers,
+            metadata.kv_cache_manager.kv_cache_pool_mapping,
+            self.kv_scale_orig_quant,
+            self.kv_scale_quant_orig,
+            self.layer_idx,
+            self.mla_params.kv_lora_rank + self.mla_params.qk_rope_head_dim,
+            metadata.kv_cache_manager.tokens_per_block,
+            metadata.kv_cache_manager.max_seq_len,
+            0,
+            1,
```
yuxianq, line 978: please assign a variable for 0 and 1 here so that we know what they mean.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -673,6 +673,150 @@ def forward_context(
...
+            full_kv,
+            past_k_nope.contiguous(),
```
yuxianq, line 772: How about calling `contiguous` inside `set_paged_kv_cache_v2_for_mla` so that it can access both contiguous or strided input tensors? The `set_paged_kv_cache_v2_for_mla` op can work instead of throw illegal memory access error when we provide strided tensors.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -658,14 +658,18 @@ def disable_optimization(backend: Backend):
                             torch.cuda.empty_cache()
 
     def _set_up_attn_metadata(self, kv_cache_manager: KVCacheManager):
+        enable_paged_context_mla = hasattr(
```
yuxianq, line 661: Can you extract `hasattr(..., "kv_lora_rank")` as `is_mla` to make it clearer?

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -140,7 +140,12 @@ def create_py_executor(executor_config: ExecutorConfig,
...
+            logger.warning(
+                f"KV cache reuse for MLA only can be enabled on SM90/SM100, change enable_block_reuse to False"
```
yuxianq, line 146: Use `disable enable_block_reuse for SM{get_sm_version()}.` instead of `change enable_block_reuse to False`.

tensorrt_llm/_torch/attention_backend/trtllm.py
zhhuang-nv: I use the cumsum trick and save the output to `host_cu_ctx_cached_kv_lens` directly, then copy from `self.host_cu_ctx_cached_kv_lens` to `self.cu_ctx_cached_kv_lens`. The first copy is necessary because cumsum is run by CPU.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -568,6 +611,45 @@ def prepare_flash_mla(self) -> None:
...
+            self.max_ctx_full_seq_len = kv_lens[:self.num_contexts].max().item()
+            self.max_ctx_uncached_seq_len = self.seq_lens[:self.
+                                                          num_contexts].max(
+                                                          ).item()
+        else:
+            self.num_ctx_cached_tokens = 0
+            self.max_ctx_cached_kv_len = 0
+            self.max_ctx_full_seq_len = 0
+            self.max_ctx_uncached_seq_len = 0
+        cu_ctx_cached_kv_lens = torch.cumsum(
```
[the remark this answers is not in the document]
zhhuang-nv: Thanks for the advice. I change `max_ctx_cached_kv_len ` to `max_ctx_cached_token_len ` because it only store information for context requests. The other names have been modified following your advice.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -233,6 +233,9 @@ def run(
         is_fused_qkv: bool = True,
         update_kv_cache: bool = True,
         attention_mask: AttentionMask = PredefinedAttentionMask.CAUSAL,
+        *,
+        mla_context_paged_kv: Optional[torch.Tensor] = None,
+        mla_context_kv_cache_block_offsets: Optional[torch.Tensor] = None,
```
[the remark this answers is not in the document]
yuxianq: I have merged https://github.com/NVIDIA/TensorRT-LLM/pull/4227 to reset all fields updated in the plan function, so there is no memory leak now.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -568,6 +611,45 @@ def prepare_flash_mla(self) -> None:
...
+                                                          num_contexts].max(
+                                                          ).item()
+        else:
+            self.num_ctx_cached_tokens = 0
+            self.max_ctx_cached_kv_len = 0
+            self.max_ctx_full_seq_len = 0
+            self.max_ctx_uncached_seq_len = 0
+        cu_ctx_cached_kv_lens = torch.cumsum(
+            cached_token_lens[:self.num_contexts], dim=0, dtype=torch.int64)
+        cu_ctx_cached_kv_lens = torch.cat(
```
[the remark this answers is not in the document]
zhhuang-nv: Good advice! Applied.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: OK, make sense since `cu_ctx_cached_kv_lens` is not a CUDA tensor. The name of `cu_ctx_cached_kv_lens` misleads me.

tensorrt_llm/_torch/pyexecutor/model_engine.py
yuxianq: Make sense. Please don't rename `_util.py` before code freeze of release 0.20 before it may cause conflicts with other huge PRs like https://github.com/NVIDIA/TensorRT-LLM/pull/4034. You can rename it after we fork release/0.20.
I also prefer to use `config_utils.py` instead of `_config_utils.py`.

tensorrt_llm/_torch/attention_backend/trtllm.py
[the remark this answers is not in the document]
zhhuang-nv: Great! Modified.

yuxianq (approved):

zhhuang-nv:
/bot run --stage-list "DGX_H100-4_GPUs-PyTorch-Others-1,DGX_H100-4_GPUs-TensorRT-1,DGX_H100-4_GPUs-TensorRT-2"

PR #3572, opened 2025-04-15
chore: Add comments to modifications that fix TP size of DeepSeek-V3/R1 when using more than 16 GPUs

jinyangyuan-nvidia:


hlu1 (approved):

hlu1:
You don't need to run CI on the one. It has no real code changes.

PR #3574, opened 2025-04-15
fix: Remove unnecessary max call

kaiyux:
Remove unnecessary `max` call since chunked context is supported now.


FrankD412 (approved):
Looks good to me -- good to know chunking is now supported! 🙂

Superjomn (approved):
LGTM

PR #3577, opened 2025-04-15
doc: Minor fixes for documents

kaiyux:
The changes are based on a user comment

> https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples/llama#run-inference the guide here says FP8 model on a single H100 node "cannot support 128k context due to memory limitations". I assume it means kv cache size limitation? Based on our test, even using BF16 kv cache we can easily have >=300k kv cache. This guide might need to be updated.


byshiue (approved):
LGTM.

PR #3582, opened 2025-04-15
feat: Integrate GPUDirect Storage (GDS) into Executor API

DomBrown:
Integrates GPUDirect Storage (GDS) into Executor API, to allow faster loading of engines on systems that support GDS.


cpp/tensorrt_llm/runtime/tllmRuntime.cpp
```diff
@@ -225,8 +405,26 @@ TllmRuntime::TllmRuntime(
...
+            TLLM_LOG_INFO("Engine load time %lld ms", elapsed_ms);
+        }
+        else
+        {
+            auto const start_time = std::chrono::high_resolution_clock::now();
+            auto reader = StreamReader(rawEngine.getPath());
+            mEngine.reset(mRuntime->deserializeCudaEngine(reader));
+            auto const elapsed_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                std::chrono::high_resolution_clock::now() - start_time);
+            TLLM_LOG_INFO("Engine load time %lld ms", elapsed_ms);
```
Funatiq: Can we unify these branches please? There is a lot of code duplication. Also please use camelCase for the variables.

cpp/tensorrt_llm/runtime/tllmRuntime.cpp
```diff
@@ -211,10 +213,188 @@ void assessLikelihoodOfRuntimeAllocation(
             numWarnings);
     }
 }
+
+class GDSStreamReader final : public nvinfer1::IStreamReaderV2
```
Funatiq: I think we should move the stream readers to a separate header or even create a separate compilation unit.

cpp/tests/unit_tests/executor/serializeUtilsTest.cpp
```diff
@@ -788,6 +788,7 @@ TEST(SerializeUtilsTest, ExecutorConfig)
         executorConfig2.getParallelConfig().value().getCommunicationMode());
     EXPECT_EQ(executorConfig.getPeftCacheConfig(), executorConfig2.getPeftCacheConfig());
     EXPECT_EQ(executorConfig.getDecodingConfig(), executorConfig2.getDecodingConfig());
+    EXPECT_EQ(executorConfig.getUseGpuDirectStorage(), executorConfig2.getUseGpuDirectStorage());
```
Funatiq, line 791: [There](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/bindings/test_executor_bindings.py) are similar tests in Python that we should extend:
* test_executor_config
* test_executor_config_pickle

Funatiq (changes requested):
Looks good so far. Can we add a functionality test in the CI for this?

DomBrown:
> Looks good so far. Can we add a functionality test in the CI for this?

I'm not sure to be honest. GDS is not supported by most systems. Unsure how we would go about it with that in mind.

DomBrown:
Also looks like the existing reader doesn't have a test either, actually

cpp/tensorrt_llm/runtime/tllmRuntime.cpp
DomBrown: The stream readers inherit from different types so I think we do need a branch here, right?
I've eliminated the duplicated timing code, however.

Funatiq: Maybe we could inherit from V2 in both cases, but I think it's fine to keep the branch.

DomBrown:
All fixed, minus the test. Not decided how to approach that yet

Funatiq (approved):
LGTM

DomBrown:
I'll run CI again then, and merge if all is (still) well. I will add coming up with a proper test for this to my todo list.

PR #3584, opened 2025-04-15
feat: Disaggregated router class

pcastonguay:
Supersedes https://github.com/NVIDIA/TensorRT-LLM/pull/3304 from Shunkang. 

Integrates a router class which can use different strategies to route requests to ctx and gen servers


tensorrt_llm/serve/openai_disagg_server.py
```diff
@@ -169,10 +179,9 @@ async def _process_context_server_request(self, ctx_req, request_type: str):
         ctx_req.stream = False
         ctx_req.stream_options = None
 
-        if request_type == "chat":
-            return await self.send_chat_request(ctx_server, ctx_req)
-        elif request_type == "completion":
-            return await self.send_completion_request(ctx_server, ctx_req)
+        response = await self.send_chat_request(ctx_server, ctx_req) if request_type == "chat" else await self.send_completion_request(ctx_server, ctx_req)
+        self.ctx_router.finish_request(ctx_req)
```
Shunkangz: Can we also wrap a try, finally code block here? In this case, we can always ensure the execution of clean up logic from router. 
```
ctx_server = await self.ctx_router.get_next_server(ctx_req)
logging.info("Sending request to ctx server: %s", ctx_server)

try:
    if request_type == "chat":
        ctx_req.max_completion_tokens = 1
    elif request_type == "completion":
        ctx_req.max_tokens = 1
    ctx_req.disaggregated_params = DisaggregatedParams(request_type="context_only")
    ctx_req.stream = False
    ctx_req.stream_options = None

    if request_type == "chat":
        response = await self.send_chat_request(ctx_server, ctx_req)
    else:
        response = await self.send_completion_request(ctx_server, ctx_req)
    return response 
finally:
    self.ctx_router.finish_request(ctx_req)
```

tensorrt_llm/serve/openai_disagg_server.py
```diff
@@ -201,6 +210,8 @@ async def _process_generation_server_request(self, gen_req, ctx_response):
             elif isinstance(gen_req, ChatCompletionRequest):
                 gen_response = await self.send_chat_request(gen_server, gen_req)
 
+            self.gen_router.finish_request(gen_req)
```
Shunkangz: The same as above.

Shunkangz:
LGTM. One more concern is that the existing integration test only has 1P1D which might not cover the scenario when we enable router.

tensorrt_llm/serve/openai_disagg_server.py
pcastonguay: Done.

tensorrt_llm/serve/openai_disagg_server.py
pcastonguay: Done.

pcastonguay:
> LGTM. One more concern is that the existing integration test only has 1P1D which might not cover the scenario when we enable router.

Added e2e tests with 2 ctx and 2 gen instances, ctx using tokens load balancing and gen using requests_load_balancing.

Shunkangz (approved):

zhengd-nv (approved):

Shunkangz:
/bot run --stage-list  "DGX_H100-4_GPUs-CPP-1,DGX_H100-4_GPUs-PyTorch-1,DGX_H100-4_GPUs-TensorRT-1,DGX_H100-4_GPUs-TensorRT-2"

PR #3530, opened 2025-04-14
refactor: Split llama4 model from llama model.

yuxianq:

PR #3586, opened 2025-04-15
fix: check for config architectures in model config

vanshilshah97:
Encountered a error while running LLama4. Hence fixed


schetlur-nv (approved):

yuxianq:
@vanshilshah97 We have already fix it in https://github.com/NVIDIA/TensorRT-LLM/pull/3530 with https://github.com/NVIDIA/TensorRT-LLM/pull/3530/files#diff-e760e2394238043b4824d0d95f7c55cd7293e6e61cc3e95f398ceaf578b2bf28R703, do we need this fix?

vanshilshah97:
makes sense @yuxianq

PR #3590, opened 2025-04-16
feat: trtllm-serve multimodal support

yechank-nvidia:
This PR is adding support for trtllm-serve + single image(image_url) inference.

For the next step (listed as priority-based),
1. Make the loading of images/videos async(because this can block the LLM inference)
2. multi-image + video support
3. Do the benchmark and optimize the procedure
4. image_embeds support (if needed)


examples/serve/curl_chat_client.sh
```diff
@@ -3,7 +3,7 @@
 curl http://localhost:8000/v1/chat/completions \
     -H "Content-Type: application/json" \
     -d '{
-        "model": TinyLlama-1.1B-Chat-v1.0,
+        "model": "TinyLlama-1.1B-Chat-v1.0",
```
yechank-nvidia, line 6: Adding "" on the model name. Without "", it gives an error.

tensorrt_llm/commands/serve.py
```diff
@@ -156,6 +158,10 @@ def launch_server(host: str, port: int, llm_args: dict):
               default=0.9,
               help="Free GPU memory fraction reserved for KV Cache, "
               "after allocating model weights and buffers.")
+@click.option("--disable_block_reuse",
```
symphonylyh: is this something can be controlled by `--extra_llm_api_options`? I'm a little worried if we have more divergence between multimodal and llm features, it's not very sustainable to add --disable_A, --disable_B one by one

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -312,7 +312,7 @@ def __call__(
         sampling_params: SamplingParams,
     ) -> Tuple[List[int], Optional[ExtraProcessedInputs]]:
         text_prompt, mm_data, mm_processor_kwargs = inputs.get("prompt"), \
-                        inputs.get("multi_modal_data"), inputs.get("mm_processor_kwargs", {})
+                        inputs.get("multi_modal_data", {}), inputs.get("mm_processor_kwargs", {})
```
symphonylyh, line 315: since this looks safe, how about change in all models so later on you can turn on trtllm-serve for all?

tensorrt_llm/commands/serve.py
symphonylyh: also did we lag behind on the `trtllm-serve` doc args https://nvidia.github.io/TensorRT-LLM/commands/trtllm-serve.html? @kaiyux which file controls the doc?

symphonylyh (approved):

tensorrt_llm/commands/serve.py
yechank-nvidia: It can be addressed with `--extra_llm_api_options`, so removed `--disable_block_reuse` and added an instruction on how trtllm-serve can be compatible with multimodal.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Added on other models, VILA, LLava-Next.

examples/serve/curl_chat_client.sh
LinPoly: Thanks for pointing out, I will add one assertion against the status code to the test when refactoring trtllm-serve CI.

examples/serve/README.md
```diff
@@ -1,3 +1,72 @@
 # Online Serving Examples with `trtllm-serve`
 
 We provide a CLI command, `trtllm-serve`, to launch a FastAPI server compatible with OpenAI APIs, here are some client examples to query the server, you can check the source code here or refer to the [command documentation](https://nvidia.github.io/TensorRT-LLM/commands/trtllm-serve.html) and [examples](https://nvidia.github.io/TensorRT-LLM/examples/trtllm_serve_examples.html) for detailed information and usage guidelines.
+
+## LLM Serving Example
```
LinPoly: Thanks for documenting, but we place a README here for redirecting users to our official doc, can you add the doc for multi-modality support to https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/commands/trtllm-serve.rst?plain=1 instead of here? Or I can commit to this PR after we finish feature review.

tensorrt_llm/commands/serve.py
```diff
@@ -33,6 +33,7 @@ def get_llm_args(model: str,
                  free_gpu_memory_fraction: Optional[float] = None,
                  num_postprocess_workers: int = 0,
                  trust_remote_code: bool = False,
+                 enable_block_reuse: bool = True,
```
LinPoly: Is this positional argument necessary? I think we don't give it a value when calling `get_llm_args` and it can be control with `echo kv_cache_config:\n  enable_block_reuse: true > ./extra-llm-api-config.yml`.

tests/unittest/llmapi/apps/_test_trtllm_serve_example.py
```diff
@@ -10,16 +10,39 @@
...
+        "LLM": "TinyLlama-1.1B-Chat",
+        "Multimodal": "Qwen2-VL-7B-Instruct"
+    }
+    return model_map[request.param]
 
 
 @pytest.fixture(scope="module")
 def server(model_name: str):
-    model_path = get_model_path(model_name)
+    model_path_map = {
```
LinPoly: Would it be better if we create another file for multimodal example? The implementation here is kind of quirky and we need two model loading by all means, creating another file won't increase test time by much.

LinPoly (approved):
Thanks for improving, besides these comments, we also need another E2E test in `tests/unittest/llmapi/apps`, to validate the output thoroughly.

examples/serve/README.md
symphonylyh: do you also happen to know what's the source file of this doc  https://nvidia.github.io/TensorRT-LLM/commands/trtllm-serve.html? It's outdated and good to update too

tensorrt_llm/commands/serve.py
symphonylyh: great. I think Pengyun has another comment here https://github.com/NVIDIA/TensorRT-LLM/pull/3590#discussion_r2046923967. I'll resolve mine and we can continue there. I think he means only multimodal needs to input this field by `extra-llm-api-config.yml` and existing models just don't need any changes

tensorrt_llm/commands/serve.py
yechank-nvidia: Removed. Thx.

examples/serve/README.md
yechank-nvidia: Moved to trtllm-serve.rst, but not sure about trtllm-serve.html's location as @symphonylyh mentioned.

tests/unittest/llmapi/apps/_test_trtllm_serve_example.py
yechank-nvidia: Created separate test file for multimodal. Thx.

tensorrt_llm/commands/serve.py
kaiyux: `docs/source/commands/trtllm-serve.rst`, currently the doc only gets updated on official releases, so for now it's only on v0.18.2.

kaiyux:
@yechank-nvidia @symphonylyh FYI - to review the compiled doc pages, you can go to the "A10-Build_TRT-LLM_Doc" stage in pipeline, download the artifacts, and open locally using browser to review.

examples/serve/README.md
kaiyux: The source file is `docs/source/commands/trtllm-serve.rst`, currently the doc only gets updated on official releases, so for now it's only on v0.18.2.

PR #3597, opened 2025-04-16
feat: add etcd dependency and interface

Shunkangz:
The etcd will be added in dockerfile in MR https://github.com/NVIDIA/TensorRT-LLM/pull/3616

### DIFF
--- a/tensorrt_llm/serve/metadata_server.py
+++ b/tensorrt_llm/serve/metadata_server.py
@@ -1,0 +1,69 @@
+import json
+from abc import ABC, abstractmethod
+
+import etcd3
+
+
+class RemoteDictionary(ABC):
+
+    @abstractmethod
+    def get(self, key: str) -> str:
+        pass
+
+    @abstractmethod
+    def put(self, key: str, value: str):
+        pass
+
+    @abstractmethod
+    def remove(self, key: str):
+        pass
+
+    @abstractmethod
+    def keys(self) -> list[str]:
+        pass
+
+
+class EtcdDictionary(RemoteDictionary):
+
+    def __init__(self, host: str, port: int):
+        self._client = etcd3.client(host, port)
+
+    def get(self, key: str) -> str:
+        return self._client.get(key)
+
+    def put(self, key: str, value: str):
+        self._client.put(key, value)
+
+    def remove(self, key: str):
+        self._client.delete(key)
+
+    def keys(self) -> list[str]:
+        # TODO: Confirm the final save key format
+        # This implementation assumes that key is in the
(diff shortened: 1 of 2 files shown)


docker/common/install_etcd.sh
```diff
@@ -0,0 +1,19 @@
...
+
+ETCD_VER=v3.5.21
+
+# choose either URL
+DOWNLOAD_URL=https://storage.googleapis.com/etcd
+
+rm -f /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz
+rm -rf /tmp/etcd-download-test && mkdir -p /tmp/etcd-download-test
+
+curl -L ${DOWNLOAD_URL}/${ETCD_VER}/etcd-${ETCD_VER}-linux-amd64.tar.gz -o /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz
```
Shixiaowei02: We need to consider the situation with ARM here.

docker/Dockerfile.multi
```diff
@@ -57,6 +57,10 @@ RUN bash ./install_polygraphy.sh && rm install_polygraphy.sh
 COPY docker/common/install_mpi4py.sh install_mpi4py.sh
 RUN bash ./install_mpi4py.sh && rm install_mpi4py.sh
 
+# Install etcd
+COPY docker/common/install_etcd.sh install_etcd.sh
```
Shixiaowei02: Could you please contact @ZhanruiSunCh to update the Docker image?

requirements.txt
```diff
@@ -48,3 +48,4 @@ flashinfer-python~=0.2.3
 opencv-python-headless
 xgrammar==0.1.16
 backoff
+etcd3
```
Shixiaowei02: Please merge the Docker image upgrade as a separate PR first, and put the changes to `requirements.txt` in the next PR. Thanks!

docker/common/install_etcd.sh
Shunkangz: Done

tensorrt_llm/serve/metadata_server.py
kaiyux, line 1: Sorry I should've missed some context - why is this file needed and where is it going to be used? Should we wait until there are changes that actually use those data structures?

Shunkangz: After discussion with @Shixiaowei02 , we plan to split the support of dynamic scaling into a few small PRs. In this PR, we want to align the interface of etcd first and there will be another MR depends on this. I will submit it soon.

PR #3600, opened 2025-04-16
fix: fix cublas_scaled_mm

dc3671:
In pytorch 2.7, cublas's workspace is modified from 32MB to 1MB on Hopper: 

- Use `_getWorkspaceSize()` rather than `_getWorkspace()` https://github.com/pytorch/pytorch/commit/203a27e0cecce5b9050218c9d6a56c8cd2ebd0a5#diff-74fcb26047c1df4024105d36ce22a36b77cf8cc93c28631d743e639b3d6066aeL1624-R1599

```diff
-   auto stream = c10::cuda::getCurrentCUDAStream();
-   size_t workspaceSize = 0;
-   auto workspace_ptr = _getWorkspace(workspaceSize);
+   size_t workspaceSize = _getWorkspaceSize();
+   auto workspace = at::empty(static_cast<int64_t>(workspaceSize),    at::TensorOptions().dtype(at::kByte).device(at::kCUDA));
```

- And in `_getWorkspace()` will use cublas's workspace size 32MB on Hopper: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/CublasHandlePool.cpp#L133

```cpp
const size_t default_size = sm90 ? 4096 * 8 * 1024 : 4096 * 1024 * 2 + 16 * 1024 * 8;
```

And this lead to different algos (especially when one choose splitK and another not) chosen by cublasLt on some GEMM shapes. So we need to align this in our tests.

In E2E test, there are other tests called torch._scaled_mm, and the workspace size variable is static in cpp code, so the modification in test_scaled_mm won't take effect and we need to add that in test_linear_fp8.py.

### DIFF
--- a/tests/unittest/_torch/thop/test_scaled_mm.py
+++ b/tests/unittest/_torch/thop/test_scaled_mm.py
@@ -38,11 +38,6 @@ from utils.util import getSMVersion
     [torch.float16, torch.float32, torch.bfloat16],
 )
 def test_fp8_scaled_mm(output_dtype, m, k_n):
-    pytest.skip("https://nvbugspro.nvidia.com/bug/5228279")
-    # Skip specific problematic case
-    if m == 228 and k_n == (28672, 8192):
-        pytest.skip("Skipping problematic case with m=228, k=28672, n=8192")
-
     k, n = k_n
     torch.random.manual_seed(0)
     shape_x = (m, k)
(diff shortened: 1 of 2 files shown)


dc3671:
/bot run --disable-fail-fast --stage-list "H100_PCIe-PyTorch-1"

yiqingy0:
/bot run --disable-fail-fast --stage-list "H100_PCIe-PyTorch-1"

dc3671:
/bot run --disable-fail-fast --stage-list "H100_PCIe-PyTorch-1"

dc3671:
/bot run --disable-fail-fast --stage-list "H100_PCIe-PyTorch-1"

dc3671:
/bot run --disable-fail-fast --stage-list "H100_PCIe-PyTorch-1"

dc3671:
/bot run --disable-fail-fast --stage-list "H100_PCIe-PyTorch-1"

tests/unittest/_torch/thop/test_scaled_mm.py
```diff
@@ -60,8 +55,10 @@ def test_fp8_scaled_mm(output_dtype, m, k_n):
         out_dtype=output_dtype,
     )
     # set pytorch's cublas workspace size to 32MB to be aligned with trtllm
+    # if anywhere else calls torch's cublas op, the static workspace size will
```
hlu1: Can you explain why the test was failing before?

dc3671: In pipeline test, the test in test_fp8_linear will run first and the workspace size will be set to 1MB, which is a static value [here](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/CUDABlas.cpp#L220). For we don't launch new process, this value will also be applied to this test.

hlu1: Can you add a comment on the splitk thing?

hlu1 (approved):

tests/unittest/_torch/thop/test_scaled_mm.py
dc3671: Sure, done.

PR #3610, opened 2025-04-16
feat: Support unfused rope in MLA.

yuxianq:
This PR helps us integrate other attention backend that supports MLA (like flashinfer) but does not support rope fusion.


litaotju (approved):

zhhuang-nv:
Do we have any L0 CI test for this code path?

yuxianq:
> Do we have any L0 CI test for this code path?

Not for now since we don't have other MLA backends now. It is a part of an effort to integrate FlashInfer SM90 MLA kernels into TRT-LLM (see https://gitlab-master.nvidia.com/ftp/tekit/-/merge_requests/8856), but the priority of this task is low now that we already have FlashMLA. So I extract the common improvements of this MR into a series of PRs to merge first. We may have an e2e test when the FlashInfer MLA MR is finished.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -680,6 +712,8 @@ def _run_bmm():
                 raise NotImplementedError(
                     f"Missing bmm impl for dtype: {self.k_b_proj_trans.dtype}.")
 
+        if not self.enable_rope_fusion:
```
zhhuang-nv: Is there any indent problem here?

yuxianq: Nice catch, Fixed.

zhhuang-nv (approved):

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -504,6 +504,25 @@ def yarn_get_mscale(scale=1, mscale=1):
...
+        self.support_unfused_qkv = not isinstance(self.mha, TrtllmAttention)
+        self.rotary_emb = None
+        if not self.enable_rope_fusion:
+            self.rotary_emb = RotaryEmbedding(
+                pos_embd_params.rope,
+                head_dim=self.qk_rope_head_dim,
+                is_neox=pos_embd_params.is_neox,
+            )
+
+    def apply_rope(self, q, k_pe, position_ids):
```
hlu1: Add type info for inputs/output

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -544,6 +563,10 @@ def forward(
             q_ctx = q[:num_ctx_tokens, ...]
             compressed_kv_ctx = compressed_kv[:num_ctx_tokens, ...]
             k_pe_ctx = k_pe[:num_ctx_tokens, ...]
+            if self.rotary_emb is not None and position_ids is not None:
```
hlu1:
```suggestion
            if apply_rotary_emb:
```
Add a bool for `apply_rotary_emb = not self.enable_rope_fusion` and a check before line 562
```
if apply_rotary_emb:
    assert self.rotary_emb is not None and position_ids is not None
```

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -554,6 +577,10 @@ def forward(
             q_gen = q[num_ctx_tokens:, ...]
             compressed_kv_gen = compressed_kv[num_ctx_tokens:, ...]
             k_pe_gen = k_pe[num_ctx_tokens:, ...]
+            if self.rotary_emb is not None and position_ids is not None:
```
hlu1:
```suggestion
            if apply_rotary_emb:
```

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -504,6 +504,25 @@ def yarn_get_mscale(scale=1, mscale=1):
...
+        if not self.enable_rope_fusion:
+            self.rotary_emb = RotaryEmbedding(
+                pos_embd_params.rope,
+                head_dim=self.qk_rope_head_dim,
+                is_neox=pos_embd_params.is_neox,
+            )
+
+    def apply_rope(self, q, k_pe, position_ids):
+        q_pe = q[..., self.qk_nope_head_dim:].reshape(
+            -1, self.num_heads * self.qk_rope_head_dim)
```
hlu1, line 529: This reshape incurs a copy internally. Just FYI.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -587,6 +614,12 @@ def forward(
                                   all_reduce_params=all_reduce_params)
         return attn_output
 
+    def convert_qkv(self, q, k, v):
```
hlu1:
```suggestion
    def _maybe_concat_qkv(self, q, k, v):
```
Rename

tensorrt_llm/_torch/modules/attention.py
yuxianq: Yes, flashinfer's `apply_rope_with_cos_sin_cache_inplace` requires the `q` shape `(nnz, num_q_heads * head_size)`, but the last dimension of `q` before reshaped is not contiguous, so we need a copy here.

tensorrt_llm/_torch/modules/attention.py
yuxianq: Added.

tensorrt_llm/_torch/modules/attention.py
yuxianq: Updated. Since we always create `self.rotary_emb` when `apply_rotary_emb` is True, I only add `assert position_ids is not None`.

tensorrt_llm/_torch/modules/attention.py
yuxianq: Updated.

tensorrt_llm/_torch/modules/attention.py
yuxianq: Renamed.

PR #3618, opened 2025-04-16
test: add INTEGRATION_TEST env var to speed up integration test

crazydemo:
set `INTEGRATION_TEST=1` to only run one batch in accuracy test

### DIFF
--- a/examples/summarize.py
+++ b/examples/summarize.py
@@ -42,6 +42,16 @@ from prompt_lookup.run_dtm_pld import run_dtm_pld
 
 
 def main(args):
+    is_integration_test = os.getenv('INTEGRATION_TEST', '0') == '1'
+    if is_integration_test:
+        logger.info(
+            "Running in integration test mode - will only run one batch and skip accuracy checks"
+        )
+        logger.info(
+            "Setting max_ite=1 and check_accuracy=False for integration test")
+        args.max_ite = 1
+        args.check_accuracy = False
+
     runtime_rank = tensorrt_llm.mpi_rank()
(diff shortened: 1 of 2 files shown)


kxdc (approved):

syuoni:
Hi @crazydemo , could you please explain the purpose of `INTEGRATION_TEST`? When should we use it?

crazydemo:
> Hi @crazydemo , could you please explain the purpose of `INTEGRATION_TEST`? When should we use it?

This is for TensorRT-LLM and TensorRT integration test. This integration test will run all the TRT flow test cases to ensure the integration's functionality. I would like to set this `INTEGRATION_TEST` env var to skip all the accuracy check, so as to speed up the integration test process. When the env var is set as `1`, all the trt flow test cases only run one batch data, instead of going through the entire dataset.

syuoni:
> > Hi @crazydemo , could you please explain the purpose of `INTEGRATION_TEST`? When should we use it?
> 
> This is for TensorRT-LLM and TensorRT integration test. This integration test will run all the TRT flow test cases to ensure the integration's functionality. I would like to set this `INTEGRATION_TEST` env var to skip all the accuracy check, so as to speed up the integration test process. When the env var is set as `1`, all the trt flow test cases only run one batch data, instead of going through the entire dataset.

So this is used from the TensorRT side, right?

crazydemo:
Yes, TensorRT's QA team will do the integration test.
> So this is used from the TensorRT side, right?

syuoni (approved):

LarryXFly (approved):

PR #3621, opened 2025-04-16
Fix: [NVBUG 5201530]Fix merge dummy request when DP + Overlap + Cuda graph.

Tracin:


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1290,12 +1290,13 @@ def _merge_tp_requests(self, new_requests: List[ExecutorRequest]):
...
+        has_context_request = len(self.active_requests) != 0
         llm_request_list = self.kv_cache_manager.add_dummy_requests(
             request_ids=list(range(num_dummy_request)),
-            is_gen=not self.has_context_request,
-            prepare_resource=not self.has_context_request,
+            is_gen=not has_context_request,
+            prepare_resource=not has_context_request,
             max_num_draft_tokens=0
-            if self.has_context_request else self.max_draft_tokens,
+            if has_context_request else self.max_draft_tokens,
```
QiJune: So the new logic is that:

1. If there is active_requests, always add a dummy generation request
2. If there is no active requests, add a dummy context request

Is my understand correct?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1290,12 +1290,13 @@ def _merge_tp_requests(self, new_requests: List[ExecutorRequest]):
         return False
 
     def _merge_dummy_request(self, num_dummy_request: int):
+        has_context_request = len(self.active_requests) != 0
         llm_request_list = self.kv_cache_manager.add_dummy_requests(
             request_ids=list(range(num_dummy_request)),
-            is_gen=not self.has_context_request,
-            prepare_resource=not self.has_context_request,
```
QiJune: It seems that `self.has_context_request`  is useless after this change. Should we remove `self.has_context_requests` related codes entirely?

tensorrt_llm/_torch/pyexecutor/py_executor.py
Tracin: That a little bit weird. My change can fix the wrong version. But I just found TOT is good without my fix. So my change is likely not the real root cause. I will un-waive those cases first and investigate if any commits really fix it. cc @jinyangyuan-nvidia

Tracin:
After some investigation. I think this commit accidently fix the issue https://gitlab-master.nvidia.com/ftp/tekit/-/commit/3041bbdab39d34cfa64f23c77274ee0f68ec7dce.
And I think those two copy operation causes the accuracy issue in previous version. cc @jinyangyuan-nvidia @QiJune

litaotju (approved):

PR #3630, opened 2025-04-16
Fix: nvbugs/5222698 variable not defined

zongfeijing:


Kefeng-Duan (approved):

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -491,7 +491,7 @@ def forward(
 
         min_latency_mode = True if hidden_states.size(
             0
-        ) <= 128 and self.fusion_config.POST_MOE_FUSION and self.is_nvfp4 else False
+        ) <= 128 and self.fusion_config.POST_MOE_FUSION and self.is_nvfp4 and not using_prev_fusion else False
```
hyukn: directly this:
min_latency_mode = hidden_states.size() <= 128 and self.fusion_config.POST_MOE_FUSION and self.is_nvfp4 and not using_prev_fusion

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -510,6 +510,7 @@ def forward(
         )
 
         if self.fusion_config.PRE_MOE_FUSION:
+            hidden_states_fp4 = None
```
hyukn: The changes on `min_latency_mode` can cover this, right?
I think mlp (either gatedmlp or moe) can not handle a None input.

tensorrt_llm/_torch/models/modeling_deepseekv3.py
zongfeijing: yes, it's a better way. Thanks.

tensorrt_llm/_torch/models/modeling_deepseekv3.py
zongfeijing: Yes, `hidden_states_fp4` is only used in `min_latency_mode`, and this variable will not be passed to `gatedmlp` module, it's only used in `fused_moe` module.

litaotju (approved):

hyukn (approved):

PR #3631, opened 2025-04-16
fix: 5197419 and removed unused runtime kernels

jdebache:


jdebache:
@Funatiq : ref. our internal chat, the kernel in question was needed it seems, as removing its call causes test failures for beam search.

I guarded the flaky test in some additional device syncs and prevented it from running on tiny devices.

Also, while attempting to remove the kernel, I noticed some others were unused, and I removed them.

Funatiq:
I only see failing beam search tests in the decoder unit tests where we explicitly check that the input tokens are set for each beam (that's what the kernel does that we wanted to remove). The e2e tests (A30-CPP-1) were all passing, right? So I still think it would be fine to remove the kernel.

jdebache:
@Funatiq it seems you were right. I adjusted the tests to only check for the input token ID in the first beam, and the end-to-end tests still pass.

Funatiq (approved):

PR #3593, opened 2025-04-16
test: Get Eagle tests working

brb-nv:
Dummy Eagle head creation for Eagle tests seems to be broken - likely because of updates to how KV cache quantization is done in ModelOpt. This MR fixes that.

I verified the following:
- examples/test_eagle.py::test_codellama_eagle_1gpu
- examples/test_eagle.py::test_llama_eagle_1gpu
- examples/test_eagle.py::test_mistral_eagle_1gpu

Additionally, it also addressed an Eagle issue related to Mistral Nemo 12B.

PR #3633, opened 2025-04-16
fix: read and set head_size and head_dim when convert Eagle checkpoint

yweng0828:
Add set `head_size` and `head_dim` when converting the Eagle checkpoint. 

This can make the `head_size` and `head_dim` of the main model and eagle model decoupled and their value are unnecessarily equal to `hidden_size / num_attention_heads` (such as mistral-nemo).


brb-nv (approved):
Changes look good to me.

yweng0828:
Cherry-pick to https://github.com/NVIDIA/TensorRT-LLM/pull/3593, close.

PR #3638, opened 2025-04-16
feat: [AutoDeploy] update rope matcher with minor variants (Deepseek)

Fridah-nv:
- [X] Update doc to summarize variants of RoPE
- [X] Support DeepSeek variants of RoPE: custom op, matcher, unit test
- [X] E2E test for DS
- [X] separate transformation matching rope input shape to one of {BSND, BNSD}

TODO:
DS RoPE optimization op and transformation


tensorrt_llm/_torch/auto_deploy/custom_ops/torch_attention.py
```diff
@@ -228,6 +228,33 @@ def apply_rotary_pos_emb_ds(q, k, cos, sin, position_ids, unsqueeze_dim=1):
...
+@torch.library.custom_op("rope::apply_rope_ds", mutates_args=())
+def apply_rope_ds(
+    q: torch.Tensor,
+    k: torch.Tensor,
+    cos: torch.Tensor,  # [B, 1, seq_len, head_dim]
+    sin: torch.Tensor,  # [B, 1, seq_len, head_dim]
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    """
+    DS-style RoPE: interleaves Q/K channels, indexes cos/sin by position_ids
+    and returns rotated (q_embed, k_embed).
```
lucaslie: There is no cos/sin indexing here with position_ids since it's done outside the op. Please update the comment accordingly

tensorrt_llm/_torch/auto_deploy/custom_ops/torch_attention.py
```diff
@@ -228,6 +228,33 @@ def apply_rotary_pos_emb_ds(q, k, cos, sin, position_ids, unsqueeze_dim=1):
     return q_embed, k_embed
```
lucaslie: Can we get rid of this `apply_rotary_pos_emb_ds` function here and use `apply_rope_ds` instead?

tensorrt_llm/_torch/auto_deploy/custom_ops/torch_attention.py
```diff
@@ -228,6 +228,33 @@ def apply_rotary_pos_emb_ds(q, k, cos, sin, position_ids, unsqueeze_dim=1):
...
 
 
+@torch.library.custom_op("rope::apply_rope_ds", mutates_args=())
+def apply_rope_ds(
```
lucaslie: Let's give it a more descriptive name like `apply_rope_with_qk_interleaving`

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,6 +1,53 @@
...
+   - Input layout: non-interleaved, [B, N, S, D] with unsqueeze_dim=1 and
+        [B, S, N, D] with unsqueeze_dim=2, default is [B, N, S, D]
+   - Frequencies are provided as separate `cos` and `sin` tensors of shape [B, S, head_dim].
+   - Source code:
+        def rotate_half(x):
+            x1 = x[..., : x.shape[-1] // 2]
+            x2 = x[..., x.shape[-1] // 2 :]
+            return torch.cat((-x2, x1), dim=-1)
+
+        def apply_rotary_pos_emb(q, k, cos, sin, position_ids=None, unsqueeze_dim=1):
```
lucaslie:
```suggestion
        def apply_rotary_pos_emb(q, k, cos, sin, unsqueeze_dim=1):
```

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,6 +1,53 @@
+"""
```
lucaslie, line 1: Fantastic description! Thank you. This is going to be a super important reference moving forward

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
lucaslie: Let's add a custom op for this in `torch_rope.py` that is just reproducing the torch implementation similar to `apply_rope_ds`

tensorrt_llm/_torch/auto_deploy/custom_ops/torch_attention.py
lucaslie, line 1: I think it's time for us to "graduate" rope functions into their own file. Let's start a new file called `torch_rope.py` and put all rope-related implementations there.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,6 +1,53 @@
...
+            sin = sin.unsqueeze(unsqueeze_dim)
+            q_embed = (q * cos) + (rotate_half(q) * sin)
+            k_embed = (k * cos) + (rotate_half(k) * sin)
+            return q_embed, k_embed
+
+2. Complex Multiplication (GPTJ/Llama-stack-style, interleaved)
+   - Input layout: [B, S, N, D] (interleaved)
+   - Frequencies are combined into a single complex-valued tensor `freqs_cis` of shape [B, S, head_dim // 2].
+   - Source code:
+        def apply_rotary_emb(
```
lucaslie, line 28: Same here. Let's add a custom op to `torch_rope.py`

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
lucaslie, line 1: I would like to suggest to also split each transformation here into two transformations akin of your fused_moe and my unfused attention PR. Here is my suggestion:

1. Pattern matcher that tries to pattern-match the rope pattern and replaces it with the reference custom op defend in `torch_rope.py` (pure pytorch implementation, no optimized kernel)
2. Perf-focused graph transformation that picks up on the torch reference op and replaces it with the optimized flashinfer op and does the `cos`/`sin` fusion and caching of the fused node to retrieve for other optimized ops (same as is already done in 2nd for each of the transformations here)

The rationale behind this is the same as for fused_moe:

Transformation (1) may fail to detect the pattern or the user has a new pattern that we don't support yet. In this case we should make it very easy for the user to have an escape hatch, i.e., use a custom op that is as close to eager torch and requires the littlest model rewrite 

Transformation (2) should always work and takes care of the tedious work of rewriting the graph/model in a way that is compatible with the optimized op/kernel. Moreover, in (2) we might want to have the flexibility of targeting different kernels. We don't want this to interfere with the pattern matching in (1)

tensorrt_llm/_torch/auto_deploy/custom_ops/torch_attention.py
Fridah-nv: @sugunav14 Could you comment on why `apply_rotary_pos_emb_ds` uses `q = q.unflatten(-1, (-1, 2)).transpose(-1, -2).reshape_as(q)` instead of `q = q.view(b, h, s, d // 2, 2).transpose(4, 3).reshape(b, h, s, d)` in original code? I believe they have same effect, is there any performance reason?

lucaslie: @sugunav14, please confirm if my memory is correct

sugunav14: Yeah, that's correct

Fridah-nv: Gotcha, thank you both!

Fridah-nv: Update the rope implementation in `fused_flattened_mla_with_cache` with the custom op. Keep this `apply_rotary_pos_emb_ds` as it's used by `fused_mla` and `fused_mla_ref`. They seem to be torch reference implementation of 'mla without kv cache' and 'mla with flatten input and kvcache'. Please correct me if I'm wrong.

Fridah-nv:
```
q = q.unflatten(-1, (-1, 2)).transpose(-1, -2).reshape_as(q)
```
Would be a requirement for `fused_flattened_mla_with_cache` as input is 3D or 4D

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: Finish separating the two major version of RoPEs. I left DS RoPE still in one transformation, I can separate it into 2 transformations after we find a kernel for DS RoPE.
Also, the transpose of input shape is postponed into the optimization transformation, should run the `eliminate_transpose` again.

lucaslie: Sumary of our conversation. We will add one more transformation here:

1. transformation to match the desired attention layout ("bsnd" or "bnsd") in the rope kernel
2. only replace torch reference op with optimized op when layout matches with the layout expected by the optimized kernel

tensorrt_llm/_torch/auto_deploy/custom_ops/torch_rope.py
```diff
@@ -0,0 +1,205 @@
...
+def rotate_half(x):
+    """Rotates half the hidden dims of the input."""
+    x1 = x[..., : x.shape[-1] // 2]
+    x2 = x[..., x.shape[-1] // 2 :]
+    return torch.cat((-x2, x1), dim=-1)
+
+
+# Copied from https://huggingface.co/deepseek-ai/DeepSeek-V3/blob/main/modeling_deepseek.py#L339
+@torch.inference_mode()
+def apply_rotary_pos_emb_ds(q, k, cos, sin, position_ids, unsqueeze_dim=1):
```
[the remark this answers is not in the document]
Fridah-nv: It's doable. I just feel that it would make the reference implementations look less straight forward. 
I think `torch_apply_rope_with_qk_interleaving` is something between a reference op and an optimized op (there's simple optimization like caching cos/sin), and that's why I'm hesitate about replacing `apply_rotary_pos_emb_ds` with it.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: Just committed an update:
1. The flash_infer op expect input to be [B*S, N*D] and the warper custom op around it expects [B, S, N, D]. For the two main RoPE version we support: `torch_apply_explicit_rope` accepts both [B,S,N,D] and [B,N,S,D], `torch_apply_complex_rope` only accept [B,S,N,D]
2. I added a transformation `match_rope_layout` that transforms layout for `torch_apply_explicit_rope` if it is [B,N,S,D]. I don't see there's a need to add support for [B,S,N,D] -> [B,N,S,D] for now. Let me know your thoughts on this. Also the optimize transformation will check if input is [B,S,N,D], if not, it will throw a warning
3. The flash_infer op requires cos_sin_cache to be in float32. (My memory was wrong). Currently casting the cos_sin_cache node to fp32 in the transformation to ensure this

lucaslie:
```
def match_rope_with_explicit_cos_sin(....):
  # detect and replace with apply_rope_with_explicit_cos_sin
  ...

   # check if the input to apply_rope_with_explicit_cos_sin contains q, k interleaving
   if no_interleaving:
      return
   else:
       # replace with apply_rope_with_qk_interleaving
```

lucaslie:
if you add `unsqueeze_dim` to both `apply_rope_with_explicit_cos_sin` and `apply_rope_with_qk_interleaved` the layout matching transformation can operator on both in the same fashion

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
```diff
@@ -24,7 +24,7 @@ def apply_rope_with_input_pos_flashinfer(
         Precomputed tensor of positional indices; it is shared across calls in the graph.
     - cos_sin_cache (torch.Tensor):
         Precomputed fused tensor created by concatenating the first half of the cosine and sine
-        components derived from the inv_freq.
+        components derived from the inv_freq. Shape [seq_len, head_dim]. Must be float32.
```
lucaslie, line 27: You could make it type `torch.FloatTensor` instead of saying "must be float32"

tensorrt_llm/_torch/auto_deploy/custom_ops/torch_rope.py
```diff
@@ -0,0 +1,187 @@
...
+    q_embed = (q * cos) + (rotate_half(q) * sin)
+    k_embed = (k * cos) + (rotate_half(k) * sin)
+    return q_embed, k_embed
+
+
+@torch_apply_rope_with_explicit_cos_sin.register_fake
+def torch_apply_rope_with_explicit_cos_sin_fake(
+    q: torch.Tensor, k: torch.Tensor, cos: torch.Tensor, sin: torch.Tensor, unsqueeze_dim: int = 1
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    return q, k
```
lucaslie, line 125:
```suggestion
    return torch.empty_like(q), torch.empty_like(k)
```

It should not reference the same tensor to be consistent with the default implementation. Having the same tensor reference has implications on graph tracing and hence we wanna be consistent

lucaslie: For example, in graph tracing something like `q.contiguous()` would not appear in the graph if `q` is already contiguous and hence `q.contiguous()` would just return a reference to the same tensor

tensorrt_llm/_torch/auto_deploy/utils/node_utils.py
```diff
@@ -307,3 +308,22 @@ def bfs(
                 visited.add(next_node)
                 queue.append(next_node)
     raise RuntimeError(f"Could not find node with target condition {target}.")
+
+
+def extract_output_tuple(node: Node, count: int = 2):
```
lucaslie, line 313: nice one! This is so annoying to deal with lol

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
lucaslie: I don't understand why `torch_apply_complex_rope` only supports [B,S,N,D]. Looks to me like `unsqueeze_dim` can be used for this just like in the other cases?

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -6,61 +53,70 @@
...
+def match_explicit_rope(gm: GraphModule) -> GraphModule:
     """
     Identify and replace legacy RoPE subgraphs (explicit cos/sin multiplication pattern):
 
-      output = (raw * unsqueeze(cos)) + (rotate_half(raw) * unsqueeze(sin))
+      - HF style: output = (raw * unsqueeze(cos)) + (rotate_half(raw) * unsqueeze(sin))
+      - DS-style: requires interleaving Q/K before the cos/sin mul
 
     If exactly two such branches (query and key) are detected within each region, they're replaced
     by a call to `torch.ops.rope.flashinfer`.
```
lucaslie, line 68: This comment looks outdated

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -69,47 +125,324 @@ def match_rope_v2(gm: GraphModule) -> GraphModule:
...
```
lucaslie, line 126: outdated comment?

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -69,47 +125,324 @@ def match_rope_v2(gm: GraphModule) -> GraphModule:
...
+
+    need_transpose = False
+    for node in list(graph.nodes):
+        if is_op(node, rope_ops):
+            if is_op(node, torch.ops.rope.torch_apply_rope_with_complex_freqs):
+                q_node, k_node, freqs_node, *rest = node.args
+                unsq = rest[0] if rest else 2
+            else:
+                q_node, k_node, cos_node, sin_node, *rest = node.args
+                unsq = rest[0] if rest else 1
```
lucaslie, line 190: So I think we should assume that the unsqueeze_dim exists. We should not infer a default here. Here is some notes on how args are represented

* Torch.export will make sure that every arg including unprovided args with default values are provided to the `node.args` field to avoid confusion
* if it's a kw-only arg it will be in `node.kwargs`
* When we insert ops like `torch.ops.rope.torch_apply_rope_with_complex_freqs` we should follow this convention as well
* Then there is no need to assume default values explicitly here

Here is a code snippet from inserting attention in my unfused attention PR doing the same thing: https://github.com/nv-auto-deploy/TensorRT-LLM/blob/68189b80e97716bc8b01c1c99f0daa9dd171aa29/tensorrt_llm/_torch/auto_deploy/transformations/library/attention.py#L508-L517

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -69,47 +125,324 @@ def match_rope_v2(gm: GraphModule) -> GraphModule:
...
+    gm = canonicalize_graph(gm)
+    return gm
+
+
+def match_rope_layout(gm: GraphModule, layout: str = "bsnd") -> GraphModule:
+    """
+    Match and transform input and output of rope ops to the layout specified.
+    Supported layout is 'bsnd' (batch, seq, head, dim).
+    """
+    supported = "bsnd"
```
lucaslie, line 166: I think we should easily be able to support `bsnd` and `bnsd` here. I don't see an issue. 

The transpose operation is always the same `.transpose(1,2)` and all you have to do below is update your check to something like

```
if unsq==2:
  current_layout = "bsnd"
elif unsq==1:
  current_layout="bsnd"
else:
    # can't match
    continue 
```

and then do something like
```
need_transpose = expected_layout != current_layout
```

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -69,47 +125,324 @@ def match_rope_v2(gm: GraphModule) -> GraphModule:
...
-            k_match,
-            rope_flash_cache,
-            rope_position_ids_cache,
+        _process_complex_rope(graph, q_match, k_match)
+
+    gm = canonicalize_graph(gm)
+    return gm
+
+
+def match_rope_layout(gm: GraphModule, layout: str = "bsnd") -> GraphModule:
```
lucaslie, line 161:
```suggestion
def match_rope_layout(gm: GraphModule, expected_layout: str = "bsnd") -> GraphModule:
```

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: The comment was outdated, sorry about the confusion. `torch_apply_complex_rope` supports both layouts now

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
Fridah-nv: This is a good suggestion, but Torch custom op seems not supporting `torch.FloatTensor`. I ran into an error when switching to `torch.FloatTensor`
Here's the supported ArgType lists I found:
https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/README.md#func

tensorrt_llm/_torch/auto_deploy/custom_ops/torch_rope.py
Fridah-nv: I see, thanks for explaining.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: > Torch.export will make sure that every arg including unprovided args with default values are provided to the node.args field to avoid confusion

This is not the case with `torch.ops.rope.torch_apply_rope_with_complex_freqs` and other torch custom rope ops when I tested them. I cannot get the `unsqueeze_dim` argument from `rope_node.args` if it is the same as the default value. `unsqueeze_dim` is not included even if I provide it explicitly when initializing the custom op.

> Here is a code snippet from inserting attention in my unfused attention PR doing the same thing:

This is helpful. I updated the code to get default value of `unsqueeze_dim` from `op._schema.arguments`.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: Included the change and updated unit tests. Thanks for the suggestion!

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
lucaslie: >I cannot get the unsqueeze_dim argument from rope_node.args if it is the same as the default value. unsqueeze_dim is not included even if I provide it explicitly when initializing the custom op.

That's interesting. I didn't know that. Thank you!

>This is helpful. I updated the code to get default value of unsqueeze_dim from op._schema.arguments.

Cool! Then we have a solution that works

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -491,22 +814,17 @@ def _get_position_ids(
...
     device = q_node.meta.get("device", "cpu")
     if isinstance(device, str):
         device = torch.device(device)
 
-    # Build positions: arange(sym_seq) -> view -> expand -> flatten.
-    positions_node = graph.call_function(
-        torch.ops.aten.arange,
-        args=(sym_seq,),
+    position_ids = graph.call_function(
+        torch.arange,  # torch.ops.aten.arange cannot take in Fake Tensor as start/end
```
Fridah-nv, line 825: I wasn't able to get this working with `torch.ops.aten.arange` , this previous code worked:
`positions_node = graph.call_function(torch.ops.aten.arange, args=(sym_seq,),)`
but this failed:
```
bs_seq = graph.call_function(torch.ops.aten.mul, args=(sym_batch, sym_seq))
positions_node = graph.call_function(torch.ops.aten.arange, args=(bs_seq,),)
```

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,115 +1,461 @@
...
+                node.args = new_args
+
+                # retrieve q and k output node from node
+                old_q, old_k = extract_output_tuple(node, 2)
+                if old_q is None or old_k is None:
+                    ad_logger.warning(
+                        f"Failed to extract all two outputs from the explicit op, \
+                            get {old_q}, {old_k}, fail to replace {node} with flashinfer rope"
+                    )
+                    return
```
lucaslie, line 256: This seems strange? Why return instead of continue? 

Also it seems to me we should extract this from the output tuple before we make any modifications to the graph so that we can skip it if something goes wrong before we modify the graph? 

`old_q`, `old_k` also doesn't seem like good variables? Maybe something like `q_rope_old` ?

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,115 +1,461 @@
...
+                with graph.inserting_after(old_q):
+                    new_q = graph.call_function(torch.ops.aten.transpose, args=(old_q, 1, 2))
+                with graph.inserting_after(old_k):
+                    new_k = graph.call_function(torch.ops.aten.transpose, args=(old_k, 1, 2))
+
+                # Preserve fake tensor in meta["val"] for the transposed inputs
+                old_q.meta["val"] = old_q.meta["val"].transpose(1, 2)
+                new_q.meta["val"] = old_q.meta["val"].transpose(1, 2)
+                old_k.meta["val"] = old_k.meta["val"].transpose(1, 2)
+                new_k.meta["val"] = old_k.meta["val"].transpose(1, 2)
```
lucaslie, line 266:
```suggestion
                new_q.meta["val"] = old_q.meta["val"].transpose(1, 2)
                new_k.meta["val"] = old_k.meta["val"].transpose(1, 2)
```

why update old_q metadata and then update new_q metadata with the updated old_q metadata? This seems wrong

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,115 +1,461 @@
...
+    the optimized `rope::flashinfer` kernel.
+    Precomputes positional IDs and the fused cosine-sine cache as explicit nodes,
+    and reuses those nodes when possible.
+    """
+    ad_logger.info("RoPE optimization")
+    graph = gm.graph
+    rope_flash_cache: DefaultDict[Any, Optional[Node]] = defaultdict(lambda: None)
+    rope_position_ids_cache: Dict[str, Node] = {}
+
+    for node in list(graph.nodes):
```
lucaslie, line 290: what's wrong with
```
for node in graph.nodes:
```
?

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
```diff
@@ -42,6 +42,10 @@ def apply_rope_with_input_pos_flashinfer(
     q_flat = q.view(batch_size * seq_len, -1)
     k_flat = k.view(batch_size * seq_len, -1)
 
+    # flatten cos_sin_cache in case cos/sin is different across batches
+    cos_sin_cache = cos_sin_cache.view(batch_size * seq_len, -1)
```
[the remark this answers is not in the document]
Fridah-nv: I see. Thanks for explaining.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: It works, thanks for the suggestion!

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,115 +1,461 @@
...
 from ...utils.logger import ad_logger
-from ...utils.node_utils import bfs, identify_regions_between_residuals, is_op
+from ...utils.node_utils import bfs, extract_output_tuple, identify_regions_between_residuals, is_op
 from .._graph import canonicalize_graph
 
 
-def match_rope_v1(gm: GraphModule) -> GraphModule:
+def match_explicit_rope(gm: GraphModule) -> GraphModule:
     """
     Identify and replace legacy RoPE subgraphs (explicit cos/sin multiplication pattern):
```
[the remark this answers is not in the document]
Fridah-nv: Removed

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,115 +1,461 @@
...
+    graph = gm.graph
+    rope_ops = {
+        torch.ops.rope.torch_apply_rope_with_explicit_cos_sin,
+        torch.ops.rope.torch_apply_rope_with_qk_interleaving,
+        torch.ops.rope.torch_apply_rope_with_complex_freqs,
+    }
+
+    need_transpose = False
+    for node in list(graph.nodes):
+        if is_op(node, rope_ops):
```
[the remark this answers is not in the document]
Fridah-nv: I agree, made the update.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: Moved the check upward and added continue. Thanks for the catch.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: When we change the layout of the input to the custom op, the output of it also changed, so I put `old_q.meta["val"] = old_q.meta["val"].transpose(1, 2)`.  `new_q` is `old_q.transpose(1,2)`, so it is actually of the same shape as previous output of the custom op. 
I changed the code a bit, hope it's less confusing.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: There could be some issue with `for node in graph.nodes:` if some nodes are deleted in the loop, but there's only node insertion here. I can iterate through it without keeping a snapshot.

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,115 +1,461 @@
...
+    if q_fake is not None and len(q_fake.shape) > 2:
+        if not (isinstance(q_fake.shape[1], torch.SymInt) and isinstance(q_fake.shape[2], int)):
+            ad_logger.warning(
+                f"""Sanity check failed: q_fake should have shape [b, s, n, d],
+                s should be symbolic and n should be int, instead got shape {q_fake.shape}"""
+            )
+    elif q_fake is not None:
+        ad_logger.warning(
+            f"Sanity check failed: q_fake should be 3D or 4D, but got shape {q_fake.shape}"
+        )
```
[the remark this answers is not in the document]
Fridah-nv: Added return, thanks for the catch!

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
```diff
@@ -1,115 +1,466 @@
...
+        q_rope_old.meta["val"] = q_rope_old.meta["val"].transpose(1, 2)
+        k_rope_new.meta["val"] = k_rope_old.meta["val"]
+        k_rope_old.meta["val"] = k_rope_old.meta["val"].transpose(1, 2)
+
+        q_rope_old.replace_all_uses_with(q_rope_new)
+        k_rope_old.replace_all_uses_with(k_rope_new)
+        q_rope_new.args = (q_rope_old, 1, 2)
+        k_rope_new.args = (k_rope_old, 1, 2)
+
+    if need_transpose:
```
lucaslie, line 274: I think the logic here is wrong. You need a boolean here that indicates whether _any_ rope needed a transpose. 

In your logic right now when all ropes except the last one need transpose, you won't call canonicalize_graph.

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
```diff
@@ -21,10 +21,10 @@ def apply_rope_with_input_pos_flashinfer(
         Tensors of shape [batch, seq_len, n_head, head_dim] (or a 3D variant)
         in half precision. Note: head_dim must be a multiple of 64.
     - position_ids (torch.Tensor):
-        Precomputed tensor of positional indices; it is shared across calls in the graph.
+        Precomputed tensor of positional indices; Shape [batch, seq_len]
```
lucaslie, line 24:
```suggestion
        Precomputed tensor of *flattened* positional indices indicating idx in flattened cos_sin_cache for each token; Shape [batch, seq_len] or [batch * seq_len]
```

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
```diff
@@ -21,10 +21,10 @@ def apply_rope_with_input_pos_flashinfer(
         Tensors of shape [batch, seq_len, n_head, head_dim] (or a 3D variant)
```
lucaslie, line 21:
```suggestion
        Tensors of shape [batch, seq_len, n_head, head_dim] or [batch*seq_len, n_head, head_dim]
```

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
```diff
@@ -21,10 +21,10 @@ def apply_rope_with_input_pos_flashinfer(
...
         in half precision. Note: head_dim must be a multiple of 64.
     - position_ids (torch.Tensor):
-        Precomputed tensor of positional indices; it is shared across calls in the graph.
+        Precomputed tensor of positional indices; Shape [batch, seq_len]
     - cos_sin_cache (torch.Tensor):
         Precomputed fused tensor created by concatenating the first half of the cosine and sine
-        components derived from the inv_freq.
+        components derived from the inv_freq. Shape [batch, seq_len, head_dim]. Must be float32.
```
lucaslie, line 27:
```suggestion
        components derived from the inv_freq. Shape [batch, seq_len, head_dim] or [batch*seq_len, head_dim]. Must be float32.
```

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
```diff
@@ -42,7 +42,9 @@ def apply_rope_with_input_pos_flashinfer(
     q_flat = q.view(batch_size * seq_len, -1)
     k_flat = k.view(batch_size * seq_len, -1)
 
-    position_ids = position_ids.to(q.device)
+    # flatten cos_sin_cache in case cos/sin is different across batches
+    cos_sin_cache = cos_sin_cache.view(batch_size * seq_len, -1)
```
lucaslie, line 46:
```suggestion
    cos_sin_cache = cos_sin_cache.view(-1, head_dim)
```

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
```diff
@@ -42,7 +42,9 @@ def apply_rope_with_input_pos_flashinfer(
...
+    position_ids = position_ids.view(batch_size * seq_len, -1).to(q.device)
```
lucaslie, line 47:
```suggestion
    position_ids = position_ids.view(-1).to(q.device)
```

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_rope.py
```diff
@@ -42,7 +42,9 @@ def apply_rope_with_input_pos_flashinfer(
     q_flat = q.view(batch_size * seq_len, -1)
     k_flat = k.view(batch_size * seq_len, -1)
```
lucaslie, line 43: let's not rely on batch_size or seq_len when we assume 3d/4d input... 

We can do the following:

```
position_ids = position_ids.view(-1).to(q.device)
num_nnz = position_ids.shape[0]  # this is batch_size * seq_len
q_flat = q.view(num_nnz, -1)
k_flat = k.view(num_nnz, -1)

lucaslie (approved):
I thought a little more about the 3d vs 4d for the flashinfer kernel and added a few comments/suggestions to make it clearer what the kernel expects. Please also make sure that the transformation is compatible with the op interface. Otherwise, looks great

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
lucaslie: gotcha. makes sense now

tensorrt_llm/_torch/auto_deploy/transformations/library/rope.py
Fridah-nv: I agree, thank you for the catch.

Fridah-nv:
/bot run --disable-fail-fast --stage-list "DGX_H100-4_GPUs-PyTorch-[Post-Merge]"

Fridah-nv:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-PyTorch-[Post-Merge]"

Fridah-nv:
/bot run --extra-stage "DGX_H100-4_GPUs-PyTorch-[Post-Merge]"

PR #3640, opened 2025-04-17
Clean up modeling_deepseek.py

hlu1:
- Rename some variables to make the code more readable
- Remove unnecessary if/else branches
- Simplify some of the logics in allreduce fusion

The `modeling_deepseek.py` files looks mostly clean after this round of cleaning. I kept the allreduce part the same because we need the allreduce op unification PR to land first.

For accuracy test, I ran the `examples/pytorch/quickstart_advanced.py` script with the R1 model with different configs.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_deepseekv3.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv3.py
@@ -99,14 +99,14 @@ class Deepseekv3RoutingImpl():
         n_group: int,
         topk_group: int,
         routed_scaling_factor: float,
-        is_thop: bool = True,
+        is_fused: bool = True,
     ):
         super().__init__()
         self.top_k = top_k
         self.topk_group = topk_group
         self.n_group = n_group
         self.routed_scaling_factor = routed_scaling_factor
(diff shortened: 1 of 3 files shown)


tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -182,7 +182,8 @@ def __init__(
         topk_group: int,
         routed_scaling_factor: float,
         dtype: Optional[torch.dtype] = None,
-        is_thop: bool = True,
+        fuse_routing_kernel: bool = True,
+        apply_routing: bool = False,
```
QiJune, line 186: In which case will `apply_routing` be `True`?  It seems tha `apply_routing` is always `False` in this `modeling_deepseekv3.py`.

hlu1: It's more like a tag to tell the user that the routing is skipped in this module.

SimengLiu-nv (approved):

QiJune (approved):
LGTM

PR #3641, opened 2025-04-17
feat: Add smart router for moe module

zongfeijing:
To address the issue of unbalanced experts, a smart router has been implemented. In the minimum latency scenario, multiple clusters hold the same expert. 

During runtime, experts are dynamically allocated to different clusters to ensure that the number of activated experts on different clusters is balanced.


tensorrt_llm/llmapi/llm_args.py
```diff
@@ -729,6 +731,11 @@ class LlmArgs(BaseModel):
     gpus_per_node: Optional[int] = Field(
         default=None, description="The number of GPUs per node.")
 
+    moe_cluster_parallel_size: Optional[int] = Field(
```
Superjomn, line 736: When modifying the LlmArgs, please update the api [references](https://github.com/superjomn/TensorRT-LLM/tree/0fd81f75be38f0772c634f896bea227e154c3404/tests/unittest/api_stability/references/llm.yaml) as well. It will check if both are aligned to keep the APIs stable.

Superjomn (approved):
LGTM for the LlmArgs, serve and bench interface modification.

Kefeng-Duan (approved):

PR #3643, opened 2025-04-17
test: remove qa benchmark test list

crazydemo:
remove benchmark test list, as the cases already covered in examples_test_list


crazydemo:
@LarryXFly Could you please help skip CI, as this PR only removes .txt

LarryXFly (approved):

PR #3649, opened 2025-04-17
fix: hmac in remote mpi session

Superjomn:


tensorrt_llm/llmapi/trtllm-llmapi-launch
```diff
@@ -23,17 +23,28 @@ function mpi_world_size {
...
+    try:
+        s.bind(("", port))
+        break
+    except OSError:
+        port += 1
+print(port); s.close()')
     export TLLM_SPAWN_PROXY_PROCESS_IPC_ADDR="tcp://127.0.0.1:${free_port}"
     log_stderr "TLLM_SPAWN_PROXY_PROCESS_IPC_ADDR: $TLLM_SPAWN_PROXY_PROCESS_IPC_ADDR"
+
+    export TLLM_SPAWN_PROXY_PROCESS_IPC_HMAC_KEY=$(openssl rand -hex 32)
```
litaotju, line 40: @yibinl-nvidia could you pls help to review this? Is if secure to use env to pass the key here? Thx

yibinl-nvidia: Thanks @litaotju for bring this to my attention. How worse is the regression? Is my understanding correct that this regression caused the timeout issue in the unit test?

If the env scope is only within mpi_session and user need to explicitly turn it on, I think it is fine for now.

yibinl-nvidia: Also do you think this will be an issue for 0.18 release? @Superjomn

Superjomn: I think the regression is minor. As the environment variables are private to the process.

tensorrt_llm/executor/utils.py
```diff
@@ -41,8 +41,12 @@ def create_mpi_comm_session(
         print_colored_debug(
             f"Using RemoteMpiPoolSessionClient to bind to external MPI processes at {get_spawn_proxy_process_ipc_addr_env()}\n",
             "yellow")
+        hmac_key = os.getenv("TLLM_SPAWN_PROXY_PROCESS_IPC_HMAC_KEY")
+        # Convert the hex string to bytes
+        if hmac_key is not None:
+            hmac_key = bytes.fromhex(hmac_key)
         return RemoteMpiCommSessionClient(
-            get_spawn_proxy_process_ipc_addr_env())
+            addr=get_spawn_proxy_process_ipc_addr_env(), hmac_key=hmac_key)
```
yibinl-nvidia, line 49: Is there a reason we don't let ZeroMqQueue to generate the key? This env flag could just be a string boolean value passed to use_hmac_encryption.

Superjomn: There are some trivial details: based on the existing implementation, both the ipc address and the hmac_key should be determined first before the client and server start, so we create both in shell, and then pass them to Python by env. As discussed with @litaotju , we will move the creation to Python in the following enhancement.

yibinl-nvidia: I see, thanks

tensorrt_llm/llmapi/trtllm-llmapi-launch
```diff
@@ -23,17 +23,28 @@ function mpi_world_size {
...
 }
 
+
 export tllm_mpi_size=$(mpi_world_size)
 log_stderr "tllm_mpi_size: $tllm_mpi_size"
 
-export_free_tcp_addr
+export_free_tcp_addr_for_spawn_proxy_process
```
litaotju, line 47: Nit @Superjomn  I think we have discussed this, it only beeds to be export in mpi rank 0?

Do you think its a safe change here? I am fine if you'd like to do it in another change.

tensorrt_llm/executor/utils.py
```diff
@@ -41,8 +41,12 @@ def create_mpi_comm_session(
         print_colored_debug(
             f"Using RemoteMpiPoolSessionClient to bind to external MPI processes at {get_spawn_proxy_process_ipc_addr_env()}\n",
             "yellow")
+        hmac_key = os.getenv("TLLM_SPAWN_PROXY_PROCESS_IPC_HMAC_KEY")
+        # Convert the hex string to bytes
+        if hmac_key is not None:
```
litaotju, line 46: @Superjomn what will happen if this is None, i.e. the user script is not started with the trtllm-llmapi-launch.

Will this code work? Thx

litaotju (approved):

tensorrt_llm/llmapi/trtllm-llmapi-launch
Superjomn: Yeah, I will perform this change if the current CI fails, or I will change it in a subsequent PR.

tensorrt_llm/executor/utils.py
Superjomn: comm_session should always work with `trtllm-llmapi-launch`, when if it is None, it should go to line 54 and doesn't need an IPC.

tensorrt_llm/llmapi/trtllm-llmapi-launch
zeroepoch: @Superjomn If you have shell access as the user running the process you can read environment variables by looking at `/proc/<PID>/environ`. Although at that point you can already run commands as the user running the process so it's not less secure.

Superjomn: > @Superjomn If you have shell access as the user running the process you can read environment variables by looking at `/proc/<PID>/environ`. Although at that point you can already run commands as the user running the process so it's not less secure.

@zeroepoch Yeah, that is a fair concern. The background is that we need to share the port and HMAC key between an IPC pair from two processes to conduct communication. In [PR3728](https://github.com/NVIDIA/TensorRT-LLM/pull/3728) I refactored this part to have a temporary file to share the information instead. I admit this is also not secure, but currently, all the IPCs we add are all intra-node, is this somewhat acceptable considering hackers need to access the node first?

PR #3652, opened 2025-04-17
fix: Fix fused_moe cache fallback issue.

hyukn:
min_latency_mode is only set to False during the warmup phase. Thus, when it becomes true during inference, all tactics fall back to the default one and cause perf regression.
We should first fix the kernel issue when setting min_latency_mode=True in profiling. Then, remove the bypass and enable real tuning for min_latency mode.


zongfeijing (approved):
LGTM.

PR #3656, opened 2025-04-17
test: change default max_batch_size to 512 in test config and dump config.json to log

ruodil:
1. change default perf config max_batch_size to 512 and max_num_tokens to 2048 in benchmarking to avoid exeeding max_batch_size of engine: https://prod.blsm.nvidia.com/swqa-tensorrt-qa-test/view/TRT-LLM-Perf-Pipelines/job/Debug-TRT-LLM-Perf-Test/25/console
2. add dump engine config to output log
3. waive failed case with known bugs.


LarryXFly (approved):

tests/integration/defs/perf/test_perf.py
```diff
@@ -283,8 +283,8 @@ def __init__(
         backend: str = "",
         mode: str = "plugin",
         data_type: str = "float16",
-        max_batch_size: int = 2048,
-        max_num_tokens: int = 8192,
+        max_batch_size: int = 512,
```
kaiyux, line 286: This is meant to be aligned with [the default setting](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/builder.py#L480) in the core library, why do we need to change this?

tests/integration/test_lists/qa/trt_llm_release_perf_cluster_test.yml
```diff
@@ -12,7 +12,7 @@ trt_llm_release_perf_cluster_test:
   - perf/test_perf.py::test_perf[llama_v3.1_70b-bench-float16-maxbs:256-input_output_len:128,128-gpus:2]
   - perf/test_perf.py::test_perf[llama_v3.1_70b-bench-float16-maxbs:256-input_output_len:512,32-gpus:2]
   - perf/test_perf.py::test_perf[llama_v2_13b-bench-float16-input_output_len:128,128-loras:8-gpus:2]
-  - perf/test_perf.py::test_perf[t5-bench-float16-input_output_len:128,20]
+  # - perf/test_perf.py::test_perf[t5-bench-float16-input_output_len:128,20]
```
kaiyux: Shouldn't waived tests be added to the waive file, instead of commenting them?

ruodil: thanks for reminder， I'll put the cases in waive list file.

tests/integration/defs/perf/test_perf.py
ruodil: if we use default setting of max_batch_size/max_num_tokens, it will raise error that batch size when benchmarking cannot exceed max_batch_size of engine, https://prod.blsm.nvidia.com/swqa-tensorrt-qa-test/view/TRT-LLM-Perf-Pipelines/job/Debug-TRT-LLM-Perf-Test/25/console, and we cannot pass max_batch_size together with dataset when building engine(https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/bench/build/build.py)

LarryXFly: @kaiyux Do you have any more comments?

kaiyux: > if we use default setting of max_batch_size/max_num_tokens, it will raise error that batch size when benchmarking cannot exceed max_batch_size of engine

But why is engine built with a smaller `max_batch_size`? We should always have a runtime `max_batch_size` not larger than engine `max_batch_size`.

LarryXFly (approved):

PR #3659, opened 2025-04-17
feat: Introduce feature properties for attention backend.

yuxianq:


tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -484,6 +484,31 @@ class PredefinedAttentionMask(str, Enum):
 AttentionMask = Union[PredefinedAttentionMask]
 
 
+class AttentionBackendFeature(IntFlag):
+    """
+    Features supported by attention backends.
+    """
+    FUSED_ROPE = auto()
+    FUSED_QKV = auto()
+    UNFUSED_QKV = auto()
```
QiJune: Is there any attention backend support both FUSED_QKV and UNFUSED_QKV? If not, I think we can keep one of them.

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -484,6 +484,31 @@ class PredefinedAttentionMask(str, Enum):
 AttentionMask = Union[PredefinedAttentionMask]
 
 
+class AttentionBackendFeature(IntFlag):
```
QiJune: Do we really need to introduce a class? Or just adding some properties to AttentionBackend:

```
class TrtllmAttentionBackend:
    @property
    def support_fused_qkv(self):
        return True
```

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -158,10 +144,23 @@ def __init__(
         self.o_lora = LoraLayer([LoraModuleType.ATTENTION_DENSE],
                                 [self.hidden_size])
 
-        if not config.skip_create_weights:
-            self.create_weights()
+        self.create_backend()
```
hlu1: Why is this if condition removed?
```
if not config.skip_create_weights:
```

hlu1: If you look at https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/modules/linear.py#L199-L201

We don't allow overwriting the weights. Once they're created, you can't change them. It's done intentionally this way to avoid recreating the weights because it can be slow.

tensorrt_llm/_torch/attention_backend/interface.py
yuxianq: Sure. This design mimics QuantMode. I will try to switch to simple properties.

tensorrt_llm/_torch/modules/attention.py
yuxianq: Because when we call `self.create_weights()`, only attention backend is created, which has not weights. So we won't overwrite weights anyway. The only reason we put the backend creation inside `create_weights` is that we change `quant_config` when we quantize, we should re-create the backend to use the new `quant_config`.

We still need to create backend even if `config.skip_create_weights`, so that we can create `rotary_emb` for non-rope-fusion backend. Let me update it to:

```
if not config.skip_create_weights:
    self.create_weights()
else:
    self.create_backend()
```

If we add new weights to attention module in the future, `create_weights` will be different from `create_backend`, this design can work in that case.

tensorrt_llm/_torch/attention_backend/interface.py
yuxianq: Not for now. Let me keep `FUSED_QKV` only.

yuxianq: Updated.

tensorrt_llm/_torch/attention_backend/interface.py
yuxianq: Updated.

tensorrt_llm/_torch/modules/attention.py
yuxianq: Updated.

hlu1 (approved):

tensorrt_llm/_torch/modules/attention.py
hlu1: Having to recreate the attention op when `quant_config` changes is a design flow of the `create_weights` feature. I'll fix it in another PR.

yuxianq: Thanks~

PR #3661, opened 2025-04-17
Support RingAttention in the BertAttention plugin and the DiT model

forrestl111:
Support RingAttention in the BertAttention plugin and the DiT model

### DIFF
--- a/cpp/tensorrt_llm/kernels/recoverFromRingAtten.cu
+++ b/cpp/tensorrt_llm/kernels/recoverFromRingAtten.cu
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2019-2023, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "tensorrt_llm/common/cudaTypeUtils.cuh"
+#include "tensorrt_llm/kernels/recoverFromRingAtten.h"
+
+#include "math.h"
+#include <cooperative_groups.h>
+#include <cuda/barrier>
+
+using namespace tensorrt_llm::common;
+
+namespace tensorrt_llm
+{
+namespace kernels
+{
+
+template <typename Tout>
+__global__ void reduce4ring_attention(
+    // this is the accumulated results for all finished ring attention blocks
+    Tout* __restrict__ accu_output,       // b x s_block x h x d
+    float* __restrict__ accu_softmax_sum, // b x s_block x h
+    float* __restrict__ accu_max,         // b x s_block x h
+    // this is the new ring attention block results
+    Tout* __restrict__ output,       // b x s_block x h x d
+    float* __restrict__ softmax_sum, // b x s_block x h
+    float* __restrict__ max,         // b x s_block x h
+    // necessary constant parameters
+    int const b, int const s_block, int const h, int const d, int const block_seq_len, int* cu_seqlens)
+{
+    auto block = cooperative_groups::this_thread_block();
+    int batchid = blockIdx.x;
+    int block_seq_idx = blockIdx.y;
+    int block_s_start = block_seq_idx * block_seq_len;
+    int block_s_end = (block_seq_idx + 1) * block_seq_len;
+    block_s_end = s_block < block_s_end ? s_block : block_s_end;
+    int64_t output_start_offset = batchid * s_block * d + block_s_start * d;
+    int64_t lm_start_offset = batchid * s_block + block_s_start;
+
+    __shared__ cuda::barrier<cuda::thread_scope::thread_scope_block> barrier;
+    if (block.thread_rank() == 0)
+    {
+        init(&barrier, block.size());
+    }
+    block.sync();
+
+    int s_len = block_s_end - block_s_start;
+    int laneid = threadIdx.x % 32;
+    int local_warpid = threadIdx.x / 32;
+    int warp_num = blockDim.x / 32;
+    int loop_on_s = (s_len + warp_num * 32 - 1) / (warp_num * 32);
+    for (int l = 0; l < loop_on_s; l++)
+    {
+        int s_ = local_warpid + warp_num * laneid + l * warp_num * 32;
+        float scaled_my_ss1_ = 1.0, scaled_my_ss2_ = 1.0;
+        if (s_ < s_len)
+        {
+            uint64_t lm_start_offset_ = lm_start_offset + s_;
+            float my_accu_ss = accu_softmax_sum[lm_start_offset_] == 0.0 ? 1.0 : accu_softmax_sum[lm_start_offset_];
+            float my_ss = softmax_sum[lm_start_offset_] == 0.0 ? 1.0 : softmax_sum[lm_start_offset_];
+
+            float cur_max = (accu_max[lm_start_offset_] > max[lm_start_offset_]) ? accu_max[lm_start_offset_]
+                                                                                 : max[lm_start_offset_];
+            float scale1 = exp(accu_max[lm_start_offset_] - cur_max);
+            float scale2 = exp(max[lm_start_offset_] - cur_max);
+            float cur_softmax_sum = my_accu_ss * scale1 + my_ss * scale2;
+            if (cur_softmax_sum == 0)
+                cur_softmax_sum = 1.0;
+            scaled_my_ss1_ = scale1 * my_accu_ss / cur_softmax_sum;
+            scaled_my_ss2_ = scale2 * my_ss / cur_softmax_sum;
+            accu_softmax_sum[lm_start_offset_] = cur_softmax_sum;
+            accu_max[lm_start_offset_] = cur_max;
+        }
+        int sid = l * warp_num * 32 + local_warpid;
+        int s_end = (l + 1) * warp_num * 32 < s_len ? (l + 1) * warp_num * 32 : s_len;
+        for (int ss = 0;; ss++)
+        {
+            uint64_t output_start_offset_ = output_start_offset + sid * d;
+            float scaled_my_ss1 = __shfl_sync(0xffffffff, scaled_my_ss1_, ss, 32);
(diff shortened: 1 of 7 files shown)


byshiue (approved):

tensorrt_llm/functional.py
```diff
@@ -4449,7 +4449,10 @@ def bert_attention(tensor: Tensor,
                    sage_attn: bool = False,
                    sage_attn_q_block_size: int = 0,
                    sage_attn_k_block_size: int = 0,
-                   sage_attn_v_block_size: int = 0) -> Tuple[Tensor]:
+                   sage_attn_v_block_size: int = 0,
+                   cp_group: list[int] = [0],
```
yuxianq: Use `cp_group: list[int] = None` instead. List is mutable, using mutable object as default value will cause all instances share the same object. Add a `cp_group = cp_group or [0]`  line to set default value.

cpp/tensorrt_llm/plugins/bertAttentionPlugin/bertAttentionPlugin.cpp
```diff
@@ -800,7 +963,7 @@ size_t BertAttentionPlugin::getSerializationSize() const noexcept
     return sizeof(mNumHeads) + sizeof(mHeadSize) + sizeof(mQScaling) + sizeof(mQKHalfAccum) + sizeof(mEnableContextFMHA)
         + sizeof(mFMHAForceFP32Acc) + sizeof(mType) + sizeof(mRelativeAttention) + sizeof(mMaxDistance)
         + sizeof(mRemovePadding) + sizeof(mSageAttn) + sizeof(mSageAttnQBlockSize) + sizeof(mSageAttnKBlockSize)
-        + sizeof(mSageAttnVBlockSize);
+        + sizeof(mSageAttnVBlockSize) + sizeof(mCpSize) + sizeof(mCpRank) + sizeof(int32_t) * mCpGroup.size();
```
yuxianq, line 979: We do not store the lenght of `mCpGroup`, so the `sizeof(int32_t)` is unnecessary.

cpp/tensorrt_llm/kernels/recoverFromRingAtten.cu
```diff
@@ -0,0 +1,149 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "tensorrt_llm/common/cudaTypeUtils.cuh"
+#include "tensorrt_llm/kernels/recoverFromRingAtten.h"
+
+using namespace tensorrt_llm::common;
```
yuxianq: Can move this line after all includes.

cpp/tensorrt_llm/kernels/recoverFromRingAtten.cu
```diff
@@ -0,0 +1,149 @@
...
+
+            float cur_max = (accu_max[lm_start_offset_] > max[lm_start_offset_]) ? accu_max[lm_start_offset_]
+                                                                                 : max[lm_start_offset_];
+            float scale1 = exp(accu_max[lm_start_offset_] - cur_max);
+            float scale2 = exp(max[lm_start_offset_] - cur_max);
+            float cur_softmax_sum = my_accu_ss * scale1 + my_ss * scale2;
+            if (cur_softmax_sum == 0)
+                cur_softmax_sum = 1.0;
+            /*if (hid == 0 && sid == 0)
+            printf("for bid %d, hid %d, sid %d, accu_ss %f, ss %f , cur_ss %f, accu_max %f, max %f, cur_max %f oa_scale
```
yuxianq: Remove debug code or transform it to debug log.

cpp/tensorrt_llm/kernels/recoverFromRingAtten.cu
```diff
@@ -0,0 +1,149 @@
...
+        int sid = l * warp_num * 32 + local_warpid;
+        int s_end = (l + 1) * warp_num * 32 < s_len ? (l + 1) * warp_num * 32 : s_len;
+        for (int ss = 0;; ss++)
+        {
+            uint64_t output_start_offset_ = output_start_offset + sid * d;
+            float scaled_my_ss1 = __shfl_sync(0xffffffff, scaled_my_ss1_, ss, 32);
+            float scaled_my_ss2 = __shfl_sync(0xffffffff, scaled_my_ss2_, ss, 32);
+            for (int eid = laneid; eid < d; eid += 32)
+            {
+                // float old_accu_output = (float)accu_output[output_start_offset + eid];
```
yuxianq: Remove debug code.

cpp/tensorrt_llm/kernels/recoverFromRingAtten.cu
```diff
@@ -0,0 +1,149 @@
...
+                accu_output[output_start_offset_ + eid]
+                    = (float) accu_output[output_start_offset_ + eid] * scaled_my_ss1
+                    + (float) output[output_start_offset_ + eid] * scaled_my_ss2;
+                /*if (rank == 0 && sid == 0 && hid == 0 && eid < 4)
```
yuxianq: Remove debug code or transform it to debug log.

cpp/tensorrt_llm/plugins/bertAttentionPlugin/bertAttentionPlugin.cpp
forrestl111: It seems the serialization size is counted by 'bytes'. And then it will be used when load data as 'length' [here](https://github.com/NVIDIA/TensorRT-LLM/blob/4b100c43163efb8c74987effbee116d28e4d61de/cpp/tensorrt_llm/plugins/bertAttentionPlugin/bertAttentionPlugin.cpp#L120). So I'm a little confused here.

yuxianq: Sorry for misleading, it is correct to use `sizeof(int32_t) * mCpGroup.size()`. I thought it was `sizeof(int32_t) + mCpGroup.size()`.

tensorrt_llm/functional.py
```diff
@@ -4570,10 +4582,32 @@ def bert_attention(tensor: Tensor,
...
+                    dim=-1).view(concat((2 * bs, seq_len, query.shape[-1])))
+        tensor = concat((query, kv),
+                        dim=0).view(concat((bs, seq_len, query.shape[-1] * 3)))
+        input_lengths = input_lengths // cp_size
+
+    cp_size = trt.PluginField("cp_size", np.array(cp_size, dtype=np.int32),
+                              trt.PluginFieldType.INT32)
+    cp_rank = trt.PluginField("cp_rank", np.array(cp_rank, dtype=np.int32),
+                              trt.PluginFieldType.INT32)
+    if cp_group is None: cp_group = [0]
```
yuxianq, line 4602: IMO `cp_group = cp_group or [0]` is shorter and simpler.

yuxianq (approved):

cpp/tensorrt_llm/plugins/bertAttentionPlugin/bertAttentionPlugin.cpp
```diff
@@ -382,176 +440,268 @@ int BertAttentionPlugin::enqueueImpl(nvinfer1::PluginTensorDesc const* inputDesc
...
+            // relative position of prev/next rank in cp group
+            int prev_rank = mCpRank > 0 ? mCpRank - 1 : mCpGroup.size() - 1;
+            int next_rank = (mCpRank == mCpGroup.size() - 1) ? 0 : mCpRank + 1;
+
+            common::check_cuda_error(cudaStreamCreate(&mNcclStream));
+            common::check_cuda_error(cudaStreamSynchronize(stream));
+
+            uint32_t* fmha_scheduler_counter_h = (uint32_t*) malloc(sizeof(uint32_t));
+            cudaMemcpyAsync(
+                fmha_scheduler_counter_h, fmha_tile_counter_ptr, sizeof(uint32_t), cudaMemcpyDeviceToHost, stream);
```
DylanChen-NV, line 466: `fmha_tile_counter_ptr` can be directly used, is `fmha_scheduler_counter_h` necessary?

ChunhuanLin: Yes, it is necessary. We need to invoke fmha multiple times in the Ring Attention feature. So we need a unique counter for the RingAttention.

tensorrt_llm/functional.py
```diff
@@ -4570,10 +4582,32 @@ def bert_attention(tensor: Tensor,
...
+        # (b, s, 3d)
+        query, key, value = chunk(tensor, 3, dim=-1)
+        bs = shape(query, 0)
+        seq_len = shape(query, 1)
+        # (b, s, d) -> (b, s, 2d) -> (2b, s, d)
+        kv = concat([key, value],
+                    dim=-1).view(concat((2 * bs, seq_len, query.shape[-1])))
+        tensor = concat((query, kv),
+                        dim=0).view(concat((bs, seq_len, query.shape[-1] * 3)))
+        input_lengths = input_lengths // cp_size
```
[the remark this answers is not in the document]
ChunhuanLin: Good point. In fact, the "input_lengths" don't need to be divisible by "cp_size". However, it must be the correct value to tell how long the seq length is in current rank to compute RingAttention. Each rank is responsible to compute part of the whole sequence,  so we need this arg. I think it better let "input_lengths" computed in the model instead of BertAttention, since we can get more message in the model level.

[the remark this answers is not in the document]
ChunhuanLin: please review this commit: https://github.com/NVIDIA/TensorRT-LLM/pull/3661/files/3bb35516c61f8018f1b87eaca0fb7c3b646ded40..ab3733834567c4bc3ccb8106a809211ce8179a54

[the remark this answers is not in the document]
DylanChen-NV: The conversation can be resolved since the divisibility is guaranteed by [line](https://github.com/NVIDIA/TensorRT-LLM/pull/3661/files/3bb35516c61f8018f1b87eaca0fb7c3b646ded40#diff-8ec1233487289523755946e10fb6cc4b0bb66915aa56ca2a3de2c25bd9187b7dR301)

cpp/tensorrt_llm/plugins/bertAttentionPlugin/bertAttentionPlugin.cpp
DylanChen-NV: Discussed with Chunhuan and found the reason to use `fmha_scheduler_counter_h` here is that the fmha kernel would modify `fmha_tile_counter_ptr` buffer. 
There is a concern that additional H2D/D2H might affect performance(previously found in weightOnlyPlugin), but it's not necessary to be addressed in this PR.

DylanChen-NV (approved):

PR #3677, opened 2025-04-18
feat: Add NIXL support

Shixiaowei02:


juney-nvidia:
Thanks @Shixiaowei02 for preparing this MR.

Thanks for the hard work from NIXL folks!

It is great to see this NIXL integration MR coming in, let's prioritize the review of this MR. @chuangz0 .

June

cpp/tensorrt_llm/batch_manager/cacheFormatter.h
```diff
@@ -65,6 +72,46 @@ class CacheFormatter final : public IOFormatter
         : mCacheManager{cacheManager}
     {
         TLLM_CHECK(mCacheManager);
+
```
ebarilanM, line 75: why not have a single pool? most of the time worker will be or sender or receiver

cpp/tensorrt_llm/batch_manager/dataTransceiverImpl.cpp
```diff
@@ -16,40 +16,110 @@
...
+#include "tensorrt_llm/batch_manager/mlaCacheFormatter.h"
+#include "tensorrt_llm/executor/cache_transmission/nixl_utils/connection.h"
 #include "tensorrt_llm/runtime/utils/mpiUtils.h"
 
 namespace tensorrt_llm::batch_manager
 {
 
 DataSenderImpl::DataSenderImpl(executor::kv_cache::ConnectionManager* manager,
-    executor::kv_cache::CacheState selfCacheState, SizeType32 selfIndex, std::unique_ptr<IOFormatter> formatter)
+    executor::kv_cache::ConnectionManager* nixlManager, executor::kv_cache::CacheState selfCacheState,
```
ebarilanM, line 30: have both mManager and mNixlManager is temp? do we plan remove it when support new commState?

zackyoray: Right, this is only temp, we will use only mManager (which will be MPI/UCX/NIXL)

PR #3681, opened 2025-04-18
Fix/executor bugs

byshiue:
1. When we estimate the kv cache memory, we need to consider that we need to add extra token when we enable the kv cache reuse. So, For each sequence of max_num_tokens length, we may need up to one extra block.

2. After getting the max_kv_cache_token by warmup, we recreate a cache manager. But we enable the start_worker there, and start it again. So, disable the first start_worker.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -1,3 +1,4 @@
+import math
 import random
 from collections.abc import Iterable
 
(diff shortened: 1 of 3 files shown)


Shunkangz (approved):

byshiue:
/bot run --stage-list "DGX_H100-4_GPUs-PyTorch-1, DGX_H100-4_GPUs-PyTorch-2"

byshiue:
/bot run --stage-list "DGX_H100-4_GPUs-PyTorch-1, DGX_H100-4_GPUs-PyTorch-2"

PR #3686, opened 2025-04-18
[fix] Fix a few issues with EAGLE3 in PyTorch backend

mikeiovine:
* Add a few more explanatory comments
* Clean up draft model hidden state saving logic. 
* Add eagle3 support to flashinfer (includes CUDA graph support). Flashinfer is still useful for debugging; it helped me identify some issues in the warmup phase.
* Fix an illegal memory access related to maximum sequence length in warmup phase. This currently does not cause crashes on the TRTLLM attention path (but it's still UB!)

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/flashinfer.py
+++ b/tensorrt_llm/_torch/attention_backend/flashinfer.py
@@ -122,7 +122,10 @@ class FlashInferAttentionMetadata(AttentionMetadata):
         super().__post_init__()
 
         if self.workspace_buffer is None:
-            self.workspace_buffer = torch.empty(256 * 1024 * 1024,
+            # Note: even though flashinfer only recommends 128 MB, we have to push it
+            # a bit higher to cover all possible CUDA graph cases. If it's too small,
+            # warmup will crash.
+            self.workspace_buffer = torch.empty(320 * 1024 * 1024,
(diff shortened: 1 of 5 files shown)


tensorrt_llm/_torch/attention_backend/flashinfer.py
```diff
@@ -122,7 +122,7 @@ def __post_init__(self) -> None:
         super().__post_init__()
 
         if self.workspace_buffer is None:
-            self.workspace_buffer = torch.empty(256 * 1024 * 1024,
+            self.workspace_buffer = torch.empty(512 * 1024 * 1024,
```
mikeiovine: Note: even though flashinfer recommends only 128MB, warmup will fail (error related to workspace size) without this change. Will add a comment on next iteration

hlu1 (approved):

lfr-0531 (approved):

lfr-0531:
Adding @yuxianq for vis of this PR. There are some changes about the flashinfer backend.

mikeiovine:
Will followup with @yuxianq offline if more changes are requested. For now I think the functional changes to the flashinfer backend are small enough to just land this (and they're pretty much limited to the speculation path only). The bulk of this PR is bug fixing.

PR #3688, opened 2025-04-18
fix: https://nvbugs/5177565 PP4 race condition

amukkara:
1. Sync send ops to avoid races in PP4
2. Avoid unmatched PP.recv() when capturing CUDA Graphs
3. Add few llama PP4 test cases
4. Minor cleanup

Relevant bugs: 5177565 and 5181511


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1153,10 +1148,7 @@ def _broadcast_new_requests(self, new_requests):
                 self.dist.recv(buf, self.dist.prev_pp_rank, tag)
 
             if not self.dist.is_last_pp_rank:
-                if self.send_new_reqs_handle[1] is not None:
-                    self.send_new_reqs_handle[1].Wait()
-                self.send_new_reqs_handle[1] = self.dist.isend(
-                    buf, self.dist.next_pp_rank, tag)
+                self.dist.send(buf, self.dist.next_pp_rank, tag)
```
chang-l, line 1151: Any idea why `isend` isn't working here?

tests/integration/test_lists/waives.txt
```diff
@@ -370,10 +370,6 @@ accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales_4gpu
 accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales_4gpus[tp4-mtp_nextn=2-attention_dp-cuda_graph-overlap_scheduler] SKIP (https://nvbugs/5201530)
 accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales_4gpus[ep4-attention_dp-cuda_graph-overlap_scheduler] SKIP (https://nvbugs/5201530)
 accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales_4gpus[ep4-mtp_nextn=2-attention_dp-cuda_graph-overlap_scheduler] SKIP (https://nvbugs/5201530)
-accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_bfloat16_4gpus[pp4-cuda_graph] SKIP (https://nvbugs/5181511)
-accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_bfloat16_4gpus[pp4-attention_dp-cuda_graph-overlap_scheduler] SKIP (https://nvbugs/5181511)
-accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales_4gpus[pp4-cuda_graph] SKIP (https://nvbugs/5181511)
```
chang-l, line 366: Do we have any pp4 test for llama as well? If not, can we add it?

chang-l (approved):
Overall looks good to me, so stamping...

tests/integration/test_lists/waives.txt
amukkara: Added in 0815f12595709c3c7f8b27588f9d04f347f3e0a6
One test-case in pre-merge and two in post-merge to control CI time.

tensorrt_llm/_torch/pyexecutor/py_executor.py
amukkara: when request has many tokens (> 1024), the recv on rank > 0 just hangs.

amukkara: root cause for hang with isend:
`buf` is a local variable that will be garbage collected after returning from this routine. to ensure isend fully copies `buf`, wait() should be called on isend before returning from this function (not in the next invocation).

```
handle = dist.isend(buf)
new_requests = dill.loads(buf.tobytes()) // This will overlap with isend
handle.wait()
```

Since work that can overlap with isend is minimal, let's use send and reconsider if there is a perf drop.

chang-l: Thanks Anurag. That makes sense.

achartier (approved):
LGTM

litaotju:
I am delegating the detailed review to @yuxianq . Thx

tensorrt_llm/_torch/pipeline_interface.py
```diff
@@ -71,7 +71,11 @@ def recv(self):
 
     def send(self):
         """Send tensors to next rank."""
+        # pp_comm.send returns after nccl send kernel is enqueued. Event sync waits till prev kernel
+        # finishes and avoids earlier PP rank executing multiple microbatches ahead of later rank.
+        self._pp_comm.send_event.synchronize()
```
yuxianq, line 76: All `pp_comm.send` are called in the default stream, so they should already be sequential. Why we need a send event to guarantee the order?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1139,10 +1137,7 @@ def _broadcast_new_requests(self, new_requests):
             self.dist.recv(metadata_arr, self.dist.prev_pp_rank, tag)
 
         if not self.dist.is_last_pp_rank:
-            if self.send_new_reqs_handle[0] is not None:
-                self.send_new_reqs_handle[0].Wait()
-            self.send_new_reqs_handle[0] = self.dist.isend(
-                metadata_arr, self.dist.next_pp_rank, tag)
+            self.dist.send(metadata_arr, self.dist.next_pp_rank, tag)
```
yuxianq, line 1140: Can we broadcast like non-pp cases with `self.dist.broadcast(new_requests, root=0)` instead of using step-by-step send-recv operation?

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -162,11 +161,11 @@ def estimate_max_kv_cache_tokens(py_executor: PyExecutor,
         req = create_dummy_context_requests(max_num_tokens, origin_seq_len,
                                             vocab_size)
         req_ids = py_executor.enqueue_requests(req)
-    req_ids = mpi_broadcast(req_ids, root=0)
+    req_ids = py_executor.dist.broadcast(req_ids, root=0)
     py_executor.start_worker()
     py_executor.await_responses(req_ids)
     # sync all ranks after processing dummy requests
-    mpi_allgather(0)
+    py_executor.dist.allgather(0)
```
yuxianq, line 168: Can we use mpi barrier insead of allgather(0)? cc @HuiGao-NV

HuiGao-NV: I tested with mpi barrier, however it always hang. I didn't dig deep on the root cause.

yuxianq: Sure. @amukkara Please help to add a comment here to indicate that allgather(0) is a trick to workaround mpi barrier's hang issue, so that we won't replace it with mpi barrier accidently.

tensorrt_llm/_torch/pyexecutor/py_executor.py
amukkara: a broadcast barrier is too restrictive for PP and will cause bubbles in PP microbatching (only one GPU will be active at any given point).

within a TP group, broadcast is used [here](https://github.com/amukkara/TensorRT-LLM/blob/df46d4dde2f7b8c17c2a74c8d87b880d50a0e06b/tensorrt_llm/_torch/pyexecutor/py_executor.py#L1121). for example with TP2PP2, ranks 0,1 enter a broadcast. rank0 then sends to rank2, rank1 to rank3.

tensorrt_llm/_torch/pyexecutor/_util.py
amukkara: added comment in b44aac800433514

tensorrt_llm/_torch/pipeline_interface.py
amukkara: the sync is not to guarantee order within a rank, but to add some back-pressure form later PP rank to earlier rank.
with this sync, earlier PP rank will be at most one microbatch ahead of later rank.

tensorrt_llm/_torch/pyexecutor/py_executor.py
yuxianq: Make sense.

tensorrt_llm/_torch/pipeline_interface.py
yuxianq: We use blocking MPI send to broadcast the next microbatch's requests to the next PP rank, which has guaranteed that the earlier PP rank will wait for the later PP rank to finish its previous microbatch, so it will be at most one microbatch ahead of later rank. Do you think it is enough for back pressure?

amukkara: IIUC blocking MPI send returns after the application buffer in the sending task (new requests in this example) is free for reuse, but does not wait for destination to finish recv. https://hpc-tutorials.llnl.gov/mpi/blocking/

yuxianq: Got it. If we want Synchronous blocking send, we should use `MPI_Ssend` instead. Thanks for your detailed explanation~

yuxianq (approved):

litaotju (approved):

PR #3692, opened 2025-04-18
Fix create_weights in attention

hlu1:
Previously, `Attention.create_weights` had to recreate the Attention op when quant_config changes. This is due to some quant states that the attention op sets up in `TrtllmAttention.__init__`. The solution is to move to the quant related changes to another function that can be called in `Attention.create_weights`. 

Separately, config `skip_create_weights` is renamed to `skip_create_weights_in_init` to make it more explicit. 
Also fixed a bug in `MLA` that's related to `MLA.create_weights`. The fp8 related tensors creation code needs to be moved to inside `MLA.create_weights`.

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/interface.py
+++ b/tensorrt_llm/_torch/attention_backend/interface.py
@@ -495,6 +495,7 @@ class AttentionBackend(Generic[TMetadata]):
         head_dim: int,
         num_kv_heads: Optional[int] = None,
         quant_config: Optional[QuantConfig] = None,
+        skip_create_weights_in_init: bool = False,
         **kwargs,
     ):
         """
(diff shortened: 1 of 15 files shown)


tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -512,6 +512,13 @@ def __init__(
         self.num_kv_heads = num_kv_heads or self.num_heads
         self.quant_config = quant_config
 
+    def set_up_quant_states(self, *args, **kwargs):
```
yuxianq: Can be `def set_up_quant_states(self)` since no argument is provided.

yuxianq (approved):

hlu1:
/bot skip -comment "The previously failed test passed in the rerun"

hlu1:
/bot skip --comment "The previously failed test passed in the rerun"

PR #3695, opened 2025-04-19
fix:  Refactor Deepseek tp_size calculation

hlu1:
- Add two new functions and docstrings
- Get block_size from checkpoint instead of hardcoding it to 128, which only applies to fp8 block scaling
- Add two FIXMEs

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_deepseekv3.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv3.py
@@ -272,22 +272,15 @@ class Deepseekv3MoE(nn.Module):
             model_config=model_config,
             aux_stream=aux_stream_dict[AuxStreamType.MoeChunkingOverlap])
 
-        self.shared_output_scale = None
-        # The block scale size is 128, which requires shared_expert_intermediate_size to be divisible by 128.
-        assert shared_expert_intermediate_size % 128 == 0
-        if self.use_dp:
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -479,6 +497,38 @@ def __init__(self, model_config: ModelConfig[PretrainedConfig],
         if not self.deepseek_allreduce_disabled:
             self.deepseek_allreduce = DeepseekAllReduce(self.mapping)
 
+    def _compute_mlp_tp_size(self, intermediate_size: int,
+                             block_size: int) -> int:
+        """
+        In the case of Deepseek-R1, the TP size of MLP is capped by intermediate_size // block_size.
+        To avoid the costly inter-node all-reduce, we further restrict TP size to be
+        divisible by gpus_per_node.
```
jinyangyuan-nvidia:
```suggestion
        a factor of gpus_per_node.
```

I made a mistake when modifying this part in [PR 3572](https://github.com/NVIDIA/TensorRT-LLM/pull/3572). Could you help correct it? Thanks.

jinyangyuan-nvidia (approved):

tensorrt_llm/_torch/models/modeling_deepseekv3.py
hlu1: Done

PR #3710, opened 2025-04-21
test: skip failed cases on B200

xinhe-nv:
skip weight only and some unsupported cases on B200

### DIFF
--- a/tests/integration/defs/accuracy/test_cli_flow.py
+++ b/tests/integration/defs/accuracy/test_cli_flow.py
@@ -19,7 +19,8 @@ from tensorrt_llm.llmapi import (EagleDecodingConfig, LookaheadDecodingConfig,
 from tensorrt_llm.quantization import QuantAlgo
 
 from ..conftest import (llm_models_root, parametrize_with_ids, skip_no_nvls,
-                        skip_pre_ada, skip_pre_blackwell, skip_pre_hopper)
+                        skip_post_blackwell, skip_pre_ada, skip_pre_blackwell,
+                        skip_pre_hopper)
 from .accuracy_core import (MMLU, CliFlowAccuracyTestHarness, CnnDailymail,
                             Humaneval, PassKeyRetrieval64k,
                             PassKeyRetrieval128k, SlimPajama6B, ZeroScrolls)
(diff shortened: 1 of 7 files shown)


kaiyux (commented):
Is there a bug filed for those changes? When should we add the coverage back?

xinhe-nv:
> Is there a bug filed for those changes? When should we add the coverage back?

Some of them are already waived in list. https://github.com/NVIDIA/TensorRT-LLM/blob/74c13ea84f3bb6f306716178eae1f28faa2206d4/tests/integration/test_lists/waives.txt#L70. But we need to waive them on blackwell with different device name.

tests/integration/defs/accuracy/test_cli_flow.py
```diff
@@ -194,9 +197,11 @@ class TestPhi2(CliFlowAccuracyTestHarness):
     MODEL_PATH = f"{llm_models_root()}/phi-2"
     EXAMPLE_FOLDER = "phi"
 
+    @skip_post_blackwell
```
syuoni, line 200: This should be a bfloat16 case, should we skip it on blackwell?

tests/integration/defs/accuracy/test_cli_flow.py
```diff
@@ -194,9 +197,11 @@ class TestPhi2(CliFlowAccuracyTestHarness):
...
     def test_auto_dtype(self):
         self.run(dtype='auto')
 
+    @skip_post_blackwell
```
syuoni, line 204: This is a bfloat16 TP=2 case, should we skip it on blackwell?

tests/integration/defs/accuracy/test_cli_flow.py
syuoni: As offline discussed, the shape is not supported by kernel.

tests/integration/defs/accuracy/test_cli_flow.py
syuoni: As offline discussed, the shape is not supported by kernel.

syuoni (approved):

litaotju (approved):

PR #3720, opened 2025-04-21
nvbugs/5232406  skip disagg deepseek test if sm!=90

chuangz0:


Shixiaowei02 (approved):

tests/integration/defs/disaggregated/test_disaggregated.py
```diff
@@ -145,6 +147,10 @@ def test_disaggregated_overlap(disaggregated_test_root, llm_venv,
 def test_disaggregated_deepseek_v3_lite_fp8(disaggregated_test_root,
                                             disaggregated_example_root,
                                             llm_venv, deepseek_v3_model_root):
+    if (get_sm_version() != 90):
+        pytest.skip(
+            f"DeepSeek FP8 is not supported in this SM version {get_sm_version()}"
+        )
```
litaotju: There are many duplication of this code.

I'd suggest to create a pytest marker like this https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/defs/conftest.py#L1910

And avoid these duplications.

litaotju: And shall blackwell run these tests?

litaotju (changes requested):

tests/integration/defs/disaggregated/test_disaggregated.py
chuangz0: blackwell may not support deepseek v3 lite fp8

litaotju (approved):

pcastonguay:
@chuangz0 why didn't we unwaive the tests in https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/test_lists/waives.txt#L473-L476?

PR #3730, opened 2025-04-21
feat/add latency support for trtllm bench

danielafrimi:
@rakib-hasan 
id like to get a review from you :) 

most of the changes are taken from your PR (found only throughput  support).


tensorrt_llm/bench/benchmark/low_latency.py
```diff
@@ -134,21 +178,71 @@ def latency_command(
...
     medusa_choices = params.pop("medusa_choices")
 
-    # Reporting Options
+    # # Reporting Options
     report_json: Path = params.pop("report_json")
     iteration_log: Path = params.pop("iteration_log")
     iteration_writer = IterationWriter(iteration_log)
 
+    # Initialize the HF tokenizer for the specified model.
+    # ignore_eos = True if runtime_config.decoding_config.decoding_mode == SpeculativeDecodingMode.NONE else False # TODO (dafrimi): nto sure where to locate this line since it requires the runConfig, but tokenizer is used to get the dataset
```
danielafrimi: @rakib-hasan WDYT? not sure where to locate it

tensorrt_llm/bench/benchmark/low_latency.py
```diff
@@ -134,21 +178,71 @@ def latency_command(
...
+
+    if modality is None:
+        # Log dataset info
+        # NOTE: This table is only accurate for non-multimodal models.
+        #       The accurate table for multimodal models will be logged after the benchmark is done.
+        logger.info(metadata.get_summary_for_print())
+
+    # Engine configuration parsing for PyTorch backend
+    kwargs = {}
+    if backend and backend.lower() in ["pytorch", "autodeploy"]:
```
Naveassaf: Id consider factoring this out to some const up top:
```
SUPPORTED_BACKENDS = ["pytorch", "autodeploy", "cpp"]
```
having this as the `--backend`'s choices, and have the default be `cpp`. IMO its confusing to have the default not be an option.

tensorrt_llm/bench/benchmark/low_latency.py
```diff
@@ -134,21 +178,71 @@ def latency_command(
...
+    max_seq_len: int = params.pop("max_seq_len")
+    backend: str = params.get("backend")
+    model_type = get_model_config(model, checkpoint_path).model_type
 
     # Runtime Options
     kv_cache_percent = params.pop("kv_cache_free_gpu_mem_fraction")
     medusa_choices = params.pop("medusa_choices")
 
-    # Reporting Options
+    # # Reporting Options
```
Naveassaf: extra `#`

tensorrt_llm/bench/benchmark/low_latency.py
Naveassaf: It **looks like** `ignore_eos` is only used for getting the `eos_id` and `pad_id`. 

Meanwhile, `eos_id` and `pad_id` are only used for creating the `SamplingParameters`. So why not move lines 198, 200 and 201 to be right before the creating of the `SamplingParams` (line ~290)? Then you could have:
```
ignore_eos = True if runtime_config.decoding_config.decoding_mode == SpeculativeDecodingMode.NONE else False
```
since youve already created the runtime config

tensorrt_llm/bench/benchmark/low_latency.py
```diff
@@ -134,21 +178,71 @@ def latency_command(
...
+
+        if metadata.max_sequence_length > engine_max_seq_len:
+            raise RuntimeError(
+                f"Engine supports a max sequence of {engine_max_seq_len}. Provided "
+                "dataset contains a maximum sequence of "
+                f"{metadata.max_sequence_length}. Please rebuild a new engine to"
+                "support this dataset.")
+
+    exec_settings["model"] = model
+    engine_tokens = exec_settings["settings_config"]["max_num_tokens"]
```
Naveassaf, line 258: Are you sure this is part of the `settings_config` for the PyTorch flow? I was under the impression its not currently supported at the attention backend level, but I definitely might be wrong.

Naveassaf (commented):
Nice change :)

tensorrt_llm/bench/benchmark/low_latency.py
rakib-hasan: IIUC, this reordering is all because of `runtime_config.decoding_config.decoding_mode`, right?
In that case, can we have an extra API just to get that early? That way `runtime_config` can be created once all the extra options are populated.
For the TRT case, it will come from the engine config (we need to read the config twice, which should be fine for now)
For the PyTorch case, it is currently empty for the throughput benchmark (possibly since SD is not used for throughput usecase) and since PyTorch doesn't support the older SD techniques yet. In that case, we can just keep that empty for now like the throughput case and say that it is a limitation until the PyTorch path supports SD techniques.
Does that sound alright?

tensorrt_llm/bench/benchmark/low_latency.py
rakib-hasan: Agree on the refactor. Not just up top, but in some utils so that both `low_latency.py` and `throughput.py` can use the same.
However, for default, I would go with pytorch.

tensorrt_llm/bench/benchmark/low_latency.py
rakib-hasan: AFAIK, `settings_config` is valid for PyTorch flow.

rakib-hasan:
@danielafrimi  The other issue you mentioned, making the tokenizer be optional for multimodal: can you add that change to both low_latency and throughput case? That should be a simple change to add. If it turns out to be complicated, then you can certainly create a new PR.

tensorrt_llm/bench/benchmark/low_latency.py
danielafrimi: done

tensorrt_llm/bench/benchmark/low_latency.py
danielafrimi: @Naveassaf @rakib-hasan reordering since we need the tokenizer for `create_dataset_from_stream` which outputs metadata, which is part of the exec_settings for the runConfig. 

@rakib-hasan commenting out this line with what you suggested. thanks

rakib-hasan: Just to confirm, with that extra API, this reorder shouldn't be needed, right?

danielafrimi:
@rakib-hasan about the issue which makes the tokenizer be optional for multimodal, its can only be optional ion the first phase when we prepare the dataset. now when im looking at this, for real_dataset, for the multimodeal we dont use the tokenzier, but for the synthetic ones it fetches the `vocab_size` for it. 

not sure what if setting the tokenzier to be optional is a good change. WDYT? can add some new files like vocab_size but it wont make the API nicer. 


for running the `trtllm-bench` for the NVILA i initialled the NVILA tokenizer as they did in modeling_vila.py (will create a different PR)

tensorrt_llm/bench/benchmark/low_latency.py
```diff
@@ -179,45 +270,41 @@ def latency_command(
...
-    logger.info(metadata.get_summary_for_print())
-    logger.info("Running experimental latency benchmark.")
+        if runtime_config.backend == 'pytorch':
+            llm = PyTorchLLM(**kwargs)
+        else:
+            llm = LLM(**kwargs)
 
-    llm = None
-    kwargs = runtime_config.get_llm_args()
+        ignore_eos = True if runtime_config.decoding_config.decoding_mode == SpeculativeDecodingMode.NONE else False
```
danielafrimi, line 305: @rakib-hasan what to you think on this change? we needed the runConfig only for getting the eos/pad id to sampling params. in this change we already have the runtime_config for getting this so TRT + pytorch will work as well.

tensorrt_llm/bench/benchmark/low_latency.py
danielafrimi: look on the comment i added (line 288). so we wont need extra API for this case. we still reorder the `exec_settings`, since we fetch need to fetch the setting in the pytorch flow which uses the metadata (from `create_dataset_from_stream`). WDYT?

BTW, checked and it works with pytorch flow as well (for TRT nothing changed)

tensorrt_llm/bench/benchmark/low_latency.py
rakib-hasan: LGTM

tensorrt_llm/bench/benchmark/low_latency.py
symphonylyh, line 1: @FrankD412 there is an offline discussion on whether this low-latency trtllm-bench is still needed. because for DS and Llama4 we just run regular throughput benchmark by setting different concurrency, so small concurrency is naturally low latency scenarios?
it's still relevant for the TRT workflow by turning on low-latency plugins, but for pytorch workflow it's irrelevant, right?

tensorrt_llm/bench/benchmark/utils/general.py
```diff
@@ -22,6 +22,9 @@
     QuantAlgo.NVFP4.value: "fp8",
 }
 
+SUPPORTED_BACKENDS = ["pytorch", "autodeploy", "cpp"]
+PYTHON_SUPPORTED_BACKENDS = ["pytorch", "autodeploy"]
```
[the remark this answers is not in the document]
danielafrimi: done

tensorrt_llm/bench/benchmark/low_latency.py
FrankD412: @symphonylyh -- yeah, this subcommand has become legacy at this point for the TRT workflow.

tensorrt_llm/bench/benchmark/low_latency.py
FrankD412: So a caveat here that was brought to my attention yesterday. Models can have more than one EOS token. For example, Llama 4 has three different EOS tokens which are used in different scenarios and can affect the behavior of speculative decoding. Are we sure we're going to get the token we want from simply asking for the tokenizer's EOS?

tensorrt_llm/bench/benchmark/low_latency.py
FrankD412: Also, for context -- I've been trying to get around to clean up the subcommands. I have an idea for it. Maybe we should have an offline discussion about what I'm thinking? @rakib-hasan @symphonylyh @danielafrimi

danielafrimi:
@rakib-hasan Can we merge it?

rakib-hasan:
Hi @danielafrimi , as Frank mentioned that `latency` side of the benchmark is just legacy at this point, I will let him decide. 

@FrankD412 do you want this PR to be merged or closed?

If merging, @danielafrimi we need to resync and rerun the CI as it has been 2 weeks since the last run.

FrankD412:
@rakib-hasan -- Sorry, I meant to schedule something with you all offline but last week was a bit crazy with travel back from the East Coast.

I'm okay with this being merged as it doesn't break the benchmark and updates it to utilize PyTorch. My only concern here is that a lot of folks are using the `throughput` benchmark for low latency (we even added an EOS option for speculation). If you want to use this you're not in-line with current methodology.

Let me also get something on the books for a sync, I think it would be useful to talk to you all about where I'd like to go with it and maybe we can join forces to help move that forward.

rakib-hasan:
Sounds good. 

I am not sure whether @danielafrimi is planning to use the latency path or it was just for feature parity with the throughput path. If it is the latter, then this PR should be closed and maybe a new PR for deprecating the latency path would reduce future confusions.

danielafrimi:
Hi @rakib-hasan @FrankD412 , regarding the latency part—I needed to benchmark a VLM for both throughput and latency. I'm currently working on supporting quantization of the NVIDIA model using TensorRT-LLM with the PyTorch backend.

My main goal here is to compare the results reported in the paper with those of the model running inside TensorRT-LLM. I created this PR as part of my modifications, building on top of the existing throughput script. That said, I’m not sure how we’ll be able to obtain metrics like TTFT, ITL, and other detailed latency numbers without proper latency support for VLMs—at least for now.

What do you think?

rakib-hasan:
Hey @danielafrimi . Apologies. I wrote a response but forgot to send. 
For latency numbers, you can enable the streaming (`--streaming`) in the throughput command. Would that be sufficient for your use-case? 

If so, we can close this one as `latency` path won't be needed. 

If not, we can certainly merge this and note what is missing so that we can add that as part of the refactor that Frank is going to work on.

FrankD412:
@rakib-hasan -- just following up on this one. What's the current status looking like?

rakib-hasan:
> @rakib-hasan -- just following up on this one. What's the current status looking like?

I don't know. Not sure if @danielafrimi is OOO or not.

danielafrimi:
Hi, sorry for the late response i was OOO. 
didnt try the `streaming` option, but if you think is sufficient, we can pass this PR.

FrankD412:
Another user was wondering on this also, so if you're okay with pushing this to the finish that would be welcomed. :)

danielafrimi:
@rakib-hasan @FrankD412 Hi, rebased the branch. can we approve this PR (review has made before)?

tensorrt_llm/bench/benchmark/throughput.py
```diff
@@ -7,7 +7,6 @@
 import click
 from click_option_group import (MutuallyExclusiveOptionGroup, OptionGroup,
                                 optgroup)
-from huggingface_hub import snapshot_download
```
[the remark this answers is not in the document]
FrankD412: Ok I see the snapshot download was removed -- that has me worried because it's there due to a previous bug with the quant config not being present if the whole snapshot wasn't downloaded.

tensorrt_llm/bench/benchmark/throughput.py
```diff
@@ -293,10 +293,11 @@ def throughput_command(
         logger.info(metadata.get_summary_for_print())
 
     # Engine configuration parsing
-    if backend and backend.lower() in ["pytorch", "_autodeploy"]:
+    if backend and backend.lower() in ALL_SUPPORTED_BACKENDS and backend.lower(
+    ) != "cpp":
         # If we're dealing with a model name, perform a snapshot download to
         # make sure we have a local copy of the model.
-        if checkpoint_path is None:
+        if bench_env.checkpoint_path is None:
```
rakib-hasan, line 300: Re-iterating the comment from the other [PR](https://github.com/NVIDIA/TensorRT-LLM/commit/d0eb47d33a7195dfcb04e83cffd42518e6452bc6#r160748049).

@danielafrimi I ran into an issue with the previous code where the model is already downloaded and the code still tries to re-download since bench_env.checkpoint_path is None even though bench_env.model contains the local path.

Can you please elaborate on why do you think the current code is wrong?

tensorrt_llm/bench/benchmark/low_latency.py
FrankD412: Pytorch is the correct default -- it just got changed [here](https://github.com/NVIDIA/TensorRT-LLM/pull/5312) recently.

danielafrimi:
@FrankD412 will work this week on this PR. thanks

FrankD412 (approved):

Issue #3305, opened 2025-04-06
Enhance the integrated robustness of scaffolding with __init__.py


WeiHaocheng:
I would like to enhance the integrated robustness of scaffolding. 

There are two goals for this work:

- If we want to change the code file structure of scaffolding, we don't want user to change their code about import something from scaffolding. So we list all the api class of scaffolding to __init__.py of scaffolding fold.

- We will periodically move some implementations from user contributed Controller/Task/Worker to the core code of scaffolding. This may cause some code incompatibility issues. For example, there are two user contributed project A and B, project A import a Controller from project B. Now we want to move project B to the core code, then the code of project B also need to be changed. This is something we don't want to see.
My plan is that:
Each project list the APIs they want to expose in the __init__.py file located in the example/scafffolding/contrib. And the example/scafffolding/contrib/__init__.py import * from tensorrt_llm.scaffolding. Then the code of project A should be:
`import XXXController form example.scaffolding.contrib`
if we move XXXController to scaffolding core code, this line will still works.

WeiHaocheng:
After the attempts of the contributors, for better convenience of cross referencing in the contrib project, I would like to move the contrib directory to tensorrt_llm/scaffolding, and just remain example part on /example/scaffolding.

PR #3731, opened 2025-04-21
feat(part 2): Enhance the integrated robustness of scaffolding with _init__.py #3305

WeiHaocheng:
feat(part 2): Enhance the integrated robustness of scaffolding with _init__.py #3305

Move the contrib directory to tensorrt_llm/scaffolding, and just remain example part on /example/scaffolding.

BTW, I found that the default model path in scratch.trt_llm_data is not friendly to contributor, so I change it as the required field.

Also, I found that dynasor does not add the deps into requirement, I create a requirement file for scaffolding and install in setup.py of tensorrt_llm.


tensorrt_llm/scaffolding/requirements.txt
```diff
@@ -0,0 +1,2 @@
+latex2sympy2
+word2number
```
dc3671: why add these two requirements? If these are required by Dynasor, can we move it into Dynasor's directory?

setup.py
```diff
@@ -88,6 +88,11 @@ def has_ext_modules(self):
     Path("requirements-dev-windows.txt"
          if on_windows else "requirements-dev.txt"))
 
+scaffolding_deps, _ = parse_requirements(
+    Path("tensorrt_llm/scaffolding/requirements.txt"))
+
+required_deps.extend(scaffolding_deps)
```
dc3671: I don't think we should directly modify repo's `setup.py` file. The requirements of scaffolding should be installed explicitly, rather than by default for all trtllm users.

dc3671 (approved):

setup.py
WeiHaocheng: OK，how about writing a setup.py for scaffolding contrib?

tensorrt_llm/scaffolding/requirements.txt
WeiHaocheng: It's fine to run Dynasor example, but it is not friendly for using Dynasor by other projects. It seems difficult to install the required dependencies on demand. 
I plan to install all the dependencies required by contrib for now, which seems acceptable in the current situation where there are not so many contrib projects.

dc3671: To follow current contrib ways in `examples/models/contrib`, I recommend to place the requirement in each contrib's directory, like in `examples/scaffolding/contrib/Dynasor`. And add `pip install -r requirement.txt` in README.md.

image

If one want to use Dynasor, he needs to follow README.md in Dynasor directory and install the requirement manually. We can discuss about this offline.

setup.py
dc3671: I think in current stage, we can just told people to install the requirement manually by `pip install -r examples/scaffolding/contrib/Dynasor/requirements.txt` for specific contrib module.

And we should not add setup.py or requirements.txt in tensorrt_llm folder, which is user's `site-packages/tensorrt_llm`. It's kind of post-installation thing.

tensorrt_llm/scaffolding/requirements.txt
WeiHaocheng: OK~This method looks better. If a project want to import Dynasor, just copy the requirements into its requirements.txt. Let me think about it again.

PR #3732, opened 2025-04-21
chore: Cleanup deprecated APIs from LLM-API (part 1/2)

Superjomn:
Remove the following SamplingParams knobs as they are deprecated for several versions:

-  beam_width
-  max_new_tokens
-  min_length
-  num_return_sequences
-  random_seed

This PR is the first part, which contains the minimum C++ change. Part 2 will focus on some remaining changes and some C++ API code deprecation.


cpp/tensorrt_llm/pybind/executor/request.cpp
```diff
@@ -118,47 +117,29 @@ void initRequestBindings(pybind11::module_& m)
...
-            py::arg("length_penalty") = py::none(), py::arg("early_stopping") = py::none(),
-            py::arg("no_repeat_ngram_size") = py::none(), py::arg("num_return_sequences") = py::none(),
-            py::arg("min_p") = py::none(), py::arg("beam_width_array") = py::none())
+            py::arg("top_p_decay") = py::none(), py::arg("seed") = py::none(), py::arg("temperature") = py::none(),
+            py::arg("min_tokens") = py::none(), py::arg("beam_search_diversity_rate") = py::none(),
+            py::arg("repetition_penalty") = py::none(), py::arg("presence_penalty") = py::none(),
+            py::arg("frequency_penalty") = py::none(), py::arg("length_penalty") = py::none(),
+            py::arg("early_stopping") = py::none(), py::arg("no_repeat_ngram_size") = py::none(),
+            py::arg("num_return_sequences") = py::none(), py::arg("min_p") = py::none(),
+            py::arg("beam_width_array") = py::none())
```
syuoni, line 133: Hi @Superjomn ,

Since all deprecated args are removed, we can use the original constructor. You can reference the `OutputConfig` below ([link](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/pybind/executor/request.cpp#L213-L217)). Please help remove Lines 101 and 102. Thanks!

cpp/tensorrt_llm/pybind/executor/request.cpp
```diff
@@ -602,7 +583,6 @@ void initRequestBindings(pybind11::module_& m)
             py::arg("guided_decoding_params") = py::none(), py::arg("language_adapter_uid") = py::none())
         .def_property_readonly("input_token_ids", &tle::Request::getInputTokenIds)
         .def_property_readonly("max_tokens", &tle::Request::getMaxTokens)
-        .def_property_readonly("max_new_tokens", &tle::Request::getMaxNewTokens)
```
syuoni, line 605: We need to update the constructor, too. Similarly, please revert to the original constructor and removes Lines 538 and 539. Thanks!

examples/llm-api/llm_inference_customize.py
```diff
@@ -35,7 +35,10 @@ def main():
     ]
 
     # With SamplingParams, you can customize the sampling strategy, such as beam search, temperature, and so on.
-    sampling_params = SamplingParams(temperature=0.8, top_p=0.95, beam_width=4)
+    sampling_params = SamplingParams(temperature=0.8,
+                                     top_p=0.95,
+                                     best_of=4,
```
syuoni, line 40: I suppose we should use `n=4`, otherwise using this `sampling_params` returns one generation, right?

tensorrt_llm/bench/benchmark/throughput.py
```diff
@@ -320,7 +320,8 @@ def throughput_command(
 
         sampling_params = SamplingParams(end_id=-1,
                                          pad_id=-1,
-                                         beam_width=beam_width)
+                                         n=beam_width,
+                                         use_beam_search=beam_width > 1)
```
syuoni, line 324: We can simplify it as `use_beam_search=True` here, since beam search with width=1 is same to greedy search.

tensorrt_llm/llmapi/_perf_evaluator.py
```diff
@@ -375,7 +375,8 @@ def run(self, end_id: int = -1, beam_width: int = 1) -> Report:
         sampling_params = SamplingParams(
             end_id=end_id,
             pad_id=end_id,
-            beam_width=beam_width,
+            n=beam_width,
+            use_beam_search=beam_width > 1,
```
syuoni, line 379: Same here, `use_beam_search=True`.

tests/unittest/llmapi/test_executor.py
```diff
@@ -88,7 +88,9 @@ def test_generation_bs2(llama_7b_bs2_path: Path):
             executor_config=tllm.ExecutorConfig(max_beam_width=2)) as executor:
         result = executor.generate(prompt_token_ids,
                                    sampling_params=SamplingParams(
-                                       max_tokens=max_tokens, beam_width=2))
+                                       max_tokens=max_tokens,
+                                       best_of=2,
```
syuoni, line 92: Since default `n=1`, so this means returning one generation. This should not be same as the original code.

tensorrt_llm/sampling_params.py
```diff
@@ -300,9 +269,8 @@ def _validate(self):
         if self.best_of is not None:
             if self.best_of > 1 and self.best_of < self.n:
                 raise ValueError(
-                    f'In beam search, beam_width ({self.beam_width}) must be '
-                    f'greater than or equal to num_return_sequences '
-                    f'({self.num_return_sequences}).')
+                    f'In beam search, beam_width ({self.best_of}) must be '
+                    f'greater than or equal to n ({self.n}).')
```
syuoni, line 273: Seems it is not specific to beam search? How about `f"best_of ({self.best_of}) must be greater than or equal to n ({self.n})."`

tensorrt_llm/bench/benchmark/throughput.py
Superjomn: Referencing the current [implementation](https://github.com/superjomn/TensorRT-LLM/blob/57944206ba403acf6ac2b3834b33ad15464ee6d5/tensorrt_llm/sampling_params.py#L254), when `beam_width=1`, the `use_beam_search` is False?

cpp/tensorrt_llm/pybind/executor/request.cpp
Superjomn: Updated accordingly, one exception is the `num_return_sequences`, I saw the Request constructor in the cpp runtime contains this, I am hesitating to remove this since that will greatly break the backward-compability (C++ function's arglist only support optional)

tensorrt_llm/sampling_params.py
Superjomn: Good suggestion, updated.

tensorrt_llm/bench/benchmark/throughput.py
syuoni: Right.

cpp/tensorrt_llm/pybind/executor/request.cpp
syuoni: The `Request::numReturnSequences` was deprecated in favor of `SamplingConfig::numReturnSequences`. This PR also removes the getter/setter of `Request::numReturnSequences` . I think we can remove `Request::numReturnSequences` completely.

syuoni (approved):
LGTM, thanks!

cpp/tensorrt_llm/pybind/executor/request.cpp
Superjomn: Yeah, I will do this in a subsequent PR.

examples/llm-api/llm_inference_customize.py
Superjomn: Yes you are right, updated.

tensorrt_llm/llmapi/_perf_evaluator.py
Superjomn: Same as above, in the current implementation, `use_beam_search=beam_width>1`.

Superjomn:
/bot skip --comment "both single-gpu and multi-gpu test pass"

PR #3735, opened 2025-04-21
Fix TLLM_CHECK(isLeaf()); assertion

akhoroshev:
When stress testing I get the error [TLLM_CHECK(isLeaf());](https://github.com/NVIDIA/TensorRT-LLM/blob/c2ffce7dbd283000255948ae04792f13f13e05dc/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp#L301)

Version is https://github.com/NVIDIA/TensorRT-LLM/tree/c2ffce7dbd283000255948ae04792f13f13e05dc

After studying, I realized that sometimes the LRUEvictionPolicy tree violates invariants about leaf nodes in mFreeQueues.

Solution works fine for me.

NOTE: I have not tested offloading to a secondary pool as I don't use such a feature in prod.

### DIFF
--- a/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
+++ b/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
@@ -130,7 +130,6 @@ KVCacheBlock::KVCacheBlock(IdType blockId, tk::KVCacheIndex blockIdx)
     , mRefCount(0)
     , mSchedulingRefCount(0)
     , mPrevBlock(nullptr)
-    , mFreeBlockIterator(std::nullopt)
     , mIsFull{false}
     , mPriority{executor::KvCacheRetentionConfig::kDefaultRetentionPriority}
     , mDurationMs{std::nullopt}
@@ -778,6 +777,10 @@ BlockPtr WindowBlockManager::getFreeBlock(
     ++mAllocTotalBlocks;
     if (!block->getUniqueTokens().empty() && canOffload && mEvictionPolicy->getNumFreeBlocks(kSecondaryLevel) > 0)
     {
+        // Since `claimBlock` resets link between parentBlock and block
+        // we need to save origin parent block.
+        const auto parentBlock = block.getPrevBlock(); 
+
         // If we're swapping a block to secondary memory, maintain the prior priority values.
         mEvictionPolicy->claimBlock(block);
         // Offload block in primary memory before repurposing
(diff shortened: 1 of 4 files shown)


cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -778,6 +777,10 @@ BlockPtr WindowBlockManager::getFreeBlock(
     ++mAllocTotalBlocks;
     if (!block->getUniqueTokens().empty() && canOffload && mEvictionPolicy->getNumFreeBlocks(kSecondaryLevel) > 0)
```
akhoroshev, line 778: I have not tested offloading to a secondary pool as I don't use such a feature in prod.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -778,6 +777,10 @@ BlockPtr WindowBlockManager::getFreeBlock(
...
     {
+        // Since `claimBlock` resets link between parentBlock and block
+        // we need to save origin parent block.
+        const auto parentBlock = block.getPrevBlock();
```
akhoroshev, line 782: I have not tested offloading to a secondary pool as I don't use such a feature in prod.

akhoroshev:
@nv-guomingz @kaiyux @byshiue

akhoroshev:
@juney-nvidia

kaiyux:
@Funatiq @thorjohnsen Can you help review? Thanks!

thorjohnsen (commented):
First of all, I apologize for not responding to this PR sooner, I am still learning how to keep track of all my assigned tasks now that we are working towards a public repository. 

Second: There is a fix for this issue in 0.20, which will be released to the public shortly. 

Third: I don't think your proposed fix will work properly with offloading enabled, but this is expected since you mention you haven't tested with offloading. The meaning of the label "leaf block" is different in eviction policy and window manager classes, in the former context it means any block that has no descendants at or above it's cache level, in the latter it means any block that has no descendants period. This is confusing but it stems from the fact that eviction policy needs to keep track of all blocks that can potentially be freed at all cache levels while window block manager actually frees the block(s). Blocks that can potentially be freed include blocks with all descendants in lower cache levels, but when one of those blocks is actually freed by window manager all descendants become orphaned and so they must be freed too.

Funatiq:
@akhoroshev the fix has been merged into main. Could you please verify that your stress test is working now?

poweiw:
Closing as no activities from requester for +10 days. Feel free to reopen 👍

PR #3737, opened 2025-04-21
fix: use cuKernelGetAttribute for wider CUDA compatibility

lucifer1004:


litaotju (approved):

lucifer1004:
Needs further check to validate working on CUDA 12.2

lucifer1004:
@litaotju I am going to close this PR since:
1. The current change is proved not enough for TRTLLM to work with CUDA 12.2. While there is a way to work around, that would be much more complex. We would have to store the kernel name to disk, otherwise we would need to use `cuobjdump` to get the exact kernel name at runtime. Either solution would contradict with our current design in some way.
2. The `undefined symbol` issue can be solved with `cuda-compat`. This can work even on an old driver which does not support CUDA >= 12.3.

PR #3743, opened 2025-04-21
feat: Add head size 72 support for QKV Preprocessing kernel

qixiang-99:
Head size 72 fmha kernels were added previously but only cross attention cases are tested. So we can reuse the added fmha kernels for head size 72 attention by removing the cross attention condition in checking.

However, with above changes, we also need to add head size 72 QKV Preprocessing kernel (v2) to support features like RoPE etc in the furture.

What this PR does: 
1. remove the cross attention condition for head size 72 fmha path
2. add head size 72 QKV Preprocessing kernel


symphonylyh:
is my understanding correctly that you're using the unfused attn kernel for the head size 72, and later on will switch to fused mha / trtllm-gen when they added 72 support?

symphonylyh (approved):

PR #3744, opened 2025-04-21
infra: add conan

tburt-nv:
This MR integrates Conan into the build system, so that it can be used to fetch dependencies in future changes.

Also installs all requirements-dev.txt inside a virtualenv instead of the system, since some of Conan's dependencies may conflict with the system packages. Virtualenv is used instead of venv because the triton server backend container has only virtualenv installed. This also allows developers to cache the requirements-dev.txt packages between container launches.


tburt-nv:
Many of these changes are identical to the changes in https://github.com/NVIDIA/TensorRT-LLM/pull/3282

nv-guomingz (approved):
LGTM

tburt-nv:
/bot skip --comment "union of pipelines is passing"

PR #3747, opened 2025-04-21
fix: Fix CUDA graphs when tp_size > 1

mikeiovine:
If the different ranks come up with different KV cache token estimates, you can get illegal memory accesses during the warm up stage. Fix by doing an MPI allreduce.

Also add CUDA graph to llama4 test case to cover this. This seems to only happen to llama4 for some reason.


mikeiovine:
Apparently we already have an allreduce inside the KV cache manager itself. This seemed to fix the issue but apparently it's just not deterministic. Still investigating, moving this PR to draft mode for now.

mikeiovine:
Close as this issue is definitely being caused by something else

PR #3759, opened 2025-04-22
infra: install Triton in the base image

Tabrizian:
It will build Triton only in the CI script to avoid increasing the build time.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -41,6 +41,7 @@ GIT_COMMIT         ?= $(shell git rev-parse HEAD)
 TRT_LLM_VERSION    ?= $(shell grep '^__version__' ../tensorrt_llm/version.py | grep -o '=.*' | tr -d '= "')
 GITHUB_MIRROR      ?=
 PYTHON_VERSION     ?=
+BUILD_TRITON       ?= 0
 
 define add_local_user
 	docker build \
@@ -79,6 +80,7 @@ endef
 		$(if $(GIT_COMMIT), --build-arg GIT_COMMIT="$(GIT_COMMIT)") \
 		$(if $(GITHUB_MIRROR), --build-arg GITHUB_MIRROR="$(GITHUB_MIRROR)") \
 		$(if $(PYTHON_VERSION), --build-arg PYTHON_VERSION="$(PYTHON_VERSION)") \
+		$(if $(BUILD_TRITON), --build-arg BUILD_TRITON="$(BUILD_TRITON)") \
 		$(if $(STAGE), --target $(STAGE)) \
 		--file Dockerfile.multi \
 		--tag $(IMAGE_WITH_TAG) \
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -41,7 +41,7 @@ GIT_COMMIT         ?= $(shell git rev-parse HEAD)
 TRT_LLM_VERSION    ?= $(shell grep '^__version__' ../tensorrt_llm/version.py | grep -o '=.*' | tr -d '= "')
 GITHUB_MIRROR      ?=
 PYTHON_VERSION     ?=
-INSTALL_TRITON     ?= 0
+INSTALL_TRITON     ?= 1
 
 define add_local_user
 	docker build \
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -41,7 +41,7 @@ GIT_COMMIT         ?= $(shell git rev-parse HEAD)
 TRT_LLM_VERSION    ?= $(shell grep '^__version__' ../tensorrt_llm/version.py | grep -o '=.*' | tr -d '= "')
 GITHUB_MIRROR      ?=
 PYTHON_VERSION     ?=
-BUILD_TRITON       ?= 0
+INSTALL_TRITON     ?= 0
 
 define add_local_user
 	docker build \
@@ -80,7 +80,7 @@ endef
 		$(if $(GIT_COMMIT), --build-arg GIT_COMMIT="$(GIT_COMMIT)") \
 		$(if $(GITHUB_MIRROR), --build-arg GITHUB_MIRROR="$(GITHUB_MIRROR)") \
 		$(if $(PYTHON_VERSION), --build-arg PYTHON_VERSION="$(PYTHON_VERSION)") \
-		$(if $(BUILD_TRITON), --build-arg BUILD_TRITON="$(BUILD_TRITON)") \
+		$(if $(INSTALL_TRITON), --build-arg INSTALL_TRITON="$(INSTALL_TRITON)") \
 		$(if $(STAGE), --target $(STAGE)) \
 		--file Dockerfile.multi \
 		--tag $(IMAGE_WITH_TAG) \
--- a/docker/common/install_triton.sh
+++ b/docker/common/install_triton.sh
@@ -1,72 +1,43 @@
 #!/bin/bash
 
 set -ex
-TRITON_VER="r25.03"
-
-for i in "$@"; do
-    case $i in
-        --TRITON_VER=?*) TRITON_VER="${i#*=}";;
-        *) ;;
-    esac
-    shift
-done
 
 install_boost() {
-	# Install boost version >= 1.78 for boost::span
-	# Current libboost-dev apt packages are < 1.78, so install from tar.gz
-	wget -O /tmp/boost.tar.gz \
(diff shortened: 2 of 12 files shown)


docker/Makefile
```diff
@@ -41,6 +41,7 @@ GIT_COMMIT         ?= $(shell git rev-parse HEAD)
 TRT_LLM_VERSION    ?= $(shell grep '^__version__' ../tensorrt_llm/version.py | grep -o '=.*' | tr -d '= "')
 GITHUB_MIRROR      ?=
 PYTHON_VERSION     ?=
+BUILD_TRITON       ?= 0
```
chzblych, line 44: Should the default value be `1`? `BUILD_TRITON` is set to `1` in all the images used by CI.

People cannot get the same env locally if building the docker image from scratch.

chzblych, line 44: Should the default value be `1`? `BUILD_TRITON` is set to `1` explicitly in all the images used by CI.

People cannot get the same development env locally if building the docker image from scratch with default settings.

docker/common/install_triton.sh
```diff
@@ -0,0 +1,72 @@
...
+    rm -rf ./build
+}
+
+# Install Triton only if base image in Ubuntu
+ID=$(grep -oP '(?<=^ID=).+' /etc/os-release | tr -d '"')
+if [ "$BUILD_TRITON" == "1" ]; then
+    if [ "$ID" == "ubuntu" ]; then
+        install_triton
+    else
+        echo "Skip Triton installation for non-Ubuntu base image"
```
chzblych, line 70: We use the RockyLinux 8 image to generate the public Linux x86 wheels. Is Triton not needed for generating the public wheels?

If Ubuntu is the only requirement, perhaps another approach is to copy the pre-compiled Triton binaries from the DLFW Triton image. For examples, https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/dockerfile/Dockerfile.trt_llm_backend#L8.

docker/common/install_triton.sh
```diff
@@ -0,0 +1,72 @@
...
+for i in "$@"; do
+    case $i in
+        --TRITON_VER=?*) TRITON_VER="${i#*=}";;
+        *) ;;
+    esac
+    shift
+done
+
+install_boost() {
+	# Install boost version >= 1.78 for boost::span
```
chzblych, line 15: The code indentation looks incorrect.

docker/common/install_triton.sh
```diff
@@ -0,0 +1,72 @@
...
+        libnuma-dev \
+        python3-build \
+        libb64-dev \
+        libarchive-dev
+
+    ARCH=$(uname -m)
+    if [ "$ARCH" = "amd64" ];then ARCH="x86_64";fi
+    if [ "$ARCH" = "aarch64" ];then ARCH="sbsa";fi
+
+	curl -o /tmp/cuda-keyring.deb \
```
chzblych, line 37: The code indentation looks incorrect.

docker/Makefile
Tabrizian: The goal was to not increase the build time of the image. That is why it is turned off in the Makefile.

docker/common/install_triton.sh
Tabrizian: I think I can copy the Triton binaries instead.

docker/common/install_triton.sh
```diff
@@ -0,0 +1,72 @@
...
+      --repo-tag=core:$TRITON_VER --repo-tag=backend:$TRITON_VER \
+      --repo-tag=thirdparty:$TRITON_VER --backend=python:$TRITON_VER \
+      --repoagent=checksum:$TRITON_VER --enable-gpu --enable-stats --enable-metrics \
+      --enable-logging --enable-cpu-metrics \
+      --no-container-build --build-dir=`pwd`/build \
+      --endpoint grpc --endpoint http --endpoint sagemaker --endpoint vertex-ai \
+      --filesystem gcs --filesystem s3 --filesystem azure_storage --backend=ensemble:$TRITON_VER
+
+    mv ./build/opt/tritonserver/ /opt/tritonserver/
+    rm -rf /tmp/boost_1_80_0
```
achartier, line 59: Nit: I would move boost cleanup inside of `install_boost`

docker/common/install_triton.sh
achartier: I'm not sure if there is an elegant way to do that directly in Dockerfile, but it'd be ideal to just copy from the Triton image.

docker/common/install_triton.sh
```diff
@@ -42,31 +35,19 @@ install_triton() {
...
-    python3 build.py --repo-tag=common:$TRITON_VER \
-      --repo-tag=core:$TRITON_VER --repo-tag=backend:$TRITON_VER \
-      --repo-tag=thirdparty:$TRITON_VER --backend=python:$TRITON_VER \
-      --repoagent=checksum:$TRITON_VER --enable-gpu --enable-stats --enable-metrics \
-      --enable-logging --enable-cpu-metrics \
-      --no-container-build --build-dir=`pwd`/build \
-      --endpoint grpc --endpoint http --endpoint sagemaker --endpoint vertex-ai \
-      --filesystem gcs --filesystem s3 --filesystem azure_storage --backend=ensemble:$TRITON_VER
 
     mv ./build/opt/tritonserver/ /opt/tritonserver/
```
achartier, line 39: Is that still required?

achartier (approved):

docker/common/install_triton.sh
Tabrizian: Fixed indentation.

docker/common/install_triton.sh
Tabrizian: It is not required. I have removed it.

docker/common/install_triton.sh
Tabrizian: Moved to `install_boost`.

docker/common/install_triton.sh
```diff
@@ -0,0 +1,43 @@
...
+    libnuma-dev \
+    python3-build \
+    libb64-dev \
+    libarchive-dev
+
+  ARCH=$(uname -m)
+  if [ "$ARCH" = "amd64" ];then ARCH="x86_64";fi
+  if [ "$ARCH" = "aarch64" ];then ARCH="sbsa";fi
+
+  curl -o /tmp/cuda-keyring.deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/${ARCH}/cuda-keyring_1.0-1_all.deb && apt install /tmp/cuda-keyring.deb && rm /tmp/cuda-keyring.deb && apt-get update && apt-get install -y datacenter-gpu-manager=1:3.2.6
```
chzblych, line 27: DLFW images are based on Ubuntu 24.04 currently.

chzblych, line 27: Do we need to pin to a specific `datacenter-gpu-manager=1:3.2.6` version?

docker/common/install_triton.sh
```diff
@@ -0,0 +1,43 @@
...
+ID=$(grep -oP '(?<=^ID=).+' /etc/os-release | tr -d '"')
+if [ "$INSTALL_TRITON" == "1" ]; then
+  if [ "$ID" == "ubuntu" ]; then
+    install_triton_deps
+  else
+    rm -rf /opt/tritonserver
+    echo "Skip Triton installation for non-Ubuntu base image"
+  fi
+else
+  rm -rf /opt/tritonserver
```
chzblych, line 42:
```suggestion
  rm -rf /opt/tritonserver
  echo "Skip Triton installation when INSTALL_TRITON is set to 0"
```

chzblych (approved):

docker/Makefile
chzblych: Since now we don't build Triton, does it make sense to make the default value be `1`?

achartier: What is the container size difference?

docker/common/install_triton.sh
Tabrizian: I used the version that Triton is using. I believe using a different version can cause issues.

docker/Makefile
Tabrizian: It looks like there is only 200MBs of size difference so I have enabled it by default.

Tabrizian: It is ~1.5GB larger because of DCGM. I enabled it by default for now. We can disable it if it causes issues.

PR #3768, opened 2025-04-22
chore: Move opencv-python import inside load_video() function

AlessioNetti:
This pull request implements a small change to render the `opencv-python` package effectively optional. 

This is, in fact, used only in `inputs/utils.py`, and moving the `cv2` import inside of the `load_video()` function is enough to hide the dependency, if video handling is not needed. As we employ TRT-LLM exclusively for text-based use cases, this allows us to save a little bit of storage for our images.

### DIFF
--- a/tensorrt_llm/inputs/utils.py
+++ b/tensorrt_llm/inputs/utils.py
@@ -1,6 +1,5 @@
 from typing import List, Union
 
-import cv2
 import numpy as np
 import requests
 import torch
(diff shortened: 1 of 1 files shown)


Funatiq (changes requested):
Thanks for the PR, @AlessioNetti! Please [sign your commits](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CONTRIBUTING.md#signing-your-work) to pass the DCO check.

AlessioNetti:
> Thanks for the PR, @AlessioNetti! Please [sign your commits](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CONTRIBUTING.md#signing-your-work) to pass the DCO check.

Sorry @Funatiq, missed this the first time around. Added the sign-off just now. Thanks for taking a look!

Funatiq (approved):

PR #3772, opened 2025-04-22
Add pre-download of checkpoint before benchmark.

FrankD412:
There is an issue when using just a model name with NVIDIA quantized checkpoints where when running the benchmark without the model cached the wrong model size is calculated. The issue stems from the NVIDIA checkpoints using the `hf_quant_config.json` to determine quantization.

With regular Huggingface checkpoints, there is no quantization information (or in the case of Deepseek they've added it to the model `config.json` file). The issue manifests for NVIDIA checkpoints because `trtllm-bench` only pulls the tokenizer and the config using the respective API calls in `transformers`. Neither of these calls downloads the `hf_quant_config.json` so when the TRT-LLM equivalent model configuration class tries to determine quantization it reports that as `None` (which is FP16). To fix this issue, `trtllm-bench` now performs a `from_pretrained` call on the whole model in the root benchmark command before calling a subcommand to make sure that the checkpoint is downloaded.


Superjomn (approved):
LGTM

FrankD412:
@Superjomn @kaiyux -- this PR is good to go. I tested it locally and it seems to work when the checkpoint is missing and does so ahead of the tuning.

kaiyux (approved):

PR #3663, opened 2025-04-17
Added NemotronH to PyTorch supported models

vegaluisjose:

PR #3781, opened 2025-04-23
chore: Fix KV cache block reuse flag name in quickstart_advanced

mikeiovine:
`quickstart_advanced.py` has KV cache reuse on by default. You can turn it off with a flag. The problem is that the name of the flag is flipped: running `quickstart_advanced.py --kv_cache_enable_block_reuse` will actually disable block reuse! This PR fixes this naming mishap.


hlu1 (approved):

examples/pytorch/quickstart_advanced.py
```diff
@@ -62,9 +62,9 @@ def add_llm_args(parser):
 
     # KV cache
     parser.add_argument('--kv_cache_dtype', type=str, default='auto')
-    parser.add_argument('--kv_cache_enable_block_reuse',
-                        default=True,
-                        action='store_false')
+    parser.add_argument('--kv_cache_disable_block_reuse',
```
QiJune: There is another MR which does similar thing: https://github.com/NVIDIA/TensorRT-LLM/pull/3663. How about renaming it to  `disable_kv_cache_reuse`?

mikeiovine: Sure, we can use that name instead.

PR #3783, opened 2025-04-23
fix: Reduce memory usage in fused moe op associated with AutoTuning.

hyukn:
* Replace the pre-defined bucket sizes with a generating function based on the tune_max_num_tokens in fused moe op tuning.
* Add free_memory logic of workspace in min_latency_mode fused moe path.


hyukn:
Retarget to release/0.19 with this PR https://github.com/NVIDIA/TensorRT-LLM/pull/3793.
Thus, close this one.

PR #3795, opened 2025-04-23
chore: make IPC port bounding atomic for slurm

Superjomn:
Before, the `trtllm-llmapi-launch` script determines the address of the IPC between RemoteMpiCommSessionServer/Client and then passes it to both server and client. This is not stable, as the port may be occupied before the server binds it. This PR changes this to be in an atomic style. The new workflow is as follows:

1. The RemoteMpiCommSessionClient will bind an empty port, and write the address (port + HMAC key) to a temporary file, this stage is **atomic**
2. The RemoteMpiCommSessionServer will wait until the address file is ready
3. The RemoteMpiCommSessionServer will read the address file and connect tothe  RemoteMpiCommClient's socket

### DIFF
--- a/tensorrt_llm/llmapi/mpi_session.py
+++ b/tensorrt_llm/llmapi/mpi_session.py
@@ -1,5 +1,6 @@
 import abc
-import itertools
+import base64
+import json
 import os
 import socket
 import sys
@@ -263,16 +264,26 @@ class RemoteMpiCommSessionClient():
     RemoteMpiCommSessionClient is a variant of MpiCommSession that is used to connect to a remote MPI pool.
     '''
 
-    def __init__(self, addr: str, hmac_key: Optional[bytes] = None):
+    def __init__(
+        self,
+        *,
+        addr: Optional[tuple[str, Optional[bytes]]] = None,
+        addr_file: Optional[str] = None,
+    ):
+        '''
+        Args:
+            addr: The address of the remote MPI pool.
+            addr_file: The file to write the address of the remote MPI pool to share with other intra-node processes.
+        '''
         # FIXME: this is a hack to avoid circular import, resolve later
         from tensorrt_llm.executor.ipc import ZeroMqQueue
-        self.addr = addr
+        self._is_shutdown = False
         print_colored_debug(
             f"RemoteMpiCommSessionClient connecting to {addr}\n", "yellow")
-        self.queue = ZeroMqQueue((addr, hmac_key),
-                                 is_server=False,
-                                 use_hmac_encryption=bool(hmac_key))
-        self._is_shutdown = False
+        self.queue = ZeroMqQueue(addr, is_server=True, use_hmac_encryption=True)
+        self.addr_file = addr_file
+        self.write_address_to_shared_file(
+        )  # for client process to get the address
 
     def submit(self, task: Callable[..., T], *args, **kwargs) -> list:
         if self._is_shutdown:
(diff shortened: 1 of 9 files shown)


tensorrt_llm/llmapi/trtllm-llmapi-launch
```diff
@@ -10,8 +10,9 @@ log_stderr "mpi_rank: $mpi_rank"
 
 pid=$(ps -o pid= -p $$ | tr -d ' ')
 
-# Tell TRTLLM to spawn a additional process for the Proxy
-export TLLM_SPAWN_PROXY_PROCESS=1
+# A file to share the IPC address between the server and client
+export TLLM_SPAWN_PROXY_PROCESS_ADDR_FILE=/tmp/llmapi-proxy-addr-${pid}
+rm -f $TLLM_SPAWN_PROXY_PROCESS_ADDR_FILE
```
litaotju, line 15: if we use a tmpfs like /dev/shm, not /tmp. Maybe you don't need to delete it? Because it won't exists?

tests/unittest/llmapi/test_mpi_session.py
```diff
@@ -53,27 +52,12 @@ def run_client(server_addr, values_to_process):
 @pytest.mark.skip(reason="https://nvbugspro.nvidia.com/bug/5179666")
```
litaotju, line 52: This still fails?

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -263,28 +264,54 @@ class RemoteMpiCommSessionClient():
...
         from tensorrt_llm.executor.ipc import ZeroMqQueue
-        self.addr = addr
+        self._is_shutdown = False
         print_colored_debug(
             f"RemoteMpiCommSessionClient connecting to {addr}\n", "yellow")
-        self.queue = ZeroMqQueue((addr, hmac_key),
-                                 is_server=False,
-                                 use_hmac_encryption=bool(hmac_key))
-        self._is_shutdown = False
+        self.queue = ZeroMqQueue(addr, is_server=True, use_hmac_encryption=True)
```
litaotju, line 283: Why the "is_server" is changed from false to true?
Its "client" process here right?

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -318,10 +345,9 @@ def __init__(self,
                  is_comm: bool = False):
         # FIXME: this is a hack to avoid circular import, resolve later
         from tensorrt_llm.executor.ipc import ZeroMqQueue
-        self.addr = addr
         self.queue = ZeroMqQueue((addr, hmac_key),
-                                 is_server=True,
-                                 use_hmac_encryption=bool(hmac_key))
+                                 use_hmac_encryption=bool(hmac_key),
+                                 is_server=False)
```
litaotju, line 350: What's the reason of switching the server/client concept here?

Its kind of confusing to me.

tensorrt_llm/llmapi/mpi_session.py
```diff
@@ -263,28 +264,54 @@ class RemoteMpiCommSessionClient():
...
+    def write_address_to_shared_file(self):
+        # This file will be deleted when client is destroyed, that is sufficient for client to read it during setup
+        with open(self.addr_file, "w") as f:
+            ip_addr, hmac_key = self.queue.address
+            if hmac_key is not None:
+                hmac_key = base64.b64encode(hmac_key).decode('ascii')
+            json.dump((ip_addr, hmac_key), f)
+
+    @staticmethod
+    def read_address_from_shared_file(path: str):
```
litaotju, line 308: Maybe delete it here after the file is used, such that you dont need to `rm` in a shell.

litaotju (changes requested):

PR #3823, opened 2025-04-24
chore: Partition LlmArgs into TorchLlmArgs and TrtLlmArgs

Superjomn:
This is the first step for landing the TorchLlmArgs, Here is the plan:

1. Partition the LlmArgs into TorchLlmArgs, TrtLlmArgs (the current stage)
2. Expand the PytorchConfig into TorchLlmArgs
3. Replace ExecutorConfig in PyT with TorchLlmArgs
4. Rename TorchLlmArgs to LlmArgs eventually when it is ready to be the default.

This PR also refined the YAML support in LlmArgs with the Pydantic builtin support in `update_llm_args_with_extra_options`.

### DIFF
--- a/tensorrt_llm/llmapi/llm.py
+++ b/tensorrt_llm/llmapi/llm.py
@@ -10,6 +10,7 @@ from typing import Any, List, Literal, Optional, Sequence, Union
 from tqdm import tqdm
 from transformers import PreTrainedTokenizerBase
 
+from tensorrt_llm.builder import BuildConfig
 from tensorrt_llm.inputs.data import TextPrompt
 from tensorrt_llm.inputs.registry import DefaultInputProcessor
 
@@ -165,8 +166,9 @@ class LLM:
             # Due to the Executor can only accept a engine path, we need to save the engine to a directory
             self._engine_dir: Optional[Path] = None
             self._executor: Optional[GenerationExecutor] = None
-            self._workspace = tempfile.TemporaryDirectory(
-                suffix="-llm-workspace", dir=self.args.workspace)
+            if self._on_trt_backend:
+                self._workspace = tempfile.TemporaryDirectory(
+                    suffix="-llm-workspace", dir=self.args.workspace)
 
             self._hf_model_dir: Optional[Path] = None
 
@@ -185,7 +187,7 @@ class LLM:
 
     @property
     def workspace(self) -> Path:
-        return Path(self._workspace.name)
+        return Path(self._workspace.name) if self._on_trt_backend else None
 
     def generate(
         self,
@@ -292,9 +294,12 @@ class LLM:
         """
         sampling_params = self._prepare_sampling_params(sampling_params)
 
-        if sampling_params.n > self.args.build_config.max_batch_size:
+        max_batch_size = self.args.max_batch_size
+        max_batch_size = max_batch_size or self.args.build_config.max_batch_size
+
+        if sampling_params.n > max_batch_size:
             raise ValueError(
-                f"SamplingParams.n ({sampling_params.n}) should not exceed max_batch_size ({self.args.build_config.max_batch_size})"
+                f"SamplingParams.n ({sampling_params.n}) should not exceed max_batch_size ({max_batch_size})"
             )
 
         inputs = prompt_inputs(inputs)
@@ -535,11 +540,19 @@ class LLM:
                                                       self.tokenizer)
         self.tokenizer = self.input_processor.tokenizer
 
-        max_batch_size = self.args.max_batch_size or self.args.build_config.max_batch_size
-        max_num_tokens = self.args.max_num_tokens or self.args.build_config.max_num_tokens
-        max_seq_len = self.args.max_seq_len or self.args.build_config.max_seq_len
+        max_batch_size = self.args.max_batch_size
+        max_num_tokens = self.args.max_num_tokens
+        max_seq_len = self.args.max_seq_len
+
+        build_config = self.args.build_config if self._on_trt_backend else BuildConfig(
+        )
+
+        max_batch_size = max_batch_size or build_config.max_batch_size
+        max_num_tokens = max_num_tokens or build_config.max_num_tokens
+        max_seq_len = max_seq_len or build_config.max_seq_len
+
         executor_config = tllm.ExecutorConfig(
-            max_beam_width=self.args.build_config.max_beam_width,
+            max_beam_width=self.args.max_beam_width,
             scheduler_config=PybindMirror.maybe_to_pybind(
                 self.args.scheduler_config),
             batching_type=PybindMirror.maybe_to_pybind(self.args.batching_type)
@@ -565,7 +578,7 @@ class LLM:
         if self.args.peft_cache_config is not None:
             executor_config.peft_cache_config = PybindMirror.maybe_to_pybind(
                 self.args.peft_cache_config)
-        elif self.args.build_config.plugin_config.lora_plugin:
+        elif self._on_trt_backend and self.args.build_config.plugin_config.lora_plugin:
             engine_config = EngineConfig.from_json_file(self._engine_dir /
                                                         "config.json")
             lora_config = engine_config.build_config.lora_config
@@ -593,10 +606,10 @@ class LLM:
         executor_config.normalize_log_probs = self.args.normalize_log_probs
         executor_config.enable_chunked_context = self.args.enable_chunked_prefill
         executor_config.max_beam_width = self.args.max_beam_width or self.args.build_config.max_beam_width
-        if self.args.extended_runtime_perf_knob_config is not None:
+        if self._on_trt_backend and self.args.extended_runtime_perf_knob_config is not None:
             executor_config.extended_runtime_perf_knob_config = PybindMirror.maybe_to_pybind(
                 self.args.extended_runtime_perf_knob_config)
-        if self.args.cache_transceiver_config is not None:
+        if self._on_trt_backend and self.args.cache_transceiver_config is not None:
             executor_config.cache_transceiver_config = PybindMirror.maybe_to_pybind(
                 self.args.cache_transceiver_config)
         from tensorrt_llm._torch.pyexecutor.config import update_executor_config
@@ -605,7 +618,8 @@ class LLM:
             backend=self.args.backend,
             pytorch_backend_config=self.pytorch_backend_config,
             mapping=self.args.parallel_config.to_mapping(),
-            build_config=self.args.build_config,
+            build_config=self.args.build_config
+            if self._on_trt_backend else None,
             speculative_config=self.args.speculative_config,
             hf_model_dir=self._hf_model_dir,
             trt_engine_dir=self._engine_dir,
@@ -613,8 +627,9 @@ class LLM:
             max_seq_len=max_seq_len)
         executor_config.llm_parallel_config = self.args.parallel_config
         return_logits = self.args.gather_generation_logits or (
-            self.args.build_config
+            self._on_trt_backend and self.args.build_config
             and self.args.build_config.gather_context_logits)
+
         self._executor = self._executor_cls.create(
             self._engine_dir,
             executor_config=executor_config,
@@ -630,6 +645,10 @@ class LLM:
             ),
             is_llm_executor=True,
             lora_config=self.args.lora_config)
+
+    @property
+    def _on_trt_backend(self) -> bool:
+        return isinstance(self.args, TrtLlmArgs)
 
     def _try_load_tokenizer(self) -> Optional[TokenizerBase]:
         if self.args.skip_tokenizer_init:
--- a/tensorrt_llm/llmapi/llm.py
+++ b/tensorrt_llm/llmapi/llm.py
@@ -115,7 +115,7 @@ class LLM:
                                                      None)
 
             llm_args_cls = TorchLlmArgs if kwargs.get(
-                'backend', None) == 'torch' else TrtLlmArgs
+                'backend', None) == 'pytorch' else TrtLlmArgs
 
             self.args = llm_args_cls.from_kwargs(
                 model=model,
--- a/tensorrt_llm/llmapi/llm.py
+++ b/tensorrt_llm/llmapi/llm.py
@@ -27,8 +27,9 @@ from ..executor.utils import (create_mpi_comm_session,
 from ..inputs import PromptInputs, create_input_processor, prompt_inputs
 from ..logger import logger
 from ..sampling_params import SamplingParams
-from .llm_args import LLMARGS_EXPLICIT_DOCSTRING, PybindMirror
-from .llm_utils import (CachedModelLoader, KvCacheRetentionConfig, LlmArgs,
+from .llm_args import (LLMARGS_EXPLICIT_DOCSTRING, PybindMirror, TorchLlmArgs,
+                       TrtLlmArgs)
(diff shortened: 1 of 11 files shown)


tensorrt_llm/llmapi/llm_args.py
```diff
@@ -791,12 +788,6 @@ class LlmArgs(BaseModel):
     calib_config: Optional[CalibConfig] = Field(
```
QiJune, line 788: Please move `calib_config` to `TrtLlmArgs` as well

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -822,24 +813,13 @@ class LlmArgs(BaseModel):
...
 
-    workspace: Optional[str] = Field(default=None,
-                                     description="The workspace for the model.")
-
     # A handful of options from PretrainedConfig
     embedding_parallel_mode: str = Field(
         default='SHARDING_ALONG_VOCAB',
         description="The embedding parallel mode.")
 
     fast_build: bool = Field(default=False, description="Enable fast build.")
```
QiJune, line 821: Please move `fast_build` to `TrtLlmArgs`

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -822,24 +813,13 @@ class LlmArgs(BaseModel):
         default=None,
         description="The maximum number of iterations for request stats.")
 
-    workspace: Optional[str] = Field(default=None,
-                                     description="The workspace for the model.")
-
     # A handful of options from PretrainedConfig
     embedding_parallel_mode: str = Field(
```
QiJune, line 817: Please move `embedding_parallel_mode` to `TrtLlmArgs`

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1282,32 +1255,58 @@ def _setup_embedding_parallel_mode(self):
...
+LlmArgs = TrtLlmArgs
 
 LLMARGS_EXPLICIT_DOCSTRING = generate_api_docs_as_docstring(LlmArgs,
                                                             indent=' ' * 4)
 
 
+class TorchLlmArgs(LlmArgsBase):
+
+    def model_post_init(self, __context):
+        super().model_post_init(__context)
```
QiJune, line 1307: It seems that `LlmArgsBase.model_post_init` still has some logic are TRT specific, for  example, `auto_parallel_config`. Could we seperate it as well?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -680,7 +683,10 @@ def model_name(self) -> Union[str, Path]:
         return self.model if isinstance(self.model, str) else None
 
 
-class LlmArgs(BaseModel):
+class LlmArgsBase(BaseModel):
```
QiJune, line 686: How about renaming to `BaseLlmArgs`? The `Base` prefix is more clear to signal the class's role as a Base/Parent class.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Done, thanks.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Done, thanks.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Done, thanks.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Done, thanks.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Good suggestion, done, thanks.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1282,32 +1235,78 @@ def _setup_embedding_parallel_mode(self):
...
+                                description="Enable auto parallel mode.")
+
+    auto_parallel_world_size: Optional[int] = Field(
+        default=None, description="The world size for auto parallel mode.")
+
+    enable_tqdm: bool = Field(default=False,
+                              description="Enable tqdm for progress bar.")
+
+    # BuildConfig is introduced to give users a familiar interface to configure the model building.
+    build_config: Optional[object] = Field(
```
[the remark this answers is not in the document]
Superjomn: The background is that we utilize [generate_api_docs_as_docstring](https://github.com/superjomn/TensorRT-LLM/blob/57944206ba403acf6ac2b3834b33ad15464ee6d5/tensorrt_llm/llmapi/llm_args.py#L1307-L1308) to generate the API doc, and it needs the JSON representation of the fields. The BuildConfig and other pybinding types, such as DecodingConfig don't support JSON type representation, so the WAR is to mark their typing as object and utilize the `json_schema_extra` to specify a real type string.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -998,16 +950,16 @@ def model_post_init(self, __context: Any):
...
-    def from_kwargs(cls, **kwargs: Any) -> "LlmArgs":
+    def from_kwargs(cls, **kwargs: Any) -> "BaseLlmArgs":
         """Create `LlmArgs` instance from kwargs.
 
         Args:
             kwargs (Any): Arguments passed to `LlmArgs` constructor.
 
         Returns:
-            tensorrt_llm.llmapi.llm_utils.LlmArgs: The `LlmArgs` instance.
+            tensorrt_llm.llmapi.llm_utils.LlmArgsBase: The `LlmArgsBase` instance.
```
syuoni, line 960: Are `LlmArgsBase` and `BaseLlmArgs` the some thing? Should we unify them?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1300,32 +1252,78 @@ def _setup_embedding_parallel_mode(self):
...
+    extended_runtime_perf_knob_config: Optional[
+        ExtendedRuntimePerfKnobConfig] = Field(
+            default=None, description="Extended runtime perf knob config.")
+
+    calib_config: Optional[CalibConfig] = Field(
+        default=None, description="Calibration config.")
+
+    embedding_parallel_mode: str = Field(
+        default='SHARDING_ALONG_VOCAB',
+        description="The embedding parallel mode.")
```
syuoni, line 1296: Some args may be supported in PyTorch flow in the future. At that time, just moving the args to `BaseLlmArgs` will work, right?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1300,32 +1252,78 @@ def _setup_embedding_parallel_mode(self):
...
+LlmArgs = TrtLlmArgs
 
 LLMARGS_EXPLICIT_DOCSTRING = generate_api_docs_as_docstring(LlmArgs,
                                                             indent=' ' * 4)
 
 
+class TorchLlmArgs(BaseLlmArgs):
+
+    def model_post_init(self, __context):
+        super().model_post_init(__context)
```
syuoni, line 1324: Should we register `pytorch_backend_config` in `TorchLlmArgs`?

tensorrt_llm/llmapi/llm_args.py
Superjomn: Yes, you are right. The BaseLlmArgs should always hold the knobs in common.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Let me do this in the second PR, which includes eliminating the PyTorchConfig and expand the knobs into TorchLlmArgs.

tensorrt_llm/llmapi/llm.py
```diff
@@ -112,7 +113,11 @@ def __init__(self,
         try:
             self.pytorch_backend_config = kwargs.pop('pytorch_backend_config',
                                                      None)
-            self.args = LlmArgs.from_kwargs(
+
+            llm_args_cls = TorchLlmArgs if kwargs.get(
```
FrankD412, line 117: This is a knit-pick, but couldn't this be put in a dictionary or a factory class?

tensorrt_llm/llmapi/llm_args.py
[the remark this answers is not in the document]
FrankD412: Got it -- I assume this won't be needed once those classes also get converted to `BaseModel`s as well.

Also can't seem to resolve this comment, so hopefully this doesn't hold up the merge.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1300,32 +1252,78 @@ def _setup_embedding_parallel_mode(self):
...
+    embedding_parallel_mode: str = Field(
+        default='SHARDING_ALONG_VOCAB',
+        description="The embedding parallel mode.")
+
+    fast_build: bool = Field(default=False, description="Enable fast build.")
+
+    def model_post_init(self, __context):
+        super().model_post_init(__context)
 
+        self.auto_parallel_config = AutoParallelConfig(
```
FrankD412, line 1303: One thing I noticed when trying to write out the class once the class was initialized was that this config automatically was written out. I'm not sure if this is the right answer, but something like this where you could pair it with a validator might work?

```
from pydantic import BaseModel, PrivateAttr

class MyModel(BaseModel):
    _internal_state: int = PrivateAttr(default=0)
```

The `PrivateAttr` prevents it from being set at initialization time and when you serialize the model it doesn't get stored. It can be modified after initialization though -- which I think we might not want? But it's something worth looking into as we're trying to make sure this can be used to initialize and having the extra class internals written out during serialization means that a config won't load properly.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1027,18 +979,18 @@ def _maybe_update_config_for_consistency(
         # max_beam_width is not included since vague behavior due to lacking the support for dynamic beam width during
         # generation
         black_list = set(["max_beam_width"])
-        executor_config_attrs = set(attr for attr in dir(ExecutorConfig)
-                                    if not attr.startswith('_')
-                                    and callable(getattr(ExecutorConfig, attr)))
+        executor_config_attrs = set(
+            attr for attr in dir(_ExecutorConfig) if not attr.startswith('_')
```
FrankD412, line 983: Might this also be something that could be handled with a `PrivateAttrt` that gets initialized under the hood? Though I suppose the goal here is to make sure that all `_ExecutorConfig` attributes are available at the top-level config?

tensorrt_llm/llmapi/llm_args.py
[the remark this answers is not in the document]
Superjomn: Yeah, the remaining non-pydantic configs are in TRT path, so ideally, the TorchLlmArgs will be pure pydantic.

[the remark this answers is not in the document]
FrankD412: Got it -- are we just leaving the TRT path alone at this point? That means we can't really serialize those configurations.

tensorrt_llm/llmapi/llm.py
Superjomn: I see, it could be fine as there are only two backends, we can change it like you mentioned if we have three or more backends in the future.

tensorrt_llm/llmapi/llm_args.py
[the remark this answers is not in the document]
Superjomn: Yes, but currently, supporting BuildConfig in YAML also need some manual WAR, it seems unnecessary to convert that to Pydantic. We can enhance the TRT path by demand and focus on the PyT path.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Good suggestion, thanks! I've adopted this. It is more intuitive to declare all the members and avoid adding members dynamically.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Sorry, I am not fully following. Could you help elaborate with some sample code?

tensorrt_llm/llmapi/llm.py
FrankD412: Sounds good -- yeah, I just wanted to call it out. :)

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -973,7 +966,7 @@ def model_post_init(self, __context: Any):
         if self.moe_expert_parallel_size is None:
             self.moe_expert_parallel_size = -1
 
-        self.parallel_config = _ParallelConfig(
+        self._parallel_config = _ParallelConfig(
```
FrankD412, line 969: This might be another candidate for `PrivateAttr`?

tensorrt_llm/llmapi/llm_args.py
FrankD412: I was thinking that you might be able to define the fields you're pulling from the executor config as a `PrivateAttr` in here so that you can hide them from being serialized. But that does mean you need to maintain that mapping whenever the `ExecutorConfig` is updated.

Superjomn: After checking the implementation of `_maybe_update_config_for_consistency`, I found it is stateless, and it doesn't change any members of BaseLlmArgs. So it seems unnecessary to declare PrivateAttr for this function.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Yes, I've updated it as a PrivateAttr.

tensorrt_llm/llmapi/llm.py
```diff
@@ -112,7 +113,11 @@ def __init__(self,
...
+                'backend', None) == 'torch' else TrtLlmArgs
```
QiJune, line 118: I think we are using `pytorch` in other places, for example:

- https://github.com/NVIDIA/TensorRT-LLM/blob/791c2090064413ab618f955519d00b2b36f62841/tensorrt_llm/_torch/llm.py#L26
- https://github.com/NVIDIA/TensorRT-LLM/blob/791c2090064413ab618f955519d00b2b36f62841/tests/unittest/llmapi/test_llm.py#L113

QiJune (approved):
LGTM

tensorrt_llm/llmapi/llm.py
Superjomn: Updated, thanks for the reminder.

hchings:
Seems lora test is flaky.

```
FAILED unittest/llmapi/test_llm_multi_gpu.py::test_llama_7b_multi_lora_tp2 - AssertionError: assert False
 +  where False = similar('华盛顿。\n\n美国的第二大城市是', '美国的首都是华盛顿。\n\n美国的')
 ```

Superjomn:
/bot skip --comment "both L0 and multi-gpu tests passed with build #11467 and #11539"

Superjomn:
> Seems lora test is flaky.
> 
> ```
> FAILED unittest/llmapi/test_llm_multi_gpu.py::test_llama_7b_multi_lora_tp2 - AssertionError: assert False
>  +  where False = similar('华盛顿。\n\n美国的第二大城市是', '美国的首都是华盛顿。\n\n美国的')
> ```

Thanks for helping rerun, finally got merged!

PR #3827, opened 2025-04-24
[https://nvbugs/5154414][fix] Balanced layer to PP rank assignment

amukkara:
## Description

For some model and PP size combinations, `num_hidden_layers % pp_size != 0`. This PR creates a balanced assignment of layers to PP ranks in such cases, with few ranks assigned just one extra layer.

For example, Deepseek-V3 with 61 layers, pp size = 8, 61 % 8 = 5:
First 5 ranks get 8 layers each, last 3 ranks get 7 layers each.

Before this change, first 7 ranks get 7 layers each, and last rank get 12 layers causing OOM on last rank.

### DIFF
--- a/tensorrt_llm/mapping.py
+++ b/tensorrt_llm/mapping.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 from typing import List
 
+import torch
+
 
 class Mapping(object):
     '''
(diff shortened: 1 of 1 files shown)


achartier (approved):

chang-l (commented):
Can you use `tensor_split` ([ref](https://pytorch.org/docs/stable/generated/torch.tensor_split.html)) here for simplicity? I think it should do the same thing.

yuxianq (approved):

amukkara:
> Can you use `tensor_split` ([ref](https://pytorch.org/docs/stable/generated/torch.tensor_split.html)) here for simplicity? I think it should do the same thing.

@chang-l done in 1c9388eecbba058020795df4901d5338c56d1389

chang-l (approved):

amukkara:
Change included in https://github.com/NVIDIA/TensorRT-LLM/pull/4399 and https://github.com/NVIDIA/TensorRT-LLM/pull/4034

PR #3828, opened 2025-04-24
refactor: skip autotune and torch compile in second warmup

HuiGao-NV:
In current kv cache memory estimation process, py_executor will be created twice with the same model_engine, and everytime py_executor is created, model_engine.warmup will be called.
Because autotune and torch compile result can be re-used in the same process, then the second warmup can skip them and only perform cudagraph capturing.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -372,6 +372,7 @@ class PyTorchModelEngine(ModelEngine):
         # with different KV cache managers.
         self.kv_cache_manager_key = KV_CACHE_MANAGER_KEY
         self.lora_model_config: Optional[LoraModelConfig] = None
+        self.warmed_up = False
 
     def set_lora_model_config(self, lora_target_modules: list[str],
                               trtllm_modules_to_hf_modules: dict[str, str]):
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -588,6 +589,7 @@ def _create_extra_inputs(bs, num_tokens_per_request):
                              resource_manager=resource_manager,
                              extra_model_inputs=_create_extra_inputs(bs, 1))
                 torch.cuda.synchronize()
+        self.warmed_up = True
```
hyukn, line 592: This flag will not be set correctly when not using cudagraph, right?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -543,7 +544,7 @@ def _create_extra_inputs(bs, num_tokens_per_request):
                                     bs, num_tokens_per_request))
                             torch.cuda.synchronize()
 
-        if self.pytorch_backend_config.autotuner_enabled:
+        if self.pytorch_backend_config.autotuner_enabled and self.warmed_up == False:
```
hyukn, line 547: After another view I realize two facts:
* The second stage can be skipped only if everything has been tuned already during the previous round of warmup.
* If a tuning config has already been in cache, it is supposed not to be tuned anymore because of the cache hit.
I am confused why the second profiling is still ongoing and might trigger more memory allocation. Does it mean that there are some distinctions between the two rounds of warmup?

HuiGao-NV: I am changing to check every step in warmup to skip the steps which have been warmed-up.
The second autotune doesn't need more memory. OOM doesn't happen in the first autotune because there is only very little memory is allocated for kv cache manager and there is enough memory left to do autotuen. However in the second autotune, most memory have been allocated for kv cache memory and no enough memory left for autotune.

HuiGao-NV:
A similar change has been merged. Then, close this PR.

PR #3845, opened 2025-04-24
feat: Mistral-Large-2 support in the Pytorch workflow

jdebache:
- Separate modelling logic for Mistral as its configuration is different than Llama models


jdebache:
@QiJune oh I see there's an issue with Nemotron in my changeset. Let me fix it.

jdebache:
@QiJune pipeline is green. Could you please have another look?

QiJune (approved):
LGTM

PR #3847, opened 2025-04-24
fix:[AutoDeploy] Patch for torch load_state_dict()

sugunav14:
Auto Deploy currently uses transformers' load_sharded_checkpoint() which calls torch.load_state_dict() with assign=False. This setting preserves the model's parameter dtypes rather than using the dtypes from the checkpoint. For Deepseek models, this causes the e_score_correction_bias parameter to be incorrectly cast to bfloat16 instead of maintaining its original dtype from the checkpoint. So we patch the torch.load_state_dict() function in the auto_deploy path to use assign=True, ensuring parameters maintain their original dtypes from the checkpoint file.

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/models/hf.py
+++ b/tensorrt_llm/_torch/auto_deploy/models/hf.py
@@ -2,7 +2,7 @@
 
 import json
 import os
-from contextlib import nullcontext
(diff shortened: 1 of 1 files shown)


lucaslie (approved):
lgtm 👍

PR #3849, opened 2025-04-25
feat: AutoDeploy fp8 quantization support for bmm

meenchen:
# Feat: AutoDeploy fp8 quantization support for bmm

## Description

Adding FP8 quantization transformation support for bmm which is used by llama4.

## Test Coverage

Unit tests for ops and transformation.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


tests/unittest/_torch/auto_deploy/_utils_test/_model_test_utils.py
```diff
@@ -204,6 +204,34 @@ def get_input(self, device, dtype=torch.bfloat16):
...
+    def __init__(self, hidden_dim, batch_size):
+        super().__init__()
+        self.hidden_dim = hidden_dim
+        self.batch_size = batch_size
+        # Create parameter weights for BMM
+        self.weight1 = nn.Parameter(torch.randn(batch_size, hidden_dim, hidden_dim))
+
+    def forward(self, x):
+        # x shape: [batch_size, seq_len, embed_dim]
+        return torch.bmm(x, self.weight1)
```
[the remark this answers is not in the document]
meenchen: yes, I think so.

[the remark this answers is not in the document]
lucaslie: can you generalize it so it can handle both weight and activation regardless of whether the argument is arg0 or arg1? Here is an example of this is done in bmm sharding: https://github.com/NVIDIA/TensorRT-LLM/pull/4163/files#r2085270651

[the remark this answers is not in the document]
meenchen: Added the support for dynamic tensors as weights for bmm @lucaslie

tensorrt_llm/_torch/auto_deploy/custom_ops/quant.py
```diff
@@ -218,4 +218,93 @@ def fp4_linear_fake(
...
 
 
-QUANT_OPS = [torch.ops.quant.fp8_linear, torch.ops.quant.fp4_linear]
+@torch.library.custom_op("quant::fp8_bmm", mutates_args=())
+@torch.compile(dynamic=True)
+def fp8_bmm(
+    input: torch.Tensor,
+    weight: torch.Tensor,
+    input_scale: torch.Tensor,
+    weight_scale: torch.Tensor,
```
lucaslie, line 227: Can we stick to [`torch.bmm`](https://docs.pytorch.org/docs/stable/generated/torch.bmm.html) naming convention for these args:
```suggestion
    input: torch.Tensor,
    mat2: torch.Tensor,
    input_scale: torch.Tensor,
    mat2_scale: torch.Tensor,
```

tensorrt_llm/_torch/auto_deploy/custom_ops/quant.py
```diff
@@ -218,4 +218,93 @@ def fp4_linear_fake(
...
+            input_fp8[b],
+            weight_fp8[b],
+            scale_a=input_scale,
+            scale_b=weight_scale,
+            bias=None,
+            scale_result=None,
+            out_dtype=input.dtype,
+            use_fast_accum=True,
+            out=batch_output,
+        )
```
lucaslie, line 284: is this the fastest fp8 bmm we have available?? 

How about we look into flashinfer's fp8 bmm: https://docs.flashinfer.ai/generated/flashinfer.gemm.bmm_fp8.html

tensorrt_llm/_torch/auto_deploy/custom_ops/quant.py
```diff
@@ -218,4 +218,93 @@ def fp4_linear_fake(
...
+@fp8_bmm.register_fake
+def fp8_bmm_fake(
+    input: torch.Tensor,
+    weight_fp8: torch.Tensor,
+    input_scale: torch.Tensor,
+    weight_scale: torch.Tensor,
+) -> torch.Tensor:
+    """Fake implementation of fp8_bmm for testing and tracing."""
+    # Use standard bmm
+    return torch.bmm(input, weight_fp8.to(input.dtype))
```
lucaslie, line 298:
```suggestion
    return torch.bmm(input.to(torch.float), mat2.to(torch.float)).to(input.dtype)
```

It seems your fake op would fail if for example is fp8 already. This is, however, supported by the default implementation above

tensorrt_llm/_torch/auto_deploy/custom_ops/quant.py
```diff
@@ -218,4 +218,93 @@ def fp4_linear_fake(
...
+    if weight.dtype in [torch.float16, torch.bfloat16]:
+        weight_fp8 = _to_fp8(weight, weight_scale)
+    else:
+        assert weight.dtype == torch.float8_e4m3fn
+        weight_fp8 = weight
+
+    # Ensure weight is in the correct memory layout
+    # For cuBLASLt, we need one matrix in row-major and one in column-major format
+    # Since PyTorch uses row-major by default, we'll transpose the weight to get a different layout
+    weight_fp8 = weight_fp8.transpose(-2, -1).contiguous().transpose(-2, -1)
```
lucaslie, line 259: we probably wanna consider moving this into the transformation where we can check:

* if `mat2` is a parameter (`get_attr` node) then we can do this as a preprocessing step in a load hook
* if it's an activation tensor then we can insert it into the graph as separate nodes.

Regardless of this optimization I mentioned above, in here we should explicitly check if `mat2` is already column-major and if not do this step optionally

tensorrt_llm/_torch/auto_deploy/transformations/library/quantization.py
```diff
@@ -81,8 +82,90 @@ def _insert_quantized_linear(
...
+    with gm.graph.inserting_before(node):
+        scales = {}
+        for scale_name in quantization_impl.scale_names():
+            scale_buffer_name = get_scale_name(scale_name)
+            scales[scale_name] = gm.graph.create_node(
+                "get_attr", f"{scale_name_prefix}{scale_buffer_name}"
+            )
+
+    # Update node arguments and kwargs
+    node.kwargs = {**node.kwargs, **scales}
```
lucaslie, line 164: It's more common to use args in torch.exported fx graph. Let's stick to this convention. Moreover, [`torch.bmm`](https://docs.pytorch.org/docs/stable/generated/torch.bmm.html) enforces positional args for input and mat2

tensorrt_llm/_torch/auto_deploy/transformations/library/quantization.py
```diff
@@ -81,8 +82,90 @@ def _insert_quantized_linear(
...
 
 
+def _insert_quantized_bmm(
+    gm: GraphModule,
+    node: Node,
+    quantization_impl: QuantizationImpl,
+    is_quantized_graph: bool = False,
+):
+    """Replaces the bmm node with a new quantized bmm node."""
+    weight_node = node.args[1]
```
lucaslie, line 92: I am still confused about this. Why don't we just generalize the implementation such that either `input` or `mat2` can both either be an activation or a weight tensor. You can take a look at the bmm sharding implementation: https://github.com/NVIDIA/TensorRT-LLM/blob/e2eea80c1dbf4043f40a4a2797234db27b018213/tensorrt_llm/_torch/auto_deploy/transformations/library/sharding.py#L447-L449

Both sides are handled the same way

lucaslie: This reminds that we need to implement bmm sharding. We can do this in a follow-up PR though if you prefer

Fridah-nv (approved):

tensorrt_llm/_torch/auto_deploy/transformations/library/quantization.py
lucaslie: >Can you share why do we need such flexibility for bmm input/mat2? Does AutoDeploy provide such flexibility for Linear as well?
I think bmm and linear are slightly different ops. Linear supports batching in one input argument and not the other. So it's clear what is the "weight" and what is the activation tensor. In bmm arguments are more interchangeable.

It's also not hard to implement (see bmm sharding) and hence can ensure the future-proofness of the implementation. 

> For FP8, I recall the bmm sharding is working ootb, but I can give it another try to verify that.
That's great. I am assuming it just works ootb because we just shard in the batch dimension which is not affected by the quantized op? Please double-check though

tensorrt_llm/_torch/auto_deploy/custom_ops/quant.py
meenchen: Updated to use fp8 bmm. That's much simpler.

tensorrt_llm/_torch/auto_deploy/custom_ops/quant.py
meenchen: Updated, now we convert weights to column-major after weight loading.

tensorrt_llm/_torch/auto_deploy/transformations/library/quantization.py
meenchen: For weight, it is handled in the current change.

lucaslie (approved):

PR #3049, opened 2025-03-25
fix: AllReduce CUDA Graph Fix + Kernel Clean up

yizhang-nv:
This PR contains following changes: 
1. Remove all allreduce kernels from `customAllreduceKernels.cu` except pre_post_norm fusion kernel. 
2. Unify the workspace of old and new fusion kernel
3. Fix a bug that allreduce kernel may produce wrong result with cuda graph enabled. 

To merge this mr, we need to wait for the new allreduce kernels in #3064, and then refactor the current call site of allreduce.

PR #3851, opened 2025-04-25
feat: Low Precision Allreduce for PCIe based GPU

kanghui0204:
# Low Precision Allreduce for PCIe based GPU

[feat] Support a new feature for low precision allreduce  for PCIe based GPU

## Description

This PR adds a customized allreduce to TensorRT-LLM. The new allreduce is used for communication on PCIe-based GPUs via low-precision quantization, which can accelerate the PCIe allreduce process.

## Test Coverage
I added a unit test for this feature: `tests/unittest/_torch/multi_gpu/test_lowprecision_allreduce.py`. I hope this test can run on L40 or L20 nodes; otherwise, the low-precision allreduce will fall back to NCCL, making the test meaningless and causing it to fail.


cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -676,7 +842,27 @@ class AllreduceOp
...
+            return AllReduceStrategyType::LOWPRECISION;
+        }
+#else
+        if (forceLowPrecision)
+        {
+            TLLM_LOG_WARNING(
+                "Can't select Low Precision Strategy when compile without ENABLE_FP8 , will fallback to Auto Strategy");
+        }
+#endif
+        // Use Auto select
```
hyukn, line 865: What will happen when `mStrategy == AllReduceStrategyType::LOWPRECISION` but fail to go into `return AllReduceStrategyType::LOWPRECISION;`? 
mStrategy will keep the value AllReduceStrategyType::LOWPRECISION. This might still fall into line 937.

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -310,6 +313,164 @@ class AllreduceOp
         return {};
     }
 
+    std::vector<torch::Tensor> runLowPrecisionAllReduce(torch::Tensor const& input,
+        torch::optional<torch::Tensor> const& residual, torch::optional<torch::Tensor> const& norm_weight,
+        torch::optional<torch::Tensor> const& scale, torch::optional<torch::Tensor> const& bias,
+        torch::optional<torch::Tensor> workspace, AllReduceStrategyType strategy) noexcept
+    {
+#ifdef ENABLE_FP8
+        auto splitNumber = [](size_t number) -> std::vector<size_t>
```
hyukn, line 322: Maybe putting this aux function separated from the dispatch function will be clearer. This makes the function body longer.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -139,7 +166,13 @@ def __init__(self,
         if self.mapping.tp_size > 1:
             # When Strategy is UB, it is guaranteed that the workspace is not used.
             if self.strategy != AllReduceStrategy.UB:
-                self.workspace = get_allreduce_workspace(self.mapping)
+                force_low_precision_env = os.environ.get(
```
hyukn, line 169: Put getting env variable and  `self.strategy == AllReduceStrategy.LOWPRECISION or force_low_precision_env is not None:` before `if self.mapping.tp_size > 1" will be clearer.

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -676,7 +842,27 @@ class AllreduceOp
...
-        // Check that heuristic is only applied when AUTO is set.
+        // Check that heuristic is only applied when AUTO/LOWPRECISION is set.
+        static char* forceLowPrecisionAllReduceStrategyChar = std::getenv("FORCE_LOW_PRECISION_ALL_REDUCE_STRATEGY");
+        bool forceLowPrecision
+            = (forceLowPrecisionAllReduceStrategyChar != nullptr) || (mStrategy == AllReduceStrategyType::LOWPRECISION);
+
+#ifdef ENABLE_FP8
+        // Use LowPrecision if PCIe and p2p support and mesaaage size is large than 4MB
+        // constexpr int LowPrecisionMinMessageSize = 2 * 1024 * 1024;
+        constexpr int LowPrecisionMinMessageSize = 1024;
```
hyukn, line 853: Why 1024 here?

kanghui0204: I made a mistake ，1024 is my debug size , real Min Message Size is 2MB

tensorrt_llm/_torch/distributed/ops.py
kanghui0204: Sure solved

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: Yes, LOWPRECISION is subject to precision loss and cannot accelerate allreduce in all cases. Therefore, when the conditions for using LOWPRECISION are not met, I want to fall back to the auto mode to select automatically. I see that after the selectImplementation function, runtimeStrategy will be used, so mStrategy will no longer be effective, correct?

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: I looked in a few places, but I'm not quite sure where the best place to put this function is. Which dispatch function are you referring to?

hyukn: I refer to runLowPrecisionAllReduce as your dispatch function. Because this function only takes 1 size_t argument, it is more like a standalone utility function. Can we put it in cpp/tensorrt_llm/thop/thUtils.cpp? Or you can directly put it in the very first in allReduceOp.cpp. This makes your major invoking logics more concrete.

cpp/tensorrt_llm/thop/allreduceOp.cpp
hyukn: I mean for this block:
  if (!isAuto)
  {
      if (mStrategy == AllReduceStrategyType::ONESHOT || mStrategy == AllReduceStrategyType::TWOSHOT)
      {
          strat = AllReduceStrategyType::MIN_LATENCY;
      }
      else
      {
          strat = mStrategy;
      }
  }
  
  `strat` will be set as LOWPRECISION if not treated properly.

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: I put splitNumber to cpp/tensorrt_llm/thop/thUtils

cpp/tensorrt_llm/thop/allreduceOp.cpp
hyukn: Just noticed that the workspace is not shared. If LOWPRECISION is not selected, I guess the same workspace is not usable by MIN_LATENCY or NCCL?

juney-nvidia:
@dongxuy04  @yuxianq  Hi Dongxu, Yuxian, can you help review this MR?

Thanks
June

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: I found I use `allocate_workspace` , and you use `allocate_allreduce_fusion_workspace` , some of the buffer is same between these 2 workspace , maybe we can discuss in next monday , to unify 2 workspace , or we now can sent 2 workspace together into torch c++ ops

cpp/tensorrt_llm/kernels/customLowPrecisionAllReduceKernels.h
```diff
@@ -0,0 +1,103 @@
...
+#include "tensorrt_llm/common/cudaUtils.h"
+#include <NvInferRuntime.h>
+#include <cuda_bf16.h>
+#include <cuda_fp16.h>
+
+namespace tensorrt_llm::kernels
+{
+
+constexpr int LP_ALLREDUCE_MAX_BLOCKS = 8;
+constexpr int LP_ALLREDUCE_WARPSIZE = 32;
```
dongxuy04, line 30: Does `LP_ALLREDUCE_WARPSIZE` aim to support other values than `WARP_SIZE` since seems there are some code like `warp_reduce_max` and `__sync_warp` only for `WARP_SIZE`? If not, maybe template arguments for some functions named `WARP_SIZE` can be removed?

tensorrt_llm/plugin/plugin.py
```diff
@@ -713,6 +713,44 @@ def max_workspace_size_auto(tp_size: int) -> int:
...
+        LP_ALLREDUCE_MAX_BLOCKS = 8
+        LP_ALLREDUCE_WARPSIZE = 32
+        LP_ALLREDUCE_DEFAULT_BLOCK_SIZE = 512
+        LP_ALLREDUCE_WARP_NUM_PER_BLOCK = 16
+        LP_ALLREDUCE_BYTES_PER_LOAD = 16
+        LP_ALLREDUCE_MAX_RANKS_PER_NUMA = 4
+        LP_ALLREDUCE_BUFFER_DUPLICATE = 16
+        LP_ALLREDUCE_BUFFER_CHUNKS = 8
+        LP_ALLREDUCE_HIER_STAGE_NUM = 3
+        LP_ALLREDUCE_MAX_ELTS_IN_WORKSPACE = 64 * 1024 * 1024
```
dongxuy04, line 728: I think It might be great if we can reuse the macros defined in C++ header files or implement this function in C++ and bind to python instead of redefining them in Python again.

cpp/tensorrt_llm/thop/thUtils.cpp
```diff
@@ -110,4 +110,48 @@ cudaDataType_t convert_torch_dtype(torch::ScalarType dtype)
     }
 }
 
+std::vector<size_t> splitNumber(size_t number)
+{
+    int const targetSize = 64 * 1024 * 1024;      // 64 MB
```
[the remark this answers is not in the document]
kanghui0204: I moved the `splitNumber` function to `customLowPrecisionAllReduceKernels.h/cu`, so now we can use the `LP_ALLREDUCE_MAX_ELTS_IN_WORKSPACE` constant int.

cpp/tensorrt_llm/thop/thUtils.cpp
```diff
@@ -110,4 +110,48 @@ cudaDataType_t convert_torch_dtype(torch::ScalarType dtype)
     }
 }
 
+std::vector<size_t> splitNumber(size_t number)
```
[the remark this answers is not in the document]
kanghui0204: When all my data regarding the maximum buffer is calculated using my macros, such as `LP_ALLREDUCE_MAX_ELTS_IN_WORKSPACE`, does that guarantee it?

cpp/tensorrt_llm/kernels/customLowPrecisionAllReduceKernels.cu
```diff
@@ -0,0 +1,1507 @@
...
+    return val;
+}
+
+template <typename QUANTIZE_T>
+struct QuantMaxValue;
+
+template <>
+struct QuantMaxValue<int8_t>
+{
+    static constexpr int8_t value = 127.0f;
```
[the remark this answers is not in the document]
kanghui0204: Sorry,my mistake, it has been corrected.

cpp/tensorrt_llm/kernels/customLowPrecisionAllReduceKernels.cu
```diff
@@ -0,0 +1,1507 @@
...
+    PackedInputType vals[output_rounds];
+
+    for (size_t start = blockIdx.x * LP_ALLREDUCE_WARP_NUM_PER_BLOCK + wid; start * NUM_ELTS_PER_WARP_IN < n_in;
+         start += LP_ALLREDUCE_WARP_NUM_PER_BLOCK * gridDim.x)
+    {
+        int32_t read_rounds = 0;
+        int32_t local_n_in = (n_in - start * NUM_ELTS_PER_WARP_IN) > NUM_ELTS_PER_WARP_IN
+            ? NUM_ELTS_PER_WARP_IN
+            : (n_in - start * NUM_ELTS_PER_WARP_IN);
+        if (local_n_in < 0)
```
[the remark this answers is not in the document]
kanghui0204: Sorry,my mistake, it has been corrected.

cpp/tensorrt_llm/kernels/customLowPrecisionAllReduceKernels.h
kanghui0204: My initial idea was to avoid using macros from other code, so I defined LP_ALLREDUCE_WARPSIZE. However, I found that the name LP_ALLREDUCE_WARPSIZE was too long and cumbersome to use in the kernel. Therefore, I used a template parameter with a default value. But now, I have followed your suggestion and removed these. In the kernel, I am now using the WARP_SIZE macro.

tensorrt_llm/plugin/plugin.py
kanghui0204: Currently, I have added a `max_workspace_size_lowprecision` C++ function in `customLowPrecisionAllReduceKernels.h/cu`, but I am unsure which namespace to use for mapping this function (since it is not a torch op). I will discuss with @hyukn  how to map this C++ function to Python and then delete the Python version of `max_workspace_size_lowprecision`.

kanghui0204: already delete python max_workspace_size_lowprecision

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: After discussing with @hyukn , we believe that it is necessary to unify the new workspace with my workspace. Here is the reference [PR](https://github.com/NVIDIA/TensorRT-LLM/pull/3049/files#diff-21f5eaec51abee38157e27f8c06e40a4a437eafca864bb2f10c65d90e25a63df). @kanghui0204  will read the PR and provide changes later.

cpp/tensorrt_llm/kernels/customLowPrecisionAllReduceKernels.cu
```diff
@@ -0,0 +1,1580 @@
...
+    size_t elts_per_thread = LP_ALLREDUCE_BYTES_PER_LOAD; // assume quant_type size is 1 bytes
+    int msg_align = elts_per_thread;
+    if (n_ranks <= 4)
+    {
+        msg_align *= n_ranks;
+    }
+    return msg_size % msg_align == 0;
+}
+
+int32_t max_workspace_size_lowprecision(int32_t tp_size)
```
hlu1, line 1528:
```suggestion
int32_t maxWorkspaceSizeLowPrecision(int32_t tp_size)
```
Be consistent with naming convention.

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -310,6 +313,120 @@ class AllreduceOp
         return {};
     }
 
+    std::vector<torch::Tensor> runLowPrecisionAllReduce(torch::Tensor const& input,
+        torch::optional<torch::Tensor> const& residual, torch::optional<torch::Tensor> const& norm_weight,
+        torch::optional<torch::Tensor> const& scale, torch::optional<torch::Tensor> const& bias,
+        torch::optional<torch::Tensor> workspace, AllReduceStrategyType strategy) noexcept
```
hlu1, line 319:
```suggestion
        torch::optional<torch::Tensor>& workspace, AllReduceStrategyType strategy) noexcept
```

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -310,6 +313,120 @@ class AllreduceOp
...
+        for (auto const& currentRank : mGroup)
+        {
+            if (cur_rank == currentRank)
+                break;
+            ++tp_rank;
+        }
+
+        int bytes_per_element = input.element_size();
+
+        int token_num = size / input.size(-1);
```
hlu1, line 339:
```suggestion
        int token_num = size / hidden_size;
```
Avoid calling Tensor.size() over and over again. It's not free.

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -310,6 +313,120 @@ class AllreduceOp
...
+
+        auto workspace_ptr = workspace.value().mutable_data_ptr();
+        auto parts = tensorrt_llm::kernels::splitNumber(size);
+
+        if (mOp == AllReduceFusionOp::RESIDUAL_RMS_NORM)
```
hlu1, line 344: What's the point of adding a non-fused path here? It calls the allreduce and norm separately.

docs/source/advanced/lowprecision-pcie-allreduce.md
```diff
@@ -0,0 +1,59 @@
...
+This feature is optimized for PCIe-based GPU topologies and may affect model accuracy. Please evaluate precision impact for your specific workload.
+```
+
+
+TRT-LLM supports `low-precision-allreduce`, a communication optimization that accelerates AllReduce operations in PCIe-based GPU environments. This feature quantizes FP16/BF16 data to FP8 during network transmission, reducing communication volume and improving performance.
+
+## Algorithm
+
+The Low-Precision-AllReduce algorithm works by:
+1. Quantizing input FP16/BF16 tensors to FP8 format before network transmission
```
hlu1, line 14: Can you add more details on what kind of FP8 quantization it uses? Is it per-tensor/per-row/per-channel/per-block?

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: Thank you, I have made the modifications according to your comment.

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: Thank you, I have made the modifications according to your comment.

docs/source/advanced/lowprecision-pcie-allreduce.md
kanghui0204: Quantization details: We use a "per-warp" quantization approach where each CUDA warp (32 threads) processes a batch of data. In each warp, 31 threads quantize FP16/BF16 values to FP8 format (16 bytes per thread), while the last thread transmits a scalar value. This results in each warp collectively quantizing 496 elements plus one scalar at a time.

I have added the above description about quantization. Could you check if it is clear?

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: The impact scenarios of low precision allreduce are similar to those of NCCL, both being used for bandwidth-bound allreduce operations. Therefore, I implement `runLowPrecisionAllReduce` as same as `runNCCLAllReduce`, support fuse op on `RESIDUAL_RMS_NORM` and `None`.

Detail of  `runNCCLAllReduce` is here : https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/thop/allreduceOp.cpp#L272

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: @kanghui0204  moved the LowPrecision workspace to a static buffer, which means it only needs to be initialized during allreduce initialization, allowing lowPrecision to use its own workspace without explicitly passing the workspace during forward calls. The benefit of this approach is that LowPrecision no longer affects other workflows. This has been discussed with @hyukn  and @yizhang-nv , and waiting for their review results.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -244,10 +278,11 @@ def forward(
...
         fusion_op = all_reduce_params.fusion_op
 
+        try_low_precision = self.strategy == AllReduceStrategy.LOWPRECISION or self.force_low_precision_env is not None
         # If the input size is larger than the max workspace size, fallback to NCCL strategy
         if input.numel() > self.max_workspace_size \
             and all_reduce_params.fusion_op != AllReduceFusionOp.NONE \
-            and all_reduce_params.fusion_op != AllReduceFusionOp.RESIDUAL_RMS_NORM:
+            and all_reduce_params.fusion_op != AllReduceFusionOp.RESIDUAL_RMS_NORM and (not try_low_precision):
             strategy = AllReduceStrategy.NCCL
             fusion_op = AllReduceFusionOp.RESIDUAL_RMS_NORM
```
hlu1, line 287: These will be removed soon by https://github.com/NVIDIA/TensorRT-LLM/pull/4080

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -679,7 +796,31 @@ class AllreduceOp
...
+        else
+        {
+            mStrategy = AllReduceStrategyType::AUTO;
+        }
+#else
+        if (forceLowPrecision)
+        {
+            TLLM_LOG_WARNING(
+                "Can't select Low Precision Strategy when compile without ENABLE_FP8 , will fallback to Auto Strategy");
+            mStrategy = AllReduceStrategyType::AUTO;
```
hlu1, line 820: Is this line redundant?

docs/source/advanced/lowprecision-pcie-allreduce.md
hlu1: What if the number of the elements is not multiples of 496?

cpp/tensorrt_llm/thop/allreduceOp.cpp
hlu1: @hyukn, what's your opinion on this?

kanghui0204: I discussed with @hyukn , and he has a PR to enable NCCL to support all fused ops. The behavior of low precision is similar to NCCL. We plan to wait until his PR is merged, and then I will modify the logic for low precision + fused ops to be similar to NCCL.

docs/source/advanced/lowprecision-pcie-allreduce.md
kanghui0204: If it is not a multiple of 496, the remaining elements will be quantized within a single warp, and a scalar will be computed among these elements.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -27,6 +28,23 @@ def get_allreduce_workspace(mapping: Mapping) -> torch.LongTensor:
     return allreduce_workspaces[mapping][1]
 
 
+def allocate_low_presicion_allreduce_workspace(mapping: Mapping) -> None:
+    if not hasattr(_thread_local, 'lowprecision_allreduce_workspaces'):
+        #raw custom allreduce can use low_presicion_allreduce's workspace , because low_presicion_allreduce have large workspace
```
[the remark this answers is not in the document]
kanghui0204: yes , you are right , I will remove this comment ,  for now low precision will not share the workspace with other allreduce operations either.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -156,9 +186,13 @@ def __init__(self,
             self.fallback_residual_rms_norm_out_quant_nvfp4,
         }
 
+        self.force_low_precision_env = os.environ.get(
+            "FORCE_LOW_PRECISION_ALL_REDUCE_STRATEGY")
         if self.mapping.tp_size > 1:
             # When Strategy is UB, it is guaranteed that the workspace is not used.
             if self.strategy != AllReduceStrategy.UB:
+                if self.strategy == AllReduceStrategy.LOWPRECISION or self.force_low_precision_env is not None:
+                    allocate_low_presicion_allreduce_workspace(self.mapping)
```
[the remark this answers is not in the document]
kanghui0204: When you call allocate_low_precision_allreduce_workspace, the address in the workspace will be registered to the low precision allreduce static variable. This way, during the forward call, low precision allreduce does not need the workspace input; it can simply find these workspace addresses from its own static variable. The purpose of this approach is to avoid interfering with other allreduce workflows.

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: My idea is this: if a user enables low precision but is using an old GPU or complie with out FP8, then low precision will not take effect. Therefore, the warning here is to alert the user that low precision is not effective. How do you think？

tensorrt_llm/plugin/plugin.py
```diff
@@ -763,6 +775,56 @@ def allocate_workspace(mapping: Mapping,
...
             device="cpu")
 
+    @staticmethod
+    def allocate_lowprecision_workspace(
+            mapping: Mapping,
+            size: int) -> Tuple[List[IpcMemory], "torch.tensor"]:
+        import torch
+
+        # Force pull mode and disable lamport when force deterministic is enabled, for reducing device memory usage.
+        force_deterministic = force_all_reduce_deterministic()
```
yizhang-nv, line 785: Is these workspace ptrs for fallback? If so, please refer to `allocate_allreduce_fusion_workspace` since `allocate_workspace` is for trt flow only. And currently we haven't add `force_all_reduce_deterministic()` support for it.

cpp/tensorrt_llm/thop/allreduceOp.cpp
hyukn: I have already revised the code for NCCL + (fallback sequential ops) in #4080. You can take a further look at how to enable LOW_LATENCY + fallback paths. @kanghui0204

cpp/tensorrt_llm/thop/allreduceOp.cpp
hlu1: It's probably overkill.

cpp/tensorrt_llm/kernels/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ endfunction()
 
 if(NOT ENABLE_MULTI_DEVICE)
   list(FILTER SRC_CU EXCLUDE REGEX "customAllReduceKernels*.*cu$")
+  list(FILTER SRC_CU EXCLUDE REGEX "customLowPrecisionAllReduceKernels*.*cu$")
```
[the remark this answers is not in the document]
hlu1: That would be a better location.

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: @hyukn , I already read your code , I think I can modify after your PR merged , and then I will modify as How NCCL allreduce run.

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: I remove this warning , and add a warning on here https://github.com/kanghui0204/TensorRT-LLM/blob/low_precision_allreduce_for_pcie/docs/source/advanced/lowprecision-pcie-allreduce.md#environment-variables

cpp/tensorrt_llm/kernels/CMakeLists.txt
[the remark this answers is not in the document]
kanghui0204: thanks to remind , I move the file to `communicationKernels` folder

tensorrt_llm/plugin/plugin.py
kanghui0204: for now I don't need fallback ,I remove all the useless buffer for low precision

cpp/tensorrt_llm/kernels/customLowPrecisionAllReduceKernels.cu
```diff
@@ -0,0 +1,1637 @@
...
+    {
+        TLLM_THROW("Unsupported tpSize for LowPrecisionCustomAllReduce");
+    }
+}
+
+void initialize_static_lowprecision_buffers(int64_t* buffer, size_t tpSize)
+{
+    void* const* buffer_ptrs = reinterpret_cast<void* const*>(buffer);
+
+    StaticLowPrecisionBuffers* static_buffers = getBufferForTpSize(tpSize);
```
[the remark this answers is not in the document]
yizhang-nv: I think the current solution is fine. No need to add an extra map to record static buffer

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: modify as PR https://github.com/NVIDIA/TensorRT-LLM/pull/4080 ， could you check @hlu1 @hyukn

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -114,38 +131,50 @@ def __init__(self,
...
+                  on specific hardware configurations.
+
+            All strategies support the following operations:
+                - NONE (AllReduce only)
+                - RESIDUAL_RMS_NORM
+                - RESIDUAL_RMS_NORM_QUANT_FP8
+                - RESIDUAL_RMS_NORM_QUANT_NVFP4
+                - RESIDUAL_RMS_NORM_OUT_QUANT_FP8
+                - RESIDUAL_RMS_NORM_OUT_QUANT_NVFP4
+
```
hyukn, line 156: Can you notify that NCCL and UB will only support consequent kernel calls, instead of the fused version?

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -766,7 +841,27 @@ class AllreduceOp
     AllReduceStrategyType selectImplementation(
         size_t seq_len, size_t message_size, int world_size, nvinfer1::DataType type) noexcept
     {
-        // Check that heuristic is only applied when AUTO is set.
+        // Check that heuristic is only applied when AUTO/LOWPRECISION is set.
```
hyukn, line 844: I suggest to wrap this into a standalone method to determine if LOWPRECISION is used
```cpp
if (isUsingLowPrecision()){
}
else{
...
{
```

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -966,10 +1061,22 @@ TORCH_LIBRARY_FRAGMENT(trtllm, m)
...
 
 TORCH_LIBRARY_IMPL(trtllm, CUDA, m)
 {
     m.impl("allreduce", &torch_ext::allreduce);
     m.impl("moe_allreduce", &torch_ext::moe_allreduce);
 }
+
+TORCH_LIBRARY_IMPL(trtllm, CPU, m)
+{
+    m.impl("initialize_static_lowprecision_buffers",
```
hyukn, line 1075: This is better. We only need the impl instead of op registration.

hyukn (approved):
Just some nits. LGTM.

cpp/tensorrt_llm/thop/allreduceOp.cpp
kanghui0204: add isUsingLowPrecision function ,could you check?

tensorrt_llm/_torch/distributed/ops.py
kanghui0204: add a Note , could you check?

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -113,38 +130,52 @@ def __init__(self,
...
+            Note: NCCL, UB, and LOWPRECISION strategies only support consequent kernel calls
+        instead of fused operations.
```
hlu1, line 157:
```suggestion
            Note: NCCL, UB, and LOWPRECISION strategies support only consecutive kernel calls instead of fused operations.  
```

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -113,38 +130,52 @@ def __init__(self,
...
+                - NCCL: Use NCCL allreduce.
+
+                - MIN_LATENCY: AllReduce uses MIN_LATENCY mode kernel.
 
                 - AUTO: AUTO chooses between NCCL and MIN_LATENCY mode based on a heuristic policy.
 
+                - LOWPRECISION: AllReduce quantizes data to lower precision for transmission.
+                  Should only be used on topologies with PCIe switches and without NVLink.
+                  This strategy may result in some precision loss but can improve performance
+                  on specific hardware configurations.
```
hlu1, line 146:
```suggestion
                - LOWPRECISION:  This strategy quantizes data to lower precision before transmission during AllReduce. It should be used only on topologies with PCIe switches and without NVLink. While it may introduce some precision loss, it can improve performance on certain hardware configurations.
```

tensorrt_llm/_torch/distributed/ops.py
kanghui0204: solved

tensorrt_llm/_torch/distributed/ops.py
kanghui0204: solved

QiJune:
Hi @kanghui0204 @hyukn , we found that multi-gpu test pipelines are not triggered as expected in this PR due to `/bot reuse-pipeline`. At the same time, the post merge multi-gpu test pipelines hangs. So, we revert this PR to fix broken CI first. Feel free to submit a new PR, and trigger multi-gpu test pipelines by `/bot run --only-multi-gpu-test`

PR #3856, opened 2025-04-25
feat: add health_generate route to openai serving

dsingal0:
Adds /health_generate for kubernetes to detect when the runtime is hung so it can restart the container/pod even if the server /health returns a 200.
This is helpful in cases when a long context request from a user has caused the llm runtime to hang or die but the fastapi server is still running.

### DIFF
--- a/examples/auto_deploy/.vscode/launch.json
+++ b/examples/auto_deploy/.vscode/launch.json
@@ -8,7 +8,7 @@
             "program": "build_and_run_ad.py",
             "args": [
                 "--config",
-                "{\"batch_size\": 2, \"page_size\": 16, \"world_size\": 2, \"compile_backend\": \"torch-simple\", \"attn_backend\": \"FlashInfer\", \"model\": \"meta-llama/Meta-Llama-3.1-8B-Instruct\", \"benchmark\": false}",
+                "{\"batch_size\": 2, \"page_size\": 16, \"world_size\": 2, \"compile_backend\": \"torch-simple\", \"attn_backend\": \"FlashInfer\",\"model_factory\": \"AutoModelForCausalLM\", \"model\": \"meta-llama/Meta-Llama-3.1-8B-Instruct\", \"benchmark\": false}",
                 "--model-kwargs",
                 "{}",
                 // "{\"num_hidden_layers\": 3}",
--- a/examples/auto_deploy/.vscode/settings.json
+++ b/examples/auto_deploy/.vscode/settings.json
@@ -4,7 +4,7 @@
     // Please also modify the .env file accordingly
     "terminal.integrated.env.linux": {
         "OMPI_MCA_opal_cuda_support": "true",
-        "LD_LIBRARY_PATH": "/lib/x86_64-linux-gnu:${env:HOME}/miniconda3/envs/tekit/lib:${env:LD_LIBRARY_PATH}",
+        "LD_LIBRARY_PATH": "/lib/x86_64-linux-gnu:<PATH-TO-CONDA-ENV>/lib:${env:LD_LIBRARY_PATH}",
     },
     // STANDARD SETTINGS, DO NOT MODIFY ////////////////////////////////////////////////////////////
     "python.envFile": "${workspaceFolder}/.vscode/.env",
@@ -23,7 +23,7 @@
     // https://code.visualstudio.com/updates/v1_81#_python
     // When that happens, we can remove the --ignore-glob flag and the gpu tests will be skipped
     "python.testing.pytestArgs": [
-        "./tests",
+        "./tests/unittest/_torch/auto_deploy",
         "--no-cov",
     ],
     "files.exclude": {
--- a/examples/auto_deploy/README.md
+++ b/examples/auto_deploy/README.md
@@ -67,21 +67,21 @@ The exported graph then undergoes a series of automated transformations, includi
 
 **Bring Your Own Model**: AutoDeploy leverages `torch.export` and dynamic graph pattern matching, enabling seamless integration for a wide variety of models without relying on hard-coded architectures.
 
-Additionally, we have officially verified and fully optimized support for the following models:
+Additionally, we have officially verified support for the following models:
 
 <details>
 <summary>Click to expand supported models list</summary>
 
-| Model Series | HF Model Card | Precision | World Size | Runtime | Compile Backend ||| Attention Backend |||
-|--------------|----------------------|-----------|------------|---------|-----------------|--------------------|--------------------|--------------------|----------|----------|
-|              |               |           |            |         | torch-simple    | torch-compile    | torch-opt          | TritonWithFlattenedInputs | FlashInfer | MultiHeadLatentAttention |
-| LLaMA        | meta-llama/Llama-2-7b-chat-hf<br>meta-llama/Meta-Llama-3.1-8B-Instruct<br>meta-llama/Llama-3.1-70B-Instruct<br>codellama/CodeLlama-13b-Instruct-hf | BF16 | 1,2,4 | demollm, trtllm | ✅ | ✅ | ✅ | ✅ | ✅ | n/a |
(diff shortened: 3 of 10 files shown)


dsingal0:
@kaiyux wdyt?

kaiyux:
> @kaiyux wdyt?

@dsingal0 Thanks for the suggestion. Some members in the team are on public holidays and will return next week, we will keep you posted.

At first glance I think we should be careful when introducing a new API that is not within OpenAI API scope, and I'll take a closer look on that.

LinPoly (commented):
For the added entrypoint function, it is acceptable to check if the runtime/executor/engine is alive. But similar logic can also be added to the client code instead of the server side, set a timeout for a simple request, then check if there is any response returned by the server and the response status. Do we have any reason to implement the logic in the server side? @dsingal0

LinPoly:
FYI: vLLM [monitors executor health with daemon thread](https://github.com/vllm-project/vllm/blob/ba7703e659e21ea376c25b872c4a80a7c82804fe/vllm/executor/mp_distributed_executor.py#L192), it is more structured and reliable for me, but we need to implement similar checking from executor level. @kaiyux

dsingal0:
@LinPoly 
>  re implementing the check server side vs client side
for deployment in kubernetes we need a health and liveness probe to detect when a pod is healthy and ready to be sent traffic during autoscaling and when its unhealthy and needs to be restarted. There is no client code to add this logic to in that case. Currently trtllm-serve exposes /health but that only checks if the server is up, not the health of the runtime. Would be great to have it at the executor/runtime level if possible

LinPoly:
> @LinPoly
> 
> > re implementing the check server side vs client side
> > for deployment in kubernetes we need a health and liveness probe to detect when a pod is healthy and ready to be sent traffic during autoscaling and when its unhealthy and needs to be restarted. There is no client code to add this logic to in that case. Currently trtllm-serve exposes /health but that only checks if the server is up, not the health of the runtime. Would be great to have it at the executor/runtime level if possible

So you mean K8S is using server side health check for error detection and auto-scaling? If so I think it is reasonable to add such a WAR, we can work on a more structured solution afterwards. @kaiyux for opinions.

tensorrt_llm/serve/openai_server.py
```diff
@@ -143,6 +144,41 @@ async def get_kv_cache_events(self) -> JSONResponse:
             # queue is empty, no more events
             pass
         return JSONResponse(content=events)
+    
+    async def health_generate(self) -> Response:
```
Superjomn, line 148: Currently, our `/health` endpoint is just a placeholder that always returns 200. If we implement it properly, returning an error code when there's an internal hang, would that address the motivation behind this WAR?

Alternatively, as a quick WAR, would it make sense to move `/health_generate` to `/health`, since they appear to serve the same purpose? Althrough we should have a more decent implementation in the future.

cc @kaiyux @penli9 for viz.

dsingal0: having /health return a non-200 response on internal error would be great.
Moving the /health_generate WAR to /health could work, but not everyone agrees with it because it introduces overhead. Eg if your health check is once every 10 seconds, you're generating 1 token and consuming one slot of the max_batch size every 10 seconds. For long running requests this can skew benchmarks numbers.

Superjomn: Got it. We are going to introduce a heartbeat mechanism to detect and update the system's status, which should return by `/health`. `/health_generate` is fine as we have no such endpoint/functionality yet.

Superjomn:
@kaiyux @penli9 , as we have no such check yet, I think a `/health_generate` is reasonable; we can refine the implementation during further iterations. BTW, we are going to introduce the heartbeat mechanism into the LLM-API level, and that could facilitate the implementation for `/health` -- it will check and update the status of the system periodically in a cheap way, but I think the `/health_generate` can always return a real-time status once it is invoked.

kaiyux:
@dsingal0 Can you also help fix the DCO check? https://github.com/NVIDIA/TensorRT-LLM/pull/3856/checks?check_run_id=41920082141

> Summary
Commit sha: [2cd150e](https://github.com/NVIDIA/TensorRT-LLM/pull/3856/commits/2cd150ecb4163fef41f56c3864cc14db0c30decf), Author: Dhruv Singal, Committer: Dhruv Singal; The sign-off is missing.
Commit sha: [a12154a](https://github.com/NVIDIA/TensorRT-LLM/pull/3856/commits/a12154a74c3e0c6ac9e5f526be607b8d3f7bbad7), Author: Dhruv Singal, Committer: Dhruv Singal; The sign-off is missing.
Commit sha: [06f3d21](https://github.com/NVIDIA/TensorRT-LLM/pull/3856/commits/06f3d2106875133ce1264cf339482f95cd501c13), Author: Dhruv Singal, Committer: Dhruv Singal; The sign-off is missing.

You can follow the guidance here to sign off https://github.com/NVIDIA/TensorRT-LLM/blob/main/CONTRIBUTING.md#signing-your-work.

Please let us know for any questions, thanks!

dsingal0:
@kaiyux done

tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -689,7 +689,7 @@ def sample_and_accept_draft_tokens(
         num_accepted_tokens = torch.ones(batch_size,
                                          dtype=torch.int,
                                          device=logits.device)
-
+
```
kaiyux, line 692: Seems to be unintentional changes?

tensorrt_llm/serve/openai_server.py
```diff
@@ -144,6 +157,40 @@ async def get_kv_cache_events(self) -> JSONResponse:
             pass
         return JSONResponse(content=events)
 
+    async def health_generate(self) -> Response:
```
kaiyux, line 160: Can you help add a test for this function? Please refer to https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/apps/_test_llm_server.py#L28

kaiyux:
@dsingal0 Thanks for fixing the DCO check. Mind squashing the commits? Not sure why there were a lot of un-related commits introduced, the diff looks good though.

dsingal0:
@kaiyux I think rebasing to add signoff added those commits to the PR, can we squash on merge instead? I added the test and removed the mtp.py change.

dsingal0:
alternatively, I could open a new PR

kaiyux:
> @kaiyux I think rebasing to add signoff added those commits to the PR, can we squash on merge instead? I added the test and removed the mtp.py change.

Thanks a lot! I think we should be fine merge this one since we will squash the commits when merge it, @chzblych do you think differently?

kaiyux:
@dsingal0 Can you help fix the failed test case?

```
[2025-05-09T14:06:26.867Z] ../../unittest/llmapi/apps/_test_llm_server.py .F....                    [100%]
[2025-05-09T14:06:26.867Z] 
[2025-05-09T14:06:26.867Z] =================================== FAILURES ===================================
[2025-05-09T14:06:26.867Z] _____________________________ test_health_generate _____________________________
[2025-05-09T14:06:26.867Z] 
[2025-05-09T14:06:26.867Z] client = <starlette.testclient.TestClient object at 0x7f632c50bf50>
[2025-05-09T14:06:26.867Z] 
[2025-05-09T14:06:26.867Z]     def test_health_generate(client):
[2025-05-09T14:06:26.867Z]         response = client.get("/health_generate")
[2025-05-09T14:06:26.867Z] >       assert response.status_code == 200
[2025-05-09T14:06:26.867Z] E       assert 404 == 200
[2025-05-09T14:06:26.867Z] E        +  where 404 = <Response [404 Not Found]>.status_code
[2025-05-09T14:06:26.867Z] 
[2025-05-09T14:06:26.867Z] ../../unittest/llmapi/apps/_test_llm_server.py:40: AssertionError
```

dsingal0:
@kaiyux  I believe that fixed it, the test was failing since I had it in the llm api server, not the openai server.

kaiyux:
@dsingal0 Can you help fix the style check? : )  I would like to fix it but I noticed the PR is based on your main branch and I think I should not push something onto that.

kaiyux:
Hi @dsingal0 , I fixed the style check and cherry-picked the changes to https://github.com/NVIDIA/TensorRT-LLM/pull/4349, I marked you as co-author. Please let me know for any questions.

dsingal0:
Amazing, thank you so much @kaiyux

kaiyux:
> Amazing, thank you so much @kaiyux

No problems, thanks a lot for the great support!

PR #3858, opened 2025-04-25
fix: [nvbug/5234873] Detect pmix and raise error when mpirun is not used.

yuxianq:


tensorrt_llm/__init__.py
```diff
@@ -14,6 +14,16 @@
 # limitations under the License.
 
 
+def _check_mpi_environment():
+    import os
```
Superjomn, line 18: Please left `import os` to the global scope.

tensorrt_llm/__init__.py
```diff
@@ -14,6 +14,16 @@
...
+    has_mpirun = "OMPI_COMM_WORLD_RANK" in os.environ
+    has_pmix = "PMIX_RANK" in os.environ
+    if has_pmix and not has_mpirun:
+        raise RuntimeError(
+            "PMIx is detected, but OMPI_COMM_WORLD_RANK is not set. "
+            "Please use mpirun to launch the program.")
```
Superjomn, line 24: Can srun also work here?

tensorrt_llm/__init__.py
yuxianq: I intentionally use local import to guarantee that all global imports are behind this detection to avoid MPI is imported before detection. If I move import `os` to the global scope, other developers (and AI agents) may be confused about which import should be placed before  `_check_mpi_environment`. `_add_trt_llm_dll_directory` in this file also follows the same principle (see https://github.com/NVIDIA/TensorRT-LLM/blob/883c8aee079aeab894b0a06798d1db8371807dd3/tensorrt_llm/__init__.py#L27).

tensorrt_llm/__init__.py
yuxianq: Yes, I test it with srun + pmix on computelab-next's b200, it can raise error when mpirun is absent, and import successfully when mpirun is added.

Superjomn: How about updating the message to "Please use mpirun or srun ..."?

tensorrt_llm/__init__.py
Superjomn: Got it, thanks for the explanation.

Superjomn (approved):
LGTM

qiaoxj07:
@yuxianq This PR will report RuntimeError when running on GB200 with slurm "srun xxx --mpi=pmix", I think some env from slurm ("SLURM_PROCID"...) should also be checked.

PR #3675, opened 2025-04-18
fix: FP8 kv accuracy

DylanChen-NV:
Multiply rope scaling to scaleSoftmaxLog2
GSM8K score of R1 (FP8 kv, Blackwell) 70%->96.06% (same as BF16 kv)

PR #3863, opened 2025-04-25
fix: Fix FMHA-based MLA in the generation phase and add MLA unit test

jinyangyuan-nvidia:
@PerkzZheng fixes a bug of FMHA-based MLA in the generation phase (cubins are generated using commit 56491367f0a36a4f7bba68d79b4c04f369dd5a91).
@peaceh-nv modifies the code for SM120.
@jinyangyuan-nvidia adds MLA unit test.
This PR also cherry-picks the modifications in https://github.com/NVIDIA/TensorRT-LLM/pull/3675.


DylanChen-NV (approved):

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -454,8 +454,8 @@ torch::Tensor attention(torch::Tensor q, torch::optional<torch::Tensor> k, torch
         int32_t const layer_num = host_kv_cache_pool_mapping.value().size(0);
         op->mIsMLAEnabled = true;
         op->mFP8GenerationMLA = op->mKVCacheQuantMode.hasFp8KvCache();
-        // only enable flash mla on sm90 and head_size == 576 and tokens_per_block == 64
-        op->mUseFlashMLA = tensorrt_llm::common::getSMVersion() == 90 && head_size == 576 && tokens_per_block == 64;
+        // only enable flash mla in the generation phase on sm90 and tokens_per_block == 64
+        op->mUseGenFlashMLA = tensorrt_llm::common::getSMVersion() == 90 && tokens_per_block == 64;
```
bobboli, line 458: Will `mUseGenFlashMLA` be true for context MLA? If so, will that cause problems?

jinyangyuan-nvidia: Yes, it will be `true` for context MLA. However, because FlashMLA is only called in `AttentionOp::mlaGeneration`, it will not cause a problem.

This modification is made to solve a problem related to FMHA validation check in context MLA. More specifically, a 576/512 FMHA kernel not found error will be raised, even though this kernel is never called in context MLA.

bobboli: Ok. I am fixing the problem in this [PR](https://github.com/NVIDIA/TensorRT-LLM/pull/3862) (if using FlashMLA the existence of FMHA kernel should not be checked). Making `mUseGenFlashMLA` for context phase to be true is a bit confusing, but to make this PR pass let's keep as-is and then I will change back in that PR.

bobboli (approved):

PR #3866, opened 2025-04-25
feat: refactoring dataset generation and adding tests

jdebache:
- Adding `output_format` to support exporting datasets for consumption in different benchmarking tools
- Refactoring the dataset generation in generators and exporters, clearly setting a boundary between generating a workload and writing it to a file
- Added tests for all of this


benchmarks/cpp/utils/generate.py
```diff
@@ -0,0 +1,316 @@
...
+
+    metadata = {
+        "num_requests": num_reqs,
+        "tokenize_vocabsize": tokenizer.vocab_size,
+        "max_input_len": max_input_len,
+        "max_output_len": max_output_len,
+    }
+
+    samples = [
+        TextSample(
```
FrankD412, line 166: Something that this breaks is the ability for the dataset tool to write out text prompts. The `trtllm-bench` format allows either input ids or prompts. That allows us to reuse datasets when the scenario is the same.

benchmarks/cpp/prepare_dataset.py
```diff
@@ -51,48 +52,65 @@ def get_tokenizer(cls,
...
-              default="info",
-              type=click.Choice(['info', 'debug']),
-              help="Logging level.")
+@click.option(
+    "--log-level",
+    default="info",
+    type=click.Choice(["info", "debug"]),
+    help="Logging level.",
+)
+@click.option(
```
FrankD412, line 88: We should update the performance guide since this changes the argument needed to create datasets for `trtllm-bench`.

benchmarks/cpp/prepare_dataset.py
```diff
@@ -51,48 +52,65 @@ def get_tokenizer(cls,
...
-    ctx.obj = RootArgs(tokenizer=kwargs['tokenizer'],
-                       output=kwargs['output'],
-                       std_out=kwargs['stdout'],
-                       random_seed=kwargs['random_seed'],
-                       task_id=kwargs['task_id'],
-                       rand_task_id=kwargs['rand_task_id'])
+    ctx.obj = RootArgs(
+        tokenizer=kwargs["tokenizer"],
+        output=kwargs["output"],
+        std_out=kwargs["stdout"],
```
FrankD412, line 108: This option can be removed since we're moving to the format option?

benchmarks/cpp/utils/abstractions.py
```diff
@@ -0,0 +1,182 @@
...
+    A dataset sample consisting of text and embedded media represented by their paths.
+    """
+
+    task_id: int
+    prompt: str
+    media_paths: Tuple[str, ...]
+    output_len: int
+
+
+Sample = TextSample | MultimodalSample
```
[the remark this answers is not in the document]
hypdeb: I wanted to use https://peps.python.org/pep-0613/. But sadly `yapf` does not support many of the Python 3.12 features, explicit type aliases being one of them. This seems to still work though.

> Also, you may be able to collect these at the top of the file if you use from __future__ import annotations which changes type checking to delay until the full definitions are parsed by the interpreter.

I'm not sure what you mean by that. The intent here is to give an alias to this union. Also, I think `from __future__ import annotations` is not needed anymore since Python 3.12 for forward declaring types, which I am **not** doing here.

benchmarks/cpp/prepare_dataset.py
hypdeb: The changes as they are should be backwards compatible. Meaning that there is no **need** for `trtllm-bench` users to change their usage at this time. However, they are now given the **possibility** of using this API instead of going through stdout. At some point, we can mark `--stdout` as deprecated, and then remove it, but it can be done at a later point.

benchmarks/cpp/prepare_dataset.py
hypdeb: See my other comment above about the backwards compatibility and deprecation path.

benchmarks/cpp/utils/generate.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+def _generate_lengths(
+    length_distribution: LengthDistribution,
+    num_reqs: int,
+    random_source: random.Random,
+) -> Tuple[int, ...]:
+    match length_distribution:
+        case UniformLengthDistribution() as uniform_length_distribution:
+            return _generate_lengths_uniform(
```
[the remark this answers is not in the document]
hypdeb: The distribution type is a descriptor for a distribution, not samples. They are not iterable themselves, they are 'recipes' for creating a sequence of length samples.

benchmarks/cpp/utils/generate.py
hypdeb: Ohhh I see what you mean. Okay that's easy to fix. We have 3 types of samples: ids, text and multi-modal instead of just 2. I'll fix that.

benchmarks/cpp/utils/generate.py
```diff
@@ -0,0 +1,316 @@
...
+    )
+
+    max_input_len = max(input_lens)
+    max_output_len = max(output_lens)
+
+    input_ids = _generate_random_tokens(input_lens, tokenizer, random_source)
+
+    metadata = {
+        "num_requests": num_reqs,
+        "tokenize_vocabsize": tokenizer.vocab_size,
```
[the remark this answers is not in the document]
hypdeb: What do you mean by tokenizer name? The name of the model it came with?

benchmarks/cpp/utils/generate.py
hypdeb: Hmmm, after all maybe I don't see it. How was one supposed to generate plain text samples instead of tokenized ones before? I cannot find this logic in the old code somehow.

FrankD412: Oh -- going back and looking I think you're right. I was thinking from when we get a dataset that has a prompt in it we might want to prefer the text version of that.

benchmarks/cpp/prepare_dataset.py
FrankD412: Got it -- that invalidates one of my other comments. If we aren't already, maybe print a deprecation warning?

benchmarks/cpp/prepare_dataset.py
FrankD412: Saw that. Resolved.

benchmarks/cpp/utils/generate.py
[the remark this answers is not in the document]
hypdeb: This whole metadata section seems to be a relic. It does not even appear in the tllm-bench format at all, so I'm thinking that it is not needed.

Also, some data in the samples is redundant. For example, it specifies the input length, while the input is a list of IDs, and it is therefore trivial to read the length from there.

benchmarks/cpp/prepare_dataset.py
hypdeb: Marked parameter as deprecated and added some more doc on it.

benchmarks/cpp/utils/abstractions.py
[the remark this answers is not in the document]
hypdeb: I added some documentation on these aliases. Is it okay to resolve this?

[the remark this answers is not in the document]
FrankD412: Yeah -- that looks fine.

benchmarks/cpp/utils/generate.py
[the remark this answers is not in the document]
FrankD412: Right -- but I guess my point is that that can be handled via a method within each distribution class (not necessarily an iterator, even though I think an iterator would be appropriate here).

It just seems like there's a very 1:1 with the `_generate_lengths_uniform`:`UniformLengthDistribution` etc. where it seems like the recipe could handle the logistics of generation. I don't think it's make or break for the PR, but it just seems like a way to condense.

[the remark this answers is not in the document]
hypdeb: I separate the logic from the recipe because the generation logic depends on state outside the recipe. In this case, that would be the state of the randomness source. In general, I avoid coupling logic that transforms data to the data itself, so that one can depend on the data types as lightweight abstractions with few or no dependencies. In this case, the description of a distribution can exist and be used to communicate within the program without needing to know how to sample from it.

FrankD412 (approved):

hypdeb:
Closing as outdated.

PR #3869, opened 2025-04-25
fix: [nvbugs/5066257] serialization improvments

coldwaterq:
## Description

Add an approve list into the pickle deserialization process to reduce the attack surface of using pickle to a subset of supported objects.

## Test Coverage

if a new object is added this PR will cause a ValueError to be raised that clearly states what module is not supported that is being serialized/deserialized.
Fully utilizing the zmq interfaces would test this code.

### DIFF
--- a/examples/llm-api/llm_logits_processor.py
+++ b/examples/llm-api/llm_logits_processor.py
@@ -3,7 +3,6 @@ from typing import List, Optional
 
 import torch
 
-import tensorrt_llm.executor.serialization as serialization
 from tensorrt_llm import LLM
 from tensorrt_llm.sampling_params import (BatchedLogitsProcessor,
                                           LogitsProcessor, SamplingParams)
@@ -16,6 +15,7 @@ from tensorrt_llm.sampling_params import (BatchedLogitsProcessor,
 # This simple callback will output a specific token at each step irrespective of prompt.
 # Refer to ../bindings/executor/example_logits_processor.py for a more
 # sophisticated callback that generates JSON structured output.
+# Please also refer to sampling_params.py for adding subclass to the approved class list for deserialization
 class MyLogitsProcessor(LogitsProcessor):
 
     def __init__(self, allowed_token_id: int):
@@ -68,8 +68,6 @@ class MyBatchedLogitsProcessor(BatchedLogitsProcessor):
 
 
 def main():
-    # Register MyLogitsProcessor as an approved class for deserialization across IPC boundaries.
-    serialization.register_approved_ipc_class(MyLogitsProcessor)
 
     # Batched logits processor (only supported in TensorRT backend)
     # should be specified when initializing LLM.
--- a/examples/llm-api/llm_logits_processor.py
+++ b/examples/llm-api/llm_logits_processor.py
@@ -68,7 +68,7 @@ class MyBatchedLogitsProcessor(BatchedLogitsProcessor):
 
 
 def main():
-
+    # Register MyLogitsProcessor as an approved class for deserialization across IPC boundaries.
     serialization.register_approved_ipc_class(MyLogitsProcessor)
 
     # Batched logits processor (only supported in TensorRT backend)
--- a/examples/llm-api/llm_logits_processor.py
+++ b/examples/llm-api/llm_logits_processor.py
@@ -3,6 +3,7 @@ from typing import List, Optional
 
 import torch
 
+import tensorrt_llm.executor.serialization as serialization
 from tensorrt_llm import LLM
 from tensorrt_llm.sampling_params import (BatchedLogitsProcessor,
                                           LogitsProcessor, SamplingParams)
@@ -68,12 +69,13 @@ class MyBatchedLogitsProcessor(BatchedLogitsProcessor):
 
 def main():
 
+    serialization.register_approved_ipc_class(MyLogitsProcessor)
+
     # Batched logits processor (only supported in TensorRT backend)
     # should be specified when initializing LLM.
     llm = LLM(
         model="TinyLlama/TinyLlama-1.1B-Chat-v1.0",
-        batched_logits_processor=MyBatchedLogitsProcessor(allowed_token_id=42),
-        additional_serializable_classes={"__main__": ["MyLogitsProcessor"]})
+        batched_logits_processor=MyBatchedLogitsProcessor(allowed_token_id=42))
 
     # Sample prompts
     prompts = [
--- a/examples/llm-api/llm_logits_processor.py
+++ b/examples/llm-api/llm_logits_processor.py
@@ -72,7 +72,8 @@ def main():
     # should be specified when initializing LLM.
     llm = LLM(
         model="TinyLlama/TinyLlama-1.1B-Chat-v1.0",
-        batched_logits_processor=MyBatchedLogitsProcessor(allowed_token_id=42))
+        batched_logits_processor=MyBatchedLogitsProcessor(allowed_token_id=42),
+        additional_serializable_classes={"__main__": ["MyLogitsProcessor"]})
 
     # Sample prompts
     prompts = [
--- a/tensorrt_llm/executor/ipc.py
+++ b/tensorrt_llm/executor/ipc.py
@@ -4,7 +4,7 @@ import os
 import time
 import traceback
 from queue import Queue
-from typing import Any, Dict, Optional
+from typing import Any, Optional
 
 import zmq
 import zmq.asyncio
@@ -33,7 +33,6 @@ class ZeroMqQueue:
                  is_server: bool,
                  is_async: bool = False,
                  name: Optional[str] = None,
-                 additional_serializable_classes: Optional[Dict] = None,
                  use_hmac_encryption: bool = True):
         '''
         Parameters:
@@ -57,11 +56,6 @@ class ZeroMqQueue:
         self._setup_done = False
         self.name = name
         self.socket = self.context.socket(socket_type)
-
-        # Initialize SERIALIZABLE_CLASSES with base classes plus provided additional classes
-        self.SERIALIZABLE_CLASSES = serialization.BASE_ZMQ_CLASSES.copy()
-        if additional_serializable_classes:
-            self.SERIALIZABLE_CLASSES.update(additional_serializable_classes)
 
         self.hmac_key = address[1] if address is not None else None
         self.use_hmac_encryption = use_hmac_encryption
@@ -169,12 +163,12 @@ class ZeroMqQueue:
                 raise RuntimeError("HMAC verification failed")
 
             obj = serialization.loads(
-                data, approved_imports=self.SERIALIZABLE_CLASSES)
+                data, approved_imports=serialization.BASE_ZMQ_CLASSES)
         else:
             # Receive data without HMAC
             data = self.socket.recv()
             obj = serialization.loads(
-                data, approved_imports=self.SERIALIZABLE_CLASSES)
+                data, approved_imports=serialization.BASE_ZMQ_CLASSES)
         return obj
 
     async def get_async(self) -> Any:
@@ -193,12 +187,12 @@ class ZeroMqQueue:
                 raise RuntimeError("HMAC verification failed")
 
             obj = serialization.loads(
-                data, approved_imports=self.SERIALIZABLE_CLASSES)
+                data, approved_imports=serialization.BASE_ZMQ_CLASSES)
         else:
             # Receive data without HMAC
             data = await self.socket.recv()
             obj = serialization.loads(
-                data, approved_imports=self.SERIALIZABLE_CLASSES)
+                data, approved_imports=serialization.BASE_ZMQ_CLASSES)
         return obj
 
     def close(self):
--- a/tensorrt_llm/executor/ipc.py
+++ b/tensorrt_llm/executor/ipc.py
@@ -1,15 +1,15 @@
 import hashlib
 import hmac
 import os
-import pickle  # nosec B403
 import time
 import traceback
 from queue import Queue
-from typing import Any, Optional
+from typing import Any, Dict, Optional
 
 import zmq
 import zmq.asyncio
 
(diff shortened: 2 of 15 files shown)


yibinl-nvidia:
Thanks a lot for making this PR! Several general comments:
1. Could you please follow https://github.com/NVIDIA/TensorRT-LLM/blob/main/CONTRIBUTING.md to sign your commits and run pre-commit hook.
2. Please add the complicated dataclass serialization unit test we discussed offline to `test_executor.py`. Also would be good to have additional tests for `serailization.py`. I can trigger a pipeline test after you added the tests to see if there is any test failure.

yibinl-nvidia:
@kaiyux This is the "approved list approach" I discussed with you offline. Let me know your thoughts.

cc @Superjomn @litaotju @juney-nvidia

tensorrt_llm/executor/ipc.py
```diff
@@ -26,18 +26,42 @@ class ZeroMqQueue:
         zmq.PUSH: "PUSH",
     }
 
+    # Base serializable types and their handlers
+    BASE_SERIALIZABLE_TYPES = {
```
Superjomn, line 30: Can you add more descriptions here to guide some developers in updating this? They may introduce a new type, break the tests, and when they finally switch here and the comments should help them to briefly know the background.

tensorrt_llm/executor/serialization.py
```diff
@@ -0,0 +1,101 @@
+import io
+import pickle
```
Superjomn, line 2: How about adding some unittests for this file, as it is the core of the feature, you can add a `executor/test_serialization.py` in the [unittest](https://github.com/Superjomn/TensorRT-LLM/tree/57944206ba403acf6ac2b3834b33ad15464ee6d5/tests/unittest) directory.

tensorrt_llm/executor/ipc.py
```diff
@@ -26,18 +26,42 @@ class ZeroMqQueue:
...
+
     def __init__(self,
                  address: Optional[tuple[str, Optional[bytes]]] = None,
                  *,
                  socket_type: int = zmq.PAIR,
                  is_server: bool,
                  is_async: bool = False,
                  name: Optional[str] = None,
+                 additional_serializable_types: Optional[Dict] = None,
                  use_hmac_encryption: bool = True):
```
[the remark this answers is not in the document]
coldwaterq: you could, however it's already implemented and provides an additional layer of security if anyone wants to use it. I'll change the flag to default to False in my commit though.

tensorrt_llm/executor/ipc.py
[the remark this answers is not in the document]
coldwaterq: In the commit I'm testing now I moved this into the serialization file as well. Since any error that would result in people updating the file would also originate from there.

tensorrt_llm/executor/serialization.py
coldwaterq: added in the latest commit

coldwaterq:
@yibinl-nvidia I fixed the DCO and ran the pre-commit hooks, added the tests, and made some improvements based on the comments. let me know if there is anything else I should address.

tensorrt_llm/executor/ipc.py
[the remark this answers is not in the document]
yibinl-nvidia: We didn't communicate to the user explicitly about this flag so most of them are unaware of this option. Let's keep this option on and I will consider to default to False if necessary in the future.

yibinl-nvidia:
I did a micro-benchmark comparing the custom serialization and deserialization against Pickle. The first result is based on a simple GenerationResult object (same one as in the `test_executor.py`) approximately 2 KB in size. The second result focused on benchmarking message with various size. This shows that the new approach introduces trivial μs-level of perf degradation.

```
Serialization and Deserialization Performance Results for GenerationRequest with 10000 iterations:
Pickle (avg/min/max): 199.038μs / 186.559μs / 1271.870μs
Custom (avg/min/max): 204.601μs / 192.718μs / 363.867μs
Average time difference: 5.56 μs

Size Comparison:
Pickle size: 1920 bytes
Custom size: 1920 bytes
Size ratio (custom/pickle): 1.00x

Serialization Performance Across Different Message Sizes:
Size            Pickle Avg (μs)         Custom Avg (μs)         Diff (μs)       Size Ratio
--------------------------------------------------------------------------------
1.0 KB               200.348                 205.710               5                    1.00x
10.0 KB              279.806                 285.365               5                    1.00x
100.0 KB            1044.078                1051.731               7                    1.00x
1.0 MB              8487.678                8505.488              17                    1.00x
10.0 MB            92523.012               93069.062             546                    1.00x
```

coldwaterq:
The last commit should resolve the issue reported by blossom.

coldwaterq:
the tests that were failed in the basic bot run should be fixed with those new commits.

tensorrt_llm/executor/postproc_worker.py
```diff
@@ -77,6 +79,7 @@ def __init__(
         tokenizer_dir: str,
         record_creator: Callable[
             ["PostprocWorker.Input", TransformersTokenizer], Any],
+        BASE_ZMQ_CLASSES: Dict,
```
coldwaterq, line 82: @yibinl-nvidia there are still a couple of changes like this in order to pass the updated approved list across process boundaries, but I think that's unavoidable and it should be invisible to the end user. Do foresee any issue with this?

yibinl-nvidia: This should be fine as long as it is not directly exposed to the user.

examples/llm-api/llm_logits_processor.py
```diff
@@ -67,6 +68,8 @@ def __call__(self, req_ids: List[int], logits: List[torch.Tensor],
 
 
 def main():
+    # Register MyLogitsProcessor as an approved class for deserialization across IPC boundaries.
+    serialization.register_approved_ipc_class(MyLogitsProcessor)
```
kaiyux, line 72: Should we somehow hide this line inside of the LLM API?

@Superjomn Thoughts on this line?

yibinl-nvidia: This is the design changes I discussed with @coldwaterq. The original idea was to append LLM API with an additional optional arg for the supplementary approved class, but this approach means API breaking change and it seems that the example above is the only place needed such change, so we ought to add `serialization.register_approved_ipc_class` instead. And I think it is necessary to highlight the changes in this example because user can define their own `LogitsProcessor` class and they would confuse why serialization fail without proper example.

coldwaterq: Yep, this solution also allows the creator of the object to simply pass it into this function which resolves how pickle will identify it even if the module of the object changes between runs. Such as if the script is run directly sometimes and used as a module in others.

Superjomn: Is it possible to make all the inheritors of `BatchedLogitsProcessor` be registered automatically? To use a logits-processor, users should always inherit that class. 

Got the following example from Cursor:
```python
# Assume this function is defined elsewhere
def register_approved_ipc_class(dtype):
  """
  Registers the given class type.
  (This is a placeholder for your actual registration logic)
  """
  print(f"Registering IPC class: {dtype.__name__}")
  # In a real scenario, you would add dtype to a list, dictionary,
  # or perform some other registration action.


class BaseClass:
  def __init_subclass__(cls, **kwargs):
    """
    This method is called when a class inherits from BaseClass.
    """
    super().__init_subclass__(**kwargs) # Call the superclass's __init_subclass__
    register_approved_ipc_class(cls)   # Register the new subclass


# Now, any class that inherits from BaseClass will automatically be registered.

class ChildClassA(BaseClass):
  pass

class ChildClassB(BaseClass):
  def __init__(self, name):
    self.name = name

class GrandChildClass(ChildClassA):
  pass

# Example of creating instances (not strictly necessary to show registration)
# child_a_instance = ChildClassA()
# child_b_instance = ChildClassB("Test")
# grand_child_instance = GrandChildClass()
```

Ideally, users should have little chance to call `register_approved_ipc_class` explicitly.

coldwaterq: That should work, however, I would propose that ideally, users should call register_approved_ipc_class anytime they create a custom sublcass that needs it. Because this is effectively our last warning that the class will be a part of the security boundary and them confirming they believe that is not a security risk.

If we want to hide the register_approved_ipc_class call from the developers creating the classes, then we should consider adding in warnings in some other way.

yibinl-nvidia: > Because this is effectively our last warning that the class will be a part of the security boundary and them confirming they believe that is not a security risk.

The argument I would have is that we should try our best to make our own codebase to be safe, but there is no way we can enforce any safety rules to the user who creates their own classes. As such, even we put a warning about this security boundary, the user needs to understand the context about how we do serialization in IPC. I doubt it is necessary to disclose this.

tensorrt_llm/sampling_params.py
```diff
@@ -69,6 +70,13 @@ def __call__(self, req_id: int, logits: torch.Tensor,
         """
         pass  # noqa
 
+    def __init_subclass__(cls, **kwargs):
```
coldwaterq, line 73: `super().__init_subclass__(**kwargs)` should probably be added to this as well.

yibinl-nvidia: good catch!

tensorrt_llm/serve/openai_protocol.py
```diff
@@ -20,6 +21,14 @@ class OpenAIBaseModel(BaseModel):
     # OpenAI API does not allow extra fields & allow to initialize by both alias and field name
     model_config = ConfigDict(extra="forbid", populate_by_name=True)
 
+    def __init_subclass__(cls, **kwargs):
```
yibinl-nvidia, line 24: @LinPoly @Superjomn could you please review the change for `OpenaiBaseModel`? My understanding is that we don't expect to add new subclass of `OpenaiBaseModel` so allow all subclass to be deserializable is safe to do.

tensorrt_llm/executor/ipc.py
```diff
@@ -116,26 +116,26 @@ def poll(self, timeout: int) -> bool:
     def put(self, obj: Any):
         self.setup_lazily()
         with nvtx_range_debug("send", color="blue", category="IPC"):
+            data = serialization.dumps(obj)
             if self.use_hmac_encryption:
```
Superjomn, line 120: @yibinl-nvidia should we eliminate the HMAC code? Is the serialization enhancement already sufficient for safety? If HMAC contributes little based on this PR, we can clean up the code to reduce the complexity.

tensorrt_llm/serve/openai_protocol.py
LinPoly: IIUC, `register_approved_ipc_class` will automatically add all sub-classes of `OpenaiBaseModel` into collections of legal IPC object, if we add new subclass, it will also be registered, why would it be unsafe if new sub-class is added?

tensorrt_llm/executor/ipc.py
yibinl-nvidia: Sure, I can file a separate PR to remove HMAC.

tensorrt_llm/serve/openai_protocol.py
yibinl-nvidia: @LinPoly The purpose of this PR is to only allow deserialization for approved class so that hacker cannot inject malicious class in deserialization. It would be unsafe if the new subclass is prone to attack. 

Put the question slightly differently, do you know if we plan to add new subclass to `OpenaiBaseModel`? If so, what kind of subclass will be added?

yibinl-nvidia:
@Superjomn @kaiyux Could you re-review this PR as pipeline finally passed. Thanks!

tensorrt_llm/executor/postproc_worker.py
Superjomn: Can we use the central ·BASE_ZMQ_CLASSES· without exposing it in the argument lists of classes and functions? Custom classes can be registered to this central dictionary as well. If all registrations happen at class definition time, we won’t need to worry about syncing the dictionary across processes.

tensorrt_llm/serve/openai_protocol.py
LinPoly: So our current subclasses are not prone to attack? If so I think it would be safe to register in this way since new subclasses (if there would be) will be very similar to what we have for now.

tensorrt_llm/executor/postproc_worker.py
yibinl-nvidia: @coldwaterq I vaguely remember the reason we do this in class initialization is that each process / worker might have different approved list, but it seems unnecessary since BASE_ZMQ_CLASSES is going to be static throughout the execution. And I agree with Chunwei's point here that it simplifies the code. Thoughts?

tensorrt_llm/serve/openai_protocol.py
yibinl-nvidia: Yes, we did assessment on the current subclass and they are safe. Thanks for the confirmation.

tensorrt_llm/executor/postproc_worker.py
coldwaterq: This is a subprocess call, so the static list that was dynamically modified by subclass initialization has to be passed into the subprocess otherwise any class initialized only by the parent process will be blocked by the child processes deserialization

Superjomn: Got it, thanks for the explanation.

Superjomn (approved):
LGTM as the first milestone. 

IIUC, there are the following TODOs:
1. Cleanup the existing HMAC logic @yibinl-nvidia 
2. Simplify the passing for BASE_ZMQ_CLASSES in intra-process scenarios @Superjomn

yibinl-nvidia (approved):

yibinl-nvidia (approved):

PR #3881, opened 2025-04-26
feat: integrate modelopt hf export into our quantization script

jdebache:
Hey @QiJune, this is an alternative path to enabling export to HF, by modifying our existing quantization logic instead of adding a new script. Would that work better?


jdebache:
Hey @QiJune , I might need your advice on this. I'm now getting a
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0! 
```
when running the quantization script. It seems there are other reports of a similar issue, both internally and externally: https://github.com/huggingface/transformers/issues/37545. Is this already on our radar?

hypdeb:
I solved the issue above. @QiJune, could you please give it another review?

hypdeb:
I worked around this by using modelopt directly, bypassing the quantization utilities from trtllm entirely.

PR #3885, opened 2025-04-27
feat: support multi lora adapters and TP

shaharmor98:
# PR title

TP & multi adapters LoRA  support

## Description

In this PR we add TP support and multi adapters support.  
Both features are tested with designated tests.   

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


tensorrt_llm/_torch/model_config.py
```diff
@@ -175,7 +174,7 @@ def get_bindings_model_config(
         from tensorrt_llm.bindings import ModelConfig as ModelConfigCpp
 
         num_heads = self.pretrained_config.num_attention_heads // (
-            tensor_parallelism * context_parallelism)
+            self.mapping.tp_size * context_parallelism)
```
byshiue, line 177: If we replace `tensor_parallelism` by `self.mapping.tp_size`, can we replace by `context_paralleism` by `self.mapping.cp_size`, too?

byshiue:
I find that https://github.com/NVIDIA/TensorRT-LLM/blob/main/examples/pytorch/quickstart_lora.py is duplicated to https://github.com/NVIDIA/TensorRT-LLM/blob/main/examples/llm-api/llm_multilora.py. So, I think we only need to keep the latter one.

tensorrt_llm/_torch/model_config.py
shaharmor98: DONE

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -422,9 +433,19 @@ def create_py_executor_instance(dist,
             max_cpu_loras,
         )
 
+        from tensorrt_llm.bindings import WorldConfig
+        world_config = WorldConfig(
+            tensor_parallelism=mapping.tp_size,
+            pipeline_parallelism=mapping.pp_size,
+            context_parallelism=1,
```
byshiue, line 442: Can set `mapping.cp_size` directly. Don't verify the correctness of cp now.

shaharmor98: DONE

tensorrt_llm/_torch/modules/mlp.py
```diff
@@ -48,12 +50,12 @@ def __init__(self,
...
             quant_config=config.get_quant_config(),
-            skip_create_weights_in_init=config.skip_create_weights_in_init)
+            skip_create_weights_in_init=config.skip_create_weights_in_init,
+            lora=self.down_lora)
 
-        self.up_lora = LoraLayer([LoraModuleType.MLP_H_TO_4H],
-                                 [self.intermediate_size])
-        self.down_lora = LoraLayer([LoraModuleType.MLP_4H_TO_H],
-                                   [self.hidden_size])
+        self.up_lora = LoraLayer(
```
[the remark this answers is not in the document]
shaharmor98: My bad, good catch

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -358,9 +359,9 @@ def __init__(self, model: TModel, *, config: ModelConfig[TConfig],
...
 
-                # TODO smor- hack
-                if hasattr(config,
-                           'lora_config') and config.lora_config is not None:
+                if hasattr(config, 'lora_config'
+                           ) and config.lora_config is not None and len(
+                               config.lora_config.lora_dir) == 1:
                     from tensorrt_llm.lora_manager import HfLoraLoader
                     lora_loader = HfLoraLoader(config.lora_config.lora_dir)
                     weight = lora_loader.lm_head
```
syuoni, line 367: Does lora adapter checkpoint always have `lm_head` weight?

tensorrt_llm/llmapi/llm.py
```diff
@@ -613,7 +613,23 @@ def _try_load_tokenizer(self) -> Optional[TokenizerBase]:
         if hasattr(
                 self.args, "backend"
         ) and self.args.backend == "pytorch" and self.args.lora_config is not None:
-            tokenizer_path = self.args.lora_config.lora_dir[0]
+            lora_dirs = len(self.args.lora_config.lora_dir)
```
syuoni, line 616: `lora_dirs` -> `num_lora_dirs`

tests/unittest/_torch/modeling/test_modeling_llama.py
```diff
@@ -398,3 +399,71 @@ def test_llama_lora(self) -> None:
...
+                               sampling_params,
+                               lora_request=[
+                                   None, lora_req1, lora_req2, None, lora_req1,
+                                   lora_req2
+                               ])
+        for output, ref in zip(outputs, references):
+            assert similar(output.outputs[0].text, ref)
+
+    @skip_single_gpu
+    def test_llama_v2_13b_lora_tp2_gpus(self) -> None:
```
syuoni, line 445: There are LLM API lora tests in https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/test_llm.py#L1321-L1393

You should use these harness to implement pytorch flow LLM API tests in `unittest/llmapi/test_llm_pytorch.py` or `unittest/llmapi/test_llm_multi_gpu_pytorch.py`.

tests/unittest/llmapi/test_llm.py
```diff
@@ -1318,6 +1318,34 @@ def test_executor_lookahead_decoding_config():
...
+        "今天天气很好，我到公园的时候，",
+    ]
+    references = [
+        "发现公园里到处都是人，有的在跑步，有的在打羽毛球，还有的",
+    ]
+    lora_req = LoRARequest("Chinese", 1, hf_lora_dir)
+    sampling_params = SamplingParams(max_tokens=20, add_special_tokens=False)
+    outputs = llm.generate(prompts, sampling_params, lora_request=[lora_req])
+    for output, ref in zip(outputs, references):
+        assert similar(output.outputs[0].text, ref)
```
syuoni, line 1346: This is the test file for single-gpu. Please add multi-gpu tests to `test_llm_multi_gpu.py`.

What is the difference of this test from `llama_v2_13b_lora_test_harness`?

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -98,6 +101,7 @@ def __init__(
             tensor_parallel_mode=TensorParallelMode.ROW,
             quant_config=config.get_quant_config(),
             skip_create_weights_in_init=config.skip_create_weights_in_init,
+            lora=self.o_lora,
```
syuoni, line 104: If we register the lora module inside the linear layer, it seems not necessary to register it at Attention module, too?

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -113,9 +117,6 @@ def __init__(
         self.fused_qkv_lora = LoraLayer([LoraModuleType.ATTENTION_QKV],
                                         [self.q_size + 2 * self.kv_size])
```
syuoni, line 118: Why the qkv lora module is treated differently? Is it possible to unify it like `o_lora`?

tensorrt_llm/_torch/modules/gated_mlp.py
```diff
@@ -86,18 +89,20 @@ def __init__(self,
             quant_config=config.get_quant_config(),
             reduce_output=reduce_output,
             skip_create_weights_in_init=config.skip_create_weights_in_init,
+            lora=self.down_lora,
         )
```
syuoni, line 93: Same here. Seems not necessary to register `down_lora` at MLP module, too.

tensorrt_llm/_torch/modules/gated_mlp.py
```diff
@@ -86,18 +89,20 @@ def __init__(self,
...
-                                            [2 * self.intermediate_size])
-        self.down_lora = LoraLayer([LoraModuleType.MLP_4H_TO_H],
-                                   [self.hidden_size])
+            [LoraModuleType.MLP_H_TO_4H, LoraModuleType.MLP_GATE], [
+                self.intermediate_size // mapping.tp_size,
+                self.intermediate_size // mapping.tp_size
+            ])
+        self.fused_gate_up_lora = LoraLayer(
+            [LoraModuleType.MLP_GATE_UP],
+            [2 * self.intermediate_size // mapping.tp_size])
```
syuoni, line 105: Oh, the `up_lora` is also managed differently from `down_lora`. It would be better to unify them.

tensorrt_llm/_torch/modules/gated_mlp.py
```diff
@@ -107,7 +112,7 @@ def forward(
         min_latency_mode: Optional[bool] = False,
         lora_params: Optional[dict] = None,
     ) -> torch.Tensor:
-        if lora_params is not None:
+        if bool(lora_params):
```
syuoni, line 115: Is there any special reason for this? Normally we prefer it to be explicit: `lora_params is not None`.

litaotju:
Thx @shaharmor98 for the change.
I am adding @hlu1 into the review, since I know Hao is leading pytorch SW. And its a chance for us to get the lora support clean/right in the first place. Thx

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -764,9 +766,17 @@ def __init__(self, model_config: ModelConfig[LlamaConfig]):
             gather_output=True,
         )
 
-        if hasattr(model_config,
-                   'lora_config') and model_config.lora_config is not None:
+        if hasattr(
+                model_config,
+                'lora_config') and model_config.lora_config is not None and len(
+                    model_config.lora_config.lora_dir) == 1:
```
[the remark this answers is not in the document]
shaharmor98: As for the first part- you already saw it in multiple places, and I've yet decided what is the right thing to do here. The problem here, and I'm not even sure it is supported in the legacy flow, what would happen if you provide multiple lora adapters, and each one of them has its own lm_head and embedding?  
There are multiple scenarios that need to be addressed, and until now I haven't handled them.  
I think it should be addressed, but in a different PR since we are in a time crunch here and want to make it to the first release on time.  

As for your suggestion- I've already mitigated some of this duplication, but in the subsequent PR which will be opened right after this one is merged. Can we revisit this in the next PR?

tensorrt_llm/_torch/models/modeling_utils.py
shaharmor98: No, it doesn't.  
Same applies here- I've already addressed that, but on the subsequent PR which will be opened right after this one

tensorrt_llm/llmapi/llm.py
shaharmor98: Fixed

tests/unittest/_torch/modeling/test_modeling_llama.py
shaharmor98: I didn't know these test files exist.  
Moved my tests to the relevant places.

tests/unittest/llmapi/test_llm.py
shaharmor98: 1. Moved
2. This test uses TP=2, there was no such a test that I could find which verifies correctness on TP>1

shaharmor98:
> Thx @shaharmor98 for the change. I am adding @hlu1 into the review, since I know Hao is leading pytorch SW. And its a chance for us to get the lora support clean/right in the first place. Thx

Great. @litaotju @hlu1 This PR is a continuation of previously merged PRs.
If anything is unclear to any of you, I’d be happy to elaborate in greater detail.

tensorrt_llm/_torch/modules/attention.py
shaharmor98: I'm not sure I understand your meaning here?  
Specifically for attention, we have two LoRA parts that should be addressed- the qkv and the dense layer.  
The main reason behind integrating specific support for linear lora is that we want the all reduce to be applied for the linear result part

tensorrt_llm/_torch/modules/attention.py
shaharmor98: The main reason is all reduce logic that is part of the Linear layer. In legacy, it goes quite the same, I haven't reformatted that

tensorrt_llm/_torch/modules/gated_mlp.py
shaharmor98: The main reason is that the lora_params dictionary could be empty, and `bool(lora_params)` will be False in that case, whereas it will be `True` for `lora_params is not None`

tensorrt_llm/_torch/modules/gated_mlp.py
shaharmor98: I'll wait to see what you reply on the previous comments. Not sure I fully understand/agree with what you clain

tensorrt_llm/_torch/modules/gated_mlp.py
shaharmor98: Same as before.  
Overall, this changes are first and foremost intend to bring initial support. I agree that we could've consider refactoring some parts, but at the moment it's 0.20 release blocker, so I think we should first converge on full working flow, and then improving the rest. What do you think about that?

tensorrt_llm/_torch/modules/attention.py
syuoni: This might be a more concerning point to me:
* For `o_proj` and `down_proj`, the LoRA module is hidden in the linear layer;
* For `qkv_proj` and `up_proj`, the LoRA is handled outside the linear layer.

We'd better to make them in a consistent way.

tensorrt_llm/_torch/modules/gated_mlp.py
syuoni: Clarify in another thread, closing this one.

tensorrt_llm/_torch/modules/gated_mlp.py
syuoni: Clarify in another thread, closing this one.

tensorrt_llm/_torch/modules/gated_mlp.py
syuoni: So it do rely on the case of an empty dict. Then, it's fine with me.

syuoni (approved):
Hi @shaharmor98 ,

For the current PR status, I'm still concerned with
* the duplication of test code;
* the difference between how LoRA module is applied to `o_proj/down_proj` and `qkv_proj/up_proj`.

Since the feature itself is functional and it's a 0.20 release blocker, it's fine with me to get this PR merged first and deal with the issues in subsequent PRs.

Thanks!

tensorrt_llm/_torch/models/modeling_llama.py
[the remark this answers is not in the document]
byshiue: I think it is fine to use multi-lora in such cases. Users can use multiple loras who have same finetuned emb/lm_head.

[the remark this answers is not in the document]
shaharmor98: It's an interesting point, and I think we should revisit this later on.  
As @syuoni said, there are at least three scenarios that should be addressed that we currently ignore.

tensorrt_llm/_torch/models/modeling_utils.py
shaharmor98: Oh, I understand now.  
It's pretty ugly, and perhaps assumes the wrong things, but- I haven't stumbled upon adapters that have custom embedding and not lm_head.  
Potentially, you're 100% correct.  
At the moment, I make a distinction of `LoraConfig` with a defined `lora_dir`, because I assume (subsequent PR mitigate that false assumption a bit), that if you provide a `lora_dir` then it's probably has it's own custom lm_head and embedding.  
The following cases are not supported, and should be:  
1. Two `lora_dirs`, one with custom lm_head, the second without.  
2. Two `lora_dirs`, both with different custom lm_head

The third option (both with same lm_head) is supported

tensorrt_llm/_torch/modules/attention.py
shaharmor98: Yes, okay, I now understand what you mean. You're again correct.  
It's worth the change- but only due to the time limits we have I think we should revisit this in a different PR. I don't want it to stall the entire release.  
But I totally agree with your point, it's a good one

tensorrt_llm/_torch/modules/attention.py
shaharmor98: Fully agree

tests/unittest/llmapi/test_llm_multi_gpu.py
```diff
@@ -493,3 +495,31 @@ def test_llm_abort_request_tp2(llm_for_sampling_params_tp2: LLM,
...
+        "今天天气很好，我到公园的时候，",
+    ]
+    references = [
+        "发现公园里到处都是人，有的在跑步，有的在打羽毛球，还有的",
+    ]
+    lora_req = LoRARequest("Chinese", 1, hf_lora_dir)
+    sampling_params = SamplingParams(max_tokens=20, add_special_tokens=False)
+    outputs = llm.generate(prompts, sampling_params, lora_request=[lora_req])
+    for output, ref in zip(outputs, references):
+        assert similar(output.outputs[0].text, ref)
```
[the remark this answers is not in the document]
shaharmor98: Okay, I didn't know the file (nor the tests) in `test_llm_multi_gpu.py`.  
Now everything makes sense to me.  
Although it's not "critical" I still think it's better for me to change it in this PR.  
You can take another look if you like, I've changed my tests accordingly (and removed the 100% duplicated one)

tensorrt_llm/_torch/models/modeling_utils.py
syuoni: I see, thanks for the explanation.

tests/unittest/llmapi/test_llm_pytorch.py
```diff
@@ -82,3 +86,85 @@ def test_llm_with_postprocess_parallel_and_result_handler(streaming):
...
+        "task-0", 0,
+        f"{llm_models_root()}/llama-models-v2/chinese-llama-2-lora-13b")
+    lora_request = [lora_req]
+
+    outputs = llm.generate(prompts, sampling_params, lora_request=lora_request)
+
+    assert similar(outputs[0].outputs[0].text, references[0])
+
+
+def llama_7b_multi_lora_test_harness(**llm_kwargs) -> None:
```
syuoni, line 119: Apologize if I'm being too nitpicking.

The tests look much better now. But the `llama_7b_multi_lora_test_harness` and `llama_v2_13b_lora_test_harness` in this file are quite similar to the two harness functions in `tests/unittest/llmapi/test_llm.py`. I believe they can be deduplicated. 

Take guided decoding as an example, it's supported in both TensorRT and PyTorch flows. Take a look at the below four links:
* https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/test_llm.py#L1102
* https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/test_llm_multi_gpu.py#L237
* https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/test_llm_pytorch.py#L18
* https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/test_llm_multi_gpu_pytorch.py#L10

The four tests (TensorRT/PyTorch x single/multi-gpu) share the same harness function.

Feel free to address this in another PR. Thanks!

shaharmor98: Thanks!  
Looks much better now :)

PR #3911, opened 2025-04-28
fix[nvbug-5228840]: Add debug log memory infomation for memory allocation error

HuiGao-NV:
CI case failures happens flaky. Add debug log so that we can check memory status when failures happen next time.


HuiGao-NV:
nvbug has been fixed by Kaiyu. Close this PR.

PR #3769, opened 2025-04-22
Nixl infra

zackyoray:
Add nixl infrastructure to trtllm

[Split from https://github.com/NVIDIA/TensorRT-LLM/pull/3677]

--------

PR #3934, opened 2025-04-29
feat: NIXL interface integration

Shixiaowei02:
This PR, in conjunction with [PR 3769](https://github.com/NVIDIA/TensorRT-LLM/pull/3769) , provides an interface solution for dynamically linking NIXL.

image

### DIFF
--- a/cpp/tensorrt_llm/CMakeLists.txt
+++ b/cpp/tensorrt_llm/CMakeLists.txt
@@ -148,6 +148,11 @@ add_subdirectory(batch_manager)
 set(EXECUTOR_TARGET tensorrt_llm_executor_static)
 set(EXECUTOR_TARGET_ARCH ${TARGET_ARCH})
 set(UCX_WRAPPER_TARGET tensorrt_llm_ucx_wrapper)
+
+if(NIXL_ROOT)
+  set(NIXL_WRAPPER_TARGET tensorrt_llm_nixl_wrapper)
+endif()
+
 add_subdirectory(executor)
 
 find_package(Threads REQUIRED)
@@ -231,6 +236,11 @@ target_link_libraries(${EXECUTOR_TARGET} INTERFACE ${SHARED_TARGET})
 if(TARGET ${UCX_WRAPPER_TARGET})
   target_link_libraries(${UCX_WRAPPER_TARGET} INTERFACE ${SHARED_TARGET})
   add_dependencies(${SHARED_TARGET} ${UCX_WRAPPER_TARGET})
+endif()
+
+if(TARGET ${NIXL_WRAPPER_TARGET})
+  target_link_libraries(${NIXL_WRAPPER_TARGET} INTERFACE ${SHARED_TARGET})
+  add_dependencies(${SHARED_TARGET} ${NIXL_WRAPPER_TARGET})
 endif()
 
 if(NOT WIN32)
--- a/cpp/tensorrt_llm/executor/cache_transmission/nixl_utils/CMakeLists.txt
+++ b/cpp/tensorrt_llm/executor/cache_transmission/nixl_utils/CMakeLists.txt
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025 NVIDIA CORPORATION &
+# AFFILIATES. All rights reserved. SPDX-License-Identifier: NVIDIA TensorRT
+# Source Code License Agreement
+#
+# NVIDIA CORPORATION, its affiliates and licensors retain all intellectual
+# property and proprietary rights in and to this material, related documentation
+# and any modifications thereto. Any use, reproduction, disclosure or
+# distribution of this material and related documentation without an express
+# license agreement from NVIDIA CORPORATION or its affiliates is strictly
+# prohibited.
+
+if(NOT CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64")
+  message(
+    STATUS
+      "The NIXL backend is temporarily unavailable on the aarch64 platform.")
+  unset(NIXL_ROOT)
+endif()
+
+if(NIXL_ROOT)
+  find_package(NIXL REQUIRED)
+  # Check if all required packages were found
+  if(NOT NIXL_FOUND)
+    message(
+      FATAL_ERROR "NIXL not found. Please install NIXL or set `NIXL_ROOT`.")
+  endif()
+
+  # Define the NIXL wrapper target
+  set(NIXL_WRAPPER_TARGET "tensorrt_llm_nixl_wrapper")
+
+  # Add the NIXL wrapper target
+  add_library(${NIXL_WRAPPER_TARGET} SHARED transferAgent.cpp)
+  target_compile_options(
+    ${NIXL_WRAPPER_TARGET} PRIVATE -U_GLIBCXX_USE_CXX11_ABI
+                                   -D_GLIBCXX_USE_CXX11_ABI=1 -Wno-error)
+
+  # Add include directories
+  target_include_directories(${NIXL_WRAPPER_TARGET} PRIVATE NIXL::nixl)
+
+  # Link against all NIXL libraries
+  target_link_libraries(${NIXL_WRAPPER_TARGET} PRIVATE NIXL::nixl)
+
+  add_custom_command(
+    TARGET ${NIXL_WRAPPER_TARGET}
+    POST_BUILD
+    COMMAND nm -C $<TARGET_FILE:${NIXL_WRAPPER_TARGET}> | grep -E
+            'tensorrt_llm.*__cxx11|__cxx11.*tensorrt_llm' && exit 1 || exit 0)
+endif()
--- a/cpp/tensorrt_llm/executor/cache_transmission/nixl_utils/transferAgent.cpp
+++ b/cpp/tensorrt_llm/executor/cache_transmission/nixl_utils/transferAgent.cpp
@@ -1,0 +1,180 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "tensorrt_llm/executor/cache_transmission/nixl_utils/transferAgent.h"
+
+namespace tensorrt_llm::executor::kv_cache
+{
+
+[[nodiscard]] nixl_mem_t NixlHelper::convert(MemoryType type)
+{
+    switch (type)
+    {
+    case MemoryType::kDRAM: return DRAM_SEG;
+    case MemoryType::kVRAM: return VRAM_SEG;
+    case MemoryType::kBLK: return BLK_SEG;
+    case MemoryType::kOBJ: return OBJ_SEG;
+    case MemoryType::kFILE: return FILE_SEG;
+    default: TLLM_THROW("Unknown MemoryType value");
+    }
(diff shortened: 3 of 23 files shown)


cpp/tensorrt_llm/executor/cache_transmission/nixl_utils/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+# NVIDIA CORPORATION, its affiliates and licensors retain all intellectual
+# property and proprietary rights in and to this material, related documentation
+# and any modifications thereto. Any use, reproduction, disclosure or
+# distribution of this material and related documentation without an express
+# license agreement from NVIDIA CORPORATION or its affiliates is strictly
+# prohibited.
+
+if(ENABLE_NIXL)
+  if(NOT CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64")
+    message(STATUS "NIXL currently only supports aarch64.")
```
zackyoray, line 14: aarch64. >> x86_64

MartinMarciniszyn (changes requested):
@Shixiaowei02 , you will need to move the installation of NIXL back into the `devel` stage in docker. Otherwise, this MR will break the build of the TRT-LLM release image since you introduce a compile time dependency if I understand this correctly.

cpp/tensorrt_llm/executor/cache_transmission/nixl_utils/transferAgent.cpp
```diff
@@ -0,0 +1,180 @@
...
+void NixlTransferAgent::registerMemory(RegisterDescs const& descs)
+{
+    nixl_status_t status;
+    status = mRawAgent->registerMem(NixlHelper::convertRegDlist(descs), &mExtraParams);
+    TLLM_CHECK(status == NIXL_SUCCESS);
+
+    std::string localMD;
+    status = mRawAgent->getLocalMD(localMD);
+    TLLM_CHECK(status == NIXL_SUCCESS);
+    mRegistrar->addAgentDesc(mName.c_str(), std::vector<char>(localMD.begin(), localMD.end()));
```
zackyoray, line 121: It seems mAgentDesc can have multiple localMD, am i correct? 
why we need to store it and cant relay on the getLocalMD() later when calling to getAgentDesc?

cpp/tensorrt_llm/executor/cache_transmission/ucx_utils/connection.h
```diff
@@ -21,6 +21,7 @@
 #include "ucxx/utils/sockaddr.h"
 #include "ucxx/utils/ucx.h"
 #include <cstdint>
+#include <future>
```
zackyoray, line 24: why its needed in this PR?

Shixiaowei02: The header file was originally in the base class file, but since it was not actually used in the base class, it was organized into the subclass.

Shixiaowei02:
> @Shixiaowei02 , you will need to move the installation of NIXL back into the `devel` stage in docker. Otherwise, this MR will break the build of the TRT-LLM release image since you introduce a compile time dependency if I understand this correctly.

@MartinMarciniszyn Okay, I have modified the installation in the image, which will increase the overall image size by approximately 150 MB. I will make further modifications to install precompiled libraries directly instead of building from source to shrink the image size. Currently, NIXL can be disabled during the build, but it will be required for testing in the pre-merge/post-merge pipeline.

docker/Dockerfile.multi
```diff
@@ -72,6 +72,13 @@ RUN bash ./install_pytorch.sh $TORCH_INSTALL_TYPE && rm install_pytorch.sh
 RUN pip3 uninstall -y opencv && rm -rf /usr/local/lib/python3*/dist-packages/cv2/
 RUN pip3 install opencv-python-headless --force-reinstall --no-deps --no-cache-dir
 
+# Install NIXL
+COPY docker/common/install_nixl.sh install_nixl.sh
+RUN bash ./install_nixl.sh && rm install_nixl.sh
```
MartinMarciniszyn, line 77: I know I encouraged this change, but I am having second thoughts on this since it will break building with docker using the multi-builder and QEMU. Instead, could we guard the usage of the NIXL library with a CMake option so that we do not use the library by default? In that case we could move it back into the `tritondevel` stage.

jenkins/L0_MergeRequest.groovy
```diff
@@ -21,10 +21,10 @@ UPLOAD_PATH = env.uploadPath ? env.uploadPath : "sw-tensorrt-generic/llm-artifac
 // Container configuration
 // available tags can be found in: https://urm.nvidia.com/artifactory/sw-tensorrt-docker/tensorrt-llm/
 // [base_image_name]-[arch]-[os](-[python_version])-[trt_version]-[torch_install_type]-[stage]-[date]-[mr_id]
-LLM_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:pytorch-25.04-py3-x86_64-ubuntu24.04-trt10.10.0.31-skip-tritondevel-202505121727-4049"
-LLM_SBSA_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:pytorch-25.04-py3-aarch64-ubuntu24.04-trt10.10.0.31-skip-tritondevel-202505121727-4049"
-LLM_ROCKYLINUX8_PY310_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:cuda-12.9.0-devel-rocky8-x86_64-rocky8-py310-trt10.10.0.31-skip-tritondevel-202505121727-4049"
-LLM_ROCKYLINUX8_PY312_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:cuda-12.9.0-devel-rocky8-x86_64-rocky8-py312-trt10.10.0.31-skip-tritondevel-202505121727-4049"
+LLM_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:pytorch-25.03-py3-x86_64-ubuntu24.04-trt10.9.0.34-skip-tritondevel-202505131405-3934"
```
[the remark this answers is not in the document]
Shixiaowei02: I have identified the issue and are currently updating the local image. Thanks.

docker/Dockerfile.multi
Shixiaowei02: It appears that NIXL encounters build errors in certain aarch64 scenarios. Its support for aarch64 is currently incomplete. I have updated the script to prevent it from building in aarch64 scenarios.

MartinMarciniszyn: Should we still have an explicit CMake option? It would be great if we could disable building this for now by default on all architectures. Since we are developing a new feature, let's enable it explicitly when needed for now.

cpp/tensorrt_llm/CMakeLists.txt
```diff
@@ -148,6 +148,7 @@ add_subdirectory(batch_manager)
 set(EXECUTOR_TARGET tensorrt_llm_executor_static)
 set(EXECUTOR_TARGET_ARCH ${TARGET_ARCH})
 set(UCX_WRAPPER_TARGET tensorrt_llm_ucx_wrapper)
+set(NIXL_WRAPPER_TARGET tensorrt_llm_nixl_wrapper)
```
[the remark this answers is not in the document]
Shixiaowei02: Done. Thanks!

docker/Dockerfile.multi
MartinMarciniszyn: Yes, this looks good. Can we still keep the installation of ETCD and NIXL in the `tritondevel` stage? This will be used in CI and for local dev. There is no need to have this the `devel` stage. Having it here will break the QEMU build.

MartinMarciniszyn: Please we will have validation that the build actually works without having the library installed.

MartinMarciniszyn: This is acceptable *as long as the QEMU based docker build is not broken by that*. You can test this by running

```
make -C docker Dngc-devel_build
```

Note that you will need to install QEMU and a multi-platform capable docker first.

In any case, I don't see exactly what we gain by moving this back to the `devel` stage. What is the issue with leaving this in `tritondevel`?

Shixiaowei02: > In any case, I don't see exactly what we gain by moving this back to the `devel` stage. What is the issue with leaving this in `tritondevel`?

I hope this dependency can also be provided in the weekly release. However, it is not necessary at the moment. I will move it to `tritondevel`, and if there is a clear user demand, we could consider including NIXL in `devel`.

MartinMarciniszyn: Note that we wouldn't need to disable the installation in the `aarch64` platform if we kept it in the `tritondevel` stage. I still don't see the issue that this creates for you.

jenkins/L0_MergeRequest.groovy
```diff
@@ -21,10 +21,10 @@ UPLOAD_PATH = env.uploadPath ? env.uploadPath : "sw-tensorrt-generic/llm-artifac
...
-LLM_ROCKYLINUX8_PY310_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:cuda-12.9.0-devel-rocky8-x86_64-rocky8-py310-trt10.10.0.31-skip-tritondevel-202505131825-4114"
-LLM_ROCKYLINUX8_PY312_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:cuda-12.9.0-devel-rocky8-x86_64-rocky8-py312-trt10.10.0.31-skip-tritondevel-202505131825-4114"
+LLM_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:pytorch-25.04-py3-x86_64-ubuntu24.04-trt10.10.0.31-skip-tritondevel-202505151430-3934"
+LLM_SBSA_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:pytorch-25.04-py3-aarch64-ubuntu24.04-trt10.10.0.31-skip-tritondevel-202505151430-3934"
+LLM_ROCKYLINUX8_PY310_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:cuda-12.9.0-devel-rocky8-x86_64-rocky8-py310-trt10.10.0.31-skip-tritondevel-202505151430-3934"
+LLM_ROCKYLINUX8_PY312_DOCKER_IMAGE = "urm.nvidia.com/sw-tensorrt-docker/tensorrt-llm:cuda-12.9.0-devel-rocky8-x86_64-rocky8-py312-trt10.10.0.31-skip-tritondevel-202505151430-3934"
```
MartinMarciniszyn, line 27: These image are referenced in at least one more `.groovy` file and a `.yml` file. Did you search and replace all instances?

docker/common/install_nixl.sh
```diff
@@ -5,12 +5,24 @@ GITHUB_URL="https://github.com"
...
 UCX_VERSION="v1.18.1"
 UCX_INSTALL_PATH="/usr/local/ucx/"
+CUDA_PATH="/usr/local/cuda"
 
 if [ ! -d ${UCX_INSTALL_PATH} ]; then
   git clone --depth 1 -b ${UCX_VERSION} https://github.com/openucx/ucx.git
   cd ucx
   ./autogen.sh
-  ./contrib/configure-release --prefix=${UCX_INSTALL_PATH}
+  ./contrib/configure-release       \
```
chuangz0, line 14: Why do we need compile and install ucx in conatiner? There are ucx from HPC in  ngc container, compiling ucx may result in  UCX lacking certain capabilities, such as the ability to use RDMA.

jenkins/L0_MergeRequest.groovy
Shixiaowei02: I have used the script you provided and made some modifications. Could you please review it again? Thank you!

cpp/tensorrt_llm/executor/cache_transmission/nixl_utils/transferAgent.cpp
Shixiaowei02: This part was originally prepared for ETCD. Let's continue to make modifications in the subsequent MRs.

MartinMarciniszyn (approved):

Shixiaowei02:
/bot skip --comment "PreMerge pipelines 4121 and 4154 have passed"

Shixiaowei02:
/bot skip --comment "PreMerge pipelines 4121 and 4154 have passed"

PR #3943, opened 2025-04-29
test: [CI] Add failed cases into waives.txt

xinhe-nv:
waive failed cases.


kaiyux (approved):
Thanks for waiving the cases.

PR #3983, opened 2025-04-30
feat: support to trace executor loop.

yuxianq:
We can trace executor loop with `TLLM_TRACE_EXECUTOR_LOOP` now.
Using `TLLM_TRACE_EXECUTOR_LOOP=<rank>`, we can select to trace which rank; using `TLLM_TRACE_EXECUTOR_LOOP=ALL`, we can trace all ranks. Each trace output will prefix with rank ID.
We also update `TLLM_TRACE_MODEL_FORWARD` to support the same usage.


tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -211,9 +211,6 @@ def estimate_max_kv_cache_tokens(py_executor: PyExecutor,
     py_executor.is_warmup = True
     py_executor.start_worker()
     py_executor.await_responses(req_ids)
-    # TODO check why call mpi_barrier() here will hang-on, but call mpi_allgather(0) is fine.
-    # sync all ranks after processing dummy requests. mpi barrier causes hang, so allgather is used.
-    py_executor.dist.allgather(0)
```
yuxianq, line 216: @byshiue The 0.19 mass integration seems to add this allgather back, I remove it again.

hyukn (approved):

PR #4034, opened 2025-05-02
feat: [TRTLLM-5623][nvbugs/5261055][nvbugs/5170160] non-invasive pipeline parallelism

yuxianq:


tensorrt_llm/mapping.py
```diff
@@ -419,16 +419,13 @@ def has_moe_tp(self):
...
     def pp_layers(self, num_layers: int) -> List[int]:
-        layers_per_pipeline_stage = num_layers // self.pp_size
-        if self.pp_rank == self.pp_size - 1:
-            layers_range = range(self.pp_rank * layers_per_pipeline_stage,
-                                 num_layers)
-        else:
-            layers_range = range(self.pp_rank * layers_per_pipeline_stage,
-                                 (self.pp_rank + 1) * layers_per_pipeline_stage)
-        return list(layers_range)
+        base_layers = num_layers // self.pp_size
```
amukkara, line 423: can you cherry pick an equivalent one-liner from https://github.com/NVIDIA/TensorRT-LLM/pull/3827

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -266,45 +239,44 @@ def forward(
...
+                                    and layer_idx - 1 == pp_layer_list[-1])
+                keep_input_layernorm = (is_next_pp_layer
+                                        and hasattr(layer, "input_layernorm"))
+                if keep_input_layernorm:
+                    input_layernorm = layer.input_layernorm
+                    layer.input_layernorm = None
+                remove_all_weights(layer)
+                if keep_input_layernorm:
+                    layer.input_layernorm = input_layernorm
+            is_first_pp_layer = (not has_pp_layer and is_last_layer) or (
```
amukkara, line 272: why is `(not has_pp_layer and is_last_layer) or ..` case needed in calculation of both `is_first_pp_layer` and `is_last_pp_layer`

tensorrt_llm/_torch/distributed/communicator.py
```diff
@@ -279,3 +279,23 @@ def recv(self,
         if src is None:
             src = self.mapping.prev_pp_rank()
         dist.recv(tensor, src, tag=tag)
+
+
+_pp_comm = None
+_pp_tag = 1234
```
amukkara, line 285: I think the tag is not needed for pp_send/recv, can be removed.

tensorrt_llm/mapping.py
yuxianq: I will rebase this PR after https://github.com/NVIDIA/TensorRT-LLM/pull/3827 merged.

tensorrt_llm/_torch/models/modeling_utils.py
yuxianq: This is for edge cases that `pp_size > num_layers`. In this case, some ranks has not pp layer, but they still need to recv-send hidden state, so we use the last layer to recv-send hidden state. Thus the last rank can still get the correct hidden state for logits processor. For example, if pp_size = 4 and num_layers = 2, then the pipeline is `rank0's layer0 -> rank1's layer1 -> rank2's layer1 (skip forward) -> rank3's layer1 (skip forward) -> rank3's logits processor`.

tensorrt_llm/_torch/distributed/communicator.py
yuxianq: Removed.

tensorrt_llm/_torch/models/modeling_utils.py
amukkara: thanks for the explanation.

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -123,102 +121,107 @@ def unpack_hidden_states(hidden_states):
...
-                       dtype=dtype,
-                       device=torch.cuda.current_device())
-            for key in keys
-        }
-        return PipelineInterface(**tensors)
-
-    return create_pipeline_interface
-
-
 class MissingLayer(torch.nn.Identity):
```
chang-l, line 124: Do we need `MissingLayer` after this refactor?

yuxianq: We don't. Will remove it.

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -266,52 +270,62 @@ def forward(
...
+            layer = self.layers[layer_idx]
+            is_last_layer = (layer_idx == num_hidden_layers - 1)
+            if layer_idx not in pp_layer_list:
+                layer.forward = layer.skip_forward
+                # keep next layer's input_layernorm's weights for fusion
+                is_next_pp_layer = (has_pp_layer
+                                    and layer_idx - 1 == pp_layer_list[-1])
+                keep_input_layernorm = (is_next_pp_layer
+                                        and hasattr(layer, "input_layernorm"))
+                remove_weights(
```
chang-l, line 308: I wonder if you call `create_weights` in the post_init method and then `remove_weights` here, would that pre-allocates and claims oversized device memory in post_init for PP?

tensorrt_llm/_torch/models/modeling_utils.py
chang-l: Ditto for `unpack_hidden_states`

tensorrt_llm/_torch/models/modeling_utils.py
yuxianq: For those weights that need to load, we use `torch.empty` to initialize, which won't trigger GPU memory allocation. Since the weight loading happens after `remove_weights`, these weights will be removed before loading. Thus the GPU memory usage won't increase. If we put `remove_weights` before `create_weights`, some weights initialized in `create_weights` cannot be removed.

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -266,52 +270,62 @@ def forward(
...
+                    if keep_input_layernorm else None,
+                )
+            is_first_pp_layer = (not has_pp_layer and is_last_layer) or (
+                has_pp_layer and layer_idx == pp_layer_list[0])
+            if is_first_pp_layer and not mapping.is_first_pp_rank():
+                layer.forward = forward_after_recv(layer.forward)
+            is_last_pp_layer = (not has_pp_layer and is_last_layer) or (
+                has_pp_layer and layer_idx == pp_layer_list[-1])
+            if is_last_pp_layer and not mapping.is_last_pp_rank():
+                layer.forward = forward_before_send(layer.forward)
```
chang-l, line 320: Thanks @yuxianq for the great effort in extracting the PP logic from the model architecture.

**Nitpick**: I wonder if it would make sense to centralize all the PP injection logic for decoder_layer within a factory impl, e.g.,
```
class PPLayerFactory:
    ...
    def create_pp_layer(layer, layer_idx,...):
        ...
        return pp_layer

# simplified loop
for layer_idx in range(num_hidden_layers):
    self.layers[layer_idx] = PPLayerFactory.create_pp_layer(self.layers[layer_idx], ...)
```
We could also potentially move all PP-related methods/class into a dedicated place, outside of modeling_utils. Maybe this can help future developers manage/extend PP-related logic in one place more easily? 
What do you think?

yuxianq: Good idea, we have already put too many code in tensorrt_llm/_torch/models/modeling_utils.py.
Since this PR is already huge, I will try this refactor in a follow-up PR.

tensorrt_llm/_torch/models/modeling_utils.py
chang-l: Got it, thanks. 
Can you add a brief comment that create/remove_weights do not trigger any memory allocation/deallocation?

tensorrt_llm/_torch/models/modeling_utils.py
yuxianq: Both updated.

tensorrt_llm/_torch/models/modeling_utils.py
amukkara: sorry for the late comment. we might be adding unnecessary code complexity to support the rare case of `pp_size > num_layers` which means some GPUs are not doing any useful work.

Is there specific use-case where this scenario arises that I am missing? if not, we can just add an `assert pp_size <= num_layers`. cc @chang-l

yuxianq: Mainly for compatibility. I don't have specific use cases now.

amukkara: in that case, I prefer we enforce `pp_size <= num_layers`.

While I appreciate the effort to support this edge case, it adds extra code and more importantly extra work in all future PP related changes to continue supporting this case.

tensorrt_llm/_torch/models/modeling_utils.py
yuxianq: Added.

tensorrt_llm/_torch/models/modeling_utils.py
yuxianq: Have enforced `pp_size <= num_layers` now.

amukkara (approved):

yuxianq:
The failed test `accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales_4gpus[pp4--attention_dp-cuda_graph-overlap_scheduler]` can be fixed by https://github.com/NVIDIA/TensorRT-LLM/pull/3878. Will wait for it to pass CI.

yuxianq:
/bot skip --comment "The failed stage DGX_H100-4_GPUs-TensorRT-1 is unrelated to this PR. The failed tests accuracy/test_llm_api_pytorch.py::TestLlama3_1_8BInstruct::test_chunked_prefill[attn_backend=TRTLLM] and accuracy/test_llm_api_pytorch.py::TestLlama3_1_8BInstruct::test_chunked_prefill[attn_backend=FLASHINFER] in DGX_H100-4_GPUs-PyTorch-Others-1 have failed before this PR. All other tests have passed. Should be safe to merge this PR."

yuxianq:
/bot skip --comment "The failed stage DGX_H100-4_GPUs-TensorRT-1 is unrelated to this PR. The failed tests accuracy/test_llm_api_pytorch.py::TestLlama3_1_8BInstruct::test_chunked_prefill[attn_backend=TRTLLM] and accuracy/test_llm_api_pytorch.py::TestLlama3_1_8BInstruct::test_chunked_prefill[attn_backend=FLASHINFER] in DGX_H100-4_GPUs-PyTorch-Others-1 have failed before this PR. All other tests have passed. Should be safe to merge this PR."

PR #4041, opened 2025-05-03
fix: [nvbug/5247414] Fix the perf regression cause by insufficient cache warmup.

hyukn:
Force tuning up to 8192 sequence length for NVFP4 linear op. Also, make this runtime-selectable with UB enabled.


juney-nvidia:
@hyukn  Thanks for the prompt fix, Yukun. Usually for release oriented fix, we will merge the fix directly onto the release branch, then it will be mass integrated into the main branch.

June

hyukn:
> @hyukn Thanks for the prompt fix, Yukun. Usually for release oriented fix, we will merge the fix directly onto the release branch, then it will be mass integrated into the main branch.
> 
> June

Sure. I will redirect this to release/0.19.

hyukn:
Close this PR to main and put it in release/0.19 in the first place.

PR #4042, opened 2025-05-03
fix: [nvbug/5247414] Fix the perf regression caused by insufficient cache warmup.

hyukn:
Force tuning up to 8192 sequence length for NVFP4 linear op. Also, make this runtime-selectable with UB enabled.


juney-nvidia (approved):

hyukn:
I just found that the lambda-style fix might cause some side effects when considering `to_userbuffer`. Will make another one.

hyukn:
> I just found that the lambda-style fix might cause some side effects when considering `to_userbuffer`. Will make another one.

NVM. I have double-checked with benchmark perf evaluation locally. This is not an issue.
Will continue on further optimization of the host code of AutoTuner.

PR #4060, opened 2025-05-05
[nvbugs/5247414][fix] Optimize the AutoTuner cache access code to reduce host code overhead.

hyukn:
The NVFP4 Linear op is very sensitive to the host overhead.
This PR introduces customizable `find_nearest_profile` and `get_cache_key_specifc`, which allow users to override the default method for generating the cache key.
For a local Llama3.3-70B benchmark, this can improve the perf by about 1.6%.


zbpatel:
Thanks @hyukn nice work

juney-nvidia (approved):

PR #4067, opened 2025-05-05
feat: Reduce branch overhead in groupRMSNorm kernels

SimengLiu-nv:
# [feat] Reduce branch overhead in groupRMSNorm kernels

## Description
Add template variable MultiRounds to avoid unnecessary branches when the last dimension of inputs can be covered in one round. MultiRounds will be true when last dimension of inputs > 1024 * sizeof(float4) / sizeof(dtype), which is 9182 for fp16 or bf16.

### DIFF
--- a/cpp/tensorrt_llm/kernels/groupRmsNormKernels/groupRmsNormKernels.cu
+++ b/cpp/tensorrt_llm/kernels/groupRmsNormKernels/groupRmsNormKernels.cu
@@ -26,7 +26,7 @@
 namespace tensorrt_llm::kernels::group_rms_norm
 {
 // Allocate more warps to deal with the second input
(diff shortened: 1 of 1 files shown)


SimengLiu-nv:
The two CI/CD failures are kinda weird, I added prints and discovered the "smem_input_mask[warp_idx] = input_idx" was not executed on the second warp with input_last_dims=[256, 512]. The default value was picked after the `__syncthreads` call. The failure happens on A30 (sm_80) and L40 (sm_89) machines. Investigating.

Kefeng-Duan (approved):

SimengLiu-nv:
/bot skip --comment "Pipeline passes for related tests. The failure is on a known CI issue."

PR #4070, opened 2025-05-05
fix: Update log query regex in perf integration test to match trtllm-bench reporting

venkywonka:
# PR title

Update log query regex in perf integration test to match trtllm-bench reporting.
I believe in this PR https://github.com/NVIDIA/TensorRT-LLM/pull/3923/files the change includes the behavior of reporting to print "tps/gpu" instead of "tokens/sec/gpu" for token throughput, but in perf integration tests, we still regex match for the latter - hence causing failures locally

### DIFF
--- a/tests/integration/defs/perf/test_perf.py
+++ b/tests/integration/defs/perf/test_perf.py
@@ -178,7 +178,7 @@ BENCH_PERF_METRIC_LOG_QUERIES = {
     PerfMetricType.INFERENCE_TIME:
     re.compile(r"Total Latency \(ms\):\s+([\d\.]+)"),
     PerfMetricType.TOKEN_THROUGHPUT:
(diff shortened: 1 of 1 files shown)


FrankD412 (approved):

FrankD412:
At some point I think these tests might want to move to the JSON output that `trtllm-bench` is able to produce. Those keys are less likely to change because we also have our own data collection that relies on those keys.

venkywonka:
Created an issue https://github.com/NVIDIA/TensorRT-LLM/issues/4071 for this cc @LarryXFly @ruodil

venkywonka:
There's a related PR: https://github.com/NVIDIA/TensorRT-LLM/pull/3531 by @amirkl94 related to improving trtllm-bench integration with perf tests.

venkywonka:
closing this, as its already a part of https://github.com/NVIDIA/TensorRT-LLM/pull/3531

PR #4076, opened 2025-05-06
Reduce computation for the top layer in chunked context.

ming-wei:
Only the last logit is required in context phase. Thus, we skip SDPA computation for all chunks (except the last one) for the top layer.

Though this PR only changes the behavior for llama models, this can be applied on any decoder-only models too.


venkywonka:
@ming-wei , does this PR need action by reviewer, or can it be closed? looks like its stale? Thanks!

ming-wei:
Hi,

This PR is a stale one. Please feel free to close it.

Thanks.

PR #4078, opened 2025-05-06
refactor: Allow models to override apply_qk_norm.

yuxianq:


tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -52,19 +51,13 @@ def __init__(
         aux_stream: Optional[torch.cuda.Stream] = None,
     ):
         config = model_config.pretrained_config
-        self.aux_stream = aux_stream
-        self.ln_events = [torch.cuda.Event(), torch.cuda.Event()]
 
         self.use_rope = not nope_layer
-        self.use_qk_norm = use_qk_norm and not nope_layer
```
hlu1, line 59: Why is this line removed?

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -191,28 +186,8 @@ def forward(
...
-                else:
-                    q_by_head = q.reshape(-1, self.head_dim)
-                    q_by_head = self.q_norm(q_by_head)
-                    q = q_by_head.view(q.shape)
-                    k_by_head = k.reshape(-1, self.head_dim)
-                    k_by_head = self.k_norm(k_by_head)
-                    k = k_by_head.view(k.shape)
-
+            if self.use_qk_norm:
+                q, k = self.apply_qk_norm(q, k)
```
hlu1, line 190: Here you apply `qk_norm` first and then `apply_rotary`, which is for QWen3, I suppose. But if you look at the llama4 code, the order is the opposite. You do `apply_rotary` first, and then `qk_norm`.

I only noticed this while reviewing the PR today. Maybe it's not the right design to try to implement `qk_norm` generically.

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -154,31 +115,6 @@ def _get_attn_scale(position_ids: torch.Tensor) -> torch.Tensor:
...
-        mrope_config: Optional[dict] = None,
-        all_reduce_params: Optional[AllReduceParams] = None,
-    ):
-        if self.use_qk_norm:
-            qkv = self.qkv_proj(hidden_states)
-            q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size],
-                                dim=-1)
-            assert self.rotary_emb is not None and not self.enable_rope_fusion, "qk_norm requires attention rope fusion disabled"
-            q, k = self.rotary_emb(position_ids, [q, k])
-            q, k = self._qk_norm(q, k)
```
hlu1, line 173: The order is important here.

tensorrt_llm/_torch/models/modeling_llama.py
yuxianq: It is unnecessary, we will set `self.use_qk_norm` in `Attention.__init__`. We won't use qk_norm in the `_forward_nope` anyway, so we don't need to skip qk_norm according to `nope_layer`.

tensorrt_llm/_torch/models/modeling_llama.py
yuxianq: Fixed by using `QkNormType.post_rope` for llama4.

byshiue (approved):
Looks good on qwen3 side.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -115,13 +121,9 @@ def __init__(
...
                                 [self.hidden_size])
 
-        self.use_qk_norm = (
-            config.pretrained_config
-            and (config.pretrained_config.model_type == 'qwen3'
-                 or config.pretrained_config.model_type == 'qwen3_moe'))
         attn_cls = get_attention_backend(self.attn_backend)
         self.enable_rope_fusion = attn_cls.support_fused_rope(
-        ) and not self.use_qk_norm
+        ) and qk_norm_type == QkNormType.none
```
hlu1, line 126: In the case of norm/rotary (not rotary/norm), we can still fuse rotary and attention, which is exactly the case in MLA

yuxianq: Updated. We fuse rope in the pre-rope qk norm case now.

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -52,86 +51,51 @@ def __init__(
...
-                                        v,
-                                        attn_metadata,
-                                        out_scale=out_scale,
-                                        attention_mask=attention_mask,
-                                        mrope_config=mrope_config)
-
-        attn_output = self.o_proj(attn_output,
-                                  all_reduce_params=all_reduce_params)
+    def apply_qk_norm(self, q, k):
+        # TODO: make this more efficient.
```
chang-l, line 90: Can we update this comment here or just remove it? It is ambiguous and provides no information.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -238,6 +226,11 @@ def forward(
 
         return attn_output
 
+    def apply_qk_norm(self, q, k):
+        raise NotImplementedError(
```
chang-l, line 230: Why we need to implement `apply_qk_norm` for each derived attn class (`llama4` or `qwen3`)?  Are there different qk_norm implementations?

tensorrt_llm/_torch/models/modeling_llama.py
yuxianq: Not added by me. I can remove it.

tensorrt_llm/_torch/modules/attention.py
yuxianq: Yes, llama4 uses the same qk_norm for both q and k, but qwen3 uses q_norm for q and k_norm for k.

tensorrt_llm/_torch/models/modeling_llama.py
yuxianq: Removed.

tensorrt_llm/_torch/modules/attention.py
chang-l: Sigh.. I see. Thanks for your explanation.
In this way, I agree we should treat `qk_norm` like `rotary_emb`, or even combine them? Maybe we can create a qk_norm module (or merge it into rope module) for all qk_norm impls. I think it may avoid qk_norm impls scattering around and ease future fusion effort. What do you think?  I know it's beyond the scope of this PR, so feel free to address it later


However, it might be

chang-l (approved):
LGTM

tensorrt_llm/_torch/modules/attention.py
hlu1: I don't think we should combine the qk_norm and rope. They are separate concepts and should stay that way.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -188,32 +194,14 @@ def forward(
...
-                    )
-                else:
-                    q_by_head = q.reshape(-1, self.head_dim)
-                    q_by_head = self.q_norm(q_by_head)
-                    q = q_by_head.view(q.shape)
-                    k_by_head = k.reshape(-1, self.head_dim)
-                    k_by_head = self.k_norm(k_by_head)
-                    k = k_by_head.view(k.shape)
-
+            q, k, v = self.split_qkv(qkv)
```
hlu1, line 201: Is this line conflicting with line 198 when both if conditions are true, i.e., FlashInder?

yuxianq: Yes, so fragile. I extend `split_qkv` to check whether q/k/v have been split and return q/k/v as is in this case. Now we can call `split_qkv` multiple times and it will only split once.

tensorrt_llm/_torch/modules/attention.py
hlu1: No, we should not do more fusions inside AttentionOp. The right direction is to make it more modular and flexible. Even rotary fusion itself already gives us a lot of trouble.

hlu1 (approved):

hlu1:
We should add more attention unit tests to test these features. Relying on e2e accuracy tests is too heavy and it takes time to debug the accuracy issues.

dongjiyingdjy:
Looks good on qwen3.

dongjiyingdjy (approved):

PR #4084, opened 2025-05-06
[fix] Fix add_dummy_requests for spec decoding cases

lfr-0531:
## Description

- Add kv space for those extra kv tokens and draft tokens
- Move the `token_num` logics into `add_dummy_requests` to make the code clean

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/resource_manager.py
+++ b/tensorrt_llm/_torch/pyexecutor/resource_manager.py
@@ -302,8 +302,6 @@ class KVCacheManager(BaseResourceManager):
                 req.state = LlmRequestState.GENERATION_IN_PROGRESS
                 req.prompt_len = token_num
                 req.py_prompt_len = req.prompt_len
-                req.add_new_token(1, 0)
-                self.impl.add_token(req_id)
                 if max_num_draft_tokens > 0:
                     req.py_draft_tokens = [1] * max_num_draft_tokens
                     for _ in range(max_num_draft_tokens):
--- a/tensorrt_llm/_torch/pyexecutor/resource_manager.py
+++ b/tensorrt_llm/_torch/pyexecutor/resource_manager.py
@@ -304,8 +304,9 @@ class KVCacheManager(BaseResourceManager):
                 req.py_prompt_len = req.prompt_len
                 if max_num_draft_tokens > 0:
                     req.py_draft_tokens = [1] * max_num_draft_tokens
-                    for _ in range(max_num_draft_tokens):
-                        self.impl.add_token(req_id)
+                    if prepare_resource:
+                        for _ in range(max_num_draft_tokens):
+                            self.impl.add_token(req_id)
             requests.append(req)
         return requests
 
--- a/tensorrt_llm/_torch/pyexecutor/resource_manager.py
+++ b/tensorrt_llm/_torch/pyexecutor/resource_manager.py
@@ -281,8 +281,7 @@ class KVCacheManager(BaseResourceManager):
         requests = []
         for i, req_id in enumerate(request_ids):
             sampling_params = SamplingParams()
-            token_num = token_nums[
-                i] if token_nums is not None else 1 + max_num_draft_tokens
+            token_num = token_nums[i] if token_nums is not None else 1
             encoder_input_tokens = [
                 1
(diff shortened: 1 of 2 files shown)


tests/unittest/_torch/speculative/test_eagle3.py
```diff
@@ -41,10 +41,16 @@ def test_llama_eagle3(use_cuda_graph: bool, attn_backend: str):
     spec_config = EagleDecodingConfig(
         max_draft_len=draft_len, pytorch_eagle_weights_path=eagle_model_dir)
 
+    # 2048 is the max_position_embeddings of the draft model
+    build_config = None
+    if attn_backend == "FLASHINFER":
+        build_config = BuildConfig(max_seq_len=2048)
+
```
mikeiovine, line 48: Why only for flashinfer?

mikeiovine (approved):

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -280,8 +280,7 @@ def add_dummy_requests(
         requests = []
         for i, req_id in enumerate(request_ids):
             sampling_params = SamplingParams()
-            token_num = token_nums[
-                i] if token_nums is not None else 1 + max_num_draft_tokens
+            token_num = token_nums[i] if token_nums is not None else 1
```
zhhuang-nv, line 283: It would be better if we add some docs for the parameters of this function, e.g., token_num should be past_kv_len + current_input_len_without_spec_decoding, this function will handle spec_decoding internally, etc.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -296,12 +295,16 @@ def add_dummy_requests(
             req.paged_kv_block_ids = []
             if prepare_resource:
                 self.impl.add_sequence(req_id, token_num, beam_width, req)
+                for _ in range(self.num_extra_kv_tokens):
+                    self.impl.add_token(req_id)
             if is_gen:
                 req.state = LlmRequestState.GENERATION_IN_PROGRESS
-                req.prompt_len = token_num - 1 + max_num_draft_tokens
+                req.prompt_len = token_num - 1
```
zhhuang-nv, line 302: We should not minus 1 for generation request as we discussed yesterday. `req.prompt_len` should be `token_num`.

lfr-0531: The `token_num = past_kv_len + current_input_len_without_spec_decoding`. For the generation requests, the `current_input_len_without_spec_decoding =1`. So we have `token_num = past_kv_len + 1`. But the `prompt_len` is not equal to the `token_num` here, and it should be the input seq_len in context phase. 

Here for a generation dummy request, we assume it is the 1st generation iteration, so the last iteration is the context phase, and the prompt length should be `token_num - 1`, right?

zhhuang-nv: Maybe we should align with some concepts first. What's the meaning of `prompt_len` for generation requests? In yesterday discussion, we think it should be always `1` and the `seq_len` for generation requests should be `1 + max_num_draft_tokens`, no matter which generation iteration they are, is that right?  I think we should keep `prompt_len` aligned within real requests and dummy requests. If that makes sense, maybe we should always have `req.prompt_len = 1` for dummy generation requests.

Anyway, the codes for now have some problems. `req.prompt_len` will be `0` if `token_num` not provided.

We can discuss it offline.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
lfr-0531: Added.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
zhhuang-nv: I got it. `prompt_len` is original input prompt len and it is unchanged during iterations typically. It may makes sense when we want to build a dummy generation requests with `token_num = 1` which doesn't have past_kv, i.e., `prompt_len = 0`. Let's keep this logic and fix other CI issues. Thanks.

PR #4133, opened 2025-05-07
[TRTLLM-5233][feat]: Add chunking to PyT heuristic for trtllm-bench.

FrankD412:
## Description

This MR adds chunking to the Pytorch flow in `trtllm-bench` -- when testing with some models it turns out that the tuning heuristic does have cases where the max tokens it recommends is less than the maximum ISL we will observe in the dataset. We also force the chunking to off in the Pytorch flow since it previously wasn't enabled. How the heuristic works now:

- Currently chunking is still off by default considering the benchmark is used for both low latency and maximum throughput
- If chunking is off and we detect that the heuristic has returned a max batch size smaller than the max ISL, we force max tokens up to the max ISL out of necessity.
- If chunking is enabled, we simply set it and set the heuristic based max tokens.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


tensorrt_llm/bench/benchmark/utils/general.py
```diff
@@ -133,6 +138,13 @@ def get_settings(params: dict, dataset_metadata: DatasetMetadata, model: str,
             f"Using heuristics or pre-defined settings: max_batch_size={max_batch_size}, max_num_tokens={max_num_tokens}."
         )
 
+        # If chunked prefill is disabled, we need to ensure that the max_num_tokens is at least the max_isl
+        if not enable_chunked_prefill and max_num_tokens < dataset_metadata.max_isl:
+            logger.warning(
```
hchings, line 143: nit: might be better to use `logger.info` when we implicitly change user configs, as other places in this function. Not sure whether by default warning will show up w/o changing log level. Otherwise MR LGTM.

hchings (approved):

kaiyux (approved):

tensorrt_llm/bench/benchmark/utils/general.py
kaiyux: I think slightly differently - we need to use `warning` level of logging to warn users that we are implicitly change their configs, does that make sense?

`--log_level` is set to `info` by default, so `warning` will show up by default as well.

FrankD412: Sorry meant to respond sooner and forgot -- yeah, I wanted `warning` since the priority of a warning is higher than `info`. I also didn't think it belonged in `info` as Kaiyu said since we are implicitly overriding something.

FrankD412:
Something that came to mind was if we wanted to expose a knob for chunking at this point. @kaiyux thoughts on that?

kaiyux:
> Something that came to mind was if we wanted to expose a knob for chunking at this point. @kaiyux thoughts on that?

@FrankD412 IIUC It should've already be able to be controlled through `--extra_llm_api_options`. We should indeed verify that can actually control that though.

FrankD412:
Will do -- I think it will. I was thinking more along the lines of a dedicated option since some folks might want to perform studies on it.

kaiyux:
> Will do -- I think it will. I was thinking more along the lines of a dedicated option since some folks might want to perform studies on it.

IMO we should really carefully review it since the initial plan was to keep to number of knobs exposed to the users minimal, I don't think we should often make some exceptions. If the extra yaml file way is not straightforward, we should consider move all parameters to cli, or, make all parameters available both in yaml file and in cli.

FrankD412:
Sounds good -- I was wondering since the throughput benchmark is now used for low latency so it would be an easier way to toggle it.

FrankD412:
I was looking at this in the context of some of our sweeps -- I think it might be beneficial to have it be on the CLI because it's very difficult to automate around it if it's in the YAML configuration. We still haven't performed the needed studies to enable it by default, but having the CLI option to see how it behaves with the rest of the tuning would be a benefit. What do you think @kaiyux?

Superjomn (approved):
LGTM

PR #4159, opened 2025-05-08
[nvbugs/5268808][fix] Fix the potential out-of-range-access issue of allreduce workspace.

hyukn:
This issue is only found for tp=ep=8 on multi-node machine. This could be caused by the race between different tp that triggers the illegal list access.

### DIFF
--- a/tensorrt_llm/_torch/distributed/ops.py
+++ b/tensorrt_llm/_torch/distributed/ops.py
@@ -12,10 +12,11 @@ _thread_local = threading.local()
 
 
 def get_allreduce_workspace(mapping: Mapping) -> torch.LongTensor:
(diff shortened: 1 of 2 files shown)


qiaoxj07:
@hyukn  I tried this fix, there will be illegal memory access error on GB200 TEP8 after apply the patch, could you have a look?

hyukn:
> @hyukn I tried this fix, there will be illegal memory access error on GB200 TEP8 after apply the patch, could you have a look?

Hi Xianjie. You mean if testing it without multi-node, the test for deepseek on single node will trigger illegal memory access issue?

yuxianq (approved):

qiaoxj07 (approved):

PR #4180, opened 2025-05-09
add changes for fp8, nemotron-nas, API

shaharmor98:
# PR title

## Description

This PR includes the changes required for feature-complete support of LoRA in PyTorch flow.
It includes:
* API changes
* FP8 test
* Nemotron-NAS dummy support

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/model_config.py
+++ b/tensorrt_llm/_torch/model_config.py
@@ -191,7 +191,27 @@ class ModelConfig(Generic[TConfig]):
             data_type=torch_dtype_to_binding(
                 self.pretrained_config.torch_dtype))
 
-        mlp_hidden_size = self.pretrained_config.intermediate_size // self.mapping.tp_size
+        mlp_hidden_size = None
+        if self.pretrained_config.intermediate_size is not None:
+            mlp_hidden_size = self.pretrained_config.intermediate_size // self.mapping.tp_size
+        else:
+            # TODO: once tensorrt_llm._torch.AutoConfig is implemented, the following logic
+            # should be moved to tensorrt_llm._torch.AutoConfig of the relevant modeling_xxx file
+            if hasattr(self.pretrained_config, "architectures"
+                       ) and self.pretrained_config.architectures is not None:
+                architectures = self.pretrained_config.architectures
+                if len(architectures
+                       ) == 1 and architectures[0] == "DeciLMForCausalLM":
+                    mlp_hidden_size = self._infer_nemotron_ffn_mult()
+                else:
+                    raise ValueError(
+                        f"Inferring mlp hidden size for model architecture: {architectures} isn't supported yet"
+                    )
+        if mlp_hidden_size is None:
+            raise ValueError(
+                f"Failed to infer mlp hidden size for model: {self.pretrained_config.model_type}"
+            )
+
         if "head_size" in self.pretrained_config:
             head_size = self.pretrained_config.head_size
         else:
@@ -201,3 +221,18 @@ class ModelConfig(Generic[TConfig]):
         model_config_cpp.size_per_head = head_size
 
         return model_config_cpp
+
+    def _infer_nemotron_ffn_mult(self):
+        # TODO smor: this is a hack to support Nemotron-Super-49B-v1 with LoRA, tracked by TRTLLM-5045 ticket
+        # Nemotron-NAS has variable ffn_mult for each layer, we need to find the maximum
+        # so that we don't set a too small mlp_hidden_size. This solution leads to a memory
+        # consumption that is higher than required.
+        biggest_ffn_mult = max(
+            [x.ffn.ffn_mult for x in self.pretrained_config.block_configs])
+
+        from tensorrt_llm._torch.models.modeling_nemotron_nas import \
+            _ffn_mult_to_intermediate_size
+        mlp_hidden_size = _ffn_mult_to_intermediate_size(
+            biggest_ffn_mult, self.pretrained_config.hidden_size)
+
+        return mlp_hidden_size
--- a/tensorrt_llm/_torch/models/modeling_llama.py
+++ b/tensorrt_llm/_torch/models/modeling_llama.py
@@ -12,6 +12,7 @@ from transformers.models.llama4.modeling_llama4 import Llama4MultiModalProjector
 from tensorrt_llm._torch.distributed import (AllReduce, AllReduceFusionOp,
                                              AllReduceParams, MoEAllReduce)
 from tensorrt_llm.functional import PositionEmbeddingType
+from tensorrt_llm.lora_manager import HfLoraLoader
 from tensorrt_llm.models.convert_utils import split_matrix_tp
 
 from ...inputs import (ExtraProcessedInputs, InputProcessor, TextPrompt,
@@ -724,14 +725,16 @@ class LlamaModel(DecoderModel):
 
         vocab_size = config.vocab_size
         # TODO smor- we load manually only if there is a single lora dir, need to come up with a better solution
+        self.has_custom_embed_tokens = False
         if hasattr(
                 model_config,
                 'lora_config') and model_config.lora_config is not None and len(
                     model_config.lora_config.lora_dir) == 1:
-            from tensorrt_llm.lora_manager import HfLoraLoader
             lora_loader = HfLoraLoader(model_config.lora_config.lora_dir)
-            weight = lora_loader.embed_tokens
-            vocab_size = lora_loader.vocab_size
+            if lora_loader.vocab_size != 0 and lora_loader.embed_tokens is not None:
+                vocab_size = lora_loader.vocab_size
+                weight = lora_loader.embed_tokens
+                self.has_custom_embed_tokens = True
 
         self.embed_tokens = Embedding(
             vocab_size,
@@ -742,10 +745,7 @@ class LlamaModel(DecoderModel):
             gather_output=True,
         )
 
-        if hasattr(
-                model_config,
-                'lora_config') and model_config.lora_config is not None and len(
-                    model_config.lora_config.lora_dir) == 1:
+        if self.has_custom_embed_tokens:
             with torch.no_grad():
                 if model_config.mapping.tp_size > 1:
                     weight = split_matrix_tp(
--- a/tensorrt_llm/_torch/models/modeling_nemotron_nas.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_nas.py
@@ -1,10 +1,12 @@
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import torch
 from torch import nn
 from transformers import PretrainedConfig
 
 from tensorrt_llm.functional import PositionEmbeddingType
+from tensorrt_llm.lora_manager import HfLoraLoader
+from tensorrt_llm.models.convert_utils import split_matrix_tp
 
 from ..attention_backend import AttentionMetadata
 from ..attention_backend.interface import PositionalEmbeddingParams, RopeParams
@@ -75,8 +77,12 @@ class LinearAttention(nn.Module):
     def forward(
         self,
         hidden_states: torch.Tensor,
+        lora_params: Optional[dict] = None,
         **kwargs: Any,
     ) -> torch.Tensor:
+        if lora_params is not None:
+            raise NotImplementedError(
+                "LinearAttention with LoRA is not supported yet")
         return self.linear_attn(hidden_states)
 
 
@@ -87,7 +93,12 @@ class LinearMLP(nn.Module):
         super().__init__()
         self.linear_mlp = _create_linear_from_configs(model_config, config)
 
-    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+    def forward(self,
+                hidden_states: torch.Tensor,
+                lora_params: Optional[dict] = None) -> torch.Tensor:
+        if lora_params is not None:
+            raise NotImplementedError(
+                "LinearMLP with LoRA is not supported yet")
         return self.linear_mlp(hidden_states)
 
 
@@ -121,7 +132,8 @@ class NemotronNASDecoderLayer(DecoderLayer):
                         self.block_config.ffn.ffn_mult, config.hidden_size),
                     bias=False,
                     dtype=config.torch_dtype,
-                    config=model_config)
+                    config=model_config,
+                    layer_idx=layer_idx)
 
     def forward(
         self,
@@ -147,7 +159,7 @@ class NemotronNASDecoderLayer(DecoderLayer):
             # Fully Connected
             residual = hidden_states
             hidden_states = self.post_attention_layernorm(hidden_states)
-            hidden_states = self.mlp(hidden_states)
+            hidden_states = self.mlp(hidden_states, **kwargs)
             hidden_states = residual + hidden_states
 
         return hidden_states
@@ -165,11 +177,35 @@ class NemotronNASModel(DecoderModel):
             for block in config.block_configs
         ]
 
+        vocab_size = config.vocab_size
+        self.has_custom_embed_tokens = False
+        if hasattr(
+                model_config,
+                'lora_config') and model_config.lora_config is not None and len(
+                    model_config.lora_config.lora_dir) == 1:
+            lora_loader = HfLoraLoader(model_config.lora_config.lora_dir)
+            if lora_loader.vocab_size != 0 and lora_loader.embed_tokens is not None:
+                vocab_size = lora_loader.vocab_size
+                weight = lora_loader.embed_tokens
+                self.has_custom_embed_tokens = True
+
         self.embed_tokens = Embedding(
-            config.vocab_size,
+            vocab_size,
             config.hidden_size,
             dtype=config.torch_dtype,
         )
+
+        if self.has_custom_embed_tokens:
+            with torch.no_grad():
+                if model_config.mapping.tp_size > 1:
+                    weight = split_matrix_tp(
+                        weight,
+                        model_config.mapping.tp_size,
+                        model_config.mapping.tp_rank,
+                        dim=0)  # split by vocabulary dimension
+                x = weight.to(self.embed_tokens.dtype)
+                self.embed_tokens.weight.data.copy_(x)
+
         self.layers = nn.ModuleList([
             NemotronNASDecoderLayer(model_config, block_config, layer_idx)
             for layer_idx, block_config in enumerate(config.block_configs)
--- a/tensorrt_llm/_torch/models/modeling_utils.py
+++ b/tensorrt_llm/_torch/models/modeling_utils.py
@@ -10,6 +10,7 @@ from torch.utils._python_dispatch import TorchDispatchMode
 from torch.utils._pytree import tree_any_only
 from tqdm import tqdm
 
+from tensorrt_llm.lora_manager import HfLoraLoader
 from tensorrt_llm.mapping import Mapping
 from tensorrt_llm.models.convert_utils import split_matrix_tp
 
@@ -338,6 +339,7 @@ class DecoderModelForCausalLM(nn.Module,
         self.model = model
         self.pp_rank = config.mapping.pp_rank
         self.pp_size = config.mapping.pp_size
+        self.has_custom_lm_head = False
 
         if config.mapping.enable_attention_dp:
             self.lm_head = LMHead(
@@ -355,14 +357,14 @@ class DecoderModelForCausalLM(nn.Module,
         else:
             # TODO(zhenhuanc): Currently lm_head Linear will not accept QuantConfig
             # will considering per layer QuantConfig in the future.
-
             if (hasattr(config, 'lora_config')
                     and config.lora_config is not None
                     and len(config.lora_config.lora_dir) == 1):
-                from tensorrt_llm.lora_manager import HfLoraLoader
                 lora_loader = HfLoraLoader(config.lora_config.lora_dir)
-                weight = lora_loader.lm_head
-                vocab_size = lora_loader.vocab_size
+                if lora_loader.lm_head is not None and lora_loader.vocab_size != 0:
+                    weight = lora_loader.lm_head
+                    self.has_custom_lm_head = True
+                    vocab_size = lora_loader.vocab_size
 
             self.lm_head = LMHead(
                 vocab_size,
@@ -373,9 +375,7 @@ class DecoderModelForCausalLM(nn.Module,
                 gather_output=True,
             )
 
-            if (hasattr(config, 'lora_config')
-                    and config.lora_config is not None
-                    and len(config.lora_config.lora_dir) == 1):
+            if self.has_custom_lm_head:
                 with torch.no_grad():
                     if config.mapping.tp_size > 1:
                         weight = split_matrix_tp(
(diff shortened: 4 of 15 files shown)


tests/unittest/llmapi/test_llm_pytorch.py
```diff
@@ -171,6 +180,131 @@ def test_llama_v2_13b_lora():
     llama_v2_13b_lora_test_harness()
 
 
+@skip_gpu_memory_less_than_40gb
```
byshiue, line 183: Can you explain what features do we test in these three CI tests? 

Besides, is it possible to fuse these tests into one to reduce the CI overhead? 

And I don't see you add the CI test in the test list, where could I find them?

tests/unittest/llmapi/test_llm_pytorch.py
```diff
@@ -171,6 +180,131 @@ def test_llama_v2_13b_lora():
...
+
+    assert similar(outputs[0].outputs[0].text, references[0])
+
+
 @skip_gpu_memory_less_than_40gb
 def test_llama_7b_multi_lora():
     llama_7b_multi_lora_test_harness()
+
+
+# TODO smor: currently Nemotron-Super-49B-v1 with LoRA memory consumption is overly high
```
byshiue, line 214: Do you plan to verify this test on which GPU? Is it possible to reduce the memory usage by using FP8 ckpt?

tensorrt_llm/lora_manager.py
```diff
@@ -147,6 +147,8 @@ class LoraConfig(DictConversion):
     max_lora_rank: int = 64
     lora_target_modules: List[str] = field(default_factory=list)
     trtllm_modules_to_hf_modules: Dict[str, str] = field(default_factory=dict)
+    max_loras: int = 2
```
byshiue, line 150: Why these two parameters are set to 2 by default instead of 1?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1219,11 +1227,41 @@ def _setup(self):
...
-                "Lora is an experimental feature and is probably subject to change before 1.0 release"
+                "enable_lora is ignored when lora_config is provided for pytorch backend."
             )
 
+        if self.lora_config is not None:
+            if len(self.lora_config.lora_dir) == 0 and len(
+                    self.lora_config.lora_target_modules) == 0:
+                logger.warning(
+                    "Both lora_dir and lora_target_modules are empty, so all LoRA modules will be expected. "
+                    "This will lead to serious memory consumption. Please provide either lora_dir or lora_target_modules."
```
byshiue, line 1258: Should be better to say "Please provide either lora_dir or lora_target_modules if this behavior is not what you expect.`

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -665,13 +667,18 @@ def filter_weights(prefix, weights: Dict):
...
+            # Skip loading weights for embedding and lm_head if LoRA is enabled and has custom values
             if hasattr(model.model_config, 'lora_config'
                        ) and model.model_config.lora_config is not None and len(
-                           model.model_config.lora_config.lora_dir) == 1 and (
-                               name == "model.embed_tokens"
-                               or name == "lm_head"):
-                continue
+                           model.model_config.lora_config.lora_dir) == 1:
+                # TODO smor- this shouldn't be here, find a better way to extract the lora_loader
+                from tensorrt_llm.lora_manager import HfLoraLoader
```
byshiue, line 675: Should move to the beginning.

tensorrt_llm/_torch/models/modeling_nemotron_nas.py
```diff
@@ -165,11 +176,37 @@ def __init__(self, model_config):
             for block in config.block_configs
         ]
 
+        vocab_size = config.vocab_size
+        has_custom_embed_tokens = False
+        if hasattr(
+                model_config,
+                'lora_config') and model_config.lora_config is not None and len(
+                    model_config.lora_config.lora_dir) == 1:
+            from tensorrt_llm.lora_manager import HfLoraLoader
```
byshiue, line 185: Should move to the beginning.

tensorrt_llm/_torch/models/modeling_nemotron_nas.py
```diff
@@ -165,11 +176,37 @@ def __init__(self, model_config):
...
+            lora_loader = HfLoraLoader(model_config.lora_config.lora_dir)
+            if lora_loader.vocab_size != 0:
```
byshiue, line 187: Similar issue.

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -360,13 +360,17 @@ def __init__(self, model: TModel, *, config: ModelConfig[TConfig],
...
 
+                has_custom_lm_head = False
                 if hasattr(config, 'lora_config'
                            ) and config.lora_config is not None and len(
                                config.lora_config.lora_dir) == 1:
                     from tensorrt_llm.lora_manager import HfLoraLoader
                     lora_loader = HfLoraLoader(config.lora_config.lora_dir)
-                    weight = lora_loader.lm_head
-                    vocab_size = lora_loader.vocab_size
+                    if lora_loader.lm_head is not None:
```
byshiue, line 369: similar issue.

tensorrt_llm/_torch/model_config.py
```diff
@@ -191,7 +191,34 @@ def get_bindings_model_config(self) -> "ModelConfigCpp":
             data_type=torch_dtype_to_binding(
                 self.pretrained_config.torch_dtype))
 
-        mlp_hidden_size = self.pretrained_config.intermediate_size // self.mapping.tp_size
+        if self.pretrained_config.intermediate_size is not None:
+            mlp_hidden_size = self.pretrained_config.intermediate_size // self.mapping.tp_size
+        else:
+            try:
+                architectures = self.pretrained_config.architectures
```
byshiue, line 198: Should we put model specific codes in `model_config.py`, which should be a general class?

tensorrt_llm/_torch/model_config.py
```diff
@@ -191,7 +191,34 @@ def get_bindings_model_config(self) -> "ModelConfigCpp":
...
+                    # TODO smor: this is a hack to support Nemotron-Super-49B-v1 with LoRA (https://jirasw.nvidia.com/browse/TRTLLM-5045)
+                    # Nemotron-NAS has variable ffn_mult for each layer, we need to find the maximum
+                    # so that we don't set a too small mlp_hidden_size. This solution leads to a memory
+                    # consumption that is higher than required.
+                    biggest_ffn_mult = max([
+                        x.ffn.ffn_mult
+                        for x in self.pretrained_config.block_configs
+                    ])
+
+                    from tensorrt_llm._torch.models.modeling_nemotron_nas import \
```
byshiue, line 210: Should move to the beginning.

tensorrt_llm/_torch/model_config.py
```diff
@@ -191,7 +191,34 @@ def get_bindings_model_config(self) -> "ModelConfigCpp":
...
 
-        mlp_hidden_size = self.pretrained_config.intermediate_size // self.mapping.tp_size
+        if self.pretrained_config.intermediate_size is not None:
+            mlp_hidden_size = self.pretrained_config.intermediate_size // self.mapping.tp_size
+        else:
+            try:
+                architectures = self.pretrained_config.architectures
+                if len(architectures
+                       ) == 1 and architectures[0] == "DeciLMForCausalLM":
+                    # TODO smor: this is a hack to support Nemotron-Super-49B-v1 with LoRA (https://jirasw.nvidia.com/browse/TRTLLM-5045)
```
byshiue, line 201: Please don't expose the internal link to github.

tensorrt_llm/_torch/model_config.py
```diff
@@ -191,7 +191,34 @@ def get_bindings_model_config(self) -> "ModelConfigCpp":
             data_type=torch_dtype_to_binding(
                 self.pretrained_config.torch_dtype))
 
-        mlp_hidden_size = self.pretrained_config.intermediate_size // self.mapping.tp_size
+        if self.pretrained_config.intermediate_size is not None:
+            mlp_hidden_size = self.pretrained_config.intermediate_size // self.mapping.tp_size
+        else:
+            try:
```
byshiue, line 197: I feel it is strange to use
```python
try:
    ...
    raise ValueError("")
except Exception as e:
    raise ValueError(...)
```

if we raise error in "try", why do you need to use "try"?

tests/unittest/llmapi/test_llm_pytorch.py
shaharmor98: All the tests in `test_llm_pytorch` and `test_llm_multi_gpu_pytorch` are the same as to those we have in the TRT flow.  
I've added exactly these same tests because, for the first iteration, we wanted to be on par with TRT, and I thought these tests are the right one to assert that.  

As for the proposed fusion- I can think of only two tests that could be fused, but still, it's not 100% logical to me. Could you please explain what overhead is caused by the tests? Perhaps I get something wrong here.  

As for the CI test list- I did't know you have to explicitly declare the tests. I know this file is configured to run on A100.  
Could you please refer me to an example you want me to adhere to? I thought tests are collected automatically

tests/unittest/llmapi/test_llm_pytorch.py
shaharmor98: You've raised two good points.  
Honestly, the adapter we currently have is pretty much useless (all `B` matrices are all zeros).  
I'm not sure if this model is supported with FP8, and unfortunately, the high memory consumption is due to a naive solution we provide (because of variable FFN).  

Do you think I should relocate this test to other file and assign it explicitly to H200?

tensorrt_llm/lora_manager.py
shaharmor98: You're right, I must have changed that accidentally when relocated the attribute.  
Since it's an API-related decision- what do you think we should set here? I'd go with 4 as the previous default.

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -665,13 +667,18 @@ def filter_weights(prefix, weights: Dict):
             if model.config.tie_word_embeddings and name.startswith("lm_head"):
                 continue
 
-            # Skip loading weights for embedding and lm_head if LoRA is enabled
+            # Skip loading weights for embedding and lm_head if LoRA is enabled and has custom values
```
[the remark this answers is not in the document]
shaharmor98: Changed, looks much better, thanks!

tensorrt_llm/_torch/models/modeling_utils.py
shaharmor98: Moved

tensorrt_llm/_torch/models/modeling_nemotron_nas.py
shaharmor98: Moved

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -772,14 +772,18 @@ def __init__(self, model_config: ModelConfig[LlamaConfig]):
...
+        has_custom_embed_tokens = False
         if hasattr(
                 model_config,
                 'lora_config') and model_config.lora_config is not None and len(
                     model_config.lora_config.lora_dir) == 1:
             from tensorrt_llm.lora_manager import HfLoraLoader
             lora_loader = HfLoraLoader(model_config.lora_config.lora_dir)
-            weight = lora_loader.embed_tokens
-            vocab_size = lora_loader.vocab_size
+            if lora_loader.vocab_size != 0:
```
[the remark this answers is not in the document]
shaharmor98: You're right. Good catch, thanks

tensorrt_llm/_torch/models/modeling_nemotron_nas.py
shaharmor98: Done

tensorrt_llm/_torch/models/modeling_utils.py
shaharmor98: DONE

tensorrt_llm/_torch/model_config.py
shaharmor98: Ideally not. I don't like this either, and I'm open for suggestions.  
However, as far as I'm concerned, it's a placeholder that is required only because of the late request to support Nemotron-NAS and due to the time limit we have.  

As I said, if you think there is a more suitable place for this "patch" I'd be happy to hear

tensorrt_llm/_torch/model_config.py
shaharmor98: It causes a circular import exception.  
Is there other alternative you think is suitable here?

tensorrt_llm/_torch/model_config.py
shaharmor98: Got it.  
How then should I reflect (as we did in the past) that this issue is tracked on an existing ticket?

tensorrt_llm/_torch/model_config.py
shaharmor98: You're right, the previous solution wasn't good enough.  
Changed that

tensorrt_llm/_torch/model_config.py
byshiue: You can say JIRA ticket TRTLLM-5045 directly. It is fine.

tests/unittest/llmapi/test_llm_pytorch.py
byshiue: Here `tests/integration/test_lists/test-db/l0_a100.yml`.

tests/unittest/llmapi/test_llm_pytorch.py
byshiue: I think we don't have H200 in CI system now.

tensorrt_llm/lora_manager.py
byshiue: We can set it to 4 first.

tensorrt_llm/_torch/model_config.py
byshiue: @QiJune  Do you have any suggestion?

QiJune: Yes, I agree that we should put model specific codes in its `modeling_xxx.py`. For example, define a `DeciLMConfig` which derives from `ModelConfig` base class. However, we don't have `tensorrt_llm._torch.AutoConfig` yet.

My suggestion is that adding a TODO in this MR. When we support `tensorrt_llm._torch.AutoConfig` in the future, we can clean up it.

Superjomn (approved):
LGTM for the API part modification.

tests/unittest/llmapi/test_llm_pytorch.py
shaharmor98: DONE

tensorrt_llm/lora_manager.py
shaharmor98: DONE

tensorrt_llm/_torch/model_config.py
shaharmor98: Done, thanks @QiJune

tensorrt_llm/_torch/model_config.py
shaharmor98: DONE

tests/unittest/llmapi/test_llm_pytorch.py
byshiue: @syuoni Do you think it is fine to put the nemotron and codellama lora tests into this file? 

Especially we run most of the tests on A100 and only plan to run nemotron test on H200.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1219,11 +1227,41 @@ def _setup(self):
...
+                    "max_loras is ignored when lora_config is provided.")
+            if self.max_cpu_loras != self.lora_config.max_cpu_loras:
+                logger.warning(
+                    "max_cpu_loras is ignored when lora_config is provided.")
+
+            if len(self.lora_config.lora_dir) == 0:
+                # TODO [TRTLLM-5173]
+                logger.warning(
+                    "lora_dir is empty, so custom embedding or lm head will not be applied."
+                )
```
syuoni, line 1246: I could be wrong. In my impression, LoRA adapters with embedding/lm_head are rare. So instead, maybe we should warn for the case with custom embedding/lm_head.

tests/unittest/llmapi/test_llm_pytorch.py
```diff
@@ -171,6 +180,131 @@ def test_llama_v2_13b_lora():
...
+    sampling_params = SamplingParams(max_tokens=10, add_special_tokens=False)
+    lora_req = LoRARequest(
+        "task-0", 0,
+        f"{llm_models_root()}/nemotron-nas/Llama-3_3-Nemotron-Super-49B-v1-lora-adapter_r64"
+    )
+    lora_request = [lora_req, None]
+
+    outputs = llm.generate(prompts, sampling_params, lora_request=lora_request)
+
+    assert similar(outputs[0].outputs[0].text, outputs[1].outputs[0].text)
```
syuoni, line 247: Does this mean the output is same regardless using LoRA or not?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1219,11 +1227,41 @@ def _setup(self):
...
+            if len(self.lora_config.lora_dir) == 0 and len(
+                    self.lora_config.lora_target_modules) == 0:
+                logger.warning(
+                    "Both lora_dir and lora_target_modules are empty, so all LoRA modules will be expected. "
+                    "This will lead to serious memory consumption. Please provide either lora_dir or lora_target_modules if this behavior is not what you expect."
+                )
+                default_trtllm_modules_to_hf_modules = get_default_trtllm_modules_to_hf_modules(
+                )
+                self.lora_config.lora_target_modules = list(
+                    default_trtllm_modules_to_hf_modules.keys())
```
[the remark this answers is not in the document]
shaharmor98: No, not at all. As the warning says, it will only result in a (redundant) increased memory usage

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -787,13 +788,22 @@ class LlmArgs(BaseModel):
...
+    max_loras: int = Field(default=4,
+                           description="The maximum number of LoRA.",
+                           deprecated="Use lora_config.max_loras instead.")
 
-    max_cpu_loras: int = Field(default=4,
-                               description="The maximum number of LoRA on CPU.")
+    max_cpu_loras: int = Field(
+        default=4,
+        description="The maximum number of LoRA on CPU.",
+        deprecated="Use lora_config.max_cpu_loras instead.")
```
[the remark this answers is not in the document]
shaharmor98: I'm not sure you're right. At the moment, both `max_cpu_lora` of `LoraConfig` and the old one is 4, so they will be equivalent.  
As for setting to `None`, I think the old behavior (of setting to 4) is the correct one. I believe @byshiue agrees (we had a different thread on that).  
@byshiue what is your opinion?

tensorrt_llm/llmapi/llm_args.py
shaharmor98: Honestly I don't have an estimate of how common adapters with embedding or lm_head are.  
However, I thought of this warning only because I'm concerned about scenarios where an embedding/lm_head exists, but the user either doesn't provide `target_modules` or `lora_dir` (due to the new default modules setting feature) or configures custom target modules (which would result in embedding/lm_head being ignored).  
WDYT?

tests/unittest/llmapi/test_llm_pytorch.py
shaharmor98: Yes,  
Unfortunately, we were only given dummy adapters (all `B` matrices are zeros). So it's more of a sanity check than a rigorous correctness check.

tensorrt_llm/llmapi/llm_args.py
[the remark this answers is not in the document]
byshiue: I think we can set them to 4 by default. And if we don't want to enable lora, we should set LoraConfig=None directly.

byshiue (approved):

tensorrt_llm/llmapi/llm_args.py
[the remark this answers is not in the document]
syuoni: I see, thanks.

tensorrt_llm/llmapi/llm_args.py
syuoni: I remember vLLM even didn't support LoRA checkpoints with lm_head/embeddings when I survey its LoRA API last time. So I suppose such LoRA checkpoints are rare.

Maybe we can survey vLLM/SGLang again for this. For now, we can keep the current implementation.

tests/unittest/llmapi/test_llm_pytorch.py
syuoni: Then, how about removing this test case for now? It doesn't effectively protect the LoRA feature on this model, right? We can add it back when we have a real adapter.

tests/integration/test_lists/test-db/l0_dgx_h200.yml
```diff
@@ -18,3 +18,4 @@ l0_dgx_h200:
   # - accuracy/test_llm_api_pytorch.py::TestDeepSeekR1::test_fp8_blockscale[throughput] # OOM
   - accuracy/test_llm_api_pytorch.py::TestDeepSeekR1::test_fp8_blockscale[latency] # 1h
   - unittest/_torch/multi_gpu_modeling/test_llama4.py::test_llama4[pp1-ep1-enable_graph-tp8-trtllm-scout]
+  - unittest/llmapi/test_llm_pytorch.py::test_nemotron_nas_lora
```
syuoni, line 21: Please make sure this test item can be correctly recognized. @VALLIS-NERIA  Could you please help double check?

syuoni (approved):
LGTM, thanks!

tests/integration/test_lists/test-db/l0_dgx_h200.yml
syuoni: Thanks @VALLIS-NERIA

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -787,13 +788,22 @@ class LlmArgs(BaseModel):
...
+
+    lora_config: Optional[LoraConfig] = Field(
```
QiJune, line 805: Is there any case that lora_config is not None, but enable_lora is True? In other words, can we remove `enable_lora` flag, and use `lora_config=None` represent lora is not enabled.

Superjomn: Agree. Maybe we can `lora_config=LoraConfig()` to have the same behaviour with the `enable_lora=True`, just from the API perspective.

shaharmor98: I'm not sure what you offered would work, since we still want to support the TRT backend, and it relies on different things then what we assume in the PyT flow.  
I could search for alternatives, or explain in greater details the current situation. You might be right after all, I'm a bit concerned because of the feature freeze timeline (and the fact I still had no green pipelines- all because of unrelated failures)

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -671,12 +672,13 @@ def filter_weights(prefix, weights: Dict):
...
 
-            # Skip loading weights for embedding and lm_head if LoRA is enabled
-            if hasattr(model.model_config, 'lora_config'
-                       ) and model.model_config.lora_config is not None and len(
-                           model.model_config.lora_config.lora_dir) == 1 and (
-                               name == "model.embed_tokens"
-                               or name == "lm_head"):
+            # Skip loading weights for embedding and lm_head if LoRA is enabled and has custom values
+            if hasattr(
+                    model.model, 'has_custom_embed_tokens'
```
amukkara, line 677: potential reason why llava test case fails.

tensorrt_llm/llmapi/llm_args.py
QiJune: Sure, we can do it in the following PRs.

QiJune (approved):
LGTM

PR #4202, opened 2025-05-10
[draft] Refactor quant in linear

HuiGao-NV:


tensorrt_llm/_torch/modules/linear.py
```diff
@@ -200,22 +202,28 @@ def __init__(
...
     def create_weights(self):
         if self._weights_created:
             return
         weight_shape = (self.out_features, self.in_features)
-        self.has_any_quant = False
         self.has_fp8_qdq = False
         self.has_fp8_block_scales = False
         self.has_nvfp4 = False
 
+        self.input_quantizer = LinearQuant.create_quantizer(self.quant_config)
```
QiJune, line 223: Based on current prototype, I would suggest we refactor Linear module or FusedMoE module seperately. The code sharing between Linear module and FusedMoE is limited. For example, this PR introduces +528/-301 diff.

tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -1366,48 +1435,23 @@ def _load_fp8_block_scales_scales(self, weights: Dict):
             [torch.stack(all_w3_scales),
              torch.stack(all_w1_scales)], dim=-2)
         self.w3_w1_weight_scaling_factor.data.copy_(w3_w1_scales)
+        self.input_quantizer.load_weights_custormized(
```
QiJune, line 1438: custormized -> customized

tensorrt_llm/_torch/modules/linear.py
HuiGao-NV: This change shall not conflict with Linear/FusedMoE modules refactoring.
And it shall make the refactored Linear/FusedMoE code more clear because it makes Linear/FusedMoE only needs to focus its own kernels selection. It also can make the quantization method selection happens at module creation phase in a static way but not at runtime (forward). 
In current code, quantCreator is just a centralized place to select quantization method which are mixed in load_weights/forward. It shall be easier to make change, such as add new quantization method, for example new int4_weight_only_per_group method. However, we can also follow current logic by removing quantCreator and create needed quantizer in create_weights in each if-quant_mode condition.

PR #4215, opened 2025-05-12
tests: PyTorch multimodal using keyword match

amukkara:
## Description

Current accuracy checks expect token-level match in reference and generated output. However,  the same prompt can give slightly different (but semantically similar) outputs on different architectures and kernel implementation. As a result, we had to allow multiple reference outputs for each prompt/model combination. This becomes unmanageable as we add more models.

This PR replaces token-level match with keyword match that just checks if output contains a small list of reference keywords.

## Test Coverage

`tests/integration/defs/test_e2e.py::test_ptp_quickstart_multimodal[]`


amukkara:
@symphonylyh should we use smaller `max_new_tokens` in test than default 64 in `quickstart_multimodal.py`? 

For several models, the outputs that we currently allow only differ in the last 10-20 tokens. smaller `max_new_tokens` will reduce any flakiness that remains even with keyword-based match.

another question: how many keywords should be in the allow list for a given prompt/model? 1, 5, 10?

yechank-nvidia (commented):
LGTM. One question is that how you set those reference keywords?

yechank-nvidia (approved):

amukkara:
> LGTM. One question is that how you set those reference keywords?

asked a chat bot to pick 10 keywords from each valid output and trimmed to around 5 words manually.

QiJune (approved):
LGTM

PR #4226, opened 2025-05-12
fix potential issues in allreduce fusion kernel and ut

yilin-void:
potential issues:
1. The `float val[4]` is not aligned to 16 bytes, which may result in a misaligned address when reinterpret_cast `val` to `float4*`.
2. There is a typo in the Lamport kernel when handling `-0` in the data.
3. FP4 quant is not supported when sm < 100.

### DIFF
--- a/cpp/tensorrt_llm/kernels/communicationKernels/allReduceFusionKernels.cu
+++ b/cpp/tensorrt_llm/kernels/communicationKernels/allReduceFusionKernels.cu
@@ -456,10 +456,10 @@ __global__ void allreduce_fusion_kernel_oneshot_lamport(AllReduceFusionParams pa
 
     for (int idx = access_id; idx < tot_access; idx += access_stride)
     {
-        float val[4];
+        alignas(16) float val[4];
         *reinterpret_cast<float4*>(val) = reinterpret_cast<float4*>(params.allreduce_in)[idx];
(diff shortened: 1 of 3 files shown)


hyukn:
Some questions about these potential issues, not sure if I understand them correctly:
1 will not cause illegal memory access, but some precision issue? Because the data is misaligned in the wrong order.
2 will cause overflow/precision issues?
3 will not be a true issue because we have high level path to guarantee that this kernel will not be triggered before Blackwell?

yilin-void:
hi @hyukn ,

>1 will not cause illegal memory access, but some precision issue? Because the data is misaligned in the wrong order. 

This does not cause any problems at present, but customers have reported that this will cause compute-sanitizer to prompt misaligned address.
>2 will cause overflow/precision issues? 

If the last 8 bytes of a consecutive 16-byte segment in the actual data, when interpreted as two floats, happen to be [-0.f, 0.f], it will cause a hang. However, this bug has not been observed to be triggered so far.
>3 will not be a true issue because we have high level path to guarantee that this kernel will not be triggered before Blackwell?

This issue is only relevant to the cpp unit test, in actual use, we will not use fp4 on pre-Blackwell.

hyukn (approved):
Thanks a lot for the detailed explanation. LGTM.

PR #4243, opened 2025-05-13
[feat][TRTLLM-5018] Dis serving python runtime trt backend

pcastonguay:
# PR title

## Description

Adds support for TRT backend when running disaggregated serving with trtllm-serve

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -889,12 +889,11 @@ class PyExecutor:
                                                       batch_outputs)
 
                     self._update_request_states(scheduled_batch)
+                    self._update_requests(sample_state)
 
                     ctx_transmission_reqs = self._send_disagg_ctx_cache(
                         scheduled_batch.context_requests
                     ) if self.kv_cache_transceiver else []
-
-                    self._update_requests(sample_state)
 
                     if num_dummy_request > 0:
                         self._finish_dummy_request(scheduled_batch)
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -1597,7 +1597,8 @@ class PyExecutor:
         if (scheduled_ctx_requests is None or len(scheduled_ctx_requests) == 0):
             return []
         for req in scheduled_ctx_requests:
-            if req.is_context_only_request and req.is_context_finished:
+            if req.is_context_only_request and (req.is_context_finished or
+                                                req.is_finished_due_to_length):
                 self.kv_cache_transceiver.respond_and_send_async(req)
                 self.resource_manager.resource_managers[
                     "seq_slot_manager"].free_resources(req)
--- a/tensorrt_llm/llmapi/llm_utils.py
+++ b/tensorrt_llm/llmapi/llm_utils.py
@@ -750,6 +750,11 @@ class CachedModelLoader:
 
                 if self.llm_args.parallel_config.is_multi_gpu:
                     assert self.mpi_session
+
+                    #mpi_session cannot be pickled so remove from self.llm_args
+                    if self.llm_args.mpi_session:
+                        del self.llm_args.mpi_session
+
                     # The engine_dir:Path will be stored to MPINodeState.state
                     build_infos = self.mpi_session.submit_sync(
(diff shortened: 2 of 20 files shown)


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -864,13 +864,12 @@ def _executor_loop(self):
                                                        batch_outputs)
 
                     self._update_request_states(scheduled_batch)
+                    self._update_requests(decoder_state)
```
Tabrizian, line 867: Just curious why do we needed to move this line?

tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -43,6 +43,10 @@
                     print_colored_debug, print_traceback_on_error)
 
 
+def dummy_task():
```
Tabrizian, line 46: Should we remove this?

tensorrt_llm/_torch/pyexecutor/py_executor.py
pcastonguay: Because we need the decoder to set the ctx request state to `COMPLETE` before we try to send the KV cache. This way, if request finishes after 1st token, the KV cache wont be sent.

tensorrt_llm/llmapi/llm_utils.py
pcastonguay: Yes will do. Thx.

tensorrt_llm/_torch/pyexecutor/py_executor.py
Tabrizian: I see, thanks.

Tabrizian (approved):

tensorrt_llm/_torch/pyexecutor/py_executor.py
Shunkangz: It appears that the KV cache transfer is completed in the `_prepare_disagg_gen_transmission_complete` function. How can we ensure that the KV cache is not allocated or transferred when it is not needed? Additionally, I am unsure whether the current implementation allows the context to cancel the transfer if the KV cache is not required.

tensorrt_llm/serve/openai_disagg_server.py
```diff
@@ -168,10 +176,6 @@ async def _process_context_server_request(self, ctx_req, request_type: str):
             return None
 
         try:
-            if request_type == "chat":
-                ctx_req.max_completion_tokens = 1
-            elif request_type == "completion":
-                ctx_req.max_tokens = 1
```
Shunkangz, line 174: May I know why we remove this?

pcastonguay: We no longer need it. In fact, with the TRT runtime, if we leave it, the generation will be marked as complete after the ctx server generates the 1st token, and the KV cache transfer won't happen.

tensorrt_llm/_torch/pyexecutor/py_executor.py
pcastonguay: If the context server marks the request as 'generation_complete` after the first token (which will happen if it encounters stop word for example), `req.is_context_finished` will return `False`, and the KV cache transfer won't get triggered. 

here: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/py_executor.py#L1578

This is the c++ `isContextFinished` https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/include/tensorrt_llm/batch_manager/llmRequest.h#L1420-L1423 which returns false if `state ==  LlmRequestState::kGENERATION_COMPLETE`. 


The proxy server (disagg server) will see the `finish_reason` as being `!= not_finished` and will never send the request to the generation server.

tensorrt_llm/serve/openai_disagg_server.py
Shunkangz: Thank you

tensorrt_llm/_torch/pyexecutor/py_executor.py
Shunkangz: Thank you!

Shunkangz (approved):

PR #4249, opened 2025-05-13
fix: add max_batch_size to the cuda graph batch sizes

lfr-0531:
## Description
When `max_batch_size` is not in the `cuda_graph_batch_sizes` list, there will be no cuda graph for such a most commonly used case, and the perf will be poor.

In this PR, we add `max_batch_size` to the cuda graph batch sizes. With this fix, trtllm will set it by default even if we forgot to add the `max_batch_size` to `cuda_graph_batch_sizes`.


kaiyux (approved):

kaiyux:
Closing as per offline discussion.

PR #4265, opened 2025-05-13
[https://nvbugspro.nvidia.com/bug/5243740][fix] deduce default max_tokens for trtllm-serve

LinPoly:
## Description

Deduce the default `max_tokens` for both TensorRT and PyTorch workflow, to improve usability, remove the default value of `max_tokens` in `openai_protocol.py`.

## Test Coverage

Original tests are good.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/executor/worker.py
+++ b/tensorrt_llm/executor/worker.py
@@ -394,11 +394,24 @@ class ExecutorBindingsWorker(GenerationExecutor):
                 )
 
         assert request.id is not None
+
+        def _deduce_max_tokens(request, executor_config):
+            if request.sampling_params.max_tokens:
+                return request.sampling_params.max_tokens
+            # deduce max_tokens when it's not set by user
+            query_token_len = len(
+                request.query_token_ids) if request.query_token_ids else 0
+            cp_size = 1 if not hasattr(
+                executor_config, "mapping") else executor_config.mapping.cp_size
+            default_max_tokens = self._executor_config.max_seq_len - len(
+                prompt_token_ids) / cp_size - query_token_len
+            return int(default_max_tokens)
+
         try:
             executor_request = tllm.Request(
                 client_id=request.id,
                 input_token_ids=prompt_token_ids,
-                max_tokens=request.sampling_params.max_tokens,
+                max_tokens=_deduce_max_tokens(request, self._executor_config),
                 streaming=request.streaming,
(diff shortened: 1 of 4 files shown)


kaiyux (approved):

tensorrt_llm/executor/worker.py
```diff
@@ -394,11 +394,24 @@ def _enqueue_request(self, request: GenerationRequest) -> int:
                 )
 
         assert request.id is not None
+
+        def _deduce_max_tokens(request, executor_config):
```
Superjomn, line 398: This is necessary for performance, right? I wonder if we need to cover this logic in the unittest, if it is broken during the iterations, it may make the performance drop without any functional error.

Superjomn: @LinPoly can you add one test? The deduce function could be modularized somewhere to be accessible. At least we should check the deduce part; an end-to-end one is better, like give a prompt and check its deduced max_tokens, but I am not sure that is possible.

LinPoly: > This is necessary for performance, right?

Actually this is for usability, we set `max_tokens` for `sampling_params`, instead of `max_num_tokens` for `executor_config`. With this change, users don't need to set `max_tokens` in OpenAI API to get reasonable response

> it may make the performance drop without any functional error

If users don't give a `max_tokens` and we cann't deduce it, then an error will be raise from [tllm.Request instantiation](https://github.com/NVIDIA/TensorRT-LLM/blob/14bfb5e0d6e81aec3306a1324cf074566646f886/tensorrt_llm/executor/worker.py#L398) because it won't accept None for `max_tokens`. So I think we don't need to worry error raising?

> At least we should check the deduce part; an end-to-end one is better, like give a prompt and check its deduced max_tokens

How about raising an error explicitly when deduced max_tokens is still None or less than 0? This means 1. users don't give a max_tokens 2. we cannot infer the max_tokens from max_seq_len, and this is abnormal, then we add tests in both trtllm-serve and LLMAPI level to check if there is an error.

Superjomn: I see, then I am fine with it as explicit errors will be raised.

Superjomn (approved):
LGTM

PR #4289, opened 2025-05-14
feat: change eagle one model flag from an env var to a config in Eagle3Config

lfr-0531:
## Description

We used `TRTLLM_EAGLE3_ONE_MODEL` to control whether to use one-model or two-model eagle3 implementation. As suggested by @hlu1, we change it to a config in Eagle3Config in this PR.

### DIFF
--- a/tensorrt_llm/_torch/speculative/eagle3.py
+++ b/tensorrt_llm/_torch/speculative/eagle3.py
@@ -21,14 +21,13 @@ class Eagle3Config(SpecConfig):
     spec_dec_name: str = "EAGLE3"
     num_layers: int = 0
     hidden_size: int = 0
+    eagle3_one_model: bool = True
 
     def __post_init__(self):
         if self.draft_model_path is None:
             raise ValueError("Path to EAGLE3 weights must be specified.")
 
-        # TODO: remove this once we have a better way to handle the one model case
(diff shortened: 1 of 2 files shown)


hlu1 (approved):

tensorrt_llm/_torch/speculative/eagle3.py
```diff
@@ -21,14 +21,13 @@ class Eagle3Config(SpecConfig):
     spec_dec_name: str = "EAGLE3"
     num_layers: int = 0
     hidden_size: int = 0
+    eagle3_one_model: bool = True
```
nv-yilinf, line 24: Since one model is enabled by default, is it better to rename this config as `eagle3_two_model` and set it to be false by default?

lfr-0531: Currently we are using one model by default. After cpu overhead optimization of two model path, we will change the default to two model approach. So I prefer to use `eagle3_one_model` here. Once the two-model path is ready, we change the default value from `True` to `False` and keep using the same interface name.  

Does this make sense to you?

nv-yilinf: I see. That makes sense. Thanks

nv-yilinf (approved):

PR #4290, opened 2025-05-14
[TRTLLM-5273]feat/Use full attention mask if Llama3 is used as encoder and fix EarlyStopDecoder unsqueeze bug

nvrohanv:
# [TRTLLM-5273]feat/Use full attention mask if Llama3 is used as encoder and fix EarlyStopDecoder unsqueeze bug

## Description

This PR adds in a flag for bidirectional_attention to `modeling_llama.py`. This is to support some customer use-cases for llama-based embedding models, see Jira TRTLLM-5273 for more details.

Additionally it fixes a bug in EarlyStopDecoder where 1d logits were being unsqueezed incorrectly, leading to incompatibility with LogitsStorage. 

Lastly it hardcodes `LlamaForTextEmbedding` as a non-generation model for now so that the customer can make their own out-of-tree example with their custom model and still leverage EarlyStopDecoder. I am in discussions with Erin Ho and Freddy Qi about how we can eliminate the need to hardcode this in the future. 

**Things To Review**

1. Should this bidirectional attention be supported via flag like the current implementation or should it be a separate class?

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_llama.py
+++ b/tensorrt_llm/_torch/models/modeling_llama.py
@@ -201,6 +201,8 @@ class LlamaAttention(Attention):
             dtype=config.torch_dtype,
             config=model_config,
         )
+
+
 class Llama4MoE(nn.Module):
 
     def __init__(
@@ -506,7 +508,7 @@ class LlamaDecoderLayer(DecoderLayer):
         self.post_attention_layernorm = RMSNorm(hidden_size=config.hidden_size,
                                                 eps=config.rms_norm_eps,
                                                 dtype=config.torch_dtype)
-        
+
         self.attention_mask = PredefinedAttentionMask.CAUSAL
         if bidirectional:
             self.attention_mask = PredefinedAttentionMask.FULL
@@ -721,7 +723,9 @@ class Llama4Model(DecoderModel):
 
 class LlamaModel(DecoderModel):
 
-    def __init__(self, model_config: ModelConfig[LlamaConfig], bidirectional=False):
+    def __init__(self,
+                 model_config: ModelConfig[LlamaConfig],
+                 bidirectional=False):
         super().__init__(model_config)
         config = self.model_config.pretrained_config
         self.padding_idx = config.pad_token_id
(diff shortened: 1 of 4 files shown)


tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -527,6 +532,7 @@ def forward(
             position_ids=position_ids,
             hidden_states=hidden_states,
             attn_metadata=attn_metadata,
+            attention_mask=self.attention_mask,
```
symphonylyh, line 535: is attention_mask usually a init-stage property at LlamaAttention() or a runtime input parameter? @qixiang-99 
if init, how about merge it inside model_config and handle that in the attention layer?

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -728,7 +734,9 @@ def forward(
 @support_pp
 class LlamaModel(DecoderModel):
 
-    def __init__(self, model_config: ModelConfig[LlamaConfig]):
+    def __init__(self,
+                 model_config: ModelConfig[LlamaConfig],
+                 bidirectional=False):
```
symphonylyh, line 739: for example, I think other people may complain about adding a dangling field here

tensorrt_llm/_torch/models/modeling_llama.py
qixiang-99: In most cases I've encountered, the attention_mask can theoretically be determined during init-stage. However, the attention forward method mandates the attention_mask as an input ([link](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/attention_backend/interface.py#L530)).

tensorrt_llm/_torch/models/modeling_llama.py
qixiang-99: I agree with this. The primary purpose of model_config is to maintain a clean argument list, so some refactoring might be necessary.

nvrohanv: Got it in that case should I make a LlamaWithBidirectionalAttention class?

tensorrt_llm/_torch/models/modeling_llama.py
nvrohanv: Resolved by just using model_config.is_generation. For now since there aren't any requests for encoder models with custom embeddings this will do and we can revisit if and when a custom embedding case comes.

symphonylyh: attention_mask --> attention_mask_type you plan to do in a follow-up chore PR, do I understand it correctly?

symphonylyh (approved):

tensorrt_llm/_torch/models/modeling_llama.py
nvrohanv: Yes, that is the plan

qixiang-99:
LGTM

PR #4344, opened 2025-05-15
feat: Low Precision Allreduce for PCIe based GPU 

kanghui0204:
last PR：https://github.com/NVIDIA/TensorRT-LLM/pull/3851
last revet PR：https://github.com/NVIDIA/TensorRT-LLM/pull/4340


yizhang-nv:
Is this set of kernels considered cuda graph support? If the barrier flag is captured, during the graph replay, it may cause issues since we depend on the value of the barrier flag and the comm buffer to make sure every gpu reach to the same barrier. 

For the captured value, if the model has odd number of ar, then it may select the same peer comm buffer here:
https://github.com/NVIDIA/TensorRT-LLM/pull/4344/files#diff-fd189077a08106939fbdaf23180ba0bc7d81d76279c632db9097381e8440b2c9R1422

kanghui0204:
> Is this set of kernels considered cuda graph support? If the barrier flag is captured, during the graph replay, it may cause issues since we depend on the value of the barrier flag and the comm buffer to make sure every gpu reach to the same barrier.
> 
> For the captured value, if the model has odd number of ar, then it may select the same peer comm buffer here: https://github.com/NVIDIA/TensorRT-LLM/pull/4344/files#diff-fd189077a08106939fbdaf23180ba0bc7d81d76279c632db9097381e8440b2c9R1422

Do all our current kernels need to support CUDA graphs? I haven't tested these kernels on CUDA graphs.

hyukn (approved):

hyukn:
Pipeline passed. Merge this PR.

PR #4348, opened 2025-05-15
Fix bias shape in weightOnlyGroupwiseQuantMatmulPlugin for TRT workflow

StudyingShao:
Thanks @AIDC-AI for reporting and locating this bug.


Barry-Delaney (approved):
LGTM.

PR #4351, opened 2025-05-15
[fix] Fixed incorrect mixed precision MoE conversion

Barry-Delaney:
This PR fixes the problem of unable to find OP and only rank 0 can be exported for mixed MoE conversion.


Barry-Delaney:
/bot skip --comment "Changes on script with no CI related"

QiJune (approved):
LGTM

PR #4378, opened 2025-05-16
[CI] update multi-gpu test triggering file list

QiJune:


lfr-0531 (approved):

lfr-0531:
Should we also add `resource_manager.py`?

QiJune:
> Should we also add `resource_manager.py`?

Ideally, we should include the whole `_torch/py_executor` directory.  It's a tradeoff between developing velocity and CI stability. Let 's merge this PR first and keep observing.

PR #4379, opened 2025-05-16
fix: fix accuracy and illegal memory access issues when using mtp + attention dp

lfr-0531:
## Description

Fixed two issues:
- MTP + attention dp illegal memory access
- MTP + attention dp + overlap scheduler accuracy drop

The illegal memory access is caused by the token_num in `add_dummy_requests`. With token_num=1, the prompt_len will be zero. But our attention kernel cannot handle such a special case, so we got illegal memory access. The fix in this PR should be a workaround to make Attention DP+MTP work.

The accuracy drop is because we didn't correctly prepare the inputs for the dummy requests when enabling overlap scheduler and speculative decoding.

Other changes:

- Change the default py_draft_tokens to an empty list.

NOTE: there is still an illegal memory access issue when running DeepSeek-V3-Lite MTP with tp + autotuner + cuda graph + overlap scheduler. We track this issue in nvbug 5304040.

## Test Coverage

The waived tests are added back.


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1298,8 +1298,7 @@ def _merge_dummy_request(self, num_dummy_request: int):
             request_ids=list(range(num_dummy_request)),
             is_gen=not self.has_context_request,
             prepare_resource=not self.has_context_request,
-            max_num_draft_tokens=0
-            if self.has_context_request else self.max_draft_tokens,
+            max_num_draft_tokens=self.max_draft_tokens,
```
zhhuang-nv, line 1301: Just curious that why we should modify these lines. If `self.has_context_request` is `True` then `is_gen` will be `False`, which will not trigger the codes modified in `tensorrt_llm/_torch/pyexecutor/resource_manager.py::add_dummy_requests`.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -334,11 +334,11 @@ def add_dummy_requests(
                 req.state = LlmRequestState.GENERATION_IN_PROGRESS
                 req.prompt_len = token_num - 1
                 req.py_prompt_len = req.prompt_len
-                if max_num_draft_tokens > 0:
-                    req.py_draft_tokens = [1] * max_num_draft_tokens
+                req.py_draft_tokens = [1] * max_num_draft_tokens
```
zhhuang-nv, line 337: Will it be ok when `req.py_draft_tokens` is an empty list?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -334,11 +334,11 @@ def add_dummy_requests(
...
+                for _ in range(max_num_draft_tokens):
+                    req.add_new_token(1, 0)
```
zhhuang-nv, line 339: It seems that we also haven't called `req.add_new_token` before https://github.com/NVIDIA/TensorRT-LLM/pull/4084, but there is a difference that the `req.prompt_len` is `token_num - 1 + max_num_draft_tokens` instead of ``token_num - 1` before that PR. So the key point is `req.prompt_len`, is that right?

BTW I totally agree that we should call `req.add_new_token` to make `LlmRequest` aligned with python request.

tensorrt_llm/_torch/pyexecutor/py_executor.py
lfr-0531: In `add_dummy_requests`, we only use `max_num_draft_tokens` when` is_gen=True`. So we don't need to set it to `0` outside the `add_dummy_requests` and just use `is_gen=False` to skip.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
lfr-0531: Yes, `req.py_draft_tokens` is with type of List.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
lfr-0531: Before https://github.com/NVIDIA/TensorRT-LLM/pull/4084, the `token_num = 1 + max_num_draft_tokens`. And we used this token_num to get the `input_ids = [1] * token_num`. So we didn't need to call `req.add_new_token` for those draft tokens. But after that PR, we moved all of the codes related to the draft tokens to the end of `add_dummy_requests`, we need to use `req.add_new_token` to update `input_ids`.

zhhuang-nv: Got it. Current codes are clean and clear. We only use `max_num_draft_tokens` for generation requests.

tensorrt_llm/_torch/pyexecutor/py_executor.py
zhhuang-nv: Thanks, that makes sense.

zhhuang-nv (approved):

PerkzZheng:
@lfr-0531 have all the illegal memory access issues been resolved or the one you mentioned still exists ? thanks.

lfr-0531:
> @lfr-0531 have all the illegal memory access issues been resolved or the one you mentioned still exists ? thanks.

No, we haven't fixed it. But looks like we didn't cover this special case in our [test list](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/test_lists/test-db/l0_h100.yml) before. I mean the setting `mtp_nextn=2-attention_dp=False-cuda_graph=True-overlap_scheduler=True`. I want to merge this PR first so that we can have other waived tests back. And we can track this special case in a new nvbug and fix it later. 

@PerkzZheng, what do you think?

PerkzZheng:
> > @lfr-0531 have all the illegal memory access issues been resolved or the one you mentioned still exists ? thanks.
> 
> No, we haven't fixed it. But looks like we didn't cover this special case in our [test list](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/test_lists/test-db/l0_h100.yml) before. I mean the setting `mtp_nextn=2-attention_dp=False-cuda_graph=True-overlap_scheduler=True`. I want to merge this PR first so that we can have other waived tests back. And we can track this special case in a new nvbug and fix it later.
> 
> @PerkzZheng, what do you think?

that makes sense to me.

syuoni (approved):

tests/integration/test_lists/waives.txt
```diff
@@ -453,18 +447,6 @@ accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales[mtp_
 accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales[mtp_nextn=2-fp8kv=False-attention_dp=False-cuda_graph=True-overlap_scheduler=False-torch_compile=False] SKIP (https://nvbugs/5286795)
 accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales[mtp_nextn=2-fp8kv=False-attention_dp=False-cuda_graph=False-overlap_scheduler=True-torch_compile=False] SKIP (https://nvbugs/5286795)
 accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales[mtp_nextn=2-fp8kv=True-attention_dp=True-cuda_graph=True-overlap_scheduler=True-torch_compile=False] SKIP (https://nvbugs/5286795)
-accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_bfloat16[mtp_nextn=0-attention_dp=False-cuda_graph=False-overlap_scheduler=False-torch_compile=False] SKIP (https://nvbugs/5285965)
```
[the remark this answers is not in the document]
lfr-0531: I added those tests back to `waive.txt` again.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1105,32 +1106,43 @@ def _prepare_tp_inputs(
...
         previous_pos_indices = []
+        request_ids_with_previous_batch = []
+        num_extend_reqs_wo_previous_batch = 0
         for request in extend_requests:
             if next_draft_tokens_device is None or request.py_batch_idx is None:
-                num_draft_tokens = len(request.py_draft_tokens)
+                # the request has no previous device tensors:
+                # (1) next_draft_tokens_device is None, which means overlap scheduler is disabled; or
+                # (2) request.py_batch_idx is None, which means the request has no previous batch.
+                # the second condition includes dummy generation requests created for CUDA graph padding or
```
QiJune, line 1116: I think extend requests do not support CUDA graph.

QiJune (approved):
LGTM

tensorrt_llm/_torch/pyexecutor/model_engine.py
lfr-0531: Those extend requests are generation requests with draft tokens. They can support CUDA graph. And the dummy requests with draft tokens are also in the extend request list here.

lfr-0531:
Got another conflict. Rerun the pipeline.

PR #4399, opened 2025-05-16
fix: [nvbugs/5287097] Align PP layer distribution between pytorch and TRT flow.

yuxianq:


tensorrt_llm/mapping.py
```diff
@@ -420,12 +422,9 @@ def has_moe_ep(self):
...
 
     def pp_layers(self, num_layers: int) -> List[int]:
-        base_layers = num_layers // self.pp_size
-        extra_layers = num_layers % self.pp_size
-        start_idx = self.pp_rank * base_layers + min(self.pp_rank, extra_layers)
-        layers_in_stage = base_layers + (1
-                                         if self.pp_rank < extra_layers else 0)
-        return list(range(start_idx, start_idx + layers_in_stage))
+        # If num_layers % pp_size = n != 0, first n ranks get one extra layer
+        return torch.tensor_split(torch.arange(num_layers),
```
yuxianq, line 426: @amukkara This PR includes your refactor in https://github.com/NVIDIA/TensorRT-LLM/pull/3827.

cpp/include/tensorrt_llm/runtime/modelConfig.h
```diff
@@ -167,23 +167,40 @@ class ModelConfig
         LayerType layerType, SizeType32 pipelineParallelism = 1, SizeType32 pipelineParallelismRank = 0) const
     {
         TLLM_CHECK_WITH_INFO(pipelineParallelism > 0, "Invalid pipelineParallelism: %d", pipelineParallelism);
-        auto const numLocalLayers = mNbLayers / pipelineParallelism; // WARNING: assume no remainder
-        auto const firstLocalLayerIt = mLayerTypes.cbegin() + (numLocalLayers * pipelineParallelismRank);
+        auto const numBaseLayers = mNbLayers / pipelineParallelism;
```
achartier, line 170: Would be nice to avoid the code duplication with `getFirstLocalLayer` if possible.

amukkara: can reuse `getNbLayers` too.

tensorrt_llm/models/generation_mixin.py
```diff
@@ -277,8 +277,7 @@ def prepare_attention_inputs(
         local_attn_layers = [i for i in layers_range if i in attn_layer_idx]
         # number of attention layers local to previous pp ranks
         num_attn_layers_lower_ranks = attn_layer_idx.index(local_attn_layers[0])
-        num_attn_layers_prev_rank = num_attn_layers_lower_ranks // mapping.pp_rank if mapping.pp_rank != 0 else len(
-            local_attn_layers)
+        num_attn_layers = len(local_attn_layers)
```
amukkara, line 280: why change the calculation from "num layers in prev rank" to "num layers in current rank"?
shouldn't the prev rank still be used, but the calculation updated to account for unequal distribution of layers to ranks?

should `num_layers_prev_rank` calculation be changed as well?

yuxianq: @achartier I don't quite understand why we use `num_attn_layers_prev_rank` instead of the current rank's one, can you help to explain your code here?

achartier: When testing this code, the model was expecting the size of this tensor to be based on the number of layers at the previous rank.

cpp/include/tensorrt_llm/runtime/modelConfig.h
yuxianq: Updated.

tensorrt_llm/models/generation_mixin.py
yuxianq: Since `Mapping`'s `pp_layers` can handle unbalanced layer number now, we should use the layer number in the current rank for the size of tensors like `host_kv_cache_pool_mapping`. We have also updated KV cache manager to align with `Mapping`'s `pp_layers`.

achartier (approved):

PR #4440, opened 2025-05-19
Downgrade the logger level for fallback tactic warning.

hyukn:
Currently, we do not want any fallback tactics to be triggered because it will hurt the performance a lot. Downgrade the logger level to mute the issue for the moment because a larger sequence length is being seen.


hyukn:
Hi Tao. Could we skip CI for this to accelerate the process? Thanks a lot. @litaotju

litaotju (approved):

litaotju (approved):

PR #4447, opened 2025-05-19
fix: skip weights defined in create_weights for pp.

yuxianq:
Since `DecoderModel`'s `__pp_init__` is called before `DecoderModelForCausalLM`'s `__post_init__`, it fails to skip weights for those weights defined in `create_weights`, which is created inside `__post_init__`.
We call `DecoderModel`'s `__pp_init__` inside `DecoderModelForCausalLM`'s `__pp_init__` to fix it, since `DecoderModelForCausalLM`'s `__pp_init__` is called after its `__post_init__`.


Barry-Delaney (approved):
LGTM. Local tests passed.

amukkara (approved):

amukkara (commented):
does any CI test fail before this change?

yuxianq:
> does any CI test fail before this change?

@amukkara No, @Barry-Delaney get OOM issue when running `python examples/pytorch/quickstart_advanced.py --model_dir /llm-models/DeepSeek-R1/DeepSeek-R1-W4AFP8 --tp_size 2 --pp_size 2 --moe_ep_size 1 --moe_tp_size 2` on H200x4. After this PR, this test can pass. Our CI does not contain any DeepSeek-R1 test now.

PR #4462, opened 2025-05-20
Adapt Sentence Transformer model to BertForSeqClassification

caronzh03:
- Change BertPooler to use mean pooling
- Retrofit a ST bert model to HF's BertForSeqClassification
- Print logits during inference

### DIFF
--- a/examples/pytorch/sbert.py
+++ b/examples/pytorch/sbert.py
@@ -1,0 +1,56 @@
+from sentence_transformers import SentenceTransformer
+from transformers import BertForSequenceClassification, BertConfig
+import torch
+
+
+my_model_path = './model/my-custom-sbert'
+
+def load_sentence_transformer():
+    # Load the model
+    model = SentenceTransformer(my_model_path)
+    print(model)
+    # SentenceTransformer(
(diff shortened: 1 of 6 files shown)


triton_backend/all_models/llmapi/tensorrt_llm/1/model.yaml
nvrohanv, line 1: same as quickstart.py and model.py is this a file you want to change?

examples/pytorch/sbert.py
```diff
@@ -0,0 +1,56 @@
...
+    num_labels=192,  # Set number of classes
+    hidden_dropout_prob=bert_model.config.hidden_dropout_prob,
+    attention_probs_dropout_prob=bert_model.config.attention_probs_dropout_prob
+)
+
+# 4. Create new classification model
+classification_model = BertForSequenceClassification(config)
+
+# 5. Copy weights
+classification_model.bert.load_state_dict(bert_model.state_dict(), strict=False)
```
nvrohanv, line 42: This might be a stupid question but I'm a bit confused why we have to copy the weights when in line 31 we created a config that had the model path and then in line 39 created classification_model from that config.

nvrohanv (commented):
The approach is good, but new files need to be created for the examples instead of permanently changing things like quickstart.py.

PR #4471, opened 2025-05-20
chore: Deprecate autopp.

yuxianq:


Superjomn (approved):
LGTM

Are there any other examples/docs that we need to add the deprecation warning as well?

yuxianq:
> Are there any other examples/docs that we need to add the deprecation warning as well?

@Superjomn I also add deprecation warning to docs. As for examples, they will give the deprecation warning once the user run it with autopp, so we don't have to modify them.

yuxianq:
/bot skip --comment "This PR only contains deprecation warning, all single-gpu tests and most multi-gpu tests are passed."

litaotju (approved):

PR #4494, opened 2025-05-20
[TRTLLM-4783][feat] Mamba2 kernel updates for Nemotron-H

tomeras91:
## Description

Import 3rd party mamba2 kernels and use them for Nemotron-H, resulting in ~35% better throughput and latency without quality degradation.

### Kernel changes

Imported Kernels are based on Tri Dao's implementation with the following modifications:
1. **Convolution kernels** (based on [Dao-AILab/causal-conv1d](https://github.com/Dao-AILab/causal-conv1d)):
1.1. Prefill kernel: Add support for valren batching, prefill with initial state, paged conv state cache
1.2 Decode kernel: Validate support for paged conv state cache

2. **SSM kernels** (based on [state-spaces/mamba](https://github.com/state-spaces/mamba)):
2.1. Prefill kernel: Add support for prefill with initial state (validate varlen batching support).  No paged SSM state cache support
2.2 Decode kernel: Validate paged SSM state cache support.

These convolution kernels didn't exist in TRTLLM and are added in this PR. The mamba2 SSM kernels already existed in TRTLLM and are just modified where needed.

### Changes in modeling code

The new kernels are now used in the Mamba2 Mixer modeling code, along with the following changes:
1. Used varlen batching for prefill instead of for-loop over the batch dimension
2. For prefill, call convolution and SSM explicitly and get conv states from conv kernel so no need for a special call to get conv states
3. Remove redundant transposes and contiguous calls. For example, change the shape of the `ssm_states` in `MambaCacheManager` so no need to transpose them before and after kernel calls
4. Pass the gate activations (`z`) to the SSM decode kernel as well, similar to as it is in the prefill phase
5. Use the standard TRTLLM RMSNorm in Nemotron-H modeling, instead of the Transformers Engine one (removes dependency on Transformers Engine).

#### Refactor
Along with the more meaningful changes above, some refactoring was also done:
1. Organize the prefill and decode branches and remove common operations out of them to reduce code duplication
2. Remove unused code (`causal_conv1d_varlen_states`, `mamba_split_conv1d_scan_combined`, ...)
3. Rename `MambaMixer` -> `Mamba2Mixer`

### Benchmarks
The modifications in this PR **improved Nemotron-H throughput by ~40%** without sacrificing quality:

**MMLU results**
| Version | MMLU Score |
|---------|------------|
| Main | 69.88 |
| PR | 70.06 |

**Performance benchmarks**

Benchmark setting:
* Used genai-perf
* Done on a single RTX6000
* ISL / OSL = 223 / 140 (similar to ShareGPT statistics)
* Max throughput setting: --request-count 500 --concurrency 500 --warmup-request-count 10
* Min latency setting: --request-count 10 --concurrency 1 --warmup-request-count 1

&nbsp;
max throughput results:
| Metric | TRTLLM main | PR | Change [%] |
|--------|-------------|-----------|------------|
| Request latency [ms] | 56,433.40 | 39,367.06 | -30% |
| Mean TTFT [ms] | 45,568.17 | 31,832.92 | -30% |
| Mean ITL [ms] | 80.81 | 56.22 | -30% |
| Output Token Throughput [tokens/sec] | 677.17 | 956.11 | +41% |
| Output Token Throughput Per User [tokens/sec/user] | 12.57 | 18.04 | +44% |
| Request Throughput [req/sec] | 5.01 | 7.09 | +42% |

&nbsp;
min latency results:
| Metric | TRTLLM main | PR | Change [%] |
|--------|-------------|-----------|------------|
| Request latency [ms] | 4,088.62 | 3,587.03 | -12% |
| Mean TTFT [ms] | 86.69 | 83.27 | -4% |
| Mean ITL [ms] | 28.79 | 26.93 | -6% |
| Output Token Throughput [tokens/sec] | 34.24 | 36.52 | +7% |
| Output Token Throughput Per User [tokens/sec/user] | 34.74 | 37.13 | +7% |
| Request Throughput [req/sec] | 0.24 | 0.28 | +17% | 

Most of the throughput improvements (~25%) can be attributed to batched prefill (instead of for-loop). The rest is from the TRTLLM RMSNorm (instead of Transformer Engine) and the other modeling code changes. Min latency results don't show a large improvement since it has no batching and there was no significant change in the decode phase.

### Tests
Unittests for the kernels are also added in `test_causal_conv1d_op.py` and `test_mamba2_chunk_ss_update.py` under `tests/unittest/_torch/thop`. They were inspired by the TRTLLM mamba kernels unittests in `test_mamba_conv1d_op.py` and `test_selective_scan_op.py`, and check kernel correctness against a native pytorch implementation in various settings (prefill, decode, batching).


tensorrt_llm/_torch/models/modeling_nemotron_h.py
```diff
@@ -3,20 +3,16 @@
...
-    RMSNorm = None
 from transformers import AutoConfig, PretrainedConfig
 
 from ..attention_backend import AttentionMetadata
 from ..model_config import ModelConfig
 from ..modules.attention import Attention
 from ..modules.decoder_layer import DecoderLayer
 from ..modules.embedding import Embedding
-from ..modules.mamba.mixer import MambaMixer
+from ..modules.mamba.mixer import Mamba2Mixer
```
vegaluisjose, line 13: This failed to import for me, I changed to `from ..modules.mamba.mamba2_mixer import Mamba2Mixer`and it worked.

tomeras91: Good catch thanks! fixed

tests/unittest/_torch/thop/test_causal_conv1d_op.py
```diff
@@ -0,0 +1,161 @@
...
+import torch
+from parameterized import parameterized
+from utils.torch_ref import mamba_conv1d_ref
+from utils.util import unittest_name_func
+
+import tensorrt_llm
+from tensorrt_llm._torch.modules.mamba import PAD_SLOT_ID
+
+
+class TestFunctional(unittest.TestCase):
```
shaharmor98, line 15: Might be nit, or wrong, but if I'm not wrong they prefer `pytest` over unittest

tests/unittest/_torch/thop/test_causal_conv1d_op.py
```diff
@@ -0,0 +1,161 @@
...
+        # long sequence tests to cover the int overflow issue
+        list(
+            product([5376], [4], ['context'], ['float16', 'bfloat16'], [2],
+                    [131072], [False, True], [False, True])),
+        name_func=unittest_name_func)
+    def test_causal_conv1d(self, dim, dconv, req_type, dtype, batch_size,
+                           max_seq_len, remove_padding, apply_silu):
+        if max_seq_len == 131072:
+            total_gpu_mem = torch.cuda.get_device_properties(0).total_memory
+            if total_gpu_mem <= 33 * 1024**3:
```
shaharmor98, line 34: Why not pre-assigning this test in advance to a machine with sufficient amount of memory ?  
Also, you could use specific fixtures to skip if there isn't enough available memory

tests/unittest/_torch/thop/test_causal_conv1d_op.py
```diff
@@ -0,0 +1,161 @@
...
+                pytest.skip(
+                    "The long sequence test needs at least 33GB memory, skipping"
+                )
+
+        device = "cuda"
+        seq_len = max_seq_len if req_type == "context" else 1
+        mean = 0.0
+        std_dev = 1.0 if dtype == "float32" else 0.5
+        torch_dtype = getattr(torch, dtype)
```
shaharmor98, line 43: nit, they have trtllm util method for that.

shaharmor98 (commented):
I didn't have any major or meaningful comments, since I don't know much about the feature.  
All of my comments are nits / general questions

tests/unittest/_torch/thop/test_causal_conv1d_op.py
tomeras91: Fixed

tests/unittest/_torch/thop/test_causal_conv1d_op.py
tomeras91: after move to pytest, skipping directly in mark.parametrize.
I don't want to run the tests only on machines with sufficient memory since the only problematic test case is the one for long sequences. I want the normal sequence length tests to run on various machines.
As for why I skipped the test instead of excluding it in the yml files of specific machines - that's the pattern I saw elsewhere in the code.

vegaluisjose (approved):
Great work @tomeras91 !

I ran the changes and was able to confirm that quality was not degraded with these changes. Also, you did some performance benchmarking and things improve.

So I am good with merging this as soon as we can.

shaharmor98 (approved):

litaotju (approved):

PR #3439, opened 2025-04-10
perf: Eliminate the need for attention DP padding when possible

jinyangyuan-nvidia:
In the current implementation, all ranks are padded to have the same number of tokens after attention when attention DP is used. Therefore, the all-gather, MoE, and reduce-scatter operations potentially process more tokens than necessary, especially when request rate is relatively low.

This PR improves the performance by eliminating the need for attention DP padding when possible. The main idea is to modify `AllgatherOp` and `ReducescatterOp` to support different sizes across different ranks in `gather_dim` and `scatter_dim`, respectively.

Performance Test with DeepSeek-R1-FP8 on 1 node with 8 * H200 (use trtllm-bench throughput: https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples/models/core/deepseek_v3#example-multi-node-benchmark-on-gb200-slurm-cluster, the number of tested requests is changed to 15360)
- w/o this PR (the achieved concurrency is 159 requests per GPU)
```
===========================================================
= PERFORMANCE OVERVIEW 
===========================================================
Request Throughput (req/sec):                     4.9006
Total Output Throughput (tokens/sec):             10036.3364
Total Token Throughput (tokens/sec):              15054.5046
Total Latency (ms):                               3134338.9501
Average request latency (ms):                     578637.2961
Per User Output Throughput [w/ ctx] (tps/user):   3.7876
Per GPU Output Throughput (tps/gpu):              1254.5420
```
- w/ this PR (the achieved concurrency is increased to 168 requests per GPU under the same `kv_cache_free_gpu_mem_fraction`) 
```
===========================================================
= PERFORMANCE OVERVIEW 
===========================================================
Request Throughput (req/sec):                     5.1881
Total Output Throughput (tokens/sec):             10625.1556
Total Token Throughput (tokens/sec):              15937.7335
Total Latency (ms):                               2960641.8052
Average request latency (ms):                     550466.8627
Per User Output Throughput [w/ ctx] (tps/user):   3.9456
Per GPU Output Throughput (tps/gpu):              1328.1445
```

Performance Test with DeepSeek-R1-FP4 on 1 node with 8 * B200 (use trtllm-bench throughput: https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples/models/core/deepseek_v3#example-multi-node-benchmark-on-gb200-slurm-cluster, the number of tested requests is changed to 15360)
- w/o this PR (the achieved concurrency is 384 requests per GPU)
```
===========================================================
[...]

PR #4511, opened 2025-05-21
[fix] Fix Llama4 allgather error due to None tensor

jinyangyuan-nvidia:
Fix a bug of Llama4 when enabling attention DP (introduced by https://github.com/NVIDIA/TensorRT-LLM/pull/3439). None tensors are handled in a more general way. The test of attention DP for LLama4 is also added to unittest (waived by now due to an existing issue). Local tests show that the error caused by None tensor has been solved.


tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -817,19 +817,27 @@ def forward_chunk(
...
-                    self.mapping,
-                    dim=0,
-                    sizes=None if use_dp_padding else all_rank_num_tokens)
+            var_dict = {
+                'x': x,
+                'x_sf': x_sf,
+                'token_selected_experts': token_selected_experts,
+                'token_final_scales': token_final_scales,
+            }
+            key_list = [key for key, val in var_dict.items() if val is not None]
```
djns99, line 826: Can this use the builtin `filter()` method, the line below can also be done with a `map()`

tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -817,19 +817,27 @@ def forward_chunk(
...
+            var_list = [var_dict[key] for key in key_list]
+            var_list = allgather(
```
djns99, line 828: Would it be better to change the allgather function itself with this functionality. It seems like this may be a common pattern needing to filter out optional items

jinyangyuan-nvidia: Good point. The code has been modified accordingly.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -62,6 +63,24 @@ def get_output_info(input: torch.Tensor, dim: int) -> List[int]:
...
+) -> Tuple[List[torch.Tensor], List[bool]]:
+    func_valid = lambda x: x is not None
+    valid_list = list(map(func_valid, input_list))
+    input_list = list(filter(func_valid, input_list))
+    return input_list, valid_list
+
+
+def restore_full_output(output_list: List[torch.Tensor],
+                        valid_list: List[bool]) -> List[torch.Tensor]:
+    index_list = list(accumulate(map(lambda x: int(x), valid_list)))
```
djns99, line 77: nit: int is a function so`map(int, valid_list)` should work right?

djns99 (approved):
LGTM, one minor nit

tensorrt_llm/_torch/distributed/ops.py
jinyangyuan-nvidia: Thanks for the suggestion. Done.

litaotju (approved):

PR #4525, opened 2025-05-21
feat: better build_wheel.py venv handling

tongyuantongyu:
Some user experience improvements for `build_wheel.py` related to virtual environment handling

- Recognize manually created environments and skip creating another virtual environment.
  - Don't use venv at all if requested by user
- Detect common PyTorch environment issue and warn user.

### DIFF
--- a/scripts/build_wheel.py
+++ b/scripts/build_wheel.py
@@ -17,6 +17,8 @@
 import os
 import platform
 import sys
+import sysconfig
+import warnings
 from argparse import ArgumentParser
 from contextlib import contextmanager
 from functools import partial
@@ -26,6 +28,11 @@ from shutil import copy, copytree, rmtree
 from subprocess import DEVNULL, CalledProcessError, check_output, run
 from textwrap import dedent
 from typing import List
+
+try:
+    from packaging.requirements import Requirement
+except (ImportError, ModuleNotFoundError):
+    from pip._vendor.packaging.requirements import Requirement
 
 build_run = partial(run, shell=True, check=True)
 
@@ -67,21 +74,31 @@ def clear_folder(folder_path):
             os.remove(item_path)
 
 
-def setup_venv(project_dir: Path, requirements_file: Path):
-    """Creates/updates a venv and installs requirements.
+def sysconfig_scheme(override_vars=None):
+    # Backported 'venv' scheme from Python 3.11+
+    if os.name == 'nt':
+        scheme = {
+            'purelib': '{base}/Lib/site-packages',
+            'scripts': '{base}/Scripts',
+        }
+    else:
+        scheme = {
+            'purelib': '{base}/lib/python{py_version_short}/site-packages',
+            'scripts': '{base}/bin',
+        }
 
-    Args:
-        project_dir: The root directory of the project.
-        requirements_file: Path to the requirements file.
+    vars_ = sysconfig.get_config_vars()
+    if override_vars:
+        vars_.update(override_vars)
+    return {key: value.format(**vars_) for key, value in scheme.items()}
 
-    Returns:
-        Tuple[Path, Path]: Paths to the python and conan executables in the venv.
-    """
+
+def create_venv(project_dir: Path):
     py_major = sys.version_info.major
     py_minor = sys.version_info.minor
-    venv_dir = project_dir / f".venv-{py_major}.{py_minor}"
+    venv_prefix = project_dir / f".venv-{py_major}.{py_minor}"
     print(
-        f"-- Using virtual environment at: {venv_dir} (Python {py_major}.{py_minor})"
+        f"-- Using virtual environment at: {venv_prefix} (Python {py_major}.{py_minor})"
     )
 
     # Ensure compatible virtualenv version is installed (>=20.29.1, <22.0)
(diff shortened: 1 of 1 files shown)


scripts/build_wheel.py
```diff
@@ -67,46 +70,80 @@ def clear_folder(folder_path):
...
+    else:
+        venv_prefix = create_venv(project_dir)
+
     # Determine venv executable paths
-    scripts_dir = venv_dir / "bin"
-    venv_python = scripts_dir / "python"
+    scripts_dir = Path(
+        sysconfig.get_path("scripts", scheme="venv", vars={'base':
+                                                           venv_prefix}))
+    venv_python = venv_prefix / sys.executable.removeprefix(sys.prefix)[1:]
```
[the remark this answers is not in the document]
tongyuantongyu: That will be an inconsistent Python installation.

scripts/build_wheel.py
```diff
@@ -67,46 +70,92 @@ def clear_folder(folder_path):
...
     build_run(f'"{venv_python}" -m pip install -r "{requirements_file}"')
 
+    purelib_dir = Path(scheme["purelib"])
+    pytorch_package_dir = purelib_dir / "torch"
+    has_non_system_pytorch = venv_prefix != sys.base_prefix and pytorch_package_dir.exists(
+    )
+    if os.environ.get(
+            "NVIDIA_PYTORCH_VERSION") is not None and has_non_system_pytorch:
+        warnings.warn(
+            "You are using the NVIDIA PyTorch container, but not using the provided PyTorch installation. "
```
nv-guomingz, line 152: Refined by Owen:
```
Using the NVIDIA PyTorch container without leveraging the included PyTorch installation may lead to compatibility issues.

If you encounter any problems, please delete the environment at `{venv_prefix}` so that `build_wheel.py` can recreate the virtual environment.
```

scripts/build_wheel.py
```diff
@@ -694,6 +745,10 @@ def add_arguments(parser: ArgumentParser):
     parser.add_argument("--generate_fmha",
                         action="store_true",
                         help="Generate the FMHA cu files.")
+    parser.add_argument(
+        "--no-venv",
+        action="store_true",
+        help="Use current Python interpreter as is, don't create venv.")
```
nv-guomingz, line 751: Use the current Python interpreter without creating a virtual environment

scripts/build_wheel.py
```diff
@@ -67,46 +70,92 @@ def clear_folder(folder_path):
...
+    )
+    if os.environ.get(
+            "NVIDIA_PYTORCH_VERSION") is not None and has_non_system_pytorch:
+        warnings.warn(
+            "You are using the NVIDIA PyTorch container, but not using the provided PyTorch installation. "
+            "This will very likely cause issues. "
+            f"If anything goes wrong, please remove the environment at {venv_prefix} "
+            "to let build_wheel.py rebuild virtual environment.")
+        print("^^^^^^^^^^ IMPORTANT WARNING ^^^^^^^^^^", file=sys.stderr)
+        time.sleep(15)
```
nv-guomingz, line 157: Could we use some more elegant behaviour  instead of time.sleep(15)?

tburt-nv: Instead of prompting the user to consider taking action, can we just run `pip remove torch` in this case (before running `pip install -r "{requirements_file}"`)?

tongyuantongyu: > Could we use some more elegant behaviour instead of time.sleep(15)?

Do you have any idea? `build_wheel.py` quickly outputs many messages around this line, so this warning is easily missed by user if we don't pause here.

> can we just run pip remove torch inside the venv in this case

The situation is more complicated than just a bad `torch` package.

This situation is likely caused by user running `build_wheel.py` on outdated Docker image. In this case pip will install dozens of packages into venv trying to override system-provided version, all of which can have compatibility issue. It's not worth the effort to fix the environment, and the only reasonable choice is to rebuild it. But I'm hesitant to do it automatically since the venv may be manually created.

nv-guomingz: > Do you have any idea? `build_wheel.py` quickly outputs many messages around this line, so this warning is easily missed by user if we don't pause here.

 If our intention is to ask the user to uninstall the torch, how about we use `input` function to pause the execution ?

tburt-nv: If the root cause is likely to be an outdated container image, I suggest modifying the comment to instruct users to obtain a newer container image (either by building one or pulling one) in addition to recreating the venv. If the developer is still using an outdated image when they encounter this error, won't the script just re-install pytorch inside the venv the next time the user runs build_wheel.py?

Comparing the image tag name won't work for NVIDIA-external developers. I think we can detect if the image is outdated by running `pip freeze | grep torch` (or some pythonic equivalent) and comparing it to the `requirements.txt` file.

The venv is an internal implementation detail of this single script; I think the script should be able to confidently maintain the venv however it chooses.

I suppose I can make a follow-up PR to attempt automatic recovery of the venv, although mentioning the problem of outdated container images in the warning would be helpful in this PR.

tongyuantongyu: > The venv is an internal implementation detail of this single script; I think the script should be able to confidently maintain the venv however it chooses.

I'd like `build_wheel.py` to not creating another venv given I already prepared a venv dedicated for TensorRT-LLM development.

> automatic recovery of the venv

I have some customizations in my venv, so I don't want it to be automatically reset.

> mentioning the problem of outdated container images

I added the detection of incompatible container image.

dcampora (approved):

PR #4558, opened 2025-05-22
[TRTLLM-3456] Speculation: Draft Target in new FW

IzzyPutterman:


examples/pytorch/quickstart_advanced.py
```diff
@@ -103,6 +103,7 @@ def add_llm_args(parser):
     parser.add_argument('--spec_decode_algo', type=str, default=None)
     parser.add_argument('--spec_decode_nextn', type=int, default=1)
     parser.add_argument('--eagle_model_dir', type=str, default=None)
+    parser.add_argument('--draft_model_dir', type=str, default=None)
```
mikeiovine, line 106: Can we merge this with `eagle_model_dir`?

tensorrt_llm/_torch/speculative/draft_target.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __post_init__(self):
+        pass
+
+    def prepare(self):
+        pass
+
+
+
+class DraftTargetSampler(TorchSampler):
```
mikeiovine, line 43: Do we need a custom decoder? It looks like we can just reuse `TorchSampler` for this one.

tensorrt_llm/_torch/speculative/interface.py
```diff
@@ -52,7 +59,7 @@ def extend_ctx(self, attention_backend: AttentionBackend):
         # Fixme: only trtllm attention backend supports eagle3 generation-phase kernels on blackwell.
         return (self.is_eagle3()
                 and not (isinstance(attention_backend, TrtllmAttention)
-                         and get_sm_version() == 100)) or self.is_ngram()
+                         and get_sm_version() == 100)) or self.is_ngram() or self.is_draft_target()
```
mikeiovine, line 62:
```python
        return ((self.is_eagle3() or self.is_draft_target())
                and not (isinstance(attention_backend, TrtllmAttention)
                         and get_sm_version() == 100)) or self.is_ngram() 
```
Should be treated the same as eagle 3

mikeiovine (commented):
Looks good. Let's add a test?

mikeiovine:
By the way, I realized that this was a bit too specific to eagle 3: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/py_executor.py#L1817

You don't have to re-compute KV cache for verified tokens in this case; that's only a thing in MTP/eagle. We can make it optional. No need to fix this now since it's a performance thing, won't hurt correctness.

tensorrt_llm/_torch/speculative/draft_target.py
IzzyPutterman: NGram used the same setup, and it seems to work :)

tensorrt_llm/_torch/speculative/interface.py
IzzyPutterman: Adjusted

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -62,7 +62,7 @@ def create_py_executor(executor_config: ExecutorConfig,
     dist = MPIDist(mapping=mapping)
 
     spec_config = executor_config.speculative_config
-    has_draft_model_engine = isinstance(spec_config, Eagle3Config)
+    has_draft_model_engine = spec_config.spec_dec_mode.has_pytorch_model()
     has_ngram_drafter = isinstance(spec_config, NGramConfig)
```
Funatiq, line 66:
```suggestion
    has_draft_model_engine = spec_config.spec_dec_mode.has_pytorch_model() if spec_config is not None else False
```

IzzyPutterman:
Some variance in output is expected (I think) specifically when a draft token is from the prefill portion.

tensorrt_llm/_torch/speculative/draft_target.py
mikeiovine: It will definitely work, I just mean that this is duplicated code that is not necessary

examples/pytorch/quickstart_advanced.py
IzzyPutterman: Made it a flag with two options so people who ran command before can still use the same script.

tensorrt_llm/_torch/speculative/interface.py
```diff
@@ -59,6 +63,9 @@ def need_load_draft_weights(self):
     def has_spec_decoder(self):
         return self.is_mtp() or self.is_eagle3() or self.is_eagle3_one_model()
 
+    def has_pytorch_model(self):
+        return self.is_eagle3() or self.is_draft_target()
```
mikeiovine, line 67: We already have `has_draft_model`; we can use that function instead

IzzyPutterman:
Depends on these fixes: https://github.com/NVIDIA/TensorRT-LLM/pull/4807

mikeiovine (approved):

hlu1 (approved):

IzzyPutterman:
/bot skip "No major changes since last CI pass, just rebase"

IzzyPutterman:
/bot skip --comment "No major changes since last CI pass, just rebase"

PR #4571, opened 2025-05-22
Chore: clean up _gather_dp_requests_num method of PyExecutor

QiJune:
This PR moves the logic of `_gather_dp_requests_num` into `fetch_requests`.

Before this PR, `_gather_dp_requests_num` is called at the end of `executor_loop`, and then, `fetch_requests` is called at the beginning of each iteration. Then, why not just fuse logic of `gather_dp_requests_num` into `fetch_requests.` After this clean up, codes become more cohensive.


yuxianq:
@QiJune Can you add a brief description about why we don't need to gather dp request number anymore?

QiJune:
> @QiJune Can you add a brief description about why we don't need to gather dp request number anymore?

@yuxianq  Please see here: https://github.com/NVIDIA/TensorRT-LLM/pull/4571#issue-3082944361

yuxianq (approved):

PR #4583, opened 2025-05-22
None: Make add_special_tokens to false for completions API

Pernekhan:
# PR title

[fix] Change default for add_special_tokens to false, since the completions API shouldn't expect to modify the prompt.

## Description

Change the default of add_special_tokens. It made some confusions, especially when running with speculative decoding which were trained with single bos tokens. The extra bos token messes up the rest of the output.


LinPoly:
> It made some confusions, especially when running with speculative decoding which were trained with single bos tokens. The extra bos token messes up the rest of the output.

So you have prepended a <BOS> token to the prompt before sending to the server, correct?

Pernekhan:
> > It made some confusions, especially when running with speculative decoding which were trained with single bos tokens. The extra bos token messes up the rest of the output.
> 
> So you have prepended a BOS token to the prompt before sending to the server, correct?

This is how it's usually done. For the completions API the prompt is already formatted, and the bos token is included most of the time. That's why I'm proposing the default to not prepend anything.

LinPoly:
Thanks for explanation, I think it is reasonable to do, could you please also change [this line](https://github.com/NVIDIA/TensorRT-LLM/blob/dd290635388199c2a5d21396d3c767acd9cf280e/tests/unittest/llmapi/apps/_test_openai_completions.py#L65) to `prompt_tokens = 5` to pass the CI?

LinPoly (approved):
Can  you please change the completion test accordingly?

Pernekhan:
@LinPoly thanks for pointing out the unit-test fix. I've just fixed it.

LinPoly:
@Pernekhan CI fails due to code style checking, could you please follow [the instructions here](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CONTRIBUTING.md#coding-style) to fix? And add @kaiyux and @nv-guomingz for comments.

LinPoly:
> I'm not sure what code style it's violating. Running pre-commit hook seems to be working fine.

I tried `pre-commit run --files tensorrt_llm/serve/openai_protocol.py` and got a yapf check failure, and the following change was suggested

image

Could you pls run the command and check again? If it doesn't work, I can create a new branch containing the fixing commit, you can cherry pick it for style checking.

PR #4596, opened 2025-05-22
[JIRA-5226219][fix] Fix Bug in KV cache manager

thorjohnsen:
## Description

The reported symptom is that a block which isn't a true leaf block (no descendants) is attempted freed. This can happen for instance when a suggested primary block has descendants in secondary memory. The preferred way to handle this is to offload the suggested primary block, but offloading isn't always possible. In the case that offloading isn't possible, the new code will explicitly free all descendants before attempting to claim suggested leaf block.

## Test Coverage

I added a unit test to kvCacheManagerTest suite that triggers the condition described above. Without the new code that explicitly frees descendants when offloading isn't possible, the new unit test will fail.


dongxuy04 (approved):
LGTM, thanks.

WeiHaocheng (approved):

PR #4614, opened 2025-05-23
Chore: refine shutdown signal of PyExecutor

QiJune:


tensorrt_llm/_torch/pyexecutor/py_executor.py
yuxianq, line 1: @QiJune LGTM. Can we use a data class instead of tuple for request queue item, e.g.,
```
@dataclass
class RequestQueueItem:
    id: int
    request: Optional[ExecutorRequest] = None
    query_token_ids: Optional[Union[torch.Tensor, np.ndarray, list]] = None

```

QiJune: @yuxianq  It's a good idea, will do it in the next PRs.

yuxianq (approved):

PR #4618, opened 2025-05-23
Reorgnize code structure in FusedMoe

HuiGao-NV:


tensorrt_llm/_torch/models/modeling_qwen3_moe.py
```diff
@@ -408,7 +409,7 @@ def filter_weights(prefix, weights: Dict):
                     module.load_weights(weights=module_weights)
                 else:
                     module_weights = filter_weights(name, weights)
-                    if isinstance(module, FusedMoE):
+                    if issubclass(type(module), FusedMoE):
```
yuxianq, line 412: Why we use issubclass instead of isinstance?

tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -294,6 +293,42 @@ class MoEWeightLoadingMode(Enum):
 
 
 class FusedMoE(nn.Module):
```
yuxianq, line 295: Can we make `FusedMoE` an abstract class, so that `create_weights/load_weights/forward` can be marked as abstract method instead of using `NotImplementedError`?

tensorrt_llm/_torch/modules/fused_moe.py
```diff
@@ -294,6 +293,42 @@ class MoEWeightLoadingMode(Enum):
...
+
+    def __init__(self):
```
yuxianq, line 297: Unnecessary to add `__init__` here.

HuiGao-NV:
Hao will take the effort to refactor FusedMoE

PR #4630, opened 2025-05-23
[TRTLLM-4971]: Use safe deserialization in ParallelConfig

yibinl-nvidia:
# Use safe deserialization in ParallelConfig

## Description

Remove unsafe pickle.load usage in ParallelConfig by using a safe deserialization method. During the tests, I realized that the current location of `serialization.py` under executor directory will cause a circular import of the files that import `ParallelConfig`, so I move the `serialization.py` outside of executor directory.

## Test Coverage

Added a new test `test_parallel_config_serialization` to check the correctness of safe serialization / deserialization of `ParallelConfig`.

### DIFF
--- a/tensorrt_llm/auto_parallel/parallelization.py
+++ b/tensorrt_llm/auto_parallel/parallelization.py
@@ -11,7 +11,7 @@ import tensorrt as trt
 import torch
 from filelock import FileLock
 
-import tensorrt_llm.serialization as serialization
+from tensorrt_llm import serialization
 from tensorrt_llm._utils import (str_dtype_to_trt, trt_dtype_to_np,
                                  trt_dtype_to_torch)
 from tensorrt_llm.functional import AllReduceParams, create_allreduce_plugin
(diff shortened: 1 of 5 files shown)


kaiyux:
@yuxianq Since it's related to auto parallel, can you help take a look?

tensorrt_llm/executor/ipc.py
```diff
@@ -9,7 +9,7 @@
 import zmq
 import zmq.asyncio
 
-import tensorrt_llm.executor.serialization as serialization
+import tensorrt_llm.serialization as serialization
```
yuxianq, line 12: Can we use `from tensorrt_llm import serialization` instead?

tensorrt_llm/serialization.py
```diff
@@ -98,6 +98,12 @@
     "torch.storage": ["_load_from_bytes"],
 }
 
+BASE_PARALLEL_CLASSES = {
```
yuxianq, line 101: Since `BASE_PARALLEL_CLASSES` is only used by `tensorrt_llm/auto_parallel/parallelization.py`, can we move it to the same file?

yibinl-nvidia: fixed

yuxianq (approved):

tensorrt_llm/auto_parallel/parallelization.py
```diff
@@ -39,6 +39,13 @@
 
 default_int_dtype = trt.int64
 
+# These dataclasses are used in ParallelConfig serialization. If there are other classes need to be serialized, please add to this list.
+base_parallelConfig_serialization_classes = {
```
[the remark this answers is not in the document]
yibinl-nvidia: fixed

yibinl-nvidia:
@kaiyux @Superjomn This MR relocates `serialization.py` to a better location and apply safe deserialization to ParallelConfig. I would expect very few updates to the BASE_AUTOPP_CLASSES (unlike IPC) so we should be safe to do this change.

PR #4642, opened 2025-05-25
Add missing serialization classes

Tabrizian:
# Add missing serialization classes

When using trtllm-launch-api I observed some failures that were fixed by adding these imports.


Superjomn (approved):
LGTM

PR #4649, opened 2025-05-26
Chore: introduce RequestQueueItem class instead of using tuple

QiJune:


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1262,15 +1270,16 @@ def _fetch_new_requests(self):
             ]
             heapq.heapify(all_ranks_new_requests_heap)
             new_requests = sorted(new_requests,
-                                  key=lambda x: len(x[1].input_token_ids),
+                                  key=lambda x: len(x.request.input_token_ids),
                                   reverse=True)
-            for request in new_requests:
+            for request_item in new_requests:
```
yuxianq, line 1275: Can use `req_item` instead of `request_item` like all other places.

yuxianq (approved):

tensorrt_llm/_torch/pyexecutor/py_executor.py
QiJune: Nice catch, will refine it in the next PRs.

PR #4689, opened 2025-05-27
Chore: fuse _merge_requests method into _fetch_new_requests method

QiJune:


yuxianq (approved):
LGTM, but still need @lfr-0531 or @dongxuy04 's confirm.

lfr-0531 (approved):
LGTM~

PR #4765, opened 2025-05-29
feat: add heuristics for checkpoint files prefetching.

yuxianq:
When the memory consumption of prefetching is less than 90% of available CPU memory, we enable prefetching.
This PR also uses local rank instead of global rank to deal with multi-node cases.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -21,7 +21,8 @@ import torch._dynamo.config
 import tensorrt_llm.bindings.internal.userbuffers as ub
 from tensorrt_llm._torch.pyexecutor.sampler import SampleStateTensors
 from tensorrt_llm._torch.speculative.mtp import SampleStateTensorsMTP
-from tensorrt_llm._utils import (is_trace_enabled, nvtx_range, release_gc,
+from tensorrt_llm._utils import (is_trace_enabled, local_mpi_rank,
+                                 local_mpi_size, nvtx_range, release_gc,
                                  torch_dtype_to_str, trace_func)
 from tensorrt_llm.bindings.executor import GuidedDecodingConfig
 from tensorrt_llm.logger import logger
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -129,6 +131,14 @@ def validate_and_set_kv_cache_quant(model_config: ModelConfig,
     model_config.quant_config.kv_cache_quant_algo = mapped_pyt_quant
 
 
+def _prefetch_one_file(file_name, rank):
```
nvpohanh, line 134: don't know what the convention is, but personally I only put `_` prefix for functions that I defined inside the function. Maybe we should remove the `_` prefix if we pull it to the top level?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -137,33 +147,32 @@ def prefetch_files(file_names: List[str], mapping: Mapping):
...
-def load_weights(checkpoint_dir: str, mapping: Mapping):
+def load_weights(
+    checkpoint_dir: str,
+    mapping: Mapping,
+):
     weights = {}
     weight_files = glob.glob(f"{checkpoint_dir}/*.safetensors")
     if weight_files:
-        prefetch_files(weight_files, mapping)
+        prefetch_size = sum(os.path.getsize(file) for file in weight_files)
```
nvpohanh, line 169: Add a comment explaining why this heuristic makes sense. Thanks!

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -137,33 +147,32 @@ def prefetch_files(file_names: List[str], mapping: Mapping):
...
+        is_prefetch = prefetch_size < psutil.virtual_memory().available * 0.9
```
nvpohanh, line 170: nit: `is_prefetch` -> `enable_prefetch`

nvpohanh (approved):

tensorrt_llm/_torch/pyexecutor/model_engine.py
yuxianq: We still need the `_` prefix for private functions in the top level to avoid other people importing them accidently.

tensorrt_llm/_torch/pyexecutor/model_engine.py
yuxianq: Added.

tensorrt_llm/_torch/pyexecutor/model_engine.py
yuxianq: Updated.

dongxuy04 (approved):

PR #4843, opened 2025-06-02
[nvbug/5314469][feat] Include the executor's max batch size in CUDA g…

mikeiovine:
…raph batch size list

## Description

See https://nvbugspro.nvidia.com/bug/5314469 for the full context. In short:

1. The perf team has a hardcoded list of CUDA graph batch sizes. In these tests, CUDA graph padding is turned on.
2. They separately use a `trtllm-bench` heuristic to tune `max_batch_size` and `max_num_tokens`.
3. Suppose a batch size `B` is included in the batch size list but `max_batch_size < B`. The `ModelEngine` cannot safely pad to to `B` because memory estimation will be done with `max_batch_size`. `B` will thus be excluded from the CUDA graph batch size list, and any requests running at the max batch size will not be able to reap the benefits of CUDA graphs.

In this case, I think it makes sense to instead allow padding to `max_batch_size`. The logic is that if the user is OK padding to `B`, they will also be OK with padding to a value less than `B`; the performance is strictly better!

## Test Coverage

Existing CUDA graph tests.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -402,11 +402,26 @@ def __init__(
...
+                and self.pytorch_backend_config.cuda_graph_batch_sizes[-1]
+                > max_cuda_graph_bs):
+            logger.warning(
+                "CUDA graph padding is enabled, but the maximum CUDA graph "
+                f"batch size ({self.pytorch_backend_config.cuda_graph_batch_sizes[-1]}) "
+                f"is larger than the executor's max batch size ({max_cuda_graph_bs}). "
+                f"We will pad batches to {max_cuda_graph_bs}.")
+            self._cuda_graph_batch_sizes.append(max_cuda_graph_bs)
+
+        self._max_cuda_graph_batch_size = (self._cuda_graph_batch_sizes[-1] if
```
hlu1, line 423: Can you move line 406 to here to a separate function and call it here?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -402,11 +402,26 @@ def __init__(
...
             bs for bs in pytorch_backend_config.cuda_graph_batch_sizes
-            if bs <= self.max_num_tokens and bs <= self.batch_size
+            if bs <= max_cuda_graph_bs
         ]
-        self._max_cuda_graph_batch_size = self._cuda_graph_batch_sizes[-1]
+
+        if (self._cuda_graph_padding_enabled and self._cuda_graph_batch_sizes
+                and self._cuda_graph_batch_sizes[-1] < max_cuda_graph_bs
+                and self.pytorch_backend_config.cuda_graph_batch_sizes[-1]
+                > max_cuda_graph_bs):
```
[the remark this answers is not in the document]
mikeiovine: Oops, this part is redundant and can be deleted:
```python
self._cuda_graph_batch_sizes self._cuda_graph_batch_sizes[-1] > max_cuda_graph_bs
```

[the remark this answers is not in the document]
mikeiovine: You could write it like this if you prefer more imperative-style code:

```python
for i, bs in enumerate(self.pytorch_backend_config.cuda_graph_batch_sizes):  # Sorted
    if bs <= max_cuda_graph_bs:
        self._cuda_graph_batch_sizes.append(bs)
    else:
        if self._cuda_graph_padding_enabled and i > 0 and self._cuda_graph_batch_sizes[i - 1] != max_cuda_graph_bs:
            self._cuda_graph_batch_sizes.append(max_cuda_graph_bs)
        break
```
The last condition in our `if` statement is implicit here because it's inside the for loop. It doesn't seem much cleaner to me though.

Another alternative is to always add the `max_cuda_graph_bs` to the `cuda_graph_batch_sizes` list when padding is enabled. I don't think this is a good idea though. Padding to smaller values than the user asked for is always OK, that can only help performance. Padding to larger values implicitly can hurt performance.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -288,6 +288,37 @@ def get_rank_model_storage(model):
...
+        if bs <= max_cuda_graph_bs:
+            result.append(bs)
+        else:
+            # One extra special case for padding. The user gave us at least
+            # one batch size to pad to which is larger than the executor's max
+            # batch size. In this case, padding to max_cuda_graph_bs is acceptable. The logic
+            # is that if the user is OK padding to a batch size B, they should also
+            # be OK with padding to some size B' < B since the performance will generally
+            # just be better in the smaller case.
+            if padding_enabled and (i == 0
```
mikeiovine, line 309: I don't really have a strong opinion on this being cleaner than the last version, but it's updated to the imperative style now.

hlu1 (approved):

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -288,6 +288,37 @@ def get_rank_model_storage(model):
     return total_bytes
 
 
+def _filter_cuda_graph_batch_sizes(cuda_graph_batch_sizes: list[int],
```
hlu1, line 291: A simple unit test would be nice.

mikeiovine: I think it's already implicitly covered by a bunch of tests that enable padding. We can test the function in a standalone way though.

PR #4848, opened 2025-06-03
enh: Enable trtllm-bench to run LoRA PyT flow

venkywonka:
# Description

This PR does some plumbing required to easily enable LoRA pathway in PyT backend from trtllm-bench itself, given an appropriate `extra-llm-api-config.yml` and `synthetic_data.json`.

## Checklist
- [x] prepare_dataset changes 
- [x] trtllm-bench and internal plumbing to accept lora metadata, locally verify it works.
- [x] unittest for prepare_dataset
- [x] documentation 
- [ ] e2e trtllm-bench with lora on pyt test (this is will be added as part of a second PR that requires some `test_perf.py`-level non-trivial changes).

## More deets

The `extra-llm-api-config.json` should include a section for lora config specifying the lora information. 
That should look something like this: 
```yaml
...
lora_config:
  lora_dir:
  - path/to/loras/0
  - /path/to/loras/1
  max_lora_rank: 64
  lora_target_modules:
  - attn_q
  - attn_k
  - attn_v
  trtllm_modules_to_hf_modules:
    attn_q: q_proj
    attn_k: k_proj
    attn_v: v_proj
...
```

The PR also includes decessary changes to `tensorrt_llm/benchmarks/cpp/prepare_dataset.py` needed to include generation requests including lora request details.
The below is an example command to generate the appropriate `synthetic_data.json` that incorporates random lora requests involving loras with task ids 0,1:

```bash
python3 /path/to/tensorrt_llm/benchmarks/cpp/prepare_dataset.py \
  --stdout \
  --rand-task-id 0 1 \
  --tokenizer=/path/to/llm-models/llama-3.1-model/Llama-3.1-8B-Instruct-FP8 \
  --lora-dir=/path/to/loras \ # assumes that the different task-specific loras are numbered subdirs (0/, 1/,..etc.) 
  token-norm-dist \
  --num-requests=100 \
  --input-mean=128 \
  --output-mean=128 \
  --input-stdev=16 \
  --output-stdev=24 \
  > path/to/synthetic_data.json
```

NOTE: The prepare_dataset.py script takes the 'parent' lora directory containing each task-specific loras, and assumes thes task-specific lora subdirectories, with their names being their respective lora task ID. 

below is how a sample request should look like when the above command is run: 
```json
{
  "task_id": 0,
  "input_ids": [
    3452, 88226, 102415, .... 
  ],
  "output_tokens": ...,
  "lora_request": {
    "lora_name": "lora_0",
    "lora_int_id": 0,
    "lora_path": "/path/to/loras/0"
  }
}
```

### DIFF
--- a/benchmarks/cpp/utils/utils.py
+++ b/benchmarks/cpp/utils/utils.py
@@ -70,13 +70,28 @@ def multimodal_dataset_dump(multimodal_texts, multimodal_image_paths,
         json.dump(workload.model_dump(), f)
 
 
-def print_text_dataset(input_ids, output_lens):
+def print_text_dataset(input_ids, output_lens, task_ids=None, lora_config=None):
     for i, input_tokens in enumerate(input_ids):
         d = {
             "task_id": i,
             "input_ids": input_tokens,
             "output_tokens": output_lens[i]
         }
+
+        # Add LoRA request if task_ids indicate LoRA usage
+        if task_ids is not None and lora_config is not None:
+            task_id = task_ids[i]
+            if task_id != -1:  # -1 means no LoRA
+                d["lora_request"] = {
+                    "lora_name":
+                    f"lora_{task_id}",
+                    "lora_int_id":
+                    task_id,
+                    "lora_path":
+                    os.path.join(lora_config.get("lora_dir", "loras"),
+                                 str(task_id))
+                }
+
         print(json.dumps(d, separators=(',', ':'), ensure_ascii=False))
 
 
--- a/docs/source/performance/perf-benchmarking.md
+++ b/docs/source/performance/perf-benchmarking.md
@@ -475,6 +475,84 @@ Total Latency (ms):             18563.6825
 
 ```
 
+#### Benchmarking with LoRA Adapters in PyTorch workflow
+
+The PyTorch workflow supports benchmarking with LoRA (Low-Rank Adaptation) adapters. This requires preparing a dataset with LoRA metadata and configuring the LoRA settings.
+
+**Preparing LoRA Dataset**
+
+Use `prepare_dataset.py` with LoRA-specific options to generate requests with LoRA metadata:
+
+```shell
+python3 benchmarks/cpp/prepare_dataset.py \
+  --stdout \
+  --rand-task-id 0 1 \
+  --tokenizer /path/to/tokenizer \
+  --lora-dir /path/to/loras \
+  token-norm-dist \
+  --num-requests 100 \
+  --input-mean 128 \
+  --output-mean 128 \
+  --input-stdev 16 \
+  --output-stdev 24 \
+  > synthetic_lora_data.json
+```
+
+Key LoRA options:
+- `--lora-dir`: Parent directory containing LoRA adapter subdirectories named by their task IDs (e.g., `0/`, `1/`, etc.)
+- `--rand-task-id`: Range of LoRA task IDs to randomly assign to requests
+- `--task-id`: Fixed LoRA task ID for all requests (alternative to `--rand-task-id`)
+
+The generated dataset will include LoRA request metadata. Below is an example of a single such request data entry:
+
+```json
+{
+  "task_id": 0,
+  "input_ids": [3452, 88226, 102415, ...],
+  "output_tokens": 152,
+  "lora_request": {
+    "lora_name": "lora_0",
+    "lora_int_id": 0,
+    "lora_path": "/path/to/loras/0"
+  }
+}
+```
+
+**LoRA Configuration**
+
+Create an `extra-llm-api-options.yaml` file with LoRA configuration:
+
+```yaml
+lora_config:
+  lora_dir:
+    - /path/to/loras/0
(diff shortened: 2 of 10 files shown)


docs/source/performance/perf-benchmarking.md
```diff
@@ -475,6 +475,84 @@ Total Latency (ms):             18563.6825
...
+
+Create an `extra-llm-api-options.yaml` file with LoRA configuration:
+
+```yaml
+lora_config:
+  lora_dir:
+    - /path/to/loras/0
+    - /path/to/loras/1
+  max_lora_rank: 64
+  lora_target_modules:
```
shaharmor98, line 531: Although I'm not an advocate for this option, we also allow "default" `lora_target_modules` which will ultimately lead to choosing all modules. To emphasize, setting the `lora_target_modules` is the preferred option.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1259,7 +1259,7 @@ def _ensure_lora_config_consistency(self):
                     "lora_dir is empty, so custom embedding or lm head will not be applied."
                 )
 
-        if self.enable_lora and self.lora_config is not None and self.backend == 'pytorch':
```
shaharmor98, line 1262: Why did you remove the `self.enable_lora`?  
At the moment we left it untouched because we haven't fully deprecated it yet

shaharmor98: You're completely right.  
The `self.enable_lora` flag pertains the TRT backend, not the PyT backend. At the end of the day, as you can see, this flag is redundant in our case, but since we only now introduced the changes to the PyT backend, we didn't want to deprecate TRT flags immediately. We had some previous discussions with `llmapi` folks, and this was the final choice we agreed on for now.  
So to sum up, this is some of a transient phase, which will be fixed in the short-term, but I don't think it should be part of your PR since it requires additional API considerations I don't think you need to tackle here.

venkywonka: cool! i'm okay with that, thanks for the context @shaharmor98 :D

tensorrt_llm/bench/utils/data.py
```diff
@@ -7,6 +7,7 @@
 from tensorrt_llm.bench.dataclasses.general import (DatasetMetadata,
                                                     InferenceRequest)
 from tensorrt_llm.bench.dataclasses.statistics import PercentileStats
+from tensorrt_llm.executor.request import LoRARequest
```
FrankD412, line 10: I have to check, but we've moved to using the LLM API Python equivalents to these classes. Is there a Python equivalent to this class?

benchmarks/cpp/utils/utils.py
```diff
@@ -70,13 +70,28 @@ def multimodal_dataset_dump(multimodal_texts, multimodal_image_paths,
...
         d = {
             "task_id": i,
             "input_ids": input_tokens,
             "output_tokens": output_lens[i]
         }
+
+        # Add LoRA request if task_ids indicate LoRA usage
+        if task_ids is not None and lora_config is not None:
+            task_id = task_ids[i]
+            if task_id != -1:  # -1 means no LoRA
```
FrankD412, line 84: Just to check -- you can use the random task id option without LoRA right? Wouldn't that break this logic?

tensorrt_llm/bench/utils/data.py
venkywonka: hmm, as per my understanding the same `LoraRequest` class is used in both LLM API and executor API 🤔

tensorrt_llm/bench/utils/data.py
```diff
@@ -7,7 +7,35 @@
...
+        inputs = default_image_loader(prompts, media, image_data_format)
+    elif modality == "video":
+        inputs = default_video_loader(prompts, media, image_data_format,
+                                      num_frames)
+    else:
+        raise ValueError(f"Unsupported modality: {modality}")
+
+    inputs = formatter(model_dir, inputs)
+
+    return inputs
```
venkywonka, line 38: need to remove this, unrelated to my PR but somehow crept in while rebasing

benchmarks/cpp/utils/utils.py
venkywonka: i don't think the task ids or rand task ids serve any other purpose... so i don't think it will get used for anything else. if they are used without any lora directory specified, then that's probably not intended - but thank you for the catch; currently that isn't handled properly...
I will either error out or basically send a warning and pretend it isn't a lora situation 👍🏾

tensorrt_llm/bench/utils/data.py
FrankD412: Yeah -- looks like there isn't a Python equivalent to this class yet. @Superjomn for vis.

venkywonka:
closing this PR as #5130 is merged 🥳

venkywonka:
closing this PR as #5130 is merged 🥳

PR #4851, opened 2025-06-03
Replace memset with data initialization within kernels

ChristinaZ:
# Remove cudaMemsetAsync()  to avoid the gap within the CUDA graph

## Description

@bobboli's profiling results show that using cudaMemsetAsync() introduces a timing gap within the CUDA graph execution. In this PR, we try to remove this gap by replacing this API with data initialization within kernels.

By the way, the previous setting for `routingData.mNormTopkProb` is not correct. It is not used in the current model. Fix the setting within this PR to pass the unit tests. 

## Test Coverage
- Simple correctness check
> pytest -k test_moe_fp4 tests/unittest/_torch/thop/test_moe.py
- To observe the gap optimization, we need to enable CUDA Graph

### DIFF
--- a/cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/RoutingKernel.cu
+++ b/cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/RoutingKernel.cu
@@ -2992,6 +2992,18 @@ __global__ void __cluster_dims__(NumBlocksPerCluster, 1, 1) __launch_bounds__(Nu
         cudaGridDependencySynchronize();
     }
 
+    // initialize the mPtrPermutedIdxToTokenIdx
+    if (params.mPtrPermutedIdxToTokenIdx != nullptr)
+    {
+        int32_t permIdxToTokenIdxNum
+            = (params.mNumTokens * NumTopExperts + (params.mNumExperts << params.mPaddingLog2) - params.mNumExperts);
+        for (int32_t i = clusterThreadIdx; i < permIdxToTokenIdxNum; i += NumThreadsPerCluster)
+        {
+            params.mPtrPermutedIdxToTokenIdx[i] = -1;
+        }
+        __threadfence();
+    }
+
     if (params.mPtrScores != nullptr)
     {
         // in this case, each warp represents a token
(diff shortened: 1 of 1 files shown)


cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/runner.cu
```diff
@@ -159,7 +159,7 @@ void Runner::run(void* routingLogits, void* routingBias, int32_t numTokens, int3
         // routingData.mDtypeElt = dtypeElt; // no-op for now as hidden_state is not input
         routingData.mUsePdl = true;
         routingData.mDoSoftmaxBeforeTopK = routingMethodType == RoutingMethodType::Qwen3;
-        routingData.mNormTopkProb = routingMethodType == RoutingMethodType::Renormalize;
+        routingData.mNormTopkProb = routingMethodType == RoutingMethodType::Qwen3;
```
rosenrodt, line 162: Good catch. Thanks!

cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/RoutingKernel.cu
```diff
@@ -2988,6 +2988,18 @@ __global__ void __cluster_dims__(NumBlocksPerCluster, 1, 1) __launch_bounds__(Nu
...
+    // initialize the mPtrPermutedIdxToTokenIdx
+    if (params.mPtrPermutedIdxToTokenIdx != nullptr)
+    {
+        int32_t permIdxToTokenIdxNum
+            = (params.mNumTokens * NumTopExperts + (params.mNumExperts << params.mPaddingLog2) - params.mNumExperts);
+        for (int32_t i = clusterThreadIdx; i < permIdxToTokenIdxNum; i += NumThreadsPerCluster)
+        {
+            params.mPtrPermutedIdxToTokenIdx[i] = -1;
+        }
+        __threadfence();
```
rosenrodt, line 3000: What is the main use of `__threadfence()`?

Also, with the lack of cluster sync, it's not obvious how we ensure the last thread finishes initalizing its share of the buffer, before the first thread start writing the output to the buffer. There is cluster sync later in this kernel but it is not clear to me if it's always enabled:

```
if (params.mPtrPermutedIdxToTokenIdx != nullptr) {
  // init buffer to -1
  __threadfence();
}
if (params.mPtrScores != nullptr) {
   ...  
  // make packed scores available to all threads in cluster
  __cluster_barrier_arrive();
  __cluster_barrier_wait(); // buffer is properly -1 from this point forward?
}
```

rosenrodt: Thanks. Makes sense. Could you add comment to clearly inform that the sync will happen later in the kernel?

In the same vein of deferring sync as late as possible to gain perf, do you think doing the same for thread fence helps?

ChristinaZ: > Could you add comment to clearly inform that the sync will happen later in the kernel?
Agree, let me add more comments.

> do you think doing the same for thread fence helps?
Actually, I am not sure about this. The overall performance is improved after removing the cudaMemsetAsync().  While I haven't test the performance variation after delaying the `__threadfence()`.

bobboli (approved):

cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/RoutingKernel.cu
rosenrodt: Could you elaborate? If some element `arr[idx]` initialized by thread X is later read by thread Y, then `__threadfence()` sounds like a must-have. Do you mean we can remove threadfence if `arr[idx]` is not read, but instead __written__ by thread Y?

byshiue (approved):

PR #4872, opened 2025-06-03
[TRTLLM-5589] feat: Integrate TRT-LLM Gen FP8 Batched GEMM with Pytorch workflow kernel autotuner

DomBrown:
## Description

Extend TRT-LLM Gen fp8 BMM torch operator to integrate with the pytorch workflow kernel autotuner to allow profiling for the best kernel config for a given kernel runner config and matrix shapes.

Slightly modifies the kernel tuner to handle None tensors, as some TRT-LLM Gen kernels have input tensors that are optional based on the kernel configuration.

## Test Coverage

`tests/unittest/_torch/thop/test_tllmg_bmm.py`

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
+++ b/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
@@ -353,6 +353,7 @@ class FP8BatchedGemmRunner(TunableRunner):
         self.low_latency_kernel = low_latency_kernel
         self.tile_size = tile_size
         self.epilogue_tile_m = epilogue_tile_m
+        self.tuning_config = self.get_tuning_config()
 
         instance_key = (output_dtype, use_deep_seek_fp8, low_latency_kernel,
                         tile_size, epilogue_tile_m)
@@ -407,7 +408,6 @@ class FP8BatchedGemmRunner(TunableRunner):
 
         return tactics
 
-
     def get_default_valid_tactic(
         self,
         inputs: List[torch.Tensor],
@@ -425,10 +425,7 @@ class FP8BatchedGemmRunner(TunableRunner):
 
         return default_tactic
 
-
-    def get_dynamic_tensor_specs(
-        self
-    ) -> Tuple[DynamicTensorSpec, ...]:
+    def get_dynamic_tensor_specs(self) -> Tuple[DynamicTensorSpec, ...]:
         """Get the dynamic tensor specs for use with the AutoTuner."""
 
         # These indices correspond to the 0th input tensor and it's first dimension
@@ -441,10 +438,9 @@ class FP8BatchedGemmRunner(TunableRunner):
         m_values = (8, 16, 32, 64, 128, 256, 512, 1024, 2048)
         round_rule = lambda x: last_positive_power_of_2(x)
 
-        spec = DynamicTensorSpec(
-            MAT1_IDX, TUNED_DIM, m_values, round_rule)
+        specs = (DynamicTensorSpec(MAT1_IDX, TUNED_DIM, m_values, round_rule), )
 
-        return (spec, )
+        return specs
 
     def get_constraint_specs(self) -> Tuple[ConstraintSpec, ...]:
         """Get the constraint specs for the dynamic tensors for use with the AutoTuner.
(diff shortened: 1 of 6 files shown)


tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -359,6 +359,236 @@ def _(
                              dtype=output_dtype)
 
 
+def div_up(a: int, b: int) -> int:
```
nekorobov, line 362: `math.ceil` does that

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -359,6 +359,236 @@ def _(
...
+                    output_dtype, use_deep_seek_fp8, low_latency_kernel,
+                    tile_size, epilogue_tile_m)
+
+        self._kernel_runner = FP8BatchedGemmRunner._runner_dict[instance_key]
+
+    def forward(
+        self,
+        inputs: List[torch.Tensor],
+        tactic: int = -1,
+        do_preparation: bool = False,
```
nekorobov, line 395: Why is that needed?

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -359,6 +359,236 @@ def _(
...
+        """Get the dynamic tensors for use with the AutoTuner."""
+
+        # These indices correspond to the 0th input tensor and it's first dimension
+        # i.e. we are tuning M where the first input tensor is of shape [B, M, K]
+
+        MAT1_IDX = 0
+        TUNED_DIM = 1
+
+        # Starting at 8 as M % tile size == 0 is required
+        m_values = (8, 16, 32, 64, 128, 256, 512, 1024, 2048)
```
nekorobov, line 482: Suggest adding 1 and 4

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -359,6 +359,236 @@ def _(
...
+    )
+
+    return kernel_runner(
+        inputs=inputs,
+        tactic=best_tactic,
+    )
+
+
+@fp8_batched_gemm_trtllmgen.register_fake
+def _(
```
nekorobov, line 562: What is the purpose of this one? Add comment

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -359,6 +359,236 @@ def _(
...
+    ) -> List[int]:
+
+        mat1, mat2, _, _, _ = inputs
+
+        b = mat1.shape[0]
+        m = mat1.shape[1]
+        n = mat2.shape[1]
+        k = mat1.shape[2]
+
+        tactics = self._kernel_runner.get_valid_configs(m, n, k, b)
```
nekorobov, line 428: b, m, n, k order is better in my opinion

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
DomBrown: We're extending the `TunableRunner` abstract class, so need the same function signature. I haven't used it here, but it still needs to be present. The fp4 GEMM runner in the same file does the same.

The doc string for that argument states:

"When True, allows one-time setup operations to be performed before tactic evaluation begins. These operations are excluded from the performance measurements during autotuning. Notice that anything prepared in this phase should be persistent in the forward and can be accessed by the following forward calls."

I assume this for allocating any persistent buffers and such.

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
DomBrown: I'll change it. I think I did it that way because it's like that for CUTLASS problem shapes. I agree that B first is better

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
DomBrown: I was following the example of the other runners already here. I believe this is used for `torch.compile` compatibility. I could probably remove it if we don't need this, but a comment is probably fine

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
nekorobov: Ah, right. Fine, let's keep it this way

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
nekorobov: Thanks for the explanation

nekorobov (approved):

dcampora (approved):

PR #4879, opened 2025-06-03
[nvbug/5319281][fix] Stop drafting when we hit the draft model's max seq len

mikeiovine:
## Description

The draft model cannot inherit the target model's `max_seq_len`. There is a corner case; the target''s max_seq_len can be larger than what the draft model actually supports. If you set `max_seq_len` too high, you can get illegal memory accesses in RoPE during warmup.

In this case, we should let the drafter infer its own value for `max_seq_len` and just stop drafting in the corner case where we hit the draft model's `max_seq_len`. The performance is not ideal, but it doesn't matter much since it's a rare corner case. We just need to avoid illegal memory accesses in warmups.

Also added some logic to avoid the user specifying a `max_seq_len` via `ExecutorConfig` which is higher than what the model supports.

## Test Coverage

Removed 2048 max seq len restriction for EAGLE3 tests.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1735,6 +1735,16 @@ def _prepare_draft_batch(
                     # No space for draft tokens.
                     continue
 
+                # Stop drafting when we hit the max seqlen. We still need dummy draft
+                # tokens attached to the requests to make sure everything works properly
+                # with CUDA graph. These dummy tokens are already added by
+                # _prepare_draft_requests to make the KV cache/scheduler aware of the fact
+                # that we want to do spec decode, so no need to do anything else here.
```
hlu1, line 1742:
```suggestion
                # that we want to do spec decoding, so no need to do anything else here.
```

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -220,7 +220,9 @@ def create_py_executor(executor_config: ExecutorConfig,
                 pytorch_backend_config,
                 batch_size=executor_config.max_batch_size,
                 max_num_tokens=executor_config.max_num_tokens,
-                max_seq_len=model_engine.max_seq_len,
+                # Note: The draft model engine will infer its own max_seq_len.
+                # We'll stop drafting when we hit the max.
+                max_seq_len=executor_config.max_seq_len,
```
hlu1, line 225: Is `executor_config.max_seq_len` different from `model_engine.max_seq_len`? What are the definitions for them?

hlu1 (approved):

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
mikeiovine: Yes. `executor_config.max_seq_len` is a user-specified maximum sequence length that is allowed to be smaller than the model's max seq len.

PR #4894, opened 2025-06-04
fix a bug of global cuda graph dummy request

QiJune:
The kv cache manager is re-created after kv cache memory estimiation. So, we should re-create the cuda graph dummy request as well.


tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -459,7 +459,6 @@ def __init__(
         # with different KV cache managers.
         self.kv_cache_manager_key = KV_CACHE_MANAGER_KEY
         self.lora_model_config: Optional[LoraModelConfig] = None
-        self.cuda_graph_dummy_request = None
```
VALLIS-NERIA, line 462: will `cuda_graph_dummy_request ` be undefined if we don't call `warmup`?

QiJune: Nice suggestion, let me keep this line.

VALLIS-NERIA (approved):

pcastonguay (changes requested):
If this is a bug fix, could we add a test to make sure we don't introduce a similar bug in the future?

lfr-0531 (approved):

QiJune:
> If this is a bug fix, could we add a test to make sure we don't introduce a similar bug in the future?

@pcastonguay Nice suggestion. I updated the corresponding test case.

pcastonguay (approved):

PR #4900, opened 2025-06-04
chore: partition LLM class into TorchLLM and TrtLLM

Superjomn:
This PR does
1. Create TorchLLM and TrtLLM with most of the code shared, making it ready to switch TorchLLM as the default LLM for the upcoming version
2. Keep alias of `tensorrt_llm._torch.llm.LLM` for `TorchLLM` and `tensorrt_llm.LLM` for `TrtLLM`
3. Fix the api reference doc accordingly, especially keep the existing LLM unbroken
    - Add separate API references for both TrtLLM and TorchLLM, add support for TrtLlmArgs and TorchLlmArgs.

Document:

LLM's API reference works as before:
image

### DIFF
--- a/tensorrt_llm/llmapi/llm.py
+++ b/tensorrt_llm/llmapi/llm.py
@@ -31,8 +31,9 @@ from ..inputs import (PromptInputs, create_input_processor,
                       create_input_processor_with_hash, prompt_inputs)
 from ..logger import logger
 from ..sampling_params import SamplingParams
-from .llm_args import (LLMARGS_EXPLICIT_DOCSTRING, PybindMirror, TorchLlmArgs,
-                       TrtLlmArgs, _AutoDeployLlmArgs)
+from .llm_args import (TORCH_LLMARGS_EXPLICIT_DOCSTRING,
+                       TRT_LLMARGS_EXPLICIT_DOCSTRING, PybindMirror,
+                       TorchLlmArgs, TrtLlmArgs, _AutoDeployLlmArgs)
 from .llm_utils import (CachedModelLoader, KvCacheRetentionConfig,
                         LlmBuildStats, ModelLoader, _ModelRuntimeContext)
 from .mpi_session import MpiPoolSession, external_mpi_comm_available
@@ -83,8 +84,7 @@ class RequestOutput(DetokenizedGenerationResultBase, GenerationResult):
         ]
 
 
-LLM_DOCSTRING = LLMARGS_EXPLICIT_DOCSTRING + """
-        kwargs (Any): Advanced arguments passed to `LlmArgs`.
+TRT_LLM_DOCSTRING = TRT_LLMARGS_EXPLICIT_DOCSTRING + """
 
     Attributes:
         tokenizer (tensorrt_llm.llmapi.tokenizer.TokenizerBase, optional): The tokenizer loaded by LLM instance, if any.
@@ -92,8 +92,17 @@ LLM_DOCSTRING = LLMARGS_EXPLICIT_DOCSTRING + """
         llm_id (str): The unique ID of the LLM instance.
 """
 
+TORCH_LLM_DOCSTRING = TORCH_LLMARGS_EXPLICIT_DOCSTRING + """
+
+    Attributes:
+        tokenizer (tensorrt_llm.llmapi.tokenizer.TokenizerBase, optional): The tokenizer loaded by LLM instance, if any.
+"""
+
 
 class BaseLLM:
+    """
+    The base class for all LLM classes.
+    """
 
     def __init__(self,
                  model: Union[str, Path],
@@ -790,9 +799,9 @@ class BaseLLM:
         self.shutdown()
 
 
-@append_docstring(LLM_DOCSTRING)
-class LLM(BaseLLM):
-    """LLM class is the main class for running a LLM model.
+@append_docstring(TRT_LLM_DOCSTRING)
+class TrtLLM(BaseLLM):
+    """LLM class is the main class for running a LLM model using TensorRT-LLM backend.
 
     Parameters:
 """
@@ -843,7 +852,12 @@ class LLM(BaseLLM):
                 shutil.copy(file, target_engine_dir / file.name)
 
 
+@append_docstring(TORCH_LLM_DOCSTRING)
 class TorchLLM(BaseLLM):
+    """LLM class is the main class for running a LLM model using PyTorch backend.
+
+    Parameters:
+"""
 
     def __init__(self,
                  model: Union[str, Path],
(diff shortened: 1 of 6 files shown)


tensorrt_llm/llmapi/llm.py
```diff
@@ -756,32 +816,54 @@ def save(self, engine_dir: str) -> None:
...
+
+
+class LLM(TrtLLM if TLLM_USE_TRT_ENGINE == "1" else TorchLLM):
+
+    def __new__(cls, *args, **kwargs):
+        """Create appropriate LLM instance based on environment configuration."""
+        if TLLM_USE_TRT_ENGINE == "1":
+            return TrtLLM(*args, **kwargs)
+        else:
+            return TorchLLM(*args, **kwargs)
```
lucaslie, line 857:
```suggestion
class LLM(TrtLLM if TLLM_USE_TRT_ENGINE == "1" else TorchLLM):
    pass
```
There shouldn't be a need to overwrite `__new__` which is pretty low-level

tensorrt_llm/llmapi/llm.py
```diff
@@ -756,32 +816,54 @@ def save(self, engine_dir: str) -> None:
...
 
-    def __exit__(self, exc_type, exc_value, traceback) -> bool:
-        del exc_value, traceback
-        self.shutdown()
-        return False  # propagate exceptions
+        kwargs_dict = dict(kwargs)
+        kwargs_dict['backend'] = 'pytorch'
+        super().__init__(model, tokenizer, tokenizer_mode, skip_tokenizer_init,
+                         trust_remote_code, tensor_parallel_size, dtype,
+                         revision, tokenizer_revision, **kwargs_dict)
```
lucaslie, line 844: `kwargs` is already a dictionary by design:
```suggestion
        kwargs['backend'] = 'pytorch'
        super().__init__(model, tokenizer, tokenizer_mode, skip_tokenizer_init,
                         trust_remote_code, tensor_parallel_size, dtype,
                         revision, tokenizer_revision, **kwargs)
```

tensorrt_llm/llmapi/llm.py
```diff
@@ -756,32 +816,54 @@ def save(self, engine_dir: str) -> None:
...
 
-    def __getstate__(self):
-        raise RuntimeError("LLM object can not be pickled.")
 
-    def __del__(self):
-        self.shutdown()
+TLLM_USE_TRT_ENGINE = os.environ.get("TLLM_USE_TRT_ENGINE", "1")
+
```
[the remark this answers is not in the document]
Superjomn: Yeah, good idea. We can extend it later, either by making the ENV `TLLM_BACKEND=PYTORCH | TENSOR_RT | AUTO_DEPLOY" or adding another ENV `TLLM_USE_AUTO_DEPLOY`.

tensorrt_llm/llmapi/llm.py
Superjomn: Correct. The original idea was to reuse either one's docstring in LLM class, but the latest code explicitly sets the `LLM.__doc__`, so we can simplify the code.

tensorrt_llm/llmapi/llm.py
Superjomn: Nice catch, updated.

tensorrt_llm/llmapi/llm.py
```diff
@@ -736,6 +738,67 @@ def tokenizer(self) -> Optional[TokenizerBase]:
...
+                 tokenizer_mode: Literal['auto', 'slow'] = 'auto',
+                 skip_tokenizer_init: bool = False,
+                 trust_remote_code: bool = False,
+                 tensor_parallel_size: int = 1,
+                 dtype: str = "auto",
+                 revision: Optional[str] = None,
+                 tokenizer_revision: Optional[str] = None,
+                 **kwargs: Any) -> None:
+        if "backend" in kwargs:
+            raise ValueError("backend is not supported for LLM")
```
lucaslie, line 792: This is gonna break `autodeploy` backend for now since autodeploy just reuses the `LLM` api with the backend argument at the moment which is inheriting from `_TrtLLM`. You can see it in the CI as well: https://prod.blsm.nvidia.com/sw-tensorrt-top-1/job/LLM/job/main/job/L0_Test-x86_64/12976/testReport/L40S-PyTorch-1/test_unittests/Test___L40S_PyTorch_1___test_unittests_v2_unittest__torch_auto_deploy_unit_singlegpu_/

There is two potential solutions I can see:

1. You remove this check from this PR so autodeploy can continue using the backend argument and I will add a follow-up PR this week to introduce a proper `_AutoDeployLLM`. 
2. Or you can do it as part of this PR if you want to keep the check.

Happy to work on 1. or any other solution you prefer

lucaslie: SG. I will do a follow-up later this week :)

lucaslie (approved):
LGTM

tensorrt_llm/llmapi/llm.py
```diff
@@ -202,10 +208,6 @@ def __init__(self,
         exception_handler.register(self, 'shutdown')
         atexit.register(LLM._shutdown_wrapper, weakref.ref(self))
 
-    @property
-    def workspace(self) -> Path:
```
shaharmor98, line 206: Nit, feel free to resolve if you don't agree.  
Calling `llm.workspace` when llm is of type `_TorchLLM` will cause an exception.  
Although it might be resonable (given that `_TorchLLM` shouldn't have a workspace), is it the desired behavior? 
Alternatively, we can keep this method here and raise `NotSupportedError`, and override it in `_TRTLLM`.

tensorrt_llm/llmapi/llm.py
```diff
@@ -789,32 +851,71 @@ def save(self, engine_dir: str) -> None:
...
+class _TorchLLM(BaseLLM):
+    """LLM class is the main class for running a LLM model using PyTorch backend.
 
-    @staticmethod
-    def _shutdown_wrapper(self_ref):
-        # Retrieve the instance if it still exists
-        instance = self_ref()
-        if instance is not None:
-            instance.shutdown()
+    Parameters:
```
shaharmor98, line 859: Shouldn't we remove this if no parameters exist?

shaharmor98 (approved):
LGTM . Left two nits, feel free to resolve if you don't agree.

tensorrt_llm/llmapi/llm.py
Superjomn: Yeah. `workspace` is a concept for TRT, it is a directory specified by the user to determine where the TRT engine and other intermediate products to stored. It is unnecessary for PyT, so it is removed from BaseLLM to keep a clean interface for PyT, as the TRT backend is to be deprecated.

tensorrt_llm/llmapi/llm.py
Superjomn: Nice catch, but the _TorchLLM will take all the fields from [TorchLlmArgs](https://github.com/nvidia/TensorRT-LLM/blob/c53bc19f5e22c25bc57209379f447d8aace76984/tensorrt_llm/llmapi/llm_args.py#L1627) which holds tens of knobs as its parameters, so there are always many parameters there.

PR #4912, opened 2025-06-04
[5310329] fix: Fix warmup phase batch size out of range.

hyukn:
This is due to the number of autotuning warmup requests exceeding the limit of max_batch_size. This PR will limit the batch size according to multiple size constraints.
Autotuning is disabled due to a race condition that was triggered in the CI. This can be fixed by https://github.com/NVIDIA/TensorRT-LLM/pull/4565. This has been locally checked on main. Will enable the config on the main after being merged.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -540,15 +540,16 @@ class PyTorchModelEngine(ModelEngine):
 
             available_blocks = kv_cache_manager.get_num_free_blocks()
 
+            maximum_tunable_num_tokens = min(
+                self.batch_size * num_tokens_per_request, self.max_num_tokens,
+                available_blocks * kv_cache_manager.tokens_per_block)
+
             # Calculate number of full-length requests and remaining tokens
             # Each request has num_tokens_per_request tokens, except possibly the last one
-            full_len_request_num = self.max_num_tokens // num_tokens_per_request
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -540,14 +540,19 @@ def get_autotune_warmup_request():
...
-            full_len_request_num = self.max_num_tokens // num_tokens_per_request
-            remaining_tokens = self.max_num_tokens % num_tokens_per_request
+            full_len_request_num = maximum_tunable_num_tokens // num_tokens_per_request
+            remaining_tokens = maximum_tunable_num_tokens % num_tokens_per_request
 
             request_num = full_len_request_num if remaining_tokens == 0 else full_len_request_num + 1
 
-            if self.max_num_tokens > available_blocks * kv_cache_manager.tokens_per_block:
+            # request num should be guaranteed to be less than available_blocks. Double check here.
+            if request_num > available_blocks:
```
litaotju, line 555: Why "request num should be guaranteed to be less than available_blocks"?

litaotju (approved):

tensorrt_llm/_torch/pyexecutor/model_engine.py
hyukn: It seems to be too loose after a second thought. It stands only when each request is a gen request. Will remove it.

litaotju (approved):

PR #4913, opened 2025-06-04
fix: params passed to BuildConfig will propagate to LlmArgs as well

tomeras91:
PR https://github.com/NVIDIA/TensorRT-LLM/issues/4600 introduced logic in `BaseLlmArgs` to extract relevant args from `BuildConfig` to the `BaseLlmArgs` class. However, the logic included a bug, that caused arguments to be extracted only if they already exist in the `kwargs` passed to the `BaseLlmArgs` constructor.

This PR fixes this issue, by extracting the relevant arguments regardless if they exist in the `kwargs` or not.


Superjomn (approved):
LGTM

tomeras91:
closing as fixed in #4972 which re-applied the logic in #4600 along with some fixes (including this one)

PR #4943, opened 2025-06-05
[fix] Fix illegal mem access and possible accuracy lose

liji-nv:
# [fix] Fix illegal mem access and possible accuracy lose

## Description

MLA is using K only KVCache, the factor is not passed down to kvCacheBlockPartialCopy. When copy data into the last block, the block only include K cache, but the kernel is trying to access 2x data. If the tokens to copy is over 32, OOB access to the KVCache memory.

For blocks before last block, it may access block that do not belong to current block. Which has potential affect on accuracy.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/cpp/tensorrt_llm/kernels/kvCachePartialCopy.cu
+++ b/cpp/tensorrt_llm/kernels/kvCachePartialCopy.cu
@@ -58,7 +58,8 @@ unsigned int ipow2(unsigned int v)
 
 template <typename T>
 void hostKVCacheBlockPartialCopy(IBuffer& dst, IBuffer const& src, unsigned int numLayers, unsigned int numHeads,
-    unsigned int tokensPerBlock, unsigned int numHidden, unsigned int numTokensToCopy, cudaStream_t stream)
+    unsigned int tokensPerBlock, unsigned int numHidden, unsigned int numTokensToCopy, int kvFactor,
+    cudaStream_t stream)
 {
     unsigned int blockX = ipow2(numHidden);         // ensure block shape is a power of 2
     blockX = std::min(blockX, 32u);                 // blockX should not exceed warp size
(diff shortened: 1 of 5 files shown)


cpp/tensorrt_llm/kernels/kvCachePartialCopy.cu
```diff
@@ -75,55 +76,56 @@ void hostKVCacheBlockPartialCopy(IBuffer& dst, IBuffer const& src, unsigned int
     auto srcData = bufferCast<T>(src);
     auto dstData = bufferCast<T>(dst);
     cuKVCacheBlockPartialCopy<<<grid, block, 0, stream>>>(
-        dstData, srcData, 2 * numLayers, numHeads, tokensPerBlock, numHidden, numTokensToCopy);
+        dstData, srcData, numLayers * kvFactor, numHeads, tokensPerBlock, numHidden, numTokensToCopy);
```
zhhuang-nv, line 79: Nice finding! BTW I think the OOB will occur once the kernel is accessing last block even if it only copies one token.

zhhuang-nv (commented):
I think the root cause is we designed KVCacheManager with both K and V stored initially. I find that some docs like https://github.com/NVIDIA/TensorRT-LLM/blob/f9082a71681ec7ad16f2b59d07cd1c5c5d2f0653/cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h#L493 is not correct, it would be better if we can modify that.
BTW do we have any problem in this line? We also use fixed kvFactor 2 here: https://github.com/NVIDIA/TensorRT-LLM/blob/f9082a71681ec7ad16f2b59d07cd1c5c5d2f0653/cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h#L339

litaotju (approved):

ming-wei (approved):

Funatiq (approved):

zhhuang-nv:
I confirm that we don't have problem here: https://github.com/NVIDIA/TensorRT-LLM/blob/f9082a71681ec7ad16f2b59d07cd1c5c5d2f0653/cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h#L339. The shape always use kvFactor 2 here to expose a tensor for attention kernels.

zhhuang-nv (approved):

bobboli:
> I confirm that we don't have problem here:
> 
> https://github.com/NVIDIA/TensorRT-LLM/blob/f9082a71681ec7ad16f2b59d07cd1c5c5d2f0653/cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h#L339
> 
> 
> The shape always uses kvFactor 2 here to expose a offset tensor for attention kernels.

Shouldn't we use KV factor 1 for MLA?

zhhuang-nv:
> > I confirm that we don't have problem here:
> > https://github.com/NVIDIA/TensorRT-LLM/blob/f9082a71681ec7ad16f2b59d07cd1c5c5d2f0653/cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h#L339
> > 
> > The shape always uses kvFactor 2 here to expose a offset tensor for attention kernels.
> 
> Shouldn't we use KV factor 1 for MLA?

We always use KV factor 2 for kv cache block offset tensor, which would be used by attention kernels. Actually the K offsets and V offsets are same in MLA cases. But we still allocate and manage KV cache blocks by KV factor 1 for MLA internally.

PR #4466, opened 2025-05-20
feat: Enhance AutoTuner inference path and code readability

hyukn:
Fix AutoTuner warmup request generating.
* The current warmup phase creates one request, which is insufficient for the warmup to cover the max_num_tokens. Revise the warmup phase to a batch of requests to cover the max_num_tokens to eliminate potential fallback cases.

Refactor AutoTuner API and reduce host overhead.
* Refine (min, opt, max) values of optimization profile setup for get_valid_tactics to achieve the correct canImplement definition.
* Refine cache key assembly process to reduce host overhead and simplify API.
* Fix lru_cache usage to reduce host overhead.
* Move tuning config initialization as one-time object in tunable runner to reduce host overhead.

Improve tuning config readability. Use dataclass to define tuning config.

PR #4954, opened 2025-06-05
[nvbug 5325284][fix] Increase Nemotron-H warmup request robustness

tomeras91:
The `Mamba2Mixer` block forward pass in Nemotron-H assumes that `state_indices` in `MambaCacheManager` is valid, which is prepared during the call to `MambaHybridCacheManager.prepare_resources()`. However, warmup runs don't call `prepare_resources()`, requiring special handling within the `Mamba2Mixer` forward pass. Previously, warmup runs were assumed to have a single request with `request_id=0` and were identified by this condition.

PR #4466 changed the behavior of warmup runs, so they no longer always have just a single request with `id=0`. Consequently, this broke the `Mamba2Mixer` forward pass during warmup runs, preventing Nemotron-H from being initialized via the LLM API.

This PR enhances the stability of `Mamba2Mixer` during warmup runs by directly checking the validity of `state_indices`. If `state_indices` is invalid, it fills it with valid dummy values, reducing the differences between regular and warmup forward passes.

Additionally, since PR #4466 broke Nemotron-H silently due to the absence of unittests using the LLM API for Nemotron-H, this PR also introduces such a test to prevent similar issues in the future.


vegaluisjose (approved):
Great work @tomeras91 !

schetlur-nv (approved):

suyoggupta (approved):

PR #4972, opened 2025-06-06
fix: build_config in TorchLlmArgs and avoid arbitrary args 

Superjomn:
This PR restores the reverted PR #4600 with some fix on the ToT

It addresses two bugs:

- The LLM class accepts arbitrary arguments without reporting errors
- `max_batch_size`, `max_seq_len` and other values from build_config are ignored

### DIFF
--- a/tensorrt_llm/llmapi/llm_args.py
+++ b/tensorrt_llm/llmapi/llm_args.py
@@ -752,7 +752,7 @@ class BaseLlmArgs(BaseModel):
     """
     model_config = {
         "arbitrary_types_allowed": True,
-        "extra": "allow",
+        "extra": "forbid",
     }
 
     # Explicit arguments
@@ -930,18 +930,17 @@ class BaseLlmArgs(BaseModel):
     # private fields those are unstable and just for internal use
     num_postprocess_workers: int = Field(
         default=0,
-        description="The number of postprocess worker processes.",
-        alias="_num_postprocess_workers")
+        description=
+        "The number of processes for generated token postprocessing, such as detokenization and so on."
+    )
 
     postprocess_tokenizer_dir: Optional[str] = Field(
         default=None,
-        description="The postprocess tokenizer directory.",
-        alias="_postprocess_tokenizer_dir")
+        description="The path to the tokenizer directory for postprocessing.")
 
     reasoning_parser: Optional[str] = Field(
         default=None,
-        description="The parser to separate reasoning content from output.",
-        alias="_reasoning_parser")
+        description="The parser to separate reasoning content from output.")
 
     # TODO[Superjomn]: To deprecate this config.
     decoding_config: Optional[object] = Field(
@@ -957,6 +956,28 @@ class BaseLlmArgs(BaseModel):
         json_schema_extra={"type": "Optional[MpiSession]"},
         exclude=True,  # exclude from serialization
         alias="_mpi_session")
+
+    _parallel_config: Optional[object] = PrivateAttr(default=None)
+    _model_format: Optional[_ModelFormatKind] = PrivateAttr(default=None)
+    _speculative_model: Optional[str] = PrivateAttr(default=None)
+    _speculative_model_format: Optional[_ModelFormatKind] = PrivateAttr(
+        default=None)
+
+    @property
+    def parallel_config(self) -> _ParallelConfig:
+        return self._parallel_config
+
+    @property
+    def model_format(self) -> _ModelFormatKind:
+        return self._model_format
+
+    @property
+    def speculative_model(self) -> Optional[_ModelFormatKind]:
+        return self._speculative_model
+
+    @property
+    def speculative_model_format(self) -> _ModelFormatKind:
+        return self._speculative_model_format
 
     @print_traceback_on_error
     def model_post_init(self, __context: Any):
@@ -990,7 +1011,7 @@ class BaseLlmArgs(BaseModel):
         if self.moe_expert_parallel_size is None:
             self.moe_expert_parallel_size = -1
 
-        self.parallel_config = _ParallelConfig(
+        self._parallel_config = _ParallelConfig(
             tp_size=self.tensor_parallel_size,
             pp_size=self.pipeline_parallel_size,
             cp_size=self.context_parallel_size,
@@ -1012,6 +1033,15 @@ class BaseLlmArgs(BaseModel):
             tensorrt_llm.llmapi.llm_utils.BaseLlmArgs: The `BaseLlmArgs` instance.
         """
         kwargs = BaseLlmArgs._maybe_update_config_for_consistency(dict(kwargs))
+        # build_config always prioritize over flattened arguments in kwargs
+        if build_config := kwargs.pop("build_config", None):
+            if kwargs.get("backend") == "pytorch":
+                logger.warning("build_config is deprecated from PyT path.")
+            kwargs["max_batch_size"] = build_config.max_batch_size
+            kwargs["max_num_tokens"] = build_config.max_num_tokens
+            kwargs["max_seq_len"] = build_config.max_seq_len
+            kwargs["max_beam_width"] = build_config.max_beam_width
+
         ret = cls(**kwargs)
         ret._setup()
         return ret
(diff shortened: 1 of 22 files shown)


QiJune (approved):

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1190,7 +1149,7 @@ def _setup_speculative_config(self):
 
                 self.build_config.max_draft_len = self.speculative_config.max_draft_len
 
-                if self.backend not in ['pytorch', '_autodeploy']:
+                if self.backend != 'pytorch':
```
lucaslie, line 1152: Why are we removing the `_autodeploy` check from here?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1210,7 +1169,7 @@ def _setup_speculative_config(self):
                         eagle3_one_model)
             elif isinstance(self.speculative_config, NGramDecodingConfig):
                 self.build_config.speculative_decoding_mode = SpeculativeDecodingMode.NGRAM
-                assert self.backend in ['pytorch', '_autodeploy']
+                assert self.backend == 'pytorch'
```
lucaslie, line 1172: same here: I think the `_autodeploy` check should remain

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1583,6 +1656,11 @@ class TorchLlmArgs(BaseLlmArgs):
         "If true, enable min-latency mode. Currently only used for Llama4.",
     )
 
+    auto_deploy_config: Optional[object] = Field(
```
lucaslie, line 1659: should be removed. Not used anymore

tests/unittest/llmapi/test_llm_utils.py
```diff
@@ -41,7 +40,7 @@ def test_CachedModelLoader():
     args = LlmArgs(model=llama_model_path,
                    kv_cache_config=KvCacheConfig(free_gpu_memory_fraction=0.4))
     args.enable_build_cache = True
-    args._setup()
+    args.model_post_init({})
```
lucaslie, line 43: why do you need to call this explicitly? The Pydantic model should call this automatically during init

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -955,11 +950,43 @@ class BaseLlmArgs(BaseModel):
...
+
+    @property
+    def speculative_model_format(self) -> _ModelFormatKind:
+        return self._speculative_model_format
+
     @print_traceback_on_error
     def model_post_init(self, __context: Any):
+        self._ensure_lora_config_consistency()
+
+        self.max_input_len = self.max_input_len or 1024
```
lucaslie, line 988: This change doesn't make sense. Why do we set the default value to `None` but then overwrite it here in the `model_post_init`. Seems to me we should keep `1024` as the default value. 

If you want to handle the case where somebody may provide `None` as a value, e.g., `BaseLlmArgs(max_input_len=None)` we should have a field validator for that

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -955,11 +950,43 @@ class BaseLlmArgs(BaseModel):
...
+        self.quant_config = self.quant_config or QuantConfig()
```
lucaslie, line 989: Same here. why not use a field validator? 

Generally speaking, I suggest we keep the `model_post_init` function at a minimum to avoid obfuscation. Field validators are supposed to provide a more scalable, modular approach to post-processing field values in pydantic models

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1407,25 +1380,47 @@ class TrtLlmArgs(BaseLlmArgs):
...
+
     # Private attributes
     _auto_parallel_config: Optional[AutoParallelConfig] = PrivateAttr(
         default=None)
     # This is used to hold the options for convert_checkpoint
     _convert_checkpoint_options: Dict[str,
                                       Any] = PrivateAttr(default_factory=dict)
 
+    # This state helps to manage the build_config initialization.
+    _build_config_init_method: Optional[Literal["from_kwargs", "from_engine",
```
FrankD412, line 1405: Couldn't this also be handled by a `model_validator` that checks if an engine directory was specified?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1446,6 +1441,89 @@ def model_post_init(self, __context):
         if self.parallel_config.auto_parallel:
             self.parallel_config.world_size = self.auto_parallel_world_size
 
+        # Setup build config after model initialization
+        self._setup_build_config()
+        self._setup_enable_build_cache()
+        self._setup_speculative_config()
+        self._setup_embedding_parallel_mode()
+
+        self.calib_config = self.calib_config or CalibConfig()
```
FrankD412, line 1450: Could be handled with `Field(default_factory=CalibConfig, ...)`

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1463,9 +1541,10 @@ class TorchLlmArgs(BaseLlmArgs):
 
     # Just a dummy BuildConfig to allow code reuse with the TrtLlmArgs
     build_config: Optional[object] = Field(
-        default=None,
+        default_factory=lambda: BuildConfig(),
```
FrankD412, line 1544: You don't need a `lambda` function here. You can just specify `BuildConfig`

tensorrt_llm/llmapi/llm_args.py
FrankD412, line 1172: I'm confused why this is if-else logic. Couldn't the respective speculative decoding configs handle their own validation and setup?

FrankD412: Though that might require another option for defining which speculation mode you're using.

tensorrt_llm/llmapi/llm_args.py
Superjomn: This is a rebase mistake, I will restore it.

tests/unittest/llmapi/test_llm_utils.py
Superjomn: Here the member is assigned after the instance creation. Will the validator be invoked automatically after that?

tensorrt_llm/llmapi/llm_args.py
Superjomn: It could work, but I just moved the original code here with minimum change. Just hope to keep the scope of this PR as it contains some bugfix and easy to get conflict.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Good suggestion, updated.

tensorrt_llm/llmapi/llm_args.py
lucaslie: I see. In that case your current approach seems reasonable to me. 

However, let's maybe work backwards here? What is the final default value we want? Is it always going to be that `build_config` overwrites LlmArgs? If not maybe we should just overwrite the value in build_config with the value in LlmArgs?

tests/unittest/llmapi/test_llm_utils.py
lucaslie: Why can't you set the argument during initialization? 

and yes you are right. field/model validators won't be called when you update a field value after init

tensorrt_llm/llmapi/llm_args.py
FrankD412: @lucaslie -- @Superjomn and I met, we agreed this can be covered by a build config specific model validator in the "after" mode (feel free to correct me if I'm misremembering).

About the warning, if both are set shouldn't that be an error? I'm hesitant that we would set that for the user (TRT-LLM produces a lot of output and it's easy for that warning to get lost).

lucaslie: Sounds good. 👍 

Either way I think it would be useful to understand what is the desired 1.0 behavior and work backwards from there to define the requirements to resolve the overlapping config values in my opinion

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -990,7 +1017,7 @@ def model_post_init(self, __context: Any):
         if self.moe_expert_parallel_size is None:
             self.moe_expert_parallel_size = -1
 
-        self.parallel_config = _ParallelConfig(
+        self._parallel_config = _ParallelConfig(
```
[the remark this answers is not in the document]
Superjomn: After a discussion with @FrankD412 , I got some takeaways. Introducing multiple model_validators can help break down the complex logic in model_post_init into smaller methods so that each one can be clearer. I will try to introduce more model_validators in this PR. But I may leave some coarse-grained to keep smaller changes to avoid conflict and complex rebase.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Good catch, updated.

tensorrt_llm/llmapi/llm_args.py
Superjomn: Removed this as they become unnecessary after breaking down the logic into a sequental model_validators, thanks for the great suggestions!

Superjomn:
Hi @FrankD412 @lucaslie , in the latest commit, I introduced more model_validator and field_validator and break down most of the mode_post_init logics. Please help review again.

pcastonguay (approved):
lgtm

chzblych (approved):

tensorrt_llm/llmapi/llm_args.py
FrankD412: This looks good! -- thanks for spending the extra time on the refactoring, I think this looks clean and we don't need to worry about calling the post init hook anymore. 😃

FrankD412 (approved):

suyoggupta (approved):

PR #4986, opened 2025-06-06
fix: Fix warmup phase batch size out of range.

hyukn:
This is due to the number of autotuning warmup requests exceeding the limit of max_batch_size. This PR will limit the batch size according to multiple size constraints.


QiJune (approved):
LGTM

PR #5027, opened 2025-06-09
refactoring: port customized kernels with public cutlass version

yunruis:
pytest -s tests/unittest/_torch/modules/test_fused_moe.py# refactoring: port customized kernels with public cutlass version

In this PR we have open-sourced some internal Cutlass kernels. Meanwhile, to ensure stability and provide an even more optimized performance experience, we have retained the previous method of calling these kernels via static libraries, as another choice.
The purpose of this document is to introduce how to use these newly open-sourced Cutlass kernels, while also supporting the option to switch back to the previously supported internal Cutlass kernels via static libraries.

**Compilation**
These open-sourced cutlass kernels are `low_latency_gemm`, `moe-gemm`, `fp4_gemm` and `allreduce_gemm`. The switch between using open-sourced Cutlass kernels and static library Cutlass kernels can be made using the macro `USING_OSS_CUTLASS_*` achieving kernel-level control. By default, the open-source Cutlass kernels are used. For example:

```bash
python3 ./scripts/build_wheel.py --skip_building_wheel --linking_install_binary --use_ccache  --cuda_architectures "90-real;100-real" --python_bindings --install --micro_benchmarks
```
This will using open-sourced cutlass kernels. 

If users prefer to use the internal Cutlass kernels from the static library, they can control this during compilation by setting marco `USING_OSS_CUTLASS_*` to OFF. For instance, if a user wants to use the static library implementation for `low_latency_gemm` and `fused_moe_gemm`, the following compilation command can be used:

```
python3 ./scripts/build_wheel.py --skip_building_wheel --linking_install_binary --use_ccache  --cuda_architectures "90-real;100-real" -D "USING_OSS_CUTLASS_MOE_GEMM=OFF;USING_OSS_CUTLASS_LOW_LATENCY_GEMM=OFF" --python_bindings --install --micro_benchmarks
```


juney-nvidia (approved):
Thanks for the hard working, Yunrui.
Let's merge this PR ASAP to unblock the dependency and keep refining it in the subsequent PRs.

cpp/CMakeLists.txt
```diff
@@ -43,6 +43,14 @@ option(ENABLE_MULTI_DEVICE
        "Enable building with multi device support (requires NCCL, MPI,...)" ON)
 option(ENABLE_UCX "Enable building with UCX (Uniform Communication X) support"
        ON)
+option(ENABLE_OPENED_CUTLASS_LOW_LATENCY_GEMM
+       "Enable open sourced Cutlass low latency gemm kernel" ON)
+option(ENABLE_OPENED_CUTLASS_FP4_GEMM
+       "Enable open sourced Cutlass fp4 gemm kernel" ON)
+option(ENABLE_OPENED_CUTLASS_MOE_GEMM
+       "Enable open sourced Cutlass moe gemm kernel" ON)
+option(ENABLE_OPENED_CUTLASS_ALLREDUCE_GEMM
```
djns99, line 52: I dont think this name is good grammar. I think we should have `ENABLE_INTERNAL_CUTLASS_*` and default to OFF

cpp/tensorrt_llm/thop/moeOp.cpp
```diff
@@ -37,7 +42,18 @@ namespace torch_ext
 {
 
 namespace common = tensorrt_llm::common;
+#if defined(ENABLE_OPENED_CUTLASS_MOE_GEMM)
```
djns99, line 45: This seems like an inefficient way of doing this. The normal way would be using macros.

We should have something like the following in a header somewhere:

```
#define INTERNAL_CUTLASS_KERNELS_NAMESPACE tensorrt_llm::kernels::internal_cutlass_kernels
#define PUBLIC_CUTLASS_KERNELS_NAMESPACE tensorrt_llm::kernels::public_cutlass_kernels

#ifdef COMPILING_INTERNAL_CUTLASS
#define USE_CUTLASS_KERNELS_NAMESPACE namespace INTERNAL_CUTLASS_KERNELS_NAMESPACE
#else
#define USE_CUTLASS_KERNELS_NAMESPACE namespace PUBLIC_CUTLASS_KERNELS_NAMESPACE
#endif

#if defined(USING_INTERNAL_CUTLASS_IN_PUBLIC_REPO) || defined(COMPILING_INTERNAL_CUTLASS)
#define CUTLASS_KERNELS_NAMESPACE INTERNAL_CUTLASS_KERNELS_NAMESPACE
#else
#define CUTLASS_KERNELS_NAMESPACE PUBLIC_CUTLASS_KERNELS_NAMESPACE
#endif
```

Then in all the implemenation we should use:

```
USE_CUTLASS_KERNELS_NAMESPACE {
  // Implementation that exists in public and internal repo
}
```

And when we use it in TRT-LLM we can use:

```
using CUTLASS_KERNELS_NAMESPACE::ActivationType; // Use whichever version based on if CMake sets USING_INTERNAL_CUTLASS_IN_PUBLIC_REPO
using PUBLIC_CUTLASS_KERNELS_NAMESPACE::ActivationType; // Always use public version
using INTERNAL_CUTLASS_KERNELS_NAMESPACE::ActivationType; // Always use internal version
```

cpp/tensorrt_llm/kernels/cutlass_kernels/include/common.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+namespace tensorrt_llm::kernels::cutlass_kernels
```
djns99, line 3: Lets change this namespace to be a public one so we don't need to do the using namespace hack everywhere we use it.
There is no reason for this to be different in internal cutlass and public since this is part of the API

cpp/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
```diff
@@ -0,0 +1,846 @@
...
+ * Expert Parallelism refers to the mode where experts are divided between the nodes. Each node will handle only the
+ * tokens that are routed to the experts it is assigned to. Only the weights for the node's experts should be provided
+ * to the plugin For example, with #experts = 8, expert parallelism = 2: Node 0 would handle experts 0-3, and node 1
+ * would handle experts 4-7
+ *
+ * Regardless of parallelism mode:
+ *  * The input routing values must be the complete routing for all tokens/experts (required for softmax)
+ *  * An allreduce must be run on the result to combine the results from different nodes if parallelism > 1
+ */
+struct MOEParallelismConfig
```
djns99, line 137: This should be outside the `cutlass_kernels` namespace and combined with internal so we dont need to do `using MOEParallelismConfig = ...`

cpp/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
```diff
@@ -0,0 +1,846 @@
...
+    friend std::ostream& operator<<(std::ostream& os, MOEParallelismConfig const& config)
+    {
+        os << "tp_size: " << config.tp_size << ", tp_rank: " << config.tp_rank << ", ep_size: " << config.ep_size
+           << ", ep_rank: " << config.ep_rank << ", cluster_size: " << config.cluster_size
+           << ", cluster_rank: " << config.cluster_rank;
+        return os;
+    }
+};
+
+struct QuantParams
```
djns99, line 201: This should be outside the `cutlass_kernels` namespace so we dont need to do `using MOEParallelismConfig = ...`

cpp/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
```diff
@@ -0,0 +1,846 @@
...
+            {}};
+    }
+
+    static QuantParams FP8BlockScaling(float const* fc1_scales, float const* fc2_scales)
+    {
+        return QuantParams{{}, {}, {}, {}, {fc1_scales, fc2_scales}};
+    }
+};
+
+struct MoeMinLatencyParams
```
djns99, line 305: This should be outside the `cutlass_kernels` namespace so we dont need to do `using MOEParallelismConfig = ...`

cpp/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
```diff
@@ -0,0 +1,846 @@
+/*
```
djns99, line 1: I think we should make all the MOE code always be public except for the cutlass launcher. This should be straightforward, we just have a public `moe_gemm` CMake target, and a private `moe_instantiations` 

```
# Public or Private
add_library(moe_instantiations STATIC ${GROUPED_CU_INSTANTIATIONS})

# Only public
add_library(moe_gemm_src STATIC ${MOE_GEMM_SRC_CU} ${GROUPED_SRC_CPP})
target_link_libraries(moe_gemm_src PRIVATE moe_instantiations)
```

But this can be a separate PR

cpp/tensorrt_llm/kernels/cutlass_kernels/python/generate_kernels.py
```diff
@@ -97,13 +97,10 @@ def GetDataTypeBits(type):
     return DataTypeSize[type]
 
 
-def GetDataTypeNames(type, is_mx_fp4=None):
-    mxprefix = ""
-    if is_mx_fp4 is not None:
-        mxprefix = "mx_" if is_mx_fp4 else "nv_"
+def GetDataTypeNames(type):
```
djns99, line 100: I would really like the MXFP4 stuff to be open sourced in the MR, but even if we don't we still need to be able to use MXFP4 from the internal repo

cpp/tests/unit_tests/kernels/mixtureOfExpertsOssTest.cu
```diff
@@ -0,0 +1,2037 @@
...
+#ifdef ENABLE_BF16
+    WeightParams<__nv_bfloat16>,
+#endif
+#ifdef ENABLE_FP8
+    WeightParams<SafeFP8, SafeFP8, half>,
+#endif
+#ifdef ENABLE_FP4
+    WeightParams<SafeFP4, SafeFP4, half>,
+#endif
+    WeightParams<half>, WeightParams<float>
```
djns99, line 1293: I am guessing we are duplicating this because we dont want MXFP4. 
Instead of duplicating, please just add an `#ifdef` around adding it to this list in the existing moe test.

If we also need to test both compile definitions we should compile the same file twice with different compile definitions

cpp/micro_benchmarks/mixtureOfExpertsBackendBenchmarkLauncherOss.cu
```diff
@@ -0,0 +1,773 @@
...
+        // Extra ones we don't want for hardcoded runs
+        BENCHMARK_BASIC_DO_REGISTER(float, float, float);
+        BENCHMARK_BASIC_DO_REGISTER(half, uint8, half);
+        BENCHMARK_BASIC_DO_REGISTER(half, uint4b_t, half);
+#ifdef ENABLE_BF16
+        BENCHMARK_BASIC_DO_REGISTER(nv_bfloat16, nv_bfloat16, nv_bfloat16);
+#endif
+#ifdef ENABLE_FP4
+        BENCHMARK_BASIC_DO_REGISTER(SafeFP4, SafeFP4, half);
+#endif
```
djns99, line 581: As with MOE test, we should just disable MXFP4 in the existing test instead of duplicating the entire file

djns99 (changes requested):
Speaking from an MOE perspective. I know this is a high priority and we want to get it done. Feel free to delay my suggestions to a future PR.

For this merge, please can we deduplicate the MOE tests/benchmarks. 

Also, if possible can we please integrate the FP8xMXFP4 changes.
Long term I think we should move the MOE boundary to only close source cutlass launcher

cpp/tensorrt_llm/thop/moeOp.cpp
yunruis: The implementation of a 3-depth-level Marco is unnecessary and introduces additional complexity, which significantly increases the difficulty of reading and debugging for upcoming developer.

cpp/CMakeLists.txt
djns99: I wouldn't describe that as a convention, see FAST_BUILD and MICRO_BENCHMARK.

How I think of it is when a user wants to search the code base what will they look for. I think they are going to search for "internal cutlass" not for "open cutlass". So we should call it `ENABLE_INTERNAL_CUTLASS_*`.

For the other cases, they are turning functionality off that is on by default so it makes sense.
A bit of a semantic difference I know, but I think it is much more natural how I suggest.


If you want to keep it this way, it should be `ENABLE_OSS_CUTLASS_*`.  
`ENABLED_OPENED_CUTLASS_*` doesnt really make sense, since "open" is the the default so "opened" doesnt make sense

cpp/tensorrt_llm/thop/moeOp.cpp
djns99: I strongly disagree, noone will need to read the macro. This way everyone needs to manually do their own macro and if they forget bad things happen

cpp/tensorrt_llm/kernels/cutlass_kernels/include/common.h
djns99: Yes I agree, my point is this doesn't need to be duplicated internally and public, we can just have a public version and put it in the regular common/ directory and reuse it public & private

cpp/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
yunruis: As I previous said, the namespace `tensorrt_llm::kernels::cutlass_kernels` follow the path structure `tensorrt_llm/kernels/cutlass_kernels`. That is also other cutlass kernels follows. For caller, the distinguish between internal-cutlass-kernel and opened cutlass kernels is by namespace. Example seen pictures. That should be the right way. 
image
image
image
image

cpp/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
yunruis: same as previous `QuantParams `

cpp/tensorrt_llm/kernels/cutlass_kernels/include/moe_kernels.h
djns99: My point is this should be shared, there should be no need to distinguish because it should be the same for both. We can refactor this later though 🙂

djns99 (approved):

cpp/CMakeLists.txt
yunruis: changed to `USING_OSS_CUTLASS_*`

cpp/tests/unit_tests/kernels/mixtureOfExpertsOssTest.cu
```diff
@@ -0,0 +1,2039 @@
...
+#include "tensorrt_llm/runtime/bufferManager.h"
+
+#include <tensorrt_llm/kernels/cutlass_kernels/cutlass_type_conversion.h>
+#include <tensorrt_llm/kernels/quantization.h>
+
+using namespace tensorrt_llm::common;
+using namespace tensorrt_llm::runtime;
+using LoraParams = tensorrt_llm::kernels::LoraParams;
+// #if defined(USING_OSS_CUTLASS_MOE_GEMM)
+#if true
```
yuxianq, line 28: Also this `#if true`. Can we remove all `#if true` for debugging?

cpp/tests/unit_tests/kernels/mixtureOfExpertsOssTest.cu
```diff
@@ -0,0 +1,2039 @@
...
+#include "tensorrt_llm/runtime/cudaStream.h"
+
+#include <algorithm>
+#include <gtest/gtest.h>
+#include <numeric>
+
+//  Temporary opend-sourced version. Will be daleted when open-sourced moe_gemm support MXFP4
+
+// #if defined(USING_OSS_CUTLASS_MOE_GEMM)
+#if true
```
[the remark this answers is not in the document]
yunruis: yes, this file is temporary used before mxfp4 moe_gemm migrating to this repo. The file would be delete after that. I add this is to provide indicating info about how to migrating this file.

yuxianq (approved):

dongxuy04 (approved):

jenkins/Build.groovy
```diff
@@ -93,7 +93,7 @@ BUILD_CORES_REQUEST = "8"
 BUILD_CORES_LIMIT = "8"
 BUILD_MEMORY_REQUEST = "48Gi"
 BUILD_MEMORY_LIMIT = "64Gi"
-BUILD_JOBS = "8"
+BUILD_JOBS = "4"
```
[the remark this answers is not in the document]
yunruis: Almostly every time. change to 4 fix this problem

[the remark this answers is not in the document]
djns99: We have had problems with MOE compile using a lot of memory in the past, we have some mitigations in `generate_kernels.py` with how we group the kernels. Sounds like this may need re-addressed, please reach out to me @ZhanruiSunCh @chzblych if you think we need to improve this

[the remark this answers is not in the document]
yuantailing: The `Build-x86_64` phase currently takes 4h17m to run if Docker image is updated. Will this build time increase to 8 hours after this PR is merged? An 8-hour duration seems excessive.

An example pipeline that docker image is updated: [/LLM/main/L0_MergeRequest_PR pipeline #6081](https://nv/trt-llm-cicd/job/main/job/L0_MergeRequest_PR/6081/)

ZhanruiSunCh:
/bot skip --comment "[PR_Github #8730](https://nv/trt-llm-cicd/job/helpers/job/PR_Github/8730/) and [PR_Github #8594](https://nv/trt-llm-cicd/job/helpers/job/PR_Github/8594/) run a full pre-merge CI Pipeline with multi GPU test"

juney-nvidia (approved):

PR #5034, opened 2025-06-09
[TRTLLM-4406][feat] LLM sleep & wakeup Part 1: virtual device memory

tongyuantongyu:
## Description

This PR implements virtual device memory that can be released and later reallocated without changing memory address. This paves the road to support LLM sleep and wakeup functionality.

Preliminary test shows we can reduce the GPU memory usage to less than 1GB during sleep.

## Test Coverage

C++: cpp/tests/unit_tests/runtime/virtualMemoryTest.cpp

Python: tests/unittest/_torch/test_virtual_memory.py

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

## Summary by CodeRabbit

* **New Features**
  * Introduced a CUDA virtual memory management system, enabling advanced allocation, release, and rematerialization of GPU memory.
  * Added Python and C++ APIs to control virtual memory allocation, including tag-based batch operations and restore modes (NONE, CPU, PINNED, MEMSET).
  * Provided PyTorch integration with context managers for scoped virtual memory allocation and utilities for releasing or materializing memory by tag.

* **Bug Fixes**
  * Improved exception safety and error tracking for virtual memory operations.

* **Tests**
  * Added comprehensive unit and integration tests for CUDA virtual memory management, covering allocation, release, restoration, and PyTorch workflows.

* **Chores**
  * Updated build configurations to include new virtual memory sources and tests.


WilliamTambellini:
Cool @tongyuantongyu  
I ve looked at the change but did nt have time to dig the detail of UVMAllocator. 
Does it use cudaMallocManaged() ?
eg.
https://developer.nvidia.com/blog/tag/unified-memory/
Best

cpp/include/tensorrt_llm/runtime/virtualMemory.h
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    ~CUDAVirtualMemory()
+    {
+        // Calling release() is necessary if materialize() succeed or threw an exception.
+        // If release() is already called by the user, whether succeed or threw an exception,
+        // we shouldn't call release() again.
+        if (mHandle != 0 && mState != INVALID_STATE)
+        {
+            release();
```
joyang-nv, line 171: A little bit concern on corner case `mHandle != 0` and `mState == INVALID_STATE`, which I expect it should not happen.
Throw exception in this case is dangerous, but we may log warn?

cpp/include/tensorrt_llm/runtime/virtualMemory.h
```diff
@@ -0,0 +1,513 @@
...
+    size_t mState = 0;
+    CUmemGenericAllocationHandle mHandle{};
+    std::unique_ptr<Creator> mCreator;
+    std::vector<std::unique_ptr<Configurator>> mConfigurators;
+};
+
+/**
+ * LocalCreator creates memory allocation locally through cuMemCreate.
+ */
+template <bool count = true>
```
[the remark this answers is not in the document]
joyang-nv: Or shouldCount

cpp/include/tensorrt_llm/runtime/virtualMemory.h
```diff
@@ -0,0 +1,513 @@
...
+    uint8_t mValue;
+    bool mFirstTime = true;
+};
+
+/**
+ * BackedConfigurator backup the content of the allocation when teardown,
+ * and restore the backup on the following setup.
+ */
+
+struct BackedConfigurator : CUDAVirtualMemory::Configurator
```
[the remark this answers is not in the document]
joyang-nv: Usually during training phase, we call this `offload`. :) Pickup what you like.

cpp/include/tensorrt_llm/runtime/virtualMemory.h
tongyuantongyu: `mHandle != 0 && mState == INVALID_STATE` is the case when a previous `release()` call threw exception, so it's correct that we don't attempt release again.

cpp/include/tensorrt_llm/runtime/virtualMemory.h
```diff
@@ -0,0 +1,513 @@
...
+        // we shouldn't call release() again.
+        if (mHandle != 0 && mState != INVALID_STATE)
+        {
+            release();
+        }
+    }
+
+    explicit operator bool() const noexcept
+    {
+        return mCreator != nullptr;
```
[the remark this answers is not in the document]
tongyuantongyu: This is to test if this `CUDAVirtualMemory` object is managing a memory block. (In contrast to default constructed ones)
Testing handle being null or not is not meaningful if you consider we have 4 `Status`. But let me add a docstring to clarify the behavior.

cpp/include/tensorrt_llm/runtime/virtualMemory.h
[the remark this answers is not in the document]
tongyuantongyu: I'm following the name we already use at
https://github.com/NVIDIA/TensorRT-LLM/blob/d2a7e92eddb72c428e29a3a7936c0159e268c4b2/cpp/tensorrt_llm/runtime/tllmBuffers.h#L46-L47

cpp/tensorrt_llm/runtime/virtualMemory.cpp
```diff
@@ -0,0 +1,425 @@
...
+        configurators.push_back(std::make_unique<MemsetConfigurator>(address, n, 0, mConfig->mBackStream->get()));
+        break;
+    case CPU:
+        configurators.push_back(
+            std::make_unique<BackedConfigurator>(address, n, MemoryType::kCPU, mConfig->mBackStream->get()));
+        break;
+    case PINNED:
+        configurators.push_back(
+            std::make_unique<BackedConfigurator>(address, n, MemoryType::kPINNED, mConfig->mBackStream->get()));
+        break;
```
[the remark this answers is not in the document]
tongyuantongyu: I think `-Wswitch` (enabled by `-Wall`) will catch this case and emit a compiler warning (And promoted to error by `-Werror`).

Adding `default:` branch would silence this warning, only to reveal it later in runtime.

cpp/tensorrt_llm/runtime/virtualMemory.cpp
```diff
@@ -0,0 +1,425 @@
...
+    static CudaVirtualMemoryManager manager;
+    return &manager;
+}
+
+void cudaVirtualAddressAllocatorDeallocate(void* ptr, std::size_t n)
+{
+    auto const address = deviceptr_cast(ptr);
+    getVirtualMemoryManager()->remove(address);
+
+    auto pageSize = getpagesize();
```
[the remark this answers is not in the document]
tongyuantongyu: I added `const`. But `getpagesize()` is explicitly documented to be a value determined during runtime and shall not be determined to some fixed value during compilation. So you will not get constant propagation anyway.

tensorrt_llm/_torch/virtual_memory.py
```diff
@@ -0,0 +1,85 @@
...
+def _virtual_address_helper(mark: str, mode: RestoreMode):
+    stream = torch.cuda.current_stream()
+    push_virtual_address_allocator(mark, mode, stream.cuda_stream)
+    try:
+        yield
+    finally:
+        pop_virtual_address_allocator()
+
+
+def _scope(mark: str, mode: RestoreMode = RestoreMode.NONE):
```
ixlmar, line 42: `-> Generator[torch.cuda.MemPool]`

tensorrt_llm/_torch/virtual_memory.py
```diff
@@ -0,0 +1,85 @@
...
+    pool = torch.cuda.MemPool(_get_torch_pluggable_virtual_memory_allocator())
+    with _virtual_address_helper(mark, mode), torch.cuda.use_mem_pool(pool):
+        yield pool
+
+
+scope = contextmanager(_scope)
+
+
+@contextmanager
+def maybe_scope(enable: bool, mark: str, mode: RestoreMode = RestoreMode.NONE):
```
ixlmar, line 58: `-> Generator[Optional[torch.cuda.MemPool]]`

tensorrt_llm/_torch/virtual_memory.py
```diff
@@ -0,0 +1,85 @@
...
+    if enable:
+        yield from _scope(mark, mode)
+    else:
+        yield
+
+
+def release_with_mark(*marks: str):
```
ixlmar, line 65: `-> int`

tensorrt_llm/_torch/virtual_memory.py
```diff
@@ -0,0 +1,85 @@
...
+
+    :param marks: The mark of the scope when the virtual memory is allocated
+    :return: Number of memory blobs released
+    """
+    manager = get_virtual_memory_manager()
+    released_blobs = sum(manager.release_with_mark(mark) for mark in marks)
+    return released_blobs
+
+
+def materialize_with_mark(*marks: str):
```
ixlmar, line 76: `-> int`

ixlmar (commented):
Initial review of PR minus the tests. So far, I only came across minor details to comment on.

tensorrt_llm/_torch/virtual_memory.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def _get_torch_pluggable_virtual_memory_allocator():
+    global _virtual_memory_allocator
+    if _virtual_memory_allocator is not None:
+        return _virtual_memory_allocator.allocator()
+
+    th_common = next(path for path in torch.classes.loaded_libraries
+                     if 'th_common' in path)
+    _virtual_memory_allocator = torch.cuda.CUDAPluggableAllocator(
```
ixlmar, line 26: As I understand the [PyTorch documentation](https://docs.pytorch.org/docs/stable/notes/cuda.html#mixing-different-cuda-system-allocators-in-the-same-program) (cf. also the [source](https://github.com/pytorch/pytorch/blob/v2.7.0/torch/cuda/memory.py#L1045)), this bypasses the PyTorch allocator caching mechanism and hence is likely to result in rather costly allocations/frees. I would not expect this to be significant, so long as we only use the virtual-memory-allocator backed `MemPool` to allocate long-lived tensors.

tongyuantongyu (commented):
Addressed some review comments.

cpp/tensorrt_llm/runtime/virtualMemory.cpp
```diff
@@ -0,0 +1,423 @@
...
+    std::size_t const pageAlignedSize = (n + mConfig->mPageSize - 1) & ~(mConfig->mPageSize - 1);
+    TLLM_CU_CHECK_FREE_RESOURCE(cuMemAddressFree(address, pageAlignedSize));
+}
+
+} // namespace tensorrt_llm::runtime
+
+namespace tensorrt_llm::runtime
+{
+
+CudaVirtualMemoryManager* getVirtualMemoryManager()
```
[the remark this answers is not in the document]
tongyuantongyu: The design of `VirtualMemoryManager` does not enforce it to be used as singleton. It's the most simple way while being sufficient for now.

I put all code using `VirtualMemoryManager` into a dedicated namespace block.

tensorrt_llm/_torch/virtual_memory.py
ixlmar: My bad: I initially had understood the documentation that way, too. Then I looked at the source I linked above and got confused by this being a simple allocator wrapper. Now, I agree with your understanding of this:
* `use_mem_pool` [calls](https://github.com/pytorch/pytorch/blob/v2.7.0/torch/cuda/memory.py#L1184C5-L1184C30) [`CUDACachingAllocator::beginAllocateToPool`](https://github.com/pytorch/pytorch/blob/main/c10/cuda/CUDACachingAllocator.cpp#L2237) which modifies `.captures_underway`.
* When allocating memory, `CUDACachingAllocator.malloc`, [calls](https://github.com/pytorch/pytorch/blob/main/c10/cuda/CUDACachingAllocator.cpp#L1389) [`CUDACachingAllocator::get_pool`](https://github.com/pytorch/pytorch/blob/main/c10/cuda/CUDACachingAllocator.cpp#L2664) which checks `.captures_underway`.

ixlmar (commented):
LGTM (selective re-review)

cpp/tensorrt_llm/pybind/runtime/bindings.cpp
```diff
@@ -425,6 +430,30 @@ void initBindings(pybind11::module_& m)
...
+        .value("CPU", tr::CudaVirtualMemoryAllocator::RestoreMode::CPU)
+        .value("PINNED", tr::CudaVirtualMemoryAllocator::RestoreMode::PINNED)
+        .value("MEMSET", tr::CudaVirtualMemoryAllocator::RestoreMode::MEMSET);
+
+    m.def("get_virtual_memory_manager", &tr::getVirtualMemoryManager, "Get the virtual memory manager",
+        py::return_value_policy::reference);
+
+    m.def(
+        "push_virtual_memory_allocator",
+        [](std::string const& mark, tr::CudaVirtualMemoryAllocator::RestoreMode mode, uintptr_t stream)
```
ixlmar, line 444: Perhaps check that `cudaStream_t` fits inside `uintptr_t`, something like `static_assert(std::is_pointer_v<cudaStream_t>)` should hopefully work.

cpp/include/tensorrt_llm/runtime/virtualMemory.h
```diff
@@ -0,0 +1,517 @@
...
+public:
+    /**
+     * Add memory to be managed by this manager.
+     * @param handle  Unique handle provided to reference this memory in `remove`.
+     * @param mark    Mark the memory, so this memory can be targeted in `releaseWithMark` and `materializeWithMark`.
+     * @param memory  The CUDAVirtualMemory object.
+     *
+     * The memory and internal state will remain valid if any exception is thrown.
+     */
+    void add(uintptr_t handle, std::string mark, CUDAVirtualMemory&& memory);
```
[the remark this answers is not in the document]
tongyuantongyu: These feel the same as "tag". Let's simply go back to the first intuition, "tag".

cpp/tensorrt_llm/pybind/runtime/bindings.cpp
tongyuantongyu: Done. I'm asserting `static_assert(sizeof uintptr_t == sizeof cudaStream_t)`, as other cases will fail `reinterpret_cast` itself during compile.

joyang-nv (approved):

cpp/tensorrt_llm/thop/virtualMemoryAllocator.cpp
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2022-2025, NVIDIA CORPORATION.  All rights reserved.
```
liji-nv, line 2:
```suggestion
 * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```

cpp/tensorrt_llm/thop/virtualMemoryAllocator.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+extern "C"
+{
+
+    void* tensorrt_llm_virtual_memory_alloc(ssize_t size, int device, cudaStream_t) noexcept
+    {
+        void* ptr{};
+        try
+        {
+            tensorrt_llm::runtime::getVirtualMemoryAllocator().allocate(&ptr, size, device);
```
liji-nv, line 29: Need an explicit static cast for size to size_t

cpp/tensorrt_llm/thop/virtualMemoryAllocator.cpp
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "tensorrt_llm/runtime/virtualMemory.h"
+#include <cuda_runtime_api.h>
```
liji-nv, line 18: Is this required? Do not see CUDA related calls.

liji-nv (approved):

cpp/tensorrt_llm/thop/virtualMemoryAllocator.cpp
tongyuantongyu: This is for the definition of `cudaStream_t`.

cpp/tensorrt_llm/runtime/virtualMemory.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+using AllocConf = CudaVirtualMemoryAllocator::Configuration;
+
+AllocConf AllocConf::backgroundConfiguration{getVirtualMemoryManager(), "", NONE, nullptr, true};
+static std::forward_list vmAllocators{
+    CudaVirtualMemoryAllocator{{std::shared_ptr<AllocConf>{}, &AllocConf::backgroundConfiguration}}};
+
+CudaVirtualMemoryAllocator const& getVirtualMemoryAllocator()
+{
+    return vmAllocators.front();
```
HuiGao-NV, line 408: There is no place to use getVirtualMemoryAllocator/pushVirtualMemoryAllocator in multiple-threaded way. IMHO, this may happen. It's better to protect them with lock.

cpp/tensorrt_llm/runtime/virtualMemory.cpp
```diff
@@ -0,0 +1,422 @@
...
+{
+    void* ret{};
+    std::memcpy(&ret, &ptr, sizeof(CUdeviceptr));
+    return ret;
+}
+
+void CudaVirtualMemoryAllocator::allocate(Pointer* ptr, std::size_t n, int device) const
+{
+    CUdeviceptr address{};
+    std::size_t const pageAlignedSize = (n + mConfig->mPageSize - 1) & ~(mConfig->mPageSize - 1);
```
HuiGao-NV, line 341: Put this as a method of config and call this function here and in deallocate. Centralized method can reduce possiblity of mismatch.

cpp/tensorrt_llm/thop/virtualMemoryAllocator.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <sys/types.h>
+
+extern "C"
+{
+
+    void* tensorrt_llm_virtual_memory_alloc(ssize_t size, int device, cudaStream_t) noexcept
+    {
+        void* ptr{};
```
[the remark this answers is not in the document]
tongyuantongyu: This is zero-initialization, which by-standard initializes pointer types to `nullptr`.

cpp/tensorrt_llm/runtime/virtualMemory.cpp
ixlmar: This is actually a very relevant point @HuiGao-NV: In addition, there is also the possibility of threading on the Python side. Imagine a race between two concurrent executions of `with allocator_scope(A): ...` and `with allocator_scope(B): ...`:

* default allocator active
* thread 1 pushes allocator A -> A active
* thread 2 pushes allocator B -> B active
* issue: now thread 1 might allocate things with the wrong allocator (B instead of A)
* thread 2 pops allocator B -> A active
* issue: now thread 2 might allocate things with the wrong allocator (A instead of default)
* thread 1 pops allocator A -> default restored

I think to get the API semantics right, this will have to use Python "thread-local" variables, i.e., something semantically equivalent to https://docs.python.org/3/library/contextvars.html. In fact, `contextvars` has a [C API](https://docs.python.org/3/c-api/contextvars.html), so perhaps that is a good way to do it. I have the impression that this whole complexity could be eliminated if there was a way to attach the allocator configuration to `torch.cuda.CUDAPluggableAllocator` (in C++ one could e.g. just bind it to the alloc/free handlers).

ixlmar (changes requested):
Current Python API appears not to be thread/concurrency/async-safe

cpp/tensorrt_llm/runtime/virtualMemory.cpp
```diff
@@ -400,22 +401,27 @@ CudaVirtualMemoryManager& getVirtualMemoryManager()
...
 AllocConf AllocConf::backgroundConfiguration{getVirtualMemoryManager(), "", NONE, nullptr, true};
+
+static std::shared_mutex vmAllocatorsMutex;
 static std::forward_list vmAllocators{
     CudaVirtualMemoryAllocator{{std::shared_ptr<AllocConf>{}, &AllocConf::backgroundConfiguration}}};
 
 CudaVirtualMemoryAllocator const& getVirtualMemoryAllocator()
 {
+    std::shared_lock lock(vmAllocatorsMutex);
     return vmAllocators.front();
```
ixlmar, line 412: The returned reference may become invalid if another thread calls `popVirtualMemoryAllocator`.

ixlmar (approved):
LGTM

Just one detail: PyTorch's `use_mem_pool` [appears to be](https://github.com/pytorch/pytorch/blob/20b5f694f8591af770c09d222b6423263807071b/torch/cuda/memory.py#L1238) thread-local. If we used thread-local variables instead of globals for the allocator, our API would align better with theirs.

PR #5076, opened 2025-06-10
fix: fix cuda graph max batch size for spec decoding cases.

lfr-0531:
## Description

- Fix the `max_cuda_graph_bs`.
- Add assertions to output more useful information.

Before this fix, if we set a small `max_num_tokens`, the `max_cuda_graph_bs` would be equal to `max_num_tokens`. But when running models with spec decoding, the real input length `(1+max_draft_len) * max_cuda_graph_bs` will exceed the 
`max_num_tokens`.


mikeiovine (approved):
Thanks for fixing! This should unblock perf testing on trunk with `trtllm-bench`.

PR #5080, opened 2025-06-10
[TRTLLM-5921][feat] Prevent serialization of entire LoRA adapters in each request

amitz-nv:
## Description

Make LoRA pytorch to broadcast weights & config tensors only on new LoRA adapter, so when the adapter is already loaded on rank 0 (and thus, should be loaded on other ranks as well), only the adapter ID is sent in the request.

NOTE: The changes in this PR cause requests with a LoRA adapter that were previously unloaded from LoRA CPU cache not to work, as they remain in `_cpp_lora_weights` in `LoraManager` python class. It was decided that for now this flow is not supported, in favor of its significant performance boost.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/executor/worker.py
+++ b/tensorrt_llm/executor/worker.py
@@ -341,13 +341,14 @@ class GenerationExecutorWorker(GenerationExecutor):
         if mpi_rank() == 0:
             self.start_thread(self.dispatch_stats_thread)
 
-    def _load_lora_adapter(self, lora_request: LoRARequest):
-        self._lora_manager.load_from_ckpt(
+    def _load_lora_adapter(self, lora_request: LoRARequest) -> bool:
+        uids = [str(lora_request.adapter_id)]
+        return uids == self._lora_manager.load_from_ckpt(
             [lora_request.path],
             model_config=self._runtime_model_config if
             self._runtime_model_config is not None else self._lora_model_config,
             runtime_mapping=None,
-            uids=[str(lora_request.adapter_id)])
+            uids=uids)
 
     def _load_prompt_adapter(self,
                              prompt_adapter_request: PromptAdapterRequest):
@@ -359,12 +360,15 @@ class GenerationExecutorWorker(GenerationExecutor):
     def _enqueue_request(self, request: GenerationRequest) -> int:
         assert request.id is not None
         if self._lora_manager is not None and request.lora_request is not None:
-            self._load_lora_adapter(request.lora_request)
+            loaded_new_lora_adapter = self._load_lora_adapter(
+                request.lora_request)
             uid = str(request.lora_request.adapter_id)
             lora_config = tllm.LoraConfig(
                 task_id=request.lora_request.adapter_id,
-                weights=self._lora_manager.cpp_lora_weights[uid],
-                config=self._lora_manager.cpp_lora_config[uid])
+                weights=self._lora_manager.cpp_lora_weights[uid]
+                if loaded_new_lora_adapter else None,
+                config=self._lora_manager.cpp_lora_config[uid]
+                if loaded_new_lora_adapter else None)
         else:
             lora_config = None
 
--- a/tensorrt_llm/executor/worker.py
+++ b/tensorrt_llm/executor/worker.py
@@ -350,7 +350,7 @@ class GenerationExecutorWorker(GenerationExecutor):
             self._runtime_model_config is not None else self._lora_model_config,
             runtime_mapping=None,
             uids=[adapter_id])
-        return newly_loaded_uids and newly_loaded_uids[0] == adapter_id
+        return adapter_id in newly_loaded_uids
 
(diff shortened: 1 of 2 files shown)


tensorrt_llm/lora_manager.py
```diff
@@ -502,16 +502,16 @@ def load_from_ckpt(
...
-            self.load_from_hf(
+            return self.load_from_hf(
                 model_dirs=model_dirs_or_files,
                 model_config=model_config,
                 runtime_mapping=runtime_mapping,
                 uids=uids,
             )
         elif ckpt_source == "nemo":
-            self.load_from_nemo(
+            return self.load_from_nemo(
```
shaharmor98, line 514: Oh, that's change scope is bigger then we think

tensorrt_llm/lora_manager.py
```diff
@@ -502,16 +502,16 @@ def load_from_ckpt(
         runtime_mapping: Optional[Mapping] = None,
         uids: Optional[List[str]] = None,
         ckpt_source: str = "hf",
-    ):
+    ) -> List[str]:
         if ckpt_source == "hf":
-            self.load_from_hf(
+            return self.load_from_hf(
```
shaharmor98, line 507: Will have implications on other places as well...

tensorrt_llm/executor/worker.py
```diff
@@ -343,13 +343,14 @@ def start(self):
         if mpi_rank() == 0:
             self.start_thread(self.dispatch_stats_thread)
 
-    def _load_lora_adapter(self, lora_request: LoRARequest):
-        self._lora_manager.load_from_ckpt(
+    def _load_lora_adapter(self, lora_request: LoRARequest) -> bool:
+        uids = [str(lora_request.adapter_id)]
+        return uids == self._lora_manager.load_from_ckpt(
```
shaharmor98, line 348: Well, might be a nit, but in terms of SW, I don't like the this `==` clause. I think you should either throw an exception (and catch here) if none was loaded, or use another layer of abstraction.  
It's debatable though.

amitz-nv: If none were loaded it just means they're not new, which is OK, so far this method just did nothing on that flow. I just want it to return whether something was actually loaded or not

tensorrt_llm/lora_manager.py
shaharmor98: I think that many tests might fail because of this change (I'm not sure that NeMo callers expects this returned value

tensorrt_llm/lora_manager.py
shaharmor98: See previous comment

tensorrt_llm/executor/worker.py
amitz-nv: Updated the function, hopefully it's clearer now, also added a short docstring about its return value.

tensorrt_llm/lora_manager.py
shaharmor98: Ok, I still think that suddenly a lot of places will "ignore" the returned value because of your PR, but let's see if something fails due to that

amitz-nv: Maybe I'm missing something, what kind of flow are you concerned may break/fail from making these functions return value, while until this change they didn't return any value?

tensorrt_llm/executor/worker.py
```diff
@@ -343,13 +343,16 @@ def start(self):
...
+        """Returns True if the adapter was loaded by this call, False if it was already loaded"""
+        adapter_id = str(lora_request.adapter_id)
+        newly_loaded_uids = self._lora_manager.load_from_ckpt(
             [lora_request.path],
             model_config=self._runtime_model_config if
             self._runtime_model_config is not None else self._lora_model_config,
             runtime_mapping=None,
-            uids=[str(lora_request.adapter_id)])
+            uids=[adapter_id])
+        return newly_loaded_uids and newly_loaded_uids[0] == adapter_id
```
shaharmor98, line 355: Theoretically, `self._lora_manager.load_from_ckpt` could load and return more than one `uid`, so I guess we'd want to check if `adatper_id in newly_loaded_uids` rather than `==[0]`, am I right?

shaharmor98 (approved):
One small last comment.  
LGTM

tensorrt_llm/executor/worker.py
amitz-nv: It should load `adapter_id` at most as no more adapter UIDs are requested here, so it can either check with `in` instead of explicitly the first element, but when checking only the first I should also assert that no more than 1 element was returned (as what if more than 1 was returned but it wasn't the first?), then maybe just `in` is shorter & easier to read. I'll change it, thanks :)

PR #4538, opened 2025-05-21
[TRTLLM-4987][feat] Support context logits in TRTLLMSampler

dcampora:
# [TRTLLM-4987] Support context logits in TRTLLMSampler

## Description

This PR adds context logits support in TRTLLMSampler:

* An algorithm `handle_context_logits` is created in Python that mimics the C++ couterpart.
* Python bindings custom caster support for all TensorPtr and TensorConstPtr.
* Removed `std::vector` opaque bindings, as no reference parameters are required.
* Adjusted `test_return_logits` to also cover what is supported in TRTLLMSampler currently.

## GitHub Bot Help

`/bot [-h] ['run', 'kill', 'skip', 'reuse-pipeline'] ...`

Provide a user friendly way for developers to interact with a Jenkins server.

Run `/bot [-h|--help]` to print this help message.

See details below for each supported subcommand.

<details>

`run  [--disable-fail-fast --skip-test --stage-list "A10-1, xxx" --gpu-type "A30, H100_PCIe" --add-multi-gpu-test --only-multi-gpu-test --disable-multi-gpu-test --post-merge --extra-stage "H100_PCIe-[Post-Merge]-1, xxx"]`

Launch build/test pipelines. All previously running jobs will be killed.

`--disable-fail-fast ` *(OPTIONAL)* : Disable fail fast on build/tests/infra failures.

`--skip-test ` *(OPTIONAL)* : Skip all test stages, but still run build stages, package stages and sanity check stages. Note: Does **NOT** update GitHub check status.

`--stage-list "A10-1, xxx"` *(OPTIONAL)* : Only run the specified test stages. Examples: "A10-1, xxx". Note: Does **NOT** update GitHub check status.

`--gpu-type "A30, H100_PCIe"` *(OPTIONAL)* : Only run the test stages on the specified GPU types. Examples: "A30, H100_PCIe". Note: Does **NOT** update GitHub check status.

`--only-multi-gpu-test ` *(OPTIONAL)* : Only run the multi-GPU tests. Note: Does **NOT** update GitHub check status.

`--disable-multi-gpu-test ` *(OPTIONAL)* : Disable the multi-GPU tests. Note: Does **NOT** update GitHub check status.

`--add-multi-gpu-test ` *(OPTIONAL)* : Force run the multi-GPU tests. Will also run L0 pre-merge pipeline.

`--post-merge ` *(OPTIONAL)* : Run the L0 post-merge pipeline instead of the ordinary L0 pre-merge pipeline.

[...]

PR #5097, opened 2025-06-10
[test] Use LLM API for Nemotron-H correctness test

tomeras91:
Nemotron-H correctness tests doesn't just check that the completions are as expected, but rather also validates that logprobs aren't too far from reference logprobs. Initially, the test was implemented using a manual generation for-loop, since the LLM API didn't support returning context and generation logprobs.

After PRs #4538 and #4819 added context and generation logits support for the LLM API, it can now be used to compare logprobs as well, removing the need for a manual generation loop. It is also better to use the LLM API in the unittest since that's the standard way to use a model in TRTLLM.

Therefore, this PR merges the old Nemotron-H correctness and LLM API unittests, such that now the LLM API is used in the Nemotron-H correctness test.

Also, the reference generation logprobs were updated, since turns out there was a bug in the position_ids in the old manual generation loop. This further emphasizes why, when possible, standard APIs should be used instead of custom code.


vegaluisjose (approved):
LGTM

omera-nv (approved):
LGTM

amitz-nv (approved):
Use of get context & generation logits LGTM.
Note that logprobs support was merged this morning: https://github.com/NVIDIA/TensorRT-LLM/pull/4836 - using it can probably save some more code from this test

PR #5121, opened 2025-06-11
[fix][test] clear cuda cache before unittests automatically

omera-nv:
# Clear torch CUDA cache before unittests

We've encountered a case in which a test failed due to OOM errors, that were resolved by adding `torch.cuda.empty_cache` at the start of the test. This PR adds this to all unittests, so each one starts with an empty torch cuda cache and can make full use of the available device memory.

### DIFF
--- a/tests/unittest/conftest.py
+++ b/tests/unittest/conftest.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 # # Force resource release after test
 import pytest
+import torch
 import tqdm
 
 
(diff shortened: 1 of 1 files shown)


tests/unittest/conftest.py
```diff
@@ -48,3 +48,12 @@ def pytest_runtest_protocol(item, nextitem):
 
                 torch.cuda.empty_cache()
             break
+
```
tomeras91, line 51: Not all unittests are using `pytest`. Some of them are written using the `unittest` framework and are just run via `pytest`. Will this autouse fixture be added also to tests written in `unittest`?

tomeras91 (approved):

tburt-nv (approved):

PR #5122, opened 2025-06-11
feat: Support post_proc for bench

kaiyux:
This allows users to set the following field in `extra-llm-api-config.yml` to enable a (experimental) feature called multiple post-processing worker, to reduce the main process pressure by splitting the post-processing to other processes.
```
_num_postprocess_workers: 2
```

### DIFF
--- a/tensorrt_llm/bench/benchmark/low_latency.py
+++ b/tensorrt_llm/bench/benchmark/low_latency.py
@@ -218,6 +218,7 @@ def latency_command(
             n=beam_width,
             use_beam_search=beam_width > 1,
         )
+        post_proc_params = None  # No detokenization
         llm = LLM(**kwargs)
 
         # Perform warmup if requested.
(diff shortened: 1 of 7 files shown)


Superjomn:
@kaiyux  I noticed you exposed the postprocessing-worker in the bench tool. Here are some questions:

1. How to activate it when using trtllm-bench?
2. Should we add a test for it?

kaiyux:
> @kaiyux I noticed you exposed the postprocessing-worker in the bench tool. Here are some questions:
> 
> 1. How to activate it when using trtllm-bench?
> 2. Should we add a test for it?

@Superjomn I updated the description, hope that answers the question.

I can add a test later.

LinPoly (approved):
LGTM

FrankD412:
Is there some upper bound that we know is good that we can set this to consistently? And is it work making an option for this in case someone needs to increase it easily?

FrankD412:
Chatted with @kaiyux and understand that this is just enabling us to use it. LGTM.

FrankD412 (approved):

nv-yilinf (approved):

PR #5141, opened 2025-06-12
[feat] Add llm args to tune python gc threshold

nv-yilinf:
# [feat] Add llm args to tune python gc threshold

## Description

Our observation is that during the TRTLLM py_executor’s create_responses stage, Python GC will be invocated multiple times and will at certain point become extremely long (~200ms). The [mechanism](https://docs.python.org/3/library/gc.html#gc.set_threshold) behind is that Python will invoke GC when it detects num_of_allocation  – num_of_deallocation is larger than a certain threshold (default value is 700). In our case, we are creating (allocating) thousands of responses at the end of an iteration before sending them away in bulk, during which GC will be triggered multiple times because we are doing tens of thousands of allocations while almost no deactivations (they will be deferred until responses are out of function scope). Those GCs are pointless because none of those new response objects are dangled, but they take time (from several to hundreds of milliseconds).
 
Our plan is to increase the GC threshold to a certain heuristic value, say 20k, which should be enough to handle 2k responses,  and make it as a configurable llm-api argument in case we need to handle larger responses in the future. To minimize the potential impact on other components (e.g., OOM issue), we first will limit such change in the create_responses stage of py_executor.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -509,7 +509,9 @@ def create_py_executor_instance(
                       if spec_config is not None else 0,
                       kv_cache_transceiver=kv_cache_transceiver,
                       draft_model_engine=draft_model_engine,
-                      start_worker=start_worker)
+                      start_worker=start_worker,
+                      garbage_collection_gen0_threshold=executor_config.
+                      garbage_collection_gen0_threshold)
 
 
 def instantiate_sampler(model_engine: PyTorchModelEngine,
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -384,7 +384,8 @@ def create_py_executor_instance(
         draft_model_engine,
         start_worker,
         sampler,
-        lora_config: Optional[LoraConfig] = None) -> PyExecutor:
+        lora_config: Optional[LoraConfig] = None,
+        garbage_collection_gen0_threshold: Optional[int] = None) -> PyExecutor:
     kv_cache_manager = resources.get(KV_CACHE_MANAGER_KEY, None)
 
     spec_config = model_engine.spec_config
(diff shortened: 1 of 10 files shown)


tensorrt_llm/llmapi/llm.py
```diff
@@ -136,6 +136,7 @@ def __init__(self,
                 revision=revision,
                 tokenizer_revision=tokenizer_revision,
                 **kwargs)
+            print(f"LLM.args: {self.args.__dict__}")
```
Superjomn, line 139: Please clean up the debugging prints.

tensorrt_llm/executor/worker.py
```diff
@@ -595,6 +601,7 @@ def worker_main(
     is_llm_executor: Optional[
         bool] = True,  # whether it's the main executor instance
     lora_config: Optional[LoraConfig] = None,
+    garbage_collection_gen0_threshold: Optional[int] = None,
```
Superjomn, line 604: How about changing the typing to `int` since it gets a value of 20000 as the default value.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -384,7 +384,8 @@ def create_py_executor_instance(
         draft_model_engine,
         start_worker,
         sampler,
-        lora_config: Optional[LoraConfig] = None) -> PyExecutor:
+        lora_config: Optional[LoraConfig] = None,
+        garbage_collection_gen0_threshold: Optional[int] = None) -> PyExecutor:
```
Superjomn, line 388: I noticed here the default value is None, while in TorchLlmArgs the default value is 20000, should we unify it to be None? That is, the default behaviour is not modifying the system's GC gen0 threshold.

nv-yilinf: I think the rationale behind is we do want to change the default GC gen0 threshold because otherwise large GC (gen2 threshold) will be triggered frequently when we are at high concurrency (e.g., >2k). That's why in BaseLlmArgs we set it to be 20000. However, in other occurrence I think it's best to set the default value to None to avoid accidental change of GC behaviour. Do this make sense?

tensorrt_llm/llmapi/llm.py
nv-yilinf: Done, thanks.

kaiyux (approved):

tensorrt_llm/_torch/pyexecutor/_util.py
Superjomn: Yeah, it makes sense.

tensorrt_llm/executor/worker.py
Superjomn: Yeah, we can keep the Optional to allow user specify None.

QiJune (approved):
LGTM

PR #4416, opened 2025-05-18
[None][feat] Multi-block mode for Hopper spec dec XQA kernel

jhaotingc:
# [feat] Multi-block mode for Hopper spec dec XQA kernel

## Description

Following [PR 3269](https://github.com/NVIDIA/TensorRT-LLM/pull/3269), it's observed that at low batch size and low draft len, the XQA hopper spec dec kernel has low CTA.

The reason is root caused to not having multi-block mode when Spec-dec is turned on.

The XQA Hopper Spec-dec kernel is launched with `gridDim = dim3{specDecBlocks, multi_block, nbKVHeads * xqaParams.batch_size}`, where 

- `specDecBlocks = divUp(specDecParams.qSeqLen, 64 / num_q_heads_over_kv)`, num_q_heads_over_kv for Llama is 8.
- `nbKVHeads` is num_kv_head per TP rank, for Llama 3, it's 8 when TP=1, 2 when TP=4, 1 when TP=8.

Before the fix, multi_block = 1.  

At a very common use case, where eagle draft length set to < 8, running TP=8, using Llama 3 70b / Llama 3.1 8b. The number of blocks launched could be only `batch_size` blocks. At BS=1, only 1 block will be launched.
```
gridDim = dim3{ divUp(7 / 8), 1, 1 * xqaParams.batch_size} # = batch_size
```

Therefore, multi-block mode is crucial for low BS, low draft length case.

## Heuristic design:

A series of sweeps was done with [xqa](https://github.com/NVIDIA/TensorRT-LLM/tree/main/cpp/kernels/xqa). The experiment showed that when original gridDim is less than a wave of SM, there is benefit for multi-block mode. Furthermore, when original block count is <= 8, 64k > ISL >= 1k, populating all SMs is not always good. The experiments are shown in Appendix.

### Kernel Speedup for ISL=32k, draft length 7, batch size 2.
By increasing gridDim from (1,1,2) to (1,32,2) yield a **7.8x speedup.**
```
before: 
kernel_mha
Begins: 39.1859s
Ends: 39.1863s (+418.620 μs)
grid:  <<<1, 1, 2>>>
block: <<<128, 1, 3>>>

after: 
kernel_mha
Begins: 33.2649s
Ends: 33.265s (+52.991 μs)
grid:  <<<1, 32, 2>>>
block: <<<128, 1, 3>>>
```
### Kernel Speedup for ISL=32k, draft length 7, batch size 8.
[...]

PR #5146, opened 2025-06-12
[https://nvbugspro.nvidia.com/bug/5329655] [feat] Pytorch path add spec dec param to attention op 

jhaotingc:
## Description

###TL;DR: 

This PR adds back spec-dec param that's in cpp runtime GPTAttentionPlugin to torch runtime AttentionOp.
Before the PR, the acceptance length of Hopper + Eagle3 one engine was either 1 (logits goes nan, no tokens accepted) or `draft_len` (logits goes nan, and all got accepted). 
After the PR:

(H200 TP8, conc=4)
- public BF16 Llama 3.3 70b + public Eagle3: average acceptance length **1.83 / 4**, tps/user speedup over regular decoding: **1.26x**
- NV FP8 Llama 4 Maverick + NV BF16 Eagle3: average acceptance length **2.90 / 4**, tps/user speedup over regular decoding: **1.67x**

Eagle3 one-engine tests added on Hopper and Blackwell.

Llama3 8b + Eagle3 Before this PR:

```
H200:
[0] Prompt: 'Hello, my name is', Generated text: ' Emily and!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'
[1] Prompt: 'The president of the United States is', Generated text: ' the!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'
[2] Prompt: 'The capital of France is', Generated text: ' a!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'
[3] Prompt: 'The future of AI is', Generated text: ' bright!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!'
```

After this PR:

```
[0] Prompt: 'Hello, my name is', Generated text: ' Emily and I am a 25-year-old freelance writer and editor. I have a passion for storytelling and a knack for crafting compelling narratives. I have been writing for over 10 years, and I have honed my skills through various writing projects, including articles, blog posts, and short stories.\nI'
[1] Prompt: 'The president of the United States is', Generated text: ' the head of state and head of government of the United States. The president serves a four-year term and is limited to two terms. The president is elected through the Electoral College system, where each state is allocated a certain number of electoral votes based on its population. The candidate who receives the majority of the'
[2] Prompt: 'The capital of France is', Generated text: ' a city of romance, art, fashion, and cuisine. Paris is a must-visit destination for anyone who loves history, architecture, and culture. From the iconic Eiffel Tower to the world-famous Louvre Museum, Paris has something to offer for every interest and age.\nThe city is divided'
[3] Prompt: 'The future of AI is', Generated text: ' bright, but it also raises concerns about bias, accountability, and the impact on jobs. Here are some of the key challenges and opportunities that AI will face in the coming years.\nThe future of AI is bright, but it also raises concerns about bias, accountability, and the impact on jobs. Here are some'

```

trtllm-bench for llama3-8b + Eagle3 not runnable for before and after this PR.


trtllm-bench for llama3-70b + Eagle3: (acceptance rate is fixed)
(H200, conc=1, aa dataset)
```
(Eagle3)
Per User Output Speed (tps/user):                 149.0368 (1.23x)
-- Acceptance Rate Details --------------------------------

[AR] MINIMUM: 1.61
[AR] MAXIMUM: 1.97
[AR] AVERAGE: 1.75
[AR] P50    : 1.74
[AR] P90    : 1.97
[AR] P95    : 1.97
[AR] P99    : 1.97

(Auto-Regressive)
Per User Output Speed (tps/user):                 121.6365

```



(H200, conc=4, aa dataset)
```
Per User Output Speed (tps/user):                 135.1499 (1.26x)
-- Acceptance Rate Details --------------------------------


[AR] MINIMUM: 1.32
[AR] MAXIMUM: 2.06
[AR] AVERAGE: 1.83
[AR] P50    : 1.92
[AR] P90    : 2.02
[AR] P95    : 2.06
[AR] P99    : 2.06

(Auto-Regressive)
Per User Output Speed (tps/user):                 107.8985


(B200 acceptance length reference)
-- Acceptance Rate Details --------------------------------

[AR] MINIMUM: 1.49
[AR] MAXIMUM: 2.08
[AR] AVERAGE: 1.80
[AR] P50    : 1.86
[AR] P90    : 2.04
[AR] P95    : 2.08
[AR] P99    : 2.08
===========================================================

```

Llama4 + Eagle3:
(bs1)
```
H200 Eagle3                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
Per User Output Speed (tps/user):                 392.0577
-- Acceptance Rate Details --------------------------------


[AR] MINIMUM: 2.66
[AR] MAXIMUM: 3.17
[AR] AVERAGE: 2.92
[AR] P50    : 2.95
[AR] P90    : 3.17
[AR] P95    : 3.17
[AR] P99    : 3.17

H200 Auto Regressive
Per User Output Speed (tps/user):                 229.3940


B200 Eagle3
-- Acceptance Rate Details --------------------------------


[AR] MINIMUM: 2.31
[AR] MAXIMUM: 3.04
[AR] AVERAGE: 2.59
[AR] P50    : 2.55
[AR] P90    : 3.04
[AR] P95    : 3.04
[AR] P99    : 3.04

```

(conc4)
```
H200 Eagle3
Per User Output Speed (tps/user):                 315.6859
-- Acceptance Rate Details --------------------------------


[AR] MINIMUM: 1.46
[AR] MAXIMUM: 3.19
[AR] AVERAGE: 2.90
[AR] P50    : 3.00
[AR] P90    : 3.17
[AR] P95    : 3.19
[AR] P99    : 3.19

H200 Auto-regressive
Per User Output Speed (tps/user):                 188.7117


B200 Eagle3
-- Acceptance Rate Details --------------------------------


[AR] MINIMUM: 2.32
[AR] MAXIMUM: 3.98
[AR] AVERAGE: 3.25
[AR] P50    : 3.47
[AR] P90    : 3.98
[AR] P95    : 3.98
[AR] P99    : 3.98

```

**I'm still not sure if llama4 + eagl3 on Hopper is supposed to have this slight drop in acceptance length, not sure if it's precision error or XQA kernel handling nope layer is different.**
**[RESOLVED] Potential perf drop when copying / computing the spec-dec mask, due to cpu-to-gpu copy.**

## How to run trtllm-serve / trtllm-bench with one-engine eagle3

`extra-llm-api-config.yml`
```
enable_attention_dp: false
disable_overlap_scheduler: true # seeing accuracy drop if false, need investigation
cuda_graph_config:
        cuda_graph_batch_sizes: [1,2,3,4]
autotuner_enabled: false
enable_min_latency: false
print_iter_log: true
# kv_cache_config:
#     enable_block_reuse: false set this to false if base and draft are running of different dtype, e.g. base fp8, head bf16

speculative_config:
    decoding_type: Eagle
    max_draft_len: 3
    eagle3_one_model: true
    pytorch_weights_path: /trt_llm/data/llm-models/EAGLE3-LLaMA3.3-Instruct-70B
```

`trtllm-bench`
```
BASE_MODEL="/trt_llm/data/llm-models/llama-3.3-models/Llama-3.3-70B-Instruct"

trtllm-bench -m $BASE_MODEL \
    --model_path $BASE_MODEL \
    throughput \
    --backend pytorch \
    --max_batch_size 4 \
    --max_num_tokens 3072 \
    --extra_llm_api_options extra-llm-api-config.yml \
    --dataset aa_gen_16k_1k_apr22.txt \
    --kv_cache_free_gpu_mem_fraction 0.7 \
    --warmup 1 \
    --streaming \
    --num_requests 20 \
    --concurrency 4 \
    --tp 8 \
    --ep 1
```

`trtllm-serve`
```
trtllm-serve serve \
    --backend pytorch \
    --extra_llm_api_options extra-llm-api-config.yml \
    --host 0.0.0.0 \
    --port 8000 \
    --max_batch_size 4 \
    --max_num_tokens 3072 \
    --tp 8 \
    $BASE_MODEL
```


## Test Coverage

* Added eagle3 one model test on B200.
* Added eagle3 one model test on H100.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -314,6 +320,22 @@ class Runner : public RunnerBase
...
+                    spec_decoding_position_offsets.has_value(), "spec_decoding_packed_mask needs value");
+                TLLM_CHECK_WITH_INFO(
+                    spec_decoding_generation_lengths.has_value(), "spec_decoding_packed_mask needs value");
+                enqueue_params.spec_decoding_packed_mask = spec_decoding_packed_mask.value().data_ptr<int32_t>();
+                enqueue_params.spec_decoding_position_offsets
+                    = spec_decoding_position_offsets.value().data_ptr<int32_t>();
+                enqueue_params.spec_decoding_generation_lengths
+                    = spec_decoding_generation_lengths.value().data_ptr<int32_t>();
+                enqueue_params.spec_decoding_is_generation_length_variable = true;
+                enqueue_params.spec_decoding_max_generation_length = input_seq_length + 1;
```
jhaotingc, line 336: Hi @lowsfer, I have question toward the computation of these masks:

Take draft_len = 4 for example, the parameters for spec-dec gen step would be:
```
spec_decoding_max_generation_length: 5
generation_input_length: 4
spec_decoding_packed_mask (shape: bs * generation_input_length * ceil(generation_input_length/32)): [1,3,7,15]
spec_decoding_position_offset (shape: bs * generation_input_length): [0,1,2,3]
```
I got this from previous [eagle](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/models/eagle/model.py#L437) implementation for the spec-dec masks. There was a series of kernels computing mask for tree-based path, but in PyTorch flow only linear tree is supported.

I adopted this simpler computation for the masks first, if the tree-based spec-dec is supported, we can add the spec-dec mask logics into torch flow. What's your opinion on this, thanks!

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -486,6 +514,10 @@ void attention_inplace(torch::Tensor q, torch::optional<torch::Tensor> k, torch:
 
     op->mAttentionChunkSize = attention_chunk_size;
 
+    op->mUseSpecDecoding = use_spec_dec; // true
+    op->mIsSpecDecodingEnabled = spec_decoding_position_offsets.has_value();
```
yuxianq, line 518: What is the difference between `mUseSpecDecoding` and `mIsSpecDecodingEnabled`? In which case will they be different?

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -741,6 +772,10 @@ TORCH_LIBRARY_FRAGMENT(trtllm, m)
         ", Tensor? mla_context_paged_kv"
         ", Tensor? mla_context_kv_cache_block_offsets"
         ", int? attention_chunk_size"
+        ", bool use_spec_dec"
+        ", Tensor? spec_decoding_position_offsets"
+        ", Tensor? spec_decoding_packed_mask"
+        ", Tensor? spec_decoding_generation_lengths"
```
yuxianq, line 778: We will have 63 arguments after this PR, which is very close to 64. Can we also pack these three `spec_decoding_xxx` to `List[Tensor]? spec_decoding_tensors`, so that other people will not reach the limit immediately?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -382,11 +393,15 @@ def run(
...
-            self.rotary_embedding_long_m_scale,
-            self.rotary_embedding_max_positions,
-            self.rotary_embedding_original_max_positions,
+            [
+                self.rotary_embedding_scale,
+                self.rotary_embedding_short_m_scale,
+                self.rotary_embedding_long_m_scale
+            ],
+            [
+                self.rotary_embedding_max_positions,
```
yuxianq, line 402: Please don't pack arguments in trtllm backend, let's hide it inside the `attention` op defined in torch_custom_ops.py. As I commented in the discussion, we can keep the `attention` op API unchanged when we switch to the packed solution.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -271,6 +273,11 @@ def forward(
             out_scale_sf = self.o_proj.input_scale
 
         q, k, v = self.convert_qkv(q, k, v)
+        use_spec_dec = (
+            spec_metadata.use_spec_dec
+            and spec_metadata.spec_dec_mode.require_multi_query_attn_kernel(
+                get_attention_backend(
```
yuxianq, line 279: If `spec_metadata` is part of `attn_metadata`, we don't need to set `use_spec_dec` according to attention backend. Only trtllm backend will handle `spec_metadata`, all other backends can ignore it.

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -486,6 +514,10 @@ void attention_inplace(torch::Tensor q, torch::optional<torch::Tensor> k, torch:
 
     op->mAttentionChunkSize = attention_chunk_size;
 
+    op->mUseSpecDecoding = use_spec_dec; // true
+    op->mIsSpecDecodingEnabled = spec_decoding_position_offsets.has_value();
+    op->mMultiBlockMode = !(op->mIsSpecDecodingEnabled);
```
nv-yilinf, line 519: Does this imply that spec decoding won't work with multi block attention?

tensorrt_llm/_torch/speculative/eagle3.py
```diff
@@ -284,6 +323,7 @@ def forward(self, input_ids, position_ids, hidden_states, logits,
         next_draft_tokens = []
         for i in range(self.max_draft_tokens):
             hidden_states, hidden_states_to_save = draft_model.model(**inputs)
+            spec_metadata.use_spec_dec = False
```
nv-yilinf, line 326: Have we verified this implementation with cuda_graph on? IIUC pure CPU operation won't be replayed after cuda_graph is captured.

tensorrt_llm/_torch/speculative/eagle3.py
```diff
@@ -183,6 +183,13 @@ def __post_init__(self):
             device='cuda',
         )
 
+        # currently Eagle3 only supports linear tree
+        self.has_spec_dec_tree = True
```
yweng0828, line 187: Hi, do you think the name `has_spec_dec_tree` may be a bit confusing? This may make people think that this bool is used to determine whether tree decoding is enabled.

We have three situations to distinguish: 1) whether to use spec; 2) whether it is a chain; and 3) whether it is a static tree.

`has_spec_dec`: used to determine whether spec is enabled, the default is false;

`is_spec_dec_tree`: used to determine whether it is a tree, the default is false. i.e., use chain;

`is_spec_dec_dynamic_tree`: used to determine whether it is a dynamic tree, the default is false, i.e., static tree;

There may be a more concise way to judge, just my two cents. : )

tensorrt_llm/_torch/speculative/eagle3.py
jhaotingc: TL;DR: If the cpu operation won't change during runtime, the engine runs exactly the same as the first time it is captured, will it be okay?

cpp/tensorrt_llm/thop/attentionOp.cpp
jhaotingc: @yweng0828 provided some context for these spec-dec mask. Will update in PR description for better documentation.

tensorrt_llm/_torch/speculative/eagle3.py
```diff
@@ -499,8 +467,8 @@ def prepare_1st_drafter_inputs(
         # get draft inputs
         input_ids = torch.concat([input_ids_ctx, input_ids_gen], dim=0)
 
-        # set use_spec_dec = true for attention.py
-        spec_metadata.use_spec_dec = True
+        # set use_spec_decoding = true for the 1st eagle head step
+        attn_metadata.use_spec_decoding = True
```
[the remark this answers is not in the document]
jhaotingc: Hi @yweng0828, sorry I'm not fully understanding the question, but I think you're talking about cases where **use_spec_decoding** needs to be `True` for the whole process including eagle layers, so no need to turn it into False.

Using the same llama 3 8b example in [this comment](https://github.com/NVIDIA/TensorRT-LLM/pull/5146#discussion_r2167314124), since this is currently a chained tree, draft engine can decode tokens auto-regressively. However, if it's Eagle2 or any tree-based spec-dec, it has to be true all the way.

The reason of this design is only because I'm not sure how eagle net spec-dec mask works now. The solution now solves the current chained-tree Eagle3, for token tree, it can be fixed in future PR.

tensorrt_llm/_torch/speculative/eagle3.py
jhaotingc: Thanks for the input! I agree with all.

For the current design, I'm thinking about Eagle1, Eagle2, Medusa, etc, that has tree-based token prediction, as well as MTP, MTP-Eagle, using chained-based tree. So your suggestions covers them all.

I further removed `has_spec_dec` since it actually doesn't need this flag, spec-dec should be enabled in SpecMetadata; and spec-dec mode should be set automatically.

cpp/tensorrt_llm/thop/attentionOp.cpp
jhaotingc: In gptAttentionPlugin, `mIsSpecDecodingEnabled` determines if spec-dec mode is enabled during plugin creation [here](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/plugins/gptAttentionCommon/gptAttentionCommon.cpp#L96).
`mUseSpecDecoding` determines if the current step / request should be run with spec-dec mode or not [here](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/plugins/gptAttentionPlugin/gptAttentionPlugin.cpp#L703)

If `mIsSpecDecodingEnabled = True`, but `mUseSpecDecoding` changed from True to False, there's an additional [prepareEnqueueGeneration](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/plugins/gptAttentionPlugin/gptAttentionPlugin.cpp#L1130) that needs to run again.

In PyTorch flow, or specifically Pytorch one-engine Eagle3, it's not necessarily to include this change, but kept these arguments since gptAttentionPlugin and thop/attentionOp.cpp uses the same common/attentionOp.cpp.

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -571,50 +566,30 @@ def attention(
...
+    qk_nope_head_dim: Optional[int], qk_rope_head_dim: Optional[int],
+    v_head_dim: Optional[int], mrope_rotary_cos_sin: Optional[torch.Tensor],
     mrope_position_deltas: Optional[torch.Tensor],
     mla_context_paged_kv: Optional[torch.Tensor],
     mla_context_kv_cache_block_offsets: Optional[torch.Tensor],
-    attention_chunk_size: Optional[int],
+    attention_chunk_size: Optional[int], use_spec_dec: bool,
+    spec_decoding_position_offsets: Optional[torch.Tensor],
+    spec_decoding_packed_mask: Optional[torch.Tensor],
+    spec_decoding_generation_lengths: Optional[torch.Tensor]
```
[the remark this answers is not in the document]
jhaotingc: Added, thanks!

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -684,50 +657,30 @@ def _(
...
```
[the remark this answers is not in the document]
jhaotingc: Added, thanks!

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -687,6 +688,7 @@ def forward(
             position_ids=position_ids,
             hidden_states=hidden_states,
             attn_metadata=attn_metadata,
+            spec_metadata=spec_metadata,
```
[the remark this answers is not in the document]
jhaotingc: This is a good suggestion, I changed the design. 
Moved the parameters into attributes of `attn_metadata`, it's more straight forward since those params are for attention backend.

cpp/tensorrt_llm/thop/attentionOp.cpp
jhaotingc: In cpp runtime, multiBlockMode is set to !specDecodingEnabled [here](https://github.com/NVIDIA/TensorRT-LLM/blob/9c012d5bf8936b040b3822b35c823460e8f63a75/cpp/tensorrt_llm/plugins/gptAttentionCommon/gptAttentionCommon.cpp#L75). 

However, there were customer complaining about it being slow, so there's another PR enabling the multi-block mode for XQA multi-query kernels. [PR 4416] (https://github.com/NVIDIA/TensorRT-LLM/pull/4416) should help both cpp and pytorch runtime.

However, it's not merged yet due to not enough bandwidth.

tensorrt_llm/_torch/attention_backend/trtllm.py
jhaotingc: Made trtllm.py api call unchanged and moved argument packing into torch custom ops.

tensorrt_llm/_torch/modules/attention.py
jhaotingc: Changed the design as comment suggested.

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -314,6 +320,22 @@ class Runner : public RunnerBase
             {
                 enqueue_params.mrope_position_deltas = mrope_position_deltas.value().data_ptr<int32_t>();
             }
+            if (op.mIsSpecDecodingEnabled && op.mUseSpecDecoding)
+            {
+                TLLM_CHECK_WITH_INFO(spec_decoding_packed_mask.has_value(), "spec_decoding_packed_mask needs value");
+                TLLM_CHECK_WITH_INFO(
+                    spec_decoding_position_offsets.has_value(), "spec_decoding_packed_mask needs value");
```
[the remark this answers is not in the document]
jhaotingc: Fixed, thanks!

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -610,7 +644,7 @@ void attention_inplace(torch::Tensor q, torch::optional<torch::Tensor> k, torch:
             host_kv_cache_block_offsets, host_kv_cache_pool_pointers, host_kv_cache_pool_mapping, cache_indirection,
             kv_scale_orig_quant, kv_scale_quant_orig, out_scale, rotary_inv_freq, rotary_cos_sin, latent_cache, q_pe,
             block_ids_per_seq, mrope_rotary_cos_sin, mrope_position_deltas, mla_context_paged_kv,
-            mla_context_kv_cache_block_offsets, softmax_stats_tensor);
+            mla_context_kv_cache_block_offsets, softmax_stats_tensor, spec_decoding_tensor_params);
```
yweng0828, line 647: Hi, although we have added this parameter here, MLA will not use this parameter now, is it right?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -849,6 +897,76 @@ def prepare_paged_context_mla(self, cached_token_lens: torch.Tensor,
...
             self.host_ctx_kv_indptr[:self.num_contexts + 1], non_blocking=True)
 
+    def update_spec_dec_param(self, is_spec_decoding_enabled, is_spec_dec_tree,
+                              is_spec_dec_dynamic_tree, max_draft_tokens):
+        # spec_dec mode should only be enabled for pre-Blackwell machines and when there's a spec-dec tree.
+        self.is_spec_decoding_enabled = is_spec_decoding_enabled and get_sm_version(
+        ) < 100
+
+        # use_spec_decoding is default to true by default, change in runtime by layers / requests
+        self.use_spec_decoding = self.is_spec_decoding_enabled
```
yweng0828, line 907: Hi, I have some questions about the assignment here. From your previous comments, I understand that `is_spec_decoding_enabled` is for the entire runtime, while `use_spec_decoding` is only for the current step.

Does the assignment here mean that as long as `is_spec_decoding_enabled=True`, then all steps will `use_spec_decoding=True`?

If this is the case, why do we need to distinguish between these two variables?

tensorrt_llm/_torch/speculative/interface.py
```diff
@@ -85,6 +85,12 @@ def extend_ctx(self, attention_backend: Type[AttentionBackend]):
                 and not (isinstance(attention_backend, TrtllmAttention)
                          and get_sm_version() == 100)) or self.is_ngram()
 
+    def attention_need_spec_dec_mode(self):
+        """
+        If true, the attention backend kernel needs to run in spec-dec mode (multi-token query mode).
+        """
+        return self.is_eagle3_one_model()
```
yweng0828, line 92: Sorry, I missed some details here. For the two-model case, does attention not need to be in spec-dec mode?

tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -255,13 +253,6 @@ def __init__(self, max_seq_len: int, config: MTPConfig):
         if config is not None:
             self.draft_len = config.num_nextn_predict_layers
 
-    def _draft_meet_max_token_stop_criteria(self, request: LlmRequest,
```
yweng0828, line 258: Hi, why are these stop conditions removed?

tensorrt_llm/_torch/speculative/eagle3.py
```diff
@@ -370,6 +376,7 @@ def forward(self, input_ids, position_ids, hidden_states, logits,
         next_draft_tokens = []
         for i in range(self.max_draft_tokens):
             hidden_states, hidden_states_to_save = draft_model.model(**inputs)
+            attn_metadata.use_spec_decoding = False
```
yweng0828, line 379: Currently the logic here is forcibly related to the chain, is it right?

That is, for the draft model, we will only use the causal mask (because it is a chain), and for the main model, we will use the spec-dec mode.

tensorrt_llm/_torch/attention_backend/trtllm.py
jhaotingc: If `is_spec_decoding_enabled=True`, `use_spec_decoding=False`, it won't do the spec-dec step, for chained tree eagle step > 1 it'll be False.
But if `is_spec_decoding_enabled=False`, it means that no matter use_spec_decoding=False or True it won't do spec-dec.
The naming is inherited from [gptAttentionPlugin](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/plugins/gptAttentionCommon/gptAttentionCommon.cpp#L76). At plugin creation, it sets is_spec_decoding_enabled, at runtime, use_spec_decoding controls the usage.

tensorrt_llm/_torch/speculative/interface.py
jhaotingc: It might need it too, haven't verify it yet. Should leave to next PR 😂.
@mikeiovine may have some insights.

tensorrt_llm/_torch/speculative/eagle3.py
jhaotingc: Yes. Slight correction, for the main model & the 1st draft model step, we use spec-dec mode.

cpp/tensorrt_llm/thop/attentionOp.cpp
jhaotingc: MLA will not use the added spec-decoding parameters, I didn't modify MLA stuff.

cpp/tensorrt_llm/thop/attentionOp.cpp
jhaotingc: It supports `int?[]`, but **not bool[]? or bool?[]** since I tried to get the spec_dec_bool_param to be also optional but ran into issue.

tensorrt_llm/_torch/speculative/mtp.py
jhaotingc: The decoding stopped when num_tokens + draft_len >= max_new_token, whereas for regular decoding it reaches max_new_token.
Say max_new_token = 32, Eagle3 stopped at 28, 29, 30 etc, but regular decoding stops at 32.
Which makes a check [here](https://github.com/NVIDIA/TensorRT-LLM/pull/5146/files#diff-5f7c2118efd2d87b6f0d6166ecce49e06c4e582a6388a264d210c73f9d4c7e03R109) guarding the same exact token output not passing.
When I removed the check in MTP sampler, it only stops when target accept the tokens, and the output would be exactly the same.

tensorrt_llm/_torch/speculative/eagle3.py
jhaotingc: Hi @yweng0828, yes, I indeed printed `use_spec_decoing` in `thop/attentionOp.cpp`, and it indeed changed according to these flags.

> even if use_spec_decoding is always True, it will not affect the calculation process

It will, the 2nd to last eagle step needs to be in regular decoding mode. The reason is that I don't know what's spec-dec mask for nth except 1st eagle step is.

yuxianq (approved):

tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -253,13 +251,6 @@ def __init__(self, args: TorchSampler.Args, *, nextn: int):
         self.mapping = None
         self.draft_len = nextn
 
-    def _draft_meet_max_token_stop_criteria(self, request: LlmRequest,
-                                            num_tokens: int, beam_idx: int):
-        if self._meet_max_token_stop_criteria(request,
-                                              num_tokens + self.draft_len):
```
yweng0828, line 259: I think we only need to remove `draft_len` here to solve the problem of short output, because the accepted tokens have been considered [above](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/speculative/mtp.py#L302).

cc @netanel-haber for more comments.

tensorrt_llm/_torch/speculative/eagle3.py
jhaotingc: image
Got an nsys for the captured steps (step after several iteration). It seems that with the current code, nth eagle step will go to MMHA, while MMHA is not runnable with use_spec_decoding, [got a check here](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/common/attentionOp.cpp#L2158).

So, the 2nd to last eagle step will be running in regular decoding mode, for the correctness in terms of this PR, but for token tree, it need to be in spec-dec mode and with an appropriated mask.

yweng0828 (approved):

jhaotingc:
Blocked by random CI issue, rebasing and trying again.

byshiue (approved):

PR #5151, opened 2025-06-12
Feat/pytorch vswa kvcachemanager

qixiang-99:
# \[Feature] PyTorch VSWA KvCacheManager Support

## Caveat
There is a [WAR](https://github.com/NVIDIA/TensorRT-LLM/pull/5151/files#diff-63c50d73505d372036cf1580e0684c27381bb26b41b49f2ce8713bdf6744d3b3R274-R376) in this PR that is specific for `Gemma3ForCausalLM`, **any model that wants to use VSWA KvCacheManager should modify accordingly.**

### TODOs:

* [x] Expand detailed description.
* [x] Clean up logging.
* [x] Finalize GSM8K benchmark scores comparing HF's and current main's Gemma 3 1b and add it.
* [x] Fix/Update current main Gemma 3 1b test
  * [x] set `enable_block_resue=False`
  * [x] add chunk unit size restriction for chunked prefill correctness 

### TL;DR
This PR adds:
1. PyTorch runtime support for VSWA KvCacheManager
2. Fix Gemma 3 1b integration test by setting `enable_block_reuse=False` explicitly because current (V)SWA is not compatible with block reuse(`enable_block_reuse=True` by default)
3. Fix Chunked prefill + (V)SWA bug, though this would not be necessary in the future. 

### Background

Previously, Variable Sliding Window Attention (VSWA) support was limited to the C++ KvCacheManager within the C++ runtime. The existing PyTorch implementation only supports a fixed `max_seq_len` window size uniformly across all attention layers. This Pull Request extends PyTorch's KvCacheManager to handle distinct attention window sizes for different layers.

### Implementation Details

To integrate VSWA into PyTorch's KvCacheManager, key changes primarily involve preparing inputs compatible with the underlying C++ Pybind class.

#### Blocks per Window

A significant [update](https://github.com/NVIDIA/TensorRT-LLM/pull/4667) introduces the `blocks_per_window` parameter in the KvCacheManager constructor, which specifies the number of memory blocks in primary and secondary memory for each distinct window size.

Currently, PyTorch's KvCacheManager uses its own implementation (`calculate_max_num_blocks`) that supports a single window size for determining block allocation. After discussion with the team, to ensure consistency between C++ and PyTorch implementations, this PR adds a new wrapper function, `calculate_max_num_blocks_from_cpp`, directly invoking C++ KvCacheManager’s calculation logic.

Notably, we apply the C++ calculation **exclusively** in VSWA contexts, retaining PyTorch's original implementation elsewhere. The rationale for this approach includes:

* Ensuring current PyTorch models remain unaffected.
* Avoiding potential performance regressions across existing models.
* Limiting the complexity of the current PR, deferring broader integration to subsequent detailed evaluations.

Future iterations can progressively transition toward broader adoption of the C++ calculation method following thorough performance analyses.

#### Inputs for C++ `calculate_max_num_blocks`

Given that the C++ `calculate_max_num_blocks` method is a Pybind class member, its inputs require precise types, necessitating conversions in the PyTorch implementation:

##### ModelConfig

PyTorch path utilizes its own `ModelConfig` class extensively. This PR enhances the existing `get_bindings_model_config` function to convert PyTorch’s `ModelConfig` to the necessary `ModelConfigCpp`.

##### ~~KvCacheConfig~~ Solved by https://github.com/NVIDIA/TensorRT-LLM/pull/5384

~~The current PyTorch codebase predominantly passes instances of `tensorrt_llm.bindings.executor.KvCacheConfig` instead of the expected `tensorrt_llm.bindings.KvCacheConfig` to KvCacheManager init function. Although structurally similar, these mismatches lead to type errors when invoking C++ methods. Given the widespread nature of this practice, this PR selectively converts `executor.KvCacheConfig` to `bindings.KvCacheConfig` only as required for compatibility with C++ method calls.~~

### GSM8K Tests for Gemma 3 1b

To verify the functional correctness of PyTorch VSWA KvCacheManager, this PR adds the GSM8K task to the Gemma-3-1b integration tests. To ensure the reference GSM8K scores are representative, GSM8K evaluations were conducted using  Huggingface's implementation of Gemma-3-1b.

Ultimately, an average accuracy of **25.52%** was selected as the reference score. Detailed evaluation steps are documented below. (`avg score = mean(flexible-extract acc, strict-match acc)` )

####  lm_eval with HF Gemma-3-1b GSM8K
This utilizes lm_eval from its [repo](https://github.com/EleutherAI/lm-evaluation-harness).
Run this command:
```bash
lm_eval --model hf \
--model_args pretrained=/home/scratch.trt_llm_data/llm-models/gemma/gemma-3-1b-it/,trust_remote_code=True \
--tasks gsm8k \
--batch_size auto \
--device cuda:0 \
--output_path $OUTPUT_DIR

```

Output:

```bash
2025-06-12:00:00:21,259 INFO     [lm_eval.loggers.evaluation_tracker:209] Saving results aggregated
hf (pretrained=/home/scratch.trt_llm_data/llm-models/gemma/gemma-3-1b-it/,trust_remote_code=True), gen_kwargs: (None), limit: None, num_fewshot: None, batch_size: auto
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.2563|±  | 0.012|
|     |       |strict-match    |     5|exact_match|↑  |0.2540|±  | 0.012|

```

## Test Coverage
Use Gemma 3 1b integration test as this is the only VSWA models supported:
`tests/integration/defs/accuracy/test_llm_api_pytorch.py::TestGemma3_1BInstruct`

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -23,9 +23,9 @@ from .kv_cache_transceiver import AttentionTypeCpp, create_kv_cache_transceiver
 from .llm_request import ExecutorResponse
 from .model_engine import PyTorchModelEngine
 from .py_executor import PyExecutor
-from .resource_manager import (KvCacheConfigCpp, KVCacheManager,
-                               MambaHybridCacheManager, PeftCacheManager,
-                               ResourceManager, ResourceManagerType)
+from .resource_manager import (KVCacheManager, MambaHybridCacheManager,
+                               PeftCacheManager, ResourceManager,
+                               ResourceManagerType)
 from .sampler import EarlyStopSampler, TorchSampler, TRTLLMSampler
 from .scheduler import (BindCapacityScheduler, BindMicroBatchScheduler,
                         SimpleScheduler)
@@ -348,8 +348,7 @@ class KvCacheCreator:
             ) if is_vswa else None
 
             kv_cache_manager = KVCacheManager(
-                # NOTE: from tensorrt_llm.bindings.executor.KvCacheConfig to tensorrt_llm.bindings.KvCacheConfig
-                KvCacheConfigCpp(executor_config.kv_cache_config),
+                executor_config.kv_cache_config,
                 tensorrt_llm.bindings.internal.batch_manager.CacheType.SELF,
                 num_layers=num_hidden_layers,
                 num_kv_heads=num_key_value_heads,
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -23,9 +23,9 @@ from .kv_cache_transceiver import AttentionTypeCpp, create_kv_cache_transceiver
 from .llm_request import ExecutorResponse
 from .model_engine import PyTorchModelEngine
 from .py_executor import PyExecutor
-from .resource_manager import (KVCacheManager, MambaHybridCacheManager,
-                               PeftCacheManager, ResourceManager,
-                               ResourceManagerType)
+from .resource_manager import (KvCacheConfigCpp, KVCacheManager,
+                               MambaHybridCacheManager, PeftCacheManager,
+                               ResourceManager, ResourceManagerType)
 from .sampler import EarlyStopSampler, TorchSampler, TRTLLMSampler
 from .scheduler import (BindCapacityScheduler, BindMicroBatchScheduler,
                         SimpleScheduler)
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -180,20 +180,6 @@ class KvCacheCreator:
             estimating_kv_cache = True
             self._executor_config.kv_cache_config.max_tokens = self._get_token_num_for_estimation(
             )
-            maw = self._executor_config.kv_cache_config.max_attention_window
-            if maw is not None:
-                unique_windows = list(set(maw))
-                if len(unique_windows) > 1:
-                    # NOTE: For sliding window, we need to allocate more tokens(max_num_tokens + window_size) for each window, need to update if this requirement changes
-                    assert len(
-                        unique_windows
-                    ) == 2, "Sliding window with more than 2 window sizes has not been tested"
-                    # sliding window is the one that is not the executor_config.max_seq_len
-                    sliding_window = unique_windows[0] if unique_windows[
-                        0] != self._executor_config.max_seq_len else unique_windows[
-                            1]
-                    self._executor_config.kv_cache_config.max_tokens += (
-                        self._executor_config.max_num_tokens + sliding_window)
         return estimating_kv_cache
 
     def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
@@ -358,6 +344,7 @@ class KvCacheCreator:
             is_vswa = executor_config.kv_cache_config.max_attention_window is not None and len(
                 set(executor_config.kv_cache_config.max_attention_window)) > 1
             binding_model_config = model_engine.model.model_config.get_bindings_model_config(
+                tokens_per_block=executor_config.tokens_per_block
             ) if is_vswa else None
 
             kv_cache_manager = KVCacheManager(
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -180,6 +180,20 @@ class KvCacheCreator:
             estimating_kv_cache = True
             self._executor_config.kv_cache_config.max_tokens = self._get_token_num_for_estimation(
             )
+            maw = self._executor_config.kv_cache_config.max_attention_window
+            if maw is not None:
+                unique_windows = list(set(maw))
+                if len(unique_windows) > 1:
+                    # NOTE: For sliding window, we need to allocate more tokens(max_num_tokens + window_size) for each window, need to update if this requirement changes
+                    assert len(
+                        unique_windows
+                    ) == 2, "Sliding window with more than 2 window sizes has not been tested"
+                    # sliding window is the one that is not the executor_config.max_seq_len
+                    sliding_window = unique_windows[0] if unique_windows[
+                        0] != self._executor_config.max_seq_len else unique_windows[
+                            1]
+                    self._executor_config.kv_cache_config.max_tokens += (
+                        self._executor_config.max_num_tokens + sliding_window)
         return estimating_kv_cache
 
     def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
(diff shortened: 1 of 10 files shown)


tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -175,6 +176,20 @@ def try_prepare_estimation(self) -> bool:
...
+                if len(unique_windows) > 1:
+                    # NOTE: For sliding window, we need to allocate more tokens(max_num_tokens + window_size) for each window, need to update if this requirement changes
+                    assert len(
+                        unique_windows
+                    ) == 2, "Sliding window with more than 2 window sizes has not been tested"
+                    # sliding window is the one that is not the executor_config.max_seq_len
+                    sliding_window = unique_windows[0] if unique_windows[
+                        0] != self._executor_config.max_seq_len else unique_windows[
+                            1]
+                    self._executor_config.kv_cache_config.max_tokens += (
```
netanel-haber, line 191: I don't understand this?
The result will be:
`self._get_token_num_for_estimation() + self._executor_config.max_num_tokens + sliding_window`.
What does this calulcation correspond to?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -158,46 +162,93 @@ def __init__(
...
+            logger.info(
+                f"Adjusted attention window size to {self.max_seq_len} in blocks_per_window"
             )
-            max_attention_window = max_atten_window_upper_bound
-            self.max_seq_len = max_atten_window_upper_bound
 
-        self.max_attention_window = max_attention_window if kv_cache_type == CacheTypeCpp.SELF else self.max_seq_len
+        # Set up temp_attention_window_inputs only for VSWA or when needed
+        temp_attention_window_inputs = None
+        if is_vswa:  # Only create when necessary
```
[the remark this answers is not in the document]
netanel-haber: @Funatiq - cc

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -501,6 +565,150 @@ def get_kv_cache_stats(self):
...
+        window_size_to_layers_map = defaultdict(list)
+
+        if not self.max_attention_window_vec:
+            # This case should ideally be prevented by earlier config validation.
+            # If num_local_layers is 0, an empty map is fine.
+            if self.num_local_layers > 0:
+                raise ValueError(
+                    "max_attention_window_vec cannot be empty if there are local layers."
+                )
+            return {
```
netanel-haber, line 584: If this is unreachable, raise an exception.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -501,6 +565,150 @@ def get_kv_cache_stats(self):
...
+            }  # Return an empty dict if no local layers or if somehow vec is empty and no layers.
+
+        # Treat max_attention_window_vec as a repeating pattern.
+        # This also correctly handles the case where len(self.max_attention_window_vec) == 1.
```
netanel-haber, line 588: I think it is more readable if single SWA is treated as a special case + early return, which also guarantees that the default path [the more common case, not VSWA] doesn't get tangled up in the messier logic below

netanel-haber: i.e., 
```python
if len(max_attention_window_vec) == 1:
    return {max_attention_window_vec[0]: list(range(self.num_local_layers))}
```

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -501,6 +565,150 @@ def get_kv_cache_stats(self):
...
+            config=kv_cache_config,
+            is_cross_attention=False,  #TODO: support cross attention
+            dtype=self.dtype,
+            model_config=model_config,
+            world_config=world_config_cpp,
+            window_size_to_layers=window_size_to_layers,
+            allotted_primary_mem_bytes=primary_pool_memory_bytes,
+            allotted_secondary_mem_bytes=secondary_pool_memory_bytes,
+            extra_cost_memory=extra_cost_memory,
+            kvFactor=self.kv_factor,
```
netanel-haber, line 642: Not your fault, but I would just fix `kvFactor` to be kv_factor in the bindings at this point

tensorrt_llm/_torch/pyexecutor/resource_manager.py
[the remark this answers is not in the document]
qixiang-99: You're right, it would be better to generalize this by creating `temp_attention_window_inputs` for the SWA case rather than limiting it only to VSWA.

Thanks also for pointing out the confusion around `max_num_tokens`. After checking both PyT and C++ implementations, I noticed PyT has a similar "confusion"—there are two separate `max_num_tokens`: one in the runtime config and another in the model config. I'll verify the logic on the PyT sid.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
qixiang-99: Fixed, thanks.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
qixiang-99: Agreed, modified, thank!

Funatiq:
Re: `KvCacheConfig` conversion: I'm trying to remove this in #5384.

qixiang-99:
> #5384

Thanks for the update @Funatiq! This change will help improve code consistency. Once you've merged https://github.com/NVIDIA/TensorRT-LLM/pull/5384, I'll update here accordingly.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -175,6 +176,20 @@ def try_prepare_estimation(self) -> bool:
...
             self._executor_config.kv_cache_config.max_tokens = self._get_token_num_for_estimation(
             )
+            maw = self._executor_config.kv_cache_config.max_attention_window
+            if maw is not None:
+                unique_windows = list(set(maw))
+                if len(unique_windows) > 1:
+                    # NOTE: For sliding window, we need to allocate more tokens(max_num_tokens + window_size) for each window, need to update if this requirement changes
+                    assert len(
+                        unique_windows
+                    ) == 2, "Sliding window with more than 2 window sizes has not been tested"
```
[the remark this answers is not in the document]
qixiang-99: removed ; )

netanel-haber (approved):

dcampora (approved):

qixiang-99:
> Nit: Adjust the PR description for `KvCacheConfig` since it always uses the `executor` one now.

Updated the description, thanks! @Funatiq

PR #5166, opened 2025-06-12
tests: Move stress tests to be Post-Merge only

amirkl94:
# L0 Tests reduction

As part of pre-merge CI optimization, moving stress tests to be Post-Merge only.


omera-nv (commented):
LGTM, but can we get an approval from the tests' author too?

chzblych (approved):

PR #5182, opened 2025-06-13
Add debug hook to support dump tensor data and add new debug functions easily

HuiGao-NV:
Add debug hook to support dump tensor data and add new debug functions easily.

To enable to dump tensors' data,
from tensorrt_llm._torch.debug.debug_hook import debugger_addon, register_tensor_dump_hook
with debugger_addon(model, DATA_FOLDER):
    register_tensor_dump_hook()
    model.forward()
    
The dumped data are put under DATA_FOLDER/rank[ID]/....
The data file name is in the pattern:
[LOOP_COUNT].[model_name]-[OPIDX_IN_MODEL].[OPNAME]-[OPIDX_IN_PRE_OP].[OPNAME]-[input|output].[PARA_NAME].pt
suc as 1.LlamaModel-24.LlamaDecoderLayer-2.LlamaAttention-2.Linear-1.AllReduce-input.input.pt.


tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def __init__(self, dest_folder=None):
+        self.pre_forward_actions = {}
+        self.after_forward_actions = {}
+
+        self.layer_names = []
+        self.layer_inner_counter = []
+
+        self.module_forward_hook_handle = None
+        self.module_forward_pre_hook_handle = None
```
QiJune, line 32: Since the outputs of layer n can be the inputs of layer n+1, do we need to dump the inputs and outputs of all layers at the same time? I guess there will be lots of duplicated results.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,461 @@
...
+
+        self.forward_hook_handles = {}  # module to handlers
+        self.forward_pre_hook_handles = {}
+        self.log_folder = dest_folder
+        self.is_forward_pre = True
+        self.dump_style: DumpStyle = DumpStyle.BINARY
+        self.log_folder_inited: bool = False
```
QiJune, line 39: Let's unifiy naming style, `is_forward_pre`, `is_log_folder_inited`

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,461 @@
...
+    def set_log_folder(self, log_folder):
+        self.log_folder = log_folder
+        self.log_folder_inited = False
+
+    def _init_log_folder(self):
+        if self.log_folder_inited:
+            return
+
+        if self.log_folder is None:
+            import os
```
QiJune, line 50: Please move `import os` to the beginning of the file.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,461 @@
...
+# Note: this method need user to disable debug by calling disable_debug
+def enable_debug(model, dest_folder: Optional[str] = None, filter=None):
+    debug_ctx = get_current_debug_ctx()
+    assert debug_ctx is None, "DebugContext shall be None when enable debugger context."
+    debug_ctx = DebuggerContext(dest_folder)
+    set_current_debug_ctx(debug_ctx)
+
+    debug_ctx.get_current_modules_tree().clear()
+    debug_ctx.get_module_indices_tree().clear()
+    for name, submodule in model.named_modules():
```
QiJune, line 315: If a `MLP` module contains 2 `Linear` modules, then, the output of `MLP` module should be the same with the output of the second `Linear` module. Is my understanding correct? If so, there will be duplicated dumpled tensors.

HuiGao-NV: Yes. This is the same question as the first one.
We can skip dump tensors for pre_forward_hook by not regitster pre_forward_hook.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: Yes. We can skip dump tensors for pre_forward_hook by not regitster pre_forward_hook. I make change in this way.

However there may be some special cases because only modules inheriting from torch.nn.Module will be injected with this hook. If there is one OP which is not a module existing between 2 modules, then the output of the first module is changed before next module's hook is invoked. Then, their values are different.
Current implementation is a simple version to dump tensors. There is no graph tracing in which it can be checked if one tensor is already dumped at runtime.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,460 @@
...
+from pathlib import Path
+from typing import Optional
+
+import torch
+import torch.nn as nn
+
+import tensorrt_llm
+
+
+class DumpStyle(enum.IntEnum):
```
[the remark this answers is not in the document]
HuiGao-NV: Removed it and dumpstyle in debugContext

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,460 @@
...
+    BINARY = 1
+    TEXT = 2
+
+
+class DebugFeatures(enum.IntEnum):
```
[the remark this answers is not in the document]
HuiGao-NV: Remove it here and enable hook functions.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,501 @@
...
+
+
+def disable_debug_all_modules():
+    assert get_current_debug_ctx() is not None, ""
+    debug_ctx.clear_state()
+    set_current_debug_ctx(None)
+
+
+@contextmanager
+def debugger_addon_all_modules(dest_folder: Optional[str] = None):
```
QiJune, line 234: In which scenario will this API be called?  Considering we already provide `debug_mode` API, I suggest we remove this `debugger_addon_all_modules`/`enable_debug_all_modules` related APIs now to keep the debug module simple at the beginning. If we really need such feature in the future, we can add it back later.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,501 @@
...
+    Returns:
+        None
+    """
+    try:
+        enable_debug(dest_folder, filter)
+    finally:
+        disable_debug()
+
+
+def pre_forward(module: nn.Module, args, kwargs):
```
QiJune, line 259: Should it be `*args, **kwargs`?

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,501 @@
...
+
+    debug_ctx.get_module_indices_tree()[
+        len(debug_ctx.get_current_modules_tree()) -
+        1] = debug_ctx.get_module_indices_tree()[
+            len(debug_ctx.get_current_modules_tree()) - 1] + 1
+    debug_ctx.do_actions(module, args, debug_ctx.get_pre_forward_action())
+    return None
+
+
+def after_forward(module: nn.Module, args, kwargs, output):
```
QiJune, line 292: The same, `*args, **kwargs`

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,501 @@
...
+    name = module.name if hasattr(module, "name") else module.__class__.__name__
+    old_name = debug_ctx.get_current_modules_tree().pop(-1)
+    assert name == old_name, "module mismatch"
+
+    debug_ctx.get_module_indices_tree().pop(-1)
+    debug_ctx.tensor_counter = 0
+    return None
+
+
+def enable_debug(model, dest_folder: Optional[str] = None, filter=None):
```
QiJune, line 319: Please add type annoations to `model` and `filter`

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,501 @@
...
+        return args
+
+    return None
+
+
+# Below is one hook for dump tensors.
+# Normally, if you want implement one hook, you need to implement one filter by
+# inheriting from base class Filter and one function which defines what to do,
+# such as dump data, modify data, inject actions, etc.
+class DumpTensorFilter(Filter):
```
QiJune, line 437: For class comments, we perfer this style: https://github.com/NVIDIA/TensorRT-LLM/blob/d13d2f460d4b545f7902ecb008d24d81199349bf/tensorrt_llm/llmapi/llm_args.py#L252-L271

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,501 @@
...
+
+    def __init__(self):
+        pass
+
+    def filter(self, module: nn.Module, debug_ctx: DebuggerContext):
+        return True
+
+
+def dump_tensor(module: nn.Module, data_tensor, debug_ctx: DebuggerContext):
+    global input_tensor_names
```
QiJune, line 447: Can we maintain `input_tensor_names`  in `debug_ctx` to avoid global variable?

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,501 @@
...
+        self.module_forward_hook_handle = None
+        self.module_forward_pre_hook_handle = None
+
+        self.forward_hook_handles = {}  # module to handlers
+        self.forward_pre_hook_handles = {}
+        self.log_folder = dest_folder
+        self.is_pre_forward = True
+        self.is_log_folder_inited: bool = False
+
+    def set_log_folder(self, log_folder):
```
QiJune, line 33: This function seems not used. `self.log_folder` is set when initializing a DebuggerContext object.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,501 @@
...
+        self.log_folder = log_folder
+        self.is_log_folder_inited = False
+
+    def _init_log_folder(self):
```
QiJune, line 37: Could we call  `_init_log_folder` inside `__init__` directly?

```
def __init__(xxx):
    xxx
    self.log_folder = self._init_log_folder(log_folder)
     xxx
```

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: Removed.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: Because it co-work with above removed function set_log_folder. User can pass in dest_folder in __init__ or set it after creating debugContext, such as reuse it for another debugger purpose (However, it isn't supported completely.)
Then, the safe place to update it is when it's accessed by get_log_folder.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: Changed.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: Added.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: This signature is from https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_hook.

I think the reason the hook code in nn.Module does pass a dict here as a parameter but not for kwars of the hook function we regitster. 
One more thing is if we use **kwargs, the after_forward hook function can't accept output as the last parameter which is required by nn.Module.register_forward_hook.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: The same reason as above.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: If we want to enable hooks on all modules but do filter on how to execute hook functions, we need to use this one.

HuiGao-NV: Removed

tensorrt_llm/_torch/debug/debug_hook.py
QiJune: Got it. 

But I think it's an "ambiguous" goal to support user-defined action hooks at the very beginning. I would suggest we control the scope of the first implementation of debug tool: only provides dumping tensors (and filters), and enable it by default. 

To put it more detail, users don't need to call `register_tensor_dump_hook` by themselves, it enabled in `with debug_mode(xxx)` by default.  They can pass filters to control the tensors.

tensorrt_llm/_torch/debug/debug_hook.py
QiJune: Seems the files are not updated yet.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: Enable dump_tensor by default now.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,385 @@
...
+            handler.remove()
+
+        for _, handler in self.forward_pre_hook_handles.items():
+            handler.remove()
+
+        self.forward_hook_handles.clear()
+        self.forward_pre_hook_handles.clear()
+
+    def register_pre_forward_action(self, filter, action):
+        self.pre_forward_actions[filter] = action
```
QiJune, line 88: To make a filter be the key of a dictionary means the `Filter` class must be hashable. Do we need to look up an action giving a filter? If no, I would suggest we use tuple to get ride of such  constraint.

```python
self.pre_forward_actions = []
self.pre_forward_actions.append((filter, action))
```

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,385 @@
...
+class Filter:
+
+    def __init__(self):
+        pass
+
+    def filter(self, module: nn.Module, debug_ctx: DebuggerContext):
+        raise NotImplementedError("Need to implement filter interface.")
+
+
+debug_ctx = None
```
QiJune, line 124: If possbible, it's not recommended to use global variable. Can we make `DebuggerContext` be a Singleton class?  Some reference:

```python
class Singleton:
    _instance = None

    def __new__(cls):
        if cls._instance is None:
            cls._instance = super().__new__(cls)
        return cls._instance

# Usage
singleton1 = Singleton()
singleton2 = Singleton()
print(singleton1 is singleton2)  # True (same instance)
```

tensorrt_llm/_torch/debug/debug_hook.py
QiJune: Thanks. 

I am still a bit of confused, `tensor_counter` and `input_tensor_names` are reset at the end of `dump_tensor`. Then, why not creating 2 local variables inside `dump_tensor`?

```python
def dump_tensor(module: nn.Module, data_tensor, debug_ctx: DebuggerContext):
    input_tensor_names = []
    tensor_counter = 0
    xxx
```

tests/unittest/_torch/debugger/test_debugger_addon.py
```diff
@@ -0,0 +1,210 @@
...
+                                   position_ids=position_ids,
+                                   attn_metadata=attn_metadata,
+                                   return_context_logits=True)
+        self.assertEqual(input_ids.shape, logits.shape[:-1])
+
+        kv_cache_manager.shutdown()
+
+
+if __name__ == "__main__":
+    unittest.main()
```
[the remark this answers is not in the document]
HuiGao-NV: Removed.

tests/unittest/_torch/debugger/test_debugger_addon.py
```diff
@@ -0,0 +1,210 @@
...
+    debug_ctx.register_after_forward_action(EmptyFilter(), print_module_name)
+
+
+class TestDebugger(unittest.TestCase):
+
+    def test_with_llama(self):
+        config_dict = deepcopy(LLAMA_3_1_8B_CONFIG)
+        # 8B * sizeof(float16) plus some extra for activations
+        mem_for_full_model = (2 + 1) * 8 * 2**(30)
+        reduce_llama_config(mem_for_full_model, config_dict)
```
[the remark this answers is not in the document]
HuiGao-NV: Thanks.
Removed.

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: By my understanding, it isn't very convenient to make it be a singleton because it requires one dest_folder as parameter. It isn't trival to handle multi-times to call instance() with different dest_folder values.
And it shall be created and reset inside this file, I think we can leave it as it is now and enhance it later when necessary.

tests/unittest/_torch/debugger/test_debugger_addon.py
```diff
@@ -0,0 +1,203 @@
...
+@dataclass(repr=False)
+class Scenario:
+    backend: str
+    use_cuda_graph: bool = False
+
+    def __repr__(self) -> str:
+        return f"backend:{self.backend.lower()}-use_cuda_graph:{self.use_cuda_graph}"
+
+
+def reduce_llama_config(mem_for_full_model: int,
```
QiJune, line 64: This function can be removed.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,374 @@
...
+        pass
+
+    def filter(self, module: nn.Module, debug_ctx: DebuggerContext):
+        return True
+
+
+def dump_tensor(module: nn.Module, data_tensor, debug_ctx: DebuggerContext):
+    global tensor_counter
+
+    input_tensor_names = []
```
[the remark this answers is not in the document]
HuiGao-NV: Removed.

tensorrt_llm/_torch/debug/debug_hook.py
```diff
@@ -0,0 +1,374 @@
...
+        torch.save(t, file_path)
+
+    def dump(t):
+        if isinstance(t, torch.Tensor):
+            dump_tensor_data(t)
+        elif isinstance(t, tuple) or isinstance(t, list):
+            for _t in t:
+                dump(_t)
+
+    tensor_counter = 0
```
QiJune, line 366: Please remove this line.

QiJune (approved):

tensorrt_llm/_torch/debug/debug_hook.py
HuiGao-NV: Removed.

PR #5202, opened 2025-06-13
[fix][test] Speedup Nemotron NAS unittests

omera-nv:
# Speedup Nemotron NAS unittests
Nemotron NAS's class is defined using remote code in transformers, so we require a chackpoint to read the class from.
Turns out `transformers` reads all of the model weights even if we set the device to `meta`, which takes a few minutes since this is a 51B model.
Instead, use `get_class_from_dynamic_module` to get the class without reading the model weights, which should speed these test cases from 2-4 minutes to a few seconds.


Naveassaf (approved):
Wow, awesome fine. 

I feel like this warrants a bugfix in `transformers` :sweat_smile:

PR #5241, opened 2025-06-16
[fix][test] remove duplicate test runs

omera-nv:
# Remove duplicate L0 tests

Remove some pytorch test runs that are duplicated across multiple hardwares.
Exceptions:
* torch unittests run on both H100 and B200
* llama, deepseek and mixtral tests run on both H100 and B200
* autodeploy tests run on both B200 and A30 (need to verify if we can remove one of these)


QiJune (approved):
LGTM

PR #5265, opened 2025-06-17
Fix : fix build for sm120

peaceh-nv:
# PR title

[TRTLLM-1000][feat] Support a new feature about cache manager

## Description

1. Add missing build env for SM120
2. New ds min latency kernel needs shared memory larger than 193kB, new SM120 hardware only has 48kB

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_deepseekv3.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv3.py
@@ -199,9 +199,8 @@ class DeepseekV3Linear(Linear):
                      lora_params: Optional[dict] | None = None,
                      layer_idx: Optional[int] | None = None):
         num_tokens = input.shape[0]
-        if (
-                not self.has_any_quant
-        ) and num_tokens >= 1 and num_tokens <= 16 and get_sm_version() != 120:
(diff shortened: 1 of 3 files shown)


kaiyux (approved):

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -198,7 +199,9 @@ def apply_linear(self,
                      lora_params: Optional[dict] | None = None,
                      layer_idx: Optional[int] | None = None):
         num_tokens = input.shape[0]
-        if (not self.has_any_quant) and num_tokens >= 1 and num_tokens <= 16:
+        if (
+                not self.has_any_quant
+        ) and num_tokens >= 1 and num_tokens <= 16 and get_sm_version() != 120:
```
hyukn, line 204: nit: `1 <= num_tokens <= 16`
nit: Equivalent to `if not self.has_any_quant and ...`

hyukn (approved):

tensorrt_llm/_torch/models/modeling_deepseekv3.py
peaceh-nv: Fixed.

PR #5266, opened 2025-06-17
Combine MNNVL allreduce with trtllm.allreduce and unify strategy selection logic.

HuiGao-NV:
Combine MNNVL allreduce with trtllm.allreduce and unify strategy selection logic.

### DIFF
--- a/cpp/tensorrt_llm/thop/allreduceOp.cpp
+++ b/cpp/tensorrt_llm/thop/allreduceOp.cpp
@@ -18,6 +18,7 @@
 #include "tensorrt_llm/common/cudaUtils.h"
 #include "tensorrt_llm/common/customAllReduceUtils.h"
 #include "tensorrt_llm/common/dataType.h"
+#include "tensorrt_llm/common/linkSupportUtil.h"
 #include "tensorrt_llm/common/mcastDevMemUtils.h"
 #include "tensorrt_llm/common/opUtils.h"
 #include "tensorrt_llm/kernels/communicationKernels/allReduceFusionKernels.h"
@@ -55,6 +56,12 @@ namespace torch_ext
 {
 
 #if ENABLE_MULTI_DEVICE
+
+at::Tensor mnnvlTwoShotAllReduce(
+    at::Tensor const& input, at::Tensor& comm_buffer, at::Tensor& buffer_flags, bool wait_for_results);
+
+std::vector<torch::Tensor> twoShotRMSNorm(torch::Tensor const& comm_buf, torch::Tensor const& gamma, double epsilon,
+    torch::Tensor const& residual, torch::Tensor& buffer_flags);
 
 namespace
 {
@@ -159,21 +166,14 @@ public:
     std::vector<torch::Tensor> run(torch::Tensor const& input, torch::optional<torch::Tensor> const& residual,
         torch::optional<torch::Tensor> const& norm_weight, torch::optional<torch::Tensor> const& scale,
         torch::optional<torch::Tensor> const& bias, bool trigger_completion_at_end,
-        torch::optional<torch::Tensor> workspace) noexcept
+        torch::optional<torch::Tensor> workspace, torch::optional<torch::Tensor> mnnvl_buffer_flag) noexcept
     {
-        size_t size = input.numel();
-        size_t seq_len = input.size(0);
-
-        // If strategy is set to UB, UB must be used as UB impl output is special and cannot be used
-        // by others.
-        AllReduceStrategyType runtime_strategy = getRuntimeStrategy(seq_len, size);
-
         // Log runtime strategy
         auto const rank = COMM_SESSION.getRank();
-        logRunTimeStrategy(runtime_strategy, rank);
+        logRunTimeStrategy(mStrategy, rank);
 
         // Dispatch to different allreduce implementations
-        switch (runtime_strategy)
+        switch (mStrategy)
         {
         case AllReduceStrategyType::UB: return runUBAllReduce(input, residual, norm_weight, scale, bias);
         case AllReduceStrategyType::NCCL: return runNCCLAllReduce(input, residual, norm_weight, scale, bias);
@@ -181,9 +181,11 @@ public:
         case AllReduceStrategyType::ONESHOT:
         case AllReduceStrategyType::TWOSHOT:
             return runFusionAllReduce(
-                input, residual, norm_weight, scale, bias, trigger_completion_at_end, workspace, runtime_strategy);
+                input, residual, norm_weight, scale, bias, trigger_completion_at_end, workspace, mStrategy);
         case AllReduceStrategyType::LOWPRECISION:
             return runLowPrecisionAllReduce(input, residual, norm_weight, scale, bias);
+        case AllReduceStrategyType::MNNVL:
+            return runMNAllReduce(input, residual, norm_weight, workspace, mnnvl_buffer_flag);
         default: TORCH_CHECK(false, "Invalid runtime strategy"); return {};
         }
     }
@@ -194,7 +196,9 @@ public:
         mNcclComm = getComm(mGroup);
         if (mStrategy != AllReduceStrategyType::NCCL && mStrategy != AllReduceStrategyType::UB)
         {
-            initGroupTopology();
+            auto supports = tensorrt_llm::common::initGroupTopology(mGroup);
+            mIsNVLINKSupported = supports[0];
+            mIsP2PSupported = supports[1];
         }
 
         TLLM_LOG_TRACE("%s stop for rank %d", __PRETTY_FUNCTION__, COMM_SESSION.getRank());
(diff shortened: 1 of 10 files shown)


tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -416,25 +352,208 @@ def __init__(self,
...
+    def _prepare_for_strategy(self, strategy, input):
+        if strategy == AllReduceStrategy.MNNVL:
+            mcast_buffer_mnnvl, buffer_mnnvl, buffer_flags_mnnvl, max_num_elements_mnnvl = get_allreduce_mnnvl_workspace(
+                self.mapping, self.dtype)
+            self.workspace = buffer_mnnvl.view(3, 2, -1, input.shape[-1])
+            self.buffer_flags_mnnvl = buffer_flags_mnnvl
+        else:
+            if strategy == AllReduceStrategy.LOWPRECISION:
+                allocate_low_presicion_allreduce_workspace(self.mapping)
+            self.workspace = get_allreduce_workspace(self.mapping)
```
liji-nv, line 555: This is accessing thread local storage so it will definitely run into 1 runtime allocation as warm up and executor are not using same thread.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -416,25 +352,208 @@ def __init__(self,
...
+        return False
+
+    def get_runtime_strategy(self, input, allreduce_params):
+        strategy = self.strategy
+        if strategy != AllReduceStrategy.AUTO:
+            if self._strategy_supports(strategy, input,
+                                       allreduce_params.fusion_op):
+                return strategy
+            elif strategy == AllReduceStrategy.LOWPRECISION or strategy == AllReduceStrategy.MNNVL or strategy == AllReduceStrategy.UB:
+                strategy = AllReduceStrategy.AUTO
```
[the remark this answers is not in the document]
HuiGao-NV: In before implementation, when strategy is UB, exception will be thrown if dtype is not supported by UB kernel.
For LOWPRECISION, the origin strategy is reset to AUTO so that it will be dispatched again to an available strategy in CPP (NCCL or MINLATENCY).
For MNNVL, in forward of python code, if it doesn't support dtype or message size, then return None and then the action is fallback in python to AllReduce (no MNNVL support). 

So the behavior keeps the same for MNNVL and LOWPRECISION.
Only for UB, the difference of behavior is that allreduce fallback to other strategy if dtype is not supported.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -416,25 +352,208 @@ def __init__(self,
...
+
+    def _prepare_for_strategy(self, strategy, input):
+        if strategy == AllReduceStrategy.MNNVL:
+            mcast_buffer_mnnvl, buffer_mnnvl, buffer_flags_mnnvl, max_num_elements_mnnvl = get_allreduce_mnnvl_workspace(
+                self.mapping, self.dtype)
+            self.workspace = buffer_mnnvl.view(3, 2, -1, input.shape[-1])
+            self.buffer_flags_mnnvl = buffer_flags_mnnvl
+        else:
+            if strategy == AllReduceStrategy.LOWPRECISION:
+                allocate_low_presicion_allreduce_workspace(self.mapping)
```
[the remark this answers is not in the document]
HuiGao-NV: it will be replace back by self.workspace = get_allreduce_workspace(self.mapping) at line 555 which is stored in thread_local.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -473,13 +592,11 @@ def forward(
...
 
-        # Try MNNVL AllReduce first if available
-        if self.mnnvl_allreduce:
-            mnnvl_output = self.mnnvl_allreduce(
-                input, all_reduce_params=all_reduce_params)
-            if mnnvl_output is not None:
-                return mnnvl_output
+        strategy = self.get_runtime_strategy(input, all_reduce_params)
+
+        self._prepare_for_strategy(strategy, input)
```
hyukn, line 597: This should not occur during runtime. If we have to select from some strategies only when the size of the input is known, then we should preallocate workspace for them. Notice that workspace are shared for different allreduce ops.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -473,13 +592,11 @@ def forward(
         if all_reduce_params is None:
             all_reduce_params = AllReduceParams()
 
-        # Try MNNVL AllReduce first if available
-        if self.mnnvl_allreduce:
-            mnnvl_output = self.mnnvl_allreduce(
-                input, all_reduce_params=all_reduce_params)
-            if mnnvl_output is not None:
-                return mnnvl_output
+        strategy = self.get_runtime_strategy(input, all_reduce_params)
```
hyukn, line 595: If we put the strategy selection logic out of the custom op, then a better warmup is required for torch.compile.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -416,25 +352,208 @@ def __init__(self,
...
+        if world_size <= 2:
+            return 16 * 1000 * 1000
+        return 8 * 1000 * 1000
+
+    def _match_min_latency(self, world_size, message_size_bytes):
+        if world_size <= 2 or (message_size_bytes < 500 * 1000) or (
+                world_size <= 4 & message_size_bytes < 1 * 1000 * 1000):
+            return True
+
+    def get_dtype_size(self, dtype):
```
hyukn, line 485: Why we do this? You can directly use `element_size()`

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -416,25 +352,208 @@ def __init__(self,
...
+    def _strategy_supports(self, strategy, input, fusion_op):
+        # Determine if this is supported in runtime.
+        if strategy == AllReduceStrategy.UB and input.dtype(
+        ) in AllReduce.get_UB_supported_dtype():
+            return True
+
+        if strategy == AllReduceStrategy.NCCL:
+            return True
+
+        override_strategy = os.getenv("OVERRIDE_HEURISTIC_ALLREDUCE_STRATEGY",
```
hyukn, line 401: We can determine all the env vars just once during init instead of forward.

tensorrt_llm/_torch/distributed/ops.py
hyukn, line 435: If strategy support returns false here, why don't we just set it as NCCL, because we have no idea whether AUTO works or not?

cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -622,33 +647,6 @@ class AllreduceOp
...
-            }
-            else
-            {
-                runtime_strategy = selectImplementation(seq_len, size, mGroup.size(), mType);
-            }
-        }
-        return runtime_strategy;
-    }
-
     void logRunTimeStrategy(AllReduceStrategyType strategy, int rank)
```
hyukn, line 650: nit: This one becomes longer and longer. Can we move it away into `customAllReduceKernels.h`?

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -416,25 +352,208 @@ def __init__(self,
...
+            raise NotImplementedError(f"Wrong datatype is set {dtype}")
+
+    def _infer_strategy(self, input, fusion_op: AllReduceFusionOp,
+                        stratey: AllReduceStrategy) -> AllReduceStrategy:
+        is_auto = stratey == AllReduceStrategy.AUTO
+        message_size_bytes = input.numel() * self.get_dtype_size(input.dtype)
+        max_workspace_size = AllReduce._get_max_required_workspace_size(
+            len(self.mapping.tp_group))
+
+        if self._fallback_nccl(message_size_bytes, max_workspace_size, is_auto):
```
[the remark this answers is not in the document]
liji-nv: Yes, the cause is the strategy is decided by size. Which is shape dependent condition.

HuiGao-NV:
Add license to new added files.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -284,79 +291,232 @@ def convert_input(x, x_info):
...
+    elif fusion_op == AllReduceFusionOp.RESIDUAL_RMS_NORM_QUANT_FP8:
+        return [torch.empty_like(input), torch.empty_like(residual)]
+    elif fusion_op == AllReduceFusionOp.RESIDUAL_RMS_NORM_OUT_QUANT_FP8:
+        return [
+            torch.empty_like(input),
+            torch.empty_like(input),
+            torch.empty_like(residual)
+        ]
+    elif fusion_op == AllReduceFusionOp.RESIDUAL_RMS_NORM_QUANT_NVFP4:
+        return [norm_quant_fp4, scale_factor, torch.empty_like(residual)]
```
liji-nv, line 516: Definition is missing.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -284,79 +291,232 @@ def convert_input(x, x_info):
...
+      norm_weight: Optional[torch.Tensor], scale: Optional[torch.Tensor],
+      bias: Optional[torch.Tensor], tp_group: list[int], strategy: int,
+      fusion_op: int, eps: float, trigger_completion_at_end: bool,
+      buffer_mnnvl_shape: Optional[list[int]],
+      workspace: Optional[torch.Tensor], mnnvl_buffer: Optional[torch.Tensor],
+      mnnvl_buff_flags: Optional[torch.Tensor]) -> List[torch.Tensor]:
+    if fusion_op == AllReduceFusionOp.RESIDUAL_RMS_NORM:
+        return [torch.empty_like(input), torch.empty_like(residual)]
+    elif fusion_op == AllReduceFusionOp.RESIDUAL_RMS_NORM_QUANT_FP8:
+        return [torch.empty_like(input), torch.empty_like(residual)]
```
liji-nv, line 508: Fake impl should correct set dtype. out is not same as input in this case.

tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -284,79 +291,232 @@ def convert_input(x, x_info):
...
+        return [
+            torch.empty_like(input),
+            torch.empty_like(input),
+            torch.empty_like(residual)
+        ]
+    elif fusion_op == AllReduceFusionOp.RESIDUAL_RMS_NORM_QUANT_NVFP4:
+        return [norm_quant_fp4, scale_factor, torch.empty_like(residual)]
+    elif fusion_op == AllReduceFusionOp.RESIDUAL_RMS_NORM_OUT_QUANT_NVFP4:
+        return [norm, norm_quant_fp4, scale_factor, torch.empty_like(residual)]
+    return torch.empty_like(input)
```
liji-nv, line 519: If the expected return type is List[torch.Tensor], it must be wrapped with a `[]`

PR #5284, opened 2025-06-17
feat: Support stream_interval

kaiyux:
You can use a `extra-llm-api-config.yml` to enable the feature:
```
stream_interval: 4
```


tensorrt_llm/_torch/pyexecutor/config.py
```diff
@@ -87,6 +87,10 @@ class PyTorchConfig:
     enable_min_latency: bool = False
     allreduce_strategy: str = "AUTO"
 
+    # The iteration interval to create responses under the streaming mode.
+    # TODO: make this a per-request parameter
+    stream_interval: int = 1
```
hypdeb, line 92: Is that what we discussed in our last call? I.e. returning only every N tokens?

jdebache (commented):
Seems like a very small number of changes to enable that feature, which is great! Do you have some numbers on how this affects performance?

pcastonguay (changes requested):
Can we add tests to verify this is working as expected?

kaiyux:
> Can we add tests to verify this is working as expected?

@pcastonguay I added tests to `tests/integration/defs/accuracy/test_llm_api_pytorch.py` (https://github.com/NVIDIA/TensorRT-LLM/pull/5284/commits/90c12f47ddabda80abc93211198ffd296028e976) to make sure that there are no tokens missed. However if you meant to verify that it indeed return responses every `N` iterations, I'm not what might be the best way to verify that for now and can take a closer look tomorrow.

tensorrt_llm/evaluate/interface.py
```diff
@@ -78,7 +78,8 @@ def _get_sampline_params(self, sampling_params: Optional[SamplingParams],
 
     def evaluate(self,
```
syuoni, line 79: Hi @kaiyux ,

In addition to this file, you also need to handle the `evaluate` method in `lm_eval.py`. Please pass the new `streaming` argument through the below places:
* https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/evaluate/lm_eval.py#L185
* https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/evaluate/lm_eval.py#L40
* https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/evaluate/lm_eval.py#L105

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1723,6 +1723,13 @@ class TorchLlmArgs(BaseLlmArgs):
         "If true, enable min-latency mode. Currently only used for Llama4.",
     )
 
+    # TODO: make this a per-request parameter
+    stream_interval: int = Field(
+        default=1,
+        description=
+        "The iteration interval to create responses under the streaming mode.",
+    )
```
Superjomn, line 1731: It is better to add a `@model_validator` to this field

```python
@model_validator(mode="after")
def validate_stream_interval(self):
    if self.stream_interval < 0:
        raise ValueError(...)
    if self.stream_interval > self.max_num_tokens: # Better to be SamplingParams.num_tokens?
        raise ValueError(...)
    return self
```

kaiyux: Added. We don't need to handle the case when `self.stream_interval > self.max_num_tokens` because that case it will just fallback as non-streaming.

syuoni (approved):

pcastonguay (approved):

QiJune (approved):
LGTM

PR #5296, opened 2025-06-17
[refactor] Remove SpecConfig boilerplate

mikeiovine:
## Description

Adding new speculation algorithms is a big pain right now. One of the reasons is that developers are forced to duplicate the config objects: one object is user-facing, and another is internal/inherits from `SpecConfig`. The internal one only has a few extra pieces of functionality. This PR deduplicates all the speculation configs - the amount of boilderplate required for adding a new algorithm is reduced. 

* Removes `SpecConfig` and its derived classes. The `DecodingConfig` that the user passes to the `LLM` can now be used directly.
* Removes extra functionality in `SpecConfig` that isn't strictly necessary. This is to avoid confusing users. This includes `update_from_model_config` (we can just pass the `ModelConfig` to `get_spec_metadata` instead) and `get_draft_prompt_tokens` (moved to `speculative/utils.py`).
* Also removed `num_extra_kv_tokens`. There's no need to put it in the config object; this can be derived from the `spec_dec_mode`. We can move the logic to `utils.py` as well. 
* `spec_dec_mode` is now derived as a `cached_property` of `DecodingConfig`, which avoids us having to remember to put it in `__post_init__` whenever we add a new spec algorithm. 

## Test Coverage

Manually tested:

- [x] EAGLE3
- [x] EAGLE3 one-model
- [x] MTP (tested model loading, crashes on main)
- [x] EAGLE-style MTP (tested model loading, crashes on main)
- [x] Draft/target

NGram is currently broken on main, will test when the fix lands. I am additionally relying on existing unit tests in CI.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -23,6 +23,7 @@ import tensorrt_llm.bindings.internal.userbuffers as ub
 from tensorrt_llm._torch.pyexecutor.llm_request import LlmRequest
 from tensorrt_llm._torch.pyexecutor.sampler import SampleStateTensors
 from tensorrt_llm._torch.speculative.mtp import SampleStateTensorsMTP
+from tensorrt_llm._torch.speculative.utils import get_num_extra_kv_tokens
 from tensorrt_llm._utils import (is_trace_enabled, local_mpi_rank,
                                  local_mpi_size, nvtx_range, release_gc,
                                  torch_dtype_to_str, trace_func)
@@ -51,7 +52,7 @@ from ..models.modeling_utils import (DecoderModelForCausalLM, MetaInitMode,
                                      timing)
 from ..modules.fused_moe.moe_load_balancer import (
     MoeLoadBalancer, MoeLoadBalancerIterContext, maybe_create_moe_load_balancer)
-from ..speculative import SpecConfig, SpecMetadata, get_spec_metadata
+from ..speculative import SpecMetadata, get_spec_metadata
 from ..utils import (get_model_extra_attrs, set_torch_compiling,
                      with_model_extra_attrs)
 from .config import LoadFormat, PyTorchConfig
@@ -331,7 +332,7 @@ class PyTorchModelEngine(ModelEngine):
         mapping: Optional[Mapping] = None,
         attn_runtime_features: Optional[AttentionRuntimeFeatures] = None,
         dist: Optional[MPIDist] = None,
-        spec_config: Optional[SpecConfig] = None,
+        spec_config: Optional["DecodingBaseConfig"] = None,
         guided_decoding_config: Optional[GuidedDecodingConfig] = None,
         lora_config: Optional[LoraConfig] = None,
         is_draft_model: bool = False,
@@ -432,8 +433,7 @@ class PyTorchModelEngine(ModelEngine):
 
         if self.is_spec_decode:
             self.spec_metadata = None
-            self.spec_config.update_from_model_config(self.model.config)
-            max_num_draft_tokens = self.spec_config.max_draft_tokens * batch_size
+            max_num_draft_tokens = self.spec_config.max_draft_len * batch_size
             self.draft_tokens_cuda = torch.empty((max_num_draft_tokens, ),
                                                  dtype=torch.int,
                                                  device='cuda')
@@ -449,7 +449,7 @@ class PyTorchModelEngine(ModelEngine):
                                                              device='cuda')
             self.without_logits = self.spec_config.spec_dec_mode.without_logits(
             )
-            self.max_draft_len = spec_config.max_draft_tokens
+            self.max_draft_len = spec_config.max_draft_len
         else:
             self.without_logits = False
             self.max_draft_len = 0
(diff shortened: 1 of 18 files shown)


tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -23,6 +23,7 @@
 from tensorrt_llm._torch.pyexecutor.llm_request import LlmRequest
 from tensorrt_llm._torch.pyexecutor.sampler import SampleStateTensors
 from tensorrt_llm._torch.speculative.mtp import SampleStateTensorsMTP
+from tensorrt_llm._torch.speculative.utils import get_num_extra_kv_tokens
```
lfr-0531, line 26: Can we add `get_num_extra_kv_tokens` to `speculative/__init__.py`, so we can import it directly by using
```
from ..speculative import get_num_extra_kv_tokens
```

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -17,6 +17,7 @@
 
 from tensorrt_llm._torch.pyexecutor.resource_manager import ResourceManagerType
 from tensorrt_llm._torch.pyexecutor.seq_slot_manager import SeqSlotManager
+from tensorrt_llm._torch.speculative.utils import get_draft_model_prompt
```
lfr-0531, line 20: The same.

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -19,7 +19,8 @@
 
 from ..attention_backend.interface import AttentionRuntimeFeatures
 from ..distributed import MPIDist
-from ..speculative import NGramConfig, get_spec_resource_manager
+from ..speculative import get_spec_resource_manager
+from ..speculative.utils import get_num_extra_kv_tokens
```
lfr-0531, line 23: The same.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -282,30 +319,52 @@ def from_dict(cls, data: dict):
...
     decoding_type: ClassVar[str] = "MTP"
 
+    def supports_backend(self, backend: str) -> bool:
+        return backend == "pytorch"
+
+    @functools.cached_property
+    def spec_dec_mode(self):
+        from tensorrt_llm._torch.speculative.interface import \
+            SpeculativeDecodingMode as TorchSpeculativeDecodingMode
+        if self.max_draft_len == 1 and not self.use_mtp_vanilla:
```
lfr-0531, line 364: For MTP spec decoding mode, we use `num_nextn_predict_layers` in `model_config` to decide between MTP Vanilla and MTP Eagle, rather than the `max_draft_len`. The `num_nextn_predict_layers` means how many MTP layers in the checkpoint, while the `max_draft_len` is the user-specified max draft length. They are different parameters. 

Seems like `model_config` isn’t accessible in `llm_args`, so maybe we can add a `num_nextn_predict_layers: Optional[int] = 0` in `MTPDecodingConfig`, and update its value in `speculative/utils.py`. Then we can change this line to
```
        if self.num_nextn_predict_layers == 1 and not self.use_mtp_vanilla:
```

tensorrt_llm/_torch/speculative/utils.py
```diff
@@ -111,3 +113,26 @@ def get_spec_worker(spec_config, mapping):
...
+    return input_tokens
+
+
+def get_num_extra_kv_tokens(spec_config):
+    """
+    Implementation detail for one model implementations of speculative decoding. Extra
+    KV cache tokens are required.
+    """
+    if spec_config.spec_dec_mode.is_eagle3_one_model(
+    ) or spec_config.spec_dec_mode.is_mtp():
```
lfr-0531, line 136: For MPT, only `MTP_Eagle` has `num_extra_kv_tokens > 0`.
```suggestion
    ) or spec_config.spec_dec_mode.is_mtp_eagle():
```

lfr-0531 (commented):
The rest LGTM!

tensorrt_llm/llmapi/llm_args.py
wili-65535: I'm a little bit confused about this.
Does it mean for MTP, `num_nextn_predict_layers` (decided by model) could be not equal to `max_draft_len` (decided by users)?

lfr-0531: > Does it mean for MTP, num_nextn_predict_layers (decided by model) could be not equal to max_draft_len (decided by users)? 
So in the HLAPI, users can specify any max_draft_len they like (for example: spec_config = MTPDecodingConfig(max_draft_len=args.spec_decode_nextn, ...)), then TRTLLM loads num_nextn_predict_layers from model checkpoint?

Yes, they are different. For example, DeepSeek-R1 only has one MTP layer the ckpt, i.e. `num_nextn_predict_layers = 1`, and we can set `max_draft_len = 3`.

> Does it mean that when num_nextn_predict_layers==1, MTP-EAGLE could be enabled and use max_draft_len as number of draft tokens; when num_nextn_predict_layers > 1, max_draft_len should be equal to num_nextn_predict_layers as number of draft tokens?

When `num_nextn_predict_layers==1`, MTP-EAGLE is the default. But if users set `use_mtp_vanilla = True`, it switches to MTP-Vanilla. When `num_nextn_predict_layers > 1`, MTP-Vanilla becomes the default, and it also supports any `max_draft_len` set by users.

mikeiovine:
This was moved to https://github.com/NVIDIA/TensorRT-LLM/pull/5639 since I was out of office last week. Closing.

PR #5334, opened 2025-06-18
doc: Include NGC release containers in quick-start-guide.md

MartinMarciniszyn:
# Update Quick Start Guide

Add explanations to the quick start guide for running the examples within the NGC release container.

### DIFF
--- a/docs/source/quick-start-guide.md
+++ b/docs/source/quick-start-guide.md
@@ -4,6 +4,9 @@
 
 This is the starting point to try out TensorRT-LLM. Specifically, this Quick Start Guide enables you to quickly get setup and send HTTP requests using TensorRT-LLM.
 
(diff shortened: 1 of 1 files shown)


laikhtewari:
The easiest (ie lightest weight) way would be with pip

We consistently hear from devs (both internal SAs and external users) that they prefer pip when quickly testing new libs

Pip should be highlighted here first in addition to ngc container I think

MartinMarciniszyn:
> The easiest (ie lightest weight) way would be with pip
> 
> We consistently hear from devs (both internal SAs and external users) that they prefer pip when quickly testing new libs
> 
> Pip should be highlighted here first in addition to ngc container I think

I agree that pip would be the easiest option in theory. However, our binary dependencies are so involved that pip install rarely works on a bare metal system. We would need to describe how to start a docker container in which one can run pip install. And that seems more complicated than starting the container with TRT-LLM pre-installed in the first place.

juney-nvidia (approved):

PR #5335, opened 2025-06-18
[fix][test] remove some cpp test cases from h100

omera-nv:
# Remove some CPP test cases from the H100 pipeline

Some CPP tests take a long time and are not required to run on an H100 anymore. Remove these.


DomBrown (approved):
LGTM based on our thread earlier.

PR #5350, opened 2025-06-19
feat(model):support qwen3 dense in trt flow

xq25478:
# feat(model) support qwen3 dense in trt flow


xq25478:
> Hello，When I use above workflow, I encountered the following problem：
> 
> v0.15.0/tensorrt_llm/models/qwen/convert.py", line 535, in convert_hf_qwen assert (k_bias.shape[0] % (mapping.tp_size * head_size)) == 0 AttributeError: 'NoneType' object has no attribute 'shape'
> 
> My model is Qwen3-0.6B，and trtllm version is v0.15.0. Have you encountered the above error? I tried migrating the conversion code for Qwen3 TRT from the main branch to TRT-LLM v0.15.0, but the aforementioned error still occurs.

please use latest trtllm version，thank you。

PR #5365, opened 2025-06-19
tests: update benchmark test lists

xinhe-nv:
update benchmark test lists for NIM


tests/integration/defs/accuracy/test_cli_flow.py
```diff
@@ -245,6 +245,7 @@ def test_fp8_prequantized(self, mocker):
                  kv_cache_quant_algo=QuantAlgo.FP8)
 
 
+@pytest.mark.timeout(7200)
```
crazydemo, line 248: this case needs at least 3 hours

LarryXFly (approved):

PR #5371, opened 2025-06-19
[TRTLLM-5838][fix] fix max batch size and max tokens in kv cache estimations for Nemotron-H

tomeras91:
Throughout the code there are a few places where the number of available kv cache tokens and/or maximum batch size are estimated. These estimations are based on the available free GPU memory and on the memory size of a single kv cache entry. These estimations didn't take into account hybrid models like Nemotron-H, in which not all layers are attention layers and require a kv cache.

Changes in this PR:
* consider mamba cache for max batch size estimation in `trtllm-bench throughput` command
* take only attention layers into account when estimating maximum number of tokens in kv cache, both in `trtllm-bench throughput` and in `KvCacheCreator`
* propagate `kv_cache_gpu_mem_fraction` from `trtllm-bench throughput` CLI arg to the function that estimates the maximum batch size
* release mamba cache memory when `MambaHybridCacheManager` is shut down (+ a small refactor to increase readability of `MambaHybridCacheManager`).


tensorrt_llm/bench/build/tuning.py
```diff
@@ -70,11 +79,39 @@ def calc_engine_setting(
...
+        num_mamba_layers = model_config.hybrid_override_pattern.count("M")
+        conv_dim = model_config.mamba_config.d_inner + 2 * model_config.mamba_config.n_groups * model_config.mamba_config.d_state
+        num_conv_state_elements = num_mamba_layers * conv_dim * (
+            model_config.mamba_config.d_conv - 1)
+        num_ssm_state_elements = num_mamba_layers * model_config.mamba_config.n_heads * model_config.mamba_config.head_dim * model_config.mamba_config.d_state
+        byte_per_state_elem = BYTES_PER_ELEM.get(quant_config.quant_algo, 2)
+        byte_per_mamba_cache = byte_per_state_elem * (
+            num_conv_state_elements + num_ssm_state_elements) / (1024**3)
+
+        print(f"byte_per_mamba_cache: {byte_per_mamba_cache}")
```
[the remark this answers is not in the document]
tomeras91: removed debug print

tensorrt_llm/bench/build/dataclasses.py
```diff
@@ -111,6 +113,29 @@ def _parse(filename: str) -> None:
...
+    head_dim: int = Field(
+        validation_alias=AliasChoices("head_dim", "mamba_head_dim"))
+    d_inner: int = Field(default=None)
+    n_heads: int = Field(default=None)
+
+    @model_validator(mode="after")
+    def set_values_if_none(self):
+        """ Set the values if cannot get values from HF config.json. """
+        if not self.d_inner:
+            self.d_inner = self.d_model * self.expand
```
tomeras91, line 133: TODO: fix this for Nemotron-H-4B

tensorrt_llm/bench/build/dataclasses.py
```diff
@@ -111,6 +113,29 @@ def _parse(filename: str) -> None:
         return huggingface_hub.get_safetensors_metadata(model_name_or_path)
 
 
+class MambaConfig(BaseModel):
```
FrankD412, line 116: Since this is being initialized by the HF config, it might be worth considering making this a derived class from `ModelConfig`. @Jie-Fang and I were syncing on this PR and we think that derivation would be a way to make it more general. This opens us to the ability to create an abstraction that let's us:

1. Isolate the calculation of additional KV cache elements required -- that would mean that we no longer need a [Mamba specific KV cache](https://github.com/NVIDIA/TensorRT-LLM/pull/5371/files#diff-6c5c812fd05818e2f21f8d111a26f34987e6c470268c5f9fff5e8e0350fa80e3R97) variable. We could generalize the calculation to be something like `(byte_per_token * target_seq_len) + extra_model_kv_cache` and generally get it from the derived configuration which contains the needed information to calculate it. The class can expose a method that can be used to calculate it. That also removes the need for the [conditional logic here](https://github.com/NVIDIA/TensorRT-LLM/pull/5371/files#diff-6c5c812fd05818e2f21f8d111a26f34987e6c470268c5f9fff5e8e0350fa80e3R84). You can also log the additional Mamba memory in the class.
2. We can return the number of heads based on the calculation in the derived class so we don't need the [conditional logic](https://github.com/NVIDIA/TensorRT-LLM/pull/5371/files#diff-6c5c812fd05818e2f21f8d111a26f34987e6c470268c5f9fff5e8e0350fa80e3R61)
3. The instantiation can be a simple `if/else` for now, but by making a derived class we can start plumbing the path to a factory or other pattern that will let us return the right class [here](https://github.com/NVIDIA/TensorRT-LLM/pull/5371/files#diff-bdac6891a96eb8ac0a6b99841b80f2f6da0c6317ee01a53278c165ae46533569R223). We now no longer need to pass in the config, just decide which `ModelConfig` class to use.

tomeras91: Hey @FrankD412, thanks for the feedback!
I do like the general idea, but with a few changes.
As far as I understand, you're suggesting to add a function to the `ModelConfig` class that calculates the kv/generalized cache size per sequence. I had a look at the code, and `ModelConfig` isn't really used anywhere except in `calc_engine_setting` to compute cache size. Therefore, assuming we add the functionality you suggested, I suggest we rename this class. This is no longer a config class (and not a datacalss / pydantic model) - it's a functional class used to calculate cache size, so let's call it accordingly. Something like `ModelCacheCalculator` or `CacheSizeEstimator` is appropriate. 

For now we'll have a base class and 2 derived classes - `KVCacheSizeEstimator` and `HybridMambaKVCacheSizeEstimator`, and instantiate using a simple `if/else`.

WDYT?

tensorrt_llm/bench/build/tuning.py
```diff
@@ -70,19 +79,48 @@ def calc_engine_setting(
...
+
+    if is_mamba_attn_hybrid:
+        num_mamba_layers = model_config.hybrid_override_pattern.count("M")
+        conv_dim = model_config.mamba_config.d_inner + 2 * model_config.mamba_config.n_groups * model_config.mamba_config.d_state
+        num_conv_state_elements = num_mamba_layers * conv_dim * (
+            model_config.mamba_config.d_conv - 1)
+        num_ssm_state_elements = num_mamba_layers * model_config.mamba_config.n_heads * model_config.mamba_config.head_dim * model_config.mamba_config.d_state
+        byte_per_state_elem = BYTES_PER_ELEM.get(QuantAlgo.NO_QUANT)
+        byte_per_mamba_cache = byte_per_state_elem * (
+            num_conv_state_elements + num_ssm_state_elements) / (1024**3)
```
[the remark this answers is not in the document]
tomeras91: you're right. Will fix

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -82,10 +82,14 @@ def _get_cache_size_per_token(model_config: ModelConfig,
                 config.hidden_size // config.num_attention_heads,
             ) * num_key_value_heads // tp_size
 
+        if is_nemotron_hybrid(config):
+            num_attention_layers = config.hybrid_override_pattern.count("*")
+        else:
+            num_attention_layers = config.num_hidden_layers
```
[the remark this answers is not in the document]
FrankD412: I think a good compromise is that if we can agree on the derived/generalized methodology that we can simply make a check here to choose which one should be used. @Jie-Fang and I discussed on Friday, and we're thinking that this probably does need to start moving towards a factory pattern design.

While the full factory/generalized implementation is outside the scope of this PR, I do think that if we start by implementing something in the right direction here means that we can roll that in to a larger generalization later.

tensorrt_llm/bench/build/tuning.py
```diff
@@ -50,12 +50,21 @@ def calc_engine_setting(
         Tuple[int, int]: Tuple containing engine configuration information for
         engine build (max_num_tokens, max_batch_size).
     """
+    is_mamba_attn_hybrid = isinstance(model_config, NemotronHybridConfig)
+    if is_mamba_attn_hybrid:
+        # Each mamba cache entry is pretty large (~50MB for 8B model), so we are more conservative when estimating the max batch size
+        kv_cache_gpu_mem_fraction *= kv_cache_gpu_mem_fraction
```
FrankD412, line 56: Might it be worth moving this into the model configuration also? It could have a generalized method for computing the KV cache scaling based on the model?

FrankD412: And by into I mean maybe a method -- not an actual assignment.

tensorrt_llm/bench/build/tuning.py
```diff
@@ -148,13 +170,16 @@ def finetune_setting(
...
+        # Set min BS to be 64.
+        if raw_bs < 256:
+            max_bs = max(64, 32 * math.ceil(raw_bs / 32))
+        elif raw_bs < 1024:
+            max_bs = 128 * math.ceil(raw_bs / 128)
+        else:
+            max_bs = 256 * math.ceil(raw_bs / 256)
     else:
-        max_bs = 256 * math.ceil(raw_bs / 256)
+        max_bs = 2 * math.floor(raw_bs / 2)
```
FrankD412, line 182: How was this math determined? Are we saying that this approximation holds in general if we are trying to be less optimistic across all models?

tomeras91: Yeah.. this is a just a floor round to the nearest even number. My thought process was that if we don't want to be optimistic (i.e. increase the max batch size), we can just keep it as is. I rounded it down to an even number just because even numbers are nicer.

tensorrt_llm/bench/build/tuning.py
tomeras91: Good idea. Fixed

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -640,6 +638,13 @@ def get_ssm_states(self, layer_idx: int) -> torch.Tensor:
         layer_offset = self.mamba_layer_offsets[layer_idx]
         return self.ssm_states[layer_offset]
 
+    def shutdown(self):
+        # release tensor memory, keeping python references as tensors
+        self.conv_states = torch.tensor([])
+        self.ssm_states = torch.tensor([])
+        self.state_indices = torch.tensor([])
+        torch.cuda.empty_cache()
```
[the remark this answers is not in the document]
tomeras91: On second thought, I prefer to leave this as is. The idea here is to clear torch memory, not to reset the entire state of the `MambaCacheManager`. Similar to how the `KVCacheManager` just calls `self.impl.release_pools()` in its shutdown method

tensorrt_llm/_torch/pyexecutor/_util.py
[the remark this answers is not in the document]
tomeras91: Added a `get_num_attention_layers()` function in `_torch/model_config.py::ModelConfig`

tensorrt_llm/bench/build/tuning.py
tomeras91: It works for Nemotron-H :) (the only relevant hybrid model)

FrankD412 (approved):
Approving since these changes from the `trtllm-bench` side.

Naveassaf (approved):
LGTM. 

Great stuff! :muscle:

PR #5382, opened 2025-06-20
[feat]: Detokenize option in /v1/completions request

Wokzy:
# Detokenize option in sampling params for trtllm-serve

## Description

Now, when making request, users can specify boolean option "detokenize", for instance:
`curl http://localhost:54321/v1/completions -H "Content-Type: application/json" -d '{"model":"DeepSeek-V3-0324", "prompt":"hello", "max_tokens":64, "detokenize":false}'`

and output will be smth like:
```
{"id":"cmpl-2027c15c7d544095908a27be101c61a2","object":"text_completion","created":1750405539,"model":"DeepSeek-V3-0324","choices":[{"index":0,"text":"","token_ids":[14, 342, 4571, 5958, 304, 5085, 270, 2502, 78157, 295, 65980, 14, 342, 1153, 396, 65980, 769, 270, 5304, 304, 5085, 295, 223, 20, 70, 14, 342, 2090, 1664, 1153, 1192, 304, 2859, 436, 377, 14, 305, 342, 4571, 3638, 260, 3375, 1014, 4985, 436, 16, 588, 5524, 1694, 678, 14, 8033, 1240, 23166, 2893, 43, 611, 260, 3295, 418, 270, 18395],"logprobs":null,"context_logits":null,"finish_reason":"length","stop_reason":null,"disaggregated_params":null}],"usage":{"prompt_tokens":2,"total_tokens":64,"completion_tokens":62},"prompt_token_ids":[[0,33310]]}
```

This option works fine with `stream: true`


Wokzy:
I could make additional field in `CompletionResponseChoice`, like `output_ids: list[int]` to put tokens there, without casting beam_output.token_ids into string

Wokzy:
@LinPoly @syuoni

syuoni (commented):
Thanks for updating this PR.

@LinPoly  Could you please take a look again? I think the current PR is changing `trtlllm-serve`'s API.

LinPoly (commented):
I think we'd better use another field to hold token_ids, leave a few suggestions for the possible change, @Wokzy feel free to comment or correct me if you have different ideas. We also need to add a test in [this file](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/apps/_test_openai_completions.py), with `detokenize=False`, and check if text is empty and token_ids is expected. You can just ping me if you are confused about the test.

Wokzy:
> I think we'd better use another field to hold token_ids, leave a few suggestions for the possible change, @Wokzy feel free to comment or correct me if you have different ideas. We also need to add a test in [this file](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/llmapi/apps/_test_openai_completions.py), with `detokenize=False`, and check if text is empty and token_ids is expected. You can just ping me if you are confused about the test.

Thanks for your feedback @LinPoly! I also thought about additional field, but in case people using completions API in some languages like GO, where they need to specify certain json pattern for response, it would require them to adapt their software after these changes.

I will add a couple of tests for non-streaming and streaming with detonenize=False

LinPoly:
>  in case people using completions API in some languages like GO, where they need to specify certain json pattern for response, it would require them to adapt their software after these changes.

If we change the type of text, from string to union of string and int list, I think they also need to change the client code if they want to use detokenize feature?

Wokzy:
> If you are blocked when adding tests, don't hesitate to let us know, we can help with that.

I'm familiar with pytest, however how am I supposed to run these tests? By simply going to `tests/unittest/llmapi/apps/_test_openai_completions.py` and run `pytest .`?

Wokzy:
> If client code has to change anyway, I would prefer the additional field way.

I think you're right, I'll make a special field: `token_ids`, as you suggested

LinPoly:
> however how am I supposed to run these tests? By simply going to tests/unittest/llmapi/apps/ and running pytest _test_openai_completions.py?

Yes you can run the test in this way, to locally verify your code, we can launch CI for you in order to merge code.

LinPoly (changes requested):
Can pass the test locally with the following modifications, can you pls take a look?

Wokzy:
> Can pass the test locally with the following modifications, can you pls take a look?

Done.

LinPoly (approved):
LGTM.

Wokzy:
> @Wokzy Thanks a lot for your support! Can you help fix the code style following the guidance here? https://github.com/NVIDIA/TensorRT-LLM/blob/main/CONTRIBUTING.md#coding-style

@kaiyux Done.

PR #5394, opened 2025-06-21
[TRTLLM-6019] feat: Remove cutlass min latency code from AutoTuner.

hyukn:
Remove the tuning process for the cutlass fused moe path under min_latency_mode.


tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -44,6 +41,7 @@ def __init__(
         enable_alltoall: bool,
         use_deepseek_fp8_block_scale: bool,
         use_w4a8_group_scaling: bool,
+        min_latency_mode: bool,
```
hlu1, line 44: Why do you need to pass the argument here? Should `self.min_latency_mode` just be false?

hlu1 (approved):

PR #4973, opened 2025-06-06
fix:https://nvbugs/5324248

nv-guomingz:
main branch has ut regression with this cmd

```python
pytest tests/unittest/_torch/test_pytorch_model_engine.py::PyTorchModelEngineTestCase
```

this pr fixed cases 

FAILED tests/unittest/_torch/test_pytorch_model_engine.py::PyTorchModelEngineTestCase::**test_layerwise_nvtx_marker** - TypeError: argument of type 'NoneType' is not iterable
FAILED tests/unittest/_torch/test_pytorch_model_engine.py::PyTorchModelEngineTestCase::**test_pad_generation_requests** - TypeError: argument of type 'NoneType' is not iterable
FAILED tests/unittest/_torch/test_pytorch_model_engine.py::PyTorchModelEngineTestCase::**test_position_id_preparation** - TypeError: argument of type 'NoneType' is not iterable
FAILED tests/unittest/_torch/test_pytorch_model_engine.py::PyTorchModelEngineTestCase::**test_warmup** - TypeError: argument of type 'NoneType' is not iterable

PR #5416, opened 2025-06-24
Fix test Pytorch model engine

Tabrizian:
Looks like the test skip was not removed when this PR was merged: https://github.com/NVIDIA/TensorRT-LLM/pull/4973

### DIFF
--- a/tests/unittest/_torch/test_pytorch_model_engine.py
+++ b/tests/unittest/_torch/test_pytorch_model_engine.py
@@ -1,7 +1,6 @@
 import unittest
 from dataclasses import dataclass
 
-import pytest
 import torch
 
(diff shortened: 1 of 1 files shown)


tests/unittest/_torch/test_pytorch_model_engine.py
```diff
@@ -43,6 +42,9 @@ def __init__(self, dtype: torch.dtype):
             torch_dtype=dtype))
         self.recorded_position_ids = None
 
+    def infer_max_seq_len(self):
```
mikeiovine, line 45: Why are we making this change?

mikeiovine (approved):

PR #5423, opened 2025-06-24
[fix][ci] trigger multigpu tests for deepseek changes

omera-nv:
# Trigger multigpu CI tests for deepseek-related changes

This PR adds moe and deepseek related files to the list of files that will trigger the multigpu tests in CI when modified.


jenkins/L0_MergeRequest.groovy
```diff
@@ -602,6 +603,8 @@ def getMultiGpuFileChanged(pipeline, testFilter, globalVars)
         "tensorrt_llm/_torch/pyexecutor/py_executor.py",
         "tensorrt_llm/_torch/pyexecutor/_util.py",
         "tensorrt_llm/_torch/models/modeling_llama.py",
+        "tensorrt_llm/_torch/models/modeling_deepseekv3.py",
```
chzblych, line 606: This is covered by this rule - https://github.com/NVIDIA/TensorRT-LLM/blob/2b34d5a2cc5421bf40c88c29be9e3ee2774b1602/jenkins/L0_MergeRequest.groovy#L528.

We only run DeepSeek related multi-GPU tests if only this file is touched.

omera-nv: Removed

omera-nv:
/bot skip --comment "Removed one file from multigpu test filter"

chzblych (approved):

PR #5431, opened 2025-06-24
[TRTLLM-5277] chore: refine llmapi examples for 1.0 (part1)

Superjomn:
## PR Description
Clean up the existing LLM-API Examples:
1. Hide all the TRT-specific examples in the `_tensorrt_engine` directory
2. Introducing subsections to the example list

Minor enhancement on docgen:
1. Introduce metadata to each example to help control the order
2. Exclude the title and other metadata from the example

image


examples/llm-api/quickstart_example.py
```diff
@@ -1,4 +1,5 @@
-from tensorrt_llm import LLM, SamplingParams
+from tensorrt_llm import SamplingParams
+from tensorrt_llm._tensorrt_engine import LLM
```
Superjomn, line 2: The test_pip_install will fail due to the CI docker env if swith to PyT, thus we pause the change.

nv-guomingz (approved):

PR #5432, opened 2025-06-24
perf: better heuristic for allreduce

yilin-void:
# better heuristic for allreduce

This PR will not significantly affect kernel performance when the token number is less than or equal to 32, and for token numbers greater than 32, it will achieve significant acceleration. Additionally, the AUTO heuristic selection logic in the allreduce torch op will use 128 tokens as the NCCL/MIN_LATENCY threshold and add an environment variable to change this threshold.


hyukn:
I think we should add multi-gpu test as `allreduceOp.cpp` is not included in the multi-GPU related list:
https://github.com/NVIDIA/TensorRT-LLM/blob/32d1573c43dc174224d191b74ec07f8e89fdef5a/jenkins/L0_MergeRequest.groovy#L555

QiJune (approved):
The L0_MergeRequest.groovy part looks good to me.

hyukn (approved):

PR #5433, opened 2025-06-24
tests: Set kv cache free memory fraction in test case

HuiGao-NV:
Refer to https://nvbugspro.nvidia.com/bug/5340908.
This failure is caused by commit 7e6d06d5d which can get more kv cache memory during memory estimation, however it means less memory left for buffer.
After testing, this case can pass with free memory fraction of 0.85 and it get more kv cache memory than code without commit 7e6d06d5d.
So we can has this commit and set a memory fraction in case.


HuiGao-NV:
/bot skip --comment="this is not a case ran in pre-merge and post-merge."

xinhe-nv (approved):

PR #5435, opened 2025-06-24
[TRTLLM-5965] perf: Optimize MoE sort kernels for large-scale EP

syuoni:
# [TRTLLM-5965] perf: Optimize MoE sort kernels for large-scale EP

## Description

This PR implements the sort logics before MoE GEMMs, and replaces the original CUB sort invocation.

In a typical large-scale EP workload (EP=32 and per-gpu batch=128):

* Before this PR: 5 kernels
    * buildExpertMapsKernel: 10.3 us
    * CUB sort (three kernels): 11.9 us
    * computeExpertFirstTokenOffsetKernel: 4.6 us
    * In addition, we see significant bubbles between CUB kernels on B200.
image

* After this PR: 3 kernels
   * blockExpertPrefixSumKernel: 2.3 us 
   * globalExpertPrefixSumKernel: 2.3 us 
   * mergeExpertPrefixSumKernel: 2.4 us
image

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
+++ b/cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
@@ -847,6 +847,31 @@ void mergeExpertPrefixSum(int const* blocked_expert_counts, int const* blocked_e
         unpermuted_row_to_permuted_row, num_tokens);
 }
 
+// threeStepBuildExpertMapsSortFirstToken uses three kernels to achieve the sort of token_selected_experts
+
+// 1. blockExpertPrefixSumKernel launches [num_experts_per_node, num_blocks_per_seq] CTAs; each CTA has
+// num_tokens_per_block threads. blocked_row_to_unpermuted_row points to a 2D buffer of size [num_experts_per_node,
+// num_tokens], which can be viewed as [num_experts_per_node, num_blocks_per_seq] blocks, and each block has
+// num_tokens_per_block tokens. Note that each CTA corresponds to a block in blocked_row_to_unpermuted_row. Within each
+// CTA, the threads leverage cub::BlockScan to compute the offsets of tokens that activate the target expert. If a
+// thread's token activates the target expert, the thread stores its unpermuted_row to the buffer block with the offset.
+// In addition, the kernel also stores the expert counts for each block to another 2D buffer blocked_expert_counts of
+// size [num_experts_per_node, num_blocks_per_seq].
+
+// 2. globalExpertPrefixSumKernel launches 1 CTA; that CTA has num_experts_per_node * num_blocks_per_seq threads.
+// The kernel views blocked_expert_counts as a 1D buffer, and leverages cub::BlockScan to compute the prefix sum of the
+// expert counts for each block. The prefix sum is stored to blocked_expert_counts_cumsum.
+
+// 3. mergeExpertPrefixSumKernel launches [num_experts_per_node, num_blocks_per_seq] CTAs; each CTA has
+// num_tokens_per_block threads. Each CTA obtains the block-level offset from blocked_expert_counts_cumsum, and thus
+// compacts blocked_row_to_unpermuted_row to permuted_row_to_unpermuted_row. In addition, with the block-level offsets,
+// the kernel fills permuted_token_selected_experts.
+
+// computeNumTokensPerBlock decides num_tokens_per_block. Note that both blockExpertPrefixSumKernel and
+// globalExpertPrefixSumKernel leverage cub::BlockScan, and their CTA sizes are num_tokens_per_block and
+// num_experts_per_node * num_blocks_per_seq, respectively. computeNumTokensPerBlock tries to find a minimum CTA size
+// for both kernels, so that the block-leval cub::BlockScan can be efficient.
+
 void threeStepBuildExpertMapsSortFirstToken(int const* token_selected_experts, int* permuted_token_selected_experts,
     int* permuted_row_to_unpermuted_row, int* unpermuted_row_to_permuted_row, int64_t* expert_first_token_offset,
     int* blocked_expert_counts, int* blocked_expert_counts_cumsum, int* blocked_row_to_unpermuted_row,
--- a/cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
+++ b/cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
@@ -305,7 +305,7 @@ void buildMinLatencyActiveExpertMaps(int* num_active_experts_per_node, float* ex
 
 template <int BLOCK_SIZE, int EXPERTS_PER_TOKEN, int LOG2_NUM_EXPERTS>
 __global__ void fusedBuildExpertMapsSortFirstTokenKernel(int const* const token_selected_experts,
-    int* const permuted_source_token_ids, int* const expanded_source_row_to_expanded_dest_row,
+    int* const permuted_source_token_ids, int* const unpermuted_row_to_permuted_row,
     int64_t* const expert_first_token_offset, int64_t const num_tokens, int const experts_per_token,
     int const start_expert, int const end_expert, int const num_experts_per_node)
 {
@@ -377,11 +377,11 @@ __global__ void fusedBuildExpertMapsSortFirstTokenKernel(int const* const token_
 #pragma unroll
         for (int i = 0; i < EXPERTS_PER_TOKEN; i++)
         {
-            // Note: permuted_source_token_ids is an alias to expanded_dest_row_to_expanded_source_row.
+            // Note: permuted_source_token_ids is an alias to permuted_row_to_unpermuted_row.
             int const source_token_id = i * num_tokens + token;
             int const dest_token_id = local_token_permuted_indices[i];
             permuted_source_token_ids[dest_token_id] = source_token_id;
-            expanded_source_row_to_expanded_dest_row[source_token_id] = dest_token_id;
+            unpermuted_row_to_permuted_row[source_token_id] = dest_token_id;
         }
     }
 
@@ -398,7 +398,7 @@ __global__ void fusedBuildExpertMapsSortFirstTokenKernel(int const* const token_
 
 template <int BLOCK_SIZE, int EXPERTS_PER_TOKEN, int LOG2_NUM_EXPERTS>
 bool fusedBuildExpertMapsSortFirstTokenDispatch(int const* token_selected_experts, int* permuted_source_token_ids,
-    int* expanded_source_row_to_expanded_dest_row, int64_t* expert_first_token_offset, int64_t const num_tokens,
+    int* unpermuted_row_to_permuted_row, int64_t* expert_first_token_offset, int64_t const num_tokens,
     int const num_experts_per_node, int const experts_per_token, int const start_expert, int const end_expert,
     cudaStream_t stream)
 {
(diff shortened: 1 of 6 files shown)


qiaoxj07 (approved):

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -539,57 +538,86 @@ bool fusedBuildExpertMapsSortFirstToken(int const* token_selected_experts, int*
...
+__global__ void blockExpertPrefixSumKernel(int const* token_selected_experts, int* block_expert_counts,
+    int* block_source_token_ids, int64_t const num_tokens, int64_t const num_experts_per_token,
+    int const start_expert_id)
+{
+    using BlockScan = cub::BlockScan<int, kNumTokensPerBlock>;
+    __shared__ typename BlockScan::TempStorage temp_storage;
+
+    // target_expert_id and expert_id are offset by start_expert_id
+    int const target_expert_id = blockIdx.x;
+    int const block_id = blockIdx.y;
```
djns99, line 572: Why are we launching an entire block for every expert? This is going to be very wasteful for large expert count + large seq length.

My understanding is you are trying to produce 2 values.

1. The offset for this token within the current block
2. The total number of elements to the current block

This can be derived from the `rank` and `exclusive_digit_prefix` parameter in `BlockRadixRank` as we do in the original `fusedBuildExpertMapsSortFirstTokenKernel`.

So are we not just better off storing those values and then calculating the correct values in `globalExpertPrefixSumKernel` (this will also be much simpler since we already have the prefix sum, so it is just a `BlockReduce` instead)

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -601,117 +629,217 @@ void buildExpertMaps(int const* token_selected_experts, int* unpermuted_token_se
...
+        num_experts_per_node, num_blocks_per_seq);
+}
+
+template <int kNumTokensPerBlock>
+__global__ void mergeExpertPrefixSumKernel(int const* block_expert_counts, int const* block_expert_counts_cumsum,
+    int const* block_source_token_ids, int* permuted_token_selected_experts, int* permuted_source_token_ids,
+    int* expanded_source_row_to_expanded_dest_row, int const num_tokens)
+{
+    int const target_expert_id = blockIdx.x;
+    int const block_id = blockIdx.y;
```
djns99, line 811: See comment above about using `BlockRadixRank` we can reduce this to only `num_tokens*topk` threads.

The final permuted idx is:

```
selected_expert = token_selected_experts[blockIdx.x * threadDim.x + threadIdx.x];
dest_token_id = expert_first_token_offset[selected_expert] + (block_rank[blockIdx.x][threadIdx.x] - block_exclusive_digit_prefix[blockIdx.x][selected_expert]);
```

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -601,117 +629,217 @@ void buildExpertMaps(int const* token_selected_experts, int* unpermuted_token_se
...
+    }
+    else if (num_elements <= 1024 * 8)
+    {
+        func = globalExpertPrefixSumKernel<1024, 8>;
+    }
+    cudaLaunchKernelEx(&config, func, block_expert_counts, block_expert_counts_cumsum, expert_first_token_offset,
+        num_experts_per_node, num_blocks_per_seq);
+}
+
+template <int kNumTokensPerBlock>
```
djns99, line 805: This template is unneeded, just pass a parameter

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -601,117 +629,217 @@ void buildExpertMaps(int const* token_selected_experts, int* unpermuted_token_se
...
+    cudaLaunchAttribute attrs[1];
+    attrs[0].id = cudaLaunchAttributeProgrammaticStreamSerialization;
+    attrs[0].val.programmaticStreamSerializationAllowed = tensorrt_llm::common::getEnvEnablePDL();
+    config.numAttrs = 1;
+    config.attrs = attrs;
 
-    // Note that expert goes [0, num_experts] (inclusive) because we want a count for the total number of active tokens
-    // at the end of the scan.
-    if (expert >= num_experts_per_node + 1)
+    // This allows accommodating 256 experts x 64k tokens; reasonable workload should not exceed this.
```
djns99, line 757: I know of some users using 1 million tokens, this is not a good assumption

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -601,117 +629,217 @@ void buildExpertMaps(int const* token_selected_experts, int* unpermuted_token_se
...
+    assert(num_elements <= 1024 * 16);
```
djns99, line 758: We should not use asserts for inputs controlled by users. This should be `TLLM_CHECK_WITH_INFO()`. This assert will never fire for a customer who passes too many tokens

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -601,117 +629,217 @@ void buildExpertMaps(int const* token_selected_experts, int* unpermuted_token_se
...
+
+#if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
+    asm volatile("griddepcontrol.launch_dependents;");
+#endif
 }
 
-// Calculates the start offset of the tokens for a given expert. The last element is the total number of valid tokens
-__global__ void computeExpertFirstTokenOffsetKernel(int const* sorted_experts, int64_t const sorted_experts_len,
-    int64_t const num_experts_per_node, int64_t* expert_first_token_offset)
+void globalExpertPrefixSum(int const* block_expert_counts, int* block_expert_counts_cumsum,
```
djns99, line 740: If we use `BlockRadixRank` we can replace this with cub `DeviceReduce`, though at that point it is probably easier just to directly implement the decoupled look back in the first kernel

djns99 (dismissed):
Thanks for this work! I think we need to get rid of this assumption before we can merge this though unfortunately:

```
// This allows accommodating 256 experts x 64k tokens; reasonable workload should not exceed this
```
I also think we should try be less wasteful with our block sizes. In the worst assumed case above (assuming topk=8) we are launching 16M threads, of which only 256k contribute anything

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
```diff
@@ -1433,8 +1619,8 @@ __global__ void finalizeMoeRoutingKernel(GemmOutputType const* expanded_permuted
         for (int k_idx = 0; k_idx < experts_per_token; ++k_idx)
         {
             int64_t const k_offset = original_row * experts_per_token + k_idx;
-            int64_t const expert_idx = expert_for_source_row[k_offset];
-            if (expert_idx >= num_experts_per_node)
+            int64_t const expert_id = expert_for_source_row[k_offset] - start_expert_id;
```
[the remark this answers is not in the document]
syuoni: Yes. The GEMM2 doesn't rely on the original mapped expert IDs (namely `unpermuted_token_selected_experts`).

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
syuoni: If I understand correctly, the major difference is `BlockRadixRank` approach handles all experts in each CTA, while the current `BlockScan` approach handles one expert in each CTA.

Actually, in a typical workload, `BlockRadixRank` approach tends to launch too few CTAs, resulting in low occupancy. Taking DeepSeek-R1 and EP8 as an example, let's say the CTA size is 256, then
* In a prefill step, the `num_tokens` is 8K 
    * The `BlockRadixRank` approach will launch 8K/256 = 32 CTAs
    * The `BlockScan` approach launches 256 / 8 \* 32 = 1024 CTAs
* In a generation step, the `num_tokens` is 256
    * The `BlockRadixRank` approach will launch 1 CTA
    * The `BlockScan` approach launches 256 / 8 = 32 CTAs

Although there are some duplicated compution between CTAs of a same target experts, launching more CTAs can parallelize the workload and better leverages the SMs.

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
syuoni: I don't think I fully understand your comment. If using `BlockRadixRank`, what is the `gridDim` and `blockDim`?

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
syuoni: Removed. Thanks for pointing this out.

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
syuoni: Removed the restriction.

I noticed the `PermuteVeryLongSequence` test which uses 1M tokens. I tested the correctness of kernel with that test.

Out of curiosity, why do the users want 1M tokens? I thought in long context, we typically use chunked context. 1M tokens is likely OOM when allocating workspaces.

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
syuoni: Removed the restriction. No assertion anymore.

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
syuoni: I think one of the motivations of the PR is removing device-level CUB function invocation, which might use `memSetAsync` and lead to bubbles when using CUDA Graph.

Agree that decoupled look back is the best solution. If I understand correctly, with decoupled look back, the global offset can be obtained in the first kernel. If so, all the three kernels can be fused together.

juney-nvidia (approved):

juney-nvidia:
Let's merge this PR to unblock the E2E optimization of Lage-EP and continue the refinements in the subsequent PRs.

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
djns99: > Actually, in a typical workload, BlockRadixRank approach tends to launch too few CTAs, resulting in low occupancy

This is true, but the problem I see here is that you are not actually adding any useful work. You are just launching more CTAs that return immediately. We should instead launch exactly enough threads (num_tokens * topk) to parallelise all the work and no more, which is what the BlockRadixRank will do. Remember that we have the shared expert overlapped so any free CTAs can in theory be utilised by that.

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
djns99: The total number of threads should be `num_tokens*topk` we can divide these into blocks however we want. Its an embarassingly parallel operation in the case of `mergeExpertPrefixSumKernel`

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
djns99: This is still used by the TRT path which I think does not support chunking. Regardless, I think MOE should not impose specific assumptions on customers as to how they want to distribute their work

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
djns99: > removing device-level CUB function invocation

This is not the issue. CUB will almost always be faster than anything we can write (many NVIDIA experts have spent years fine tuning it). The real problem is that we were using a Radix Sort which is a much more complicated algorithm than we actually needed in this situation. If we actually needed to sort I doubt we would be writing anything faster

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
syuoni: I see your point. Yes, `mergeExpertPrefixSumKernel` can be optimized as I reply above, thanks!

cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/moe_kernels.cu
syuoni: > If we actually needed to sort I doubt we would be writing anything faster

For sure I agree this. The problem is, device-level CUB function invocation might cause bubbles when using CUDA Graph. This issue has been verified on `cub::DeviceRadixSort`.

PR #5449, opened 2025-06-24
feature: support to nemotron-h-4b

vegaluisjose:
# feature: support to nemotron-h-4b

## Description

This PR adds support for [Nemotron-H-4B](https://arxiv.org/pdf/2504.11409) @tomeras91 @suyoggupta 

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_nemotron_h.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_h.py
@@ -137,7 +137,7 @@ class NemotronHLayer(DecoderLayer):
             self.mixer = Mamba2Mixer(d_model=config.hidden_size,
                                      d_state=config.ssm_state_size,
                                      d_conv=config.conv_kernel,
-                                     num_heads=config.mamba_num_heads,
+                                     nheads=config.mamba_num_heads,
                                      n_groups=config.n_groups,
                                      head_dim=config.mamba_head_dim,
                                      chunk_size=config.chunk_size,
--- a/tensorrt_llm/_torch/models/modeling_nemotron_h.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_h.py
@@ -137,7 +137,7 @@ class NemotronHLayer(DecoderLayer):
             self.mixer = Mamba2Mixer(d_model=config.hidden_size,
                                      d_state=config.ssm_state_size,
                                      d_conv=config.conv_kernel,
-                                     expand=config.expand,
+                                     num_heads=config.mamba_num_heads,
                                      n_groups=config.n_groups,
                                      head_dim=config.mamba_head_dim,
                                      chunk_size=config.chunk_size,
@@ -247,6 +247,8 @@ class NemotronHForCausalLM(DecoderModelForCausalLM[NemotronHModel,
                 for k in model_config.quant_config.exclude_modules
             ]
 
+        model_config.pretrained_config.head_dim = model_config.pretrained_config.attention_head_dim
+
         super().__init__(
             NemotronHModel(model_config),
             config=model_config,
@@ -258,7 +260,7 @@ class NemotronHForCausalLM(DecoderModelForCausalLM[NemotronHModel,
         config = self.model_config.pretrained_config
         tp_size = self.model_config.mapping.tp_size
         tp_rank = self.model_config.mapping.tp_rank
-        d_inner = config.hidden_size * config.expand
+        d_inner = config.mamba_head_dim * config.mamba_num_heads
         n_groups = config.n_groups
         d_state = config.ssm_state_size
         nheads = d_inner // config.mamba_head_dim
--- a/tensorrt_llm/_torch/models/modeling_nemotron_h.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_h.py
@@ -247,6 +247,9 @@ class NemotronHForCausalLM(DecoderModelForCausalLM[NemotronHModel,
                 for k in model_config.quant_config.exclude_modules
             ]
 
+        # Nemotron-H config.json uses attention_head_dim instead of head_dim which makes the
+        # attention module use the wrong head_dim, based on hidden_size and num_attention_heads.
+        # for this model, we need to override the head_dim to attention_head_dim.
         model_config.pretrained_config.head_dim = model_config.pretrained_config.attention_head_dim
 
         super().__init__(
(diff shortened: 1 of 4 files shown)


suyoggupta (approved):

tensorrt_llm/_torch/models/modeling_nemotron_h.py
```diff
@@ -247,6 +247,8 @@ def __init__(
                 for k in model_config.quant_config.exclude_modules
             ]
 
+        model_config.pretrained_config.head_dim = model_config.pretrained_config.attention_head_dim
```
tomeras91, line 250: Why is this added? Where is it used?

tensorrt_llm/_torch/modules/mamba/mamba2_mixer.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         tp_rank = config.mapping.tp_rank
         tp_size = config.mapping.tp_size
 
-        d_inner = d_model * expand
+        d_inner = head_dim * num_heads
         nheads = d_inner // head_dim
```
tomeras91, line 64: if `num_heads` is passed as input, we don'y need to compute it again from d_inner..

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -532,7 +532,7 @@ def __init__(
         tp_size = mapping.tp_size
 
         # derive mamba parameters for conv and ssm states
-        d_inner = d_model * expand
+        d_inner = head_dim * num_heads
         conv_dim = d_inner + 2 * n_groups * d_state
         nheads = d_inner // head_dim
```
tomeras91, line 537: same here. No need to compute `nheads` if it is passed as input (`num_heads`)

tensorrt_llm/_torch/models/modeling_nemotron_h.py
```diff
@@ -258,7 +260,7 @@ def load_weights(self, weights: Dict):
         config = self.model_config.pretrained_config
         tp_size = self.model_config.mapping.tp_size
         tp_rank = self.model_config.mapping.tp_rank
-        d_inner = config.hidden_size * config.expand
```
tomeras91, line 261: Generally, why is the change from computing `d_inner` using `expand` to computing it by `mamba_head_dim*mamba_num_heads` necessary? Can you add an explanation in the PR description?

tomeras91 (commented):
Overall looks OK, but I don't understand why removing `expand` is necessary.

tensorrt_llm/_torch/models/modeling_nemotron_h.py
vegaluisjose: All NemotronH models use `attention_head_dim` instead of `head_dim` in their config file. Also, our attention module check if this is present otherwise it does `hidden_size // num_heads` which is not correct for the upcoming model, see here https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/modules/attention.py#L69-L70

tensorrt_llm/_torch/models/modeling_nemotron_h.py
vegaluisjose: this is what @JRD971000 mentioned about it:

"Previously, the number of mamba heads was affected by the hidden size of the model; mamba_num_heads = expand * hidden_size / mamba_head_dim. With this refactor, mamba_num_heads is set directly, and it is decoupled from the hidden size of the model."

Link [here](https://github.com/vllm-project/vllm/pull/19766#issuecomment-2982020537)

tensorrt_llm/_torch/pyexecutor/resource_manager.py
vegaluisjose: fixed

tensorrt_llm/_torch/models/modeling_nemotron_h.py
tomeras91: Got it. Worth adding an explanation in PR description for easier tracking in the future

tomeras91:
Cool
Given the new config of Nemotron-H-4B, I think it's worth adding it to the unittests. The changes in this PR are crucial for its correctness, and without adding it to the unittests they will cause a silent failure of Nemotron-H-4B if accidentally reverted. I guess we can use HF or this initial implementation to get reference generations / logprobs.
I'd even go the extra mile and replace the Nemotron-H-8B test with one for Nemotron-H-4B (as it's a smaller model), but I guess we don't have M-core logprobs for the 4B model.

BTW, now that `expand` is irrelevant for Nemotron-H-4B, it means that the huggingface implementation of the hybrid cache is broken. See [here](https://huggingface.co/nvidia/Nemotron-H-4B-Base-8K/blob/main/modeling_nemotron_h.py#L174)
I don't know who's in charge of that but it should be fixed.

vegaluisjose:
> Given the new config of Nemotron-H-4B, I think it's worth adding it to the unittests. The changes in this PR are crucial for its correctness, and without adding it to the unittests they will cause a silent failure of Nemotron-H-4B if accidentally reverted. I guess we can use HF or this initial implementation to get reference generations / logprobs.
I'd even go the extra mile and replace the Nemotron-H-8B test with one for Nemotron-H-4B (as it's a smaller model), but I guess we don't have M-core logprobs for the 4B model.

Yes, we will need the logprobs and also put the model weights in `llm_models_root`. I ran the unittest locally with 8b and it seems to be working.

>BTW, now that expand is irrelevant for Nemotron-H-4B, it means that the huggingface implementation of the hybrid cache is broken. See [here](https://huggingface.co/nvidia/Nemotron-H-4B-Base-8K/blob/main/modeling_nemotron_h.py#L174). I would actually suggest to remove expand from config.json and configuration_nemotron_h.py as it doesn't have any effect and is just misleading.

Yes, I think this is part of the TODO list.

tomeras91:
> BTW, now that expand is irrelevant for Nemotron-H-4B, it means that the huggingface implementation of the hybrid cache is broken. See [here](https://huggingface.co/nvidia/Nemotron-H-4B-Base-8K/blob/main/modeling_nemotron_h.py#L174). I would actually suggest to remove expand from config.json and configuration_nemotron_h.py as it doesn't have any effect and is just misleading.

Another option - just change the value of `expand` such that `mamba_head_dim*mamba_num_heads = hidden_size * expand`. That's probably the correct thing to do - keep `expand`'s original intent. If we do that, no need for any code changes here

PR #5464, opened 2025-06-25
Fix:  fix nvbug 5356427

HuiGao-NV:
fix typo in  in "with" syntax to make debugMode exit and enable test cases


QiJune (approved):

tests/unittest/_torch/thop/test_moe.py
```diff
@@ -579,8 +579,6 @@ def quant_dequant_per_tensor_fp8(a):
                          ids=["autotune", "no_autotune"])
 def test_moe_fp8(num_tokens, expert_info, hidden_size, intermediate_size,
                  use_autotune):
-    if use_autotune:
```
QiJune, line 582: Should we revert this change?

HuiGao-NV: No need to revert this.
This change also skip above test case, then all cases in test_moe can pass.

PR #5470, opened 2025-06-25
[fix][test] remove test in global scope

omera-nv:
# [fix][test] Remove test from running in global scope


lucaslie (approved):
Thank you! That must have been a typo that slipped through

PR #5476, opened 2025-06-25
feat: reduce unnecessary kernel generation

tongyuantongyu:
## Description

Reduce compile time and binary size by only compile for architectures each kernel designed for.

Notable extra changes: 

- Switch to CMake's  `CUDA_ARCHITECTURES` support for convenient architecture configuration.
- Compile for family conditional targets, to prepare for future architectures.
- Switch to Ninja when building on CI for better job parallelization.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


cpp/CMakeLists.txt
```diff
@@ -362,6 +363,9 @@ if(BUILD_PYT)
     )
   endif()
   unset(ENV{TORCH_CUDA_ARCH_LIST})
+  # Placeholder.
+  message(STATUS "Set TORCH_CUDA_ARCH_LIST to placeholder value \"8.0\".")
+  set(TORCH_CUDA_ARCH_LIST "8.0")
```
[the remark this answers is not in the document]
yuantailing: There should be a comment, or I don't know the placeholder is to be removed. And I prefer to remove `message(STATUS ...)` because it misleads the user to think some modules are compiled with 8.0 while 8.0 has no effects.

cpp/tensorrt_llm/kernels/cutlass_kernels/CMakeLists.txt
```diff
@@ -46,64 +49,94 @@ set(INSTANTIATION_GENERATION_DIR
...
+    target_compile_definitions(${target_name} PUBLIC COMPILE_HOPPER_TMA_GEMMS)
+    target_compile_definitions(${target_name}
+                               PUBLIC COMPILE_HOPPER_TMA_GROUPED_GEMMS)
+  endif()
 
-# Sort both lists to ensure order doesn't matter
-list(SORT _KERNEL_GEN_OUTPUT)
-list(SORT INSTANTIATIONS_GENERATED)
+  if(${enable_blackwell} AND ("100" IN_LIST CMAKE_CUDA_ARCHITECTURES_ORIG
+                              OR "120" IN_LIST CMAKE_CUDA_ARCHITECTURES_ORIG))
```
djns99, line 74: What is the point of including 120 in this if condition if it doesnt do anything

cpp/tensorrt_llm/kernels/cutlass_kernels/allreduce_gemm/kernel/sm100_gemm_allreduce_tma_warpspecialized.hpp
```diff
@@ -418,7 +420,19 @@ class GemmARUniversal<ProblemShape_, CollectiveMainloop_, CollectiveEpilogue_, C
     CUTLASS_DEVICE
     void operator()(Params const& params, char* smem_buf)
     {
+        if constexpr (tensorrt_llm::kernels::arch::is_major_v<10>)
+        {
+            _invoke(params, smem_buf);
+        }
+        else
+        {
+            printf("ERROR : This kernel shall only run on SM10x devices.\n");
```
djns99, line 429: Can we add an assert too so that with NDEBUG off we actually get a crash.
Also to make it easier to debug build issues in the future can we print the function name and file/line number using`__PRETTY_FUNCTION__` ` __FILE__` and `__LINE__` macros

djns99 (approved):
LGTM, excellent work thank you!

1. How are you validating we aren't introducing any performance regressions?
2. You mentioned Ninja gives better parallelisation? Could you elaborate a bit further on what this entails i.e. is this an opt-in feature developers should be aware of, or does ninja just have a better implementation?

cpp/tensorrt_llm/kernels/cutlass_kernels/allreduce_gemm/kernel/sm100_gemm_allreduce_tma_warpspecialized.hpp
djns99: It would be good if we could perhaps have a way of checking this on the host side too to correctly propagate an error

cpp/tensorrt_llm/kernels/cutlass_kernels/CMakeLists.txt
tongyuantongyu: This is some leftover that became meaningless after we open sourced these kernels. Removed.

cpp/tensorrt_llm/kernels/cutlass_kernels/allreduce_gemm/kernel/sm100_gemm_allreduce_tma_warpspecialized.hpp
djns99: > This is device code so I want to keep the printf as simple as possible

It shouldn't be compiled in the `if constexpr` case right?

Can you do __PRETTY_FUNCTION__ instead of file? This way it should contain the relevant template instantiation as this may be relevant to debugging failures

tongyuantongyu:
@djns99 
> How are you validating we aren't introducing any performance regressions?

I didn't perform any test, but these changes should be conservative enough that I don't expect any difference in generated kernels. The only place that can have regression is we switching to family-conditional targets, and we can easily go back to arch-conditional if any regression is found during perf validation.


> You mentioned Ninja gives better parallelization? Could you elaborate a bit further on what this entails i.e. is this an opt-in feature developers should be aware of, or does ninja just have a better implementation?

If you `make -j N targets1 targets2` they are built one by one: https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#why-cant-i-make-multiple-targets-on-the-command-line-in-parallel. There's no such issue for Ninja.

Besides that, Ninja generally loads build rule much faster, always buffer the output, and by default use system core count as job count.

yuantailing:
CMake error at second build.

For the first build, CMAKE_CUDA_ARCHITECTURES is `100-real` and it's OK. But for the second build, CMAKE_CUDA_ARCHITECTURES is rewritten to `100f-real` so it cannot pass the check.

```
CMake Error at /usr/local/cmake-3.30.2-linux-aarch64/share/cmake-3.30/Modules/Internal/CMakeCUDAArchitecturesValidate.cmake:9 (message):
  CMAKE_CUDA_ARCHITECTURES:

    100f-real

  is not one of the following:

    * a semicolon-separated list of integers, each optionally
      followed by '-real' or '-virtual'
    * a special value: all, all-major, native

Call Stack (most recent call first):
  /usr/local/cmake-3.30.2-linux-aarch64/share/cmake-3.30/Modules/CMakeDetermineCUDACompiler.cmake:112 (cmake_cuda_architectures_validate)
  CMakeLists.txt:48 (project)


-- Configuring incomplete, errors occurred!
```

PR #5485, opened 2025-06-25
[5356427] fix: Remove the seq_len of 4096 from FP8 block scale MoE tuning configs.

hyukn:
The seq_len of 4096 will cause some unknown CUDA illegal memory access issue if run with some other tests consecutively.
Put a saturated upper bound for any sequence length larger than it.


DomBrown (approved):
LGTM, Yukun. Many thanks for helping to debug this while I've been OOTO

QiJune (approved):

PR #5519, opened 2025-06-26
fix: [https://nvbugspro.nvidia.com/bug/5349343] Fix mPtrExpertCounts allocation in MoE TRT-LLM backend (nvfp4)

ChristinaZ:
# Fix mPtrExpertCounts allocation in MoE TRT-LLM backend (nvfp4)

## Description
1. Fix mPtrExpertCounts allocation in MoE TRT-LLM backend (nvfp4)
@zhhuang-nv find that: for DeepSeek Lite, the MoE TRT-LLM backend with nvfp4 reports the following error:
```
[TensorRT-LLM][ERROR] CUDA runtime error in cudaMemsetAsync( data.mPtrExpertCounts, 0, static_cast<size_t>(2 * NumThreads) * sizeof(int32_t), (cudaStream_t) stream): invalid argument (/home/scratch.zhhuang_sw/project/TensorRT-LLM/cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/RoutingKernel.cu:1254).
```

Through our analysis, we identified the root cause: 
The data size used in cudaMemset (` cudaMemsetAsync(
            data.mPtrExpertCounts, 0, static_cast<size_t>(2 * NumThreads) * sizeof(int32_t)...`) exceeds the actual allocated memory size (`at::Tensor expert_count_histogram = at::detail::empty_cuda({((num_experts * 2 + 255) / 256) * 256},...`).  
 
So I tried to fix this bug by revising the code to 

```
    int const size_of_expert_count_histogram = max(num_experts * 2, 256 * 2);
    at::Tensor expert_count_histogram = at::detail::empty_cuda({size_of_expert_count_histogram},
       ...
```

2. Some routing kernels require a compute capability of 9.0 or higher, which the A30 does not support. I had to skip certain tests for now.  In this PR, I revised the unit test to use a better approach for skipping the unit tests.

### DIFF
--- a/cpp/tensorrt_llm/thop/fp4BlockScaleMoe.cpp
+++ b/cpp/tensorrt_llm/thop/fp4BlockScaleMoe.cpp
@@ -123,7 +123,8 @@ std::vector<torch::Tensor> fp4_block_scale_moe_runner(torch::Tensor const& routi
         {args.num_tokens, args.top_k}, routing_bias_dtype, routing_logits.device(), std::nullopt);
     at::Tensor expert_indexes = at::detail::empty_cuda(
         {args.num_tokens, args.top_k}, at::ScalarType::Int, routing_logits.device(), std::nullopt);
-    at::Tensor expert_count_histogram = at::detail::empty_cuda({((num_experts * 2 + 255) / 256) * 256},
+    int64_t const size_of_expert_count_histogram = std::max(num_experts * 2, int64_t(256 * 2));
+    at::Tensor expert_count_histogram = at::detail::empty_cuda({size_of_expert_count_histogram},
         at::ScalarType::Int, // 256 is the max number of threads per block and max number of experts
         routing_logits.device(), std::nullopt);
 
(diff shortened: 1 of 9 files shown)


cpp/tensorrt_llm/thop/fp4BlockScaleMoe.cpp
```diff
@@ -123,7 +123,8 @@ std::vector<torch::Tensor> fp4_block_scale_moe_runner(torch::Tensor const& routi
         {args.num_tokens, args.top_k}, routing_bias_dtype, routing_logits.device(), std::nullopt);
     at::Tensor expert_indexes = at::detail::empty_cuda(
         {args.num_tokens, args.top_k}, at::ScalarType::Int, routing_logits.device(), std::nullopt);
-    at::Tensor expert_count_histogram = at::detail::empty_cuda({((num_experts * 2 + 255) / 256) * 256},
+    int64_t const size_of_expert_count_histogram = std::max(num_experts * 2, int64_t(256 * 2));
```
zhhuang-nv, line 126: Should we just use `2 * 256` like what we do in `fp8BlockScaleMoe.cpp`?

ChristinaZ: It works for now. 
While for Llama4 and Qwen3 (they also use this code for moe nvfp4), the size of variable `expert_count_histogram` is `num_experts*2`. 
So I modified the code in case there might be a larger expert number in the future.

ChristinaZ: Yes, agree. Let me revise it.

zhhuang-nv (approved):

byshiue (approved):

PR #5532, opened 2025-06-27
[nvbug/5302638][nvbugs/5310314] fix _handle_cancelled_requests

QiJune:


LinPoly (commented):
Thanks for quick fix! Is this fix only for tp_size > 1? If not, can you please also unwaive [our request cancellation test](https://github.com/NVIDIA/TensorRT-LLM/blob/980030c8167b08cb4a4df83e13a9af9e0b60492b/tests/unittest/llmapi/apps/_test_openai_misc.py#L89) to validate the fix?

QiJune:
> Thanks for quick fix! Is this fix only for tp_size > 1? If not, can you please also unwaive [our request cancellation test](https://github.com/NVIDIA/TensorRT-LLM/blob/980030c8167b08cb4a4df83e13a9af9e0b60492b/tests/unittest/llmapi/apps/_test_openai_misc.py#L89) to validate the fix?

The test is unwaived. I had a offline discussion with @Shunkangz , he will revisit the cancel logic of attention dp.

kaiyux (approved):

PR #5563, opened 2025-06-28
Fix GEMM+AR fusion on blackwell

xavier-nvidia:
## Description

- Fix 'Assertion failed: file.is_open()' in GemmAllReducePlugin that occurs when TRT-LLM runs inside different container than used for building the engine - GemmAllReducePlugin would create kernel heuristics file in /tmp/, which won't exist in a different container. There are two solutions implemented 1) use static heuristics (default) which doesn't require heuristics file, and 2) specify directory of heuristics file via GEMM_AR_PLUGIN_ENGINE_DIR env variable.
- Adds GB200 support

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


cpp/tensorrt_llm/plugins/gemmAllReducePlugin/gemmAllReducePluginProfiler.cpp
```diff
@@ -58,10 +58,18 @@ void GemmAllReducePluginProfiler::deserializeFromOwnFile(GemmIdCore gemmId, Gemm
...
+    return envDir != nullptr;
+}
+
 std::string GemmAllReducePluginProfiler::getCacheFileName(GemmIdCore gemmId)
 {
     std::stringstream fileName;
-    fileName << "/tmp/gemm-AR";
+    char const* envDir = getenv("GEMM_AR_PLUGIN_ENGINE_DIR");
+    std::string directory = envDir ? std::string(envDir) : "/tmp/";
+    fileName << directory + "/gemm-AR";
```
symphonylyh, line 72: this is one way to fix it. the other way, do you know which dockerfile doesn't have `/tmp` on certain devices? I think it can also be fixed from dockerfile side, that we explicitly create the `/tmp` directory so it's always exist -- this can be a long-term solution to consider that doesn't block the merge of this PR.

symphonylyh (approved):

cpp/tensorrt_llm/plugins/gemmAllReducePlugin/gemmAllReducePluginProfiler.cpp
xavier-nvidia: The problem arises (before this fix), when the container used for building was different than used for running. During build I would create a profile file under `/tmp` but then during execution (in different container) this file wouldn't exist anymore. So whether `/tmp` exists isn't the problem. This environment variable will allow the user to specify a directory which is persistent across different containers. 

By default this is not used as it will use static heuristics, so no profile file needed.

PR #5566, opened 2025-06-28
[feat][test] reuse MPI pool executor across tests

omera-nv:
# [feat][test] reuse MPI pool executor across tests

Initializing an `MPIPoolExecutor` involves some delay, and it accumulates with parameterization. This PR makes the pool executor a module-scoped fixture to be reused by multiple tests, reducing test times significantly.


Naveassaf (approved):
LGTM. Nice find.

PR #5567, opened 2025-06-28
Refactor the topk parallelization part for the routing kernels

ChristinaZ:
# Refactor the topk part for the routing kernels in the MoE TrtLLMGen backend

## Description

This is the first pull request (PR) for refactoring the routing kernels in the MoE TrtLLMGen backend.
In this PR, I initially relocated the topK parallelization logic from `cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/RoutingKernel.cu` to a new CUDA header file: `cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/RoutingKernelTopK.cuh`.

Also, to facilitate future refactoring efforts, I have adjusted the namespace configuration.

## Test Coverage
```
cd cpp/build
make -j$(nproc) google-tests
./tests/unit_tests/kernels/routingKernelsTest

pytest -k test_moe_fp4 tests/unittest/_torch/thop/test_moe.py
```


cpp/tensorrt_llm/kernels/trtllmGenKernels/blockScaleMoe/RoutingKernel.h
```diff
@@ -26,14 +26,23 @@
...
 {
-////////////////////////////////////////////////////////////////////////////////////////////////////
 
 namespace tg = batchedGemm::trtllm::gen;
 
+template <typename TypeExpW>
+struct PackedScoreIdx
+{
+    TypeExpW score;
+    int16_t idx; // @TODO: Might use int8_t as the number of experts is 128
```
MatthiasKohl, line 38: note: I thought about this before, but because there are only a few possibilities for the score:

- 4-byte floating point (or more like double): it doesn't matter whether we use 1-byte or 2-byte for idx because `PackedScoreIdx` cannot fit in register / coupled instructions anyway
- 2-byte floating point (or lower): it doesn't matter whether we use 1-byte or 2-byte for idx because it will always fit in 4 bytes, and there are no better instructions for e.g. 2-byte structures compared to 4-byte

MatthiasKohl (approved):
LGTM, just added one minor comment/note

byshiue (approved):

PR #5571, opened 2025-06-29
[CI] reduce mamba2 ssm test parameterization

tomeras91:
Mamba2 SSM unittests were over parametrized and took a long time to run. This PR reduces the number of test by changing one parameter at a time instead of running the full cartesian product. Reduced the number of tests from 144 to 61.

Locally, this reduces test time of tests/unittest/_torch/thop/test_mamba2_chunk_ss_update.py from ~80secs to ~50secs.


omera-nv (approved):
This seems to save 10 minutes of test time on H100 - nice!

PR #5578, opened 2025-06-29
[ci] remove MMLU if followed by GSM8K

omera-nv:
# [ci] remove MMLU if followed by GSM8k

We do not expect any model to fail MMLU while passing GSM8K, so we do not need to evaluate MMLU if we evaluate GSM8K.

This should reduce test time by about 15%.


syuoni (approved):
LGTM, thanks!

PR #5610, opened 2025-06-30
chore: enhance yaml loading arbitrary options in LlmArgs

Superjomn:
Enhance the YAML loading in LLM, and it can accept any knobs now.

### DIFF
--- a/tensorrt_llm/llmapi/llm_args.py
+++ b/tensorrt_llm/llmapi/llm_args.py
@@ -1937,15 +1937,9 @@ def update_llm_args_with_extra_dict(
         "quant_config": QuantConfig,
         "calib_config": CalibConfig,
         "build_config": BuildConfig,
-        "kv_cache_config": KvCacheConfig,
         "decoding_config": DecodingConfig,
         "enable_build_cache": BuildCacheConfig,
-        "peft_cache_config": PeftCacheConfig,
-        "scheduler_config": SchedulerConfig,
         "speculative_config": DecodingBaseConfig,
-        "batching_type": BatchingType,
-        "extended_runtime_perf_knob_config": ExtendedRuntimePerfKnobConfig,
-        "cache_transceiver_config": CacheTransceiverConfig,
         "lora_config": LoraConfig,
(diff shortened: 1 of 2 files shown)


tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1937,15 +1937,9 @@ def update_llm_args_with_extra_dict(
         "quant_config": QuantConfig,
         "calib_config": CalibConfig,
         "build_config": BuildConfig,
-        "kv_cache_config": KvCacheConfig,
```
FrankD412, line 1940: To make sure I'm clear, this is because these can be feed to the underlying pydantic `BaseModel` directly without any special handing right?

Superjomn: Yes, you are right. The remaining ones are not Pydantic, thus we leave them there to do explicit parsing.

tests/unittest/llmapi/test_llm_args.py
```diff
@@ -40,27 +40,100 @@ def test_LookaheadDecodingConfig():
...
+
+        llm_args = TrtLlmArgs(model=llama_model_path)
+        llm_args_dict = update_llm_args_with_extra_dict(llm_args.to_dict(),
+                                                        dict_content)
+        with pytest.raises(ValueError):
+            llm_args = TrtLlmArgs(**llm_args_dict)
+
+    def test_llm_args_with_build_config(self):
+        # build_config isn't a Pydantic
+        yaml_content = """
```
kaiyux, line 87: Nit: Should we just use dict to simplify the test implementation? `update_llm_args_with_extra_dict` does not accept raw yaml file anyway.

kaiyux (approved):

FrankD412 (approved):

tests/unittest/llmapi/test_llm_args.py
Superjomn: With a raw YAML string, I just hope to confirm that those strings are as expected.

Issue #3700, opened 2025-04-20
[Call for contributions] Feature parity between TensorRT and PyTorch backend - Sampling


juney-nvidia:
- [x] greedy search  
  - [ ] beam search
  - [x] top-k sampling
  - [x] top-p sampling
  - [x] stop words
  - [ ] penalty
  - [ ] bad words
  - [ ] return context logits
  - [ ] return generation logitx
  - [ ] return logprobs
  - [ ] num_return_sequences
  - [ ] top_p_min
  - [ ] no_repeat_ngram_size
  - [ ] min_p

karljang:
I'm closing this issue as completed. Thanks!

PR #5620, opened 2025-06-30
[fix] fix log probs and add for mtp and completion requests

pathorn:
fix log probs and add for mtp and completion requests
Part of feature parity with pytorch #3700

Note: Based on top of #4923 since changes to how `output.logprobs_diff` is calculated was required to avoid a race condition with mismatched token id lengths.

## Description

Logprobs currently only works for chat completion, and non-mtp backend. It also has some race conditions which causes the number of returned logprobs to be incorrect sometimes.

## Test Coverage

TestMTPSampleAndAcceptDraftTokens calls mtpworker.sample_and_accept_draft_tokens but does not verify the resulting log_probs.


pathorn:
We are still seeing slow performance with logprobs in production. However, this PR is stale and does not fix the actual performance issues, so I'm closing it.

PR #5623, opened 2025-06-30
[None][fix] Ignore context for presence and frequency penalties. Matches calculation in vllm.

pathorn:
[fix] Ignore context for presence and frequency penalties. Matches calculation in vllm.

This is an old change that only affects the c++ sampler, but we forgot to submit a PR. Given that the c++ sampler is still going to be the main implementation, it makes sense to fix it.

## Description

The presence penalty does not use tokens in the prompt, while repetition penalty does.

This changes two things in `penaltyKernels.cu` .

1. Adds a sentinel value -1 which will be used when a given token exists in the prompt. It will be ignored when counting output tokens and start from 1. In either case, this guaranteees `penaltyWorkspace[penaltyIndex]` is nonzero if a token exists in the prompt, but does not count the number of tokens in the prompt for the purpose of frequency and presence penalties.
2. Changes the condition for repetition penalty to `if (numOccurences != 0)` to allow for occurrences in the prompt to match. This should cause the above change to have no impact on repetition penalty, since it just cares if at least one match exists.

## Test Coverage

None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of token occurrence counts during penalty application, ensuring more accurate distinction between context and generated tokens.
  * Adjusted logic for applying repetition, presence, and frequency penalties for better token generation behavior.

### DIFF
--- a/cpp/tensorrt_llm/kernels/penaltyKernels.cu
+++ b/cpp/tensorrt_llm/kernels/penaltyKernels.cu
@@ -123,7 +123,7 @@ __global__ void batchApplyPenalty(T const* const* inputLogits, T* outputLogits, 
                 auto penaltyIndex = outputIdsPtr[batchSlot][beamIdx * maxSeqLen + step];
                 if (penaltyIndex < vocabSize)
                 {
-                    atomicAdd(&penaltyWorkspace[penaltyIndex], 1);
+                    penaltyWorkspace[penaltyIndex] = -1;
                 }
             }
         }
@@ -145,7 +145,7 @@ __global__ void batchApplyPenalty(T const* const* inputLogits, T* outputLogits, 
                 auto penaltyIndex = outputIdsPtr[batchSlot][beamIdx * maxSeqLen + currentStep - 1];
                 if (penaltyIndex < vocabSize)
                 {
-                    penaltyWorkspace[penaltyIndex] += 1;
+                    penaltyWorkspace[penaltyIndex] = max(0, penaltyWorkspace[penaltyIndex]) + 1;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


Pernekhan:
Any updates here?

jinyangyuan-nvidia:
Hi @jhaotingc, the previous implementation referred to both Huggingface and vLLM. Because vLLM used the same tokens  when computing repetition, presence, and frequency penalties at that time, the support for presence and frequency penalties in TensorRT-LLM also used the same tokens as repetition penalty in TensorRT-LLM and Huggingface (prompts are included).

Pernekhan:
This is a docs from vllm:
```
        presence_penalty: Float that penalizes new tokens based on whether they
            appear in the generated text so far. Values > 0 encourage the model
            to use new tokens, while values < 0 encourage the model to repeat
            tokens.
        frequency_penalty: Float that penalizes new tokens based on their
            frequency in the generated text so far. Values > 0 encourage the
            model to use new tokens, while values < 0 encourage the model to
            repeat tokens.
        repetition_penalty: Float that penalizes new tokens based on whether
            they appear in the prompt and the generated text so far. Values > 1
            encourage the model to use new tokens, while values < 1 encourage
            the model to repeat tokens.
```

I think if we initially followed vllm, we should keep following. I don't think this will change again in future.

jhaotingc:
cc @laikhtewari

jhaotingc:
Pasting to our internal channel to get some more vis. 💀

Pernekhan:
Any updates on this PR?

jhaotingc:
> Any updates on this PR?

@kaiyux do you have any thought?

pathorn:
This was already implemented in main behind a flag. Thanks for fixing it! I'll close the stale PR.

PR #5627, opened 2025-06-30
[feat] Implement pytorch sampler for MTP

pathorn:
[feat] Implement pytorch sampler for MTP

## Description

Previously, speculative decoding was always greedy and does not support sampling parameters.

This PR implements temperature, top-p, top-k and min-p sampling parameters in python when using MTP speculative decoding (for DeepSeek).

This also adds a log-probs return value from the pytorch sampler (Intended to be used together with #5620)

Future work: support this pytorch sampler for non-MTP models. Figure out why it is faster than the C++ sampler.

## Test Coverage

None

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -16,6 +16,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import psutil
 import safetensors
+from tensorrt_llm._torch.pyexecutor.llm_request import LlmRequest
 import torch
 import torch._dynamo.config
 
@@ -336,6 +337,7 @@ class PyTorchModelEngine(ModelEngine):
         lora_config: Optional[LoraConfig] = None,
         is_draft_model: bool = False,
     ):
+        torch.manual_seed(0)
         self.ub_buffers = None
         self.batch_size = batch_size
         self.max_num_tokens = max_num_tokens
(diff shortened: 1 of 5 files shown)


tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -852,7 +852,7 @@ def sample_and_accept_draft_tokens(
 
         # Strict acceptance
         else:
-            if self.is_thop:
+            if False:
```
netanel-haber, line 855: What is happening here?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -336,6 +337,7 @@ def __init__(
         lora_config: Optional[LoraConfig] = None,
         is_draft_model: bool = False,
     ):
+        torch.manual_seed(0)
```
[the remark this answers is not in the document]
pathorn: Without this line, TRTLLM would malfunction because each GPU would produce a different answer.

This was necessary to prevent divergent calculations on different GPUs with different seeds.

There's a larger problem that currently the user cannot pass in a seed for individual requests. Ideally we would not depend on the global torch seed, but for some reason this is difficult within PyTorch since its cuda graph seems to make some assumptions that break the ability to use a custom RNG.

tensorrt_llm/_torch/speculative/mtp.py
pathorn: We did not implement the C++ MTP sampling kernel for thop, so this is a hack to use the pytorch implementation of the sampler. I'll revert this line for now, but I'm not sure if the code added by this PR will work properly without forcing this case.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -16,6 +16,7 @@
 
 import psutil
 import safetensors
+from tensorrt_llm._torch.pyexecutor.llm_request import LlmRequest
```
QiJune, line 19: This line seems duplicate with line 24.

jhaotingc:
Hi, thanks @pathorn for contribution, @QiJune, @netanel-haber for reviewing, @nvxuanyuc and I will help getting this merged these weeks.

PR #5636, opened 2025-07-01
[fix] WAR to fix the illegal memory access issue in moe gemm on SM120

peaceh-nv:
## Description

Avoid using void elementc to fix the illegal memory access issue in moe gemm on SM120 after cutlass upgrade and add related unit tests on SM120


cpp/tensorrt_llm/kernels/cutlass_kernels/moe_gemm/launchers/moe_gemm_tma_ws_launcher.inl
```diff
@@ -348,7 +348,7 @@ using SafeBF16 = void;
                 EpilogueTileShape, ClusterShape,                                                        /**/                                                                                                                                      \
                 cutlass::epilogue::collective::EpilogueTileAuto,                                        /**/                                                                                                                                      \
                 ElementAccumulator, ElementAccumulator,                                                 /**/                                                                                                                                      \
-                ElementC, LayoutC*, AlignmentC,                                                         /**/                                                                                                                                      \
```
djns99, line 351: Since this only happens for SM120 Can you use:
`using EpilogueElementC = std::conditional_t<IsSM120, ElementCSafe, ElementC>;`

I believe setting it to void can have a small perf benefit in the case of `SM90/SM100` so lets not disable it in this case

peaceh-nv: Fixed in the latest commit. Thx!

djns99 (approved):

PR #5128, opened 2025-06-11
[TRTLLM-5835][feat] Optimized Mamba2Mixer prefill

tomeras91:
## Description

Currently on main, the mamba2 block forward pass has some dynamic memory allocations and host-to-device copies, greatly hurting its performance. This PR improves performance by minimizing these memory operations:

1. pre-allocate the `state_indices` tensors on device instead of moving it to device during forward pass
2. introduce `MambaMetadata` class holding `cu_seqlens` and `seq_idx` needed for varlen batched prefill of the SSM op. Compute them from `attn_metadata` at the start of the model forward pass instead of doing it inside the `Mamba2Mixer` block of each mamba layer. This also means we create these tensors once and not multiple times in each layer.
3. Remove the redundant for-loop on request type (prefill/decode) in the `Mamba2Mixer` forward pass. Replace it with 2 if statements, checking if we have prefills / decodes in the current batch.

These changes removed many of the GPU bubbles present in the mamba forward block, as seen in these profiles. They lead to **40% latency reduction**.

image
image

### Benchmarks
The modifications in this PR **improved Nemotron-H max throughput by ~15% and min prefill latency by ~40%** without sacrificing quality:

**MMLU results**
| Version | MMLU Score |
|---------|------------|
| Main | 69.98 |
| PR | 69.98 |

**Performance benchmarks**

Benchmark setting:
* Used genai-perf
* Done on a single RTX6000
* ISL / OSL = 223 / 140 (similar to ShareGPT statistics)
* nax_batch_size = 64
* Max throughput setting: --request-count 500 --concurrency 500 --warmup-request-count 10
* Min latency setting: --request-count 10 --concurrency 1 --warmup-request-count 1

&nbsp;
max throughput results:
| Metric | TRTLLM main | PR | Change [%] |
|--------|-------------|-----|------------|
| Request Latency [ms] | 30,614.38 | 26,482.98 | -13.5% |
| Time To First Token [ms] | 25,107.02 | 21,563.04 | -14.1% |
| Inter Token Latency [ms] | 40.91 | 36.97 | -9.6% |
| Output Token Throughput [tokens/sec] | 1,235.88 | 1,400.71 | +13.3% |
| Output Token Throughput Per User [tokens/sec/user] | 24.84 | 27.68 | +11.4% |
| Request Throughput [req/sec] | 9.14 | 10.43 | +14.1% |

&nbsp;
min latency results:
| Metric | TRTLLM main | PR | Change [%] |
|--------|-------------|-----|------------|
[...]

PR #5646, opened 2025-07-01
[TRTLLM-4923][feat] Enable CUDA graphs for Nemotron-H

tomeras91:
## Description

This PR adds CUDA graph support for Nemotron-H, by making adjustments to the mamba2 block forward pass and the mamba cache resource manager.

Replaying a CUDA graph correctly requires that all memory buffers are static, exist during graph capture, and no memory allocations are made in the graphed code. PR #5128 already removed most of the dynamic memory allocations in the mamba2 mixer forward. The only buffer left that did not exist during the graph capture phase was the `state_indices` tensor that holds the corresponding indices of the current requests in the mamba cache.

In this PR, we pre-allocate `state_indices` when creating the mamba cache with the maximum possible number of elements (max batch size). Before the forward pass, we assign the correct values in the first `batch_size` locations of `state_indices`. Therefore, state_indices is a static buffer that exists during graph capture. Just the values stored in that memory are changed before each graph replay.

### Tests
The current implementation assumes that the requests_ids sent to `prepare_mamba_cache_blocks` are the same ones that the forward pass is actually done on. `prepare_mamba_cache_blocks` is called during `resource_manager.prepare_resources(scheduled_batch)` in the executor loop. Then, the inputs to the forward pass are created inside the PyExecutor `_forward_step()`. To make sure the overlap scheduler, which adds inputs from the previous batch, doesn't break CUDA graph correctness (i.e. - doesn't change the request_ids between the call to `prepare_mamba_cache_blocks` and the forward pass itself), I added a unittest that asserts the outputs are the same with/without CUDA graph and/or overlap scheduler.

### Benchmarks
The modifications in this PR improved Nemotron-H min latency in settings where kernel launch overhead is substantial ( **+45% better output throughput, 30% better latency on  2 X RTX6000**). Max throughput performance, which is compute-bound, is un-affected. Quality does not change.

**MMLU results**
| Version | MMLU Score |
|---------|------------|
| No CG | 69.98 |
| With CG | 69.98 |

**Performance benchmarks**

Benchmark setting:
* Used genai-perf
* Done on 2 X RTX6000 (TP=2)
* ISL / OSL = 223 / 140 (similar to ShareGPT statistics)
* max_batch_size = 128
* Max throughput setting: --request-count 500 --concurrency 500 --warmup-request-count 10
* Min latency setting: --request-count 10 --concurrency 1 --warmup-request-count 1

&nbsp;
max throughput results
| Metric | No CG | With CG | Change [%] |
|--------|-------------|-----|------------|
| Request Latency [ms] | 15,533.22 | 15,436.64 | -0.6% |
| Time To First Token [ms] | 10,506.53 | 10,475.58 | -0.3% |
| Inter Token Latency [ms] | 38.62 | 37.98 | -1.7% |
| Output Token Throughput [tokens/sec] | 2,745.18 | 2,761.35 | +0.6% |
| Output Token Throughput Per User [tokens/sec/user] | 27.08 | 27.40 | +1.2% |
| Request Throughput [req/sec] | 20.46 | 20.57 | +0.5% |

&nbsp;
min latency results:
| Metric | No CG | With CG | Change [%] |
|--------|-------------|-----|------------|
| Request Latency [ms] | 2,044.73 | 1,412.66 | -30.9% |
| Time To First Token [ms] | 51.91 | 40.76 | -21.5% |
| Inter Token Latency [ms] | 14.34 | 9.87 | -31.2% |
| Output Token Throughput [tokens/sec] | 68.45 | 99.07 | +44.7% |
| Output Token Throughput Per User [tokens/sec/user] | 69.85 | 101.32 | +45.0% |
| Request Throughput [req/sec] | 0.49 | 0.71 | +44.9% |


shaharmor98 (approved):
LGTM

PR #5684, opened 2025-07-02
support TRTLLM_DEEP_EP_TOKEN_LIMIT to allow run deep-ep on memory-con…

ttyio:
# none: support TRTLLM_DEEP_EP_TOKEN_LIMIT to allow run deep-ep on memory-constrained GPUs.

## Description

DeepEP requires additional RDMA memory for communication, and on memory-constrained GPUs, we may not have enough memory to enable DeepEP for both the context and decoding phases. In disaggregated serving scenarios, it's straightforward to enable DeepEP only on the decoding server. However, for inflight batching, we need to apply a token limit so that DeepEP is only used during decoding.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_deepseekv3.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv3.py
@@ -512,11 +512,16 @@ class Deepseekv3MoE(nn.Module):
         # max-throughput
         use_dp_padding = False
         if self.use_dp and self.mapping.tp_size > 1:
+            # MoE use static heuristic to check alltoall enabled or not, however, for wide_ep, the alltoall could also be dynamically disabled when chunking is used or TRTLLM_DEEP_EP_TOKEN_LIMIT is hit.
+            is_wide_ep_alltoall_disabled = isinstance(
+                self.experts, WideEPMoE) and not self.experts.can_use_alltoall(
+                    hidden_states, all_rank_num_tokens)
+            alltoall_enabled = self.experts.enable_alltoall and not is_wide_ep_alltoall_disabled
+
             # FP4 all_gather moves this bf16 allgather in to after topk and fp4 quantization
             # to reduce allreduce BW
-            if (disable_fp4_allgather()
-                    and not self.experts.can_use_alltoall(hidden_states_fp4 or hidden_states)) or isinstance(
-                        self.experts, TRTLLMGenFusedMoE):
+            if (disable_fp4_allgather() and not alltoall_enabled) or isinstance(
+                    self.experts, TRTLLMGenFusedMoE):
                 hidden_states = allgather(hidden_states,
                                           self.mapping,
                                           dim=0,
--- a/tensorrt_llm/_torch/models/modeling_deepseekv3.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv3.py
@@ -515,7 +515,7 @@ class Deepseekv3MoE(nn.Module):
             # FP4 all_gather moves this bf16 allgather in to after topk and fp4 quantization
             # to reduce allreduce BW
             if (disable_fp4_allgather()
-                    and not self.experts.enable_alltoall) or isinstance(
+                    and not self.experts.can_use_alltoall(hidden_states_fp4 or hidden_states)) or isinstance(
                         self.experts, TRTLLMGenFusedMoE):
                 hidden_states = allgather(hidden_states,
                                           self.mapping,
--- a/tensorrt_llm/_torch/models/modeling_deepseekv3.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv3.py
@@ -62,8 +62,7 @@ from ..modules.multi_stream_utils import maybe_execute_in_parallel
 from ..modules.rms_norm import RMSNorm
 from ..peft.lora.layer import LoraLayer
 from ..speculative import MTPEagleWorker, MTPSpecMetadata, MTPWorker
-from ..utils import (AuxStreamType, EventType, Fp4QuantizedTensor,
-                     disable_fp4_allgather)
+from ..utils import AuxStreamType, EventType, Fp4QuantizedTensor
 from .modeling_utils import (DecoderModel, DecoderModelForCausalLM,
                              EagerFusionConfig, filter_weights,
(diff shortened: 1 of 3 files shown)


yilin-void (commented):
The code related to `self.use_postquant_alltoall` should include the `use_all_to_all` check as well. Otherwise, there will be a scenario where the AG/RS path is taken, but the post quant all2all logic is triggered, just like the following code snippet.

```
if not disable_fp4_allgather() or self.use_postquant_alltoall:
    if isinstance(x, Fp4QuantizedTensor):
        x, x_sf = x.fp4_tensor, x.scaling_factor
        x_row = x.shape[0]
        # note: we use uint8 to store 2 fp4 values
        x_col = x.shape[1] * 2
    else:
        sf_swizzle = not self.use_postquant_alltoall
        x_row = x.shape[0]
        x_col = x.shape[1]
        x, x_sf = torch.ops.trtllm.fp4_quantize(
            x, self.fc31_input_scale, self.scaling_vector_size,
            False, sf_swizzle)
        if self.use_postquant_alltoall:
            x_sf = x_sf.view((x_row, -1))
```

yuantailing (commented):
~There are a lot variables related to communication method selection: `enable_alltoall`, `use_all_to_all` (from this PR), and `use_allgather` (will be introduced by #5684 ), which makes the logic hard to follow.~

~This comment is not to block this PR, but suggest a future refactor on communication method selection.~

(Discuss in the following comment)

tensorrt_llm/_torch/modules/fused_moe/fused_moe_wide_ep.py
```diff
@@ -277,6 +281,16 @@ def enable_alltoall(self):
...
+    def can_use_alltoall(self, input):
+        num_tokens = input.shape[0]
+
+        # For DeepEPLowLatency, check if tokens exceed the threshold
+        if (self.alltoall_method_type == AlltoallMethodType.DeepEPLowLatency
+                and num_tokens > self.deep_ep_max_num_tokens):
+            return False
+
+        return self.enable_alltoall
+
```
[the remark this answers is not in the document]
yuantailing: I mean refactor in future PRs. Maybe `enable_alltoall` is the thing to be removed, need discussion after both PRs are marged.

Keep things simple in this PR, no need to change these stuffs.

yuantailing:
Here is some code related to `enable_alltoall`:

https://github.com/NVIDIA/TensorRT-LLM/blob/3e3b1769adc7c1158525f8116e910e99abad1d4b/tensorrt_llm/_torch/models/modeling_deepseekv3.py#L517-L521

Should this condition agree with which communication method is really used? Did you test this code path with the combination that `use_all_to_all=False` and `enable_alltoall=True`?

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -514,7 +514,8 @@ def compute_routed_output(self, hidden_states, hidden_states_fp4,
         if self.use_dp and self.mapping.tp_size > 1:
             # FP4 all_gather moves this bf16 allgather in to after topk and fp4 quantization
             # to reduce allreduce BW
-            if disable_fp4_allgather() and not self.experts.enable_alltoall:
+            if disable_fp4_allgather() and not self.experts.can_use_alltoall(
+                    hidden_states_fp4 or hidden_states):
```
yuantailing, line 518: 1. The method `can_use_alltoall` is not a method of the base class. It raises if `self.experts` is a instance of `CutlassFusedMoE`.
2. If hidden_states_fp4 is a Tensor, `hidden_states_fp4 or hidden_states` will raise.

yilin-void (approved):

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -512,11 +512,16 @@ def compute_routed_output(self, hidden_states, hidden_states_fp4,
...
+                    hidden_states, all_rank_num_tokens)
+            alltoall_enabled = self.experts.enable_alltoall and not is_wide_ep_alltoall_disabled
+
             # FP4 all_gather moves this bf16 allgather in to after topk and fp4 quantization
             # to reduce allreduce BW
-            if (disable_fp4_allgather()
-                    and not self.experts.enable_alltoall) or isinstance(
-                        self.experts, TRTLLMGenFusedMoE):
+            if (disable_fp4_allgather() and not alltoall_enabled) or isinstance(
+                    self.experts, TRTLLMGenFusedMoE):
```
hlu1, line 524:
```suggestion
            if isinstance(self.experts, TRTLLMGenFusedMoE):
```

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -512,11 +512,16 @@ def compute_routed_output(self, hidden_states, hidden_states_fp4,
         # max-throughput
         use_dp_padding = False
         if self.use_dp and self.mapping.tp_size > 1:
+            # MoE use static heuristic to check alltoall enabled or not, however, for wide_ep, the alltoall could also be dynamically disabled when chunking is used or TRTLLM_DEEP_EP_TOKEN_LIMIT is hit.
+            is_wide_ep_alltoall_disabled = isinstance(
+                self.experts, WideEPMoE) and not self.experts.can_use_alltoall(
+                    hidden_states, all_rank_num_tokens)
+            alltoall_enabled = self.experts.enable_alltoall and not is_wide_ep_alltoall_disabled
```
hlu1, line 519: Only WideEPMoE supports alltoall. If `alltoall_enabled` is only used on line 524, you can delete line 515 to 519

tensorrt_llm/_torch/models/modeling_deepseekv3.py
ttyio: updated, thanks!

yuantailing (approved):

yuantailing:
Hi @ttyio ,
Do you have a suggested value for `TRTLLM_DEEP_EP_TOKEN_LIMIT`? Should it be equal to `max_batch_size * (1 + MTP)` or something else?

yuantailing:
Thank you! @ttyio

PR #5696, opened 2025-07-03
feat: EXAONE4.0 support

yechank-nvidia:
## Description

This is the model support newly announced by LG-AI. Ref [HF](https://github.com/huggingface/transformers/pull/39129)


jaedeok-nvidia (approved):
Thanks for your hard work, Yechan!

tensorrt_llm/_torch/models/modeling_exaone4.py
```diff
@@ -0,0 +1,322 @@
...
+    ) -> torch.Tensor | Tuple[torch.Tensor, Optional[torch.Tensor]]:
+
+        residual = hidden_states
+
+        hidden_states = self.self_attn(
+            position_ids=position_ids,
+            hidden_states=hidden_states,
+            attn_metadata=attn_metadata,
+            all_reduce_params=AllReduceParams(
+                enable_allreduce=not (self.mapping.tp_size == 1)),
```
yizhang-nv, line 232: We can remove this line and leave the params to be `None` since we've already checked this at the op level: https://github.com/NVIDIA/TensorRT-LLM/blob/ce3940953076d0f7f8544c9460676caa7ff85606/tensorrt_llm/_torch/distributed/ops.py#L456

This only needs to be set when we want to do some AR-related fusion and call the allreduce outside of attention.

tensorrt_llm/_torch/models/modeling_exaone4.py
```diff
@@ -0,0 +1,322 @@
...
+            **kwargs,
+        )
+
+        hidden_states = self.post_attention_layernorm(hidden_states)
```
yizhang-nv, line 236: We can leverage the flashinfer rms+residual add fusion by passing in the residual tensor: https://github.com/NVIDIA/TensorRT-LLM/blob/ce3940953076d0f7f8544c9460676caa7ff85606/tensorrt_llm/_torch/models/modeling_deepseekv3.py#L803 (The comment's `no fusion` means no AR + RMS Norm fusion, but still have rms + residual fusion)

yizhang-nv (approved):
Overall LGTM. Some code style and performance-related code changes can be fixed in the following pr if we want to merge this pr ASAP.

tensorrt_llm/_torch/models/modeling_exaone4.py
yechank-nvidia: Will address on the following PR. Thx for the comment.

PR #5713, opened 2025-07-03
hopper-style context MLA

zhou-yuxin:
Hopper-style Context MLA


qsang-nv (approved):

cpp/kernels/fmha_v2/setup.py
```diff
@@ -3414,43 +3424,7 @@ def get_lname_from_kname(kname: str) -> str:
 # The source code of paged context fmha kernels are not in this repo, but we have cubins for them.
 # Other kernels are for passing CI cases.
 def modify_cubin_header(cubin_header):
-    # for paged context fmha cases
```
PerkzZheng, line 3417: you might want to revert the changes here as those left cubins are needed right now (for some accuracy reansons). And Qidi will fix it later.

cpp/tensorrt_llm/kernels/contextFusedMultiHeadAttention/cubin/fmha_v2_flash_attention_bf16_64_128_S_q_paged_kv_192x128_tma_ws_sm90.cubin.cpp
PerkzZheng, line 1: you might also need to update the other left cubins as I have shared with you (the internal fmha_v2 branch). Thanks.

PerkzZheng (commented):
the changes on the fmha_v2 source codes have already been reviewed internall, so I will just skip them. 

I think you still need to update the `fmhaRunner.cpp` about how the TMA descriptors are set. We still use that path even though most of the kernels are compiled from the source codes instead of cubins.

PerkzZheng (approved):

PR #5714, opened 2025-07-03
[ci] parallelize torch unittests

omera-nv:
# [ci] parallelize torch unittests

This PR marks torch unittests that utilize high cuda memory, or that simply don't work with xdist, so we can parallelize the execution of the rest of the tests in the future.

### DIFF
--- a/tests/unittest/pytest.ini
+++ b/tests/unittest/pytest.ini
@@ -19,5 +19,5 @@ markers =
     gpu2: this test uses 2 GPUs
     gpu4: this test uses 4 GPUs
     post_merge: this test should only run in post merge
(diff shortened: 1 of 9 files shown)


tests/unittest/pytest.ini
```diff
@@ -19,3 +19,5 @@ markers =
     gpu2: this test uses 2 GPUs
     gpu4: this test uses 4 GPUs
     post_merge: this test should only run in post merge
+    high_cuda_memory: this test uses a lot of CUDA memory
```
joyang-nv, line 22: @omera-nv , can you help to put some description on the rule to set this mark? Like >10GB GPU memory, we should mark it.

joyang-nv (approved):

Issue #5249, opened 2025-06-16
Refactor the signature of AD graph transformations


nzmora-nvidia:
AD graph transformations, such as `canonicalize_graph(gm: GraphModule, ...) -> GraphModule` have a signature that implies that they return a transformed graph instance while leaving the `gm` unchanged. These  transformations actually mutate the `GraphModule` instance in-place, which can be confusing.

Change all  `transform_graph(gm: GraphModule, ...) -> GraphModule` to `transform_graph(gm: GraphModule, ...) -> None` and fix the code accordingly.

see slack conv: https://nvidia.slack.com/archives/C08T73CRBA6/p1749730434523839

PR #5719, opened 2025-07-03
[#5249][AutoDeploy] Refactor the signatures of AD graph transformations to return None

galagam:
## Description

Refactor signatures of AD graph transformations from
  gm = transformation(gm)
to
  transformation(gm)

Since the AD graph transformations modify the input GraphModule
in-place. Previous signature style was misleading.

## Test Coverage

Refactor-only change. Using existing unit tests to verify.

</details>


galagam:
Should be targeting AD branch, closing.

PR #5726, opened 2025-07-03
[ci] speedup fused moe tests

omera-nv:
# [ci] speedup fused moe tests


djns99 (approved):
The tolerances will need to be improved on these tests, but any issues would appear to be already underlying.
I do worry about the implication that a torch randn update could randomly cause these tests to fail as this would be a very confusing bug to track down

PR #5763, opened 2025-07-04
Fix docker cache mount

MartinMarciniszyn:
# Docker automation fix

This should fix the docker run automation when the `.cache` directory is not writeable by the local user.


ixlmar (approved):
LGTM

PR #5784, opened 2025-07-07
doc: add Deprecation Policy section

QiJune:
## Summary by CodeRabbit

* **Documentation**
  * Added a "Deprecation Policy" section explaining how deprecated features are communicated, including runtime warnings and a 3-month migration period before removal.


laikhtewari (commented):
One reasonable scenario: you want to deprecate / make breaking change to some of the params of the LLM API, but you won't deprecate the whole LLM API/method/class

I don't think this scenario is explicitly covered in this deprecation policy but should be

syuoni (approved):

laikhtewari (approved):
Nice addition of partial deprecation. Lgtm

Issue #5912, opened 2025-07-10
fix head_dim calculation in Qwen config for better compatibility


gkswns0531:
## Problem
The current Qwen model config calculates head_dim using simple division, which causes QKV weight shape mismatch for certain models like Qwen3-14B.

### Current Implementation
```python
head_dim = hf_config.hidden_size // hf_config.num_attention_heads
```

### Error Details
For Qwen3-14B model:
- Expected by TensorRT-LLM: (64 + 2 * 8) * 80 = 6400
- Actual model weights: (64 + 2 * 8) * 128 = 10240
- Error: `Parameter transformer.layers.0.attention.qkv.weight has invalid shape torch.Size([10240, 5120]) compared with expected shape (6400, 5120)`

### Root Cause
Starting from Qwen3-14B, the head_dim is explicitly defined in config.json as 128, not 80 (5120/64). The calculation method works for most models but fails when the model has non-standard head dimensions.

### Models Affected
- Qwen3-14B: Works (head_dim calculated correctly)
- Qwen3-32B: Fails due to explicit head_dim=128 in config

## Solution
Use getattr to check for explicit head_dim values first before calculating:
```python
head_dim = getattr(hf_config, 'head_dim', hf_config.hidden_size // hf_config.num_attention_heads)
```

This approach:
- Supports models with explicit head_dim values
- Prevents QKV weight shape mismatches

## Impact
- Fixes conversion failures for Qwen3-14B and potentially other models
- Ensures consistent behavior across different model sizes
- Improves robustness of the model loading process

## Realted PR: #5913

Shattered217:
I tried to quantitatively deploy Qwen3-32B for a day, but still encountered the error "Updated: (5120, 2560), original: (5120, 1600)". 😭

PR #5650, opened 2025-07-01
[feat] Add TensorRT-Engine Qwen3 (dense) model support

gkswns0531:
- Add Qwen3ForCausalLM mapping in models/__init__.py
- Update config.py to support Qwen3 architecture with qwen3/qwen3_moe types
- Add Qwen3 weight conversion logic in convert.py
- Implement Qwen3-specific model modifications in model.py
- Support attention_bias=False and qk_layernorm=True for Qwen3
- Enable FP16 and FP8 quantization for Qwen3

Tested with Qwen3-1.7B model successfully. (Support for the Qwen3 MoE architecture will be added in a future update)

Dear Maintainers,

I would like to kindly submit a pull request that adds support for building the Qwen3 model as a TensorRT engine.
I would be truly grateful if you could take the time to review it at your convenience.

Thank you very much for your consideration.

ref)
vllm vs tensorrt-llm latency (Qwen3-1.7b, L4 GPU, 2048 input_length, 3072 max_length, batch 1)

FP16
- trt-llm (tensorrt): TTFT: 25.7ms & TPS: 60.9
- vllm:                      TTFT: 25.0ms & TPS: 59.2

FP8
- trt-llm: TTFT: 18.3ms & TPS: 104.9
- vllm: TTFT: 20.6ms & TPS: 80.2

Related to #5673

PR #5913, opened 2025-07-10
[fix] improve head_dim calculation in Qwen config

gkswns0531:
This is follow-up PR for #5650 

## Problem

The current Qwen model config calculates head_dim using simple division, which causes QKV weight shape mismatch for certain models like Qwen3-32B.

### Current Implementation
```python
head_dim = hf_config.hidden_size // hf_config.num_attention_heads
```

### Root Cause
Starting from Qwen3-32B, the head_dim is explicitly defined in config.json as 128, not 80 (5120/64). The calculation method works for most models but fails when the model has non-standard head dimensions.

## Solution
Use getattr to check for explicit head_dim values first before calculating:
```python
head_dim = getattr(hf_config, 'head_dim', hf_config.hidden_size // hf_config.num_attention_heads)
```

This approach:
- Supports models with explicit head_dim values  
- Prevents QKV weight shape mismatches

Related Issues: #5912 

image

image

## Summary by CodeRabbit

* **Refactor**
  * Improved how configuration values are determined to support more flexible settings for model parameters.


gkswns0531:
@kaiyux
Just a gentle follow-up on the PR when you have a moment. I completely understand how busy things can get, so no pressure at all — just wanted to keep it on your radar.

Please let me know if there’s anything I can clarify or improve further.
Thanks again for your time and support!

gkswns0531:
@kaiyux 
just a gentle reminder on this PR. Please take a look when you get a chance. thank you!

amukkara (approved):

gkswns0531:
Closed this PR as the changes are covered in #6344. Thanks for the update! @amukkara

PR #5933, opened 2025-07-10
[None] [fix] improve kvcache allocation in PyTorch runtime

qixiang-99:
# fix/improve kvcache allocation in PyTorch runtime

## Background
We found significant memory under-allocation when using VSWA KvCacheManager. One reason is that PyTorch runtime uses estimated 'max_tokens' to constrain KV cache allocation, while 'max_tokens' is a strange(and potentially ambiguous) concept in Variable Sliding Window Attention KV Cache.

## TL;DR
Previously, we used `max_tokens` to limit the KV cache memory usage, but that method didn't work for VSWA's KV cache. This PR refines the KV cache estimation logic (mainly in `KVCacheCreator`) by directly enforcing a memory size limit. 
Besides, this PR also contains these minor fixes:
1. Improve estimation phase to avoid potential OOM issues for long sequence with CUDA Graph.
2. Add optional `torch.cuda.empty_cache` pytest fixture to avoid potential OOM in CI
Big shout out to @jaedeok-nvidia, @ixlmar and @Funatiq for their suggestions and reviews!

## Description

Currently, the PyTorch path includes a [step](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/py_executor_creator.py#L396) that estimates available free memory for the KV cache, calculates the corresponding 'max_tokens', and [updates](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/_util.py#L267) the kv_cache_config accordingly.

However, the existing logic significantly limits memory allocation, as it assumes a homogeneous attention model ([nvbug](https://nvbugspro.nvidia.com/bug/5380887)). 

Additionally, the 'max_tokens' concept doesn't align well with VSWA scenarios, as its definition is "_The maximum number of tokens that should be stored in the KV cache_". In VSWA case, "tokens stored in KV cache" is a bit "coarse-grained", as the in-window tokens are "fully in KV cache", while the out-of-window tokens are "partially in KV cache" because of non-homogeneous attention.

## Proposed Solution 

An alternative approach could be to **directly compute the allocated memory size during the estimation step**.

### Advantages:
1. Aligns naturally with the block distribution function `KVCacheManagerCpp.calculate_max_num_blocks` ([link](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/resource_manager.py#L642)), which already expects a memory size parameter.
3. Provides greater flexibility, particularly beneficial for future scenarios such as dynamic memory allocation in VSWA.

### Implementation Details

1. This PR introduces a new field, `mMaxGpuTotalBytes` (exposed as `max_gpu_total_bytes` in Python), in `KvCacheConfig`.  
2. Currently, `KvCacheCreator` is responsible for building and estimating KV caches. With this PR, it also calculates the estimated in-memory size of the KV cache and uses the `max_gpu_total_bytes` setting from `KvCacheConfig` to limit memory allocation.  
4. Within `KVCacheManager`, the VSWA path uses a C++ binding function (see `calculate_max_num_blocks_from_cpp`) that takes a memory size as input to determine KV blocks. For non-VSWA cases, it still uses its own Python function, but future efforts aim to unify both approaches, as discussed [here](https://nvidia.slack.com/archives/C059LSY62BT/p1748549582863269).

### ~~"Hack" to achieve this~~
~~One hack is to interpret `freeGpuMemoryFraction` as **the fraction of total GPU memory** reserved for the KV cache. This allows us to avoid adding an extra field, but we should clearly document this usage to prevent confusion. (Perhaps add a clarifying comment in the code?)~~
 
## Test Coverage
`tests/unittest/_torch/test_resource_manager.py`

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

## Summary by CodeRabbit

- **New Features**
  - Add option to cap KV-cache GPU memory via max_gpu_total_bytes (used with existing fraction limit).
  - KV-cache stats now report allocated GPU bytes; exposed in Python.
  - Automatic KV-cache sizing moved to a memory-based approach, accounting for main/draft models and VSWA.

- **Refactor**
  - Public API renamed: estimate_max_tokens → configure_kv_cache_capacity; executor creation uses the new flow.

- **Tests**
  - New fixture clears CUDA cache before tests; applied to an existing disaggregated single-GPU test.

### DIFF
--- a/cpp/include/tensorrt_llm/executor/executor.h
+++ b/cpp/include/tensorrt_llm/executor/executor.h
@@ -1006,7 +1006,8 @@ public:
         std::optional<RetentionPriority> secondaryOffloadMinPriority = std::nullopt, size_t eventBufferMaxSize = 0,
         bool enablePartialReuse = true, bool copyOnPartialReuse = true, bool useUvm = false,
         SizeType32 attentionDpEventsGatherPeriodMs = 5,
-        std::optional<tensorrt_llm::runtime::RuntimeDefaults> const& runtimeDefaults = std::nullopt);
+        std::optional<tensorrt_llm::runtime::RuntimeDefaults> const& runtimeDefaults = std::nullopt,
+        uint64_t const& maxGpuTotalBytes = 0);
 
     [[nodiscard]] bool getEnableBlockReuse() const;
     [[nodiscard]] bool getEnablePartialReuse() const;
@@ -1022,11 +1023,12 @@ public:
     [[nodiscard]] size_t getEventBufferMaxSize() const;
     [[nodiscard]] bool getUseUvm() const;
     [[nodiscard]] SizeType32 getAttentionDpEventsGatherPeriodMs() const;
+    [[nodiscard]] uint64_t getMaxGpuTotalBytes() const;
 
     void setEnableBlockReuse(bool enableBlockReuse);
     void setEnablePartialReuse(bool enablePartialReuse);
     void setCopyOnPartialReuse(bool copyOnPartialReuse);
-    void setMaxTokens(SizeType32 maxTokens);
+    void setMaxTokens(std::optional<SizeType32> maxTokens);
     void setMaxAttentionWindowVec(std::vector<SizeType32> maxAttentionWindowVec);
     void setSinkTokenLength(SizeType32 sinkTokenLength);
     void setFreeGpuMemoryFraction(FloatType freeGpuMemoryFraction);
@@ -1037,6 +1039,7 @@ public:
     void setEventBufferMaxSize(size_t eventBufferMaxSize);
     void setUseUvm(bool useUvm);
     void setAttentionDpEventsGatherPeriodMs(SizeType32 attentionDpEventsGatherPeriodMs);
+    void setMaxGpuTotalBytes(uint64_t maxGpuTotalBytes);
 
     void fillEmptyFieldsFromRuntimeDefaults(tensorrt_llm::runtime::RuntimeDefaults const& runtimeDefaults);
 
@@ -1095,6 +1098,10 @@ private:
 
     /// @brief The period in milliseconds to gather attention DP events across ranks
     SizeType32 mAttentionDpEventsGatherPeriodMs;
+    /// @brief The maximum size in bytes of GPU memory that can be allocated for the KV cache.
+    /// If both mMaxGpuTotalBytes and mFreeGpuMemoryFraction are specified, memory corresponding to the minimum will
+    /// be allocated.
+    uint64_t mMaxGpuTotalBytes;
 };
 
 /// @brief Configuration class for the runtime perf knobs
--- a/cpp/tensorrt_llm/executor/kvCacheConfig.cpp
+++ b/cpp/tensorrt_llm/executor/kvCacheConfig.cpp
@@ -28,7 +28,7 @@ KvCacheConfig::KvCacheConfig(bool enableBlockReuse, std::optional<SizeType32> co
     std::optional<FloatType> const& crossKvCacheFraction, std::optional<RetentionPriority> secondaryOffloadMinPriority,
     size_t eventBufferMaxSize, bool enablePartialReuse, bool copyOnPartialReuse, bool useUvm,
     SizeType32 attentionDpEventsGatherPeriodMs,
-    std::optional<tensorrt_llm::runtime::RuntimeDefaults> const& runtimeDefaults)
+    std::optional<tensorrt_llm::runtime::RuntimeDefaults> const& runtimeDefaults, uint64_t const& maxGpuTotalBytes)
     : mEnableBlockReuse(enableBlockReuse)
     , mHostCacheSize(hostCacheSize)
     , mOnboardBlocks(onboardBlocks)
@@ -38,6 +38,7 @@ KvCacheConfig::KvCacheConfig(bool enableBlockReuse, std::optional<SizeType32> co
     , mCopyOnPartialReuse{copyOnPartialReuse}
     , mUseUvm{useUvm}
     , mAttentionDpEventsGatherPeriodMs(attentionDpEventsGatherPeriodMs)
+    , mMaxGpuTotalBytes{maxGpuTotalBytes}
 {
     if (maxTokens)
     {
@@ -62,6 +63,10 @@ KvCacheConfig::KvCacheConfig(bool enableBlockReuse, std::optional<SizeType32> co
     if (runtimeDefaults)
     {
         fillEmptyFieldsFromRuntimeDefaults(runtimeDefaults.value());
+    }
+    if (maxGpuTotalBytes)
+    {
+        setMaxGpuTotalBytes(maxGpuTotalBytes);
     }
     TLLM_CHECK_WITH_INFO(
         mAttentionDpEventsGatherPeriodMs > 0, "Attention DP events gather period must be greater than 0");
@@ -137,6 +142,11 @@ SizeType32 KvCacheConfig::getAttentionDpEventsGatherPeriodMs() const
     return mAttentionDpEventsGatherPeriodMs;
 }
 
+uint64_t KvCacheConfig::getMaxGpuTotalBytes() const
+{
+    return mMaxGpuTotalBytes;
+}
+
 void KvCacheConfig::setEnableBlockReuse(bool enableBlockReuse)
 {
     mEnableBlockReuse = enableBlockReuse;
@@ -152,9 +162,12 @@ void KvCacheConfig::setCopyOnPartialReuse(bool copyOnPartialReuse)
     mCopyOnPartialReuse = copyOnPartialReuse;
 }
 
-void KvCacheConfig::setMaxTokens(SizeType32 maxTokens)
+void KvCacheConfig::setMaxTokens(std::optional<SizeType32> maxTokens)
 {
-    TLLM_CHECK(maxTokens > 0);
+    if (maxTokens)
+    {
+        TLLM_CHECK(maxTokens.value() > 0);
+    }
     mMaxTokens = maxTokens;
 }
 
@@ -217,6 +230,11 @@ void KvCacheConfig::setAttentionDpEventsGatherPeriodMs(SizeType32 attentionDpEve
 {
     TLLM_CHECK(attentionDpEventsGatherPeriodMs > 0);
     mAttentionDpEventsGatherPeriodMs = attentionDpEventsGatherPeriodMs;
+}
+
+void KvCacheConfig::setMaxGpuTotalBytes(uint64_t maxGpuTotalBytes)
+{
+    mMaxGpuTotalBytes = maxGpuTotalBytes;
 }
 
 void KvCacheConfig::fillEmptyFieldsFromRuntimeDefaults(tensorrt_llm::runtime::RuntimeDefaults const& runtimeDefaults)
--- a/cpp/tensorrt_llm/pybind/executor/executorConfig.cpp
+++ b/cpp/tensorrt_llm/pybind/executor/executorConfig.cpp
@@ -104,11 +104,11 @@ void initConfigBindings(pybind11::module_& m)
             self.getSinkTokenLength(), self.getFreeGpuMemoryFraction(), self.getHostCacheSize(),
             self.getOnboardBlocks(), self.getCrossKvCacheFraction(), self.getSecondaryOffloadMinPriority(),
             self.getEventBufferMaxSize(), self.getEnablePartialReuse(), self.getCopyOnPartialReuse(), self.getUseUvm(),
-            self.getAttentionDpEventsGatherPeriodMs());
+            self.getAttentionDpEventsGatherPeriodMs(), self.getMaxGpuTotalBytes());
     };
     auto kvCacheConfigSetstate = [](py::tuple const& state)
     {
-        if (state.size() != 14)
+        if (state.size() != 15)
         {
             throw std::runtime_error("Invalid state!");
         }
@@ -117,20 +117,21 @@ void initConfigBindings(pybind11::module_& m)
             state[4].cast<std::optional<float>>(), state[5].cast<std::optional<size_t>>(), state[6].cast<bool>(),
             state[7].cast<std::optional<float>>(), state[8].cast<std::optional<tle::RetentionPriority>>(),
             state[9].cast<size_t>(), state[10].cast<bool>(), state[11].cast<bool>(), state[12].cast<bool>(),
-            state[13].cast<SizeType32>());
+            state[13].cast<SizeType32>(), std::nullopt, state[14].cast<uint64_t>());
     };
     py::class_<tle::KvCacheConfig>(m, "KvCacheConfig")
         .def(py::init<bool, std::optional<SizeType32> const&, std::optional<std::vector<SizeType32>> const&,
                  std::optional<SizeType32> const&, std::optional<float> const&, std::optional<size_t> const&, bool,
                  std::optional<float> const&, std::optional<tle::RetentionPriority>, size_t const&, bool, bool, bool,
-                 SizeType32, std::optional<RuntimeDefaults> const&>(),
+                 SizeType32, std::optional<RuntimeDefaults> const&, uint64_t const&>(),
             py::arg("enable_block_reuse") = true, py::arg("max_tokens") = py::none(),
             py::arg("max_attention_window") = py::none(), py::arg("sink_token_length") = py::none(),
             py::arg("free_gpu_memory_fraction") = py::none(), py::arg("host_cache_size") = py::none(),
             py::arg("onboard_blocks") = true, py::arg("cross_kv_cache_fraction") = py::none(),
             py::arg("secondary_offload_min_priority") = py::none(), py::arg("event_buffer_max_size") = 0, py::kw_only(),
             py::arg("enable_partial_reuse") = true, py::arg("copy_on_partial_reuse") = true, py::arg("use_uvm") = false,
-            py::arg("attention_dp_events_gather_period_ms") = 5, py::arg("runtime_defaults") = py::none())
+            py::arg("attention_dp_events_gather_period_ms") = 5, py::arg("runtime_defaults") = py::none(),
+            py::arg("max_gpu_total_bytes") = 0)
         .def_property(
             "enable_block_reuse", &tle::KvCacheConfig::getEnableBlockReuse, &tle::KvCacheConfig::setEnableBlockReuse)
         .def_property("max_tokens", &tle::KvCacheConfig::getMaxTokens, &tle::KvCacheConfig::setMaxTokens)
@@ -140,6 +141,8 @@ void initConfigBindings(pybind11::module_& m)
             "sink_token_length", &tle::KvCacheConfig::getSinkTokenLength, &tle::KvCacheConfig::setSinkTokenLength)
         .def_property("free_gpu_memory_fraction", &tle::KvCacheConfig::getFreeGpuMemoryFraction,
             &tle::KvCacheConfig::setFreeGpuMemoryFraction)
+        .def_property(
+            "max_gpu_total_bytes", &tle::KvCacheConfig::getMaxGpuTotalBytes, &tle::KvCacheConfig::setMaxGpuTotalBytes)
         .def_property("host_cache_size", &tle::KvCacheConfig::getHostCacheSize, &tle::KvCacheConfig::setHostCacheSize)
         .def_property("onboard_blocks", &tle::KvCacheConfig::getOnboardBlocks, &tle::KvCacheConfig::setOnboardBlocks)
         .def_property("cross_kv_cache_fraction", &tle::KvCacheConfig::getCrossKvCacheFraction,
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -98,8 +98,7 @@ class KvCacheCreator:
             fraction = 0.9
         return fraction
 
-    def _cal_max_tokens(self, peak_memory, total_gpu_memory, fraction,
-                        alloc_kv_tokens: int) -> int:
+    def _get_kv_size_per_token(self):
         model_config = self._model_engine.model.model_config
         mapping = self._mapping
         kv_size_per_token = self._get_cache_size_per_token(
@@ -108,18 +107,25 @@ class KvCacheCreator:
             draft_model_config = self._draft_model_engine.model.model_config
             kv_size_per_token += self._get_cache_size_per_token(
                 draft_model_config, mapping)
+        return kv_size_per_token
+
+    def _cal_max_memory(self, peak_memory, total_gpu_memory, fraction,
+                        allocated_bytes: int) -> int:
+        """
+        Calculate the max KV cache capacity.
+
+        NOTE: `allocated_bytes` is the total KV-cache memory that must be pre-allocated during the estimation phase (for both the main and draft models) so the estimation run can complete successfully. When computing `available_kv_mem`, add this amount back in.
+        """
+        kv_size_per_token = self._get_kv_size_per_token()
 
         available_kv_mem = (total_gpu_memory - peak_memory +
-                            alloc_kv_tokens * kv_size_per_token) * fraction
+                            allocated_bytes) * fraction
         logger.info(
             f"Peak memory during memory usage profiling (torch + non-torch): {peak_memory / (GB):.2f} GiB, "
             f"available KV cache memory when calculating max tokens: {available_kv_mem / (GB):.2f} GiB, "
             f"fraction is set {fraction}, kv size is {kv_size_per_token}. device total memory {total_gpu_memory / (GB):.2f} GiB, "
-            f", tmp kv_mem { (alloc_kv_tokens * kv_size_per_token) / (GB):.2f} GiB"
-        )
-        max_tokens = int((available_kv_mem) // kv_size_per_token)
-        max_tokens = max(max_tokens, 0)
-        return max_tokens
+            f", tmp kv_mem { (allocated_bytes) / (GB):.2f} GiB")
+        return int(available_kv_mem)
 
     def _create_dummy_context_requests(
             self, input_seq_len: int) -> List[trtllm.Request]:
@@ -187,12 +193,17 @@ class KvCacheCreator:
         estimating_kv_cache = False
         if 'cp_type' not in self._mapping.cp_config:
             estimating_kv_cache = True
-            self._executor_config.kv_cache_config.max_tokens = self._get_token_num_for_estimation(
-            )
+            max_attention_window_from_config = self._executor_config.kv_cache_config.max_attention_window
+            max_seq_len = max(
+                max_attention_window_from_config
+            ) if max_attention_window_from_config is not None else self._executor_config.max_seq_len
+            self._executor_config.kv_cache_config.max_tokens = max(
+                self._get_token_num_for_estimation(), max_seq_len)
         return estimating_kv_cache
 
-    def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
+    def configure_kv_cache_capacity(self, py_executor: PyExecutor) -> None:
         """Perform KV cache capacity estimation.
+        NOTE: for VSWA case, we calculate and set kv cache memory instead of using max_tokens in kv_cache_config.
 
         This updates `kv_cache_config`.
         """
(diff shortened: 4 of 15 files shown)


cpp/include/tensorrt_llm/executor/executor.h
```diff
@@ -1085,6 +1088,12 @@ class KvCacheConfig
 
     /// @brief Whether to use UVM for the KV cache.
     bool mUseUvm;
+
+    /// @brief The maximum size in bytes of GPU memory that can be allocated for the KV cache.
+    /// This is only used for VSWA case for now as a alternative to mMaxTokens.
+    /// If both mMaxFreeGpuMemorySize and mFreeGpuMemoryFraction are specified, memory corresponding to the minimum will
+    /// be allocated.
+    std::optional<uint64_t> mMaxFreeGpuMemorySize;
```
jaedeok-nvidia, line 1096: I think `mMaxFreeGpuMemorySize` is a redundant field, as we can calculate it from `fraction` ([here](https://github.com/NVIDIA/TensorRT-LLM/blob/8628139625f608b708857384d5952dbcb30179f4/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp#L2145-L2178)). If it's necessary, why don't we calculate at the torch backend side and pass `freeGpuMemoryFraction` instead? That allows us to keep the fields minimal.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -94,8 +94,13 @@ def _get_free_gpu_memory_fraction(self) -> float:
...
         return fraction
 
-    def _cal_max_tokens(self, peak_memory, total_gpu_memory, fraction,
-                        alloc_kv_tokens: int) -> int:
+    def _cal_max_tokens_and_memory(self, peak_memory, total_gpu_memory,
+                                   fraction,
+                                   alloc_kv_tokens: int) -> tuple[int, int]:
+        """
+        Calculate the max KV cache capacity in max_tokens and max_free_gpu_memory_size.
+        For VSWA case, we use max_free_gpu_memory_size instead of max_tokens.
```
jaedeok-nvidia, line 102: Can I know why we need memory as well? Since, in VSWA case, we will set `max_tokens=None` before calculating the number of cache blocks, so `available_kv_mem` seems unnecessary.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -115,7 +120,7 @@ def _cal_max_tokens(self, peak_memory, total_gpu_memory, fraction,
         )
```
jaedeok-nvidia, line 120: How about not printing `tmp kv_mem` field in case of VSWA not to make a confusion? It depends on maxTokens as well as adjusted window sizes which will be done later.

tensorrt_llm/_torch/pyexecutor/_util.py
ixlmar: I'd propose to keep this simple: Make this `_calculate_available_memory(...) -> int` and let it return an upper limit on the total memory footprint (in bytes) of the draft+model KV caches.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -185,8 +190,9 @@ def try_prepare_estimation(self) -> bool:
             )
         return estimating_kv_cache
 
-    def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
+    def estimate_max_tokens_or_memory(self, py_executor: PyExecutor) -> None:
```
ixlmar, line 193: Is there a need to make VSWA a special case here? I think it should be sufficient to always `estimate_max_bytes(...) -> int`

ixlmar: This implies that `KvCacheConfig.max_tokens` should be removed completely in favor of `max_total_bytes`.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -796,6 +796,11 @@ class KvCacheConfig(BaseModel, PybindMirror):
     )
     use_uvm: bool = Field(default=False,
                           description="Whether to use UVM for the KV cache.")
+    max_free_gpu_memory_size: Optional[int] = Field(
```
ixlmar, line 799: I think `kv_cache_config.max_total_bytes` would be a more fitting name for this. Mainly, I do not think that the name should include "free".

tensorrt_llm/_torch/pyexecutor/_util.py
jaedeok-nvidia: Could you please help me understand why we need `max_tokens` or `max_total_bytes` here? Is this related to the draft model? IIUC, the process is as follows: KVCacheCreator
  1. sends a dummy request with the maximum sequence length,
  2. gathers the peak memory and cache usage, and
  3. sets the estimated max_tokens in the kv_cache_config.
But, even if we do not explicitly set `max_tokens` on the PyTorch side, the C++ KvCacheManager will calculate the maximum available tokens at [calculateMaxNumBlocks](https://github.com/NVIDIA/TensorRT-LLM/blob/6d4b045d1f9196e3ea2b0b4a422ac5fb512165ce/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp#L2210). 

Regarding `max_total_bytes`, it seems flexible, however, I am not sure if it will be useful for others since it cannot be directly used for calculating number of blocks.

One additional point is that for VSWA, the KV cache pools are currently separated per window. Therefore, if we want to estimate the KV cache size, we need to estimate the cache size per window.

qixiang-99: Hi @jaedeok-nvidia, sure, it would be great to have some clarifications to ensure we're aligned. Let me summarize my current understanding below:

Regarding the `KVCacheCreator`:
1. It sends dummy request(s) with a total token count of `min(maximum sequence length, max_num_tokens)`. My understanding is that this step determines the KV cache capacity required to successfully run **one iteration**, preventing OOM issues during subsequent dummy runs.
2. It doesn't directly estimate the KV cache required by a model, but rather estimates how much memory the model can utilize for the KV cache ([reference](https://github.com/NVIDIA/TensorRT-LLM/blob/6d4b045d1f9196e3ea2b0b4a422ac5fb512165ce/tensorrt_llm/_torch/pyexecutor/_util.py#L108-L109)).
3. I'm not entirely sure why we need dummy forward steps for KV cache estimation, but my guess is due to the difficulty in precisely calculating and controlling memory usage when running inference with PyTorch.

Regarding [`calculateMaxNumBlocks`](https://github.com/NVIDIA/TensorRT-LLM/blob/6d4b045d1f9196e3ea2b0b4a422ac5fb512165ce/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp#L2210):
1. My understanding is this function prefers memory size over token count (`max_tokens`), as indicated by [this warning](https://github.com/NVIDIA/TensorRT-LLM/blob/6d4b045d1f9196e3ea2b0b4a422ac5fb512165ce/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp#L2220-L2227). With the memory size, the C++ KvCacheManager will first split the memory size across different windows, and "_calculate the maximum available tokens (for each window size) at [calculateMaxNumBlocks](https://github.com/NVIDIA/TensorRT-LLM/blob/6d4b045d1f9196e3ea2b0b4a422ac5fb512165ce/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp#L2210)_"

Funatiq: Regarding `free_gpu_memory_fraction`, `max_tokens`, `max_total_bytes`, they all make sense to me from a user perspective.
* `free_gpu_memory_fraction` - user doesn't know or care how much memory in bytes are available, but wants to have some control how much memory should be left free.
* `max_tokens` - user knows/estimates how many requests/tokens are processed at a time and wants to reserve specific token budget. To me this only makes sense if it considers the full request length, so either full attention (non-SWA) or VSWA with full attention layers (also see the warning [here](https://github.com/NVIDIA/TensorRT-LLM/blob/6d4b045d1f9196e3ea2b0b4a422ac5fb512165ce/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp#L2220)). We could simply disallow this option if no full attention layer is used.
* `max_total_bytes` - user wants to reserve a specific amount of memory, but doesn't care how many requests can run.

Therefore I don't think we should remove one in favor of another. What we could do, however, is restricting the `KvCacheConfig` so that we don't have to deal with all possible combinations of the options. E.g. restrict to either use `max_tokens` or `max_total_bytes`, or have a function that calculates the resulting bytes from `max_tokens` and take the minimum with `max_total_bytes` (and `free_gpu_memory_fraction`). Then we could always call `calculateMaxNumBlocks` with a single bytes value. WDYT?

jaedeok-nvidia: Thank @qixiang-99 for kind explanation and thank @Funatiq / @ixlmar for discussions.

I agree both @Funatiq / @ixlmar. 

Currently, KvCacheManager splits cache pool per window, so that in practice VSWA model uses inefficient cache pool. For instance, for VSWA model having windows [1k, 100k], then KvCacheManager allocates pools proportional to window size. If we have 1) ISL 1k x 100 samples vs 2) ISL 100k x 1 samples. Number of ISL tokens are the same, however, case 1) will consume much more memory. And if cache pool for window=1k is no enough, then batch size will be limited because of unused pool of window=100k. So, unifying cache pool will be necessary. This is out-of-topic here, but someday we need to do.

Considering these, for now, I hoped we can have a minimal change (by minimal effort) at this moment.

And, after the above discussions, each parameter seems meaningful, and the current implementation looks good to me (I didn't want to against).

cpp/include/tensorrt_llm/executor/executor.h
```diff
@@ -1030,6 +1032,7 @@ class KvCacheConfig
     void setSecondaryOffloadMinPriority(std::optional<RetentionPriority> secondaryOffloadMinPriority);
     void setEventBufferMaxSize(size_t eventBufferMaxSize);
     void setUseUvm(bool useUvm);
+    void setMaxFreeGpuMemorySize(uint64_t maxFreeGpuMemorySize);
```
jaedeok-nvidia, line 1035: Should we allow `std::optional<unin64_t> maxFreeGpuMemorySize`? none means unset the limit.

tensorrt_llm/llmapi/llm_args.py
jaedeok-nvidia: Accordingly, it may need to be changed C++ part - `maxFreeGpuMemorySize` to `max_total_bytes`.

cpp/include/tensorrt_llm/executor/executor.h
jaedeok-nvidia: Thanks! As we discussed offline and below thread, it looks ok for me now.

tensorrt_llm/_torch/pyexecutor/_util.py
ixlmar: @qixiang-99 Understood. To reduce technical debt given the direction suggested in https://github.com/NVIDIA/TensorRT-LLM/pull/5933#discussion_r2210770747, I propose to make the following hopefully quick changes to `KvCacheCreator`:

* Rename `estimate_max_tokens_or_memory` to `configure_kv_cache_capacity`.

* Change `_cal_max_tokens_and_memory` to `_cal_max_memory`. Handle everything except `max_tokens` there as done currently.

* In `configure_kv_cache_capacity`:

  * If user provided `max_tokens`,  use `_get_cache_size_per_token` to convert to bytes and update max. memory obtained via `_cal_max_memory` accordingly.
  * If VSWA, raise warning or error as done now. Since we are validating user-inputs we should probably `raise` a meaningful error rather than `assert`.

  * Only pass max. memory to C++ bindings in all cases.

jaedeok-nvidia (approved):
Thank @qixiang-99 for your contribution to VSWA fixes. Hope this PR fix our underutilization issue in VSWA models. And, let's consider how to improve the efficient further.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -255,16 +261,29 @@ def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
...
             peak_memory, total_gpu_memory, fraction,
             kv_stats.max_num_blocks * kv_stats.tokens_per_block)
 
         if self._max_kv_tokens_in is not None:
             kv_cache_max_tokens = min(kv_cache_max_tokens,
                                       self._max_kv_tokens_in)
 
-        logger.info(f"Estimated max tokens in KV cache : {kv_cache_max_tokens}")
-        executor_config.kv_cache_config.max_tokens = kv_cache_max_tokens
+        if executor_config.kv_cache_config.max_attention_window is not None:
```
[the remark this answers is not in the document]
qixiang-99: Updated, thanks!

cpp/include/tensorrt_llm/executor/executor.h
qixiang-99: Thanks for the suggestion. I've removed the optional and rename the variable as `maxGpuTotalBytes`

tensorrt_llm/_torch/pyexecutor/_util.py
qixiang-99: Thanks for the discussion, team! I've done these changes thanks to @ixlmar 's detailed suggestion:
1. Rename `estimate_max_tokens_or_memory` to `configure_kv_cache_capacity` ([here](https://github.com/NVIDIA/TensorRT-LLM/pull/5933/files#diff-5f0e349b724e8ad501e0ae859c992801e631217a8d52acfa40e24a8fe95bbbcbR248))
2. Change `_cal_max_tokens_and_memory` to `_cal_max_memory` and it will only return memory size in bytes.
3. Added function to handle the conversion from user-defined `max_tokens` to memory in bytes, and will raise Warning if it is VSWA model ([here](https://github.com/NVIDIA/TensorRT-LLM/pull/5933/files#diff-5f0e349b724e8ad501e0ae859c992801e631217a8d52acfa40e24a8fe95bbbcbR345-R353))

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -187,8 +245,9 @@ def try_prepare_estimation(self) -> bool:
             )
         return estimating_kv_cache
 
-    def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
+    def configure_kv_cache_capacity(self, py_executor: PyExecutor) -> None:
```
jaedeok-nvidia, line 248: Love this new function name. It makes more sense than before, as this function sets the cache capacity for the class.

tensorrt_llm/_torch/pyexecutor/_util.py
jaedeok-nvidia: Thanks for catching this point. 

One more concern was, a dummy request could be too short to fill the sliding window size. When testing EXAONE4 model, I found dummy req has a seq of length 2080, which is much shorther than the window size 4096. That means estimating via `kv_stats` can be incorrect. 

I feel we may need to redesign related parts 😂 . Let's discuss it offline later.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -254,19 +313,72 @@ def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
...
+        peak_memory += extra_memory_for_attention_metadata
+
         kv_stats = py_executor.resource_manager.resource_managers.get(
             ResourceManagerType.KV_CACHE_MANAGER).get_kv_cache_stats()
 
-        kv_cache_max_tokens = self._cal_max_tokens(
+        # calculate max memory from peak memory and free gpu memory fraction
+        kv_cache_max_memory = self._cal_max_memory(
             peak_memory, total_gpu_memory, fraction,
             kv_stats.max_num_blocks * kv_stats.tokens_per_block)
```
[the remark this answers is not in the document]
qixiang-99: This is a great observation! After reviewing `kv_stats.max_num_blocks` in the VSWA context, it does indeed sum `max_num_blocks` for each window size manager. Fundamentally, `_cal_max_memory` needs the total allocated memory for the KV cache during estimation, which counts toward available KV memory in a real run. We already calculate this allocated KV memory size in existing code (see [the relevant code section](https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp#L1733-L1759)), so I've exposed this value through `kv_stat`. This approach should be able to addresses the issue from "first principles" without having to manage a more complex calculation : )

cpp/tensorrt_llm/pybind/executor/executorConfig.cpp
```diff
@@ -138,6 +138,8 @@ void initConfigBindings(pybind11::module_& m)
             "sink_token_length", &tle::KvCacheConfig::getSinkTokenLength, &tle::KvCacheConfig::setSinkTokenLength)
         .def_property("free_gpu_memory_fraction", &tle::KvCacheConfig::getFreeGpuMemoryFraction,
             &tle::KvCacheConfig::setFreeGpuMemoryFraction)
+        .def_property(
```
ixlmar, line 141: The Nanobind bindings need to be updated accordingly.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -97,8 +97,60 @@ def _get_free_gpu_memory_fraction(self) -> float:
...
+          and the tensor's footprint appears modest.
+
+        • Real inference: `max_blocks_per_seq` can increase to
+          `max_seq_len / tokens_per_block`.  For long-context models this is
+          orders of magnitude larger, so the tensor consumes significantly more
+          GPU memory.
+
+        • CUDA graphs: when graph capture is enabled the full
+          `kv_cache_block_offsets` tensor must be pre-allocated,
+          making the extra memory grow linearly with the number of graphs.
```
ixlmar, line 121: Can the memory for this tensor be shared between CUDA graphs?

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -255,19 +315,80 @@ def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
...
+                logger.warning(
+                    "max_tokens should not be set for VSWA case as it is ambiguous concept for VSWA."
+                )
+            # calculate max memory from max_tokens
+            kv_cache_max_memory_from_max_tokens = self._max_kv_tokens_in * self._get_kv_size_per_token(
+            )
+            kv_cache_max_memory = min(kv_cache_max_memory,
+                                      kv_cache_max_memory_from_max_tokens)
+            logger.info(
+                f"max_tokens is provided, max_memory is set to {kv_cache_max_memory / (GB):.2f} GiB"
```
ixlmar, line 365: Perhaps include `self._max_kv_tokens_in` in the log message.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -255,19 +315,80 @@ def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
...
+        # ---------------------------handle max_tokens---------------------------------
+
+        # ---------------------------handle max_gpu_total_bytes---------------------------------
+        # if user provided max_gpu_total_bytes, set max memory from max_gpu_total_bytes
+        if executor_config.kv_cache_config.max_gpu_total_bytes > 0:
+            kv_cache_max_memory = min(
+                kv_cache_max_memory,
+                executor_config.kv_cache_config.max_gpu_total_bytes)
+            logger.info(
+                f"max_gpu_total_bytes is provided, max_memory is set to {kv_cache_max_memory / (GB):.2f} GiB"
```
ixlmar, line 383: Print `executor_config.kv_cache_config.max_gpu_total_bytes` as well.

tensorrt_llm/_torch/pyexecutor/_util.py
qixiang-99: Updated, thanks!

tensorrt_llm/_torch/pyexecutor/_util.py
qixiang-99: This is a good point, and I did look into it. As I'm not a CUDA Graph expert, please take this with a grain of salt: my understanding is that we really only need one `kv_cache_block_offsets` that's large enough for all block offset tracking; **from that perspective, it could likely be shared.**

However, implementing this would be non-trivial because the current hierarchy is: `PyTorchModelEngine` owning multiple `DecodingCUDAGraphRunner` instances for different batch sizes, each with its own `AttentionMetadata` (where `kv_cache_block_offsets` resides). 

I recommend addressing this in a separate pull request to keep this PR focused, WDYT?

ixlmar: Yes, makes sense.

ixlmar (commented):
Reviewed latest changes

ixlmar (approved):
LGTM

Note that https://github.com/NVIDIA/TensorRT-LLM/blob/e9be1e9de4836f86e6fa36080270ff236b47a06e/tensorrt_llm/_torch/pyexecutor/_util.py#L244-L247 might currently be over-dimensioning the initial KV cache because it does not account for VSWA. OTOH, if it can be left as it is, this reduces code complexity.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -188,8 +247,9 @@ def try_prepare_estimation(self) -> bool:
             )
         return estimating_kv_cache
 
-    def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
+    def configure_kv_cache_capacity(self, py_executor: PyExecutor) -> None:
```
QiJune, line 250: @HuiGao-NV  Could you please help review this change? Thanks

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -97,8 +97,60 @@ def _get_free_gpu_memory_fraction(self) -> float:
...
+            max_batch_size = 1
+        max_attention_window_from_config = self._executor_config.kv_cache_config.max_attention_window
+        max_window_size = max(
+            max_attention_window_from_config
+        ) if max_attention_window_from_config is not None else self._executor_config.max_seq_len
+        tokens_per_block = self._executor_config.tokens_per_block
+        num_pools = kv_cache_manager.num_pools
+
+        # calculate the max_blocks_per_seq in real inference phase
+        real_phase_max_blocks_per_seq = int(
```
HuiGao-NV, line 138: I suggest to make the calculation of max_blocks_per_seq be moved into modelEngine or inside of TrtllmAttentionMetadata. This shall be one of their attribute and responsibility.
Then, we can get the value at multiple places such as here and warmup stage to pre allocate all  max_blocks_per_seq.

cpp/tensorrt_llm/executor/kvCacheConfig.cpp
```diff
@@ -36,6 +36,7 @@ KvCacheConfig::KvCacheConfig(bool enableBlockReuse, std::optional<SizeType32> co
     , mEnablePartialReuse{enablePartialReuse}
     , mCopyOnPartialReuse{copyOnPartialReuse}
     , mUseUvm{useUvm}
+    , mMaxGpuTotalBytes{maxGpuTotalBytes}
```
HuiGao-NV, line 39: I didn't see how this mMaxGpuTotalBytes impact on kvCacheManager or any logic else. Why do we need it?

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -255,19 +315,80 @@ def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
...
+            executor_config.kv_cache_config.max_tokens = int(
+                kv_cache_max_memory // self._get_kv_size_per_token())
+        # ---------------------------handle max_tokens---------------------------------
+
+        # ---------------------------handle max_gpu_total_bytes---------------------------------
+        # if user provided max_gpu_total_bytes, set max memory from max_gpu_total_bytes
+        if executor_config.kv_cache_config.max_gpu_total_bytes > 0:
+            kv_cache_max_memory = min(
+                kv_cache_max_memory,
+                executor_config.kv_cache_config.max_gpu_total_bytes)
```
HuiGao-NV, line 381: I see what max_gpu_total_bytes means and what it does. I suggest to change its name to max_kv_bytes.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -97,8 +97,60 @@ def _get_free_gpu_memory_fraction(self) -> float:
...
+        • Estimation phase: we run a dry-run with requests of length
+          `max_num_tokens` (e.g. 8192).  Consequently, `max_blocks_per_seq` is small
+          and the tensor's footprint appears modest.
+
+        • Real inference: `max_blocks_per_seq` can increase to
+          `max_seq_len / tokens_per_block`.  For long-context models this is
+          orders of magnitude larger, so the tensor consumes significantly more
+          GPU memory.
+
+        • CUDA graphs: when graph capture is enabled the full
```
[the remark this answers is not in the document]
qixiang-99: Hi @HuiGao-NV, thanks for reviewing. Let me put things together here to make sure we are all on the same page:
1. Regarding the memory usage of `kv_cache_block_offsets`(shape=`[num_pools, max_num_sequences, 2, max_blocks_per_seq]`): `PyTorchModelEngine` owning multiple `DecodingCUDAGraphRunner` instances for different batch sizes, each with its own `AttentionMetadata` (where `kv_cache_block_offsets` resides).
2. When I refer to "Kv cache estimation phase", I mean these [codes](https://github.com/qixiang-99/TensorRT-LLM/blob/e61c6862e6f48c179ec24fc5b0f4cf38339ecb15/tensorrt_llm/_torch/pyexecutor/_util.py#L264-L317). And yes, you are right, the memory should be reflected in the Peak memory usage stats. However, **in Kv cache estimation phase**, we only set a small KV Cache(set max_tokens [here](https://github.com/qixiang-99/TensorRT-LLM/blob/e61c6862e6f48c179ec24fc5b0f4cf38339ecb15/tensorrt_llm/_torch/pyexecutor/_util.py#L237-L248), e.g. usually 8192) to run a small dummy request(e.g. one req with input_seq_len=8192). So the `max_blocks_per_seq` is small(e.g. `8192/32(tokens_per_block)=256`). So **the Peak memory usage only reflects the `kv_cache_block_offsets`, for exmaple, with shape=`[num_pools, max_num_sequences, 2, 256]`.**
3. When I refer to "Real inference", I mean the real py_executor creation in these [codes](https://github.com/qixiang-99/TensorRT-LLM/blob/636265a35b5aa23a2e3e51a8f8188e90ad196c77/tensorrt_llm/_torch/pyexecutor/py_executor_creator.py#L399-L436). In this phase, we will re-create KvCacheManager for the real inference([here](https://github.com/qixiang-99/TensorRT-LLM/blob/636265a35b5aa23a2e3e51a8f8188e90ad196c77/tensorrt_llm/_torch/pyexecutor/py_executor_creator.py#L408)). In real inference, the `max_blocks_per_seq` could be large for long sequence model--for example, for max_seq_len=128k, the `max_blocks_per_seq` could be `128k/32(tokens_per_block)=4096`. So `kv_cache_block_offsets.shape=[num_pools, max_num_sequences, 2, 4096]`. 
4. So this function is used to **calculate the difference in size between `[num_pools, max_num_sequences, 2, 256]`(estimation phase) and `[num_pools, max_num_sequences, 2, 4096]`** for each `kv_cache_block_offsets` , and the final result will **multiply the number of graphs**.
5. **In most cases we wouldn't hit any issue with this difference**, because we usually have some free memory and the diff wouldn't be too large.  I found this issue because in this [nvbug](https://nvbugspro.nvidia.com/bug/5380887), the reporter used 1310972 as the max_seq_len, which make the real run has 162x larger `kv_cache_block_offsets` than the estimation phase's one, and in total 162x34(num of graphs)=**5505x** larger.

TBH I am not an expert about the CUDA Graph and ModelEngine part, so please let me know if above explanation is missing something or incorrect. Thank you!

HuiGao-NV (approved):
LGTM

HuiGao-NV (commented):
After discussion with members, the problem may be max_seq_len is overwritten by max_tokens if max_tokens is smaller.
If so, the better place is in function try_prepare_estimation in file _util.py, change  block
```
self._executor_config.kv_cache_config.max_tokens = self._get_token_num_for_estimation(
            ) 
```
to 
```
max_attention_window_from_config = self._executor_config.kv_cache_config.max_attention_window
        max_window_size = max(
            max_attention_window_from_config
        ) if max_attention_window_from_config is not None else self._executor_config.max_seq_len
self._executor_config.kv_cache_config.max_tokens = self._executor_config.kv_cache_config.max_attention_window
```

Then, KV cache will take correct max_seq_len during memory estimation.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -255,19 +315,80 @@ def estimate_max_tokens(self, py_executor: PyExecutor) -> None:
...
+        allocated_bytes = kv_stats.allocated_bytes + (
+            kv_stats_draft.allocated_bytes if kv_stats_draft is not None else 0)
+
+        # calculate max memory from peak memory and free gpu memory fraction
+        kv_cache_max_memory = self._cal_max_memory(peak_memory,
+                                                   total_gpu_memory, fraction,
+                                                   allocated_bytes)
+
+        max_attention_window = executor_config.kv_cache_config.max_attention_window
+        is_vswa = max_attention_window and len(max_attention_window) > 1
```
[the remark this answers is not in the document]
qixiang-99: Yes, [128, 128]  wouldn't trigger error here. I have confusion here, what does [128, 128] stand for?

nv-guomingz (approved):
LGTM for LLM API part

QiJune (approved):

tests/unittest/_torch/test_resource_manager.py
```diff
@@ -527,6 +534,13 @@ def test_adjust_window_sizes_for_vswa(self):
                     f"Memory bytes: {memory_bytes}\n"
                     f"Actual: {adjusted}\n"
                     f"Expected: {expected_window_sizes}")
+                self.assertEqual(
+                    adjusted_max_attention_window_vec,
+                    expected_max_attention_window_vec,
+                    f"Test case '{description}' failed.\n"
+                    f"Memory bytes: {memory_bytes}\n"
```
eopXD, line 541: We also need test coverage on memory fraction calculated through `calculate_max_num_blocks_from_cpp`

jaedeok-nvidia: Thanks @eopXD for the suggestion. Improving test coverage is important. However, if it's not super critical, let's add the test in a separate PR, because this PR has already been delayed for too long due to CI infra failures, while it's necessary to fix the vswa memory underutilization issue. If we add a new commit now, it might take a few more days 😢

QiJune: There is a duplicated PR that address memory fraction part (https://github.com/NVIDIA/TensorRT-LLM/pull/6807)  so we will add the test at there.

tensorrt_llm/_torch/pyexecutor/_util.py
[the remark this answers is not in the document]
qixiang-99: Previously I assumed the user will provide `max_attention_window` , now I see your point here, thanks! Will update to:
```python
is_vswa = max_attention_window and len(set(max_attention_window)) > 1
```
Sounds good?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1036,6 +1036,37 @@ def validate_max_gpu_total_bytes(cls, v: int):
...
+            if not isinstance(i, int):
+                raise ValueError(
+                    "kv_cache_config.max_attention_window must contain only integers"
+                )
+            if i <= 0:
+                raise ValueError(
+                    "kv_cache_config.max_attention_window values must be positive"
+                )
+
+        # Must not be a redundant repetition of a shorter pattern
```
jaedeok-nvidia, line 1061: Should we prevent this case? Of course, most models do not have such case, but I'm not sure we have to prevent it from the argument.

jaedeok-nvidia:
@qixiang-99 thanks for addressing the comments and test to improve test coverage. I left a question, but overall it looks good to me. Thanks!

tensorrt_llm/llmapi/llm_args.py
qixiang-99: Thanks for the thoughtful review. I originally added this to simplify inputs by enforcing the shortest pattern. After reconsidering, I agree with you—the validation is too strict because:

- The description doesn’t require user to do it.
- Some HF model configs don’t use the shortest pattern.

I’ll remove this validation. Thanks for flagging it!

jaedeok-nvidia: Thanks for the quick update 😄

jaedeok-nvidia:
Thanks @qixiang-99 for addressing the last-minute reviews as well.  @eopXD could you help with another round of review?

eopXD (approved):
LGTM, thank you for the test coverages!

PR #5947, opened 2025-07-11
[nvbugs/5333742] fix MTP illegal memory access in cuda graph warmup

lfr-0531:
# [nvbugs/5333742] fix MTP illegal memory access in cuda graph warmup

## Description

We found an illegal memory access issue in the CUDA graph warmup. The root cause:

When enabling the overlap scheduler, the generation requests should use device tensors from the last iteration to prepare inputs. But in warmup, those dummy generation requests don't have the previous `new_tensors_device`. So the tensor `previous_pos_id_offsets_cuda` and `previous_kv_lens_offsets_cuda` won't be set to zeros https://github.com/NVIDIA/TensorRT-LLM/blob/fbb4cc7379733163f0d0140d11069f8a21a99ae6/tensorrt_llm/_torch/pyexecutor/model_engine.py#L1441 Then in `_preprocess_inputs` the `kv_lens_cuda` will be set to a wrong value https://github.com/NVIDIA/TensorRT-LLM/blob/fbb4cc7379733163f0d0140d11069f8a21a99ae6/tensorrt_llm/_torch/pyexecutor/model_engine.py#L1172 Then when writing key/values to the KV cache in `applyMLARopeAndAssignQKVKernelGeneration` kernel, there will be illegal memory access.

## Test Coverage
A new test to L0:
```
accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_bfloat16_4gpus[tp4-mtp_nextn=2-attention_dp=False-cuda_graph=True-overlap_scheduler=True-torch_compile=False]
```

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -1408,6 +1408,16 @@ class PyTorchModelEngine(ModelEngine):
                     self.previous_batch_indices_cuda[:previous_batchs]],
                 non_blocking=True)
 
+        if (not self._disable_overlap_scheduler
+                and next_draft_tokens_device is None
+                and len(extend_requests) > 0):
+            # During warmup, for those generation requests, we don't have previous tensors,
+            # so we need to set the previous_pos_id_offsets and previous_kv_lens_offsets to zeros
+            # to skip the value changes in _preprocess_inputs. Otherwise, there will be illegal memory access
+            # when writing key/values to the KV cache.
+            self.previous_pos_id_offsets_cuda *= 0
+            self.previous_kv_lens_offsets_cuda *= 0
+
         position_ids = torch.tensor(position_ids,
                                     dtype=torch.int,
                                     pin_memory=True)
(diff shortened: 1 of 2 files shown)


litaotju (approved):

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1408,6 +1408,16 @@ def _prepare_tp_inputs(
...
                 non_blocking=True)
 
+        if (not self._disable_overlap_scheduler
+                and next_draft_tokens_device is None
+                and len(extend_requests) > 0):
+            # During warmup, for those generation requests, we don't have previous tensors,
+            # so we need to set the previous_pos_id_offsets and previous_kv_lens_offsets to zeros
+            # to skip the value changes in _preprocess_inputs. Otherwise, there will be illegal memory access
+            # when writing key/values to the KV cache.
+            self.previous_pos_id_offsets_cuda *= 0
```
juney-nvidia, line 1418: Is it possible to catch such corner case based on test cases?

juney-nvidia: To be noted, this doesn't need to block the merge of this PR since we want to unblock 0.21 release ASAP.
While it can be a good follow-up discussion and potential action to reduce the future cost.

litaotju: We have one of the bug reported on CI. I told @lfr-0531 to unwaive the test cases

lfr-0531: I added the test reported in nvbugs/5304040.

achartier (approved):

mikeiovine (approved):

schetlur-nv:
@lfr-0531 can you please add an Mr description and Pr title please?

PR #5975, opened 2025-07-11
[fix] Release slots with spec decode + disagg

Tabrizian:
# [fix] Release slots with spec decode + disagg

For disagg with eagle the resources where not freed running into no free slot errors.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -869,8 +869,8 @@ class PyExecutor:
 
                 self._pad_attention_dp_dummy_request()
 
-                if self.draft_model_engine is not None or is_ngram or hasattr(
-                        self, 'drafter') and self.drafter is not None:
+                if self.draft_model_engine is not None or is_ngram or (hasattr(
+                        self, 'drafter') and self.drafter is not None):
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -869,8 +869,8 @@ def _executor_loop(self):
 
                 self._pad_attention_dp_dummy_request()
 
-                if self.draft_model_engine is not None or is_ngram or hasattr(
-                        self, 'drafter') and self.drafter is not None:
+                if self.draft_model_engine is not None or is_ngram or (hasattr(
```
mikeiovine, line 872: The `self.drafter` attribute is always defined for `PyExecutor`: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/py_executor.py#L198

So the `hasattr` part is not necessary

mikeiovine (approved):

tensorrt_llm/_torch/pyexecutor/py_executor.py
Tabrizian: It looks like on 0.21 branch drafter doesn't exist, I can probably get rid of it here.

pcastonguay (approved):

Issue #5210, opened 2025-06-13
Broken checkpoint converter for whisper


SerhiiArtemuk:
Hi. I'm trying to recreate this demo: https://github.com/NVIDIA/TensorRT-LLM/tree/e88da961c51b300e6b9c931476428a2de908830d/examples/whisper. 
I use the large-3 model. However, when I try to build the decoder, I get this error:
```python
Traceback (most recent call last):
  File "/usr/local/bin/trtllm-build", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/commands/build.py", line 621, in main
    parallel_build(model_config, ckpt_dir, build_config, args.output_dir,
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/commands/build.py", line 419, in parallel_build
    passed = build_and_save(rank, rank % workers, ckpt_dir,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/commands/build.py", line 384, in build_and_save
    engine = build_model(build_config,
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/commands/build.py", line 361, in build_model
    model = model_cls.from_checkpoint(ckpt_dir, config=rank_config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/models/modeling_utils.py", line 661, in from_checkpoint
    model.load(weights, from_pruned=is_checkpoint_pruned)
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/models/modeling_utils.py", line 684, in load
    raise RuntimeError(
[...]

PR #5984, opened 2025-07-13
feat: Simplify and Improve Whisper Example

MahmoudAshraf97:
Since v0.27.0 https://github.com/NVIDIA/TensorRT-Model-Optimizer/pull/125, ModelOpt started supporting the conversion and quantization of whisper huggingface checkpoints to TRT-LLM checkpoints, This PR removes the custom conversion checkpoints that existed in TRTLLM before and uses the unified conversion path provided by ModelOpt, it's also simpler since the users are no longer required to download the models manually and can use any HF whisper checkpoints including finetuned ones

It also simplifies the whisper example by removing duplicated functions that were originally copied from the original whisper implementation and importing the original ones instead, the whisper package was already in the requirements so no new requirements were added cc. @yuekaizhang


MahmoudAshraf97:
This PR fixes #5210 can someone review it?

MahmoudAshraf97:
@juney-nvidia I want to continue this PR, why was it closed?

PR #5986, opened 2025-07-13
perf: Enable 128x256 tile shapes for FP4 MOE CUTLASS backend

djns99:
Adds support for MOE tile shape 128x256x256 when using NVFP4 on Blackwell.


hlu1 (approved):
LGTM

Issue #5927, opened 2025-07-10
[PERF] Qwen-VLs. Speedup by avoiding MPI overhead


vadiklyutiy:
### Problem 
I investigated performance of Qwen-VL models family. Here I will provide data for Qwen2.5-VL-3B, but the issue is actual for all Qwen2-VL and Qwen2.5-VL model. The workload is really high throughput with 1 image per request. 

I found that performance is very low compared to vllm and GPU utilizations is around 10%. 

The problem that default build is with `ENABLE_MULTI_DEVICE=1`. In `_broadcast_new_requests` in worker process we call `mpi_comm().bcast(obj, root)`. We have only one GPU and one worker. Nothing to broadcast but there still exist sufficient overhead to pickle and unpickle tensor with image(this includes useless D2H and H2D memory transfers).
Code is here 
https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_utils.py#L495-L496
```python
def mpi_broadcast(obj, root=0):
    return mpi_comm().bcast(obj, root) if ENABLE_MULTI_DEVICE else obj
```
So, we avoid `bcast` call if `ENABLE_MULTI_DEVICE=0` what is not the default. 

nsys screenshot is below

<img width="1133" height="223" alt="Image" src="https://github.com/user-attachments/assets/3665f343-659d-4cac-9081-ae584cb836e3" />

I made a experiment and build with `--extra-cmake-vars "ENABLE_MULTI_DEVICE=0" `. Got around 15% improvement in throughput. `_broadcast_new_requests` fully disappeared from profile. 

### Solution 
I don't think that another build `"ENABLE_MULTI_DEVICE=0"` is a good solution. 

I propose in MPI interfaces in `tensorrt_llm/_utils.py` in every place where we check  `if ENABLE_MULTI_DEVICE` also add `tmp.Get_size() > 1` .

### Other 
Run on 1xH100

```
trtllm-serve serve "Qwen/Qwen2.5-VL-3B-Instruct"  --backend pytorch --extra_llm_api_options ./extra-llm-api-config.yml  --max_seq_len 8192 --log_level error
```

appeared with any workload. I used https://github.com/CentML/flexible-inference-bench
```
[...]

PR #6004, opened 2025-07-14
[PERF] Move calculation Qwen2-VL's rotary_cos_sin to LLM worker process

vadiklyutiy:
## Description

Qwen2-VL `rotary_cos_sin` tensor calculated on main process and sent to worker via IPC. This tensor size is 64MB and transferring it via socket was a bottleneck.  

Move this calculations to worker(LLM) process. Also moved this calculation to GPU. 

## Test Coverage
Ran Qwen2.5-VL with several thousands of random request with 1 image per request. 

Also tested with following debug feature: I left both transfer of `rotary_cos_sin` and calculation on worker and check that tensors are `torch.allclose`

## Performance results
I tested Qwen2.5-3B-VL with high throughput workload. 

For test used https://github.com/CentML/flexible-inference-bench 
```
fib benchmark -rps 20 --input-token-distribution uniform 250 300     --output-token-distribution uniform 150 250 --num-of-imgs-per-req 1     --img-ratios-per-req 512x512  -n 1000 --base-url http://localhost:8000     --endpoint v1/chat/completions --backend openai-chat
```
this command send 1000 requests(20 reqs/s) with one 512x512 image per request. 

trt-llm  built with `--extra-cmake-vars "ENABLE_MULTI_DEVICE=0"` to avoid impact of #5927
```
trtllm-serve serve "Qwen/Qwen2.5-VL-3B-Instruct"  --backend pytorch --extra_llm_api_options ./extra-llm-api-config.yml  --max_seq_len 8192 --log_level error --host 0.0.0.0
```

Number of processed requests per second
**Before: 2.5 reqs/sec
After: 12.3 reqs/sec
Speedup around 5x times.** 

## Summary by CodeRabbit

* **New Features**
  * Enhanced model profiling capabilities during input processing.
  * Rotary embedding computations are now handled dynamically during model execution, improving flexibility for multimodal inputs.

* **Refactor**
  * Rotary embedding initialization and processing have been streamlined for better performance and maintainability.


tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -567,6 +586,43 @@ def _parse_and_concat_mrope_config(
 
         return batched_mrope_config
 
+    def add_rotary_cos_sin(self, multimodal_params: List[MultimodalParams]):
+        if self.cos_ori.device != self.device:
+            self.cos_ori = self.cos_ori.to(self.device)
+            self.sin_ori = self.sin_ori.to(self.device)
```
vadiklyutiy, line 592: By some reason, during `Qwen2VLModelBase.__init__` `self.device == 'cpu'`. Due to that implemented this workaround. 
If you have better solution would be glad to change.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -513,6 +515,23 @@ def __init__(
         logger.info(f"{self.dtype=} {self.model_dtype=}")
         self.post_config()
         self.is_loaded = True
+        self.init_rotary_cos_sin_ori()
+
+    def init_rotary_cos_sin_ori(self):
```
vadiklyutiy, line 520: This is from `Qwen2VLInputProcessorBase._post_init_`

yechank-nvidia (commented):
Thx for the work! Ultimately, we should move the entire `get_mrope_config` to the model forward side, but seem like we are having lots of benefits from moving portion of it. Comments are below.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
vadiklyutiy: Unfortunately we can't. When `init_rotary_cos_sin_ori` is executed `self.device == 'cpu'`. 
For me it is a bit weird, but likely there should some reason for it.

yechank-nvidia: Can you move `init_rotary_cos_sin_ori` under `load_weights`?

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -42,7 +43,6 @@ def __init__(self,
             trust_remote_code=trust_remote_code)
 
         self.tllm_multimodal_token_id = self.model_config.vocab_size + 1
-        self._post_init_()
```
yechank-nvidia, line 45: Can you also delete `-post_init()` from InputProcessor?

yechank-nvidia (approved):
Other than my last comment, LGTM. Thx for the work.

brb-nv (approved):
LGTM.

PR #6032, opened 2025-07-14
[fix] Release slots with spec decode + disagg (#5975)

Tabrizian:
# [fix] Release slots with spec decode + disagg

## Summary by CodeRabbit

* **New Features**
  * Added support for new model variants "Llama-3.1-8B-Instruct" and "EAGLE3-LLaMA3.1-Instruct-8B" in disaggregated single-GPU serving tests.
  * Introduced a new test to verify correct batch slot release when using speculative decoding in disaggregated setups.

* **Bug Fixes**
  * Improved resource management by ensuring both sequence and speculative resources are properly released for completed context-only requests.
  * Enhanced draft request preparation to include models using n-gram decoding mode.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -896,6 +896,10 @@ class PyExecutor:
 
     def _executor_loop(self):
         torch.cuda.set_device(self.device_id)
+        is_ngram = hasattr(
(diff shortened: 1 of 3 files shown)


pcastonguay (changes requested):
Could we have a test that would have caught this issue? Maybe enqueue enough requests until we run out of batch slots?

pcastonguay (approved):

PR #6062, opened 2025-07-15
fix: Add $HOME/.local/bin to PATH when running docker in local user mode

MartinMarciniszyn:
# Minor fix for Docker with `LOCAL_USER=1`

This adds `$HOME/.local/bin` to the `PATH` environment variable when `docker` is executed in local user mode. This enables building without a separate virtual environment for Python.


ixlmar (approved):
Just realizing that because of the `getent passwd ${USER_ID} || useradd ...` the name of the user might actually differ from `${USER_NAME}`. However, this already affects L23 and thus this PR does not make a difference with regards to this corner case.

BTW, apparently `Dockerfile` allows `USER ${USER_ID}`, and afterwards one should be able to get the name from `$USER`.

PR #6110, opened 2025-07-17
add safe chunked broadcast

QiJune:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting during request broadcasting by including detailed error messages and payload size information.
  * Enhanced broadcasting of large Python objects by introducing chunked transmission for reliable data transfer and error handling.


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1253,13 +1254,19 @@ def _fetch_new_requests(self) -> List[RequestQueueItem]:
...
+                # Preserve original `new_requests` on rank 0 since it may contain
+                # Python-only objects (e.g., custom logits processors) not serializable by pybind.
+                _ = self._broadcast_new_requests(new_requests,
+                                                 py_request_objects)
+            else:
+                new_requests, py_request_objects = self._broadcast_new_requests(
+                    new_requests, py_request_objects)
+        except Exception as e:
+            raise ValueError(
+                f"broadcast new requests failed: {str(e)}, payload size: {len(pickle.dumps((new_requests, py_request_objects)))} bytes"
```
liji-nv, line 1268: The `str` for `e` may not be needed. I think fstring should have handled the conversion.

liji-nv: :) Perplexity gives me an opposite answer when I ask "Do I need to call str(object) when using python f-string?"

In f-string it seems it will call obj.__format__ if available or it will call obj.__str__.

qiaoxj07 (approved):

leslie-fang25 (approved):

PR #6111, opened 2025-07-17
[fix] Fix Triton build (#6076)

Wanli-Jiang:
Introduce https://github.com/NVIDIA/TensorRT-LLM/pull/6076 to fix build error


Wanli-Jiang:
Duplicated with https://github.com/NVIDIA/TensorRT-LLM/pull/6117, and the latter one contains more fixes, so we closed this PR.

PR #3070, opened 2025-03-25
feat: Add BW measurement

BatshevaBlack:
Request BW : This metric is measured from the generation side for each request. 
It is be calculated as:
Request BW = (Total KVCacheSize of Request​)/(Total Time for all Generations to Complete)

PR #6135, opened 2025-07-17
[TRTLLM-6549] chore: record delay introduced by disaggregated serving in kv cache measure

zhengd-nv:
## Description

Add delay metrics in KV cache measure helper to track extra TTFT introduced by disaggregated serving. Also add transmission time recording in the generation instance, using the context request id for comparation.

The delay refers to:

- Context instances: The time period between the first (also the only) token generated and the start time of cache sending.
- Generation instances: The time period between the arrival of the generation-only request and the start time of cache receiving. This includes generation-side initialization and sending request info to the context instance.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

## Summary by CodeRabbit

* **New Features**
  * Enhanced performance metrics tracking for KV cache transfers, including detailed timing, delay, and bandwidth measurements during cache send and receive operations.
  * Performance metrics collection can now be enabled via an environment variable for chat and completion endpoints.

* **Improvements**
  * Metrics output now includes delay, duration, and bandwidth for each KV cache transfer, providing more comprehensive insights into system performance.
  * Improved accuracy of transfer size reporting and refined logging for request context IDs.
  * Unified metric helper usage across cache formatters and data transceivers for consistent measurement handling.


tensorrt_llm/serve/openai_server.py
```diff
@@ -402,6 +406,9 @@ async def generator_wrapper(generator: AsyncIterator[Any]):
             promises: List[RequestOutput] = []
             postproc_params_collection: List[Optional[PostprocParams]] = []
             sampling_params = request.to_sampling_params()
+            # TODO: better way to enable metrics
+            if len(os.getenv("TRTLLM_KVCACHE_TIME_OUTPUT_PATH", "")) > 0:
+                sampling_params.return_perf_metrics = True
```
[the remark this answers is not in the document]
Shunkangz: I agreed that the KV cache time should be included in prometheus metrics. But here, you are providing something like per-request level metric. I still think that we should separate them.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
```diff
@@ -687,6 +712,15 @@ void CacheFormatter::unformat(TransferSession& session)
...
                 }
+                auto endTime = std::chrono::steady_clock::now();
+                double delay = 0.0;
+                if (arrivalTime != invalidTime)
+                {
+                    delay = std::chrono::duration<double, std::milli>(startTime - arrivalTime).count();
+                }
+                double cacheTransferTime
+                    = std::max(0.0, std::chrono::duration<double, std::milli>(endTime - startTime).count());
+                kvCacheMeasureHelper.appendKVCacheTransfer(ctxReqId, delay, cacheTransferTime, size);
```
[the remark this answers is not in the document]
Shunkangz: Even if we launch by MPI, we might still get conflict if we use separate mpi comm world. Is it correct? What do you think that we can append "ctx" or "gen" as part of the file name.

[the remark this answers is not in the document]
zhengd-nv: Actually we can also specify different `TRTLLM_KVCACHE_TIME_OUTPUT_PATH` for each mpirun to avoid conflict. But I agree that including ctx/gen in filename is better.
Note that CacheFormatter is not tagged with ctx/gen itself, so I have to do that in DataSender/Receiver. I believe it is a temporary state and ultimately transit to per-request metrics without tagging the formatter.

tensorrt_llm/serve/openai_server.py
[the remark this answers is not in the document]
zhengd-nv: Actually it is making "enable recording" and "enable returning the result" two different options, which affects all metrics involved in LlmRequest. This is far beyond the scope of this PR.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
```diff
@@ -350,9 +353,14 @@ void CacheFormatter::format(TransferSession& session)
...
             auto endTime = std::chrono::steady_clock::now();
+            double delay = 0.0;
+            if (recordDelay)
+            {
+                delay = std::chrono::duration<double, std::milli>(startTime - lastTokenTime).count();
+            }
             double cacheTransferTime
                 = std::max(0.0, std::chrono::duration<double, std::milli>(endTime - startTime).count());
-            kvCacheMeasureHelper.appendKVCacheTransfer(llmRequest.mRequestId, cacheTransferTime, size);
+            kvCacheMeasureHelper.appendKVCacheTransfer(llmRequest.mRequestId, delay, cacheTransferTime, size);
```
Tabrizian, line 363: We already capture KVCache time as a part of request stats using @pcastonguay's PR. I was curious what is the difference here? Can we use the existing metrics instead of introducing a new one? https://github.com/NVIDIA/TensorRT-LLM/pull/4156

zhengd-nv: The transfer time in `kvCacheMeasureHelper` is to measure the bandwidth of the transmission ( #3070 ), so the time is the actual transmission time (send/recv calls). The one in the per-request stats includes: 1) D sending request info to P, 2) P sending cache to D (KV cache measure here), 3) D move cache from recv buffers to cache storage.

Ideally, these metrics should be moved into per-request stats to provide a more detailed timing of cache transmission, but the per-request metrics are not accessible in disaggregated serving, because the underlying workers are manipulated via openai api and no per-request metrics interface available.

pcastonguay: @zhengd-nv I'm not sure I understand the following statement: "and no per-request metrics interface available". I have added per-request stats to the perf metrics which are accessible via the `/metrics` endpoint. 

This is the PR that added the functionality: https://github.com/NVIDIA/TensorRT-LLM/pull/4156

Could we add those KV cache stats to the per-request stats instead of writing them to a file and having to rely on an env variable? I think we should just enable them for disagg if 
```
   enable_iter_perf_stats
   enable_iter_req_stats
 ```
are set to `True`.

Comment (pcastonguay) [reply cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp:363]: Could we include the `ctx_in_transmission_requests` in the req stats then? I think we could set the request `return_perf_metrics` to `True` if  `enable_iter_req_stats` is True, and return those stats via the `/metrics` endpoint. Or am I missing some details?

Could having metrics only for the 1st rank be acceptable? I would want to avoid doing unnecessary broadcasts that could impact actual performance. 

I just feel that the current solution is a bit of a hack, where some metrics are controlled by parameters (`enable_iter_perf_stats` and `enable_iter_req_stats`) and accessed via the `metrics` endpoint, while others are controlled via env variables and dumped to a file.

**Comment by Shunkangz** [comment]
LGTM. As @pcastonguay said, let's merge this PR first and then after we merge the customer's metric PR, we can see how to refactor this part and make it clean.

**Review by Shunkangz — APPROVED** [review]

File: cpp/tensorrt_llm/batch_manager/dataTransceiver.h [hunk 4]
```diff
@@ -269,12 +269,24 @@ class DataRequester
...
+        double bandwidth; // in Gbps
+    };
+
     KvCacheMeasureHelper(std::string output_path)
         : mOutputPath(std::move(output_path))
     {
     }
 
-    void appendKVCacheTransfer(LlmRequest::RequestIdType requestId, double duration, size_t size)
+    void markAsContext(bool isContext)
```
Shixiaowei02, line 284: Would it be better to write a more general tagging function here? Something like `setName()`?

Shixiaowei02:
Thanks for your work. I do have some confusion regarding the `isContext()` member and would appreciate if that part could be adjusted. Everything else looks good to me.

cpp/tensorrt_llm/batch_manager/dataTransceiver.h
Shixiaowei02: It might be a bit early to merge this PR in its current intermediate state. Perhaps we could wait until more work is incorporated and the transitional aspects are resolved. In the meantime, if there's a need for temporary testing, we could consider using a side-branch rebased onto the main branch.

cpp/tensorrt_llm/batch_manager/cacheFormatter.h
```diff
@@ -76,6 +76,15 @@ class BaseCacheFormatter
...
     virtual ~BaseCacheFormatter() = default;
+
+    // TODO: better way for context/generation tagging
+    void markAsContext(bool isContext)
+    {
+        kvCacheMeasureHelper.markAsContext(isContext);
+    }
+
+protected:
+    KvCacheMeasureHelper kvCacheMeasureHelper{common::getEnvKVCacheTransferOutputPath()};
```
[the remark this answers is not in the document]
zhengd-nv: As an intermediate version, put it here to avoid duplication. Will be removed in metrics integration.

cpp/tensorrt_llm/batch_manager/dataTransceiver.h
zhengd-nv: After discussed @Shixiaowei02 offline, I made some changes to make the PR acceptable for merge, despite some intermediate codes exist for fast accessibility. The KV cache measures is planned to be integrated into incoming better-defined metrics in a later PR, and all intermediate code (tagging formatter, setting perf knobs by env, dumping measures to file) will be removed then.

Shixiaowei02 (approved):

PR #6147, opened 2025-07-17
[nvbug/5393888][nvbug/5393042] Always use `py_seq_slot`

netanel-haber:
Buggy flow prior to fix, for given request X:
**State:**
1. scheduler is MAX_UTILIZATION = pausing is enabled
2. overlap scheduler is enabled = previous batch is processed *after* current batch is scheduled

**Flow:**

1. [`self._pause_requests(scheduled_batch.paused_requests)`](https://github.com/NVIDIA/TensorRT-LLM/blob/a7184869001d28ca70a738e9862ea91cb147da8c/tensorrt_llm/_torch/pyexecutor/py_executor.py#L1110) is called for request X: `seq_slot` is deleted [**by cpp**](https://github.com/NVIDIA/TensorRT-LLM/blob/d71c6fe5267f4b61c51cc39d4594cdcb417f0703/cpp/include/tensorrt_llm/batch_manager/llmRequest.h#L830) when pause is called.
2. **After Pausing X**, [`self._update_requests(self.previous_batch.sample_state)`](https://github.com/NVIDIA/TensorRT-LLM/blob/d71c6fe5267f4b61c51cc39d4594cdcb417f0703/tensorrt_llm/_torch/pyexecutor/py_executor.py#L1138) is called, which also holds X in `self.previous_batch.sample_state.scheduled_requests.all_requests()`.
3. `update_requests` reads the seq_slot off of X, but it has been erased in 2.

cpp deletes the seq_slot for its uses, [as an implementation detail within the scheduler](https://github.com/NVIDIA/TensorRT-LLM/blob/d71c6fe5267f4b61c51cc39d4594cdcb417f0703/cpp/tensorrt_llm/batch_manager/assignReqSeqSlots.cpp#L33). `py_seq_slot` is not deleted when the request is paused, so we opt to use it. For consistency, we also transition to use this across the whole codebase where `seq_slot` was used before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the handling of sequence slot identifiers in request processing, ensuring consistent use of the appropriate attribute across model execution and sampling components. This improves reliability in request mapping and token assignment during inference.


QiJune (approved):
LGTM

Funatiq (approved):
This fix seems good enough. At some point we should also refactor the `seq_slot` handling in C++ or do things differently once we move away from the C++ slot manager.

PR #6210, opened 2025-07-21
[None][feat] Deepseek: Start Eagle work

IzzyPutterman:
## Summary by CodeRabbit

* **New Features**
  * Added support for tracking speculative decoding metadata in DeepseekV3 and Eagle3 models.
  * Introduced a configuration option to set the number of capture layers for Eagle decoding.

* **Improvements**
  * Enhanced flexibility in model behavior based on speculative decoding modes and capture layer settings.
  * Improved configuration handling for draft length synchronization in MTP decoding.

* **Bug Fixes**
  * Adjusted model weight loading to skip draft model parameters, preventing unintended loading.


tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -743,23 +744,34 @@ def forward(
         )
 
         if isinstance(self.mlp, Deepseekv3MoE):
+
+            if spec_metadata is not None and spec_metadata.is_layer_capture(
```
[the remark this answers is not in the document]
IzzyPutterman: If layer capture is False, then we enable the fusion.  Only on layers where Eagle/MTP asks for the hidden state are the fusion disabled.

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -1027,6 +1048,8 @@ def forward(
         input_ids: Optional[torch.IntTensor] = None,
         position_ids: Optional[torch.IntTensor] = None,
         inputs_embeds: Optional[torch.FloatTensor] = None,
+        spec_metadata: Optional[SpecMetadata] = None,
+        **kwargs,
```
mikeiovine, line 1052: nit: extra kwargs not necessary?

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -1108,19 +1131,22 @@ def forward(
         input_ids: torch.IntTensor = None,
         position_ids: Optional[torch.IntTensor] = None,
         inputs_embeds: Optional[torch.FloatTensor] = None,
-        spec_metadata: Optional[MTPSpecMetadata] = None,
+        spec_metadata: Optional[SpecMetadata] = None,
         return_context_logits: bool = False,
         **kwargs,
     ) -> torch.Tensor:
+        #TODO Izzy check this
```
mikeiovine, line 1138: nit: comment can be removed?

tensorrt_llm/_torch/speculative/eagle3.py
```diff
@@ -197,8 +200,8 @@ class Eagle3OneModelSpecMetadata(SpecMetadata):
     batch_indices_cuda: Optional[torch.Tensor] = None
 
     def __post_init__(self):
-        if self.num_layers == 1:
-            self.layers_to_capture = (1, )
+        if self.num_layers == 1 or self.num_capture_layers == 1:
+            self.layers_to_capture = (self.num_layers - 1, )
```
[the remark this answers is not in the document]
mikeiovine: I'm pretty sure you can't have `num_layers == 1` on the 1 model path. This looks like some code that was copied over from 2-model and then mangled. 

One 1-model, `num_layers` is always equal to the number of layers in the target model https://github.com/NVIDIA/TensorRT-LLM/blob/ee3cbb073e118d4f8b6e40ee7a1f17a14b5b4496/tensorrt_llm/_torch/speculative/utils.py#L45

tensorrt_llm/_torch/speculative/eagle3.py
```diff
@@ -90,8 +91,8 @@ class Eagle3SpecMetadata(SpecMetadata):
     eagle3_resource_manager: Optional[Eagle3ResourceManager] = None
 
     def __post_init__(self):
-        if self.num_layers == 1:
-            self.layers_to_capture = (0, )
+        if self.num_layers == 1 or self.num_capture_layers == 1:
+            self.layers_to_capture = (self.num_layers - 1, )
```
mikeiovine, line 95: Can we just get rid of `num_capture_layers`?  It should be equal to `len(layers_to_capture)`. Feel free to add a read-only `@property` that derives it from this field for convenience though.

tensorrt_llm/_torch/speculative/interface.py
```diff
@@ -85,12 +85,19 @@ def extend_ctx(self, attention_backend: Type[AttentionBackend]):
...
         """
-
         # Fixme: only trtllm attention backend supports eagle3 generation-phase kernels on blackwell.
-        return ((self.is_eagle3() or self.is_draft_target())
-                and not (isinstance(attention_backend, TrtllmAttention)
-                         and get_sm_version() == 100)
-                ) or self.is_ngram() or self.is_user_provided()
+
+        if self.is_eagle3() or self.is_draft_target():
+            if get_sm_version() == 100 and isinstance(attention_backend,
```
mikeiovine, line 91: Should be `issubclass`. That is technically also applicable to the `attention_backend is TrtllmAttention` below (though the usage of `issubclass` vs `is` is a minor issue, both are functionally the same since subclass of `TrtllmAttention` are not likely)

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -326,6 +326,7 @@ class EagleDecodingConfig(DecodingBaseConfig):
     num_eagle_layers: Optional[int] = None
     max_non_leaves_per_layer: Optional[int] = None
     eagle3_one_model: Optional[bool] = True
+    num_capture_layers: Optional[int] = 3
```
mikeiovine, line 329: Type annotation should be `int` if it can't be `None`.

Why are we making this part of the part of the spec config? EAGLE3 should capture hidden states from 3 layers. It's not eagle 3 if it doesn't do that

tensorrt_llm/_torch/speculative/interface.py
```diff
@@ -178,6 +185,9 @@ def create_cuda_graph_metadata(self, max_batch_size: int):
         cuda_graph_metadata.__post_init__()
         return cuda_graph_metadata
 
+    def is_layer_capture(self, layer_id: int):
+        return False
+
```
mikeiovine, line 190: This could use a docstring. Something that explains its relationship to `maybe_capture_hidden_states`

tensorrt_llm/_torch/speculative/interface.py
```diff
@@ -85,12 +85,19 @@ def extend_ctx(self, attention_backend: Type[AttentionBackend]):
...
-                         and get_sm_version() == 100)
-                ) or self.is_ngram() or self.is_user_provided()
+
+        if self.is_eagle3() or self.is_draft_target():
+            if get_sm_version() == 100 and isinstance(attention_backend,
+                                                      TrtllmAttention):
+                return False
+            elif get_sm_version(
+            ) == 100 and attention_backend is TrtllmAttention:
+                return False
```
mikeiovine, line 96: This can be removed when you fix the check above

tensorrt_llm/_torch/speculative/interface.py
```diff
@@ -85,12 +85,19 @@ def extend_ctx(self, attention_backend: Type[AttentionBackend]):
...
+            else:
+                return True
+        else:
+            return self.is_ngram() or self.is_user_provided()
```
mikeiovine, line 100: Out of scope for this PR but it doesn't look right to me - ngram at the very least should have the same kernel restrictions as eagle3

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -1108,19 +1131,22 @@ def forward(
...
         hidden_states = self.model(
             input_ids=input_ids,
             attn_metadata=attn_metadata,
             position_ids=position_ids,
             inputs_embeds=inputs_embeds,
+            spec_metadata=spec_metadata,
         )
 
         if spec_metadata and spec_metadata.spec_dec_mode.is_mtp():
+            # TODO Merge API with EagleWorker in modeling_speculative.py
```
mikeiovine, line 1149: We should do this now that there are multiple possible 1-model paths. Otherwise I think it's a bit too hard to follow

mikeiovine (commented):
We seem to be allowing the use of 1 capture layer instead of 3 - what's the motivation? 

That means we're effectively supporting EAGLE2, not EAGLE3. It's yet another thing to add on the support matrix so I would rather stick to the convention of only using EAGLE3 if possible. However, I'm not opposed to adding EAGLE2 support if there's a strong motivation for it

tensorrt_llm/_torch/speculative/interface.py
IzzyPutterman: I assume all of them should have the same restrictions, not sure why in the OG they were not included in the parethesis.

tensorrt_llm/_torch/models/modeling_deepseekv3.py
mikeiovine: Per our discussion offline we can do this on a followup. Let's open a JIRA though so it doesn't get swept under the rug

tensorrt_llm/_torch/models/modeling_deepseekv3.py
IzzyPutterman: modeling_speculative passes kwargs in FW, it is needed

mikeiovine (dismissed):

mikeiovine (approved):
Accept to unblock

syuoni (approved):
Reviewed the changes to `llm_args.py`; looks good to me.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -443,6 +445,12 @@ def spec_dec_mode(self):
             return TorchSpeculativeDecodingMode.EAGLE3_ONE_MODEL
         return TorchSpeculativeDecodingMode.EAGLE3
 
+    @functools.cached_property
+    def num_capture_layers(self):
+        if self.eagle3_layers_to_capture is not None:
+            return len(self.eagle3_layers_to_capture)
+        return 3
```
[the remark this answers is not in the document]
IzzyPutterman: Updated

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -523,7 +531,9 @@ class MTPDecodingConfig(DecodingBaseConfig):
 
     @classmethod
     def from_dict(cls, data: dict):
-        return cls(**data)
+        out = cls(**data)
+        out.max_draft_len = out.num_nextn_predict_layers
```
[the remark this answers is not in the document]
IzzyPutterman: max_draft_len is a property of `DecodingBaseConfig`, which MTP inherits from.  In the MTP case, `num_nextn_predict_layers` should be equal to max_draft_len I believe.

SimengLiu-nv (approved):

PR #6241, opened 2025-07-22
[PERF] Don't use hmac encryption for loopback interfaces

vadiklyutiy:
## Description

In `class ZeroMqQueue` disable encryption for loopback interfaces. 

**The Reason**
For Qwen-2.5-3B-VL main process (that get requests, pre-process images, send request to worker) is bottleneck. Images preprocessing and sending takes more time than inference. 

`send` is around 35% from all time 
image

inside send we spend around half to sign image with `hmac`. 

I think we can avoid encryption for loopback interfaces

 
## Performance 
Tested Qwen-2.5-3B-VL model. H100. Images 512x512
Before image processing took **23ms** per image. 
After **17ms**
  
image

## Summary by CodeRabbit

* **New Features**
  * Automatically disables encryption for local or loopback IPC connections to improve performance.
* **Bug Fixes**
  * Ensures correct detection of local addresses to apply appropriate security settings.

### DIFF
--- a/tensorrt_llm/executor/ipc.py
+++ b/tensorrt_llm/executor/ipc.py
@@ -17,7 +17,7 @@ from ..llmapi.utils import (ManagedThread, enable_llm_debug, print_colored,
                             print_colored_debug)
 
 
-def is_address_is_loopback(address: str) -> bool:
+def is_address_loopback(address: str) -> bool:
     """Check if the address represents a local/loopback connection."""
     # IPC and inproc are always local
     if address.startswith(("ipc://", "inproc://")):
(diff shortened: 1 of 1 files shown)


karljang (changes requested):
@vadiklyutiy,
Thank you for the contribution and apologies for the delayed response.

I believe that security settings should never be automatically downgraded based on network address assumptions. The current approach can create a significant security vulnerability that could be exploited in common deployment scenarios, such as SSH tunneling and Docker or Kubernates port forwarding.

So, could you propose an alternative solution that makes 'use_hmac_encryption' configurable by users?

PR #6245, opened 2025-07-22
[None][feat] Implement advanced sampling for one model path mtp/eagle

nvxuanyuc:
[feat] Implement pytorch sampler for MTP

## Description
- Adds support for advanced sampling in the PyTorch path for MTP with speculative decoding
   - Previously, only greedy mode was supported.
   - Implements temperature, top-p, top-k, and min-p sampling parameters in Python when using MTP speculative decoding (for DeepSeek) @pathorn [[#5627]](https://github.com/NVIDIA/TensorRT-LLM/pull/5627)
   - Adds support for returning log-probs from the Pytorch sampler [related to [#5620]](https://github.com/NVIDIA/TensorRT-LLM/pull/5620)

The default behavior of the MTP pytorch decoder remains greedy sampling. Advanced sampling can be enabled via the `enable_mixed_sampler` flag in `TorchLlmArgs`.

## Test Coverage

- Added test for greedy mode `temperature <= 1e-2` using the new PyTorch sampler.
- Tests for advanced sampling modes are not yet included.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -1234,16 +1234,21 @@ class PyTorchModelEngine(ModelEngine):
         def collect_req_spec_dec_sampling_params(request: LlmRequest,
                                                  draft_len: int = 0):
 
-            def get_request_temperature(request: LlmRequest) -> float:
+            def get_request_temperature(request: LlmRequest,
+                                        is_greedy: bool) -> float:
+                if is_greedy:
+                    return 0.01  # avoid numerical errors and keep the same behavior as greedy sampling
                 if not request.sampling_config.temperature:
                     return 1.0
                 temperature = request.sampling_config.temperature[0]
-                if 0 < temperature < 1e-2:
+                if 0 <= temperature < 1e-2:
                     # temperature less than 0.01 may cause numerical errors
                     temperature = 0.01
                 return temperature
 
-            def get_request_top_k(request: LlmRequest) -> int:
+            def get_request_top_k(request: LlmRequest, is_greedy: bool) -> int:
+                if is_greedy:
+                    return 1
                 if not request.sampling_config.top_k:
                     top_k = 0
                 else:
@@ -1269,9 +1274,20 @@ class PyTorchModelEngine(ModelEngine):
                     min_p = request.sampling_config.min_p[0]
                 return min_p
 
-            temperatures.extend([get_request_temperature(request)] *
-                                (draft_len + 1))
-            top_k.extend([get_request_top_k(request)] * (draft_len + 1))
+            def is_greedy_sampling(request: LlmRequest) -> bool:
+                if (request.sampling_config.temperature is None
+                        and request.sampling_config.top_p is None
+                        and request.sampling_config.top_k is None):
+                    return True
+                return False
+
+            is_greedy_req = is_greedy_sampling(request)
+
+            temperatures.extend(
+                [get_request_temperature(request, is_greedy_req)] *
+                (draft_len + 1))
+            top_k.extend([get_request_top_k(request, is_greedy_req)] *
+                         (draft_len + 1))
             top_p.extend([get_request_top_p(request)] * (draft_len + 1))
             min_p.extend([get_request_min_p(request)] * (draft_len + 1))
 
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -14,6 +14,7 @@ import torch
 import torch._dynamo.config
 
 import tensorrt_llm.bindings.internal.userbuffers as ub
+from tensorrt_llm._torch.pyexecutor.llm_request import LlmRequest
 from tensorrt_llm._utils import (is_trace_enabled, nvtx_range, release_gc,
                                  torch_dtype_to_str, trace_func)
 from tensorrt_llm.inputs.multimodal import (MultimodalParams,
@@ -172,6 +173,12 @@ class PyTorchModelEngine(ModelEngine):
         self.sparse_attention_config = sparse_attention_config
         self.enable_spec_decode = self.is_spec_decode
         self.is_draft_model = is_draft_model
+
+        self.is_advanced_spec_dec_sampler = (self.is_spec_decode and (
+            (self.spec_config.spec_dec_mode.is_mtp_one_model()
+             and self.spec_config.use_advanced_spec_dec_sampling) or
+            (self.spec_config.spec_dec_mode.is_eagle3_one_model()
+             and self.spec_config.use_advanced_spec_dec_sampling)))
 
         self.attn_runtime_features = attn_runtime_features or AttentionRuntimeFeatures(
         )
@@ -1211,10 +1218,62 @@ class PyTorchModelEngine(ModelEngine):
         num_cached_tokens_per_seq = []  # per sequence
         draft_tokens = []
         draft_lens = []
+
         gen_request_seq_slots = []  # per generation request
         multimodal_params_list = []
         mrope_position_ids = []
         num_accepted_draft_tokens = []  # per request
+
+        if self.is_advanced_spec_dec_sampler:
+            temperatures = []
+            top_k = []
+            top_p = []
+            min_p = []
+
+        # advanced mtp sampling's request preprocessing helper functions
+        def collect_req_spec_dec_sampling_params(request: LlmRequest,
+                                                 draft_len: int = 0):
+
+            def get_request_temperature(request: LlmRequest) -> float:
+                if not request.sampling_config.temperature:
+                    return 1.0
+                temperature = request.sampling_config.temperature[0]
+                if 0 < temperature < 1e-2:
+                    # temperature less than 0.01 may cause numerical errors
+                    temperature = 0.01
+                return temperature
+
+            def get_request_top_k(request: LlmRequest) -> int:
+                if not request.sampling_config.top_k:
+                    top_k = 0
+                else:
+                    top_k = request.sampling_config.top_k[0]
+
+                # set k to a very large value (larger than vocab size) to disable top_k sampling
+                TOP_K_DISABLED = torch.iinfo(torch.int32).max
+                if top_k <= 0:
+                    top_k = TOP_K_DISABLED
+                return top_k
+
+            def get_request_top_p(request: LlmRequest) -> float:
+                if not request.sampling_config.top_p:
+                    top_p = 1.0
+                else:
+                    top_p = request.sampling_config.top_p[0]
+                return top_p
+
+            def get_request_min_p(request: LlmRequest) -> float:
+                if not request.sampling_config.min_p:
+                    min_p = 0.0
+                else:
+                    min_p = request.sampling_config.min_p[0]
+                return min_p
+
+            temperatures.extend([get_request_temperature(request)] *
+                                (draft_len + 1))
+            top_k.extend([get_request_top_k(request)] * (draft_len + 1))
+            top_p.extend([get_request_top_p(request)] * (draft_len + 1))
+            min_p.extend([get_request_min_p(request)] * (draft_len + 1))
 
         for request in scheduled_requests.context_requests:
             request_ids.append(request.py_request_id)
@@ -1233,7 +1292,6 @@ class PyTorchModelEngine(ModelEngine):
             past_seen_token_num = begin_compute
             num_cached_tokens_per_seq.append(past_seen_token_num)
             request.cached_tokens = num_cached_tokens_per_seq[-1]
-
             # Multimodal
             py_multimodal_runtime = MultimodalRuntimeData(
                 mm_token_lengths=request.multimodal_lengths,
@@ -1268,6 +1326,9 @@ class PyTorchModelEngine(ModelEngine):
                 #re-assign the multimodal_data to the request after to_device for generation requests
                 request.py_multimodal_data = multimodal_params.multimodal_data
                 multimodal_params_list.append(multimodal_params)
+
+            if self.is_advanced_spec_dec_sampler:
+                collect_req_spec_dec_sampling_params(request)
 
             request.py_batch_idx = request.py_seq_slot
 
@@ -1344,6 +1405,11 @@ class PyTorchModelEngine(ModelEngine):
                               past_seen_token_num + 1 + num_draft_tokens)))
                 num_cached_tokens_per_seq.append(past_seen_token_num)
                 request.cached_tokens = num_cached_tokens_per_seq[-1]
+
+                if self.is_advanced_spec_dec_sampler:
+                    collect_req_spec_dec_sampling_params(
+                        request, num_draft_tokens)
+
                 # update batch index
                 request.py_batch_idx = request.py_seq_slot
             else:
@@ -1378,6 +1444,9 @@ class PyTorchModelEngine(ModelEngine):
                     num_cached_tokens_per_seq.append(past_seen_token_num +
                                                      self.runtime_draft_len + 1)
                 request.cached_tokens = num_cached_tokens_per_seq[-1]
+                if self.is_advanced_spec_dec_sampler:
+                    collect_req_spec_dec_sampling_params(
+                        request, self.runtime_draft_len)
                 if self.enable_spec_decode and spec_config.spec_dec_mode.extend_ctx(
                         self.attn_backend):
                     prompt_lengths.append(1 + self.runtime_draft_len)
@@ -1407,6 +1476,10 @@ class PyTorchModelEngine(ModelEngine):
 
             # update batch index
             request.py_batch_idx = request.py_seq_slot
+
+            if self.is_advanced_spec_dec_sampler:
+                collect_req_spec_dec_sampling_params(
+                    request, self.original_max_draft_len)
 
         for request in generation_requests:
             request_ids.append(request.py_request_id)
@@ -1466,6 +1539,10 @@ class PyTorchModelEngine(ModelEngine):
                                 "mrope_config.mrope_position_deltas"
                             ])
                         multimodal_params_list.append(multimodal_params)
+
+            if self.is_advanced_spec_dec_sampler:
+                collect_req_spec_dec_sampling_params(request,
+                                                     self.runtime_draft_len)
 
             request.py_batch_idx = request.py_seq_slot
             # Do not add a gen_request_seq_slot for CUDA graph dummy requests
@@ -1701,6 +1778,11 @@ class PyTorchModelEngine(ModelEngine):
                                                                 total_draft_lens]
             spec_metadata.request_ids = request_ids
             spec_metadata.gather_ids = self.gather_ids_cuda[:len(gather_ids)]
+
+            if self.is_advanced_spec_dec_sampler:
+                spec_metadata.update_advanced_sampling_params(
+                    temperatures, top_k, top_p, min_p)
+
             spec_metadata.num_generations = len(
                 scheduled_requests.generation_requests)
             spec_metadata.num_tokens = total_num_tokens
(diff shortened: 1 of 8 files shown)


tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1129,6 +1154,45 @@ def _prepare_tp_inputs(
...
 
+        if self.is_advanced_mtp_sampler:
+            temperatures = []
+            top_k = []
+            top_p = []
+            min_p = []
+
+        def get_request_temperature(request: LlmRequest) -> float:
+            if not request.sampling_config.temperature:
+                return 0.7
```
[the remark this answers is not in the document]
nvxuanyuc: Will change to 1.0 for consistency with tllm sampler.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -151,6 +151,118 @@ def top_p_sampling_batch(logits: torch.Tensor, top_p: float = 0.9):
...
+    p: Optional[torch.Tensor],
+) -> torch.Tensor:
+    """Apply top-k and top-p masks to the logits.
+
+    If a top-p is used, this function will sort the logits tensor,
+    which can be slow for large batches.
+
+    The logits tensor may be updated in-place.
+    """
+    logits_sort, logits_idx = logits.sort(dim=-1, descending=False)
```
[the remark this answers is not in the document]
nvxuanyuc: We may add a check in `forward_native` to skip top-k and top-p filtering when not needed.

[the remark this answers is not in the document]
nvxuanyuc: We can’t skip top-k/top-p filtering logic even when their values are set to default for all requests because:

- top-k/top-p are runtime inputs, not static config — can vary per batch and must be handled dynamically. As a result:
    - Device-side conditional execution isn’t feasible with native torch ops, since we are not using custom kernels.
    - Host-side checks (e.g., conditional flag) would break CUDA graph capture and introduce CPU-GPU sync overhead.

To maintain compatibility with CUDA graphs and avoid performance penalties from branching, top-k and top-p filtering are still applied unconditionally.

tensorrt_llm/llmapi/tokenizer.py
```diff
@@ -94,8 +94,13 @@ def convert_ids_to_tokens(
             self,
             ids: Union[int, List[int]],
             skip_special_tokens: bool = False) -> Union[str, List[str]]:
-        return self.tokenizer.convert_ids_to_tokens(
+        # DeepSeek vocabulary has token ids not mapped to any tokens, these will get converted to None
+        # by the tokenizer. We need to filter them out.
+        tokens = self.tokenizer.convert_ids_to_tokens(
             ids, skip_special_tokens=skip_special_tokens)
```
syuoni, line 100: Currently, the default path is `hf_decode_incrementally`, which doesn't reach here. So how did you find this issue? (Just concerning if there is some unexpected issues.)

tensorrt_llm/llmapi/tokenizer.py
```diff
@@ -94,8 +94,13 @@ def convert_ids_to_tokens(
...
+        if isinstance(ids, int):
+            return tokens  # Single token, return as-is (could be None)
```
syuoni, line 102: Should we return an empty list if the single id is mapped to None? If not, will this None cause problems?

tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -822,8 +822,15 @@ def sample_and_accept_draft_tokens(
...
                     mtp_num_modules, batch_size, num_contexts, logits.shape[-1])
             else:
-                # Do greedy sampling for the input logits
-                target_tokens = torch.argmax(logits, dim=-1)
+                if self.spec_config.use_advanced_mtp_sampler:
+                    # Do advanced sampling for the input logits
+                    # target_log_probs currently unused but kept for future log probs support in MTP
+                    target_tokens, target_log_probs = sampling_batch(
+                        logits, spec_metadata.temperatures, spec_metadata.top_k,
+                        spec_metadata.top_p, spec_metadata.min_p)
```
syuoni, line 830: Is it possible to check if there are any requests with topk/topp sampling in the current batch, and decide whether to call `sampling_batch` automatically? If so, we can save one flag `use_advanced_mtp_sampler`.

I think we have done something similar: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/sampler.py#L412

nvxuanyuc: Unlike `fast_path` referred here, MTP sampling runs inside the forward pass, so we can't use host-side branching without breaking cuda graph capture and will cause CPU-GPU sync. Since mtp sampling uses native torch ops, device-side branching isn't possible either.

tensorrt_llm/llmapi/tokenizer.py
nvxuanyuc: It's a failure case found by DeepInfra in production. It seems a rare case and unrelated to sampler side. Will revert in this PR.

tensorrt_llm/llmapi/tokenizer.py
nvxuanyuc: Returning [] makes more sense. But these changes will be reverted in this PR as stated above.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -262,6 +263,10 @@ def __init__(
         lora_config: Optional[LoraConfig] = None,
         is_draft_model: bool = False,
     ):
+        # Set deterministic seed for consistent multi-GPU sampling using PyTorch RNG
+        # operations that avoid torch.multinomial's CPU-GPU sync overhead
+        torch.manual_seed(0)
```
netanel-haber, line 268: Isn't this a breaking change for all users? Why is this strictly necessary to be done at the global level? Can this be done before and after multinomial?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1132,6 +1156,45 @@ def _prepare_tp_inputs(
...
+            return temperature
+
+        def get_request_top_k(request: LlmRequest) -> int:
+            if not request.sampling_config.top_k:
+                top_k = 0
+            else:
+                top_k = request.sampling_config.top_k[0]
+
+            if top_k <= 0:
+                top_k = 2147483647
```
netanel-haber, line 1181: Use int64.max or something else, for searchability/readability, instead of a magic number

tensorrt_llm/_torch/speculative/mtp.py
netanel-haber: [Motivation being, it will reduce surface area]

syuoni: Makes sense to me. Please consider @netanel-haber 's suggestion, too. Thanks

syuoni (approved):

tensorrt_llm/_torch/pyexecutor/model_engine.py
nvxuanyuc: We now only [set the seed ](https://github.com/NVIDIA/TensorRT-LLM/pull/6245/files#diff-4394e1671109642e2baad3fde529f5bdc0f5e1d09e2395e10f5b54391bf6486aR211)when we enable this mtp advanced sampler.

tensorrt_llm/_torch/pyexecutor/model_engine.py
nvxuanyuc: Since python doesn't have built-in max constant, we now define [`TOP_K_DISABLED`](https://github.com/NVIDIA/TensorRT-LLM/pull/6245/files#diff-203c6491ca0bedf455eeee3ed7939c3209f3f14c383594c26261e1dc48a00395R1187) as a large value (greater than vocab size). Any better suggestions?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -374,6 +381,23 @@ def __init__(
             self.without_logits = self.spec_config.spec_dec_mode.without_logits(
             )
             self.max_draft_len = spec_config.max_draft_len
+
+            if self.is_advanced_mtp_sampler:
```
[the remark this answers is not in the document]
nvxuanyuc: We retain the logic for collecting request-specific speculative parameters in model_engine, as moving it to the MTP side would require duplicating the loop over all requests and passing multiple parameters into model_engine. Instead, we encapsulate this logic in a [helper function](https://github.com/NVIDIA/TensorRT-LLM/pull/6245/files#diff-203c6491ca0bedf455eeee3ed7939c3209f3f14c383594c26261e1dc48a00395R1168) to keep the code cleaner.
Other initialization steps and SpecMetadata related settings have been moved to the MTP side.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -405,6 +405,7 @@ class MTPDecodingConfig(DecodingBaseConfig):
     relaxed_topk: int = 1
     relaxed_delta: float = 0.
     use_mtp_vanilla: bool = False
+    use_advanced_mtp_sampler: Optional[bool] = False
```
Superjomn, line 408: Should the type be `bool`? Does `None` have a special meaning beyond `False`?

tensorrt_llm/_torch/pyexecutor/model_engine.py
nvxuanyuc: Now uses [torch.iinfo(torch.int32).max](https://github.com/NVIDIA/TensorRT-LLM/pull/6245/files#diff-203c6491ca0bedf455eeee3ed7939c3209f3f14c383594c26261e1dc48a00395R1199)

tensorrt_llm/_torch/speculative/mtp.py
nvxuanyuc: Updated the logic to reuse the `enable_mixed_sampler` flag. This flag is now passed to the model via `model_config` to enable MTP worker setup and control the sampling mode. [[load_model](https://github.com/NVIDIA/TensorRT-LLM/pull/6245/files#diff-203c6491ca0bedf455eeee3ed7939c3209f3f14c383594c26261e1dc48a00395R301); [mtp worker init](https://github.com/NVIDIA/TensorRT-LLM/pull/6245/files#diff-4394e1671109642e2baad3fde529f5bdc0f5e1d09e2395e10f5b54391bf6486aR380); [mtp sampler branching](https://github.com/NVIDIA/TensorRT-LLM/pull/6245/files#diff-4394e1671109642e2baad3fde529f5bdc0f5e1d09e2395e10f5b54391bf6486aR899)]

tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -324,6 +376,11 @@ def __init__(self, spec_config: "MTPDecodingConfig", model_config=None):
         self.model_config = model_config
         self.is_thop = False
 
+        # Default to greedy mode. If true, use advanced pytorch sampling strategy.
+        self.enable_mixed_sampler = False
+        if self.model_config is not None:
+            self.enable_mixed_sampler = self.model_config.enable_mixed_sampler
```
ixlmar, line 382: This could be a `@property` rather than a copy, to avoid potential consistency issues in the future.

tensorrt_llm/_torch/pyexecutor/model_engine.py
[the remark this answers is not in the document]
Funatiq: I agree with @netanel-haber, I don't think collecting sampling parameters in `model_engine` is a good design choice. Do these parameters even belong into `spec_metadata` or should we create a separate class for them? They seem general enough to justify an independent class. (They also don't seem to be specific to MTP so the new functions shouldn't include mtp in the names)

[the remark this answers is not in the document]
jhaotingc: Hi @Funatiq @netanel-haber, I agree it's a suboptimal solution putting the samplings in spec_metadata.
Is there a plan to have [make_decoding_batch_input](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/sampler.py#L666) or similar design in TorchSampler?
What's the roadmap for TorchSampler having advanced sampling feature? Or a design we can follow?

One reason of having those in speculative related class and also parsing it in model_engine is because, [sample_async](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/py_executor.py#L744) in py_executor happens after model forward, while MTP is part of model forward. So a better design is add extra logics in py_executor, iterating through scheduled_batch, parsing the sampling params, before model forward, or iterate that during MTPWorker logit sampling.

This PR is here because the feature has been lacking for long, we received many customer requests wanting this feature, so we help merging this community contributing PR to main. @laikhtewari

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -235,6 +235,113 @@ def is_generation_model(self) -> bool:
...
+
+def apply_min_p(
+    logits: torch.Tensor,
+    min_p: torch.Tensor,
+) -> torch.Tensor:
+    """
+    Filters logits using adaptive probability thresholding.
+    """
+    # Convert logits to probability distribution
+    probability_values = torch.nn.functional.softmax(logits, dim=-1)
```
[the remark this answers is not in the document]
IzzyPutterman: nvm im wrong here, misread something

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -235,6 +235,113 @@ def is_generation_model(self) -> bool:
...
+
+
+def apply_top_k_top_p(
+    logits: torch.Tensor,
+    k: Optional[torch.Tensor],
+    p: Optional[torch.Tensor],
+) -> torch.Tensor:
+    """Apply top-k and top-p masks to the logits.
+
+    If a top-p is used, this function will sort the logits tensor,
```
longlee0622: As a perf optimization, should we skip the expensive sorting / softmax / cumsum ops for top_p >=1?

longlee0622: @IzzyPutterman @ziyixiong-nv

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -235,6 +235,113 @@ def is_generation_model(self) -> bool:
...
+    # Reshape min_p for broadcasting
+    adjusted_min_p = min_p.unsqueeze(1) * max_probabilities
+    # Identify valid tokens using threshold comparison
+    valid_token_mask = probability_values >= adjusted_min_p
+    # Apply mask using boolean indexing
+    logits[~valid_token_mask] = -float("inf")
+    return logits
+
+
+def apply_top_k_top_p(
```
ixlmar: Note that we already have https://github.com/NVIDIA/TensorRT-LLM/blob/f81caf54912b2381ceecde21390a0daa98351857/tensorrt_llm/_torch/pyexecutor/sampling_utils.py#L142

and some related functions. We should not duplicate this basic functionality, so let's use the existing functions and extend them as necessary (adding the `Tensor`-type k, p, temperature, etc.).

Also note that I am working on FlashInfer.sampling based alternatives for those functions. This upcoming PR brings support for `Tensor`-type k, p, temperature, etc. when FlashInfer is used. If you integrate the improvements made here for the non-FlashInfer case, this could give a quite nice feature set.

Ideally, this PR could (i) improve the existing sampling routines and (ii) use them via https://github.com/NVIDIA/TensorRT-LLM/blob/f81caf54912b2381ceecde21390a0daa98351857/tensorrt_llm/_torch/pyexecutor/sampling_utils.py#L312

ixlmar (changes requested):
Partially reviewed

tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar: See also TRTLLM-7723 (and TRTLLM-7152) for scope of ongoing work.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -235,6 +235,113 @@ def is_generation_model(self) -> bool:
...
+    k: Optional[torch.Tensor],
+    p: Optional[torch.Tensor],
+) -> torch.Tensor:
+    """
+    PyTorch-native implementation of top-k and top-p sampling.
+
+    The logits tensor may be updated in-place.
+    """
+    logits = apply_top_k_top_p(logits, k, p)
+    probs = logits.softmax(dim=-1, dtype=torch.float32)
```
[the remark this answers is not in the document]
IzzyPutterman: I think I am also wrong here

tensorrt_llm/_torch/pyexecutor/sampler.py
IzzyPutterman: I think I tested flashinfer with cuda graphs and it was breaking a bunch.  With the generator objects its quite annoying in TRTLLM becuase in warmup we alternate between cuda graph warmup and non-cuda graph warmup, which breaks

IzzyPutterman: Worth a double check ofc, perhaps there is an easy way around it

jhaotingc:
@ixlmar @IzzyPutterman would you mind sharing the latest plan regarding this PR? Is the plan to merge this in then add in more performant kernels?

@ixlmar we can give you the access to this branch if you'd like to develop on top of this.

ixlmar:
> @ixlmar @IzzyPutterman would you mind sharing the latest plan regarding this PR? Is the plan to merge this in then add in more performant kernels?
> 
> @ixlmar we can give you the access to this branch if you'd like to develop on top of this.

@jhaotingc Thanks. As far as I can tell, the code that I think we should be extending and reusing (`top_k_top_p_sampling_batch`) is already in main and my ongoing and planned work is so far orthogonal to what is happening here. My main concern is to improve the existing sampling code rather than introducing the separate but functionally overlapping `apply_top_k_top_p` from this PR, which I fear will make things harder to maintain in the long run.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -235,6 +235,113 @@ def is_generation_model(self) -> bool:
...
+    probs: torch.Tensor,
+) -> torch.Tensor:
+    """Randomly sample from the probabilities.
+
+    We use this function instead of torch.multinomial because torch.multinomial
+    causes CPU-GPU synchronization.
+    """
+    q = torch.empty_like(probs)
+    q.exponential_()
+    return probs.div_(q).argmax(dim=-1).view(-1)
```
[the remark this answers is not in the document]
ixlmar: (Disclaimer: I might have well overlooked that `torch.multinomial` is syncing so far, so I would be curious to hear more on this.)

tensorrt_llm/_torch/pyexecutor/sampler.py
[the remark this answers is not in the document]
ixlmar: I think @IzzyPutterman is right: `apply_min_p` evaluates the softmax of the temperature-scaled logits and uses that to mask out some of the logits (set to `-inf`). The probs could be masked in the same way (set to `0`). The resulting probs can (mind paragraph below) then be reused in `apply_top_k_top_p`, which masks out more logits/probs.

Every time logits/probs are masked, it is sufficient to renormalize the probs such that they sum to one, which is much cheaper than computing softmax. This is probably also why https://docs.flashinfer.ai/api/sampling.html uses function names like `..._renorm_probs`.

Note that much of this is already worked out in #8581, albeit using `flashinfer.sampling`.

ixlmar (commented):
I will be unavailable until next week and cannot "un-request changes" on this PR, so please feel free to merge without my explicit approval if this is needed to unblock other tasks.

PR #6247, opened 2025-07-22
doc: update known issues

QiJune:
## Summary by CodeRabbit

* **Documentation**
  * Updated release notes to include a new known issue regarding performance regression with full chunked attention support for LLaMA4 models on sequences longer than 8K tokens.

### DIFF
--- a/docs/source/release-notes.md
+++ b/docs/source/release-notes.md
@@ -73,6 +73,7 @@ All published functionality in the Release Notes has been fully tested and verif
 ### Known Issues
 - accuracy/test_cli_flow::TestGpt2::test_beam_search_large is broken.
 - Enabling disaggregated serving, MTP, and the overlap scheduler at the same time can lead to accuracy problems.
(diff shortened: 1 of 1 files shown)


zhuolingwang (approved):

litaotju (approved):

laikhtewari:
@QiJune @juney-nvidia can we amend the comment to specify which scenarios are affected by the perf issue (ie specifically for seq Len <8k)

zhuolingwang:
In 0.21, full chunked attention support has been added to make sure LLaMA4 model can functionally run with > 8K seq length, while there is a known performance regression(only affect LLaMA4 model) due to this functional enhancement. The root cause of the regression has been identified already and the fix will be part of the future release

@laikhtewari  Are you OK with above statement or need to add more words?

QiJune:
cc @nv-yilinf  to comment on the known issue.

PR #6272, opened 2025-07-22
[None][feat] Add PyTorch Runtime Support for MoE Weight Prefetching

nvxuanyuc:
## Description

**Overview**
This PR introduces PyTorch runtime support for MoE model weight prefetching, primarily targeting:

- Maximum throughput scenarios during the prefill stage in disaggregated serving
- Small-scale Large MoE model testing and deployment, where GPU memory is constrained

Currently supported models include DeepSeek, LLaMa 4, and Mixtral, with minimal effort required to extend to other MoE models.

**Background**
Previously, the full set of MoE layer weights had to be stored in device memory, placing significant pressure on available memory — especially for models with hundreds of experts per layer. This PR introduces an alternative: keeping only a limited number of MoE layers in device memory, and dynamically prefetching weights from host memory on demand. 

When performance is a critical concern, prefetching is recommended only during the prefill phase, where memory copy latency can be effectively overlapped with computation. This makes the approach well-suited for disaggregated serving scenarios.

### Implementation Details

**Core Concepts**
- moe_prefetch_capacity: Number of MoE layers whose weights can be buffered concurrently in GPU memory.
    - Layers with the same prefetch_layer_idx % moe_prefetch_depth share the same devive-side weight buffer.
- moe_prefeth_stride: Controls which MoE layers enable weight offloading.
    - Enables sparse prefetching, where one layer’s weights are offloaded every stride layers to improve overlap between H2D memory ops and compute.

**Prefetching Mechanism**
- A full copy of MoE layer weights is maintained in host memory.
- A fixed-size device weight buffer is allocated with size: ‘moe_prefetch_depth x single_layer_moe_weight_size’.
- At runtime, we perform host-to-device memcpy to prefetch weights in advance, overlapping it with preceding layer computations.
    - Specifically, layer i triggers prefetching for layer ‘i+moe_prefetch_stride x moe_prefetch_depth’ once its MoE GEMM operations complete.

**Design Architecture**
- A global moe_prefetch_manager orchestrates:
    - Device buffer allocation
    - Shared host-to-device weight mapping across layers
- Each prefetched MoE layer uses a moe_prefetch_proxy, which:
    - Interfaces with the manager
    - Launches the actual prefetch operation

This separation improves modularity and enables easy future extensions (e.g. ratio-based or priority-based prefetching strategies).

**MoE Backend Support**
- Currently enabled only in the CUTLASS and TRTLLM-GEN MoE backend, where max-throughput is focus.
- Fully compatible with fp8 and nvfp4 quantization modes and tensor/expert/mixed parallelism.
- Pipeline parallelism is not supported, as it addresses memory pressure by scaling out across more devices — which contrasts with the core performance advantage of prefetching — achieving comparable throughput with fewer resources.

The implementation is designed to be extensible with minimal integration overhead. For reference, Mixtral 8x7B is a simple and direct example.

**Brief Performance Results**

- Enabling prefetching for the same workload on the same resources results in slight performance degradation, primarily due to 20-30% slowdown in NCCL AllReduce when overlapped with concurrent H2D memcpy. Even at a doubled maximum batch size enabled by prefetching compared to the original limit, the same configuration fails to improve throughput.
- Significant **per-GPU throughput improvements** are observed when running the **same workload (prefill-only) using fewer GPU resources**:
    - DeepSeek R1 [fp8] on H200: tep4 w/ prefetching achieves 1.35x higher per-gpu output throughput (tps/gpu) compared to tep8 w/o prefetching at ISL=2k.
    - Llama 4 maverick [fp8] on H20: tep4 w/ prefetching achieves 1.29x higher per-gpu output throughput (tps/gpu) compared to tep8 w/o prefetching at ISL=4k.

## Test Coverage

The prefetching logic is integrated within the fused MoE module, and model-side integration only requires setting the appropriate prefetching config values. To avoid redundancy, we validate correctness using only the DeepSeek-V3-Lite integration test in the current pipeline.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

## Summary by CodeRabbit

* **New Features**
  * Introduced Mixture of Experts (MoE) weight prefetching support for improved GPU memory efficiency and performance in supported transformer models.
  * Added new configuration options for enabling and tuning MoE prefetching, including prefetch depth and stride.
  * Integrated MoE prefetching into DeepSeekV3, Llama4, and Mixtral models, with support for asynchronous weight loading during inference.

* **Tests**
  * Added integration tests to validate MoE weight prefetching functionality across various quantization types and parallelism settings.

* **Chores**
  * Extended configuration and argument parsing to support MoE prefetching parameters.


tensorrt_llm/_torch/modules/fused_moe/quantization.py
```diff
@@ -174,16 +174,29 @@ def create_weights(
         w3_w1_bias_shape: Optional[tuple[int, int]] = None,
         w2_bias_shape: Optional[tuple[int, int]] = None,
     ):
-        # Fused gate_up_proj (column parallel)
-        w3_w1_weight = nn.Parameter(torch.empty(w3_w1_weight_shape,
-                                                dtype=weight_dtype),
-                                    requires_grad=False)
-        module.register_parameter("w3_w1_weight", w3_w1_weight)
+        if module.use_prefetch:
```
djns99, line 177: Should we handle scaling factors for NVFP4 too?
I suppose we can handle this in a future MR, as long as this design accomodates it

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+
+    def __init__(self, num_hidden_layers: int, moe_layer_freq: int,
+                 add_one_to_layer_idx: bool, first_k_dense_replace: int,
+                 moe_prefetch_depth: int, moe_prefetch_stride: int):
+        # tp and ep support only
+        self.prefetch_depth = moe_prefetch_depth  # of buffers
+        self.prefetch_stride = moe_prefetch_stride  # prefetch every [stride] moe layers
+        self.weights = {}
+        # set on first register_weight call
```
djns99, line 19: Have you tested with any of the auto quant modes from model opt that dynamically quantize some layers and not others?
I'm assuming this won't support that currently as it expects the buffers to be the same size, but maybe something to consider

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
```diff
@@ -0,0 +1,183 @@
...
+        self.weight_dtype = None
+        self.weight_shapes = None
+        self.device_weight_buffers = []
+
+        # moe layers indices
+        moe_layer_offset = 1 if add_one_to_layer_idx else 0
+        self.moe_layer_indices = [
```
djns99, line 26: Would taking this list as a parameter be a more flexible API. This API seems a little convoluted

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def register_weight(self, layer_id: int, weights: List[torch.Tensor]):
+        assert len(weights) == 2, "Experted two weight tensors per moe layer"
+
+        if self.weight_dtype is None or self.weight_shapes is None:
+            self.weight_dtype = weights[0].dtype
+            self.weight_shapes = {
+                "w3_w1_weight": weights[0].shape,
+                "w2_weight": weights[1].shape
+            }
```
djns99, line 70: Can we have a list of weight_names instead of duplicating these everywhere

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
```diff
@@ -0,0 +1,183 @@
...
+                torch.empty(w3_w1_shape,
+                            dtype=self.weight_dtype,
+                            device=torch.device("cuda")),
+                "w2_weight":
+                torch.empty(w2_shape,
+                            dtype=self.weight_dtype,
+                            device=torch.device("cuda"))
+            })
+
+    def prefetch_weights(self, cur_stream: torch.cuda.Stream):
```
djns99, line 101: This name is not clear that it only works for the initial layers. Maybe `prefetch_initial_weights`?

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
```diff
@@ -0,0 +1,183 @@
...
+        buffer_idx = self.layer_to_buffer[layer_idx]
+        return self.prefetch_streams[buffer_idx]
+
+    def get_dst_device_buffers(self, layer_idx: int):
+        assert layer_idx in self.layer_to_buffer, f"Layer {layer_idx} not found in registered prefteched layers"
+        buffer_idx = self.layer_to_buffer[layer_idx]
+        return self.device_weight_buffers[buffer_idx][
+            'w3_w1_weight'], self.device_weight_buffers[buffer_idx]['w2_weight']
+
+    def get_buffer_n_source(self, layer_idx: int):
```
djns99, line 130: This name is unclear to me. Should it be `get_next_prefetch_buffers()` maybe?

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
```diff
@@ -0,0 +1,183 @@
...
+    def __init__(self, layer_id: int, prefetch_manager: MoEPrefetchManager):
+        self.layer_id = layer_id
+        self.prefetch_manager = prefetch_manager
+        # set when creating weights
+        self.w3_w1_weight_shape = None
+        self.w2_weight_shape = None
+        self.weight_dtype = None
+        self.w3_w1_dst_buffer = None
+        self.w2_dst_buffer = None
+        # set first time when retrieving from prefetch manager
```
djns99, line 158: This comment is unclear to me

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
```diff
@@ -0,0 +1,183 @@
+from typing import List, Tuple
+
+import torch
+
+
+class MoEPrefetchManager:
```
djns99, line 6: I don't see anything here that is unique to MOE (i.e no JIT loading). Will it be easy to generalise this to support regular dense layers too?

djns99 (approved):
Excellent work! This is a very elegant implementation and seems relatively easy to use. I have left a few suggestions of minor improvements, but I don't think any of these need to block this MR going in as a version 1.

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
nvxuanyuc: The current design doesn’t support variable-sized buffers. A possible approach is to size the shared buffer to the largest weight among the layers that share it, and then give each layer a tensor view with its exact shape backed by that buffer.

tensorrt_llm/_torch/modules/fused_moe/quantization.py
nvxuanyuc: After discussion with @symphonylyh, we think that it’s only necessary if we run into a case where the model just slightly exceeds total device memory. So far we haven’t seen such cases. Offloading those scaling factors doesn’t save much memory but does add extra copy overhead.

tensorrt_llm/_torch/modules/fused_moe/moe_prefetch_manager.py
nvxuanyuc: [Done](https://github.com/NVIDIA/TensorRT-LLM/pull/6272/files#diff-7e749285b2fd0d991aed6bb062c64933ac93d96bf7919482cd342f42af706078R94).

tensorrt_llm/_torch/modules/fused_moe/quantization.py
djns99: Scaling factors are 1/8 the size of the weights for NVFP4, this still seems decently significant. If we have 64 layers that is ~8 layers worth of weights that could be pinned to the GPU instead.
We can concatenate them together into one buffer if copy overhead is a concern. But I agree it is not our top priority

symphonylyh (commented):
in the PR description and the code comments, I think we need to better explain the depth and stride ideas to the users because these are unfamiliar concepts invented by us

Some naming disucssions:

1. I think prefetch_depth is a confusing term, is prefetch_budget a better name? and can you elaborate more on the PR description to make sure the audience can understand the core concepts
2. I think prefetch has been a anti-intuitive term, can we change it all to MoE offloading? offloading is a feature which a user can immediately understand that MoE weights are offloaded to CPU, whereas prefetching is a implementation detail that they don't need to know. thoughts? @djns99 
3. would be nice if we can attach any diagram to showcase how depth/stride means & how offload/prefetch works. can be added after the merge though

symphonylyh (approved):

tensorrt_llm/_torch/model_config.py
```diff
@@ -61,6 +61,12 @@ def get_layer_initial_global_assignments(self, layer_idx: int) -> List[int]:
             return None
 
 
+@dataclass
+class MoEPrefetchConfig:
+    prefetch_depth: int = 2
+    prefetch_stride: int = 1
```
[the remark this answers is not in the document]
nvxuanyuc: Detailed explanations are provided in [`MoeConfig`](https://github.com/NVIDIA/TensorRT-LLM/pull/6272/files#diff-bf140a489175e4c6e5931cd19cc57e06c1634483517f06c4e6fda96d5d26b056R201). `MoEPrefetchConfig` serves as a wrapper passed to model based on `MoeConfig`. It's not meat to be set directly, instead, should be constructed from provided settings.

`prefetch_depth` (renaming as `prefetch_capacity)` can be any integer value. While double buffer (i.e. depth 2) typically delivers the best performance (thus used as default), it is not mandatory.

examples/llm-api/quickstart_advanced.py
```diff
@@ -111,6 +111,13 @@ def add_llm_args(parser):
     parser.add_argument('--apply_chat_template',
                         default=False,
                         action='store_true')
+    parser.add_argument(
+        '--use_moe_prefetch',
+        default=False,
+        action='store_true',
+        help='Enable MoE weight prefetching to reduce GPU memory usage')
+    parser.add_argument('--moe_prefetch_depth', type=int, default=None)
+    parser.add_argument('--moe_prefetch_stride', type=int, default=None)
```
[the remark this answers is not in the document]
nvxuanyuc: [Added](https://github.com/NVIDIA/TensorRT-LLM/pull/6272/files#diff-18e6a4fa610461f95f42816b266f1872e94fd0c98ffdf77c2d271343e622a579R124).

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -255,6 +255,48 @@ class MoeConfig(StrictBaseModel):
         "Use low precision combine in MoE operations (only for NVFP4 quantization). When enabled, uses lower precision for combining expert outputs to improve performance."
     )
 
+    # moe weight prefetching config
```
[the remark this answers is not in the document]
nvxuanyuc: Changed.

Superjomn (approved):
LGTM on the llmapi changes.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -279,6 +279,48 @@ class MoeConfig(StrictBaseModel):
...
+    use_prefetch: bool = Field(default=False,
```
QiJune: `use_prefetch` can be seen as `prefetch_config is not None`. Let's remove this field.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -279,6 +279,48 @@ class MoeConfig(StrictBaseModel):
...
+        "Number of MoE layers (weights) that can be buffered concurrently on device"
+    )
+
+    prefetch_stride: Optional[int] = Field(
+        default=None,
+        description=
+        "Stride of MoE prefetching layers -- offload weights for one layer every stride layers"
+    )
+
+    prefetch_config: Optional[object] = Field(
```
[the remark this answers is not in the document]
nvxuanyuc: Thanks for the suggestion. Have refactored all fields into a [single config](https://github.com/NVIDIA/TensorRT-LLM/pull/6272/files#diff-bf140a489175e4c6e5931cd19cc57e06c1634483517f06c4e6fda96d5d26b056R282) now.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -2682,6 +2724,16 @@ def validate_load_balancer(self) -> 'TorchLlmArgs':
                 ) from e
         return self
 
+    @model_validator(mode="after")
+    def validate_moe_prefetch_config(self):
+        from .._torch.model_config import MoEPrefetchConfig
+        if self.moe_config.use_prefetch and self.moe_config.prefetch_config is None:
```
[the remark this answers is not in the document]
nvxuanyuc: As @djns99 says, this feature is mainly targeting the prefill stage, particularly in disaggregated serving scenarios. Have also updated the naming to offloading to make it more intuitive.

tensorrt_llm/_torch/model_config.py
```diff
@@ -75,6 +75,26 @@ def get_layer_initial_global_assignments(self, layer_idx: int) -> List[int]:
             return None
 
 
+@dataclass
+class MoEOffloadConfig:
```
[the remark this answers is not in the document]
nvxuanyuc: Renamed.

QiJune (approved):
LGTM

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -191,6 +191,34 @@ class MoeConfig(StrictBaseModel):
...
+        default=False, description="Whether to use MoE prefetching.")
+
+    moe_prefetch_depth: Optional[int] = Field(
+        default=None,
+        description="The depth of MoE prefetching device buffers.")
+
+    moe_prefetch_stride: Optional[int] = Field(
+        default=None, description="The stride of MoE prefetching layers.")
+
+    moe_prefetch_config: Optional[object] = Field(
```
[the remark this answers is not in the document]
nvxuanyuc: A usage example has been added in [quickstart_advanced.py](https://github.com/NVIDIA/TensorRT-LLM/pull/6272/files#diff-18e6a4fa610461f95f42816b266f1872e94fd0c98ffdf77c2d271343e622a579R258). The default config should git most use cases.

PR #6273, opened 2025-07-23
fix: Tuning method for VSWA models in `trtllm-bench`

jaedeok-nvidia:
This PR corrects the calculation of required KV cache size in case of sliding window enabled. It may affect actual max_batch_size and max_num_tokens, affecting throughput performance. 

## Summary by CodeRabbit

* **New Features**
  * Added support for sliding window and variable sliding window attention configurations in model settings.
  * Enhanced calculation of KV cache memory requirements for models using sliding window attention, enabling more accurate resource estimation.

* **Bug Fixes**
  * Improved accuracy of maximum batch size and token count calculations for different attention models.

* **Tests**
  * Introduced comprehensive unit tests to validate tuning logic and ensure consistency across various model configurations.

### DIFF
--- a/tensorrt_llm/bench/build/dataclasses.py
+++ b/tensorrt_llm/bench/build/dataclasses.py
@@ -1,5 +1,5 @@
 from transformers import AutoConfig
-from typing import Optional, Literal
+from typing import Optional, Literal, List
 from pydantic import AliasPath, BaseModel, Field, AliasChoices, model_validator
 import huggingface_hub
 from huggingface_hub.constants import (
@@ -152,6 +152,8 @@ class ModelConfig(BaseModel):
                                          "attention_head_dim",
                                          AliasPath("text_config", "head_dim"),
                                      ))
+    sliding_window: Optional[int] = Field(default=None)
+    sliding_window_pattern: Optional[int | str] = Field(default=None)
     max_position_embeddings: Optional[int] = Field(
         default=None,
         validation_alias=AliasChoices(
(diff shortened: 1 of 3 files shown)


yechank-nvidia (approved):
LGTM. Thx for the work! One question below.

FrankD412:
I'm looking over this PR, and that's a lot of changes to the math. I'm chatting with @venkywonka about the tuning heuristic returning different values in 1.0 vs 0.20 -- we think that the bug is related to support for Nemotron-H in the tuning heuristic adding some new math.

I'm hesitant to approve this just yet as we already have an instance of the math changing unintentionally in a prior PR -- I see that you have added tests which I appreciate.

If we are going to keep augmenting this area of the code, I think we need a more modular approach than detecting if specific properties are returned with non-None values. The more modular approach would be more unit-testable and would hopefully isolate other model types. I need to review this PR further, but I welcome suggestions about how to make this more modular.

venkywonka:
This line: 
```
gb_per_token = 2 * model_config.num_attention_layers * adjusted_num_kv_heads \
        * model_config.head_size * byte_per_kv_elem / (1024 ** 3)
```

is actually causing perf regressions (as the tuning is conservative and calculates much lesser `max_bs` and `max_nt` than it used to). 
It was previously:
```
    byte_per_token = 2 * model_config.num_hidden_layers * adjusted_num_kv_heads \
        * model_config.head_size * byte_per_kv_elem / (1024 ** 3)
```

- which is something we should anyways revert for non-`NemotronHybird` models. 

This PR adds complications on top of that, unfortunately...

as @FrankD412 suggested, this is a hotspot that could lead to regressions in default behaviors of many existing workflows if unverified - 

@jaedeok-nvidia  does these changes gaurantee that the tuning math surgically only is specific to VSWA models and doen't have side-effects on others?

PR #6293, opened 2025-07-23
test: skip llama3.3 70b test on cg4

xinhe-nv:
skil llama3.3 70B test on CG4

## Summary by CodeRabbit

* **Tests**
  * Improved test stability by skipping tests on systems with less than 1TB host memory.
  * Updated device compatibility markers for specific test methods.
  * Enhanced GPU memory management during testing for better resource utilization.


tests/integration/defs/accuracy/test_llm_api_pytorch.py
```diff
@@ -339,6 +339,7 @@ def test_fp8_prequantized(self):
 
 
 @pytest.mark.timeout(7200)
+@pytest.mark.skip_by_host_memory(1000000)
```
yuxianq, line 342: Why we need 1TB host memory to run a 70B model?

xinhe-nv: 1TB  is basic requirement for our large model tests environment. CG4 120G only has 800G host memory, and 480G is shared with GPUs. the test will cause the system crash. we have to use this condition to skip the test on CG4.

yuxianq (approved):

PR #6334, opened 2025-07-24
[TRTLLM-4921][feat] Enable chunked prefill for Nemotron-H

tomeras91:
## Description

This PR adds chunked prefill support for Nemotron-H. This is done by using the `initial_states` option of the mamba2 prefill kernels (conv and SSM). This should allow running extremely long sequence lengths without exploding activation memory.

The option to use `initial_states` was already implemented, but included some bugs in the mamba prefill kernels, manifested in cases where the prefix chunking size (max_num_tokens) wasn't a multiple of the mamba chunk size:
1. In the state-passing kernel (`_state_passing_fwd_kernel`), the dA_cumsum was computed from the start of the mamba chunk instead of from the start of the current sequence
2. In the chunked scan kernel (`_chunk_scan_fwd_kernel`), the dA_cumsum_boundary was loaded from a different memory address in each block (in the chunk_size dimension) even though it should be loaded from a constant location.

The PR includes fixes for these 2 issues.

### Tests
* Added unittest for `torch.ops.trtllm.causal_conv1d_fwd` with initial states
* Added unittest for `mamba_chunk_scan_combined` with initial states
* Added unittest for `mamba_chunk_scan_combined` with initial states for parametrized sequence lengths and batch sizes. This test failed before the kernel bugfixes.
* Added a uniitest for Nemotron-H asserting that results with and without prefix chunking don't change.

### Benchmarks
**MMLU results**
Here are MMLU results for Nemotron-H 8B Base with and without prefix chunking. `max_num_tokens` was set to two different values to test different cases:
1. A low value of 64 to ensure a lot of prompts are chunked into multiple chunks.
2. A higher value of 1000 (which is not a multiple of the mamba chunk size (128 for Nemotron-H 8B)) to ensure batching of multiple prompts and sequence changes in the middle of mamba chunks.

| Version | MMLU Score |
|---------|------------|
| main - no chunking | 69.98 |
| PR - no chunking | 69.98 |
| PR - with chunking (64) | 70.08 | 
| PR - with chunking (1000) | 69.71 | 

**note**: Chunking with `max_num_tokens=1000` is the case that failed before the kernel bugfixes. It had a score of 63.42

## Summary by CodeRabbit

* **New Features**
  * Added support for chunked prefill requests with initial state handling, enabling more precise and efficient processing of long prompts and sequences.
  * Enhanced model and kernel logic to compute and manage chunk boundaries and initial states during chunked sequence processing.

* **Bug Fixes**
  * Improved accuracy and correctness of state passing and cumulative sum calculations at chunk boundaries during continuous batching.

* **Tests**
  * Introduced new and extended tests to validate chunked prefill, initial state handling, and chunked prefix processing, ensuring output consistency and correctness across modes.

* **Chores**
  * Updated internal logic and attributes to support chunked processing and initial state management in relevant modules.


tensorrt_llm/_torch/modules/mamba/mamba2_metadata.py
```diff
@@ -13,15 +13,74 @@
...
+import math
+from typing import Tuple
+
 import torch
 
 from tensorrt_llm._torch.attention_backend.interface import AttentionMetadata
 
 
+def cu_seqlens_to_chunk_indices_offsets(
+        cu_seqlens: torch.Tensor,
```
shaharmor98, line 25: It's a great documentation, I think that adding another sentence per parameter would also help (i.e., add docs on cu_seqlens, and chunk_size).  
Given my lack of context, the fact that you jumped to the implementation (by mentioning chunk_indices for example) confused me a bit.

tensorrt_llm/_torch/modules/mamba/mamba2_metadata.py
```diff
@@ -44,4 +112,17 @@ def prepare(self, attn_metadata: AttentionMetadata):
                              dtype=torch.int,
                              device=self.cu_seqlens.device),
                 repeats=context_lens,
-                output_size=self.cu_seqlens[num_contexts]).unsqueeze(0)
+                output_size=num_ctx_tokens).unsqueeze(0)
+
+            num_cached_tokens_per_seq = attn_metadata.kv_cache_params.num_cached_tokens_per_seq
```
shaharmor98, line 117: I haven't seen any related changes in this PR, and I'm not aware of the greater concepts, but the default value of `num_cached_tokens_per_seq` is None, is it guaranteed that it won't be None here when you use it?

tests/unittest/_torch/modeling/test_modeling_nemotron_h.py
```diff
@@ -336,3 +340,62 @@ def test_nemotron_h_cuda_graph_overlap_scheduler():
             msg=lambda x:
             f"Prompt {i}: with/without overlap scheduler (with CG) logprobs for all selected tokens {x}"
         )
+
+
+def test_nemotron_h_chunked_prefill():
+    # Long prompts (~100 tokens) to make sure chunked prefill is enabled
+    # (At the time of development, tokens_per_block isn't configurable from the LLM API,
+    # and max_tokens (i.e. chunk size) needs to be a multiple of tokens_per_block)
+    prompts = [
```
shaharmor98, line 349: And also, I assume you didn't thought of these prompts by yourself.  
If they are copied, I think that (not in terms of copyrighting) mentioning this could be useful.

tests/unittest/_torch/modeling/test_modeling_nemotron_h.py
```diff
@@ -336,3 +340,62 @@ def test_nemotron_h_cuda_graph_overlap_scheduler():
...
+        assert output.outputs[0].text == chunked_prefill_output.outputs[0].text
+
+        # assert same prefill logprobs. Same atol as diff between mcore and initial impl
+        prefill_logprobs = extract_prefill_logprobs(output)
+        chunked_prefill_logprobs = extract_prefill_logprobs(
+            chunked_prefill_output)
+        torch.testing.assert_close(
+            prefill_logprobs,
+            chunked_prefill_logprobs,
+            atol=0.3,
```
shaharmor98, line 388: out of curiosity, how did you choose these atol and rtol?

tomeras91: Same as it is for the correctness test comparing logprobs of TRTLLM vs those of the initial Mcore implementation

tests/unittest/_torch/modeling/test_modeling_nemotron_h.py
tomeras91: I don't exactly remember, but I think they are AI generated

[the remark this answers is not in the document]
tomeras91: I see your point, but I disagree. It's only 5 prompts, and I think adding another file to store them actually clutters the code more.
If this was a longer list, or we would've used the same prompts in many test, I agree it's better to move it elsewhere.

tensorrt_llm/_torch/modules/mamba/mamba2_metadata.py
tomeras91: Done

shaharmor98 (approved):
LGTM

danielafrimi (approved):

PR #6346, opened 2025-07-25
[None][fix] Fix KV event consumption

jthomson04:
This MR fixes two related issues in KV Cache Event handling.

1. There's a time.sleep(0.2) in the workers which emit events on the queue. This is fine for iteration stats, but very bad for kv event consumption. This adds an optional `urgency` argument to the `_iteration_result_task`. We retain the 0.2s sleep for iteration stats, but not for kv events. 
2. KV Cache Events are started up on all MPI ranks, even though we only consume them from rank 0. This causes warnings to be emitted by all non-zero ranks.

## Summary by CodeRabbit

  * Reduced unnecessary delays for urgent tasks so high-priority operations process faster.
  * Avoided needless locking and work when there are no events to flush, improving responsiveness.
  * Limited event-manager initialization to a single processing rank in distributed runs to reduce overhead.


jthomson04:
Can I please get a review on this?

laikhtewari:
@Superjomn @juney-nvidia can you please take a look at merging this PR in?

pcastonguay (approved):

PR #6368, opened 2025-07-25
chore: Improve the AutoTuner log information.

hyukn:
* Change the fallback alert from DEBUG to WARNING level and only do it once.
* Add debug information for profiling cache right after the warmup phase.
* Change the level of exception message during tactic profiling from ERROR to WARNING level. All exception details are pushed to the DEBUG level.
* Other trivial refinements and cleanups.

## Summary by CodeRabbit

* **New Features**
  * Added a method to display the contents of the profiling cache for debugging purposes.

* **Improvements**
  * Enhanced logging for cache misses, profiling failures, and tuning errors to provide clearer and more actionable information.
  * Reformatted and clarified log messages related to profiling and autotuner cache status.
  * Added detailed warnings and summaries to assist users in understanding tuning and inference issues.
  * Standardized autotuner cache size reporting and included profiling cache details after warmup.

### DIFF
--- a/tensorrt_llm/_torch/autotuner.py
+++ b/tensorrt_llm/_torch/autotuner.py
@@ -197,13 +197,13 @@ def autotune(tune_mode: bool = True):
     AutoTuner.get().is_tuning_mode = tune_mode
     autotune_enabled = tune_mode and not old_mode
     if autotune_enabled:
-        logger.info("[Autotuner]: Autotuning process starts ...")
+        logger.info("[Autotuner] Autotuning process starts ...")
     try:
         yield
     finally:
         AutoTuner.get().is_tuning_mode = old_mode
         if autotune_enabled:
-            logger.info("[Autotuner]: Autotuning process ends")
+            logger.info("[Autotuner] Autotuning process ends")
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/autotuner.py
```diff
@@ -426,7 +422,11 @@ def choose_one(self, custom_op: str, runners: List[TunableRunner],
                         custom_op] = self.stats.tuned_op_successful_configs.get(
                             custom_op, 0) + 1
                     logger.debug(
-                        f"[Autotuner]: profiling chosen runner: {runners[runner_id]} {tactic} for {cache_key}"
+                        f"[Autotuner]: Profiling runner={runners[runner_id]}, tactic={tactic} for cache_key={cache_key}"
+                    )
+                else:
+                    logger.error(
+                        f"[Autotuner]: No valid runner/tactic was found for custom_op={custom_op}, input_shapes={input_shapes}. At least one valid (runner, tactic) pair is required. Please check the implementation of the runner for the custom_op."
```
djns99, line 429: Are warnings on by default? If not we should also indicate that the user should increase the log level to see the details of why it failed

djns99 (approved):

tensorrt_llm/_torch/autotuner.py
djns99: > I will change the log level for the failed tuning tactics to 'ERROR' to indicate that failures occur.

I thought the whole point of this change was to NOT do that because it is only a warning, this is not an error. The only error is if none of them succeed

hyukn: After a second thought, I changed it to this way:
* In the ERROR level, only display one line showing how many tactic tuning failures occur for each custom op.
* In the WARNING/DEBUG level, display details of the failed tactics.

tensorrt_llm/_torch/autotuner.py
```diff
@@ -426,8 +425,24 @@ def choose_one(self, custom_op: str, runners: List[TunableRunner],
                         custom_op] = self.stats.tuned_op_successful_configs.get(
                             custom_op, 0) + 1
                     logger.debug(
-                        f"[Autotuner]: profiling chosen runner: {runners[runner_id]} {tactic} for {cache_key}"
+                        f"[Autotuner] Profiling runner={runners[runner_id]}, tactic={tactic} for cache_key={cache_key}."
                     )
+                else:
+                    logger.warning(
```
djns99, line 431: This should definitely be an error, potentially even an exception. It is much more severe an issue than the case below when only some have failed

hyukn: Because we have some cases that allowed `get_valid_tactics` to return an empty list, and the fallback tactic is not automatically included in this part. Then it could be a false alarm if some profile is determined to be invalid (to be filtered out of the cache). Therefore, I make a WARNING level here.

litaotju (approved):

tensorrt_llm/_torch/autotuner.py
hyukn: I agree. I will change this [line](https://github.com/NVIDIA/TensorRT-LLM/pull/6368/files#diff-32958195044352737fb712f99a390c7a236d2ebb6fc5d691d03e5c70541065beR440:~:text=if%20new_tuning_failure_occured%3A-,logger.error(,-f%22%5BAutotuner%5D%20New) as a warning to avoid confusion.

PR #6447, opened 2025-07-29
[nvbug 5380101][fix] Fix nemotronNAS loading for TP>1

tomeras91:
## Summary by CodeRabbit

* **New Features**
  * Improved support for distinguishing between key-value heads and attention heads in model configurations.

* **Bug Fixes**
  * Simplified internal logic for handling head dimensions and key-value head duplication in model weight mapping.

* **Tests**
  * Re-enabled a previously skipped integration test for automatic data type handling.

## Description

This PR fixes weight loading for NemotronNAS with TP>1.

The issue was in layers where the number of KV heads is smaller than TP size. In this case, KV heads should be duplicated to match the TP size, so each TP rank can hold a full KV head. The logic to call `_duplicate_kv` where `_num_kv_heads` is a list was flawed.

The fix is to compute the number of KV heads from the shape of the K or V weight and the head dimension. The list logic that was present in `_duplicate_kv_weights` was added in a [PR](https://github.com/NVIDIA/TensorRT-LLM/pull/5459) relevant only for Qwen3MOE, so it is now encapsulated only in `Qwen3MoeHfWeightMapper`.

Other than this, this PR also fixes the type hint for `BaseWeightMapper._config`.

## Test Coverage

`test_llm_api_pytorch.TestNemotronNas.test_auto_dtype_tp8` now passes


tensorrt_llm/_torch/models/checkpoints/hf/weight_mapper.py
```diff
@@ -59,17 +59,14 @@ def should_skip_module(self, module_name: str) -> bool:
     def _duplicate_kv_weights(self, module: nn.Module, new_name: str,
                               weights: dict):
         if new_name in ['k_proj', 'v_proj']:
-            num_kv_heads_list = [self._num_kv_heads
-                                 ] * len(weights) if isinstance(
-                                     self._num_kv_heads,
-                                     int) else self._num_kv_heads
+            num_kv_heads = weights['weight'].shape[0] // self._head_dim
```
shaharmor98, line 62: Not a bug, but could you explain `weights['weight'].shape[0]` ?

shaharmor98 (approved):
Short and sweet.  
One little ask, approved anyway

tensorrt_llm/_torch/models/checkpoints/hf/weight_mapper.py
tomeras91: Added a comment

PR #6475, opened 2025-07-30
[TRTLLM-6222][feat] NVFP4 Quantization for DeepSeek-R1 MTP and Attention 

binghanc:
## Description

This PR only modifies `tensorrt_llm/_torch/models/modeling_deepseekv3.py` to support NVFP4 quantization for DeepSeek-R1's MTP and Attention.

Specifically, this PR supports NVFP4 quantization for all linear layers in Attention, but the BMM is still BF16. The quantization configs of MTP are consistent with DeepSeek-R1.

## Summary by CodeRabbit

* **New Features**
  * Added support for NVFP4 quantization format when loading and initializing DeepseekV3 model weights.

* **Improvements**
  * Enhanced weight loading logic to handle NVFP4-specific scaling and requantization for attention projection weights, ensuring compatibility with new quantization formats.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_deepseekv3.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv3.py
@@ -45,6 +45,9 @@ from tensorrt_llm.llmapi.utils import enable_llm_debug
 from tensorrt_llm.mapping import Mapping
 from tensorrt_llm.models.modeling_utils import QuantConfig
 
+import tensorrt_llm.quantization.utils.fp4_utils as fp4_utils
+
+
 from ..attention_backend import AttentionMetadata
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -45,6 +45,9 @@
 from tensorrt_llm.mapping import Mapping
 from tensorrt_llm.models.modeling_utils import QuantConfig
 
+import tensorrt_llm.quantization.utils.fp4_utils as fp4_utils
```
kaiyux, line 48: Since this is using @binghanc 's locally exported checkpoint, we need to converge the checkpoint related discussion before merging this PR.

Will follow up with @Kefeng-Duan @binghanc @FelixXidddd .

PR #6369, opened 2025-07-25
feat: Add support for disaggregation with pp with pytorch backend

pcastonguay:
## Summary by CodeRabbit

* **New Features**
  * Enhanced integration tests for disaggregated model serving with expanded support for various tensor and pipeline parallelism configurations.
  * Added new YAML configuration files to enable flexible context and generation server setups.
  * Introduced dynamic retrieval of accuracy task classes by name in integration tests.

* **Bug Fixes**
  * Improved error handling when removing files during build processes.

* **Tests**
  * Added comprehensive parameterized tests covering symmetric and asymmetric parallelism scenarios for disaggregated serving.
  * Extended test coverage with new test functions and configurations for TinyLlama-1.1B-Chat-v1.0.

* **Chores**
  * Improved logging for configuration mismatches and KV cache handling in pipeline parallel execution.

* **Refactor**
  * Internal enhancements to resource management for key-value cache parallelism tracking.

## GitHub Bot Help

`/bot [-h] ['run', 'kill', 'skip', 'reuse-pipeline'] ...`

Provide a user friendly way for developers to interact with a Jenkins server.

Run `/bot [-h|--help]` to print this help message.

See details below for each supported subcommand.

<details>

`run  [--reuse-test (optional)pipeline-id --disable-fail-fast --skip-test --stage-list "A10-PyTorch-1, xxx" --gpu-type "A30, H100_PCIe" --test-backend "pytorch, cpp" --add-multi-gpu-test --only-multi-gpu-test --disable-multi-gpu-test --post-merge --extra-stage "H100_PCIe-TensorRT-Post-Merge-1, xxx" --detailed-log --debug(experimental)]`

Launch build/test pipelines. All previously running jobs will be killed.

`--reuse-test (optional)pipeline-id ` *(OPTIONAL)* : Allow the new pipeline to reuse build artifacts and skip successful test stages from a specified pipeline or the last pipeline if no pipeline-id is indicated. If the Git commit ID has changed, this option will be always ignored. The DEFAULT behavior of the bot is to reuse build artifacts and successful test results from the last pipeline.

`--disable-reuse-test ` *(OPTIONAL)* : Explicitly prevent the pipeline from reusing build artifacts and skipping successful test stages from a previous pipeline. Ensure that all builds and tests are run regardless of previous successes.

`--disable-fail-fast ` *(OPTIONAL)* : Disable fail fast on build/tests/infra failures.

[...]

PR #6489, opened 2025-07-30
[None][fix] disagg ctx pp4 + gen pp4 integ test

raayandhar:
## Summary by CodeRabbit

* **Tests**
  * Updated integration tests to use llama-3.1-8b instead of TinyLlama-1.1B-Chat-v1.0.
  * Enabled previously skipped tests, allowing them to run as part of the test suite.

## Description

Per [this PR](https://github.com/NVIDIA/TensorRT-LLM/pull/6369) for disagg PP support on the PyTorch backend, we faced an issue where the ctx pp4 + gen pp4 integration test would fail.

This failure was because TinyLlama has 22 hidden layers, which is not divisible by 4. Switching this to Llama3-8B (has 32 hidden layers) should fix this. The test is registered under H200s, so the increased space needed for the model size should not be an issue. 

## Test Coverage

n/a

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


Tabrizian (approved):
> This failure was because TinyLlama has 22 hidden layers, which is not divisible by 4.

Is it a generic limitation of PP for AGG or this limitation only exists for disaggregated serving?

pcastonguay (commented):
What was the behavior with Tiny Llama? Would it just hang? If so, could we catch this issue and return a clear error message?

raayandhar:
> What was the behavior with Tiny Llama? Would it just hang? If so, could we catch this issue and return a clear error message?

Just running the test it would timeout. In the actual output logs, we actually get an output message complaining about how we can't divide the layers into the stages, but I think the timeout comes faster than the error propagation. I can look into seeing if we can return a clearer error message earlier. In the non-pytest case this error returns pretty quickly.

tests/integration/defs/disaggregated/test_disaggregated.py
```diff
@@ -616,7 +616,6 @@ def test_disaggregated_ctxpp2_gentp2(disaggregated_test_root, llm_venv,
 def test_disaggregated_ctxtp2pp2_gentp2pp2(disaggregated_test_root, llm_venv,
                                            disaggregated_example_root,
                                            llama_model_root):
-    pytest.skip(f"8 GPU test times out currently, skipping")
```
raayandhar, line 619: This test actually always worked since the issue is related to PP=4 only (cannot do 22/4 and remain int)

tests/integration/defs/disaggregated/test_disaggregated.py
```diff
@@ -631,14 +631,13 @@ def test_disaggregated_ctxtp2pp2_gentp2pp2(disaggregated_test_root, llm_venv,
 
 
 @pytest.mark.skip_less_device(8)
-@pytest.mark.parametrize("llama_model_root", ['Llama-3.1-8B-Instruct'],
-                         indirect=True)
+@pytest.mark.parametrize("llama_model_root", ['llama-3.2-1b'], indirect=True)
```
raayandhar, line 634: This is because in `conftest.py` it lists the actual models and their paths that it will pick up. And I found this smaller model that has 16 layers, so should work with PP=4, but I will verify this once I grab an 8x node.

Tabrizian: If the path is correct, it will be available on all the GPUs that the CI choses.

raayandhar: Perfect, should be good then. Thanks Iman.

Tabrizian (approved):

PR #6500, opened 2025-07-31
Feat/mtp opt 2 lamport allgather

ameynaik-hub:
## Summary by CodeRabbit

* **New Features**
  * Added ALLGATHER fusion operation across runtime, operator, and speculative sampling (MTP) with workspace support and heuristic selection.

* **Bug Fixes**
  * Relaxed parameter validation to allow ALLGATHER without a residual tensor.

* **Tests**
  * Added unit and multi-GPU tests, utilities, and benchmarks covering ALLGATHER correctness and performance.

* **Chores**
  * Updated enums, bindings (Python/nanobind/pybind), workspaces, fake ops, and operator mappings to expose ALLGATHER consistently.

Utilize Lamport-based allgather. A new option has been introduced in Lamport-based allreduce to provide the output after only the gather operation. When employed in MTP module after lm_head and local_argmax ops, the performance of Deepseek R1 BS1 improves by approximately 2 TPS on B200x8. Considering the message size is small, the previously used nccl-based allgather is approximately 2-3 times slower than the newly introduced Lamport-based allgather.


cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -471,6 +472,16 @@ class AllreduceOp
             allreduce_fusion_params.allreduce_out = reduce_out.mutable_data_ptr();
             allreduce_fusion_params.pattern = tensorrt_llm::kernels::ar_fusion::AllReduceFusionPattern::kAllReduce;
         }
+        // Handle AllGather operation
+        else if (mOp == AllReduceFusionOp::ALLGATHER)
```
liji-nv, line 476: I don't think it is a correct way to implement ALLGATHER inside the allreduce op.

liji-nv (changes requested):
From the code, I see you are reusing so many allreduce code. But adding allgather into allreduce breaks semantic of the allreduce Op. This is even worse than a copy paste.

If you really want to share the code, you need to make fundamental refactor to make the common code agnostic to allreduce and allgather.

PR #6512, opened 2025-07-31
refactor: Return the first token ahead of time

Shixiaowei02:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming response by excluding the initial chunk for smoother output.
  * Enhanced context response formatting with cleaner JSON output and removal of sensitive information.

### DIFF
--- a/tensorrt_llm/serve/openai_disagg_server.py
+++ b/tensorrt_llm/serve/openai_disagg_server.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 import asyncio
 import copy
+import json
 import os
 import signal
 import traceback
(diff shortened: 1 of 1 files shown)


Shixiaowei02:
Many thanks to @Tabrizian for the modification. Since @Shunkangz had previously made changes (https://github.com/NVIDIA/TensorRT-LLM/pull/4692) to this, I discussed with him again. During our discussion, we concluded that, due to the current implementation, reducing TTFT can actually increase the time between the first and second tokens. As other frameworks return the first token via generation worker and calculate TTFT based on that, it might be necessary to introduce an option or environment variable to support different interpretations of the TTFT metric.

Tabrizian (approved):

PR #6537, opened 2025-08-01
[https://nvbugs/5394392][fix] Enlarge scheduler capacity under disagg bs == 1

yifeizhang-c:
## Description

Under disagg, if generation server runs with `bs == 1`, the dummy `GENERATION_IN_PROGRESS` request for attention dp will prevent `DISAGG_GENERATION_INIT` from being scheduled, thus letting generation server running in an endless cycle. This PR enlarges scheduler capacity and related resources to be with at least capacity == 2.

Besides, originally `py_executor` logic assign new `SEQ_SLOT` resource on `DISAGG_GEN_INIT` state. This PR delays the assignment to `DISAGG_TRANS_COMPLETE` state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a scheduling stall that could occur with single-sequence generation when using KV cache by adjusting capacity and sampler sizing, improving stability.
  * Prevented initialization-only requests in disaggregated generation from blocking sequence-slot assignment, enhancing throughput and reliability.

* **Tests**
  * Added an integration test for disaggregated generation with batch size 1 (TinyLlama) and included it in the l0_dgx_h100 pre-merge test suite.


Shixiaowei02 (approved):

yifeizhang-c:
/bot run --stage-list "H100_PCIe-PyTorch-1, H100_PCIe-PyTorch-2"

yifeizhang-c:
> @yifeizhang-c Do you mind responding to the question here? [#6537 (comment)](https://github.com/NVIDIA/TensorRT-LLM/pull/6537#issuecomment-3176634742)
> 
> Increasing `max_num_sequences` by 1 when ADP is enabled is like a workaround, no?

Increasing `max_num_sequences` by 1 can be regarded as a workaround.

Sorry for not replying the previous comment in time, as the suggestion indeed will make the code changes more concise, and I was analyzing into it.

To resolve the disagg bs==1 non-stop issue, increasing the `CapacityScheduler`'s capacity to be at least 2 seems to be a must as the active request pool may have both `DISAGG_GENERATION_INIT` and `GENERATION_IN_PROGRESS` requests. If `CapacityScheduler`'s capacity is only 1, then according to request scheduling priority, `GENERATION_IN_PROGRESS` will be scheduled first as it is closer to the final `GENERATION_COMPLETE` status, and thus `DISAGG_GEN_INIT` will be stuck unscheduled forever, and generation workers will never start. `CapacityScheduler` currently cannot distinguish dummy `GENERATION_IN_PROGRESS` from normal `GENERATION_IN_PROGRESS`.

Tabrizian:
Thanks @yifeizhang-c, I think the pytorch runtime doesn't use the C++ batch scheduler (feel free to correct me). I think the reason that `_schedule` function doesn't schedule the generation init requests is that as soon as it hits the max batch size, it would stop looking at the other requests.

In this case, it would just schedule the dummy request and never reach the generation init request since there is a dummy request added in every iteration.

Here is a draft of the scheduler changes: https://github.com/NVIDIA/TensorRT-LLM/compare/main...Tabrizian:TensorRT-LLM:user/imant/schedulerChange?expand=1 It

yifeizhang-c:
> I think the pytorch runtime doesn't use the C++ batch scheduler (feel free to correct me).

Currently `py_executor` uses `BindCapacityScheduler`, which eventually invokes C++ `GuaranteedNoEvictScheduler`.

> I think the reason that `_schedule` function doesn't schedule the generation init requests is that as soon as it hits the max batch size, it would stop looking at the other requests.
> 
> In this case, it would just schedule the dummy request and never reach the generation init request since there is a dummy request added in every iteration.
> 
> Here is a draft of the scheduler changes which I think is needed for this: https://github.com/NVIDIA/TensorRT-LLM/compare/main...Tabrizian:TensorRT-LLM:user/imant/schedulerChange?expand=1 Please let me know if you have any feedback.

I am not clear about the logic behind the implementation of `GuaranteedNoEvictScheduler`. But I wonder whether we can schedule `DISAGG_GEN_INIT` totally separately from `GEN_IN_PROGRESS`. Will there be cases where the (N-1)-th request is still in `GEN_IN_PROGRESS` and we received the N-th request’s `DISAGG_GEN_INIT`, and will scheduling `DISAGG_GEN_INIT` cause trouble in this case?

schetlur-nv (approved):

PR #6557, opened 2025-08-01
[https://nvbugs/5410391][bug] Support to share device buffers in attention meta

HuiGao-NV:
## Summary by CodeRabbit

* **New Features**
  * Runtime buffer APIs added to expose attention-related CPU/GPU buffers and accept supplied buffers during initialization.

* **Improvements**
  * Buffer-caching and reuse to reduce allocations and improve memory reuse/performance.
  * Initialization flow refactored to materialize and reuse provided buffers.
  * Memory-accounting comment clarified for token estimation.

* **Bug Fixes**
  * Safer finalization and clearing of stored metadata buffers to avoid stale-state errors.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/flashinfer.py
+++ b/tensorrt_llm/_torch/attention_backend/flashinfer.py
@@ -167,12 +167,6 @@ class FlashInferAttentionMetadata(AttentionMetadata):
                                                  device='cuda',
                                                  dtype=torch.int)
 
-    def get_runtime_buffers(self):
-        return {}
-
-    def post_init_with_buffers(self, buffers) -> None:
-        self.__post_init__()
-
     def create_cuda_graph_metadata(self,
                                    max_batch_size: int,
                                    sub_cross_metadata: bool = False,
--- a/tensorrt_llm/_torch/attention_backend/flashinfer.py
+++ b/tensorrt_llm/_torch/attention_backend/flashinfer.py
@@ -167,10 +167,17 @@ class FlashInferAttentionMetadata(AttentionMetadata):
                                                  device='cuda',
                                                  dtype=torch.int)
 
+    def get_runtime_buffers(self):
+        return {}
+
+    def post_init_with_buffers(self, buffers) -> None:
+        self.__post_init__()
+
     def create_cuda_graph_metadata(self,
                                    max_batch_size: int,
                                    sub_cross_metadata: bool = False,
-                                   max_draft_tokens: int = 0) -> Self:
+                                   max_draft_tokens: int = 0,
+                                   buffers=None) -> Self:
         metadata = super().create_cuda_graph_metadata(max_batch_size,
                                                       sub_cross_metadata,
                                                       max_draft_tokens)
--- a/tensorrt_llm/_torch/attention_backend/interface.py
+++ b/tensorrt_llm/_torch/attention_backend/interface.py
@@ -137,6 +137,7 @@ class AttentionMetadata:
 
     # This buffer is currently only used for TrtllmAttentionMetadata.
     cache_indirection: Optional[torch.Tensor] = None
+    cuda_graph_buffers: dict[str, list[torch.Tensor]] = None
 
     def __post_init__(self) -> None:
         if self.is_cross:
@@ -295,6 +296,7 @@ class AttentionMetadata:
 
         cuda_graph_metadata = copy.copy(self)
         cuda_graph_metadata.is_cuda_graph = True
+        cuda_graph_metadata.cuda_graph_buffers = buffers
         if self.has_cross_sub_metadata:
             cuda_graph_metadata.cross = cuda_graph_metadata.cross.create_cuda_graph_metadata(
                 max_batch_size, True)
@@ -319,7 +321,7 @@ class AttentionMetadata:
                 )
 
         cuda_graph_metadata.num_contexts = 0
-        cuda_graph_metadata.post_init_with_buffers(buffers)
+        cuda_graph_metadata.__post_init__()
         return cuda_graph_metadata
 
     def update_spec_dec_param(self, is_spec_decoding_enabled, is_spec_dec_tree,
--- a/tensorrt_llm/_torch/attention_backend/interface.py
+++ b/tensorrt_llm/_torch/attention_backend/interface.py
@@ -282,7 +282,8 @@ class AttentionMetadata:
     def create_cuda_graph_metadata(self,
                                    max_batch_size: int,
                                    sub_cross_metadata: bool = False,
-                                   max_draft_tokens: int = 0) -> Self:
+                                   max_draft_tokens: int = 0,
+                                   buffers=None) -> Self:
         """
         Creates metadata for CUDA graph execution.
         CUDA graphs require to use pre-allocated buffers for all tensors in fields.
@@ -318,7 +319,7 @@ class AttentionMetadata:
                 )
 
         cuda_graph_metadata.num_contexts = 0
-        cuda_graph_metadata.__post_init__()
+        cuda_graph_metadata.post_init_with_buffers(buffers)
         return cuda_graph_metadata
 
     def update_spec_dec_param(self, is_spec_decoding_enabled, is_spec_dec_tree,
--- a/tensorrt_llm/_torch/attention_backend/trtllm.py
+++ b/tensorrt_llm/_torch/attention_backend/trtllm.py
@@ -600,9 +600,9 @@ class TrtllmAttentionMetadata(AttentionMetadata):
 
     def __post_init__(self) -> None:
         super().__post_init__()
-        self.__post_init_with_buffers(self.cuda_graph_buffers)
+        self._post_init_with_buffers(self.cuda_graph_buffers)
 
-    def __post_init_with_buffers(self, buffers) -> None:
+    def _post_init_with_buffers(self, buffers) -> None:
 
         # Set a default value, as max_num_sequences is not always set.
         if self.max_num_sequences is None:
@@ -624,8 +624,6 @@ class TrtllmAttentionMetadata(AttentionMetadata):
             Args:
                 tensor_shape: The required shape.
                 dtype: The required dtype.
-                buffers: A dictionary mapping cache names to lists of buffer tensors.
-                        Can be `None` or empty.
                 cache_name: The key for the specific list of buffers to search in.
 
             Returns:
@@ -652,20 +650,26 @@ class TrtllmAttentionMetadata(AttentionMetadata):
 
         def get_empty_like(like_tensor: torch.Tensor,
                            cache_name: str) -> torch.Tensor:
-            return get_empty(like_tensor.shape,
-                             cache_name=cache_name,
-                             dtype=like_tensor.dtype)
+            return get_empty(
+                like_tensor.shape,
+                cache_name=cache_name,
+                dtype=like_tensor.dtype,
+            )
 
-        self.prompt_lens_cuda = get_empty((self.max_num_sequences, ),
-                                          cache_name="prompt_lens_cuda",
-                                          dtype=torch.int)
+        self.prompt_lens_cuda = get_empty(
+            (self.max_num_sequences, ),
+            cache_name="prompt_lens_cuda",
+            dtype=torch.int,
+        )
         self.prompt_lens_cpu = torch.empty_like(
             self.prompt_lens_cuda,
             device='cpu',
             pin_memory=True,
         )
-        self.kv_lens_cuda = get_empty_like(self.prompt_lens_cuda,
-                                           cache_name="kv_lens_cuda")
+        self.kv_lens_cuda = get_empty_like(
+            self.prompt_lens_cuda,
+            cache_name="kv_lens_cuda",
+        )
         self.kv_lens = torch.empty_like(self.kv_lens_cuda,
                                         device='cpu',
                                         pin_memory=True)
@@ -685,7 +689,8 @@ class TrtllmAttentionMetadata(AttentionMetadata):
                     self.kv_cache_manager.max_blocks_per_seq
                 ],
                 cache_name="kv_cache_block_offsets",
-                dtype=torch.int32)
+                dtype=torch.int32,
+            )
             self.host_kv_cache_block_offsets = torch.empty_like(
                 self.kv_cache_block_offsets,
                 device='cpu',
@@ -700,20 +705,23 @@ class TrtllmAttentionMetadata(AttentionMetadata):
                         self.kv_cache_manager.max_blocks_per_seq
                     ],
                     cache_name="block_ids_per_seq",
-                    dtype=torch.int32)
+                    dtype=torch.int32,
+                )
                 self.kv_block_ids_per_seq = get_empty(
                     [
                         self.kv_cache_manager.max_batch_size,
                         self.kv_cache_manager.max_blocks_per_seq
                     ],
                     cache_name="kv_block_ids_per_seq",
-                    dtype=torch.int32)
+                    dtype=torch.int32,
+                )
             if self.enable_paged_context_mla:
                 # for kv cache reuse/chunked context in MLA
                 self.ctx_cached_token_indptr = get_empty(
                     (self.max_num_requests + 1, ),
                     cache_name="ctx_cached_token_indptr",
-                    dtype=torch.int64)
+                    dtype=torch.int64,
+                )
                 self.host_ctx_cached_token_indptr = torch.zeros_like(
                     self.ctx_cached_token_indptr,
                     device='cpu',
@@ -722,16 +730,19 @@ class TrtllmAttentionMetadata(AttentionMetadata):
                 self.ctx_uncached_token_indptr = get_empty(
                     (self.max_num_requests + 1, ),
                     cache_name="ctx_uncached_token_indptr",
-                    dtype=torch.int64)
+                    dtype=torch.int64,
+                )
                 self.host_ctx_uncached_token_indptr = torch.zeros_like(
                     self.ctx_uncached_token_indptr,
                     device='cpu',
                     pin_memory=True,
                 )
                 # context full seqlens include cached tokens and uncached tokens
-                self.ctx_kv_indptr = get_empty((self.max_num_requests + 1, ),
-                                               cache_name="ctx_kv_indptr",
-                                               dtype=torch.int64)
+                self.ctx_kv_indptr = get_empty(
+                    (self.max_num_requests + 1, ),
+                    cache_name="ctx_kv_indptr",
+                    dtype=torch.int64,
+                )
                 self.host_ctx_kv_indptr = torch.zeros_like(
                     self.ctx_kv_indptr,
                     device='cpu',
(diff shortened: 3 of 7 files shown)


HuiGao-NV:
B200 cases meet "RuntimeError: CUDA error: CUDA-capable device(s) is/are busy or unavailable".
Need to rerun.

HuiGao-NV:
/bot run --stage-list="A30-PyTorch-2,A30-PyTorch-1,H100_PCIe-PyTorch-1"

tensorrt_llm/_torch/attention_backend/flashinfer.py
```diff
@@ -167,10 +167,17 @@ def __post_init__(self) -> None:
...
+
+    def __post_init_with_buffers__(self, buffers) -> None:
+        self.__post_init__()
+
     def create_cuda_graph_metadata(self,
                                    max_batch_size: int,
                                    sub_cross_metadata: bool = False,
-                                   max_draft_tokens: int = 0) -> Self:
+                                   max_draft_tokens: int = 0,
+                                   buffers=None) -> Self:
```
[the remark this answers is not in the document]
HuiGao-NV: This is formatted by pre-commit script and "buffers = None" makes pre-commit report error.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -604,101 +604,184 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
...
-                    ],
-                    dtype=torch.int32,
-                    device='cuda',
-                )
+                tmp_cpu_tensor = torch.empty([
+                    self.kv_cache_manager.max_batch_size,
+                    self.kv_cache_manager.max_blocks_per_seq
+                ],
+                                             device='cpu',
+                                             dtype=torch.int32)
```
QiJune, line 743: It's easy to extend `find_or_create_buffer` to support tensor shapes as inputs to avoid creating a temporary tensor.

```python
ef find_or_create_buffer(
    like_tensor_or_shape: Union[torch.Tensor, torch.Size, List[int], Tuple[int, ...]],
    buffers: Optional[Dict[str, List[torch.Tensor]]],
    cache_name: str,
    dtype: Optional[torch.dtype] = None
) -> torch.Tensor:
    """
    Finds a compatible, reusable buffer from a cache or creates a new one.

    This function can be called with either a tensor or a shape. It searches for a
    pre-allocated buffer that can be reused for the operation.

    The compatibility rules are:
    - For 1D shapes: The buffer's total elements must be >= the required elements.
    - For N-D shapes (>1D): The buffer's total elements must be a perfect
      multiple of the required elements.

    Args:
        like_tensor_or_shape: A tensor or a shape (tuple, list, torch.Size)
                              defining the required buffer properties.
        buffers: A dictionary mapping cache names to lists of buffer tensors.
                 Can be `None` or empty.
        cache_name: The key for the specific list of buffers to search in.
        dtype: The required data type for the buffer. This parameter is
               **required** if `like_tensor_or_shape` is a shape, but ignored
               if it's a tensor (the dtype is inferred from the tensor).

    Returns:
        An existing compatible buffer or a newly created one.
    
    Raises:
        ValueError: If a shape is provided without a corresponding `dtype`.
    """
    # --- 1. Unify input to get shape, numel, dim, and dtype ---
    if isinstance(like_tensor_or_shape, torch.Tensor):
        shape = like_tensor_or_shape.shape
        numel_like = like_tensor_or_shape.numel()
        dim = like_tensor_or_shape.dim()
        final_dtype = like_tensor_or_shape.dtype
    else:
        # Input is a shape (list, tuple, or torch.Size)
        if dtype is None:
            raise ValueError(
                "The 'dtype' argument is required when passing a shape instead of a tensor."
            )
        shape = torch.Size(like_tensor_or_shape)
        numel_like = shape.numel()
        dim = len(shape)
        final_dtype = dtype

    # --- 2. Search for a compatible buffer in the cache ---
    if buffers is not None:
        candidate_buffers = buffers.get(cache_name, [])
        for buffer in candidate_buffers:
            # Skip buffers with the wrong dtype
            if buffer.dtype != final_dtype:
                continue
                
            numel_buffer = buffer.numel()
            # Condition for 1D tensors: buffer just needs to be large enough.
            if dim == 1:
                if numel_buffer >= numel_like:
                    return buffer
            # Condition for N-D tensors: buffer size must be an exact multiple.
            else:
                if numel_like > 0 and numel_buffer % numel_like == 0:
                    return buffer

    # --- 3. If no suitable buffer was found, create a new one ---
    return torch.empty(shape, dtype=final_dtype, device='cuda')
```

tensorrt_llm/_torch/pyexecutor/cuda_graph_runner.py
```diff
@@ -76,7 +76,12 @@ def __init__(
         self.optional_extra_model_inputs = ["mrope_position_deltas"]
 
     def __del__(self):
-        self._graph.reset()
+        del self.attn_metadata
+        del self.spec_metadata
+        del self.input_ids
+        del self.position_ids
```
yuxianq, line 82: Why should we delete these tensors manually? If all tensors in `DecodingCUDAGraphRunner` should be deleted, should we also delete `self.mrope_position_deltas`?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -604,101 +604,184 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
...
+                numel_like = like_tensor.numel()
+
+                for buffer in candidate_buffers:
+                    numel_buffer = buffer.numel()
+
+                    # Condition for 1D tensors: buffer just needs to be large enough.
+                    if like_tensor.dim() == 1:
+                        if numel_buffer >= numel_like:
+                            return buffer  # Found a fit, return immediately.
+                    # Condition for N-D tensors: buffer size must be an exact multiple.
```
QiJune, line 680: What's the motivation of "buffer size must be an exact multiple for a N-D tensor"?

tensorrt_llm/_torch/attention_backend/flashinfer.py
```diff
@@ -167,10 +167,17 @@ def __post_init__(self) -> None:
                                                  device='cuda',
                                                  dtype=torch.int)
 
+    def get_runtime_buffers(self):
+        return {}
+
+    def __post_init_with_buffers__(self, buffers) -> None:
```
yuxianq, line 173: Can we move the default implementation of `get_runtime_buffers` and `__post_init_with_buffers__` to the base class `AttentionMetadata`?

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -318,7 +319,7 @@ def create_cuda_graph_metadata(self,
                 )
 
         cuda_graph_metadata.num_contexts = 0
-        cuda_graph_metadata.__post_init__()
+        cuda_graph_metadata.__post_init_with_buffers__(buffers)
```
yuxianq, line 322: `__post_init__` is a standard API for data class, so it has `__xxx__` name convention. `__post_init_with_buffers__` is added by us, we can use `post_init_with_buffers` directly.

tensorrt_llm/_torch/attention_backend/trtllm.py
HuiGao-NV: This makes this function confusing because it may received conflicted parameters of tensor and dtype.
I used to add another function to accept only shape and dtype and create an empty tensor inside and call find_or_create_buffer. The two locations need it use the same shape and dtype, a new helper function doesn't help clean this code.

tensorrt_llm/_torch/attention_backend/flashinfer.py
```diff
@@ -167,10 +167,17 @@ def __post_init__(self) -> None:
                                                  device='cuda',
                                                  dtype=torch.int)
 
+    def get_runtime_buffers(self):
```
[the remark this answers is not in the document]
HuiGao-NV: Change the base branch to 1.0 directly is not possible because 1.0 and main have too many different commits. I can cherry-pick it to 1.0.

tensorrt_llm/_torch/pyexecutor/cuda_graph_runner.py
HuiGao-NV: This is not a must. I used to think that this can help destroy objects and its sub-objects more quickly.
After checking this online, I find it may not help because GC is always required to free objects.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -443,6 +443,7 @@ def __init__(
         self.kv_cache_manager_key = ResourceManagerType.KV_CACHE_MANAGER
         self.lora_model_config: Optional[LoraModelConfig] = None
         self.cuda_graph_dummy_request = None
+        self.cuda_graph_meta_buffers: dict[str, List[torch.Tensor]] = {}
```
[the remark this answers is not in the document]
HuiGao-NV: Thanks. Change it to list.

tensorrt_llm/_torch/attention_backend/flashinfer.py
HuiGao-NV: I didn't do this now because only part of AttentionMetadata types update with provided share buffers. It isn't an action of AttentionMetadata, but special actions of some sub-classes.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -765,6 +848,9 @@ def prepare(self) -> None:
                 self.host_kv_cache_block_offsets,
                 self.request_ids[self.num_contexts:], self.beam_width,
                 self.num_contexts)
+            host_kv_shape = self.host_kv_cache_block_offsets.shape
+            self.kv_cache_block_offsets.view(host_kv_shape[0], -1,
```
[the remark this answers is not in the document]
HuiGao-NV: Because this copy action needs a proper shape to work:
self.kv_cache_block_offsets[:, :self.num_seqs].copy_(
                self.host_kv_cache_block_offsets[:, :self.num_seqs],
                non_blocking=True)

tensorrt_llm/_torch/attention_backend/trtllm.py
HuiGao-NV: The _copy action in trtllm.py requires a proper shape to work and there is no interface on tensor to copy data on a given offset of storage. Then a view op is applied. N-D tensors just likes array of numbers with N dimensions. Then, it has to be multiple size of lower dimensions' product.

Let's have a look at one possible situation in theory: if one buffer tensor requires [4, 32, 16] but give a buffer tensor with shape [10, 32, 8]. tensor with [10, 32, 8] can't be viewed as [?, 32, 16].

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: I agree with June that we could mimic `torch.empty`'s API, and I propose to simplify the implementation as follows:

1. Always create a 1D buffer and view it as like_shape, so that we don't have to handle 1D and ND differently
2. We don't need to create new buffer when shape mismatch, we can extend it to larger buffer with `tensor.resize_` and slice it to smaller buffer with `tensor[n:]` when returned, and keep the extended buffer in the cache
3. We can update `buffers` inside `find_or_create_buffer` instead of inside model engine, so that `get_runtime_buffers` is unnecessary.

tensorrt_llm/_torch/attention_backend/flashinfer.py
yuxianq: We already have used `post_init_with_buffers` inside `AttentionMetadata.create_cuda_graph_metadata`, so it assumes all subclass should support `post_init_with_buffers`.

yuxianq: I have another proposal: in `create_cuda_graph_metadata`, we can store `buffers` inside `cuda_graph_metadata.cuda_graph_buffers`, and use the original `__post_init__` instead of `post_init_with_buffers`. In trtllm backend, we reuse `buffers` if `cuda_graph_metadata.cuda_graph_buffers` is not None. No extra API is needed in this case.

tensorrt_llm/_torch/attention_backend/trtllm.py
HuiGao-NV: 1. The reason why to handle 1D and ND differently is there will be, in theory, cases that  buffer tensors can't be viewed to required shape of tensor during runtime. Such as buffer is [9, 2, 32], required shape of tensor is [:, 4, 32]. 1D can work for such situation but ND can't.  
    This problem is related to the question #2.
    Since the problem is only in theory but doesn't happen in current implementation, so I relax the check of numel. This change makes question 2 not exist more.
    
2. By my understanding, tensor.resize_ to a bigger number of elements, the underlying storage will be resized, if current storage can't be extended with continuous storage, a new storage will be allocated from allocator. Thus, this will break previous captured graphs, right?
```
>>> a = a.resize_(16)
>>> a.data_ptr()
140117696
>>> a = a.resize_(64)
>>> a.data_ptr()
38059712
```

Slice operation can help to get smaller storage from a tensor with more elements.

3. This is a good point. I thought this before too. If do this, buffers need to be class variable. What I worried is if there are multiple engines and some of them want to use buffers while some don't. I am not sure which way is better.

However, I don't prefer to add more parameters to find_or_create_buffer. This makes the function hard to understand.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -599,101 +599,186 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
...
+
+            # If we get here, no suitable buffer was found in the cache. Create a new one.
+            return torch.empty_like(like_tensor, device='cuda')
+
+        def find_or_create_buffer_by_shape(tensor_shape: list[int],
+                                           dtype: torch.dtype,
+                                           buffers: Optional[dict[
+                                               str, list[torch.Tensor]]],
+                                           cache_name: str) -> torch.Tensor:
+            tmp_cpu_tensor = torch.empty(tensor_shape,
```
QiJune, line 683: Could you clarify the requirement for instantiating a tensor at this stage? The implementation appears to only leverage its metadata (shape, dtype), suggesting a potential optimization to avoid unnecessary object creation and memory allocation

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -599,101 +599,186 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
...
+                                  cache_name: str) -> torch.Tensor:
+            """
+            Finds a compatible, reusable buffer from a cache or creates a new one.
+
+            This function searches for a pre-allocated tensor (buffer) that can be
+            reused for an operation involving a tensor with the shape of `like_tensor`.
+
+            The compatibility rules are:
+            - For 1D tensors: The buffer's total elements must be >= `like_tensor`'s.
+            - For N-D tensors (>1D): The buffer's total elements must be a perfect
```
QiJune, line 647: Since the implementation changes, the comments need to be changed accordingly.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: 2. We create CUDA graphs by the order of larger bs to smaller bs, right? So the required buffer sizes should be smaller and smaller, we can assert it before resize_ to guarantee it.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -599,101 +599,186 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
...
+    def post_init_with_buffers(self, buffers) -> None:
         super().__post_init__()
+
         # Set a default value, as max_num_sequences is not always set.
         if self.max_num_sequences is None:
             self.max_num_sequences = self.max_num_requests
 
-        self.prompt_lens_cuda = torch.empty(
+        def find_or_create_buffer(like_tensor: torch.Tensor,
+                                  buffers: Optional[dict[str,
```
yuxianq, line 636: find_or_create_buffer is an inner function, we can remove this buffers argument and directly use the outer one.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -599,101 +599,186 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
...
+                    numel_buffer = buffer.numel()
+
+                    # buffer just needs to be large enough.
+                    if numel_buffer >= numel_like:
+                        return buffer  # Found a fit, return immediately.
+
+            # If we get here, no suitable buffer was found in the cache. Create a new one.
+            return torch.empty_like(like_tensor, device='cuda')
+
+        def find_or_create_buffer_by_shape(tensor_shape: list[int],
```
yuxianq, line 678: We can mimic the `empty/empty_like` API, rename `find_or_create_buffer/find_or_create_buffer_by_shape` to `get_empty/get_empty_like`

tensorrt_llm/_torch/attention_backend/trtllm.py
HuiGao-NV: Fixed

tensorrt_llm/_torch/attention_backend/trtllm.py
HuiGao-NV: Fixed

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -600,21 +600,72 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
...
+
+            The compatibility rules are: The buffer's total elements must be >= tensor_shape's.
+
+            If a compatible buffer is found, it's returned immediately. Otherwise, a new
+            buffer is allocated on the 'cuda' device with the give properties of 'tensor_shape' and 'dtype'.
+
+            Args:
+                tensor_shape: The required shape.
+                dtype: The required dtype.
+                buffers: A dictionary mapping cache names to lists of buffer tensors.
```
yuxianq, line 627: `buffers` argument is removed now.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -600,21 +600,72 @@ def host_kv_cache_pool_mapping(self) -> Optional[torch.Tensor]:
 
     def __post_init__(self) -> None:
         super().__post_init__()
+        self.__post_init_with_buffers(self.cuda_graph_buffers)
+
+    def __post_init_with_buffers(self, buffers) -> None:
```
yuxianq, line 605: Use `_post_init_with_buffers` instead.

yuxianq (approved):

QiJune (approved):
LGTM

kaiyux (approved):

PR #6648, opened 2025-08-06
[https://nvbugs/5430124][fix] Mistral mixture_text_image test case fix

yechank-nvidia:
Bug fix for nvbug5404005

## Summary by CodeRabbit

* **Tests**
  * Updated test inputs for multimodal media fields to use string values instead of lists.
  * Revised expected keywords for specific model and modality in test cases.

### DIFF
--- a/tests/integration/defs/test_e2e.py
+++ b/tests/integration/defs/test_e2e.py
@@ -2226,7 +2226,7 @@ def test_ptp_quickstart_multimodal(llm_root, llm_venv, model_name, model_path,
             ],
             "mixture_text_image":
             [["invention", "person", "scientists", "Lick", "engineers"],
-             ["landscape", "dome", "yosemite", "altitude", "scattered"]]
+             ["landscape", "trees", "road", "natural", "rock"]]
         },
         "gemma-3-27b-it": {
             "image": [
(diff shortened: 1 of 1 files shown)


tests/integration/defs/test_e2e.py
```diff
@@ -2226,7 +2226,7 @@ def test_ptp_quickstart_multimodal(llm_root, llm_venv, model_name, model_path,
             ],
             "mixture_text_image":
             [["invention", "person", "scientists", "Lick", "engineers"],
-             ["landscape", "dome", "yosemite", "altitude", "scattered"]]
+             ["landscape", "trees", "road", "natural", "rock"]]
```
amukkara, line 2229: not blocking: we should remove these keyword based checks and add an evaluation dataset (like MMMU) instead. I think with every upgrade to transformers, the reference output changes breaking the test.

amukkara (approved):

tests/integration/defs/test_e2e.py
yechank-nvidia: I see. The MMMU support is on the multimodal JIRA list. Maybe I can create a new JIRA for that after we implement MMMU. @chang-l for viz.

PR #6776, opened 2025-08-11
[https://nvbugs/5412456][fix] Fix an illegal instruction was encountered

zhou-yuxin:
pytest tests/unittest/trt/attention/test_gpt_attention.py occurs 'RuntimeError: CUDA error: an illegal instruction was encountered'

## Summary by CodeRabbit

* **Bug Fixes**
  * Improves stability of fused multi-head attention by safely handling a rare edge case where the key/value sequence length could be reported as zero on some GPU/driver configurations. A fallback now ensures a valid, non-zero value, reducing the risk of launch/setup failures (e.g., during descriptor initialization). This change does not affect outputs, performance, or other behaviors in normal scenarios and involves no public API changes.


zhou-yuxin:
pytest unittest/trt/attention/test_gpt_attention.py

image

PerkzZheng (approved):

PR #6842, opened 2025-08-12
[None][fix]Make logprobs faster by not copying over everything

Pernekhan:
## Summary by CodeRabbit

* **Refactor**
  * Sampling now reads probability data directly from on-device runtime state, eliminating redundant host-side conversions.
  * Reduced latency and memory overhead during token generation.
  * Log-prob recording is now conditional and more robust (only recorded when available).
  * No changes to public APIs or external behavior.

## Description

This makes it faster when asking for logprobs. TRTLLM returns the entire logprobs list everytime, and copying over entire list again and again makes it slower. I've profiled with nsys and it's faster now. 

update_requests_single_beam_single_step  19ms (before) -> 62microseconds(after)

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/sampler.py
+++ b/tensorrt_llm/_torch/pyexecutor/sampler.py
@@ -909,10 +909,6 @@ class TRTLLMSampler(Sampler):
         sequence_lengths_host_data = state.host.sequence_lengths.flatten(
         ).tolist()
         finish_reasons = state.host.finish_reasons.flatten().tolist()
-        log_probs_host = state.host.log_probs.tolist(
-        ) if state.host.log_probs is not None else None
-        cum_log_probs_host = state.host.cum_log_probs.tolist(
-        ) if state.host.cum_log_probs is not None else None
 
         reqs = [
             r for r in state.scheduled_requests.context_requests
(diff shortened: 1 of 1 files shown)


nvxuanyuc:
The current `update_requests_single_beam_single_step` implementation incurs significant performance overhead proportional to the maximum batch size due to the [.tolist()](https://github.com/NVIDIA/TensorRT-LLM/pull/6842/files#diff-76ef6eceb82cb64d4ea6d49fca82d42ac6c942a45ad9fc8d3ff6a9b49e3b466bL912) operation, creating a large CPU workload (“bubble”) that cannot be hidden with overlap scheduling.

E.g. Llama 3.1 8B on single GPU
- Max BS=1 → ~2.05 ms
- Max BS=4 → ~11.5 ms
- Max BS=64 → ~270 ms

Replacing .tolist() with [.item()](https://github.com/NVIDIA/TensorRT-LLM/pull/6842/files#diff-76ef6eceb82cb64d4ea6d49fca82d42ac6c942a45ad9fc8d3ff6a9b49e3b466bR943) logic avoids unnecessary CPU memory allocation and copies, reducing execution time substantially and making it scale with the number of active requests. With a single active request, execution time of `update_requests_single_beam_single_step` drops to ~36 µs.
Screenshot 2025-08-14 at 11 46 08 AM
Screenshot 2025-08-14 at 11 49 43 AM

dcampora (approved):

dcampora:
Looks good to me.

PR #6850, opened 2025-08-13
[None][fix] Complete the last missing allreduce op in Llama3/4.

hyukn:
The allreduce op of the last decoder layer is missing in some circumstances for the models Llama3 and Llama4.

## Summary by CodeRabbit

- **Bug Fixes**
  - Disabled fusion optimizations on the final decoder layer for both MLP and MoE to prevent end-of-network issues.
  - Refined post-MLP and post-MoE fusion gating to depend on the presence of the next layer’s normalization and next-layer quantization state, improving correctness across quantized and non-quantized paths.

- **Chores**
  - Preserved public APIs while improving runtime behavior and unpacking logic for fused outputs; clarified comments and formatting.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_llama.py
+++ b/tensorrt_llm/_torch/models/modeling_llama.py
@@ -550,50 +550,60 @@ class Llama4DecoderLayer(DecoderLayer):
                                                       hidden_states, residual)
 
         if (self.fusion_config.POST_MOE_FUSION
-                or self.fusion_config.POST_MLP_FUSION
-            ) and self.next_layer_layernorm is not None:
-            # Get the scale for the next allreduce fusion op
-            if self.next_attn is not None and (self.is_nvfp4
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -706,6 +707,11 @@ def forward(
                 scale = self.mlp.gate_up_proj.input_scale
             else:
                 scale = None
+
+            if self.mapping.tp_rank == 0:
```
litaotju, line 711: Lets delete this log?

It prints on every forward, its very verbose.

litaotju (approved):

tensorrt_llm/_torch/models/modeling_llama.py
hyukn: Done~

Issue #5861, opened 2025-07-09
Quantization-aware Transformation Pipeline via Early Fusion + Node Annotation


Fridah-nv:
Background

Currently, quantization transformation is applied at the beginning of the transformation pipeline. It rewrites standard linear ops into quantized ops (e.g., fused quant_linear). While this works, it introduces coupling between transformation passes and quantization backends. I.e. Downstream passes (e.g. pattern matching, weight fusion, sharding) must be quantization-aware.

Goals
- Introduce an annotation-based approach to preserve quantization metadata across passes. Unify and simplify the handling of quantized ops in the transformation pipeline.
- Leverage the InferenceOptimizer transformation system (staged passes, config inheritance) to modularize quantization-aware logic.
- Consider alternative ways other than node.meta to preserve quant info

Exploration:
- Configurable Quantization Backend options similar to attention backend

PR #6876, opened 2025-08-13
[#5861][feat] Implement Annotation-Based Quantization Transformation Pipeline

therealnaveenkamal:
## Summary by CodeRabbit

- New Features
  - Introduces a two-stage quantization workflow (annotation and late fusion) enabling backend-aware quantization and performance optimizations.
- Chores
  - Quantization is now disabled by default; enable it via configuration when needed.
  - Added configuration entries to control the new quantization stages and their execution order.

## Description

This PR (addressing #5861) refactors the QAT pipeline to decouple quantization logic from other graph optimization passes by introducing annotations.

`quantize_annotate`: A pass runs early to annotate nodes with quantization metadata without actually modifying the graph structure.

`quantize_late_fusion`: Applies actual quantization transformations based on the preserved annotations

The intermediate passes only need to preserve the node metadata, ideally they need not be quantization-aware.

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/config/default.yaml
+++ b/tensorrt_llm/_torch/auto_deploy/config/default.yaml
@@ -23,7 +23,6 @@ transforms:
     stage: post_export
     backend: real_quant
     requires_clean_graph: false
-  # old quantization transform - for backward
   quantize:
     stage: pattern_matcher
     enabled: false
--- a/tensorrt_llm/_torch/auto_deploy/config/default.yaml
+++ b/tensorrt_llm/_torch/auto_deploy/config/default.yaml
@@ -19,8 +19,14 @@ transforms:
     stage: post_export
   cleanup_input_constraints:
     stage: post_export
+  quantize_annotate:
+    stage: post_export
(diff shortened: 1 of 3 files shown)


therealnaveenkamal:
cc @Fridah-nv 
I've tried to implement annotation based quantization, let me know your thoughts.

Fridah-nv:
> Hi @Fridah-nv thanks for getting back. My goal with this PR was to help with issue #5861 - to annotate nodes and transform wherever required.
> 
> I'd love to help out with the new quantization architecture you're working on instead. What's the best way for me to get involved and contribute to the current effort?
> 
> Please let me know how I can best help. Thanks!

Hi @therealnaveenkamal, thank you for the offer to help. At this stage, the refactor is already in the finalization phase, so there’s limited room for additional changes without disrupting the current plan.

If you’re looking to get involved, I’d recommend checking out issues start with `Call for contributions` — these are great entry points where community contributions can have a strong impact right now. We’d really value your expertise there.

Thanks again for your effort on this PR and for being an active part of the project!

PR #6927, opened 2025-08-15
[TRTLLM-6835][fix] Fix potential hang caused by python multiprocessing when prefetching weights

lancelly:
This PR replaces multiprocessing with multithreading when prefetching weights, since the workload is I/O-bound. Python’s default fork start method for multiprocessing carries a risk of hangs (see https://github.com/python/cpython/issues/84559), and when used together with MPI it can deadlock. In theory, multithreading is the safer choice; we have verified on an 8-GPU node that loading DeepSeek-R1 takes no longer than before. See the following pics:
截屏2025-08-15 10 39 46
截屏2025-08-15 10 41 23

## Summary by CodeRabbit

* **Refactor**
  * Improved prefetching performance by switching to thread-based parallelism for I/O tasks, enhancing reliability across environments.
  * Ensures all prefetch operations complete before continuing, resulting in smoother and potentially faster model initialization.
  * No configuration changes required; worker limits remain automatically tuned to system resources, helping reduce startup overhead on multi-core systems.

### DIFF
--- a/tensorrt_llm/_torch/models/checkpoints/hf/weight_loader.py
+++ b/tensorrt_llm/_torch/models/checkpoints/hf/weight_loader.py
@@ -1,6 +1,7 @@
 import glob
 import multiprocessing
 import os
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any, List
 
 import psutil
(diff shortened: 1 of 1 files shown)


yuxianq (approved):

djns99 (approved):
LGTM, I have confirmed this appears to fix the issue

litaotju (approved):

PR #6996, opened 2025-08-18
[https://nvbugs/5458874][fix] Fix Nemotron-H flaky CUDA graph / overlap scheduler test

tomeras91:
## Summary by CodeRabbit

- **Tests**
  - Re-enabled a previously skipped CUDA-graph overlap scheduling test.
  - Added precise assertions comparing behavior with and without overlap under CUDA graphs, including checks on generated logits and decode log probabilities.
  - Tightened numerical tolerances to increase sensitivity of comparisons.
  - Updated assertion messages to clearly reflect CUDA-graph scenarios.
  - Overall, improves test coverage and confidence in CUDA-graph overlap behavior without affecting user-facing functionality.

## Description

Make test less flaky by not comparing all logits for all generated steps, but just all logits for first generation step and just generated logits for all generation steps (similar to what is tested for CUDA graphs in this test).
Still use a small tolerance since overlap scheduler should not affect logits.

## Test Coverage

`tests/unittest/_torch/modeling/test_modeling_nemotron_h.py::test_nemotron_h_cuda_graph_overlap_scheduler` now passes and should not be flaky

### DIFF
--- a/tests/unittest/_torch/modeling/test_modeling_nemotron_h.py
+++ b/tests/unittest/_torch/modeling/test_modeling_nemotron_h.py
@@ -247,7 +247,7 @@ def test_nemotron_h_correctness(mamba_ssm_cache_dtype):
         nemotron_h.shutdown()
 
 
-@pytest.mark.skip(reason="https://nvbugs/5458874")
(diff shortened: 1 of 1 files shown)


danielafrimi (approved):
LGTM

PR #7003, opened 2025-08-18
[None][fix] Fix build of tritonbuild/tritonrelease image

dbari:
## Summary by CodeRabbit

* **Chores**
  * Container builds now embed a resolved short-tag for backend builds.
  * Example files ending with .safetensors are excluded from container contexts.
  * Cache directory resolution is canonicalized to an absolute path to improve mount and path handling.

* **Refactor**
  * Build tooling now uses absolute paths, clearer tag selection, and explicit install steps for more reliable, reproducible builds.

## Description

Building `tritonbuild` or `tritonrelease` image failed due to mismatching directories in `triton_backend/inflight_batcher_llm/scripts/build.sh`, which is called from `docker/Dockerfile.multi`. Also, some files from the `docker` and `jenkins` directories are needed for `build.sh` to get the desired tag of the Triton repositories used when compiling. These have been added to `docker/Dockerfile.multi`.

A couple of small usability improvements:
- `docker/Makefile` now resolves `${HOME_DIR}/.cache` if it's a symlink. Without this, `make -C docker xyz_run LOCAL_USER=1` used to fail.
- `examples/**/*.safetensors` was added to `.dockerignore` in case there are checkpoints generated by integration tests in the working directory.

Edit: Saw recently merged PR #6898 by @Tabrizian with a similar fix. As discussed with @MartinMarciniszyn, I addressed the comments and rebased onto `release/1.0`.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


dbari:
Ready for review. This should not affect anything that is built/tested during the pre-merge run, as it touches lines only relevant to `tritonbuild`/`tritonrelease`, which are not part of the pre-merge. Should I start a full build despite this?

tburt-nv (commented):
I don't have expertise here. I'll remove my review in favor of Martin's review, since his review has already been requested.

docker/Dockerfile.multi
```diff
@@ -175,6 +175,8 @@ FROM wheel AS tritonbuild
 
 WORKDIR /src/tensorrt_llm
 RUN pip install /src/tensorrt_llm/build/tensorrt_llm*.whl
+COPY ./jenkins/scripts/get_triton_tag.sh ./jenkins/scripts/get_triton_tag.sh
+COPY ./docker/Dockerfile.multi ./docker/Dockerfile.multi
```
MartinMarciniszyn, line 179: This is very cumbersome: we pass a script and `Dockerfile.multi` to the docker build just to extract a variable which is defined above in the same file within `build.sh` in the stage below. Instead, pass the tag of the triton image as an argument to `build.sh` below. You may need to adjust some Jenkins scripts.

MartinMarciniszyn (changes requested):

docker/Dockerfile.multi
dbari: Done as requested. No Jenkins scripts had to be changed, as `build.sh` is only ever called from here.

MartinMarciniszyn (approved):

triton_backend/inflight_batcher_llm/scripts/build.sh
```diff
@@ -35,14 +38,22 @@ done
 echo "Using TRT_ROOT=${TRT_ROOT}"
 echo "Using BUILD_UNIT_TESTS=${BUILD_UNIT_TESTS}"
 
+DIRNAME="$(dirname "$(realpath "$0")")"
+if [ -z "$TRITON_SHORT_TAG" ]; then
+  # Get TRITON_SHORT_TAG from docker/Dockerfile.multi
+  LLM_ROOT="${DIRNAME}/../../.."
+  TRITON_SHORT_TAG=$("$LLM_ROOT/jenkins/scripts/get_triton_tag.sh" "$LLM_ROOT")
```
hypdeb, line 45: Do I understand correctly that the dockerfile is the source of truth for base images versions, and that is why we do this here? Is there maybe a better way? Maybe a `.env` file?

docker/Makefile
```diff
@@ -137,7 +137,7 @@ CODE_DIR          ?= /code/tensorrt_llm
 EXTRA_VOLUMES     ?=
 CCACHE_DIR        ?= $(CODE_DIR)/cpp/.ccache
 CONAN_DIR         ?= $(CODE_DIR)/cpp/.conan
-USER_CACHE_DIR    ?= $(HOME_DIR)/.cache
+USER_CACHE_DIR    ?= $(shell readlink -f "${HOME_DIR}/.cache")
```
hypdeb, line 140: Could we think about using those: https://specifications.freedesktop.org/basedir-spec/latest/? The home directory tends to be really small for us. I think `pip` and some other package managers follow it already.

jdebache (commented):
Added some suggestions, maybe out of scope, I'll leave it to your judgement.

docker/Makefile
dbari: Good point. It could be changed to the following:
```suggestion
USER_CACHE_DIR    ?= $(shell readlink -f "${XDG_CACHE_HOME:=${HOME_DIR}/.cache}")
```
However I don't know how if it's safe in combination with the override in Docker rootless mode (mentioned [here](https://github.com/NVIDIA/TensorRT-LLM/blob/d17f13a305f51100262b351f0fd3fc335f6cfe8d/docker/Makefile#L12))

triton_backend/inflight_batcher_llm/scripts/build.sh
MartinMarciniszyn: The Dockerfile was always the source of truth for the tags of the base images. I looked at using `.env` files in the first implementation, but all tools support these well. For now, I suggest keeping it as is. We can approach this in a separate MR if there is available bandwidth.

docker/Makefile
dbari: > Nevermind, just checked, these env variables are not set to any defualt in standard Ubuntu...

That's by design. If `XDG_CACHE_HOME` is not defined, it should default to `$HOME/.cache` per the spec. Leaving it as is for now.

jdebache (approved):

PR #6220, opened 2025-07-21
[TRTLLM-5830][feat] Improve LoRA cache memory control

amitz-nv:
## Description

1. Added support for configuring LoRA cache sizes on pytorch flow.
2. Changed `LoraConfig.max_loras` and `LoraConfig.max_cpu_loras` to be optional. When they're not set, the cache size would be determined by the `PeftCacheConfig`.
3. Implemented a "merge" logic of fields in `peft_cache_config` and `lora_config` - when cache size fields in `lora_config` have a value, they would take precedence over the relevant fields in `peft_cache_config`
4. Removed deprecated LoRA LLM args, as they're already specified inside `lora_config: LoraConfig` LLM arg: `max_lora_rank`, `max_loras`, `max_cpu_loras`.
5. For user clarity - defined default values in python `PeftCacheConfig` class for `device_cache_percent` to 2% and `host_cache_size` to 1GiB, the same default values that the CPP code uses when these fields have no value.
6. In TRT-python flow, `lora_config` in LLM args would take precedence over `lora_config` from the engine build config.
7. Raise an exception when `peft_cache_config.lora_prefetch_dir` has a value, as currently it's not supported.
8. Added tests that verify the LoRA cache size LLM args take effect in the expect order.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of LoRA and PEFT cache configurations, allowing cache size parameters in LoRA configs to override those in PEFT cache configs.
  * Enhanced flexibility for specifying LoRA adapter cache sizes by supporting optional values.
  * Added validation to prevent unsupported LoRA prefetch directory usage in PEFT cache configuration.
  * Refactored LoRA config integration to merge existing PEFT cache settings and support explicit overrides.
  * Simplified LoRA adapter naming in multimodal models for consistent identification.

* **Tests**
  * Added tests verifying correct behavior when LoRA and PEFT cache configurations interact, including error handling for insufficient cache sizes and precedence of LoRA config values.
  * Updated existing tests and fixtures to include new LoRA cache size parameters for comprehensive coverage.

* **Examples**
  * Updated example usage to pass LoRA configuration directly with explicit cache size parameters, removing deprecated fields.

* **Documentation & Validation**
  * Clarified PEFT cache configuration field descriptions and tightened default values.
  * Removed deprecated LoRA-related fields and warnings from configuration validation.
  * Added model validators to ensure compatibility of PEFT cache settings with backend capabilities.

## Test Coverage

[...]

PR #7014, opened 2025-08-18
[https://nvbugs/5464088] [fix] Guard against fp8 activations in lora forward; update perf test config

venkywonka:
## Summary by CodeRabbit

- New Features
  - Automatically derives LoRA count from model labels and sets max_loras and max_cpu_loras accordingly.
- Bug Fixes
  - Prevents runtime failures when LoRA receives unsupported input types by auto-casting to FP16/BF16, with a debug log for visibility.
- Tests
  - Perf configuration updated to reflect dynamic LoRA count and new presets, improving flexibility for PyTorch model runs.

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>

### DIFF
--- a/tensorrt_llm/_torch/modules/gated_mlp.py
+++ b/tensorrt_llm/_torch/modules/gated_mlp.py
@@ -5,6 +5,7 @@ import torch
 import torch.nn.functional as F
 from torch import nn
 
+from tensorrt_llm.logger import logger
 from tensorrt_llm.mapping import Mapping
 
 from ..distributed import AllReduceParams
@@ -95,12 +96,23 @@ class GatedMLP(nn.Module):
             [LoraModuleType.MLP_GATE_UP],
             [2 * self.intermediate_size // mapping.tp_size])
 
-    def _apply_activation(self, x):
+    def _apply_activation(self, x, *, for_lora: bool = False):
         if self.activation == F.silu:
             if self.down_proj.has_fp8_qdq:
-                return swiglu(x,
-                              quant_scale=self.down_proj.input_scale,
-                              quant_type=torch.float8_e4m3fn)
+                if for_lora:
+
+                    target = torch.bfloat16 if torch.cuda.is_bf16_supported(
+                    ) else torch.float16
+                    logger.debug(
+                        f"GatedMLP._apply_activation: LoRA path active; forcing non-FP8 activation dtype {target} (keeping activations in bf16/fp16), layer_idx={self.layer_idx}"
+                    )
+                    return swiglu(x,
+                                  quant_scale=self.down_proj.input_scale,
+                                  quant_type=target)
+                else:
+                    return swiglu(x,
+                                  quant_scale=self.down_proj.input_scale,
+                                  quant_type=torch.float8_e4m3fn)
             else:
                 return swiglu(x)
         elif callable(self.activation):
@@ -152,7 +164,7 @@ class GatedMLP(nn.Module):
         if h1_lora is not None:
             h1 = h1 + h1_lora
 
-        h2 = self._apply_activation(h1)
+        h2 = self._apply_activation(h1, for_lora=True)
         output = self.down_proj(h2,
                                 all_reduce_params=final_all_reduce_params,
                                 lora_params=lora_params,
--- a/tensorrt_llm/_torch/modules/gated_mlp.py
+++ b/tensorrt_llm/_torch/modules/gated_mlp.py
@@ -96,19 +96,17 @@ class GatedMLP(nn.Module):
             [LoraModuleType.MLP_GATE_UP],
             [2 * self.intermediate_size // mapping.tp_size])
 
-    def _apply_activation(self, x, *, for_lora: bool = False):
+    def _apply_activation(self, x, *, has_lora: bool = False):
         if self.activation == F.silu:
             if self.down_proj.has_fp8_qdq:
-                if for_lora:
-
-                    target = torch.bfloat16 if torch.cuda.is_bf16_supported(
-                    ) else torch.float16
-                    logger.debug(
-                        f"GatedMLP._apply_activation: LoRA path active; forcing non-FP8 activation dtype {target} (keeping activations in bf16/fp16), layer_idx={self.layer_idx}"
+                if has_lora:
+                    # NOTE: This is a WAR, since LoRA grouped_gemm does not support FP8 yet.
+                    # TODO: Remove this path when LoRA grouped_gemm supports FP8
+                    # see: cpp/tensorrt_llm/thop/loraOp.cpp::lora_grouped_gemm
+                    logger.warning(
+                        f"GatedMLP._apply_activation: LoRA path active; forcing non-FP8 activation dtype bf16/fp16, layer_idx={self.layer_idx}"
                     )
-                    return swiglu(x,
-                                  quant_scale=self.down_proj.input_scale,
-                                  quant_type=target)
+                    return swiglu(x)
                 else:
                     return swiglu(x,
                                   quant_scale=self.down_proj.input_scale,
@@ -164,7 +162,7 @@ class GatedMLP(nn.Module):
         if h1_lora is not None:
             h1 = h1 + h1_lora
 
-        h2 = self._apply_activation(h1, for_lora=True)
+        h2 = self._apply_activation(h1, has_lora=True)
         output = self.down_proj(h2,
                                 all_reduce_params=final_all_reduce_params,
                                 lora_params=lora_params,
--- a/tensorrt_llm/_torch/peft/lora/layer.py
+++ b/tensorrt_llm/_torch/peft/lora/layer.py
@@ -3,8 +3,6 @@ from typing import Dict, List, Optional
 
 import torch
 
-from tensorrt_llm._utils import logger
-
 
 class LoraModuleType(IntEnum):
     """Enum class representing different types of modules that can have LoRA adapters.
@@ -121,15 +119,6 @@ class LoraLayer(torch.nn.Module):
             if len(active_lora_module_ids) == 0:
                 return None
             else:
-                # Guard: LoRA custom op only supports FP16/BF16 activations.
-                # If upstream produced FP8 (e.g., FP8 SwiGLU), cast here to avoid runtime failure.
-                if x.dtype not in (torch.float16, torch.bfloat16):
-                    target_dtype = torch.bfloat16 if torch.cuda.is_bf16_supported(
-                    ) else torch.float16
-                    logger.debug(
-                        f"lora_grouped_gemm supports only FP16/BF16. Casting input from {x.dtype} to {target_dtype}."
-                    )
-                    x = x.to(target_dtype).contiguous()
                 lora_outputs = torch.ops.trtllm.lora_grouped_gemm(
                     x,
                     lora_params['host_request_types'][:num_seqs],
--- a/tensorrt_llm/_torch/peft/lora/layer.py
+++ b/tensorrt_llm/_torch/peft/lora/layer.py
@@ -3,6 +3,8 @@ from typing import Dict, List, Optional
 
 import torch
 
+from tensorrt_llm._utils import logger
+
 
 class LoraModuleType(IntEnum):
     """Enum class representing different types of modules that can have LoRA adapters.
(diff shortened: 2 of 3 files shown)


litaotju (approved):

tensorrt_llm/_torch/peft/lora/layer.py
```diff
@@ -119,6 +121,15 @@ def forward(
             if len(active_lora_module_ids) == 0:
                 return None
             else:
+                # Guard: LoRA custom op only supports FP16/BF16 activations.
+                # If upstream produced FP8 (e.g., FP8 SwiGLU), cast here to avoid runtime failure.
+                if x.dtype not in (torch.float16, torch.bfloat16):
```
amitz-nv, line 126: Is there a test that verifies this fixes the issue?

JunyiXu-nv: This is kind of werid. The input type of the down_proj will be checked before calling the kernel, as we can see here: 

https://github.com/NVIDIA/TensorRT-LLM/blob/655d0f48d0d2d0659c54e86db87a56a3a0229f1e/tensorrt_llm/_torch/modules/gated_mlp.py#L103

Maybe we should revise the logic here to avoid using fp8 activation? I was using the `self.down_proj.has_fp8_qdq` to determine whether we use fp8 activation. Does the model set this argument here?

The input type will keep the same as the input `x` if no dtype is set when calling swiglu.

https://github.com/NVIDIA/TensorRT-LLM/blob/655d0f48d0d2d0659c54e86db87a56a3a0229f1e/tensorrt_llm/_torch/modules/swiglu.py#L62

tensorrt_llm/_torch/peft/lora/layer.py
```diff
@@ -119,6 +121,15 @@ def forward(
...
+                    target_dtype = torch.bfloat16 if torch.cuda.is_bf16_supported(
```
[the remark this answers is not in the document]
venkywonka: I'm not sure myself, but the tests usually have dummy loras which aren't low precision. 
But either way the `lora_grouped_gemm` [[these lines](https://github.com/NVIDIA/TensorRT-LLM/blob/82276167e6a67cae84a7904a211de82f1593b7b6/cpp/tensorrt_llm/thop/loraOp.cpp#L147-L151
)] only supports fp16 anyways.

tensorrt_llm/_torch/peft/lora/layer.py
venkywonka: > I was using the self.down_proj.has_fp8_qdq to determine whether we use fp8 activation

@JunyiXu-nv , yes `self.down_proj.has_fp8_qdq` evaluates to true, but it shouldn't, when using lora (because our bindings dont' support it, not because it can't be quantized).

https://github.com/NVIDIA/TensorRT-LLM/blob/655d0f48d0d2d0659c54e86db87a56a3a0229f1e/tensorrt_llm/_torch/modules/swiglu.py#L62

oh, didn't notice that. then that would be a simpler place to modify is it? if we just override the `dtype` there in swigly, to ensure the outputs are half-precision instead of fp8 (specifically in lora path)?

venkywonka: @JunyiXu-nv , so would you say its better to make this modification/change at the site of `swiglu` : 
```python
# in gated_mlp.py
def _apply_activation(self, x, *, for_lora: bool = False):
    if self.activation == F.silu:
        if self.down_proj.has_fp8_qdq:
            if for_lora:
                target = torch.bfloat16 if torch.cuda.is_bf16_supported() else torch.float16
                return swiglu(x, quant_scale=self.down_proj.input_scale, quant_type=target)
            else:
                return swiglu(x, quant_scale=self.down_proj.input_scale, quant_type=torch.float8_e4m3fn)
        else:
            return swiglu(x)
    elif callable(self.activation):
        return self.activation(x)
    elif self.activation is None:
        return x
    else:
        raise NotImplementedError(...)
```

venkywonka: @JunyiXu-nv in the latest commit i have done the same as you suggested. can you confirm if this is what you meant. confirmed that this fixes the bug too.

JunyiXu-nv: Yes, that's what I mean. The performance could be better if we don't involve the redundent quant/dequant.

JunyiXu-nv (approved):

tensorrt_llm/_torch/modules/gated_mlp.py
```diff
@@ -95,12 +96,23 @@ def __init__(self,
...
-    def _apply_activation(self, x):
+    def _apply_activation(self, x, *, for_lora: bool = False):
         if self.activation == F.silu:
             if self.down_proj.has_fp8_qdq:
-                return swiglu(x,
-                              quant_scale=self.down_proj.input_scale,
-                              quant_type=torch.float8_e4m3fn)
+                if for_lora:
+
+                    target = torch.bfloat16 if torch.cuda.is_bf16_supported(
```
shaharmor98, line 104: I think it can be really error-prone. If one expects to use FP8 and you cast it to BF16/FP16, it will most likely affect the final output, no?

tensorrt_llm/_torch/modules/gated_mlp.py
```diff
@@ -95,12 +96,23 @@ def __init__(self,
...
+                    ) else torch.float16
+                    logger.debug(
```
shaharmor98, line 106: I'm not sure that adding a debug log level is sufficient here. TRTLLM has many debug logs, and I don't really expect anyone to catch this kind of log. It might also have serious implications in terms of the model's output. I can think of many correctness tests and use cases that could be negatively affected by this change, and it could be extremely hard for practically anyone to catch this forced casting.

tensorrt_llm/_torch/modules/gated_mlp.py
```diff
@@ -95,12 +96,23 @@ def __init__(self,
             [LoraModuleType.MLP_GATE_UP],
             [2 * self.intermediate_size // mapping.tp_size])
 
-    def _apply_activation(self, x):
+    def _apply_activation(self, x, *, for_lora: bool = False):
```
shaharmor98, line 99: nit, a matter of personal test, feel free to resolve- instead of `for_lora` --> `has_lora` ?

tensorrt_llm/_torch/modules/gated_mlp.py
```diff
@@ -95,12 +96,23 @@ def __init__(self,
...
+                if for_lora:
+
+                    target = torch.bfloat16 if torch.cuda.is_bf16_supported(
+                    ) else torch.float16
+                    logger.debug(
+                        f"GatedMLP._apply_activation: LoRA path active; forcing non-FP8 activation dtype {target} (keeping activations in bf16/fp16), layer_idx={self.layer_idx}"
+                    )
+                    return swiglu(x,
+                                  quant_scale=self.down_proj.input_scale,
+                                  quant_type=target)
```
JunyiXu-nv, line 111: Just realize that the quant_scale/quant_type should not be set here. The output from `gate_up_proj` should already be bf16/fp16, and set quant_scale will let the kernel multiply the scale with x, which would involve accuracy bug.

Maybe revise like this:

```suggestion
                    return swiglu(x)
```

Or change the above condition `if self.down_proj.has_fp8_qdq:` to `if self.down_proj.has_fp8_qdq and not for_lora:`

tensorrt_llm/_torch/peft/lora/layer.py
amitz-nv: I've conducted benchmarks recently (even today) on `Llama-3.1-8B-Instruct-FP8` with `Llama-3-Chinese-8B-Instruct-v2-LoRA` adapter and it works, what's different in the test perf that it fails there?

tensorrt_llm/_torch/peft/lora/layer.py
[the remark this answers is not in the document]
amitz-nv: TBH I don't think I know... maybe Shahar or Nave know

tests/integration/defs/perf/pytorch_model_config.py
```diff
@@ -181,10 +181,19 @@ def get_model_yaml_config(model_label: str,
 
     # lora-specific change for pytorch
     if 'pytorch' in model_label and 'loras' in model_label:
+        # Derive the requested number of adapters from model_label (segment like "loras:X")
+        lora_count = 1
+        for part in model_label.split('-'):
+            if part.startswith('loras:'):
```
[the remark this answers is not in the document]
venkywonka: the test and QA side guarantees this. that's the string format agreed upon in the way we dynamically parse the test perf test cases, see `tests/integration/defs/perf/test_perf.py::PerfTestConfig.to_string`

tests/integration/defs/perf/pytorch_model_config.py
```diff
@@ -181,10 +181,19 @@ def get_model_yaml_config(model_label: str,
...
+                lora_count = max(1, int(part.split(':', 1)[1]))
+                break
+
         lora_config = {
             'lora_config': {
                 'lora_dir': lora_dirs if lora_dirs is not None else [],
-                'max_lora_rank': 64
+                'max_lora_rank': 64,
+                'max_loras': lora_count,
```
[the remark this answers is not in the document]
venkywonka: My understainding is the below behavior was after @amitz-nv 's #6220 
- Default (no `max_loras` specified): LoRA GPU cache is sized by `PeftCacheConfig.device_cache_percent` (default 0.02 = 2% of free GPU memory after engine load), with paging determined by `optimal_adapter_size/max_adapter_size`
- With `max_loras` specified: it overrides the percentage and sets `num_device_module_layer = max_lora_rank × num_lora_modules × max_loras`, sizing the cache to hold that many adapters.

so TLDR; 
if we don't override with `max_loras`, the gpu cache set by default by `device_cache_percent` is not enough when running `test_perf[llama_v3.1_8b_instruct_fp8-bench-pytorch-float8-input_output_len:128,128-loras:1-reqs:100-con:2-gpus:1` and so overriding makes it pass.

I'd say its a WAR and the underlying issue is we probably lack the recovery and fallback.

@amitz-nv how do you think we should track the above identified issue? i can create a bug for this if you think it makes sense?

tensorrt_llm/_torch/modules/gated_mlp.py
venkywonka: but we don't have have a choice right? as fp8 anyways will break in `lora_grouped_gemm()` inside `tensorrt_llm/_torch/peft/lora/layer.py`

venkywonka: ideally, we should find a way to add `fp8` support to `lora_grouped_gemm()` and then remove this guard. i can add a TODO there, but tha's beyond the scope of `release/1.0`.

venkywonka: also, before #6540 , the `swiglu()` was emitting fp16 activations only all this while, so the change is merely restoring old behavior (only for lora path) and *not* adding new behavior.

tensorrt_llm/_torch/modules/gated_mlp.py
venkywonka: thank you @JunyiXu-nv  for catching that! i'm not familiar with this area of the code, so missed it. This should ideally have been caught right?
I guess L0 didnt' catch it but do you know if there is any accuracy coverage in L1+ for such an accuracy bug? 

maybe i should run a post merge to make sure this PR doesn't break anything (even after the above suggestion)?

tests/integration/defs/perf/pytorch_model_config.py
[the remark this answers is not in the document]
venkywonka: yep, its right below this comment :)

tensorrt_llm/_torch/peft/lora/layer.py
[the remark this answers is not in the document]
venkywonka: for now, either way, the only dtype for activations that can _not break_ `lora_grouped_gemm` is `bf16/fp16` ... so we're limited by that issue. 
I have added simplified the guarding and added a NOTE + TODO for this.

resolving this hence

shaharmor98 (approved):
LGTM

PR #7021, opened 2025-08-19
[None][chore] Wrap the swiglu into custom op to avoid redundant device copy.

hyukn:
A redundant D2D copy is observed when enabling torch.compile for the Llama model due to the swiglu triton kernel, which brings perf overhead. Use a custom op to wrap the swiglu op to avoid this overhead.

## Summary by CodeRabbit

- **New Features**
  - Adds a public SiLU-and-multiply operation with optional scaling for direct use in models and pipelines; input width must be even.
  - Provides a fake-path compatibility fallback that returns a placeholder result when the native extension isn't available.

- **Refactor**
  - SwiGLU now delegates to the new operation for consistent behavior across environments, simplifying the Python-level implementation.

### DIFF
--- a/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
+++ b/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
@@ -1,7 +1,8 @@
 from functools import lru_cache
-from typing import List, Optional, Tuple
+from typing import List, Mapping, Optional, Tuple
 
 import torch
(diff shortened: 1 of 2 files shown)


litaotju (approved):

JunyiXu-nv (approved):

liji-nv (approved):

HuiGao-NV (approved):
LGTM

PR #7133, opened 2025-08-21
[None][fix] Fix the incorrect header file import in dataType.h

Fan-Yunfan:
### Problem

Today, while reading the `dataType.h` file, I noticed that the methods defined in this file use `TLLM_THROW` and `throw std::runtime_error` to handle cases where the input is an unknown `nvinfer1::DataType`.

To use these two exception handling functions, the header file `tensorrt_llm/common/logger.h` is included. **However, the `TLLM_THROW` macro are actually defined in `tensorrt_llm/common/tllmException.h`.** The reason no compilation error occurred is that the `logger.h` includes the `assert.h`, which in turn includes `tllmException.h`. 

Given that the `dataType.h` file does not use any logging-related macros or functions from `tensorrt_llm/common/logger.h`, I believe it would be more intuitive and reasonable to replace the included header from `logger.h` to `tllmException.h`. 

This would avoid ambiguity and potential issues caused by indirect inclusion.

### Current Implementation

`cpp/include/tensorrt_llm/common/dataType.h
`
```
#pragma once

#include "tensorrt_llm/common/logger.h"
#include <NvInferRuntime.h>
#include <map>
```

### Solution

Replace `#include "tensorrt_llm/common/logger.h"` with `#include "tensorrt_llm/common/tllmException.h"`.

```
#pragma once

- #include "tensorrt_llm/common/logger.h"
+ #include "tensorrt_llm/common/tllmException.h"
#include <NvInferRuntime.h>
#include <map>
```

### Additional info

`tllmException.h` import  `<stdexcept>` as same, so don't worry `throw std::runtime_error` will occur compilation error.

```
#pragma once

#include "tensorrt_llm/common/stringUtils.h"

#include <array>
#include <cstddef>
#include <stdexcept>
#include <string>
```


Fan-Yunfan:
@coderabbitai review 

I pushed a fix in commit https://github.com/NVIDIA/TensorRT-LLM/pull/7133/commits/a8c19567fc3e7b27596fc2bf78234ed0f3747dbc, please review it.

karljang:
@tongyuantongyu,  
LGTM, but could you help review this PR~ 😄

tongyuantongyu:
> @tongyuantongyu (Yuan Tong), LGTM, but could you help review this PR~ 😄

I don't mind making this change in a larger PR, but I feel it's too trivial on its own and not worth a dedicate PR.

Fan-Yunfan:
In the Sglang community, even if you just modify a variable name, someone will review it and merge it into the main branch.

These minor changes, though not significant feature updates, still hold value in my opinion.

If only those who have an in-depth understanding of the project and possess exceptional skills can contribute, I don't believe it constitutes true open-source in the real sense.

Additionally, contributing to the TensorRT-LLM open-source community does not generate any direct benefits for me. As I am already employed full-time, my involvement stems solely from using the framework in my work and a genuine desire to learn through participation.

karljang (approved):

karljang:
@Fan-Yunfan , thanks for your contribution! What makes things complete are these small changes anyway~
Please understand above comment, I think, it's due to some congestion in our CI pipeline lately.

PR #7495, opened 2025-09-03
[https://nvbugs/5488582][fix] Avoid unexpected Triton recompilation in DG fused_moe.

hyukn:
Triton will trigger unexpected recompilation for DG fused_moe _preprocess_after_permute_kernel, where the token number was defined as a constant, which might be treated as a specialization by Triton. Replacing it with a normal int value will solve the issue.

For concurrency 4096: w/. fixing vs. w/o. fixing = 51029.35 vs. 44774.58 TOPS.

## Summary by CodeRabbit

- Refactor
  - Updated token handling to be dynamic at runtime, improving flexibility across varying batch sizes and sequence lengths and reducing reliance on compile-time constants.
  - Enhances compatibility with a wider range of configurations without manual tuning.

- Bug Fixes
  - Resolves errors and incorrect masking seen with larger or variable token counts.
  - Improves reliability and stability in edge cases involving diverse token distributions.

### DIFF
--- a/tensorrt_llm/_torch/modules/fused_moe/fused_moe_deepgemm.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/fused_moe_deepgemm.py
@@ -220,7 +220,7 @@ def _preprocess_after_permute_kernel(
     expert_offsets_ptr,
     masked_m_ptr,
     token_map_ptr,
-    TOTAL_TOKENS: tl.constexpr,
+    total_tokens,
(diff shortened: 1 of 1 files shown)


yuxianq (approved):

lfr-0531 (approved):

litaotju:
Screenshot 2025-09-04 at 23 36 10


I have to by pass and merge, the pipeline passed, only the H100 pending but this code path does not affect any of H100, its only B200.

PR #7590, opened 2025-09-07
[None][fix] Fix the incorrect code style in dataType.h

Fan-Yunfan:
### Problem
The TensorRT-LLM coding style requires the storage-class specifier `static` to appear before the type qualifier `constexpr`. 
Such as:
`include/tensorrt_llm/common /logger.h
`
```
class Logger
{
......
private:
    static auto constexpr kPREFIX = "[TensorRT-LLM]";
};
```
`cpp/include/tensorrt_llm/common/tllmException.h
`
```
namespace tensorrt_llm::common
{
......
/// @brief Constant for unknown request ID
static constexpr uint64_t kUNKNOWN_REQUEST_ID = std::numeric_limits<uint64_t>::max();

} // namespace tensorrt_llm::common
```

`cpp/include/tensorrt_llm/common/cudaUtils.h
`
```
......
template <>
struct CudaDataType<float>
{
    static constexpr cudaDataType_t value = cudaDataType::CUDA_R_32F;
};
```

However, the return-type declarations of `getDTypeSize()` and `getDTypeSizeInBits()` in `dataType.h` currently place `constexpr` first, violating this rule.

### Current Implementation
`cpp/include/tensorrt_llm/common/dataType.h 
`
```
constexpr static size_t getDTypeSize(nvinfer1::DataType type)
{
    ......
    return 0;
}

constexpr static size_t getDTypeSizeInBits(nvinfer1::DataType type)
{
    ......
    return 0;
}
```
### Solution
Replace `constexpr static ` with `static constexpr`.
```
static constexpr size_t getDTypeSize(nvinfer1::DataType type)
{
    ......
    return 0;
}

static constexpr size_t getDTypeSizeInBits(nvinfer1::DataType type)
{
    ......
    return 0;
}
```

## Summary by CodeRabbit

* **Refactor**
  * Harmonized qualifier order on compile-time utility functions to improve consistency and readability across the codebase. This change has no effect on behavior, performance, configuration, or user workflows. No API, CLI, or UI changes; existing integrations continue to work as before. No action is required from users or administrators after updating.


Fan-Yunfan:
> @Fan-Yunfan , Thanks for your contribution. Consistent coding style definitely matters! I noticed there are over 100 instances using "constexpr static" instead of "static constexpr", so I'm curious: where was the preferred coding style actually documented? 😄
In the project, there seems to be no explicitly defined coding style, nor any clear preference for using "constexpr static" versus "static constexpr." However, "static constexpr" appears significantly more frequently (found in 3,370 files with 16,246 occurrences) compared to "constexpr static" (found in 113 files with 322 occurrences). Personally, I believe maintaining a consistent convention is preferable, which is why I submitted this PR. That said, both forms are technically correct. If the project has not formally standardized this formatting, this PR can be disregarded.  😂

Fan-Yunfan:
By the way，would you like to help to review another PR（https://github.com/NVIDIA/TensorRT-LLM/pull/7133）which releated with the same file  in your free time？🌹🌹🌹

karljang:
> > @Fan-Yunfan , Thanks for your contribution. Consistent coding style definitely matters! I noticed there are over 100 instances using "constexpr static" instead of "static constexpr", so I'm curious: where was the preferred coding style actually documented? 😄
> > Thanks for your replay. In the project, there seems to be no explicitly defined coding style, nor any clear preference for using "constexpr static" versus "static constexpr." However, "static constexpr" appears significantly more frequently (found in 3,370 files with 16,246 occurrences) compared to "constexpr static" (found in 113 files with 322 occurrences). Personally, I believe maintaining a consistent convention is preferable, which is why I submitted this PR. That said, both forms are technically correct. If the project has not formally standardized this formatting, this PR can be disregarded.  😂


Yes, consistency does matter, but if it's not formally defined, let's leave it as is for now, especially since everything is evolving so quickly. 
Thanks again for your flexibility!

PR #7646, opened 2025-09-09
[https://nvbugs/5503440][fix] Fix potential hang due to wrong type of ZMQ socket and protocol for worker_init_status_queue

lancelly:
Currently, our proxy and workers communicate via ZMQ during startup: after a worker successfully starts, it sends a ready message, and the proxy polls for it. The current implementation uses ZMQ socket type PAIR with the TCP protocol, which is explicitly discouraged in the ZMQ documentation. The official documentation states the following reason:

> https://libzmq.readthedocs.io/en/latest/zmq_socket.html
> While 'ZMQ_PAIR' sockets can be used over transports other than zmq_inproc, their inability to auto-reconnect coupled with the fact that new incoming connections will be terminated while any previous connections (including ones in a closing state) exist makes them unsuitable for TCP in most cases.

In my understanding, there are two potential issues in the current implementation:
1. zmq_connect() is a non-blocking call. A successful call does not guarantee that the TCP connection has been fully established. Sending a message immediately after connecting using TCP + ZMQ_PAIR may not succeed. Moreover, there is no additional retry mechanism, so the proxy ends up waiting for the message until it times out.
As mentioned in the documentation:
> https://libzmq.readthedocs.io/en/latest/zmq_connect.html
> For most transports and socket types, the connection is not performed immediately but as needed by ØMQ. Thus a successful call to zmq_connect() does not mean that the connection was or could actually be established. Because of this, for most transports and socket types the order in which a 'server' socket is bound and a 'client' socket is connected to it does not matter. The ZMQ_PAIR and ZMQ_CHANNEL sockets are an exception, as they do not automatically reconnect to endpoints.

I have modified the worker_init_status_queue to use the REP/REQ pattern, added a simple retry mechanism on the worker side (with a default 5-second timeout).

Although it hasn’t been possible to reproduce the issue to confirm that this completely resolves it, at least the incorrect usage has been eliminated in accordance with the official documentation.

## Summary by CodeRabbit

- Bug Fixes
  - Improved startup reliability with an acknowledgment handshake and retry logic between workers and the proxy, reducing intermittent initialization failures and lost signals.
  - Enhanced shutdown and error handling to ensure deterministic cleanup of resources, preventing hangs and leaks during quits or failures.


tensorrt_llm/executor/worker.py
```diff
@@ -708,6 +709,37 @@ def __del__(self):
         self.shutdown()
 
 
+def notify_proxy_with_retry(queue, message, max_retries=5, timeout=1):
```
Superjomn, line 712: There is a [retry](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/executor/ipc.py#L131) in ipc.py, can you reuse/enhance that?

tensorrt_llm/executor/worker.py
```diff
@@ -880,7 +922,13 @@ def notify_proxy_threads_to_quit():
                                                    mp_stats_queue)
                 worker._set_iteration_result_queue(worker.kv_events_queues,
                                                    kv_cache_events_queue)
-                worker_init_status_queue.put((ready_signal, None))
+                # Send ready signal with confirmation
+                ready_msg = (ready_signal, None)
+                if not notify_proxy_with_retry(worker_init_status_queue,
```
Superjomn, line 927: So is here.

tensorrt_llm/executor/worker.py
lancelly: The retry here refers to retries when no ACK is received, which is slightly different from the original retry on send failure.

tests/unittest/_torch/executor/test_router_dealer_ipc.py
```diff
@@ -0,0 +1,373 @@
+import multiprocessing
+import pickle
```
Superjomn, line 2: You may need to add this to a pre-merge test list, as far as I can see, the dangling test file won't be tested.

Superjomn (approved):
LGTM

tensorrt_llm/executor/ipc.py
```diff
@@ -56,6 +58,9 @@ def __init__(self,
         self.name = name
         self.socket = self.context.socket(socket_type)
 
+        # For ROUTER sockets, track the last identity to enable replies. For now we assume there is only one client in our case.
+        self._last_identity = None if socket_type == zmq.ROUTER else None
```
yuxianq, line 62: Why self._last_identity is None for both cases?

tensorrt_llm/executor/proxy.py
```diff
@@ -320,6 +322,9 @@ def mpi_done_callback(future: concurrent.futures.Future):
         while True:
             if self.worker_init_status_queue.poll(1):
                 ready_signal, error_trace = self.worker_init_status_queue.get()
+                # Send ACK to the worker
+                self.worker_init_status_queue.put("ACK")
+                logger.info("get signal from executor worker")
```
yuxianq, line 327: Use `logger.debug` instead if verbose.

tensorrt_llm/executor/ipc.py
lancelly: I'll fix this typo. `self._last_identity = None` should be fine

yuxianq (approved):

tests/unittest/_torch/executor/test_router_dealer_ipc.py
lancelly: Unit tests under existed folder will be tested now, thx~

PR #8791, opened 2025-10-30
[https://nvbugs/5612504][chore] refactor cluster env check logic

Shixiaowei02:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for MPI configuration issues. The system now outputs explicit diagnostic messages and exits clearly when required MPI environment settings are missing, instead of silently failing.

### DIFF
--- a/tensorrt_llm/llmapi/trtllm-llmapi-launch
+++ b/tensorrt_llm/llmapi/trtllm-llmapi-launch
@@ -19,7 +19,8 @@ function mpi_world_size {
     elif [ -n "$OMPI_COMM_WORLD_SIZE" ]; then
         echo "$OMPI_COMM_WORLD_SIZE"
     else
-        echo "1"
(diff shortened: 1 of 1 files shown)


kaiyux (approved):

tensorrt_llm/llmapi/trtllm-llmapi-launch
```diff
@@ -19,7 +19,8 @@ function mpi_world_size {
     elif [ -n "$OMPI_COMM_WORLD_SIZE" ]; then
         echo "$OMPI_COMM_WORLD_SIZE"
     else
-        echo "1"
+        echo "Error: Unable to determine MPI world size. Neither SLURM_NTASKS nor OMPI_COMM_WORLD_SIZE is set." >&2
+        exit 1
```
pcastonguay: So are we implying here that we only support `trtllm-llmapi-launch` on Slurm clusters or with OpenMPI? Do we know if customers use it in different environments?

PR #8822, opened 2025-10-31
[None][perf] AutoDeploy optimize _get_unique_value

suyoggupta:
## Summary by CodeRabbit

* **Performance**
  * Optimized attention interface operations to improve efficiency and reduce memory overhead during value assignment calculations through streamlined lookup mechanisms.


lucaslie (approved):
Great to see that this is sufficient. 

I am still gonna keep #8631 open because it seems like we eventually we should do the right thing and just get that information from the cache manager

PR #8862, opened 2025-11-02
[https://nvbugs/5478158][fix] Disable paged context FMHA for INT4/AWQ + FP8 KV-cache in torch flow

danielafrimi:
INT4/AWQ (FP16/BF16 compute) + FP8 KV-cache + paged context FMHA requests a fused kernel meta that isn’t precompiled -> “FMHA kernels are not found” at context prefill, despite scales/KV-cache being correct.

root cause:
Fused context FMHA key: D=128, causal, Layout=Q_PAGED_KV (ragged), Q=FP16/BF16, KV=FP8.

disabling aged context FMHA make it works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of incompatible quantization configurations. When FP8 KV cache is used with W4A16 AWQ quantization and paged context FMHA, the system now logs a warning and automatically disables paged context FMHA to ensure stable operation.

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/trtllm.py
+++ b/tensorrt_llm/_torch/attention_backend/trtllm.py
@@ -6,6 +6,8 @@ from typing import TYPE_CHECKING, Optional, Tuple, Union
 
 import torch
 
+from tensorrt_llm.quantization.mode import QuantAlgo
+
 if TYPE_CHECKING:
     from ..speculative.utils import SpecDecodingTensor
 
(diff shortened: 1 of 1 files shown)


PerkzZheng:
>> Fused context FMHA key: D=128, causal, Layout=Q_PAGED_KV (ragged), Q=FP16/BF16, KV=FP8

I think it is not supported because the output dtype is BF16/FP16. we will convert Q to FP8 in this case, so it is not a problem.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1292,6 +1294,13 @@ def forward(
             # Context MLA uses separate qkv instead of paged_context_fmha
             use_paged_context_fmha = False
 
+        if self.quant_config.layer_quant_mode.has_fp8_kv_cache(
+        ) and self.quant_config.quant_algo == QuantAlgo.W4A16_AWQ and use_paged_context_fmha:
+            logger.warning(
+                "Disabling Paged context FMHA, as it is not supported with FP8 KV cache, for W4A16 AWQ quantization"
+            )
+            use_paged_context_fmha = False
```
PerkzZheng: IMO, I think we should throw out an error and let user disable chunked perfill and kv cache reuse (so paged context FMHA will not be enabled). I am not quite sure if adding a warning is suggested or not.

At the same time, we support FP8 QKV + BF16 output on blackwell, so this cannot be simply disabled. we check if it is on hopper or not.

brnguyen2:
👋 Thanks for this contribution! As part of a one-time effort to trim the TensorRT-LLM open-PR backlog, we're closing PRs that have had no activity in over 180 days and currently have merge conflicts with `main`, and this one qualifies. This isn't a reflection on the work itself: the branch, commits, and discussion are all preserved. If you'd like to revive it, just rebase onto `main` and reopen (or ping a maintainer) and we'll pick it back up. Thanks again!

PR #8931, opened 2025-11-05
[https://nvbugs/5556998][fix] init_hf_modules in worker_main for models with trust_remote=true

lancelly:
Similar to: https://github.com/vllm-project/vllm/pull/871

> If we load a model with trust_remote_code=True, hf will download model code or load local model code and cache dynamic model modules in its cache directory (it's usually in ~/.cache/huggingface/modules/). When a server starts, it will serialize some configs and send them to worker sub process. The model config contains dynamic modules imported from hf cache directory which is already added in sys.path by init_hf_modules function. So workers have to init_hf_modules once they start in case of ModuleNotFoundError.

After adding this initialization, locally running `trtllm-llmapi-launch trtllm-eval --trust_remote_code` with Kimi-K2(as described in https://nvbugs/5556998) no longer raises:
`ModuleNotFoundError: No module named 'transformers_modules'`

Interestingly, IT coverage seems to have already included `trtllm-llmapi-launch trtllm-eval --trust_remote_code` usage , yet no error occurred:
https://github.com/NVIDIA/TensorRT-LLM/blob/eeb56c2848a23bb60acd82c158d086c4305b249b/tests/integration/defs/test_e2e.py#L3407

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced initialization handling for models requiring remote code execution to ensure proper module setup during process startup and worker spawning.


tensorrt_llm/executor/worker.py
```diff
@@ -37,6 +37,28 @@
...
+        from transformers.dynamic_module_utils import init_hf_modules
+        init_hf_modules()
+        logger.debug("HF modules initialized")
+    except ImportError as e:
+        logger.warning(f"ImportError initializing HF modules: {e}")
+    except Exception as e:
+        logger.error(f"Exception initializing HF modules: {e}")
+
+
+_init_hf_modules()
```
Superjomn: Can you move this to base_worker.py? Currently, we have two other kinds of workers (Ray, RPC) those all reuse the code in base_worker.py.

Superjomn (approved):
LGTM

PR #9344, opened 2025-11-21
[None][feat] AutoDeploy: Precompute the A log for mamba layers

nvchenghaoz:
The A log is a per-layer parameter and it is a constant during inference. So precompute the -exp(A_log) during the compile time to have a better perf. 

## Summary by CodeRabbit

* **New Features**
  * Added new transform capability to the auto-deploy configuration system for model optimization.
  * Expands available graph transformation patterns during automated model deployment phases.
  * Provides additional optimization flexibility in the deployment workflow for compatible model types.


tensorrt_llm/_torch/auto_deploy/config/default.yaml
```diff
@@ -96,6 +96,8 @@ transforms:
   ############################################################################################
   # RUN POST-LOAD FUSION AND OPTIMIZATIONS
   ############################################################################################
+  fuse_mamba_a_log:
+    stage: post_load_fusion
```
suyoggupta: instead of adding it to default.yaml, how about adding this to examples/auto_deploy/nano_v3.yaml ?

tensorrt_llm/_torch/auto_deploy/transform/library/fuse_mamba_a_log.py
```diff
@@ -0,0 +1,140 @@
+"""Transform to fuse A_log into A for Mamba/NemotronH models."""
```
suyoggupta: looks like there is a memory leak. Could you also add a unit test that checks for mem usage before after this transformation

tensorrt_llm/_torch/auto_deploy/transform/library/fuse_mamba_a_log.py
```diff
@@ -0,0 +1,140 @@
...
+
+@TransformRegistry.register("fuse_mamba_a_log")
+class FuseMambaALog(BaseTransform):
+    """Fuse A_log parameter into A constant/parameter.
+
+    Replaces:
+        A = -torch.exp(self.A_log.float())
+    With:
+        A = self.A_fused
+    """
```
lucaslie: can we use the new pattern matcher - style replacement here? This is going to be really hard to maintain. Please check with @Fridah-nv if you have any questions

nzmora-nvidia (approved):

PR #9430, opened 2025-11-25
[None][chore] AutoDeploy add multi stream moe pass to default.yaml

suyoggupta:
## Release Notes

* **Chores**
  * Added new compile-stage transform configuration option (disabled by default) to expand optimization capabilities while maintaining backward compatibility.


nvchenghaoz (approved):

tensorrt_llm/_torch/auto_deploy/config/default.yaml
```diff
@@ -163,6 +163,9 @@ transforms:
   ############################################################################################
   fuse_causal_conv_activation:
     stage: compile
+  multi_stream_moe:
```
nvchenghaoz: As we discussed in my PR... I am not sure if we want to use the default.yaml for all models going forward. I think a better design - have a per model yaml + the shared yaml (default.yaml).. Not a blocker for this PR.

nvchenghaoz (approved):

PR #9509, opened 2025-11-27
[None][fix] Introduce inline namespace to avoid symbol collision

yihwang-nv:
## Description

This patch introduce a configuration header file `cpp/include/tensorrt_llm/common/config.h`, which was used to implicit add ABI namespace for tensorrt_llm namespace. It was useful to isolation at the ABI level, since other project(Eg. flashinfer) use TensorRT-LLM's kernel implementation but not update the code's namespace, it's caused a symbol collision issue between flashinfer and TensorRT-LLM. Therefore, we introduce this feature to fix this issue, and downstream projects which use TensorRT-LLM's kernel implementation can conveniently control the namespace by redefining macros.

After this patch:
- `tensorrt_llm::` -> `tensorrt_llm::_v1::`
- `torch_ext::` -> `tensorrt_llm::_v1::torch_ext::`

The major changes in this patch is:
Update the following code pattern
```cpp
namespace tensorrt_llm 
{
  // ....
} // namespace tensorrt_llm
```
into

```cpp
#include "tensorrt_llm/common/config.h"

TRTLLM_NAMESPACE_BEGIN
// ....
TRTLLM_NAMESPACE_END

```


eopXD (approved):
Looks good. Thank you.

QiJune (approved):

PR #9541, opened 2025-11-28
[None][fix] Introduce inline namespace to avoid symbol collision

yihwang-nv:
## Description

This patch is very similar with https://github.com/NVIDIA/TensorRT-LLM/pull/9509, but it's revert the changes for trtllm-gen cubin kernels, `runtime`, `executor`, `batch_manager`, `layers`, `ub`, `plugins`, `mpi`, `plugins` and `pg_utils` namespaces, we can resolve the symbol collision issue without modify the code under `cpp/tensorrt_llm/kernels/trtllmGenKernels` directory, this because `batchedGemm`, `gemm` and`gemmGatedAct` has a top-level additional namespace around `namespace tensorrt_llm` (Eg. `batchedGemm::tensorrt_llm::kernels`), and it can avoid namespace ambiguous. 

However, Fmha kernels do not have this additional namespace; the `tensorrt_llm` namespace is at the top level. We need some special handling to avoid namespace conflicts and resolve symbolic link issues.

We introduced a new file KernelMetaInfo.cpp, which defines some redundant data structures and enumeration types, and did not include any header files containing the inline namespace _v1(_v1 namespace is not a inline namespace in this file), thus avoiding cubins linking errors caused by namespace pollution.

This method allows us to significantly reduce the patch size, making reviews easier.

This patch introduce a configuration header file `cpp/include/tensorrt_llm/common/config.h`, which was used to implicit add ABI namespace for tensorrt_llm namespace. It was useful to isolation at the ABI level, since other project(Eg. flashinfer) use TensorRT-LLM's kernel implementation but not update the code's namespace, it's caused a symbol collision issue between flashinfer and TensorRT-LLM. Therefore, we introduce this feature to fix this issue, and downstream projects which use TensorRT-LLM's kernel implementation can conveniently control the namespace by redefining macros.

After this patch:
- `tensorrt_llm::` -> `tensorrt_llm::_v1::`
- `torch_ext::` -> `tensorrt_llm::_v1::torch_ext::`
- This changes exclude `tensorrt_llm::_v1::{runtime, batch_manager, executor, layers}`.
- The inner namespace of `ub::` (`userbuffer`) only change `tensorrt_llm::_v1::ub::kernels`.

The major changes in this patch is:
Update the following code pattern
```cpp
namespace tensorrt_llm 
{
  // ....
} // namespace tensorrt_llm
```
into

```cpp
#include "tensorrt_llm/common/config.h"

TRTLLM_NAMESPACE_BEGIN
// ....
TRTLLM_NAMESPACE_END

```

This patch also made changes for fmha_v2 kernel generation script `cpp/kernels/fmha_v2/setup.py`. All the symbols in generated *.cu files were in `tensorrt_llm::_v1::kernel` namespace. In order to avoid changes of `cpp/tensorrt_llm/kernels/contextFusedMultiHeadAttention/cubin`, we made a hack in the generated fmha_cubin.{h. cpp}:

fmha_cubin.h
```cpp
// copyright
#include "tensorrt_llm/common/config.h"

TRTLLM_NAMESPACE_BEGIN
namespace kernels {

struct FusedMultiHeadAttentionKernelMetaInfoV2 {
  ...
};

extern const FusedMultiHeadAttentionKernelMetaInfoV2 sMhaKernelMetaInfosV2[];
extern const int sMhaKernelMetaInfosV2Size;

} // namespace kernels
TRTLLM_NAMESPACE_END
```

fmha_cubin.cpp
```cpp
// copyright
#include "tensorrt_llm/common/config.h"

#include <cstddef>
#include <cstdint>
#include <cuda_runtime_api.h>

namespace tensorrt_llm::kernels {
Cubin arrays forward declarations...
Cubin lengths forward declarations...
} // namespace tensorrt_llm::kernels

namespace tensorrt_llm::TRTLLM_ABI_NAMESPACE::kernels {
Kernel launcher forward declarations...

// Duplicated constant declaration, to avoid include header.
constexpr int32_t kSM_* = *;
enum Data_type { ... };

struct FusedMultiHeadAttentionKernelMetaInfoV2 { ... };

extern const FusedMultiHeadAttentionKernelMetaInfoV2 sMhaKernelMetaInfosV2[] = {
Kernel metadata infos...
};
extern const int sMhaKernelMetaInfosV2Size = sizeof(sMhaKernelMetaInfosV2) / sizeof(sMhaKernelMetaInfosV2[0]);

} // namespace tensorrt_llm::TRTLLM_ABI_NAMESPACE::kernels

```


cpp/include/tensorrt_llm/common/config.h
```diff
@@ -0,0 +1,113 @@
...
+
+#if (defined(TRTLLM_NAMESPACE_PREFIX) || defined(TRTLLM_NAMESPACE_POSTFIX)) && !defined(TRTLLM_NAMESPACE_QUALIFIER)
+#error TRTLLM requires a definition of TRTLLM_NAMESPACE_QUALIFIER when TRTLLM_NAMESPACE_PREFIX/POSTFIX are defined.
+#endif
+
+/**
+ * \def TRTLLM_WRAPPED_NAMESPACE
+ * If defined, this value will be used as the name of a namespace that wraps the
+ * `tensorrt_llm::` namespace.
+ * If THRUST_TRTLLM_WRAPPED_NAMESPACE is set, this will inherit that macro's value.
```
tongyuantongyu: Cleanup all references to `thrust` and `cub` in this file.

yihwang-nv: Remove unnecessary macros and clean the references.

QiJune (approved):
LGTM

tongyuantongyu (approved):

hyukn (approved):
LGTM.

PR #9557, opened 2025-11-30
[None][fix] Set symbols default visibility to hidden to avoid symbol collision

yihwang-nv:
## Summary by CodeRabbit

* **New Features**
  * Added `getLevel()` and `setLevel()` methods to logger configuration for improved control over logging behavior.

* **Chores**
  * Applied consistent symbol visibility controls across the C++ library to properly support shared library exports on all platforms.
  * Updated CMake configuration to standardize how library symbols are exposed externally.

## Description

This patch set the default symbol visibility to hidden and introduce a configuration header file `cpp/include/tensorrt_llm/common/config.h` defines `TRTLLM_API` to allow developers explicit export symbols.

Since other project(Eg. flashinfer) use TensorRT-LLM's kernel implementation but not update the code's namespace, it's caused a symbol collision issue between flashinfer and TensorRT-LLM. We can hide symbols to avoid symbol collision.


QiJune (approved):
LGTM

Issue #9432, opened 2025-11-25
[Bug]: NameError in memory profiler on Jetson device.


bonginn:
### System Info

- **Hardware:** Jetson Orin NX (16GB)
- **OS:** L4T 36.4.0 (JetPack 6.1)
- **CUDA Version:** 12.6
- **TensorRT Version:** 10.3.0
- **TensorRT-LLM version:** 0.12.0
- **Python Version:** 3.10

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

1. Setup environment on a Jetson Orin NX device (JetPack 6.x).
2. Install `nvidia-ml-py` to support NVML calls (or to suppress the `pynvml` missing warning).
3. Run the python benchmark script:    
```bash
python3 benchmarks/python/benchmark.py [and your arguments]
```

4. Observe the traceback in the console output.
```bash
Traceback (most recent call last):

  File "/home/jetson/miniconda3/envs/trtllm-tiny/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap

    self.run()

  File "/home/jetson/miniconda3/envs/trtllm-tiny/lib/python3.10/multiprocessing/process.py", line 108, in run

    self._target(*self._args, **self._kwargs)

  File "/home/jetson/project/TensorRT-LLM/benchmarks/python/mem_monitor.py", line 72, in _upd_peak_memory_usage

    peak_host_used, peak_device_used = self.get_memory_usage()

  File "/home/jetson/project/TensorRT-LLM/benchmarks/python/mem_monitor.py", line 84, in get_memory_usage

    device_used, _, _ = device_memory_info()

  File "/home/jetson/miniconda3/envs/trtllm-tiny/lib/python3.10/site-packages/tensorrt_llm/profiler.py", line 163, in device_memory_info

    mem_info = _device_get_memory_info_fn(handle)

NameError: name '_device_get_memory_info_fn' is not defined
```

### Expected behavior

The profiler should handle the Jetson platform gracefully. Since `_device_get_memory_info_fn` is skipped during initialization on Jetson, the `device_memory_info` function should check for the platform and return early (or initialize correctly), rather than attempting to call an undefined function.

### actual behavior

[...]

PR #9610, opened 2025-12-02
[#9432][fix] Resolve NameError in memory profiler for v0.12.0-jetson branch.

bonginn:
## Description
This PR fixes the issue #9432 specifically for the `v0.12.0-jetson` branch.

On Jetson platforms (where `on_jetson_l4t` is True), the memory profiler currently throws a `NameError` because `_device_get_memory_info_fn` is undefined. This occurs specifically when `nvidia-ml-py` (or `pynvml`) is installed, causing the benchmark script's background monitoring process to fail. This change is necessary to allow benchmarks to run correctly on edge devices without crashing.

## Summary of Changes
This PR modifies `tensorrt_llm/profiler.py` to add a guard clause in the `device_memory_info` function. It explicitly checks if the platform is Jetson (`on_jetson_l4t`). If true, it returns early (reporting 0 memory usage) and logs a one-time warning to inform the user that memory monitoring is skipped, preventing the `NameError` and avoiding log flooding.

### DIFF
--- a/tensorrt_llm/profiler.py
+++ b/tensorrt_llm/profiler.py
@@ -154,6 +154,14 @@ def host_memory_info(pid: Optional[int] = None) -> Tuple[int, int, int]:
 def device_memory_info(
         device: Optional[Union[torch.device,
                                int]] = None) -> Tuple[int, int, int]:
+    if on_jetson_l4t:
+        if not getattr(device_memory_info, "_has_logged_jetson_warning", False):
+            logger.warning(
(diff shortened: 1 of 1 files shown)


karljang:
@bonginn , thanks for your contribution!
@Shixiaowei02 , could you help review this PR?  Please note that it targets the old branch for Jetson tests, not the main branch.

karljang (approved):
LGTM; I'll merge it with the old jetson branch~
Thank you for your contribution

PR #9693, opened 2025-12-04
[None][fix] Fix MTP rewind corner case

axxx03:
# Description
When using overlap scheduler with MTP, the request may have been terminated and removed from mSequences before rewindKVCache is called, we should avoid this.

### DIFF
--- a/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
+++ b/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
@@ -2909,6 +2909,18 @@ void KVCacheManager::removeToken(RequestIdType requestId)
 
 void KVCacheManager::rewindKVCache(RequestIdType requestId, SizeType32 rewindLengths)
 {
+    // Check if the sequence still exists before rewinding
+    // In overlap mode with MTP, the request may have been terminated and removed
+    // from mSequences before rewindKVCache is called
+    {
+        std::scoped_lock lck(mSequencesMtx);
+        if (mSequences.find(requestId) == mSequences.end())
+        {
+            TLLM_LOG_DEBUG("Request %lu has already been removed from KV cache manager, skipping rewind", requestId);
+            return;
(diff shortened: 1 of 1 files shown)


Shixiaowei02 (approved):

eopXD (commented):
Slightly confused about the corner case here.

Rewind is only called if the generation request is not in a complete state. This should be a hard assertion rather than returning with only a warning.

```
        # rewind kv cache
        for request in scheduled_batch.generation_requests:
            if request.state != LlmRequestState.GENERATION_COMPLETE:
                if request.py_rewind_len > 0:
                    self.rewind_kv_cache(request, request.py_rewind_len)
```

May I ask what specific case have you encountered that requires the code change of this MR?

axxx03:
> Slightly confused about the corner case here.
> 
> Rewind is only called if the generation request is not in a complete state. This should be a hard assertion rather than returning with only a warning.
> 
> ```
>         # rewind kv cache
>         for request in scheduled_batch.generation_requests:
>             if request.state != LlmRequestState.GENERATION_COMPLETE:
>                 if request.py_rewind_len > 0:
>                     self.rewind_kv_cache(request, request.py_rewind_len)
> ```
> 
> May I ask what specific case have you encountered that requires the code change of this MR?

Hi, this corner case also happens when using TensorRT-LLM with FlexKV. 

In this case, when finishing kv cache offloading, the correspoding resources will be released. However, when combing with overlap scheduling + MTP, in the overlap step, if the MTP verification fails, a KV cache rewind may be triggered. At this point, the KV cache has already been released due to the earlier offloading, so the rewind operation attempts to access non-existent KV cache entries, which leads to an IndexError.

jthomson04:
@axxx03 Can you please run pre-commit and push the changes?

pcastonguay (approved):

dagil-nvidia:
@axxx03 it's failing pre-commit check

PR #10127, opened 2025-12-18
[TRTLLM-10591][feat] Add programmatic model-feature support matrix & runtime auto-fallback support

venkywonka:
## Summary

This PR introduces a programmatic model support matrix (`model_support_matrix.py`) that serves as the single source of truth for model-feature compatibility in TensorRT-LLM. It also introduces an associated LLMAPI arg `auto_disable_unsupported_features` to toggle the auto-fallback behavior.

### Motivation
* Currently we don’t have a programmatic way to track which features are supported by which models. We expect the user to configure them instead of giving an OOTB optimal config by referring to the docs [here](https://nvidia.github.io/TensorRT-LLM/models/supported-models.html#model-feature-support-matrix-key-models).
* For example:
  * QwenNext will break with the default setting of enable_block_reuse=True , so user has to disable explictly.
  * DSv3.2 and VLMs will break with `enable_chunked_prefill=True` even though for all the other models, enabling it as a default would result in OOTB long-context support for free.
* The PR attempts to make a subset of inference features that are clearly mapped to an LLMAPI argument (chunked prefill, kv cache reuse, cuda graphs, ... etc.) programmatically tracked - so that one can toggle intelligent fallback in LLMAPI usage (when that feature is enabled by default, but not supported in that specific model), as well as autogeneration of the [model support matrix](https://nvidia.github.io/TensorRT-LLM/models/supported-models.html#model-feature-support-matrix-key-models) in documentation. 
* In the future, this would enable us to make auto-fallback behavior default and then go ahead and confidently enable certain LLMAPI flags globally that would give good OOTB performance for the majority of models, without breaking the models that don't support it. Specifically,
  * `enable_chunked_prefill` can be enabled `True` globally by default, so all models that support it can benefit from long-context support - while models that don't support it can auto-fallback.
  * `enable_block_reuse=True` (currently default) will not break QwenNext or other models that don't support it.

### Things to note
* These changes are not breaking. the added param `auto_disable_unsupported_features` is set to False by default to be safe and not change the behavior of existing workloads (what if some downstream workload breaks when trtllm runtime auto-disables an llmapi flag because it wasn't supported in the model)? One can always toggle them later to a breaking change.
* Any new model or feature support update to the docs now has to be done in-code, and the docs will autogenerate it. Direct modifications to the `supported-models.md` file will break the test that checks for parity with script.
 
### PR items:
  1. Optional runtime feature gating - Automatically disable unsupported features based on model architecture, if `auto_disable_unsupported_features=True`. Else just a warning and continue with current behavior (failure).
  2. Documentation generation - Auto-generate `docs/source/models/supported-models.md` from the matrix
  3. Sync validation - CI check to ensure docs stay in sync with the programmatic matrix

### Test Plan

  - Unit tests for all fallback scenarios (`test_model_feature_fallbacks.py`)
  - Docs sync validation test (`test_supported_models_sync.py`)

## Summary by CodeRabbit

* **New Features**
  * Added `auto_disable_unsupported_features` flag to automatically disable unsupported model features with warnings instead of errors.
  * Introduced comprehensive model support matrix tracking feature compatibility across architectures.

* **Documentation**
  * Updated Supported Models documentation with auto-generated content reflecting current architecture and feature support.

## PR Checklist

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


tensorrt_llm/llmapi/model_support_matrix.py
```diff
@@ -0,0 +1,576 @@
...
+
+    EPD_DISAGG_SERVING = "EPD Disaggregated Serving"
+    MODALITY = "Modality"
+
+
+@dataclass(frozen=True)
+class SupportedModel:
+    architecture: str
+    model: str
+    huggingface_example: str
```
venkywonka:
```suggestion
    huggingface: str
```

tensorrt_llm/llmapi/llm.py
```diff
@@ -252,6 +254,55 @@ def __init__(self,
         exception_handler.register(self, 'shutdown')
         atexit.register(LLM._shutdown_wrapper, weakref.ref(self))
 
+    def _apply_model_feature_fallbacks(self) -> None:
```
anish-shanbhag: Could we move this logic to a `model_validator` of `BaseLlmArgs` / `TorchLlmArgs` / etc.?

We could define one validator in `BaseLlmArgs` for the common attributes, and then additional validators in `TorchLlmArgs` / `TrtLlmArgs` etc. for the backend-specific validators. Ideally this would mean we avoid the use of `getattr` entirely.

tensorrt_llm/llmapi/model_support_matrix.py
```diff
@@ -0,0 +1,562 @@
...
+    ),
+    SupportedModel(
+        architecture="DeciLMForCausalLM",
+        model="Nemotron",
+        huggingface="nvidia/Llama-3_1-Nemotron-51B-Instruct",
+    ),
+    SupportedModel(
+        architecture="DeepseekV3ForCausalLM",
+        model="DeepSeek-V3",
+        huggingface="deepseek-ai/DeepSeek-V3",
```
anish-shanbhag: How will this currently work with quantized variants like `Deepseek-R1-FP4`?

tensorrt_llm/llmapi/model_support_matrix.py
```diff
@@ -0,0 +1,562 @@
...
+    Feature.TORCH_SAMPLER,
+    Feature.TLLM_CPP_SAMPLER,
+    Feature.KV_CACHE_REUSE,
+    Feature.SLIDING_WINDOW_ATTENTION,
+    Feature.LOGITS_POST_PROCESSOR,
+    Feature.GUIDED_DECODING,
+)
+
+KEY_MODEL_MATRIX: Mapping[str, Mapping[Feature, FeatureCell]] = {
+    "DeepseekV3ForCausalLM": {
```
anish-shanbhag: It looks like the current assumption is that all models under the same architecture will have the same feature support - is this always true?

tensorrt_llm/llmapi/model_support_matrix.py
```diff
@@ -0,0 +1,562 @@
...
+        architecture="Qwen3MoeForCausalLM",
+        model="Qwen3MoE",
+        huggingface="Qwen/Qwen3-30B-A3B",
+    ),
+    SupportedModel(
+        architecture="Qwen3NextForCausalLM",
+        model="Qwen3Next",
+        huggingface="Qwen/Qwen3-Next-80B-A3B-Thinking",
+    ),
+)
```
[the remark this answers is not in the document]
Superjomn: It may be even better to reuse the existing [register_auto_model](https://github.com/Superjomn/TensorRT-LLM/blob/fix-create-engine-timeout/tensorrt_llm/_torch/models/modeling_utils.py#L628), which is used to register a new model arch. Just adding the new fields and applying the changes for all the arches.

tensorrt_llm/llmapi/model_support_matrix.py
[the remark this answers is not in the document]
venkywonka: yea, unfortunately - but i believe the manual effort to modify the code is not too much, about the same as the manual effort to modify the docs would you agree?

tensorrt_llm/llmapi/model_support_matrix.py
```diff
@@ -0,0 +1,562 @@
...
+    Feature.EAGLE3_TWO_MODEL_ENGINE,
+    Feature.TORCH_SAMPLER,
+    Feature.TLLM_CPP_SAMPLER,
+    Feature.KV_CACHE_REUSE,
+    Feature.SLIDING_WINDOW_ATTENTION,
+    Feature.LOGITS_POST_PROCESSOR,
+    Feature.GUIDED_DECODING,
+)
+
+KEY_MODEL_MATRIX: Mapping[str, Mapping[Feature, FeatureCell]] = {
```
Superjomn: How about scatter this large table to `register_auto_model`? Like upgrade it to support:
```python
supported_features = [
    Feature.OVERLAP_SCHEDULER,
    ...
]

@register_auto_model("LlamaForCausalLM", features=supported_features, ...other_fields)
class LlamaForCausalLM(...):
```

The advantage is that registration and updates are already handled within each model, with model PICs tracking them. This way, we avoid duplicating model architecture entries.

cc @QiJune, what do you think?

tensorrt_llm/llmapi/model_support_matrix.py
```diff
@@ -0,0 +1,562 @@
...
+
+    lines: List[str] = []
+    lines.append(_row(headers))
+    lines.append(_sep())
+    for row in rows:
+        lines.append(_row(row))
+    return "\n".join(lines)
+
+
+def render_supported_models_markdown() -> str:
```
Superjomn: A minor suggestion—perhaps overkill—but using a Jinja2 template could help clarify the document structure. The headings and some parts of the content are fixed, and we're essentially just inserting content into predefined fields.

tests/unittest/api_stability/references/llm.yaml
```diff
@@ -231,6 +231,10 @@ methods:
         annotation: int
         default: 1000
         status: prototype
+      auto_disable_unsupported_features:
```
Superjomn: How about shortening this to `ignore_unsupported_features` , I am not sure if "auto" + "disable" = "ignore"?

tensorrt_llm/llmapi/llm.py
anish-shanbhag: That's a good point - might end up adding unnecessary complexity.

Another option is to not use a `model_validator` but instead add something like `def update_from_feature_matrix(architecture: str)` to each of the `LlmArgs` classe, which would update the right attributes using the `architecture` that's passed in.

I think my main initial concern was just with the `getattr` / `hasattr` logic which could end up being harder to maintain.

tensorrt_llm/llmapi/model_support_matrix.py
venkywonka: But practically, for example, say a feature is marked as supported for `DeepseekV3ForCausalLM`, and is supported for `DeepSeekV3` but is not supported in `DeepSeek-R1-0528-NVFP4-v2` - then the auto-fallback will miss this and we're atleast as good as today.

PR #10166, opened 2025-12-19
[None][fix] [fix] Make NCCL resource manager destructor exception-safe

nv-lschneider:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability during application shutdown by strengthening exception handling in resource cleanup sequences.
  * Enhanced robustness of cleanup operations to prevent crashes in static destruction scenarios through safer locking patterns and guarded exception handling.
  * Made logging operations during teardown more resilient to prevent failure propagation during shutdown procedures.

## Description

It was reported that the clean up of NCCL resources is not always clean and could end in segfaults.

I am struggling to reproduce this exact error.
But upon inspection of the code, I can see why they is a potential problem when tearing down the static ResourceManager.

In particular the unprotected access of the Logger is potentially the issue here.

I updated the code, to be a lot more conservative in the tear-down process to prevent problem in the future.

## Test Coverage

No new test needed.

## PR Checklist


kmk142789 (approved):

hyukn (approved):

hyukn:
Maybe we should add multi-gpu stages for more coverage?

PR #10330, opened 2025-12-29
[TRTLLM-10171][fix] Correct attention handling in ModelConfig and KVCacheManager

jaedeok-nvidia:
## Summary

**[Bug fix]** Fixed OOM issue of VSWA KV cache manager when enabling ADP of VSWA models.

## Description

This PR fixes the model config binding so KVCacheManager can compute required cache blocks with accurate head and hidden sizes. 

The Attention DP logic hasn't been properly propagated to KvCacheManager. [get_bindings_model_config](https://github.com/NVIDIA/TensorRT-LLM/blob/965578ca219d2329d296714dd5d71cb9881e3e2a/tensorrt_llm/_torch/model_config.py#L498-L499) manually calculates the number of local attention heads in various places, but it hasn't account attn_tp_size and attn_cp_size from the parallelism extensions, as well as the behavior of attention data parallel.

For Attention DP, attn_tp_size should effectively be 1. However, the current logic calculates it using the global TP size (e.g., 8), resulting in smaller number of local attention heads and causing the system to attempt allocating an excessive number of blocks. So it leads to OOM if config contains `max_attention_window` explicitly, e.g.,
```yaml
kv_cache_config:
    max_attention_window: [128, 128, 1000]
```

Changes:
 - Updated hidden size and number of attention head calculations with the correct attention TP and CP sizes.
 - Added enable_attention_dp parameter to KVCacheManager for the correct resource management.

### DIFF
--- a/tensorrt_llm/_torch/model_config.py
+++ b/tensorrt_llm/_torch/model_config.py
@@ -495,10 +495,15 @@ class ModelConfig(Generic[TConfig]):
         # TODO smor- currently assuming no rnn layers, no MOE
         from tensorrt_llm.bindings import ModelConfig as ModelConfigCpp
 
-        num_heads = self.pretrained_config.num_attention_heads // (
-            self.mapping.tp_size * self.mapping.cp_size)
+        # Attention DP should not shard attention heads; use attn_tp_size=1 in that case
+        # so downstream KV calculations see the full (non-partitioned) head count.
+        attn_tp_size = self.mapping.attn_tp_size if not self.mapping.enable_attention_dp else 1
+        attn_cp_size = self.mapping.attn_cp_size
 
-        hidden_size = self.pretrained_config.hidden_size // self.mapping.tp_size
+        num_heads = self.pretrained_config.num_attention_heads // (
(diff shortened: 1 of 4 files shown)


tensorrt_llm/_torch/model_config.py
```diff
@@ -495,10 +495,15 @@ def get_bindings_model_config(self,
         # TODO smor- currently assuming no rnn layers, no MOE
         from tensorrt_llm.bindings import ModelConfig as ModelConfigCpp
 
+        # Attention DP should not shard attention heads; use attn_tp_size=1 in that case
+        # so downstream KV calculations see the full (non-partitioned) head count.
+        attn_tp_size = self.mapping.attn_tp_size if not self.mapping.enable_attention_dp else 1
+        attn_cp_size = self.mapping.attn_cp_size
```
eopXD: Is there a default value for this?

eopXD (approved):
LGTM. We should have test coverage and add it into L0.

dongxuy04 (approved):

PR #10522, opened 2026-01-07
[None][chore] update flashinfer to 0.6.0

nvchenghaoz:
Update the FI to 0.6.0 for TRTLLM.


suyoggupta (approved):

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_attention.py
```diff
@@ -523,7 +373,7 @@ def flashinfer_mha_with_cache(
             num_seq=num_prefill,
             page_size=k_cache.shape[1],
             q_dtype=q_prefill.dtype,
-            kv_dtype=k_cache.dtype,
+            kv_dtype=q_prefill.dtype if k_cache.dtype == torch.float8_e4m3fn else k_cache.dtype,
```
lucaslie: how about `v.dtype`

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_attention.py
```diff
@@ -535,10 +385,17 @@ def flashinfer_mha_with_cache(
             kv_lens_arr_host=seq_len_with_cache_host[:num_prefill],
             plan_params=pp_prefill,
         )
+        # Convert FP8 KV cache to BF16 for compute (FlashInfer 0.6.0 doesn't support mixed Q=BF16, KV=FP8)
+        if k_cache.dtype == torch.float8_e4m3fn:
+            k_cache_compute = k_cache.to(q.dtype)
+            v_cache_compute = v_cache.to(q.dtype)
+            kv_cache_for_run = (k_cache_compute, v_cache_compute)
+        else:
+            kv_cache_for_run = (k_cache, v_cache)
```
lucaslie: this is the entire kv-cache with all pages. We cannot convert that every time. That's a massive perf and memory hit. 

I also cannot imagine that flashinfer doesn't support mixed. Otherwise, fp8 is basically useless. This is a blocker from merging the PR

lucaslie (changes requested):

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_attention.py
suyoggupta: can we check if using fa2 backend works?

juney-nvidia (approved):

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_attention.py
suyoggupta: also, looks like the limitation only exists on Hopper (and not for blackwell.. ) so if this change is absolutely necessary, we should limit the perf impact to Hopper only

tensorrt_llm/_torch/auto_deploy/custom_ops/flashinfer_attention.py
lucaslie: but are you using `q_prefill.dtype`?

suyoggupta (approved):

security_scanning/pyproject.toml
```diff
@@ -57,7 +57,7 @@ ordered-set = "^4.1.0"
 peft = "^0.18.1"
 patchelf = "^0.17.2.4"
 einops = "^0.8.1"
-flashinfer-python = ">=0.3.0,<0.4.0"
+flashinfer-python = "^0.6.0"
```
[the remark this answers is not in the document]
yibinl-nvidia: @nv-guomingz what is the difference between ^0.6.0 and ~0.6.0? Gemini said they are pretty much equal for < 1.0.0?
Screenshot 2026-01-15 at 10 22 05

brb-nv (approved):

lucaslie (approved):

PR #10646, opened 2026-01-13
[None][fix] Fix FakeTensorMode mismatch in pattern matcher for FP8 models

karthikvetrivel:
# [None][fix] Fix FakeTensorMode mismatch in pattern matcher for FP8 models

## Release Notes

* **Bug Fixes**
  * Enhanced the model export and compilation process to properly support additional tensor input types and configurations during the tracing phase, improving reliability and robustness when deploying models across diverse use cases, environments, and computational backends while reducing potential failures in edge scenarios.

### Problem

FP8 models like `nvidia/Llama-3.1-8B-Instruct-FP8` failed during the `fuse_fp8_linear` transform with the following error:

```
AssertionError: fake mode (<torch._subclasses.fake_tensor.FakeTensorMode object at 0x...>) 
from fake tensor input 0 doesn't match mode (<torch._subclasses.fake_tensor.FakeTensorMode object at 0x...>) 
from fake tensor input 1
```

### Root Cause

During pattern matching in `fuse_quant.py`, PyTorch's pattern matcher calls `trace_fn(replace_fn, args)` to trace replacement patterns. The `args` are `FakeTensor` objects from the model's `FakeTensorMode`. However, `_trace_to_gm` was calling `torch_export_to_gm`, which creates a **new** `FakeTensorMode` context. This caused PyTorch's `detect_fake_mode` to assert that all fake tensors must come from the same mode.

**Call stack:**
```
fuse_quant.py:289 → patterns.apply()
  → torch/_inductor/pattern_matcher.py:1982 → entry.extra_check(m)
    → pattern_matcher.py:1437 → detect_fake_mode(args)  ← CONFLICT
      → pattern_matcher.py:1520 → trace_fn(replace_fn, args)
        → _trace_to_gm → torch_export_to_gm()  ← CREATES NEW MODE
```

### Solution

Modified `_trace_to_gm` in `pattern_matcher.py` to detect if input tensors are `FakeTensor`s and use `make_fx` (which respects the existing fake mode context) instead of `torch_export_to_gm` (which creates a new context).

```python
def _trace_to_gm(fn: Callable, args: Sequence[torch.Tensor]) -> GraphModule:
    """Exports a function or Module into a GraphModule via torch_export_to_gm."""
    from torch._guards import detect_fake_mode
    from torch.fx.experimental.proxy_tensor import make_fx

    module = fn if isinstance(fn, torch.nn.Module) else _WrapperModule(fn)

    # Use make_fx for FakeTensors to avoid FakeTensorMode mismatch during pattern matching
    if detect_fake_mode(args) is not None:
        return make_fx(module, tracing_mode="fake")(*args)

    return torch_export_to_gm(module, tuple(args))
```

## Test Coverage

Verified on H100 system with:

```bash
python3 examples/auto_deploy/build_and_run_ad.py \
  --model nvidia/Llama-3.1-8B-Instruct-FP8 \
  --args.yaml-extra examples/auto_deploy/model_registry/configs/dashboard_default.yaml \
  --args.yaml-extra examples/auto_deploy/model_registry/configs/world_size_2.yaml
```

**Results:**
- `fuse_fp8_linear`: 224 pattern matches (previously failed)
- `fuse_allreduce_residual_rmsnorm`: 64 pattern matches
- Model inference produces coherent outputs

## PR Checklist

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/utils/pattern_matcher.py
+++ b/tensorrt_llm/_torch/auto_deploy/utils/pattern_matcher.py
@@ -81,10 +81,23 @@ class _WrapperModule(torch.nn.Module):
 
 
 def _trace_to_gm(fn: Callable, args: Sequence[torch.Tensor]) -> GraphModule:
-    """
-    Exports a function or Module into a GraphModule via torch_export_to_gm.
-    """
+    """Exports a function or Module into a GraphModule via torch_export_to_gm."""
+    import torch.utils._python_dispatch as pd
+    from torch._guards import detect_fake_mode
+
     module = fn if isinstance(fn, torch.nn.Module) else _WrapperModule(fn)
+
+    # Convert FakeTensors to meta tensors to avoid mode mismatch with torch.export
+    if detect_fake_mode(args) is not None:
+        with pd._disable_current_modes():
+
+            def _to_meta(t):
+                if not isinstance(t, torch.Tensor):
+                    return t
+                return torch.empty(tuple(int(s) for s in t.shape), dtype=t.dtype, device="meta")
+
+            args = tuple(_to_meta(t) for t in args)
+
     return torch_export_to_gm(module, tuple(args))
 
 


tensorrt_llm/_torch/auto_deploy/utils/pattern_matcher.py
```diff
@@ -84,7 +84,15 @@ def _trace_to_gm(fn: Callable, args: Sequence[torch.Tensor]) -> GraphModule:
...
     Exports a function or Module into a GraphModule via torch_export_to_gm.
     """
+    from torch._guards import detect_fake_mode
+    from torch.fx.experimental.proxy_tensor import make_fx
+
     module = fn if isinstance(fn, torch.nn.Module) else _WrapperModule(fn)
+
+    # Use make_fx for FakeTensors to avoid FakeTensorMode mismatch during pattern matching
+    if detect_fake_mode(args) is not None:
+        return make_fx(module, tracing_mode="fake")(*args)
```
lucaslie: @Fridah-nv , please take a look

Fridah-nv (commented):
This is an interesting fix, thank you!
A while ago when I tried to use `make_fx` as the `trace_fn` to trace python code, I found that the graph output by `make_fx` is different from `torch_export_to_gm`/`torch.export.export`,  I wonder if you observe something similar?
Since our initial graph is produced by `torch_export_to_gm`, there could be more matching failure if the pattern graph is generated with `make_fx` and be slightly different from the actual pattern in the target graph.
Please double check if this can be an issue for other pattern matchers, for `fuse_fp8_linear` it should be fine as it only contains one op in the pattern.

karthikvetrivel:
@Fridah-nv Thanks for taking a look! I think you're right—if we need to maintain strict graph semantics, then `make_fx` could eventually give us a different graph. I think this could be a band-aid fix for now to make models work where the replacement patterns are simple single-op substitutions (like `fuse_fp8_linear`) and have this fake tensor issue.  These models wouldn't work w/ AutoDeploy as-is now.  

I was exploring a few larger ideas, but they didn't seem worth the complexity. I'm open to exploring other ideas though and would love to hear your thoughts.

Fridah-nv (approved):

karthikvetrivel:
@Fridah-nv Yes right, good catch. WIll do after another approval.

brnguyen2:
👋 As part of an effort to reduce the TensorRT-LLM open-PR backlog, we're checking in on PRs with no activity in over 120 days. This one qualifies.

Could you let us know whether you still plan to land it?

- **If yes:** please rebase onto the latest `main`, resolve any merge conflicts, and re-request review (or just leave a comment with where it stands).
- **If it's no longer needed:** you can close it, or let us know and we'll close it for you.

If we don't hear back within 14 days, we'll close this PR to keep the review queue manageable. Closing isn't a rejection: the branch, commits, and discussion are all preserved, and you can reopen or resubmit at any time.

Thanks for the contribution!

brnguyen2:
👋 Closing this as part of the TensorRT-LLM open-PR backlog cleanup. We pinged on 2026-08-10 about 120+ days of inactivity and didn't hear back within the 14-day window.

This isn't a rejection — your branch, commits, and discussion are all preserved. If it's still relevant, just **reopen the PR** any time and we'll pick it back up; please rebase onto the latest `main` when you do. Thanks again for the contribution!

PR #10867, opened 2026-01-21
[None][feat] AutoDeploy: Flashinfer kernels bringup

nvchenghaoz:
1. Update the attention prefill kernel to use the "auto" mode when the GPU is not H100
2. Remove some redundant test debug print
3. Major update: Introduced the SSM kernel from Flashinfer and add a new field in the yaml file to specify the kernel. {The main reason to add this instead of introducing a new one is that Flashinfer only has this decode kernel. Adding a new FI backend will introduce many duplicate code and hard to maintain. In the future if the FI supports all prefill / decode / etc, we can consider adding a new FI backend. Given this change, the next step is to rename triton in the file name and the source code.}. Update: Per offline discussion, splitted the ssm into two backends: Flashinfer and triton. Move most code into a shared file. 

## Summary by CodeRabbit

* **New Features**
  * Added configurable decode kernel selection (Triton or FlashInfer) for SSM attention operations to optimize performance based on hardware capabilities.
  * Implemented automatic backend optimization that selects the best kernel variant for your GPU architecture.

* **Tests**
  * Removed unnecessary debug logging from test suite.


tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/triton_backend_mamba.py
```diff
@@ -205,19 +206,40 @@ def _triton_cached_ssm(
...
-            C_decode,
-            D=D_full,
-            z=None,
-            dt_bias=dt_bias_hp,
-            dt_softplus=True,
-            state_batch_indices=slot_idx_decode,
-        )  # [nd, H, D]
+        # Use decode_kernel parameter to select backend
+        if decode_kernel == "flashinfer":
+            import flashinfer
```
suyoggupta: feels hacky.. we're basically handling the if-else inside the custom- op.. but a more sound approach may be to insert the correct custom op depending on the chosen backend?

lucaslie (commented):
please add a separate attention backend for flashinfer ssm kernels

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/mamba_backend_common.py
```diff
@@ -0,0 +1,309 @@
+from typing import List, Optional, Tuple
```
nvchenghaoz: Add the copyright to this file

lucaslie (approved):

PR #11035, opened 2026-01-27
[TRTLLM-10845][feat] Add dynamic llmapi defaults system

venkywonka:
**Motivation and Design:** #11083

**Summary:**
This PR introduces a dynamic model-specific defaults system that allows models to declare their required configurations, eliminating the need for users and tests to manually specify model-specific constraints.

**Problem:**
Previously, users had to manually configure model-specific settings like disabling KV cache block reuse for SSM models. This was error-prone and required deep knowledge of model architectures.

**Solution:**
- Added `get_model_defaults()` classmethod to model classes
- ModelLoader automatically applies these defaults before model initialization
- User overrides are still respected

Explicitly disabling `enable_block_reuse` will no longer be necessary for below models:
- Qwen3Next
- NemotronH

**Benefits:**
- A way for models to be tweaked to work correctly out-of-the-box
- Self-documenting model requirements
- User overrides still possible when needed

**Other changes**
- Cleaned up the way `trtllm-serve` applies CLI overrides to prevent pydantic defaults from getting contaminated by serve-frontend's own defaults (which are redundant and same as pydantic defaults) 
- Moved `fail_fast_on_attention_window_too_large` attribute to `TrtLlmArgs` from `BaseLlmArgs`

## Release Notes

* **New Features**
  * Added support for `LlavaLlamaForCausalLM` model type alias.
  * Enhanced model architecture resolution with improved handling for specialized model variants.
  * Implemented automatic model-specific configuration defaults for optimized performance.

* **Improvements**
  * Improved model initialization flow with better configuration management.
  * Extended test coverage for model defaults and argument handling.


tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -6,19 +6,20 @@
 import weakref
```
anish-shanbhag: FYI, my comment at https://github.com/NVIDIA/TensorRT-LLM/pull/11035/changes#r2761505427 might make some of the other comments in this file irrelevant, if we refactor along that route

anish-shanbhag (commented):
Great progress toward improving the OOTB experience!

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -338,6 +329,25 @@ def allocation_scope(current_stage: ExecutorMemoryType,
                     vm_pools[stage] = memory_pool
                     yield
 
+    from tensorrt_llm._torch.pyexecutor.model_loader import ModelLoader, _construct_checkpoint_loader
+
+    # Apply model-specific defaults before creating AttentionRuntimeFeatures
+    checkpoint_loader = _construct_checkpoint_loader(llm_args.backend,
```
QiJune: I noticed that `ModelEngine` instantiates its own checkpoint loader. To avoid the overhead of constructing it twice, could we update the `ModelEngine` constructor to accept an optional checkpoint_loader argument? That way, we can reuse the existing instance. 

For example:

```
    model_engine = PyTorchModelEngine(
        model_path=checkpoint_dir,
        llm_args=llm_args,
        mapping=mapping,
        attn_runtime_features=attn_runtime_features,
        dist=dist,
        spec_config=spec_config,
        # OPTIMIZATION: Pass the existing loader to avoid re-creating it
        checkpoint_loader=checkpoint_loader, 
    )
```

tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -484,6 +488,20 @@ def _load_model_from_hf(self):
...
 
+        # Apply model-specific defaults if the model class provides them
+        if hasattr(model_cls, 'get_model_defaults'):
+            model_defaults = model_cls.get_model_defaults(self.llm_args)
+
+            if model_defaults:
+                model_defaults = validate_model_defaults(
+                    model_defaults, self.llm_args)
+
+                applied_defaults = apply_model_defaults_to_llm_args(
```
leslie-fang25: why we need another apply here, feels like it will also hit the `apply_model_defaults_to_llm_args` invoked from `py_executor_creator` afterwards?

tensorrt_llm/commands/serve.py
```diff
@@ -86,6 +87,40 @@ def _signal_handler_cleanup_child(signum, frame):
...
+        "postprocess_tokenizer_dir"
+    }
+
+    if param_name in always_include:
+        return True
+
+    if value is None:
+        return False
+
+    llm_args_class = TrtLlmArgs if backend == "tensorrt" else TorchLlmArgs
```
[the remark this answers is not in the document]
venkywonka: i'll add a note about that here - but since auto-deploy's `LlmArgs` inherits from `TorchLlmArgs` - the code path currently is what is expected

tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -985,4 +1004,208 @@ class LlmBuildStats:
...
+    merged_args = merge_llm_configs_with_defaults(llm_args_dict,
+                                                  model_defaults,
+                                                  user_overrides=user_overrides,
+                                                  llm_args_class=type(llm_args))
+
+    for key in model_defaults:
+        if key in merged_args:
+            try:
+                setattr(llm_args, key, merged_args[key])
+            except AttributeError:
```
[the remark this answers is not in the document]
venkywonka: 100% - that's claude's defensive programming :cry-laugh:

tensorrt_llm/llmapi/llm_utils.py
venkywonka: That was originally so that both TRT backend and pytorch backend could benefit from this - and that is also why i added a `_model_defaults_applied` so the second time the code hits that same function for the pytorch backend its a no-op.

but now that i think about it - this workaround is not elegant and this need not support TRT backend anymore. I'll go ahead and simplify it - thanks for the catch!

tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -985,4 +1004,208 @@ class LlmBuildStats:
...
+            setattr(temp_args, key, value)
+        except (ValidationError, AttributeError) as e:
+            raise ValueError(
+                f"Invalid model default field '{key}' with value {value}. Error: {e}"
+            ) from e
+
+    return defaults_dict
+
+
+def apply_model_defaults_to_llm_args(
```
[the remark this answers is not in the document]
venkywonka: hey @anish-shanbhag  - i have trimmed the logic - can you have another pass? thanks!

tensorrt_llm/llmapi/llm_utils.py
venkywonka: in the latest commit - i have removed its call from `_load_model_from_hf` so only pytorch path would have this functionality

anish-shanbhag (approved):

tensorrt_llm/commands/serve.py
```diff
@@ -141,33 +169,76 @@ def get_llm_args(
...
+        "num_postprocess_workers":
+        num_postprocess_workers,
+        "enable_chunked_prefill":
+        enable_chunked_prefill,
+        "revision":
+        revision,
+        "reasoning_parser":
+        reasoning_parser,
+        "otlp_traces_endpoint":
+        otlp_traces_endpoint,
```
Superjomn: Can this dict be out-of-date? Is it possible to add some checks here to ensure it is always aligned with the arglist?

tensorrt_llm/_torch/models/modeling_qwen3_next.py
```diff
@@ -1251,6 +1254,10 @@ def __init__(
         )
         self.preload_weight_modules = self.model.preload_weight_modules
 
+    @classmethod
+    def get_model_defaults(cls, llm_args: 'TorchLlmArgs') -> dict:
+        return {"kv_cache_config": {"enable_block_reuse": False}}
```
Superjomn: One concern: the dict of strings might be outdated, and there’s no clear reference between the strings and the target members in xxxConfigs. Although Pydantic should eagerly catch such issues—since all models are tracked in CI—it still feels like a potential leak.

tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -985,4 +988,65 @@ class LlmBuildStats:
...
+    Returns the defaults that were actually applied.
+    """
+    if not model_defaults_dict:
+        return {}
+
+    user_overrides = llm_args.model_dump(exclude_unset=True)
+    base_state = llm_args.model_dump()
+    merged_state = _deep_merge(base_state, model_defaults_dict, user_overrides)
+
+    new_args = TorchLlmArgs(**merged_state)
```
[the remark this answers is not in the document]
venkywonka: yes great catch!

Superjomn (approved):
Overall LGTM as the first step. 

I have some comments in the long run:
1. There needs to be a mechanism for module PICs/developers to keep updating the model defaults to keep offering good OOTB model defaults. cc @QiJune and @kaiyux for viz. 
2. There are many strings as symbol names referenced in dicts, which may be easy to become out-of-date

tensorrt_llm/_torch/models/modeling_nemotron_h.py
```diff
@@ -528,6 +531,10 @@ def load_weights(self, weights: Dict, weight_mapper: BaseWeightMapper):
         new_weights = weight_mapper.preprocess_weights(weights)
         super().load_weights(weights=new_weights, weight_mapper=weight_mapper)
 
+    @classmethod
+    def get_model_defaults(cls, llm_args: 'TorchLlmArgs') -> dict:
```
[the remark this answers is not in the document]
venkywonka: agreed, we don't have a model developer's checklist - i think its good to add a separate doc for new model developers and add this there. for now i think i'll add it in the contribution guide and coding rules. thanks for the catch!

tensorrt_llm/commands/serve.py
venkywonka: definitely agreed - in fact i noticed that `fail_fast_on_attention_window_too_large` was such a param that isn't even supposed to be in pytorch backend - such things seem to need manual cleanup. 

i believe we can add automation but that might be beyond the scope of this PR.

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
arysef: Do we need to worry about llm_args later being mutated to change these values and therefore ending up with divergent values? For example a situation where the post-merged llm_args max_seq_len is different from this one and so we end up with two values for it floating around

tensorrt_llm/commands/serve.py
```diff
@@ -141,33 +169,76 @@ def get_llm_args(
...
+                    max_beam_width=max_beam_width,
+                    max_seq_len=max_seq_len) if backend == "tensorrt" else None,
+        "scheduler_config":
+        SchedulerConfig(capacity_scheduler_policy=CapacitySchedulerPolicy.
+                        GUARANTEED_NO_EVICT,
+                        dynamic_batch_config=DynamicBatchConfig(
+                            enable_batch_size_tuning=True,
+                            enable_max_num_tokens_tuning=False,
+                            dynamic_batch_moving_average_window=128))
+        if backend == "tensorrt" else None,
```
arysef: I believe we also accept "trt" as a backend - maybe we can canonicalize from trt->tensorrt in this function to avoid any places where trt is missed?

juney-nvidia (approved):

tests/unittest/_torch/modeling/test_modeling_nemotron_h.py
```diff
@@ -46,16 +46,19 @@ def create_nemotron_h_llm(model_folder,
...
+            mamba_ssm_cache_dtype=mamba_ssm_cache_dtype)
+
     return LLM(
         model=model_dir,
         tensor_parallel_size=1,
         max_batch_size=max_batch_size,
         cuda_graph_config=CudaGraphConfig() if use_cuda_graph else None,
         disable_overlap_scheduler=disable_overlap_scheduler,
-        kv_cache_config=KvCacheConfig(
-            enable_block_reuse=False,
```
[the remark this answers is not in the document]
nv-guomingz: I mean we don't need to add this https://github.com/NVIDIA/TensorRT-LLM/pull/11035/changes#diff-dd8e405f922cc00ef03a97330c37df6f6ddd958bbc4cd01ffd283d6a10df50d4R49-R55 out of LLM class.

QiJune (approved):
LGTM

nv-guomingz (approved):
LGTM

PR #11044, opened 2026-01-27
[None][doc] Document torch.export limitations for Mamba SSM (dynamic shapes and control flow)

### LINKED ISSUE #10849: [Feature]:  AutoDeploy: linear attention pattern matching

Currently, AutoDeploy supports the following attention types:
* Quadratic (sdpa) attention
* Causal conv
* Mamba2 (SSM)
* DeltaNet

However, only sdpa attention comes with a set of pattern matcher transforms to automatically detect the attention pattern from a given source model, see the following transforms:
https://github.com/NVIDIA/TensorRT-LLM/blob/3a894951e7defa7fd44c4b972953b327c9cd8047/tensorrt_llm/_torch/auto_deploy/config/default.yaml#L33-L45

For other attention types, the model currently needs to be patched with the `torch.ops.auto_deploy.torch_*` reference op  in order for the compiler to correctly recognize the attention op for later transforms such as sharding and caching. 

For example, here is an example patch:
https://github.com/NVIDIA/TensorRT-LLM/blob/ae8f74b6206fb74d7924dacf69017e3491ec2812/tensorrt_llm/_torch/auto_deploy/models/patches/bamba.py#L87-L98

The goal of this ticket is to develop pattern matching transforms akin of the quadratic attention pattern matchers for our linear attention support.

Roughly, the implementation plan could look as follows:
1. Develop an initial pattern matcher with 1-2 patterns and 1-2 example models
2. Write a unit test to check if pattern matcher works as expected
3. Put together a list of relevant hybrid models
4. Collect all linear attention patterns present in these models
5. Extent the pattern matcher to handle more patterns
6. Extend and validate with extended unit test
7. Test e2e with a couple of integration tests and by removing patches that are now covered by the pattern matcher

### Alternatives

_No response_

### Additional context

_No response_

karthikvetrivel:
## Description

This PR documents why the Mamba SSM implementation cannot be traced by `torch.export` when sequence length is symbolic. The intent of this PR is to provide a explanation so future contributors understand why the SSM is intentionally encapsulated behind a custom op and isn't currently partially traced. We aim to brainstorm how we can port some of this behavior to a pattern matcher.

The SSM chunking logic depends on runtime sequence length (`seq_len`) and includes:

**ISSUE 1: modulo arithmetic on symbolic dimensions**

https://github.com/NVIDIA/TensorRT-LLM/blob/c26a8f764cbfc96846a50c7b7221795e424ee586/tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/torch_mamba.py#L96

**ISSUE 2: reshapes with data-dependent sizes**

In this case, the -1 dimension's value depends on `pad_size`, which depends on `seq_len % chunk_size`.
With symbolic `seq_len`: PyTorch can't compute what -1 should be at trace time because `(seq_len + pad_size) / chunk_size` involves a symbolic modulo result.

https://github.com/NVIDIA/TensorRT-LLM/blob/c26a8f764cbfc96846a50c7b7221795e424ee586/tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/torch_mamba.py#L36

https://github.com/NVIDIA/TensorRT-LLM/blob/c26a8f764cbfc96846a50c7b7221795e424ee586/tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/torch_mamba.py#L40-L42

https://github.com/NVIDIA/TensorRT-LLM/blob/c26a8f764cbfc96846a50c7b7221795e424ee586/tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/torch_mamba.py#L160

**ISSUE 3: control flow conditioned on symbolic values.**

https://github.com/NVIDIA/TensorRT-LLM/blob/c26a8f764cbfc96846a50c7b7221795e424ee586/tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/torch_mamba.py#L158-L159

These patterns are fundamentally incompatible with `torch.export`’s symbolic tracing model and cannot be refactored away without changing SSM semantics.

No functional or behavioral changes are introduced.

## Test Coverage

N/A — documentation only. No code paths were modified.

## PR Checklist

- [x] PR description clearly explains what and why
- [x] No code changes or behavior changes
- [x] Documentation-only update
- [x] No new dependencies introduced
- [x] No test updates required

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/torch_mamba.py
+++ b/tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/torch_mamba.py
@@ -33,6 +33,7 @@ def _reshape_into_chunks(input_tensor, pad_size, chunk_size):
 
     if len(input_tensor.shape) == 3:
         # [bsz, seq_len multiple of chunk_size, num_heads] -> [bsz, -1, chunk_size, num_heads]
+        # ISSUE 2: -1 dim depends on pad_size (symbolic).
         return input_tensor.reshape(input_tensor.shape[0], -1, chunk_size, input_tensor.shape[2])
     else:
         # [bsz, seq_len multiple of chunk_size, num_heads, head_dim or state_size] ->
@@ -93,6 +94,7 @@ def _torch_ssm_prefill(
     C = C.reshape(batch_size, seq_len, n_groups, ssm_state_size).float()
     B = B.repeat_interleave(num_heads // n_groups, dim=2, output_size=num_heads)
     C = C.repeat_interleave(num_heads // n_groups, dim=2, output_size=num_heads)
+    # ISSUE 1: seq_len is symbolic, so pad_size is symbolic
     pad_size = (chunk_size - seq_len % chunk_size) % chunk_size
 
     D_residual = D[..., None] * _pad_tensor_by_size(hidden_states, pad_size)
@@ -102,6 +104,7 @@ def _torch_ssm_prefill(
     A = A.to(hidden_states.dtype) * dt
 
     # Rearrange into blocks/chunks
+    # The -1 dimesion's size depends on pad_size, which is data-dependent.
     hidden_states, A, B, C = [
         _reshape_into_chunks(t, pad_size, chunk_size) for t in (hidden_states, A, B, C)
     ]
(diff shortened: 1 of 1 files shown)


lucaslie (commented):
Thank you for going through this is in so much detail. I agree that those are fundamental limitations. And we don't want to rely on partial manual patching and pattern matching. Either we do 
1. full pattern matching, or
2. full patching

a combination of both (i.e. patching to make the pattern matcher work) seems undesirable. 

So I agree with your conclusion that it's currently infeasible to export + pattern match SSM. Maybe in the future if there is a more torch-native way to represent this but for now let's stick to patches

PR #11307, opened 2026-02-05
[TRTLLM-9778][feat] Implement python based kv_cache_v2 scheduler

peaceh-nv:
Implement python based max utilization scheduler based on kv cache v2
 - Move KVCacheV2DummyScheduler and KVCacheV2MaxUtilizationScheduler to SimpleUnifiedScheduler
 - KV cache manager V2 based Max util scheduler need to do prepare_resources inside schedule_request and no need to do prepare_resources afterwards.


tensorrt_llm/_torch/pyexecutor/scheduler.py
```diff
@@ -230,6 +234,281 @@ def schedule_request(
...
+                break
+
+            if req_state == LlmRequestState.DISAGG_GENERATION_INIT:
+                scheduled_disagg_gen_init_requests.append(request)
+            else:
+                scheduled_requests.append(request)
+
+        return scheduled_requests, scheduled_disagg_gen_init_requests, []
+
+    def prepare_resources(self, context_requests: RequestList,
```
yizhang-nv: The overall design exposes lots of details of allocating kv cache which should be hidden inside kv cache manager. (e.g. direct access of internal variables like `kv_cache_map`). The ideal solution would be using apis like `prepare_blocks_if_schedulable` to hide the resource allocation details and only tell the scheduler whether we have enough resources to schedule this req or not. 

If this will take long time to design the API and overall structure, we can do this in the following pr.

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
```diff
@@ -949,6 +1013,82 @@ def _try_scheduling_request(
...
+    ) -> tuple[RequestList, RequestList]:
+        # Lazy initialization of delegate scheduler
+        if self.delegate_scheduler is None:
+            self.delegate_scheduler = KVCacheV2MaxUtilizationScheduler(
+                max_num_requests=scheduler.max_num_requests,
+                kv_cache_manager=scheduler.kv_cache_manager,
+            )
+
+        # Delegate to KVCacheV2MaxUtilizationScheduler
+        scheduled_requests, scheduled_disagg_gen_init_requests, paused_requests = (
```
lancelly: Both KVCacheV2MaxUtilizationPolicy and KVCacheV2DummyPolicy delegate 1:1 to a lazily-initialized Scheduler class. The Scheduler classes don't use any logic that can't be done directly in the Policy. For example, KVCacheV2MaxUtilizationScheduler.schedule_request is just state filtering + append, which can be inlined into the Policy's schedule() method — and it would also benefit from using scheduler._can_be_scheduled() instead of reimplementing the state check.
How about removing both Scheduler classes and inlining their logic into the Policy classes? This eliminates 2 classes, the lazy init pattern, and one layer of indirection in both the schedule and prepare_resources call chains.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -1984,6 +2001,218 @@ def _kv_connector_should_add_sequence(self, request: LlmRequest) -> bool:
...
+
+                        if kv_cache.resize(req.prompt_len):
+                            new_context_batch.append(req)
+                        else:
+                            kv_cache.suspend()
+                            continue
+
+                        if self.kv_connector_manager is not None:
+                            block_ids = self.get_cache_indices(req)
+                            self.kv_connector_manager.update_state_after_alloc(
```
lancelly: Does kv cache manager v2 supoort this?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -1984,6 +2001,218 @@ def _kv_connector_should_add_sequence(self, request: LlmRequest) -> bool:
...
+                                req.py_request_id, req.lora_task_id,
+                                req.get_tokens(0)[:-1]
+                                if self.enable_block_reuse else None)
+                            assert beam_width == 1, "Currently, KVCacheManagerV2 only supports beam width 1"
+                        if not self.enable_block_reuse:
+                            assert kv_cache.num_committed_tokens == 0
+                            kv_cache.stop_committing()
+                        else:
+                            req.context_current_position = kv_cache.num_committed_tokens
+                            chunk_size = req.context_chunk_size
```
lancelly: Should chunked prefill and non-chunked paths have different approaches?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -1956,6 +1977,226 @@ def _kv_connector_should_add_sequence(self, request: LlmRequest) -> bool:
...
+                        if evicted is None:
+                            # No more requests to evict
+                            break
+                        if evicted in new_generation_batch:
+                            new_generation_batch.remove(evicted)
+                        evicted_requests.append(evicted)
+
+                if not capacity_increased:
+                    # Could not increase capacity even after evicting all possible requests
+                    continue
```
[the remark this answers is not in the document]
peaceh-nv: Fixed this in the latest commit. Add "kv_cache.suspend()" for this case.

PR #11472, opened 2026-02-12
[None][perf] Enable fused MoE routing kernel and reduce Python overhead in decode hot path

kaiyux:
## Description

Optimize DeepSeek V3/V3 Lite serving performance through two areas:

### 1. Enable fused routing kernel for single-group MoE configs
The Python guard in `Deepseekv3RoutingImpl.noaux_tc()` was overly restrictive, blocking the fused CUDA kernel for `n_group=1, topk_group=1` configurations unless `top_k==22`. The underlying C++ kernel (`noAuxTcKernels.cu`) already supports `n_group<=1` with `top_k<=8` and `num_experts<=512` via the `is_single_group` path.

Remove the unnecessary condition to allow models like DeepSeek V3 Lite (72 experts, top_k=6) and Kimi K2 (384 experts, top_k=8) to use the fused kernel instead of falling back to separate PyTorch ops (sigmoid, topk, gather, sum, div).

### 2. Reduce Python overhead in decode hot path
- Pre-allocate pinned host buffers for `input_ids`, `position_ids`, `previous_batch_indices`, and `seq_lens` to avoid per-iteration `mlock()` syscalls from `torch.tensor(..., pin_memory=True)`
- Cache D2H copy destination buffers in `TRTLLMSampler` by source tensor `data_ptr` to eliminate repeated pinned memory allocation
- Replace `.tolist()` with `.numpy()` in sampler hot path for zero-copy tensor-to-host conversion
- Add early-exit guard for `log_probs` when no requests need them
- Return `generation_requests` directly from `all_requests()` in decode-only case to avoid list concatenation called 3-5x/iteration
- Cache `HandleLogits`/`HandleAdditionalOutputs` instances on executor
- Replace per-iteration inner function creation in `_forward_step` with `nvtx_range` context manager
- Replace `active_requests` `clear()`+`extend()` with direct assignment
- Avoid list concatenation when building reqs in sampler update paths

### Performance (DeepSeek V3 Lite NVFP4 / B200, ISL=1024, OSL=512)
| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Output throughput | 1776 tok/s | 2109 tok/s | **+18.8%** |
| TPOT | 6.72 ms | 5.91 ms | **-12.2%** |
| TTFT | 504 ms | 283 ms | **-44.0%** |

## Test Coverage

- Existing MoE routing unit tests cover kernel dispatch logic
- `trtllm-bench` end-to-end benchmarks used for performance validation

## PR Checklist

### DIFF
--- a/tensorrt_llm/_torch/model_config.py
+++ b/tensorrt_llm/_torch/model_config.py
@@ -247,9 +247,8 @@ class ModelConfig(Generic[TConfig]):
         if moe_backend.upper() != "AUTO":
             return moe_backend
 
-        sm_version = get_sm_version()
-
         if architecture == "GptOssForCausalLM":
+            sm_version = get_sm_version()
             # Select the best performing backend based on SM version
             if 100 <= sm_version < 120:  # Blackwell
                 return "TRTLLM"
@@ -257,12 +256,6 @@ class ModelConfig(Generic[TConfig]):
                 return "TRITON"
             else:
                 return "CUTLASS"  # Fallback to CUTLASS for other SM versions (e.g., SM120)
-
-        if architecture in ("DeepseekV3ForCausalLM", "DeepseekV32ForCausalLM"):
-            if 100 <= sm_version < 120:  # Blackwell
-                return "TRTLLM"
-            elif 90 <= sm_version < 100:  # Hopper
-                return "DEEPGEMM"
 
         return "CUTLASS"
 
--- a/tensorrt_llm/_torch/model_config.py
+++ b/tensorrt_llm/_torch/model_config.py
@@ -247,8 +247,9 @@ class ModelConfig(Generic[TConfig]):
         if moe_backend.upper() != "AUTO":
             return moe_backend
 
+        sm_version = get_sm_version()
+
         if architecture == "GptOssForCausalLM":
-            sm_version = get_sm_version()
             # Select the best performing backend based on SM version
             if 100 <= sm_version < 120:  # Blackwell
                 return "TRTLLM"
@@ -256,6 +257,12 @@ class ModelConfig(Generic[TConfig]):
                 return "TRITON"
             else:
                 return "CUTLASS"  # Fallback to CUTLASS for other SM versions (e.g., SM120)
+
+        if architecture in ("DeepseekV3ForCausalLM", "DeepseekV32ForCausalLM"):
+            if 100 <= sm_version < 120:  # Blackwell
+                return "TRTLLM"
+            elif 90 <= sm_version < 100:  # Hopper
+                return "DEEPGEMM"
 
         return "CUTLASS"
 
--- a/tensorrt_llm/_torch/modules/fused_moe/routing.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/routing.py
@@ -264,9 +264,10 @@ class Deepseekv3RoutingImpl:
                         "The configuration is not supported by the fused routing kernel. We have to use the original pytorch implementation."
                     )
                 self.is_fused = False
-        elif (num_experts > 512 or (self.top_k > 8 and self.top_k != 22)
-              or (self.topk_group == 1 and self.top_k != 22)):
-            # We have special implementation for n_group == 1, top_k == 22 and num_experts == 512 for Nemotron Super v3.
+        elif (num_experts > 512 or (self.top_k > 8 and self.top_k != 22)):
+            # The fused CUDA kernel supports:
+            #   - n_group == 1 with top_k <= 8 and num_experts <= 512
+            #   - n_group == 1 with top_k == 22 and num_experts == 512 (Nemotron Super v3)
             if self.is_fused:
                 warnings.warn(
                     "The configuration is not supported by the fused routing kernel. We have to use the original pytorch implementation."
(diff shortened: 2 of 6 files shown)


kaiyux (commented):
## Code Review: [None][perf] Reduce Python overhead in decode hot path and auto-select MoE backend for DeepSeek V3

Good PR overall — the optimizations are well-targeted and the performance numbers are impressive (+18.8% output throughput, -12.2% TPOT). A few items worth addressing:

---

### 1. `all_requests()` aliasing contract (scheduler.py) — Medium Risk

```python
def all_requests(self) -> list[LlmRequest]:
    if not self.context_requests:
        return self.generation_requests  # returns the ACTUAL list, not a copy
    return self.context_requests + self.generation_requests
```

This returns the **live `generation_requests` list** in the decode-only fast path. The comment says "Callers must not structurally modify (append/remove) the returned list", but the return type `list[LlmRequest]` doesn't enforce this. Today's callers appear safe (they iterate or pass to functions), but this is a fragile invariant.

**Suggestion**: Consider returning `Sequence[LlmRequest]` in the type hint to signal read-only intent at the type level, or add an assertion in debug builds.

---

### 2. `seq_lens` pre-allocation partially defeated (model_engine.py) — Low Risk

```python
self.seq_lens_host[:num_seqs].copy_(torch.as_tensor(sequence_lengths, dtype=torch.int))
attn_metadata.seq_lens = self.seq_lens_host[:num_seqs].clone()
```

The `.clone()` produces a **non-pinned** CPU tensor. The `AttentionMetadata.seq_lens` setter (in `interface.py:202`) then calls `.pin_memory()`, which allocates *another* pinned buffer. Net result: 2 allocations per iteration (clone + pin) vs. 1 in the old code (`torch.tensor(..., pin_memory=True)`).

Since `seq_lens` is small (batch_size ints), this is a minor inefficiency — the real wins come from the other three buffers (`input_ids`, `position_ids`, `previous_batch_indices`) which don't go through a re-pinning setter. Consider either:
- Documenting this as an accepted trade-off, or
- Passing the pinned slice directly (but `.pin_memory()` on an already-pinned tensor returns `self`, which would alias the persistent buffer — so `.clone()` IS needed for correctness if the setter doesn't copy)

---

### 3. `_host_buffer_cache` invariants (sampler.py) — Suggestion

```python
self._host_buffer_cache: dict[int, torch.Tensor] = {}
```

The cache relies on two undocumented invariants:
1. Decoder state tensors have **stable `data_ptr`** across iterations
2. Each `data_ptr` is used at most **once per sample step**

Both are valid today. Consider adding a brief comment documenting these assumptions at the cache declaration site to prevent future breakage.

Also: the cache grows unboundedly if tensor addresses change (e.g., model reconfiguration). In practice this shouldn't happen during serving, but a `_host_buffer_cache.clear()` in the appropriate lifecycle method would be defensive.

---

### 4. `.numpy()` vs `.tolist()` — Verified Safe

Traced all downstream usage of `sequence_lengths_host_data`, `finish_reasons`, and `new_tokens` after the `.numpy()` conversion. Numpy scalars work correctly with:
- Integer comparisons and arithmetic
- Bitwise operations in `FinishedState`
- Nanobind `std::vector<int32_t>` conversion in `add_new_tokens_to_requests`
- Dict key usage and `Logprob` constructor

No issues here — good micro-optimization.

---

### 5. `HandleLogits` / `HandleAdditionalOutputs` caching — Verified Safe

Both classes are completely stateless (no `__init__`, no instance attributes). Caching them on the executor is correct and eliminates per-iteration object construction overhead.

---

### 6. Fused routing kernel guard fix (routing.py) — Correct

The removed condition `(self.topk_group == 1 and self.top_k != 22)` was overly restrictive. The CUDA kernel (`noAuxTcKernels.cu`) supports `n_group <= 1` with `top_k <= 8` and `num_experts <= 512` via the `is_single_group` path. This correctly enables the fused kernel for DeepSeek V3 Lite (72 experts, top_k=6) and Kimi K2 (384 experts, top_k=8).

---

### 7. MoE backend auto-selection (model_config.py) — Clean

The `DeepseekV3ForCausalLM` / `DeepseekV32ForCausalLM` handling falls through to `CUTLASS` for non-Hopper/Blackwell architectures, consistent with the existing `GptOssForCausalLM` pattern. One minor note: the `sm_version` variable is now extracted before the architecture check, which is fine (called once, cheap).

---

### 8. `_forward_step` refactoring (py_executor.py) — Clean

Replacing the per-iteration nested closure + `@nvtx_range` decorator with a `nvtx_range` context manager is a good simplification that eliminates closure creation, decorator overhead, and an extra call frame.

---

### Summary

| Concern | Severity | Action |
|---------|----------|--------|
| `all_requests()` aliasing | Medium | Consider `Sequence[LlmRequest]` return type |
| `seq_lens` double-pin | Low | Document as accepted trade-off |
| `_host_buffer_cache` invariants | Low | Add comment documenting assumptions |
| `.numpy()` conversion | None | Verified safe |
| Handler caching | None | Verified safe |
| Routing guard fix | None | Correct |
| MoE backend selection | None | Clean |
| `_forward_step` refactor | None | Clean |

Overall: well-motivated optimizations with solid performance gains. The aliasing concern in `all_requests()` is the main item to consider before merge.

ChristinaZ (approved):

PR #11589, opened 2026-02-19
[TRTLLM-10004][feat] Enable GEMM -> AR with GEMM output in registered buffers

nv-lschneider:
## Summary by CodeRabbit

* **New Features**
  * Added an output-buffer-kind system (Default, Userbuffers, NCCL window) with optional group selection for GEMM/output allocation.
  * New NCCL “window” tensors to enable in-place, multi-device output routing; model and linear layers can opt into this to reduce temporaries.
  * AutoTuner helper to query cached tactics.

* **Refactor**
  * Unified output allocation across GEMM and related ops; replaced boolean flags with the new buffer-kind API.
  * Updated model/attention/linear paths to propagate buffer-kind and a prefer-window-output hint.

## Description

The goal is to use registered buffer for the NCCL_SYMMETRIC AllReduce strategy.
Output buffers are generated and not a problem, but input buffers must come from the previous operation directly into the registered buffer.

This PR implements this in fused python operations (GEMM and MOE) to enable output directly in a custom buffer if NCCL_SYMMETRIC is chosen.
This approach of modifying the python ops enables this to work independent of execution mode.
I.e. it works directly with default settings (torch compile is not required).

But as a downside, it manually needs to modify each operation individually that wants to make use of this speed up.
The FP4 GEMM can currently not be modified like this, since it already uses a differently defined custom output buffer.
(There might a future upside to unify this on a ncclWindowTensor as well.)

A little bit of plumbing is added to conservatively obtain the output tensors from the existing machinery.

I also modified the auto-tuner to use registered buffers for NCCL_SYMMETRIC assuming we can get those.
And the AR benchmark is modified to enable registered buffers as well.

## Test Coverage

No explicit test for this is checked in, but literally any multi-GPU run is using this because of the auto-tuner.

## PR Checklist

## GitHub Bot Help

<details>

### kill

`kill  `

### skip

`skip --comment COMMENT `

### reuse-pipeline

`reuse-pipeline `

</details>


nv-lschneider:
I changed the behavior to use NCCLWindowTensors now always with AUTO AllReduce strategy.
I just want to know if CI is green with that change included.
Using a NCCLWindowTensor for non-nccl comm strategies should be fine with no perf degradation.
So this may be a valid solution. Happy to discuss more.

tensorrt_llm/_torch/models/modeling_llama.py
```diff
@@ -541,6 +560,9 @@ def forward(
             all_rank_num_tokens=attn_metadata.all_rank_num_tokens,
             final_all_reduce_params=AllReduceParams(
                 enable_allreduce=not self.disable_feed_forward_allreduce),
+            allow_window_output=(self.disable_feed_forward_allreduce
+                                 and (self.fusion_config.POST_MLP_FUSION
+                                      or self.fusion_config.POST_MOE_FUSION)),
```
Tabrizian: `Llama4MoE.forward()` (line 333) doesn't accept `allow_window_output` and has no `**kwargs`, so this crashes for MoE layers:
```
TypeError: Llama4MoE.forward() got an unexpected keyword argument 'allow_window_output'
```
Same for `Llama4MinLatencyMoE.forward()`. Could you add it to both signatures?

Also, should we rename this to `prefer_window_output`? "allow" reads like a permission gate, but it's more of a hint that the caller prefers window output.

cpp/tensorrt_llm/thop/ncclWindowTensor.cpp
```diff
@@ -0,0 +1,112 @@
...
+#endif // ENABLE_MULTI_DEVICE
+}
+
+} // namespace torch_ext
+
+TRTLLM_NAMESPACE_END
+
+TORCH_LIBRARY_FRAGMENT(trtllm, m)
+{
+    m.def("create_nccl_window_tensor(Tensor like, int[] group, int[]? shape=None) -> (Tensor out, bool is_valid)");
```
Tabrizian: This op is missing a `register_fake` in `cpp_custom_ops.py`. CI fails with:
```
AssertionError: Custom ops missing fake impl: trtllm::create_nccl_window_tensor
```

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -1230,15 +1289,15 @@ def w4a8_mxfp4_fp8_gemm(
...
+        act_fp4: torch.Tensor,
+        weight: torch.Tensor,
+        act_sf: torch.Tensor,
+        weight_scale: torch.Tensor,
+        alpha: torch.Tensor,
+        output_dtype: torch.dtype,
+        output_buffer_kind: int = int(OutputBufferKind.DEFAULT),
 ) -> torch.Tensor:
-    return act_fp8.new_empty((act_fp8.size(0), weight.size(0)),
+    return act_fp4.new_empty((act_fp4.size(0), weight.size(0)),
```
Tabrizian: I wonder how this code used to work before.

cpp/tensorrt_llm/thop/outputTensor.h
```diff
@@ -0,0 +1,101 @@
...
+#include <vector>
+
+TRTLLM_NAMESPACE_BEGIN
+
+namespace torch_ext
+{
+
+using CreateTensorFn = std::function<at::Tensor(at::IntArrayRef shape, at::ScalarType dtype, c10::Device device)>;
+
+enum class OutputBufferKind : int
```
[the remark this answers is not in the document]
nv-lschneider: Added a comment above the enum noting it's exposed to Python via the nanobind binding and that integer values must stay in sync.

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -136,9 +136,12 @@ def weight_dequant(x: torch.Tensor,
 
 
 @torch.compile(dynamic=True)
-def moe_reduce_add_shared_output(routed_output, shared_output):
-    routed_output = torch.sum(routed_output, dim=1, keepdim=False)
-    return shared_output + routed_output
+def moe_reduce_add_shared_output(routed_output, shared_output, out=None):
```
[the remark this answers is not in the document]
nv-lschneider: The `out=None` argument adds one extra graph variant in `torch.compile`: one for `out=None` (non-TP or non-window path) and one for `out=<tensor>` (TP + NCCL window path). Each variant is compiled once since the branch is determined statically per rank — the window path is always taken or never taken for a given rank. Within the tensor variant, `dynamic=True` already handles shape variation, so no further recompilation occurs. No incompatibility introduced.

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -367,7 +392,7 @@ def __init__(
 
     def unique_id(self):
         return (
-            self.to_userbuffers,
+            self.output_buffer_kind,
             self.output_dtype,
```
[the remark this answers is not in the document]
nv-lschneider: You're right, we should add `group`. The group size determines the scale of the AllReduce that follows the GEMM, which affects the overall GEMM+AR balance and thus which GEMM tactic is globally optimal. Runners with different groups should not share tactic results. Fixed.

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -1820,10 +1879,31 @@ def tunable_allreduce(
...
+        if not inputs:
+            return inputs
+        input_tensor = inputs[0]
+        if not isinstance(input_tensor,
+                          torch.Tensor) or not input_tensor.is_cuda:
+            return inputs
+        window_tensor, is_valid = torch.ops.trtllm.create_nccl_window_tensor(
+            input_tensor, group)
+        if not is_valid:
+            return inputs
```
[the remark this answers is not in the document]
nv-lschneider: If the input is not in a registered (window) buffer, `NCCL_SYMMETRIC` will copy it to a registered buffer before the AllReduce, which adds overhead and reduces performance. However, this is the pre-PR behavior — this PR is specifically about eliminating that copy by writing the GEMM output directly into the window buffer. If window allocation fails at tune time (and at runtime), we fall back to that copy path, which is not a regression.

cpp/tensorrt_llm/thop/ncclWindowTensor.cpp
nv-lschneider: Added in `cpp_custom_ops.py`.

tensorrt_llm/_torch/modules/linear.py
```diff
@@ -469,6 +401,21 @@ def apply(self, module: Linear, input: torch.Tensor,
             output = F.linear(input, module.weight, bias)
         return output
 
+    def apply_out(self, module: Linear, input: torch.Tensor,
```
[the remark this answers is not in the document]
nv-lschneider: Handled — a pre-check at the call site in `forward` nulls out `window_out` before calling `apply_out` when the input is float8 and `use_custom_cublas_mm` is False (the path where `torch.matmul(out=)` is unsupported). So `apply_out` never reaches the fallback for a live window buffer, and `window_success` correctly reflects False in that case.

tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
```diff
@@ -1125,3 +1133,10 @@ def _(k_cache: torch.Tensor, slot_mapping_fp8: torch.Tensor,
         k_fp8 = k_cache.new_empty([num_tokens, 128], dtype=torch.float8_e4m3fn)
         k_scale = k_cache.new_empty([num_tokens, 1], dtype=torch.float32)
         return k_fp8, k_scale
+
+    @torch.library.register_fake("trtllm::create_nccl_window_tensor")
+    def _(like: torch.Tensor,
+          group: List[int],
+          shape: Optional[List[int]] = None):
+        out_shape = shape if shape is not None else list(like.shape)
+        return like.new_empty(out_shape), True
```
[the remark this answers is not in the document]
nv-lschneider: The design is intentional: is_valid=True in the fake is correct because under torch.compile(fullgraph=True) the compiled graph must handle only one path, and the C++ op now always returns a defined tensor (falling back to a regular CUDA tensor on failure, which allreduce handles transparently). No code changes needed — just the explanatory comment.

cpp/tensorrt_llm/thop/cublasScaledMM.cpp
```diff
@@ -21,6 +21,7 @@
 #include "tensorrt_llm/plugins/common/plugin.h"
 #include "tensorrt_llm/plugins/gemmPlugin/gemmPlugin.h"
 #include "tensorrt_llm/runtime/torchUtils.h"
+#include "tensorrt_llm/thop/outputTensor.h"
```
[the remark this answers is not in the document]
nv-lschneider: This is addressed now.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
```diff
@@ -562,12 +570,23 @@ def forward(
             a_tensor, b_tensor, a_sf_tensor, b_sf_tensor, alpha_tensor = inputs
             m, k, n = a_tensor.shape[0], a_tensor.shape[1], b_tensor.shape[0]
 
-            # Allocate output tensor from UserBuffers or regular CUDA memory
-            if self.to_userbuffers:
+            # Allocate output tensor based on output_buffer_kind
+            if self.output_buffer_kind == int(BufferKind.NCCL_WINDOW):
+                window, is_valid = torch.ops.trtllm.create_nccl_window_tensor(
+                    a_tensor, self.group or [], [m, n])
```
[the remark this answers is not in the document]
nv-lschneider: Addressed in a different way

hyukn (commented):
Thanks for the great effort. This is truly a big change and has large scope of impact. My major concern is the complexity of `use_window` and the perf part. Because we might not have a very solid benchmarking to illustrate the impacts of allocating NCCL window, like some extra overhead (which still exists when window is not usable).

I am not sure if adding a environment at this moment to keep the featrue opt-in by using env variable will be better? This may gives us an easier way to switch on the GEMM + AR buffer switching. Because we already have a green CI, default-off will not affect the pipeline results and it will be easy to pass. Also we can turn on the feature only in some tests (allreduce and e2e tests like in test_llm_api_pytorch.py) to verify the changes. What do you think @nv-lschneider

StanleySun639 (approved):

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
```diff
@@ -562,14 +570,24 @@ def forward(
...
+            elif self.output_buffer_kind == int(BufferKind.USERBUFFERS):
                 c_tensor = torch.ops.trtllm.create_userbuffers_tensor(
                     [m, n], self.output_dtype)
             else:
-                c_tensor = torch.empty(*(m, n),
+                c_tensor = torch.empty(m,
+                                       n,
                                        dtype=self.output_dtype,
-                                       device="cuda")
+                                       device=a_tensor.device)
```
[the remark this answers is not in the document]
nv-lschneider: I changed the design and eliminated create_nccl_window_tensor.
Instead we have now an allocateOutput function and it returns a buffer and the bufferKind that is acutally allocated, instead of `valid`.
More flexible and clean design. That eliminates the full if / else switch here on the python side.

The function does need to wrap the previous implementation, so that we can cleanly expose it to python from C++.

tests/unittest/_torch/multi_gpu/test_linear.py
```diff
@@ -358,6 +358,8 @@ def fp4_row_linear_allreduce(tp_size, local_rank, seq_len, output_size,
                              hidden_size, dtype, output_ref, x_sf_global,
                              w_sf_global, x_fp4s, w_fp4, x_sf_blocks,
                              w_sf_block_unswizzled):
+    import tensorrt_llm._torch.distributed.ops as _ops
+    _ops._NCCL_SYMMETRIC_ZERO_COPY = True
```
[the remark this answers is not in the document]
nv-lschneider: done

tests/integration/defs/accuracy/test_llm_api_pytorch.py
```diff
@@ -4904,6 +4904,8 @@ def test_w4_4gpus(self, v2_kv_cache, kv_cache_reuse, kv_cache_dtype,
                       moe_backend, tp_size, pp_size, ep_size, attention_dp,
                       cuda_graph, overlap_scheduler, mocker):
 
+        mocker.patch.dict(os.environ, {"TLLM_NCCL_SYMMETRIC_ZERO_COPY": "1"})
```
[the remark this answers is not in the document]
nv-lschneider: I reverted this change

nv-lschneider:
> I see your concern. I added a killswitch and disable this path for now. Rebasing and rerunning CI.
> 
> It is default off, except for a few tests, where I enabled it.
> 
> After a full perf study to see the impact, we can change the default of the kill switch and turn it on. Thanks for taking a look @hyukn

I reverted the test change only after discussion with Iman. We had a green CI on the ON before. Now we are going to get green CI on the OFF.
The API doesn't change with the switch, so I am not concerned about someone introducing a breaking change.

And once we can enable this change, we will test with a new default again.

Tabrizian (approved):

mikeiovine (approved):
Stamp on behalf of llama devs, did not review changes in other files

brb-nv:
> This PR implements this in fused python operations (GEMM and MOE) to enable output directly in a custom buffer if NCCL_SYMMETRIC is chosen.

Naive question: Is it possible to always use registered buffers regardless of NCCL_SYMMETRIC being chosen?

cpp/tensorrt_llm/thop/allocateOutput.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+TRTLLM_NAMESPACE_BEGIN
+
+namespace torch_ext
+{
+
+std::tuple<at::Tensor, int64_t> allocateOutputOp(at::Tensor const& like, int64_t output_buffer_kind,
+    c10::optional<torch::List<int64_t>> group, c10::optional<at::IntArrayRef> shape,
+    c10::optional<at::ScalarType> out_dtype)
+{
```
brb-nv, line 43: Nit: Can we use const wherever possible?

```
std::tuple<at::Tensor, int64_t> allocateOutputOp(at::Tensor const& like, int64_t output_buffer_kind,
    c10::optional<torch::List<int64_t>> group, c10::optional<at::IntArrayRef> shape,
    c10::optional<at::ScalarType> out_dtype)
{
    std::vector<int64_t> const outShape = (shape.has_value() && !shape->empty())
        ? std::vector<int64_t>(shape->begin(), shape->end())
        : std::vector<int64_t>(like.sizes().begin(), like.sizes().end());
    at::ScalarType const dtype = out_dtype.value_or(like.scalar_type());
    auto const [tensor, actual_kind] = torch_ext::allocate_output(
        outShape, dtype, like.device(), static_cast<torch_ext::BufferKind>(output_buffer_kind), group);
    return {tensor, static_cast<int64_t>(actual_kind)};
}
```

nv-lschneider:
> > This PR implements this in fused python operations (GEMM and MOE) to enable output directly in a custom buffer if NCCL_SYMMETRIC is chosen.
> 
> Naive question: Is it possible to always use registered buffers regardless of NCCL_SYMMETRIC being chosen?

Yes — using a registered buffer instead of a non-registered one does not cause performance degradation, regardless of whether NCCL_SYMMETRIC is enabled.

The main drawback is that memory for symmetrically registered buffers is currently managed outside of Python, so it’s possible to over-allocate and hit an OOM error if used unnecessarily. It’s mostly safe to always enable this for AllReduce, unless you’re already close to memory limits.

You could use a symmetric registered memory pool for all operations, but doing this through PyTorch requires a newer version than the one currently in the container — so that’s planned for later.

Some inference frameworks register their entire memory pool, which works but adds complexity since communication operations must align buffer offsets across ranks. With careful coordination, it’s still a viable approach overall.

TLDR. Yes we could turn it on for all strategies. But in order to not risk OOM problems it would be better to have a graceful fallback.

tests/unittest/_torch/thop/parallel/test_fp4_linear.py
```diff
@@ -469,7 +469,7 @@ def test_nvfp4_gemm_unified_all_tactics(dtype, mnk):
                                                  weight_scale=w_sf_block,
                                                  alpha=alpha_tensor,
                                                  output_dtype=dtype,
-                                                 to_userbuffers=False,
+                                                 output_buffer_kind=0,
```
brb-nv, line 472: This seems to be the only test updated and all tests seem to be testing only the `Default` buffer kind. Should we parameterize this test to check `userbuffers` and `nccl_window` buffer kinds also?

tensorrt_llm/_torch/autotuner.py
```diff
@@ -1025,6 +1025,27 @@ def choose_one(
                 custom_op, 0) + tuning_end_time - tuning_start_time
         return (runners[runner_id], tactic)
 
+    def get_cached_tactic(
```
brb-nv: Can you please help me understand who is calling this function? If it's dead code, can it wait for a future MR where it's actually put to use?

tensorrt_llm/_torch/modules/linear_common.py
```diff
@@ -0,0 +1,99 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Shared tensor-parallel types and weight-loading utilities with minimal dependencies.
+
+This module exists to avoid circular imports: it does not depend on
+tensorrt_llm._torch.custom_ops or tensorrt_llm._torch.modules.fused_moe.
+Consumers that need only TensorParallelMode and load_weight_shard (e.g. fused_moe
+quantization) should import from here; linear.py re-exports these for backward
```
brb-nv: Maybe I'm missing something but I currently see duplication of `TensorParallelMode` and `load_weight_shard` here and `linear.py`. Is this intentional? Not sure what re-export means.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -249,6 +249,7 @@ def _helix_cp_output_projection(
     mapping_o: Mapping,
     layer_idx: int,
     lora_params: Optional[dict] = None,
+    prefer_window_output: Optional[bool] = None,
```
brb-nv: Nit: Can be more explicit: `prefer_nccl_window_output`. Else, it could be confused with sliding window used in attention.

brb-nv: Extends to other modules with this variable name.

cpp/tensorrt_llm/thop/fp8RowwiseGemm.cpp
```diff
@@ -52,8 +52,8 @@ void check_input_dtypes(torch::Tensor const& mat, torch::Tensor const& matScale)
 
 template <typename OutputType>
 torch::Tensor fp8_rowwise_gemm_launch(torch::Tensor const& mat1, torch::Tensor const& mat2,
-    torch::Tensor const& mat1Scale, torch::Tensor const& mat2Scale, bool to_userbuffers = false,
-    tkc::CutlassGemmConfig const* maybe_config = nullptr)
+    torch::Tensor const& mat1Scale, torch::Tensor const& mat2Scale, int64_t output_buffer_kind = 0,
```
brb-nv, line 55: Dumb question: Why not use `BufferKind` enum instead of `int64_t` for output_buffer_kind? Current usage seems error prone and we seem to be casting to `BufferKind` anyway.

Extends to all files using `int64_t output_buffer_kind`.

cpp/tensorrt_llm/thop/weightOnlyQuantGemm.cpp
```diff
@@ -113,15 +115,8 @@ at::Tensor WeightOnlyQuantGemmRunner::runGemm(at::Tensor const& mat_a, at::Tenso
...
-    at::Tensor out;
-    if (to_userbuffers)
-    {
-        out = torch_ext::create_userbuffers_tensor({m, real_n}, dtype).first;
-    }
-    else
-    {
-        out = at::detail::empty_cuda({m, real_n}, dtype, mat_a.device(), std::nullopt);
-    }
+    auto [out, _] = torch_ext::allocate_output(
```
brb-nv, line 120: Some of the calls to `allocate_output` pass down a group while the others don't (like this one). Can you please let me know when `group` is needed and when it's not?

tensorrt_llm/_torch/custom_ops/__init__.py
```diff
@@ -1,22 +1,24 @@
...
 from ..cute_dsl_utils import IS_CUTLASS_DSL_AVAILABLE
 from ..flashinfer_utils import IS_FLASHINFER_AVAILABLE
-from ..modules.attention import attn_custom_op_inplace, mla_custom_op_inplace
 from .cpp_custom_ops import _register_fake
-from .torch_custom_ops import bmm_out
+from .torch_custom_ops import BufferKind, bmm_out
 from .trtllm_gen_custom_ops import fp8_block_scale_moe_runner
 from .userbuffers_custom_ops import add_to_ub, copy_to_userbuffers, matmul_to_ub
 
+# Attention custom ops (attn_custom_op_inplace, mla_custom_op_inplace) are defined in
```
brb-nv, line 9: Can you please let me know what about this MR is introducing circular imports (here and linear_common.py)? If this wasn't a problem earlier, is it possible to modify this MR to avoid circular imports?

brb-nv (commented):
Left some comments.

brb-nv:
> > > This PR implements this in fused python operations (GEMM and MOE) to enable output directly in a custom buffer if NCCL_SYMMETRIC is chosen.
> > 
> > 
> > Naive question: Is it possible to always use registered buffers regardless of NCCL_SYMMETRIC being chosen?
> 
> Yes — using a registered buffer instead of a non-registered one does not cause performance degradation, regardless of whether NCCL_SYMMETRIC is enabled.
> 
> The main drawback is that memory for symmetrically registered buffers is currently managed outside of Python, so it’s possible to over-allocate and hit an OOM error if used unnecessarily. It’s mostly safe to always enable this for AllReduce, unless you’re already close to memory limits.
> 
> You could use a symmetric registered memory pool for all operations, but doing this through PyTorch requires a newer version than the one currently in the container — so that’s planned for later.
> 
> Some inference frameworks register their entire memory pool, which works but adds complexity since communication operations must align buffer offsets across ranks. With careful coordination, it’s still a viable approach overall.
> 
> TLDR. Yes we could turn it on for all strategies. But in order to not risk OOM problems it would be better to have a graceful fallback.

I feel it's cleaner (if-else in modeling_xyz.py could be avoided. Hence, the question).

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -1169,17 +1172,33 @@ def _compute_routed_output():
         if not do_finalize:
             return [shared_output, *routed_output]
         else:
+            if not isinstance(shared_output, torch.Tensor):
+                final_hidden_states = shared_output + routed_output
+                if not self.use_dp and self.mapping.tp_size > 1:
+                    final_hidden_states = self.allreduce(
+                        final_hidden_states,
+                        all_reduce_params=final_all_reduce_params)
+                return final_hidden_states
```
kaiyux, line 1182: are there existing tests to protect the changes here?

kaiyux (approved):
approving on behalf of trt-llm-torch-models-deepseek-devs

tensorrt_llm/_torch/autotuner.py
nv-lschneider: Isn't used anymore. Removing.
We may have to reintroduce to differentiate .AUTO autotuning choice. Later.

tensorrt_llm/_torch/modules/linear_common.py
nv-lschneider: This is rebasing mistake. And I am removing it.
Thanks for catching it.

tensorrt_llm/_torch/modules/attention.py
nv-lschneider: The NCCL name for this is registered symmetric windows.
They are backed by NCCL windows. So I don't see how we can eliminate window from the anme.
But I am changing it to nccl symmetric memory window to make this more clear.

cpp/tensorrt_llm/thop/fp8RowwiseGemm.cpp
nv-lschneider: This is deliberate choice for the compilation in to pytorch/dynamo graphs.
The graphs compile better with native int types, than custom types.

cpp/tensorrt_llm/thop/weightOnlyQuantGemm.cpp
nv-lschneider: Good catch. This is actually deliberate here and the only time no group is passed.

The `Int8WeightOnlyLinearMethod` that calls this, does not support a nccl window tensor output.
And group is not needed for non-nccl window tensors.
But doing quietle like it was here is probably not the best idea.
So, I changed the API that you need to pass something (and not forget).
So now it passes c10:nullopt to expllictly signal group is intentionally not passed here.
And I added a comment to explain that.

tests/unittest/_torch/thop/parallel/test_fp4_linear.py
nv-lschneider: This is not a good test place for it.
The test is not multi-GPU and UserBuffers needs extra setup and multiGPUs.
And NCCL needs multi-GPU setup, so would just default back to Default anyways.

I am considering if it makes sense to have a dedicated and useful test for this.

brb-nv (approved):
Approving to unblock. I think the only open comment is related to adding more tests. Kindly take care of that before merging.

tests/unittest/_torch/thop/parallel/test_fp4_linear.py
nv-lschneider: I added a dedicated test to run with all 3 buuferkinds.

tensorrt_llm/_torch/models/modeling_deepseekv3.py
nv-lschneider: Can you clarify your comment?
I didn't added extra tests for this change, but expect the existing CI to catch issues if they exist.

nv-lschneider:
Waiting for multi-gpu CI runs to merge #13085 to enable ray based failing multi-gpu bugs.

Wanli-Jiang (approved):

hyukn (approved):
Thanks for the great effort. LGTM.

PR #11677, opened 2026-02-24
[TRTLLM-9939][perf] Short-sequence MHA optimization for DSA MLA prefill

kaiyux:
## Summary

For DSA models (DeepSeek V3.2), the MLA prefill absorption path requires two extra BMMs (Q-absorption and V-projection) and uses a larger head_dim (`kv_lora_rank + qk_rope_head_dim` = 576 vs 192). For short sequences, this overhead dominates. This PR adds an alternative dense MHA path that bypasses the absorption path for short context batches.

**Key changes:**
- When total packed context tokens <= threshold, `forward_context_dsa` dispatches to `forward_context()` (kv_b_proj expansion + standard attention) instead of the absorption/sparse MLA path
- `forward_context()` handles both fresh and cached tokens by dispatching to `forward_context_with_cached_kv` or `forward_context_with_chunked_prefill` as needed
- A dedicated `TrtllmAttention` dense backend (`self.mha`) is initialized with `sparse_attention_config=None` for the standard MHA path, unified with the existing non-DSA MHA creation
- When the short MHA path handles all context and there are no generation tokens, the DSA indexer (linear projections + attention + top-k selection) is skipped entirely
- Enabled by default: `TRTLLM_MLA_SHORT_SEQ_MHA_THRESHOLD=10240` (total packed tokens). Only activates when `rope_fusion=True` and `cp_size == 1`

## Details

**Dispatch guard** (`_should_use_short_mha`; all must be true):
- `short_seq_mha_threshold > 0`
- `apply_rotary_emb is False` (rope_fusion path — prevents double-RoPE)
- `cp_size == 1` (Context Parallelism not active)
- `position_ids` available
- `num_ctx_tokens <= threshold`

**Files changed:**
- `tensorrt_llm/_torch/modules/attention.py` — `_should_use_short_mha` guard, short-MHA dispatch in `forward_context_dsa`, unified `self.mha` initialization, indexer skip logic
- `tests/unittest/_torch/attention/sparse/test_short_seq_mha.py` — unit tests covering single/multi-sequence correctness, chunked context correctness, threshold boundary conditions, and A/B comparison against the absorption path

## Test plan

- [x] Unit tests pass (`pytest tests/unittest/_torch/attention/sparse/test_short_seq_mha.py`)
- [x] Verify correctness: run DeepSeek V3 inference with `TRTLLM_MLA_SHORT_SEQ_MHA_THRESHOLD=0` (disabled) and compare outputs with threshold enabled (default `10240`)
- [x] Benchmark short-sequence prefill latency with threshold enabled vs disabled
- [x] Verify generation quality is unaffected (KV cache populated correctly)
- [x] Test with TP > 1 to ensure head partitioning is correct

### DIFF
--- a/tensorrt_llm/_torch/modules/attention.py
+++ b/tensorrt_llm/_torch/modules/attention.py
@@ -1507,10 +1507,10 @@ class MLA(nn.Module):
 
         # Check if the short-seq MHA path will handle context, in which case
         # the indexer (topk_indices) is not needed for context tokens.
-        # Disable when cached tokens exist: forward_context_default does not
-        # attend over cached KV from previous chunks/turns.
-        has_cached_ctx = getattr(attn_metadata, 'num_ctx_cached_tokens', 0) > 0
-        use_short_mha_for_ctx = (num_contexts > 0 and not has_cached_ctx
+        # The MHA path handles cached tokens via forward_context(), which
+        # dispatches to forward_context_with_cached_kv or
+        # forward_context_with_chunked_prefill as needed.
+        use_short_mha_for_ctx = (num_contexts > 0
                                  and self._should_use_short_mha(
                                      num_ctx_tokens, position_ids))
 
@@ -1625,9 +1625,10 @@ class MLA(nn.Module):
                               position_ids: Optional[torch.Tensor]) -> bool:
         """Check if the short-seq MHA optimization should be used for context.
 
-        This checks only token count vs threshold. Callers MUST separately
-        verify that there are no cached tokens (num_ctx_cached_tokens == 0),
-        because forward_context_default does not attend over cached KV.
+        This checks only new token count vs threshold.  When cached tokens
+        exist, the caller should route through forward_context() which
+        dispatches to the appropriate cached-KV handler
+        (forward_context_with_cached_kv or forward_context_with_chunked_prefill).
         """
         return (self.short_seq_mha_threshold > 0 and not self.apply_rotary_emb
                 and self.mapping.cp_size == 1 and position_ids is not None
@@ -1665,13 +1666,12 @@ class MLA(nn.Module):
         # using kv_b_proj expansion + standard attention instead.
         # See __init__ comment for rationale. topk_indices is not used
         # because dense attention is faster than sparse routing at this scale.
+        # forward_context() handles cached tokens by dispatching to
+        # forward_context_with_cached_kv or forward_context_with_chunked_prefill.
         num_ctx_tokens = q.shape[0]
-        has_cached = getattr(attn_metadata, 'num_ctx_cached_tokens', 0) > 0
-        if not has_cached and self._should_use_short_mha(
-                num_ctx_tokens, position_ids):
-            return self.forward_context_default(q, compressed_kv, k_pe,
-                                                position_ids, attn_metadata,
-                                                output, latent_cache)
+        if self._should_use_short_mha(num_ctx_tokens, position_ids):
+            return self.forward_context(q, compressed_kv, k_pe, position_ids,
+                                        attn_metadata, output, latent_cache)
 
         if get_sm_version() >= 100:
             return self.forward_absorption_context(q,
--- a/tensorrt_llm/_torch/modules/attention.py
+++ b/tensorrt_llm/_torch/modules/attention.py
@@ -28,7 +28,7 @@ from ..distributed import (AllReduceParams, HelixAllToAllNative, alltoall_helix,
 from ..model_config import ModelConfig
 from ..peft.lora.layer import LoraLayer, LoraModuleType
 from ..utils import (Fp4QuantizedTensor, get_model_extra_attrs,
-                     is_torch_compiling, maybe_compile, maybe_compiled_cat,
+                     is_torch_compiling, maybe_compiled_cat,
                      maybe_compiled_copy_)
 from .linear import Linear, TensorParallelMode, WeightMode, WeightsLoadingConfig
 from .multi_stream_utils import maybe_execute_in_parallel
@@ -872,22 +872,6 @@ def fp8_block_scaling_bmm_out(
                       out=out)
     else:
         raise NotImplementedError(f"SM{sm_version} is not supported")
-
-
-@maybe_compile(dynamic=True)
-def _short_mha_reshape_kv(kv, k_pe_roped, num_heads_tp, qk_nope_head_dim,
-                           v_head_dim, qk_rope_head_dim, qk_head_dim):
-    """Fused split + reshape + expand + cat for short-seq MHA KV construction."""
-    k_nope, v = kv.split(
-        [num_heads_tp * qk_nope_head_dim, num_heads_tp * v_head_dim], -1)
-    k_nope_r = k_nope.view(-1, num_heads_tp, qk_nope_head_dim)
-    k_pe_expanded = k_pe_roped.view(-1, 1,
-                                    qk_rope_head_dim).expand(
-                                        -1, num_heads_tp, -1)
-    k = torch.cat([k_nope_r, k_pe_expanded], dim=-1)
-    k = k.view(-1, num_heads_tp * qk_head_dim)
-    v = v.view(-1, num_heads_tp * v_head_dim)
-    return k, v
 
 
 class MLA(nn.Module):
@@ -1769,15 +1753,31 @@ class MLA(nn.Module):
         ctx_position_ids = position_ids[..., :num_ctx_tokens]
         [k_pe_roped] = self._rotary_emb_mha(ctx_position_ids, [k_pe])
 
-        # Step 3: Expand KV via kv_b_proj, construct full K, flatten.
-        # The reshape/cat ops are compiled via _short_mha_reshape_kv to fuse
-        # split + view + expand + cat into fewer kernels.
+        # Step 3: Expand KV via kv_b_proj.
         kv = self.kv_b_proj(compressed_kv)
-        k, v = _short_mha_reshape_kv(kv, k_pe_roped, self.num_heads_tp,
-                                      self.qk_nope_head_dim, self.v_head_dim,
-                                      self.qk_rope_head_dim, self.qk_head_dim)
+        k_nope, v = kv.split(
+            [
+                self.num_heads_tp * self.qk_nope_head_dim,
+                self.num_heads_tp * self.v_head_dim
+            ],
+            -1,
+        )
 
-        # Step 4: Compute attention via the fused MHA backend.
+        # Step 4: Construct full K = [k_nope, k_pe_roped] per head.
+        k_nope_r = k_nope.view(-1, self.num_heads_tp, self.qk_nope_head_dim)
+        k_pe_expanded = k_pe_roped.view(-1, 1,
+                                        self.qk_rope_head_dim).expand(
+                                            -1, self.num_heads_tp, -1)
+        k = maybe_compiled_cat([k_nope_r, k_pe_expanded], dim=-1)
+        # k: [num_ctx_tokens, num_heads_tp, qk_head_dim]
+
+        # Step 5: Flatten Q, K, V to 2D packed layout for the fused attention
+        # backend, which natively handles variable-length sequences via
+        # attn_metadata (no padding/batching needed).
+        k = k.view(-1, self.num_heads_tp * self.qk_head_dim)
+        v = v.view(-1, self.num_heads_tp * self.v_head_dim)
+
+        # Step 6: Compute attention via the fused MHA backend.
         # latent_cache=None signals that RoPE and KV cache append were already
         # done in Step 1 (mla_rope_append_paged_kv_assign_q), matching the
         # pattern used by forward_context_with_cached_kv.
--- a/tensorrt_llm/_torch/modules/attention.py
+++ b/tensorrt_llm/_torch/modules/attention.py
@@ -1507,7 +1507,10 @@ class MLA(nn.Module):
 
         # Check if the short-seq MHA path will handle context, in which case
         # the indexer (topk_indices) is not needed for context tokens.
-        use_short_mha_for_ctx = (num_contexts > 0
+        # Disable when cached tokens exist: forward_context_default does not
+        # attend over cached KV from previous chunks/turns.
+        has_cached_ctx = getattr(attn_metadata, 'num_ctx_cached_tokens', 0) > 0
+        use_short_mha_for_ctx = (num_contexts > 0 and not has_cached_ctx
                                  and self._should_use_short_mha(
                                      num_ctx_tokens, position_ids))
 
@@ -1620,7 +1623,12 @@ class MLA(nn.Module):
 
     def _should_use_short_mha(self, num_ctx_tokens: int,
                               position_ids: Optional[torch.Tensor]) -> bool:
-        """Check if the short-seq MHA optimization should be used for context."""
+        """Check if the short-seq MHA optimization should be used for context.
+
+        This checks only token count vs threshold. Callers MUST separately
+        verify that there are no cached tokens (num_ctx_cached_tokens == 0),
+        because forward_context_default does not attend over cached KV.
+        """
         return (self.short_seq_mha_threshold > 0 and not self.apply_rotary_emb
                 and self.mapping.cp_size == 1 and position_ids is not None
                 and num_ctx_tokens <= self.short_seq_mha_threshold)
@@ -1658,7 +1666,9 @@ class MLA(nn.Module):
         # See __init__ comment for rationale. topk_indices is not used
         # because dense attention is faster than sparse routing at this scale.
         num_ctx_tokens = q.shape[0]
-        if self._should_use_short_mha(num_ctx_tokens, position_ids):
+        has_cached = getattr(attn_metadata, 'num_ctx_cached_tokens', 0) > 0
+        if not has_cached and self._should_use_short_mha(
+                num_ctx_tokens, position_ids):
             return self.forward_context_default(q, compressed_kv, k_pe,
                                                 position_ids, attn_metadata,
                                                 output, latent_cache)
--- a/tensorrt_llm/_torch/modules/attention.py
+++ b/tensorrt_llm/_torch/modules/attention.py
@@ -1126,29 +1126,6 @@ class MLA(nn.Module):
         mscale = yarn_get_mscale(scaling_factor, mscale_all_dim)
         q_scaling = 1.0 / (mscale * mscale)
 
-        if not self.is_dsa:
-            self.mha = create_attention(
-                config.attn_backend,
-                self.layer_idx,
-                self.num_heads_tp,
-                head_dim=self.qk_head_dim,
-                num_kv_heads=self.num_key_value_heads_tp,
-                pos_embd_params=pos_embd_params,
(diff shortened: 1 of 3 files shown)


tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1557,6 +1603,118 @@ def forward_context_dsa(
...
+            # lengths correctly with causal masking.
+            offset = 0
+            for i in range(num_contexts):
+                slen = seq_lens[i].item()
+                # [1, H, slen, D]
+                q_seq = q_r[offset:offset + slen].unsqueeze(0).transpose(1, 2)
+                k_seq = k[offset:offset + slen].unsqueeze(0).transpose(1, 2)
+                v_seq = v_r[offset:offset + slen].unsqueeze(0).transpose(1, 2)
+
+                attn_out_seq = F.scaled_dot_product_attention(
```
PerkzZheng: just curious why SDPA is adopted. Is it easier to implement with ? Also we might want to understand its impacts on task accuracy (seqlen > 2k).

kaiyux (commented):
## Code Review: Short-Sequence MHA Optimization for DeepSeek V3.2 MLA

### Overview

This PR adds an alternative MHA path for DeepSeek V3.2 MLA during short-sequence prefill. Instead of the absorption path (which requires extra BMMs and uses head_dim=576), it expands KV via `kv_b_proj` and uses standard SDPA with head_dim=192. Well-motivated optimization — the crossover point where absorption overhead exceeds the benefit is real for short sequences.

**Overall: Good change with solid design.** A few issues below, ranging from correctness concerns to nits.

---

### Issues

#### 1. [Medium] `kv_b_proj` cost may dominate for the threshold range

`kv_b_proj` is a large projection (`kv_lora_rank → num_heads * (qk_nope_head_dim + v_head_dim)`), which for DeepSeek V3 is `512 → 128 * (128 + 128) = 32768`. For very short sequences this matmul is cheap, but as you approach the threshold, this projection's cost grows linearly with `num_tokens`. The default threshold of 0 (disabled) is wise, but the PR description and comments should mention that the optimal threshold is model/hardware-dependent and needs benchmarking to set properly.

#### 2. [Medium] Per-sequence Python loop for packed batches (lines 1709-1724)

The multi-sequence fallback iterates in Python, issuing a separate SDPA kernel per sequence. For packed batches with many short sequences, this Python-loop overhead and lack of kernel fusion could negate the benefit. The TODO mentions NestedTensor — that would be worth doing before merging since the packed-batch case is actually the most common deployment scenario (continuous batching packs multiple requests).

At minimum, consider using `torch.nn.attention.flex_attention` or `torch._nested_tensor_from_mask` to batch the variable-length sequences.

#### 3. [Low] Environment variable read at `__init__` time — no hot-reload

The threshold is read from `os.environ` once during model construction. This means changing the env var requires restarting the server. This is fine for now but worth noting in the documentation. An alternative would be reading it per-forward (with caching) or making it a config parameter in `LlmArgs`.

#### 4. [Low] `self.softmax_scale` correctness for MHA path

The existing `self.softmax_scale` is set based on the absorption path's head dimensions. Verify that `self.softmax_scale` equals `1/sqrt(qk_head_dim)` (i.e., `1/sqrt(192)`) and not `1/sqrt(kv_lora_rank + qk_rope_head_dim)` (i.e., `1/sqrt(576)`). If `softmax_scale` was tuned for the absorption path, this MHA path would produce incorrect attention weights.

#### 5. [Nit] `position_ids` slicing assumes batch-first layout

Line 1659: `ctx_position_ids = position_ids[..., :num_ctx_tokens]` — the `...` makes this work for both `[S]` and `[B, S]` shapes, but the comment should clarify the expected shape contract. If `position_ids` is `[B, S]` where `S` includes generation tokens, does slicing to `num_ctx_tokens` correctly handle the packed layout?

#### 6. [Nit] Agree with CodeRabbit's suggestions
- The `raise ... from err` fix has already been addressed in commit `55c277ad`.
- The `isinstance` assertion before `cast(TrtllmAttention, self.mqa)` is a good defensive practice.

---

### Questions (echoing @PerkzZheng's review comment)

- **Why SDPA over FlashAttention directly?** SDPA dispatches to FlashAttention under the hood for causal masks, but the dispatch has overhead. For a short-seq optimization, calling the FA kernel directly (if available) would be more aligned with the goal.
- **Accuracy at boundary thresholds:** Has the output been validated for numerical equivalence (within tolerance) between the MHA path and absorption path at the threshold boundary? The two paths compute mathematically equivalent results but with different numerical orderings — worth a regression test.

---

### What's good

- Clean separation: new method `forward_context_short_mha` is well-isolated with an NVTX marker
- Thorough guard conditions (rope_fusion, position_ids, max_ctx_seq_len) prevent misuse
- Good docstring explaining the 5-step algorithm
- Env-var gating with default=disabled is the right approach for an experimental optimization
- KV cache is still populated correctly (via `mla_rope_append_paged_kv_assign_q`), ensuring generation phase works

kaiyux (commented):
## Review: Short-sequence MHA optimization

Good iteration since the self-review — the batched SDPA approach for multi-sequence is a solid improvement over the per-sequence loop. One potential correctness issue and a couple of performance observations below.

### Summary of findings

| # | Severity | Issue |
|---|----------|-------|
| 1 | **High** | Chunked prefill correctness — SDPA misses cached KV tokens |
| 2 | Low | CPU→GPU sync in multi-sequence path |
| 3 | Nit | `pad_sequence` + `transpose` produces non-contiguous tensors for SDPA |

kaiyux (commented):
## Code Review: Short-Sequence MHA Optimization for DeepSeek V3.2 MLA

Nice optimization — the insight that the absorption path's overhead (extra BMMs + head_dim 576 vs 192) dominates for short sequences is well-motivated, and the implementation has evolved nicely across the 13 commits (the progression from per-sequence SDPA loops → batched pad+SDPA → fused TrtllmAttention backend is good). A few issues below:

---

### 1. [High] Duplicated 5-condition dispatch guard — divergence risk

The exact same guard logic appears in two places:
- `forward_impl_with_dsa` (lines 1518-1524) — decides whether to skip the indexer
- `forward_context_dsa` (lines 1659-1663) — decides whether to call `forward_context_short_mha`

If someone updates one without the other, the indexer-skip and the actual dispatch will disagree. For example, adding a new guard condition to `forward_context_dsa` but forgetting `forward_impl_with_dsa` would cause `topk_indices=None` to flow into the absorption path, which requires it.

**Suggestion:** Extract into a helper method:
```python
def _should_use_short_mha(self, num_ctx_tokens: int, position_ids) -> bool:
    return (self.short_seq_mha_threshold > 0
            and not self.apply_rotary_emb
            and self.mapping.cp_size == 1
            and position_ids is not None
            and num_ctx_tokens <= self.short_seq_mha_threshold)
```

### 2. [High] Default threshold of 10240 enables the optimization silently

`TRTLLM_MLA_SHORT_SEQ_MHA_THRESHOLD` defaults to `'10240'` (line 1083), meaning this path is **on by default** for all DSA MLA models. For a new experimental optimization that changes the compute path and produces numerically different results (the A/B test tolerance is 15%), defaulting to enabled is risky. If a user upgrades TRT-LLM and gets subtly different prefill outputs, debugging would be difficult since there's no log message when the path activates.

**Suggestion:** Default to `'0'` (disabled, opt-in) until the optimization has been validated in production, or at minimum add a `logger.info` when the threshold is configured and a `logger.debug` on first dispatch hit.

### 3. [Medium] Chunked prefill correctness — still relevant?

The self-review (commit `58c4258`) flagged that the short-seq MHA path only attends over new context tokens and misses cached KV from previous chunks. The later switch to a fused TrtllmAttention backend (commit `7fbf9c3`) may have addressed this since TrtllmAttention uses the paged KV cache. But the threshold guard at line 1659 uses `num_ctx_tokens = q.shape[0]`, which is the number of *new* tokens in the current chunk, not the full sequence length. Could you confirm the fused backend correctly attends to cached tokens from prior chunks? If so, a comment documenting this would help.

### 4. [Medium] `_short_seq_mha` backend created even for non-DSA MLA models

The guard at line 1096 is `self.short_seq_mha_threshold > 0 and not self.apply_rotary_emb`. A non-DSA model with `rope_fusion=True` would create the extra RotaryEmbedding + TrtllmAttention backend unnecessarily. While the runtime dispatch in `forward_context_dsa` prevents actual use (non-DSA doesn't call that path), it wastes memory.

**Suggestion:** Add `and self.is_dsa` (or equivalent) to the init-time guard.

### 5. [Low] `position_ids` passed unsliced to `forward_context_short_mha`

At line 1562, `position_ids` is passed in full (covering all tokens including generation), while `q`, `compressed_kv`, and `k_pe` have been sliced to `[:num_ctx_tokens]`. Inside `forward_context_short_mha`, line 1733 slices it: `ctx_position_ids = position_ids[..., :num_ctx_tokens]`. This works but is inconsistent with how all other tensors are handled — slice at the call site for clarity.

### 6. [Low] No negative threshold validation

A user can set `TRTLLM_MLA_SHORT_SEQ_MHA_THRESHOLD=-1`. The `int()` parse succeeds, and the guard `> 0` correctly prevents activation, but the semantically invalid value is stored silently. A bounds check after parsing would be cleaner.

### 7. [Nit] `_apply_rope_single_packed` — `is_neox` branching at trace time

The `if is_neox` branch (line 56) will be a Python bool at trace time, so `@maybe_compile` will create separate compiled graphs for each case. This is fine for correctness but worth noting — if the model's `is_neox` value changes between layers (unlikely but possible with mixed configs), each variant gets separately compiled.

---

### Tests (test_short_seq_mha.py)

**Good:**
- Comprehensive coverage: single/multi-sequence, threshold boundary, disabled path, A/B vs absorption
- Proper SM90 skip annotation
- KV cache cleanup with `kv_cache_manager.shutdown()`
- Thread leak suppression documented

**Concerns:**

1. **A/B test tolerance is very relaxed** — `rtol=0.15, atol=0.15` (line ~793-794). A 15% relative tolerance could mask real regressions. Consider tightening or adding a mean-error assertion alongside the max-error check.

2. **No test for mixed context+generation batch** — the most complex dispatch path (lines 1528-1530 where `topk_indices` is computed only for generation, and line 1561 where it's sliced with a None check) is untested. This is the continuous-batching scenario where the indexer-skip matters.

3. **`kv_cache_manager.shutdown()` not in `finally` blocks** — if an assertion fails before shutdown, resources leak in CI.

4. **`print()` statements** (lines ~466, 572, 666, 795) are effectively dead code since pytest captures stdout. Consider `logging.info()` or removing them.

---

### What's good

- Clean method isolation with NVTX marker
- Good docstring with numbered algorithm steps
- Correct KV cache handling via `mla_rope_append_paged_kv_assign_q`
- Progressive optimization across commits (especially the shift from SDPA to fused TrtllmAttention)
- `maybe_compiled_cat` for the K construction avoids kernel launch overhead
- Guard conditions are thorough (rope_fusion, CP, position_ids)

kaiyux (commented):
## Code Review: Short-sequence MHA optimization for DeepSeek V3.2 MLA prefill

Overall this is a well-structured optimization with clear rationale. The approach of bypassing the absorption path for short prefills by reusing `forward_context_default` is clean. A few issues to address:

---

### 1. Redundant `self.mha` creation for non-DSA models (bug-adjacent)

The new short-seq MHA initialization block runs unconditionally for any MLA model with threshold > 0 and `apply_rotary_emb=False`:

```python
if self.short_seq_mha_threshold > 0 and not self.apply_rotary_emb:
    self.mha = create_attention(...)
```

For non-DSA MLA models (e.g., standard DeepSeek V2), `self.mha` was already created a few lines earlier by the `if not self.is_dsa:` block. The new code silently **overwrites** it with a second `create_attention(...)` call. Both happen to be equivalent (since `config.sparse_attention_config` is `None` for non-DSA), so this is functionally harmless but wasteful — it creates the attention backend twice per layer.

**Suggested fix:** Gate on `self.is_dsa`:
```python
if self.is_dsa and self.short_seq_mha_threshold > 0 and not self.apply_rotary_emb:
    self.mha = create_attention(..., sparse_attention_config=None)
```

This also makes the intent clearer: the short-seq MHA backend is only needed for DSA models.

---

### 2. `create_weights` guard change is correct but deserves a comment

The change from `if not self.is_dsa:` to `if self.mha is not None:` is semantically correct — it now correctly handles the new DSA-with-short-mha case. However, the comment above it (`# self.mha/mqa has no weights but has states...`) should be updated to note that `self.mha` is now non-None for DSA models when the short-seq optimization is active.

---

### 3. Commit history needs squashing

The PR has 18 commits showing the full development iteration (initial impl → fix review comments → fix tests → fix device placement → batch SDPA → use fused kernel → skip indexer → change threshold semantics → inline reshape → extract guard → reuse forward_context_default → cleanup). Per the project guidelines ("squash fix-on-fix commits before PR submission"), this should be consolidated into 1-3 logical commits before merge, e.g.:
1. Core implementation: short-seq MHA dispatch + indexer skip
2. Unit tests

---

### 4. Relaxed A/B test tolerance (rtol=0.15, atol=0.15)

The `test_short_mha_agrees_with_absorption_path` uses 15% relative tolerance, which is quite loose. Given both paths compute the same mathematical operation (just with different intermediate representations), the tolerance should be tighter. If the two paths genuinely diverge by 15% on some elements, that could mask real correctness issues. Consider:
- Investigating whether the tolerance can be tightened (e.g., 0.05-0.08)
- If the large tolerance is inherent to the bf16 absorption path's numerics, document why in the test

---

### 5. `forward_context_default`: uninitialized k_pe region is subtle

When the short-seq path calls `forward_context_default` with `apply_rotary_emb=False`, the rope portion of `k` (`k[..., qk_nope_head_dim:]`) remains uninitialized from `torch.empty_like`. This works because TrtllmAttention with rope fusion handles k_pe via the `latent_cache` internally — but it's non-obvious and could surprise future readers. A one-line comment at the `if self.apply_rotary_emb:` guard in `forward_context_default` noting "rope portion handled by fused kernel via latent_cache when rope_fusion=True" would help.

---

### 6. CI: Docstring coverage check failing

The "Check PR Checklist Resolution" CI check is failing, with docstring coverage at 60% (required 80%). The new `forward_context_dsa` signature change (added `position_ids` parameter) and the dispatch logic in `forward_impl_with_dsa` may need docstring updates.

---

### Minor nits:
- The env var default of `'10240'` means this optimization is **enabled by default** for all DSA models. The PR description and commit messages mention this, but it might be worth adding a `logger.info` at init time when the short-seq path is activated, similar to how other optimization toggles are logged in TRT-LLM.
- Tests are thorough — good coverage of boundary conditions, disabled path, and A/B comparison.

lfr-0531 (approved):

kaiyux (commented):
## Code Review: Short-Sequence MHA Optimization for DeepSeek V3.2 MLA Prefill

### Overall Assessment

Well-motivated optimization with a clean implementation. The insight that the absorption path's overhead (extra BMMs + head_dim 576 vs 192) dominates for short sequences is solid, and the approach of redirecting through `forward_context` → `forward_context_default` (the existing dense MHA path) is the right way to reuse infrastructure.

The evolution visible from PR comments (per-sequence SDPA → batched SDPA → reusing TrtllmAttention via `forward_context`) shows good iteration. The final design is much cleaner than the alternatives.

---

### Issues

#### 1. [Medium] Env var read at `__init__` time — not configurable at runtime

The threshold is read from `os.environ` once at module construction:

```python
_threshold_str = os.environ.get('TRTLLM_MLA_SHORT_SEQ_MHA_THRESHOLD', '10240')
self.short_seq_mha_threshold = int(_threshold_str)
```

This means:
- Changing the env var after model initialization has no effect
- The `self.mha` backend is conditionally created based on the threshold — so even if you could change the threshold later, the backend might not exist

This is fine for deployment but worth noting explicitly in the docstring/comment that the threshold is **read once at init time**. Currently the comment says "Enabled by default with `TRTLLM_MLA_SHORT_SEQ_MHA_THRESHOLD` env var" but doesn't clarify this is init-time-only.

#### 2. [Medium] `_should_use_short_mha` checks `num_ctx_tokens` (total packed tokens) — not per-sequence length

The guard checks total packed context tokens vs threshold:
```python
num_ctx_tokens <= self.short_seq_mha_threshold
```

For a batch of many short sequences that individually would benefit from the MHA path but collectively exceed the threshold (e.g., 20 sequences × 600 tokens = 12000 > 10240), the optimization won't activate. This is a reasonable design choice for simplicity (avoids partial-batch dispatch), but the threshold semantics should be documented clearly — users might expect it to be a per-sequence threshold.

Consider adding a one-line comment in `_should_use_short_mha`: `# Note: threshold applies to total packed tokens, not per-sequence length.`

#### 3. [Low] Indexer still runs when `use_short_mha_for_ctx=True` but `num_generations > 0`

```python
if use_short_mha_for_ctx and num_generations == 0:
    topk_indices = None
else:
    topk_indices = self.indexer(...)  # runs even if only gen needs it
```

When the short MHA path handles context but there are also generation tokens, the indexer runs on the full batch (all tokens including context tokens). The context portion of `topk_indices` is then unused since `forward_context_dsa` will take the short MHA path. This is correct but wasteful — the indexer computes routing for context tokens that won't use them.

This is probably fine for now since the mixed context+generation case is less common during prefill-heavy workloads, but could be a follow-up optimization.

#### 4. [Low] `sparse_attention_config` handling for non-DSA MLA

```python
_short_seq_mha = (self.is_dsa and self.short_seq_mha_threshold > 0
                  and not self.apply_rotary_emb)
if not self.is_dsa or _short_seq_mha:
    self.mha = create_attention(
        ...
        sparse_attention_config=(None if _short_seq_mha else
                                 config.sparse_attention_config),
    )
```

For non-DSA models (`self.is_dsa = False`), this passes `config.sparse_attention_config` to `self.mha`. In the original code, non-DSA also passed `sparse_attention_config`. So this is equivalent, but it's a bit subtle — the `_short_seq_mha` variable name suggests it's only about DSA, yet the `if not self.is_dsa` branch also uses it for flow control. A small comment would help.

---

### Nits

1. **`import os` at module top**: The diff adds `import os` at line 1 of the file. This is fine, but given that `os.environ` is only read once during `__init__`, you could also use `tensorrt_llm._utils` if there's already an env-reading utility there. Not blocking.

2. **Test tolerances are generous**: `rtol=0.05, atol=0.05` in the reference comparison and `rtol=0.08, atol=0.08` for the A/B comparison. These are reasonable for bf16 attention with different code paths, but if the team has tighter tolerance standards, might be worth tightening.

3. **Test file is comprehensive**: The test coverage is solid — pure prefill, threshold boundary, threshold=0 disabled, above-threshold fallback, A/B vs absorption, and chunked context. Good use of parametrized specs.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | Medium | Env var threshold is init-time-only — document this |
| 2 | Medium | Threshold is total-packed-tokens, not per-sequence — clarify semantics |
| 3 | Low | Indexer runs on context tokens even when short MHA will handle them |
| 4 | Low | `sparse_attention_config` flow for non-DSA could use a comment |

The design is sound — reusing `forward_context` → `forward_context_default` handles chunked prefill, cached KV, and edge cases correctly without duplicating logic. The indexer skip when `num_generations == 0` is a nice bonus optimization. Ship it with the doc/comment improvements.

yuxianq (approved):

PR #11729, opened 2026-02-25
Fix multimodal + speculative decoding crash (Eagle3/MTP)

yilian49:
## Summary

Fixes a crash when running **any multimodal model** (Llama 4, Gemma 3 VL, Qwen 2/3 VL, etc.) with **speculative decoding** (Eagle3 one-model, Eagle two-model, or MTP).

### Root Cause

`fuse_input_embeds()` in `modeling_multimodal_utils.py` returns `input_ids=None` after fusing text and vision/audio embeddings into `inputs_embeds`. This is correct for text-only inference (the backbone model only needs one of `input_ids` or `inputs_embeds`), but the speculative decoding worker still needs the original token IDs — specifically, `prepare_1st_drafter_inputs()` in `eagle3.py` calls `input_ids.shape[0]` on the returned `None`, causing an `AttributeError`.

### Fix

Two changes:

1. **`fuse_input_embeds()`**: Return the original `input_ids` instead of `None`. The token IDs are still valid after embedding fusion — they just aren't needed by the backbone model when `inputs_embeds` is already computed.

2. **`SpecDecOneEngineForCausalLM.forward()`, `Eagle3ForCausalLM.forward()`, `MistralLarge3EagleForCausalLM.forward()`**: Pass `input_ids=None` to the backbone model when `inputs_embeds` is set (to avoid the XOR validation that many models enforce: "You cannot specify both input_ids and inputs_embeds"). The original `input_ids` is preserved and passed to the speculative decoding worker which needs it.

### Affected Models

All 10 multimodal architectures that call `fuse_input_embeds`: Llama 4, Gemma 3 VL, Qwen 2 VL, Qwen 3 VL, Mistral, LLaVA-Next, VILA, Phi-4 MM, Nemotron Nano, HyperCLOVA X.

## Test plan

- [x] Verified Llama 4 Scout (17B-16E) + Eagle3 speculative decoding with image inputs no longer crashes
- [ ] Verify other multimodal models with speculative decoding
- [ ] Run existing multimodal unit tests to confirm no regression

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token ID handling in multimodal model processing to ensure input data is consistently propagated through the pipeline.
  * Improved speculative decoding compatibility with multimodal inputs by refining parameter forwarding logic.


2ez4bz (commented):
Thanks for the fix! How involved would it be to have this scenario checked in CI?

PR #11826, opened 2026-03-02
[None][fix] Fix typos, grammar, and formatting in comments and docstrings

kaiyux:
## Summary
- Fix "it's" → "its" (possessive) across multiple files
- Fix various typos in comments and docstrings (e.g., "enale" → "enable", "recieve" → "receive", "concatnated" → "concatenated", missing spaces)
- Fix grammar errors and improve clarity in inline comments
- Auto-formatted long lines in `openai_server.py` (triggered by pre-commit hooks)

## Test plan
- [x] No functional code changes — all modifications are in comments, docstrings, and formatting
- [x] Pre-commit hooks pass

## Release Notes

* **Bug Fixes**
  * Corrected logic condition in speculative inference error handling
  * Fixed typos and grammar in user-facing error messages

* **Documentation**
  * Improved clarity in docstrings and comments across the codebase

* **Breaking Changes**
  * `get_stats()` return type changed from dictionaries to JSON strings
  * Enhanced context manager support in request handling

### DIFF
--- a/tensorrt_llm/_torch/autotuner.py
+++ b/tensorrt_llm/_torch/autotuner.py
@@ -201,7 +201,7 @@ class TunableRunner(ABC):
 
         tactic==-1 has special meaning, means the fallback kernel which should be able to implement any shapes
         This fallback tactic is needed for 2 reasons:
-            * when the autotuner cannot find a valid tactic in it's cache.
+            * when the autotuner cannot find a valid tactic in its cache.
             * in eager mode, w/o autotunning the custom op should have at least one kernel, which makes the autotuning
               process an optional process, such that user can opt out.
 
--- a/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
+++ b/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
@@ -1069,7 +1069,7 @@ class FP8BatchedGemmRunner(TunableRunner):
     def get_dynamic_tensor_specs(cls) -> Tuple[DynamicTensorSpec, ...]:
         """Get the dynamic tensor specs for use with the AutoTuner."""
 
-        # These indices correspond to the 0th input tensor and it's first dimension
+        # These indices correspond to the 0th input tensor and its first dimension
         # i.e. we are tuning M where the first input tensor is of shape [B, M, K]
 
         MAT1_IDX = 0
--- a/tensorrt_llm/_torch/modules/attention.py
+++ b/tensorrt_llm/_torch/modules/attention.py
@@ -811,7 +811,7 @@ class Attention(nn.Module):
 
     def apply_qk_norm(self, q, k):
         raise NotImplementedError(
-            f"QK norm is not implemented for {self.__class__.__name__}."
+            f"QK norm is not implemented for {self.__class__.__name__}. "
             "Please override the `apply_qk_norm` method in the subclass.")
 
 
@@ -959,7 +959,7 @@ class MLA(nn.Module):
                 self)
             self.register_to_config = True
 
-        # only support one kind of sparse attention, dsa now.
+        # Currently only DSA sparse attention is supported.
         if config is not None and config.sparse_attention_config is not None and config.sparse_attention_config.algorithm == "dsa":
             self.is_dsa = True
         else:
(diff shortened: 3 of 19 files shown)


nvpohanh (approved):

QiJune (approved):
LGTM

kaiyux (commented):
## Review

Nice cleanup PR. Most of the typo/grammar fixes are correct and valuable. A few items worth addressing:

### Bug: `get_stats()` return type docstring is now wrong

In `llm.py:662`, the docstring was changed from `List[dict]` to `List[str]`, but the return type annotation is still `-> List[dict]`, and the underlying `IterationResult.get_results()` (`result.py:930`) calls `json.loads(data)` which returns Python dicts, not strings. The docstring should stay as `List[dict]`.

### Highlight: Real bug fixes buried in "typo" PR

Several changes here are genuine bug fixes, not just comment corrections. Consider calling these out more prominently in the PR description (and splitting if reviewers want separate CI signal):

1. **`py_executor.py:651` — `__exit__` signature fix**: `def __exit__(self)` → `def __exit__(self, exc_type, exc_val, exc_tb)`. This fixes a broken context manager protocol — Python passes 3 args to `__exit__`, so the old signature would `TypeError` when a `with` block encounters an exception.

2. **`mtp.py:1312` — inverted error message**: `token_count < max_num_requests` → `token_count > max_num_requests`. The code reaches this branch when `pad_len < 0`, which means `token_count > max_num_requests`. Original message was backwards.

3. **`sampler.py:2870` and `linear.py:1505` — misplaced docstrings**: Code statements before the docstring caused Python to treat the docstring as a dead string literal (not accessible via `__doc__`/`help()`). Moving them is a functional fix.

4. **`resource_manager.py:2204` — inverted assertion message**: The assert checks `!= SELFKONLY` but the message said "SELFKONLY is the only supported type" — logically backwards.

5. **`llm_utils.py:579` — missing negation**: `"model is loaded yet."` → `"model has not been loaded yet."` (assert fires when model is `None`).

### Minor nit

The `openai_server.py` changes are 453 additions / 282 deletions of pure yapf reformatting. Consider splitting this into its own commit or squashing separately — it makes the meaningful changes in this PR harder to review in the diff.

Overall: good catches, especially the `__exit__` fix, the misplaced docstrings, and the implicit string concatenation space fixes in `llm_args.py`. Just fix the `get_stats()` return type doc and this LGTM.

hyukn (approved):

kaiyux (commented):
Good PR — thorough pass over comments, docstrings, and error messages. I verified all the substantive (non-typo) changes and they look correct:

**Verified code-level fixes (not just comment/typo):**

1. **`mtp.py:1312` — error message comparison flipped** (`<` → `>`): Correct. The condition is `pad_len < 0` where `pad_len = max_num_requests - token_count`, so when it fires, `token_count > max_num_requests`. The old message had it backwards.

2. **`py_executor.py:651` — `__exit__` signature fixed** to accept `(exc_type, exc_val, exc_tb)`: Correct per Python's context manager protocol (`__enter__` is defined, so this class is used with `with` statements).

3. **`py_executor.py:3471` — `_terminate_requests` param renamed** `requests_to_pause` → `requests_to_terminate`: Safe — both call sites use positional args only.

4. **`modeling_utils.py:1548` — `set_fp8_context_fhma` → `set_fp8_context_fmha`**: Only used locally in `modeling_utils.py` (definition + one call in `optimize_model`), both updated. No external callers.

5. **`llm.py:796` — `built_enging_cfg` → `built_engine_cfg`**: Purely local variables, all references updated.

6. **`linear.py:2700` — assertion message `"non-unquantized"` → `"unquantized"`**: Correct. The guard is `not isinstance(self.quant_method, UnquantizedLinearMethod)`, so the feature is only for unquantized methods. The old double-negative was confusing.

7. **`sampler.py:2868` — docstring moved before first executable statement**: Correct per PEP 257 (old placement after `pin_memory = ...` made it a dead string, not a docstring).

8. **`llm.py:659` — `get_stats()` return type docstring `List[str]` → `List[dict]`**: Matches the actual type annotation `-> List[dict]` and the `json.loads()` in the implementation.

9. **`create_moe.py:381` — warning message now interpolates `{requested_backend}`** instead of hardcoding `"TRTLLM"`: Correct, since this branch also handles CUTEDSL requests.

10. **`linear.py:1654` — removed stray `*6`** from assertion f-string: Good catch.

**One minor nit** (non-blocking): The `openai_server.py` reformatting is ~450 added / ~280 deleted lines, which is a lot of diff noise for a typo-fix PR (just the `yapf: enale` → `yapf: enable` fix on line 91 was the real fix — the rest is auto-formatter reflow). Consider splitting formatter-only changes into a separate commit in future PRs so the substantive fixes are easier to review. But not worth re-splitting now.

LGTM.

sunnyqgg (approved):
LGTM

PR #11843, opened 2026-03-03
[None][feat] Small optimizations for mamba_mixer2.py decode

hnover-nv:
## Description
Add 2 optimizations for mamba2_mixer.py decoding path, with the groundwork for a third.  Saves about 2% on bf16, ssm fp32, concurrency 1 with the triton MTP kernels (default at the time I wrote it).

1. Eliminate the computation every pass of intermediate_state_indices.

We just need an arange up to :num_decodes.  The free way to do this is just have an arange you know is big enough, and make a view with a slice, as none of the boundary information gets into cuda code anyway.  This is what's done inside PythonMambaCacheManager.

But the largest value is the size of the intermediate state cache from the MambaCacheManager, which gets initialized with max_batch_size.  And we never have access to max_batch_size, and only have access to MambaCacheManager during the forward pass, and in theory MambaCacheManager could change and the new one could have a new configuration, although in practice it doesn't.  So we @functools.cache the "big" arange to the size of the MambaCacheManager's intermediate state, which is accessible through get_max_num_resources.  Actually today they are potentially different values, but this is guaranteed to be at least as large.  If the size of this vector becomes an issue, we can instead inspect the dimensions of the cache, but this seems cleaner.  This saves over a microsecond / layer.

2. Convert dt_d to fp32 in parallel to running conv1d, on a second stream.  At least when using cuda graphs.

This conversion was added in https://github.com/NVIDIA/TensorRT-LLM/pull/10757.  From context it may only be needed for the flashinfer kernel, which we don't currently use on the MTP path but will be using soon.  This adds about .2 microseconds to conv1d but totally hides the conversion time, saving over another microsecond per layer.  For the future, it's probably better to have the kernel do the conversion on load.

These two changes together achieve a little under a 2% speedup for bf16, ssm state fp32, concurrency 1, MTP=5: 
```
Request Generation Step Time
Baseline: mean: 0.007153, std: 0.003610, quantiles: {0.25: 0.006951, 0.5: 0.007037, 0.75: 0.007116}
Changes:  mean: 0.007025, std: 0.003645, quantiles: {0.25: 0.006813, 0.5: 0.006908, 0.75: 0.006994}
```

4. Add support for PDL between the triton selective_state_update and triton causal_conv1d_update, but don't turn it on.

selective_state_update should be largely memory bound.  The biggest memory read is the old state, which was finalized way back during the last forward pass.  So we can have causal conv1d launch dependent kernels, which will be state update, and then launch state update with PDL and load the old state, blocking when we start needing x, B, or C.  Due to the loop structure we actually are a little less aggressive, and can't easily load A or dt components before waiting.

But we don't turn this on here yet.  https://github.com/NVIDIA/TensorRT-LLM/pull/11376 is in progress, and will replace the triton selective state update with a FlashInfer implementation, which doesn't yet have PDL support.  So better to enable a cleaner merge.  If it was on, it saves about 0.2% additional time.  In isolation it was more, 0.7%, but multi-stream dt eats into the savings.

## Test Coverage

No new test coverage.

## PR Checklist

## GitHub Bot Help

## Summary by CodeRabbit

  * Enhanced Mamba2 model kernel scheduling with configurable dependent-kernel launches
  * Implemented parallel execution for convolution and data type conversion operations
  * Optimized kernel caching and execution stream management to reduce computational overhead


sunnyqgg (approved):
LGTM

nv-guomingz (approved):

PR #11899, opened 2026-03-04
[TRTLLM-10421][perf] Add fused cat+fp8_quantize CUDA kernel for DSA indexer

kaiyux:
## Summary

Add a fused CUDA kernel (`trtllm::fused_cat_fp8`) that combines BF16 tensor concatenation and FP8 E4M3 1×128 quantization into a single kernel launch. This replaces the sequential `cat → fp8_quantize_1x128` path in the DSA indexer's `_prep_q_or_k()`, reducing per-call kernel launches from 2+ down to 1.

### Changes
- **New CUDA kernel** (`fusedCatFp8.cu/h`): 1 warp (32 threads) per row, 8 rows per block. Vectorized 8-byte `int2` loads for BF16 inputs, 4-byte `uint32_t` stores for FP8 outputs. Per-row FP8 quantization with warp-reduced amax.
- **Template specialization** on `UseUe8m0`: eliminates runtime branching for scale format. UE8M0 path uses IEEE 754 bit manipulation instead of `ceilf(log2f(...))` + `exp2f(...)`. Hardware `rcp.approx.ftz` for fast reciprocal.
- **Non-contiguous input support**: accepts row strides directly, handling `torch.split()` views without a prior `.contiguous()` copy.
- **Torch op registration**: registered via `TORCH_LIBRARY_FRAGMENT` with fake tensor support for `torch.compile`.
- **DSA integration** (`dsa.py`): `_prep_q_or_k()` now calls the single fused op, removing imports for `maybe_compiled_cat` and `fp8_utils`.
- **`rotate_activation` restored** as a standalone utility in `dsa.py` for future use (not called in the current code path).

### Files
| File | Description |
|------|-------------|
| `cpp/tensorrt_llm/kernels/fusedCatFp8.cu` | Fused CUDA kernel implementation |
| `cpp/tensorrt_llm/kernels/fusedCatFp8.h` | Kernel header with API docs |
| `cpp/tensorrt_llm/thop/fusedCatFp8Op.cpp` | Torch op binding (TORCH_LIBRARY_FRAGMENT) |
| `cpp/tensorrt_llm/thop/CMakeLists.txt` | Build integration |
| `tensorrt_llm/_torch/attention_backend/sparse/dsa.py` | Use fused op in `_prep_q_or_k()` |
| `tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py` | Fake tensor registration for compile |
| `tests/unittest/_torch/thop/serial/test_fused_cat_fp8.py` | Comprehensive unit tests |

## Test plan
- [x] Unit test: `tests/unittest/_torch/thop/serial/test_fused_cat_fp8.py` — contiguous, non-contiguous, odd M values, UE8M0 and standard scale formats, stride alignment validation
- [x] Integration: DeepSeek-V3/R1 decode serving with fused path
- [x] CI: `/bot run`

### DIFF
--- a/cpp/tensorrt_llm/kernels/fusedCatFp8.cu
+++ b/cpp/tensorrt_llm/kernels/fusedCatFp8.cu
@@ -194,6 +194,12 @@ void invokeFusedCatFp8(__nv_fp8_e4m3* fp8_out, float* scale_out, __nv_bfloat16 c
         pe_row_stride >= pe_dim, "fusedCatFp8: pe_row_stride (%d) must be >= pe_dim (%d)", pe_row_stride, pe_dim);
     TLLM_CHECK_WITH_INFO(nope_row_stride >= nope_dim, "fusedCatFp8: nope_row_stride (%d) must be >= nope_dim (%d)",
         nope_row_stride, nope_dim);
+    TLLM_CHECK_WITH_INFO(pe_row_stride % ELEMS_PER_THREAD == 0,
+        "fusedCatFp8: pe_row_stride (%d) must be a multiple of %d for aligned vectorized access", pe_row_stride,
+        ELEMS_PER_THREAD);
+    TLLM_CHECK_WITH_INFO(nope_row_stride % ELEMS_PER_THREAD == 0,
+        "fusedCatFp8: nope_row_stride (%d) must be a multiple of %d for aligned vectorized access", nope_row_stride,
+        ELEMS_PER_THREAD);
 
     int num_blocks = (M + ROWS_PER_BLOCK - 1) / ROWS_PER_BLOCK;
     dim3 grid(num_blocks);
--- a/cpp/tensorrt_llm/kernels/fusedCatFp8.cu
+++ b/cpp/tensorrt_llm/kernels/fusedCatFp8.cu
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-#include "fusedCatHadamardFp8.h"
+#include "fusedCatFp8.h"
 #include "tensorrt_llm/common/assert.h"
 #include "tensorrt_llm/common/config.h"
 #include "tensorrt_llm/common/cudaUtils.h"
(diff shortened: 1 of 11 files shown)


kaiyux (commented):
## PR Review: [TRTLLM-10421][perf] Fuse cat+fp8_quantize in DSA indexer _prep_q_or_k

**Overall: Clean, well-structured PR. The kernel design is solid and the test coverage is thorough. A few issues to address.**

### Issues

**1. (Medium) `rotate_activation` removal — is this intentional?**

The old `_prep_q_or_k` called `rotate_activation(q_or_k)` (Hadamard transform) between cat and FP8 quantize. The new fused kernel skips this entirely. The PR description mentions "Eliminates ~366 kernel launches" but doesn't call out that the Hadamard transform is being dropped. The commit history shows a prior commit `d3626594ba [None][refactor] Remove Hadamard transform from fused cat+fp8 kernel` — so this seems intentional and was a separate decision, but:

- Was the Hadamard transform confirmed unnecessary for DSV3.2 correctness? The old code had it as a functional step, not just a perf optimization.
- The `fast_hadamard_transform` import and `rotate_activation` function are removed in this PR — if this is a separate semantic change, it should be in a separate commit/PR for clarity.

**2. (Low) Output always flattened to 2D**

The thop computes `M = pe.numel() / pe_dim` and always returns `[M, head_dim]` shaped output. The old code did `q_or_k.view(-1, self.head_dim)` explicitly before quantization, so this is semantically equivalent. But the 3D test (`test_fused_cat_fp8_3d_input`) confirms the caller must reshape after — this is fine since the call site does `q_fp8.view(-1, self.n_heads, self.head_dim)` at line 1559.

**3. (Low) `fp8_quantize_1x128_sf_transpose` does scale transposition — fused kernel doesn't**

The reference function name includes "sf_transpose" suggesting it does a scale format transposition. Looking at the reference, `fp8_quantize_1x128_sf_transpose` calls `torch.ops.trtllm.fp8_quantize_1x128` then does padding/reshaping of the scale. The fused kernel returns raw `[M, 1]` scales. Verify that downstream consumers (`_weight_scale`, `_update_k_cache`, `sparse_attn_indexer`) expect this format and not the transposed/padded format from the old path.

**4. (Nit) Hardcoded `HEAD_DIM = 128`**

Fine for DSV3.2 but worth a comment in the header doc saying this kernel is purpose-built for the DSA indexer path. Already has the runtime check, so no functional risk.

**5. (Nit) UE8M0 exponent overflow not guarded**

In the UE8M0 path (lines 137-143 of the kernel), if `mantissa != 0`, you add `0x00800000u` to `exp_bits`. If the input contains `inf` or extremely large values, `exp_bits` could overflow the 8-bit exponent field. This is extremely unlikely in practice (inputs are BF16, max ~3.4e38 → amax/448 < 7.6e35, well within float range), but a comment noting the assumption would help.

### Strengths

- **Vectorized I/O**: `int2` loads (8B) for BF16 and `uint32_t` stores (4B) for FP8 are well-matched to the data widths.
- **Branchless PE/NOPE selection**: The ternary on `from_pe` compiles to SELP — avoids warp divergence since all threads in a warp have the same split boundary.
- **`rcp.approx.ftz`**: Good use of hardware approximate reciprocal — FP8 E4M3 only has 3 mantissa bits so ~2^-23 relative error is more than sufficient.
- **Non-contiguous stride support**: Properly handles `torch.split()` views without requiring `.contiguous()` — avoids an extra copy.
- **Thorough test matrix**: Covers contiguous, non-contiguous, 3D, mixed contiguity, zero input, and both UE8M0/standard scale formats.
- **Test tolerance rationale**: Good explanation of why fused vs reference differ (fp32 vs bf16 amax precision) and appropriate tolerance thresholds.

### Summary

The kernel is well-optimized and the integration is clean. The main question is whether dropping `rotate_activation` is validated as correct for DSV3.2 model quality (issue #1), and whether the scale output format is compatible with downstream consumers without the transposition step (issue #3). Everything else looks good to ship.

yuxianq (approved):

lfr-0531 (approved):
LGTM~

hyukn (approved):

PR #11943, opened 2026-03-05
[TRTLLM-12427][perf] Qwen2.5/3/3.5-VL Performance Optimization

yechank-nvidia:
## Summary

This PR reworks the Qwen2-VL / Qwen2.5-VL / Qwen3-VL PyTorch-backend implementations to cut host (CPU) overhead in the vision tower and input processors, enables piecewise CUDA graph for these models, and fixes several correctness issues around mRoPE and vision-block RoPE. The optimizations target the high-concurrency serving regime, where launch overhead and host↔device syncs dominate.

### Vision tower / rotary embedding
  - Drop the HF rotary dependency and memoize the frequency table; pre-compute `cos`/`sin` into init-time buffers so the forward path no longer calls `.cos()`/`.sin()` per step.
  - Add an L2 per-tile GPU rotary cache for Qwen2.5/3-VL vision and annotate its measured GPU footprint.
  - Remove dead code and a redundant batched pos-embed kernel from the vision tower.

### Host-overhead reduction
  - Add an `async_tensor_h2d` helper and route all Qwen2.5/3-VL H2D copies (vision `pos_ids`, `window_index`, `rope_position_ids`) through it.
  - Skip redundant pinning in `maybe_pin_memory` when the input is already pinned.
  - Pre-allocate deepstack scratch and skip vision-encoder host syncs.
  - Add a text-only fast path in the Qwen2.5/3-VL input processors so text-only requests avoid vision-path work.

### CUDA graph
  - Enable piecewise CUDA graph for LLM prefill on Qwen2/3-VL.
 
### Refactor
  - Inherit the Qwen3-VL input processor from the Qwen2-VL base to remove duplication.

## Performance

  Model: `Qwen3VLForConditionalGeneration` (FP8 weights + KV cache, bf16 vision encoder). Hardware: H200 ×1. Workload: image+text, ISL=1000, OSL=1000, 512×512 image, KV block reuse off, 3-run mean. Comparison is against upstream `main` with identical serving config (`max_batch_size=256`, `max_num_tokens=8192`, `num_postprocess_workers=4`, `cuda_graph_config.enable_padding=true`, chunked prefill on).

### System output-token throughput (tok/s)

  | concurrency | upstream | this PR | Δ |
  |---|---|---|---|
  | 32  | 4735 | 4838 | +2.2% |
  | 64  | 6948 | 7359 | +5.9% |
  | 128 | 8553 | 9767 | **+14.2%** |

### Other metrics

  - **Per-user throughput (tok/s/user)** — c=128: 77.5 → 89.8 (**+15.9%**)
  - **TTFT (ms)** — c=1: 147 → 81 (**−45%**); c=32: 815 → 690 (−15%); c=128: 1953 → 1804 (−8%)
  - **ITL (ms/token)** — c=128: 12.96 → 11.18 (**−13.7%**)
  - **Request latency (ms)** — c=128: 14901 → 12975 (**−12.9%**)

Gains are concentrated at high concurrency, where the host-side savings (fewer CPU launches, async H2D, higher CUDA-graph hit ratio) and the lower-overhead mRoPE path matter most. Low-concurrency TTFT also improves substantially (c=1 nearly halved) from the text fast path and removed vision-encoder syncs.

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Fixed rotary cache index computation in attention kernels.
  * Corrected request broadcasting logic for distributed inference configurations.

* **Performance**
  * Optimized multimodal model inference through vectorized RoPE position indexing and GPU memory operations.
  * Added Triton kernel support for position embedding interpolation in vision models.
  * Improved device transfer efficiency with async host-to-device operations and optimized tensor placement.
  * Enhanced distributed tensor parallelism handling.

* **Tests**
  * Expanded test coverage for multimodal RoPE configurations and vision component equivalence validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/11943?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tensorrt_llm/_torch/pyexecutor/request_utils.py
```diff
@@ -527,7 +527,10 @@ def _broadcast_requests(
             return payloads
 
         if not self.dist.has_pp:
-            return self.dist.broadcast(payloads, root=0)
+            if self.dist.has_tp:
+                return self.dist.broadcast(payloads, root=0)
+            else:
+                return payloads
```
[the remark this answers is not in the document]
yechank-nvidia: TODO: Need to remove this.

cpp/tensorrt_llm/kernels/unfusedAttentionKernels/unfusedAttentionKernels_2_template.h
```diff
@@ -903,9 +903,8 @@ __global__ void applyBiasRopeUpdateKVCacheV2(QKVPreprocessingParams<T, KVCacheBu
         [[maybe_unused]] float2 const* rotary_coef_cache_buffer = nullptr;
         if (params.mrope_rotary_cos_sin != nullptr)
         {
-            rotary_coef_cache_buffer = params.mrope_rotary_cos_sin
-                + batch_idx * params.rotary_embedding_max_positions * params.half_rotary_dim
-                + static_cast<size_t>(rotary_position) * params.half_rotary_dim;
+            rotary_coef_cache_buffer
```
2ez4bz, line 915: @coderabbitai Why was this change safe? The math on paper is not the same.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -50,6 +51,18 @@
...
+
+# Vision RoPE on Qwen2-VL/2.5-VL/3-VL uses head_dim 72/80 which doesn't satisfy
+# FlashInfer's "head_size % 64 == 0" precondition. Fall back to flash_attn's
+# fused Triton rotary kernel: one launch instead of the ~7 elementwise launches
+# of the PyTorch path (mul x4, add/sub x2, cat x1).
+try:
+    from flash_attn.ops.triton.rotary import \
+        apply_rotary as _flash_attn_apply_rotary  # type: ignore
+    _FLASH_ATTN_ROTARY_AVAILABLE = True
+except ImportError:  # pragma: no cover - flash_attn is part of the default deps
```
2ez4bz: What does this comment mean? Do we need this import guard if `flash_attn` is part of the default deps?

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -360,6 +375,23 @@ def __call__(
     ) -> Tuple[List[int], Optional[ExtraProcessedInputs]]:
         text_prompt, mm_data, mm_processor_kwargs = inputs.get("prompt"), \
                         inputs.get("multi_modal_data", {}), inputs.get("mm_processor_kwargs", {})
+
+        # Text-only fast path: skip the multi-modal HF processor and call
+        # the tokenizer directly (their outputs match bit-exactly when
+        # ``images`` / ``videos`` are ``None``). mrope_config still needs
```
2ez4bz: Nit:

```
sed -i 's/``/`/g'
```

on the changed lines.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -645,13 +677,39 @@ def apply_rope(self,
                     err,
                 )
 
-        cos = cos.to(dtype=q.dtype)
-        sin = sin.to(dtype=q.dtype)
+        # cos/sin are pre-cast to ``q.dtype`` upstream (``get_rope_and_window_
+        # index_by_thw`` for Qwen2.5-VL; ``rope_cos_cache`` buffer dtype for
+        # Qwen3-VL). When that holds, ``.to(dtype=q.dtype)`` is a no-op; keep
+        # it as a safety net for callers/quantization paths that didn't pre-cast.
```
2ez4bz: I don't get it - if `.to(dtype=q.dtype)` is a no-op when the types already match, do we still need the `if x.dtype != y.dtype` style checks below?

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -475,9 +477,9 @@ def get_prompt_token_ids(
 
 class Qwen3VLVisionAttention(Qwen2_5_VLVisionAttention):
     def __init__(self, model_config, layer_idx):
-        # Qwen3-VL keeps `torch_dtype` only on `text_config` under transformers 5.x
```
2ez4bz, line 527: Nit: can we keep the original comment? Or was it outdated?

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -132,109 +134,95 @@ def get_rope_index(
...
-            total_input_ids = input_ids
-            if attention_mask is None:
-                attention_mask = torch.ones_like(total_input_ids)
-            position_ids = torch.ones(
-                3,
-                input_ids.shape[0],
-                input_ids.shape[1],
-                dtype=input_ids.dtype,
-                device=input_ids.device,
+            # numpy-based vectorized impl: ~2-3× faster than the torch loop
```
2ez4bz: There seems to be a lot of overlap between what's here and in the Qwen2 file. Is there any way the common parts could be extracted and shared?

Or do the different video + text token handling schemes make it impractical?

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -618,63 +705,42 @@ def pos_embed_interpolate_native(
...
-    h_ceil = torch.clamp(h_floor + 1, max=num_grid_per_side - 1)
-    w_ceil = torch.clamp(w_floor + 1, max=num_grid_per_side - 1)
-
-    dh = h_idxs - h_floor
-    dw = w_idxs - w_floor
-
-    dh_grid, dw_grid = torch.meshgrid(dh, dw, indexing="ij")
-    h_floor_grid, w_floor_grid = torch.meshgrid(h_floor, w_floor, indexing="ij")
-    h_ceil_grid, w_ceil_grid = torch.meshgrid(h_ceil, w_ceil, indexing="ij")
+    h_scale = float(num_grid_per_side - 1) / float(h - 1) if h > 1 else 0.0
```
2ez4bz: Out of curiosity, why do we need the `float` conversions? Doesn't `/` return `float`?

2ez4bz (approved):

tensorrt_llm/_torch/models/modeling_qwen3vl.py
yechank-nvidia: Removed. Thx.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
yechank-nvidia: Refactored into Qwen3-VL inherits Qwen-2VL's inputprocessor. Thx for suggestion.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
yechank-nvidia: Reverted.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: I think claude was too cautious about this. Removed.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Done.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Found that our docker already install flash_attn natively, so I think we don't need guard on this. Removed.

tensorrt_llm/_torch/pyexecutor/request_utils.py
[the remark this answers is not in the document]
yechank-nvidia: Removed. Can you check?

eopXD (approved):
Changes under `tensorrt_llm/_torch/pyexecutor/resource_manager.py` looks good to me. They look like drive-by fixes though, I don't think they directly contribute to what this MR is trying to achieve.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Turns out we need this. The SBSA build does not have flash_attn.

xxi-nv (approved):

cpp/tensorrt_llm/common/attentionOp.cpp
```diff
@@ -1638,6 +1638,11 @@ int AttentionOp::enqueueContext(EnqueueContextParams<T> const& params, cudaStrea
     invokeBuildDecoderInfo(decoder_params, stream);
     sync_check_cuda_error(stream);
 
+    int32_t const* contextCuQSeqlens = params.cu_q_seqlens != nullptr ? params.cu_q_seqlens : workspaceViews.cuQSeqlens;
```
PerkzZheng: should we skip computing cuSeqlens in invokeBuildDecoderInfo if provided ?

yechank-nvidia: Yes, we can skip recomputing cuSeqlens if provided. Let me add that part.

hyukn (approved):

cpp/tensorrt_llm/common/attentionOp.cpp
yechank-nvidia: Added.

PerkzZheng (approved):

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -459,6 +460,19 @@ def __init__(
                         self.model.model,
                         backend=self._torch_compile_backend,
                         fullgraph=torch_compile_fullgraph)
+                elif hasattr(self.model, "llm") and isinstance(
+                        getattr(self.model.llm, "model", None),
+                        torch.nn.Module):
```
QiJune: I’m concerned about making this an engine-level heuristic based on `.llm.model`.

This branch does not only match Qwen2/3-VL. It matches any multimodal wrapper that happens to expose an inner causal LM as `self.llm.model`. Several multimodal models in this repo have that shape, but they do not all satisfy the same assumptions as Qwen VL.

There are two concrete risks here:
1. The compiled decoder attention custom ops look up layer metadata through the root model attrs populated by `model_engine.model_forward()`. Qwen explicitly wires the inner LLM `extra_attrs` to the outer model config, but many other multimodal wrappers build the inner LLM from a copied `ModelConfig` and do not appear to propagate the inner `attn_layers` / `mla_layers` back to the root model. For those models, compiling `self.model.llm.model` can make the compiled decoder fail to find its registered attention layers.

2. `set_torch_compiling(True)` is global. The outer multimodal encoder still runs eager while that flag is enabled. Qwen’s vision attention has a specific unregister guard to avoid taking the compile-only TRT-LLM attention custom-op path, but other multimodal vision towers that use TRT-LLM `Attention` do not necessarily have the same guard.

I think the safer design is to make decoder-only compile an explicit model opt-in hook rather than inferring it from `.llm.model`. For example, the engine could check for `self.model.apply_torch_compile(...)`; Qwen VL can implement that hook and compile only `self.llm.model`, while models without the hook keep the existing generic behavior. This lets each multimodal model own the required setup, such as sharing `extra_attrs`, unregistering vision attention when needed, and preserving any outer references to `self.llm.model.embed_tokens`.

yechank-nvidia: Thanks, that makes sense. I removed the engine-level `.llm.model` heuristic and changed this to an explicit model opt-in hook.

Qwen VL now implements `apply_llm_torch_compile()`, so only the inner decoder is compiled while the outer multimodal/vision path remains eager. This keeps the Qwen-specific setup owned by the model instead of applying the assumption to every wrapper that happens to expose `self.llm.model`.

I also added TODOs to move this contract to `MultimodalModelMixin` once the other multimodal models consistently inherit from it.

QiJune (approved):
LGTM

2ez4bz:
first try

PR #11960, opened 2026-03-05
[https://nvbugs/6115560][fix] catch OSError in config_file_lock for NFS compatibility

sara4dev:
## Summary

`config_file_lock()` in `tensorrt_llm/_torch/model_config.py` crashes when `HF_MODULES_CACHE` resides on an NFS-mounted filesystem. On NFS, `filelock` operations can raise `OSError` with errno `ENOLCK` (No locks available) or `ESTALE` (Stale file handle) instead of `PermissionError`. Since the current exception handler only catches `PermissionError` and `filelock.Timeout`, these NFS-specific errors bypass the existing tempdir fallback and crash the process.

This is particularly impactful in multi-node GPU clusters where a shared NFS cache is standard practice — every pod that tries to load a model config concurrently hits this crash.

## Changes

Add `OSError` to both exception handlers in `config_file_lock()`:

1. **Primary lock attempt** (line 51): `except (PermissionError, filelock.Timeout)` → `except (PermissionError, OSError, filelock.Timeout)` — triggers the tempdir fallback for NFS errors
2. **Tempdir fallback** (line 66): `except (PermissionError)` → `except (PermissionError, OSError)` — handles the unlikely case where tempdir also fails

Since `PermissionError` is a subclass of `OSError`, catching `OSError` technically covers both, but keeping `PermissionError` explicit preserves the original intent and readability.

## Root Cause

NFSv3 uses the Network Lock Manager (NLM) protocol for file locking, which is unreliable for cross-node `flock()`/`fcntl()` operations. When `filelock.FileLock` attempts to acquire a lock on an NFS path:

- Lock acquisition can fail with `OSError: [Errno 37] No locks available` (ENOLCK)
- Lock release can fail with `OSError: [Errno 116] Stale file handle` (ESTALE)

The existing fallback to `/tmp` (local ephemeral storage) is the correct behavior for this case — it just wasn't being triggered.

## Reproduction

```python
# On any NFS-mounted path, run from multiple nodes concurrently:
import filelock
lock = filelock.FileLock("/nfs-mount/test.lock")
with lock:  # Raises OSError: [Errno 37] No locks available
    pass
```

## Workaround (for users on affected versions)

Set `HF_MODULES_CACHE=/tmp/hf_modules` as an environment variable to redirect the lock file to local storage.

Fixes #11958

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced file locking behavior to catch additional OS and permission-related errors. Added temporary directory-based fallback locking with timeout handling.

### DIFF
--- a/tensorrt_llm/_torch/model_config.py
+++ b/tensorrt_llm/_torch/model_config.py
@@ -86,8 +86,10 @@ def config_file_lock(timeout: int = 10):
     try:
         with lock:
             yield
-    except (PermissionError, filelock.Timeout):
-        # Fallback to tempdir
+    except (PermissionError, OSError, filelock.Timeout):
+        # Fallback to tempdir when primary lock path is unusable (e.g.,
+        # NFS locking failures like ENOLCK/ESTALE, permission issues,
+        # or lock acquisition timeouts)
         tmp_dir = Path(tempfile.gettempdir())
         tmp_dir.mkdir(parents=True, exist_ok=True)
         tmp_lock_path = tmp_dir / "_remote_code.lock"
@@ -101,7 +103,7 @@ def config_file_lock(timeout: int = 10):
             )
             # proceed without lock
             yield
-        except (PermissionError) as e:
+        except (PermissionError, OSError) as e:
             logger.warning(
                 f"tempdir config lock unavailable due to OS/permission issue: {e}, proceeding without lock"
             )
--- a/tensorrt_llm/_torch/model_config.py
+++ b/tensorrt_llm/_torch/model_config.py
@@ -1,4 +1,5 @@
 import contextlib
+import errno
 import json
 import os
 import tempfile
@@ -86,10 +87,14 @@ def config_file_lock(timeout: int = 10):
     try:
         with lock:
             yield
-    except (PermissionError, OSError, filelock.Timeout):
+    except (PermissionError, OSError, filelock.Timeout) as e:
         # Fallback to tempdir when primary lock path is unusable (e.g.,
         # NFS locking failures like ENOLCK/ESTALE, permission issues,
         # or lock acquisition timeouts)
+        if isinstance(e,
+                      OSError) and e.errno not in (errno.EACCES, errno.EPERM,
+                                                   errno.ENOLCK, errno.ESTALE):
+            raise
         tmp_dir = Path(tempfile.gettempdir())
         tmp_dir.mkdir(parents=True, exist_ok=True)
         tmp_lock_path = tmp_dir / "_remote_code.lock"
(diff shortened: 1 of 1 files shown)


chang-l (approved):

chienchunhung (approved):
The PR looks good to me. PS: We might need to clean up something in the follow-up PR.

chang-l:
@sara4dev @pengbowang-nv — fresh CI ([L0_MergeRequest_PR #39025](https://prod.blsm.nvidia.com/sw-tensorrt-top-1/job/LLM/job/main/job/L0_MergeRequest_PR/39025/)) confirms `OSError` is too broad in its current placement. 29 failures in `unittest._torch.executor.test_pytorch_model_engine`, all with `RuntimeError: generator didn't stop after throw()`.

Representative stack trace (e.g. `test_warmup`):

```
huggingface_hub.errors.RepositoryNotFoundError: 404 Client Error for url
    https://huggingface.co/dummy/resolve/main/config.json
tensorrt_llm/_torch/model_config.py:79 in config_file_lock
    yield        ← second yield (tempdir-fallback path)
```

The test calls `from_pretrained("dummy")` to exercise the 404 path. HF raises `RepositoryNotFoundError` (a subclass of `OSError`) *inside* the `with config_file_lock():` body. The new broader catch swallows that, enters the tempdir fallback, and `yield`s a second time — violating `@contextmanager`'s `throw()` semantics.

Two minimal fix options, both keep the bug-fix intent for ENOLCK/ESTALE:

**A. Narrow to specific errnos** (matches the original review suggestion):
```python
except OSError as e:
    if e.errno not in {errno.ENOLCK, errno.ESTALE, errno.EACCES, errno.EPERM}:
        raise
    # ... tempdir fallback
```

**B. Scope the broad catch to `lock.acquire()` only, not the yielded body:**
```python
try:
    lock.acquire(timeout=timeout)
except (PermissionError, OSError, filelock.Timeout):
    # tempdir fallback
    ...
else:
    try:
        yield
    finally:
        lock.release()
```

I think either resolves the failing tests – @sara4dev can you please take a look?

chang-l:
Applied **Option A** from [earlier comment](https://github.com/NVIDIA/TensorRT-LLM/pull/11960#issuecomment-4500402045) since author was unresponsive past Friday EOB. New commit narrows the `OSError` catch to errno set {EACCES, EPERM, ENOLCK, ESTALE} so non-lock OSErrors (e.g. HF `RepositoryNotFoundError`) propagate out of the lock body cleanly. Triggering fresh CI.

/bot run

chang-l:
Rebased with sign-offs to satisfy DCO (force-pushed; content unchanged — verified tree hash equals pre-rebase). New HEAD: `4d86ed2089`. Triggering fresh CI.

/bot run

yufeiwu-nv:
hi @chang-l , option-A lead to deepseek_r1_0528 error in GB300.  I create a pr to apply option-B. Do you think it's OK?
`Error: filelock._error.Timeout: The file lock '/root/.cache/huggingface/modules/_remote_code.lock' could not be acquired. [executor][RANK 0] Failed to initialize executor on rank 0: The file lock '/root/.cache/huggingface/modules/_remote_code.lock' could not be acquired. [F] [llmapi] MpiSession shutdown timeout, aborting... perf/test_perf.py::test_perf[deepseek_r1_0528_fp4-bench-pytorch-float4-maxbs:1000-maxnt:5000-kv_frac:0.85-input_output_len:5000,500-reqs:2000-ep:4-tp:4-gpus:4]`
#15213

PR #11987, opened 2026-03-06
[None][infra] Reduce Docker image layer count in release stage

janbernloehr:
## Summary

Follow-up to #11720 which already consolidated the devel stage layers using bind mounts.

This PR applies the same pattern to the **release stage**:

- Replace 7 `COPY` layers + 2 `RUN` layers with a single `RUN` using `--mount=type=bind` and `cp`, consolidating all file copies, symlink creation, and cleanup into one layer
- Replace the OSS attribution `COPY` + `RUN` + `rm` with a single `RUN` using a bind mount

Reduces the release stage from ~13 layers to 3 (`pip install`, copy+setup+cleanup, OSS attribution), saving ~10 layers in the published image and leaving more headroom under Docker's 128-layer overlay2 limit for downstream consumers.

## Test plan

- [x] Build release image and verify layer count reduction (`docker history`)
- [x] Verify `trtllm-serve` starts correctly from built image
- [x] Verify symlinks to bin/lib are created correctly
- [x] Verify OSS attribution file is generated in release image

### DIFF
--- a/docker/Dockerfile.multi
+++ b/docker/Dockerfile.multi
@@ -120,35 +120,41 @@ WORKDIR /app/tensorrt_llm
 RUN --mount=type=cache,target=/root/.cache/pip --mount=type=bind,from=wheel,source=/src/tensorrt_llm/build,target=/tmp/wheel \
     pip install /tmp/wheel/tensorrt_llm*.whl
 
-COPY README.md ./
-COPY --from=wheel /src/tensorrt_llm/build/tensorrt_llm*.whl ./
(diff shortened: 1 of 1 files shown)


dpitman-nvda (approved):
LGTM

MartinMarciniszyn (changes requested):
We need to run the CI with the modified docker images before merging this.

tburt-nv:
superseded by https://github.com/NVIDIA/TensorRT-LLM/pull/14972

Issue #11694, opened 2026-02-24
[Feature]: AutoDeploy - Piecewise CG peak memory improvement


nvchenghaoz:
### 🚀 The feature, motivation and pitch

Currently the Piecewise CG implementation is around 1.5x - 2x comparing to the eager run. This leads to some limitation for kv cache or mamba cache since the user might need to lower the free_gpu_memory_fraction during the usage.. 

The goal for this task is to do the peak memory analysis / reduce the memory usage for piecewise CG. 

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and checked the [documentation](https://nvidia.github.io/TensorRT-LLM/) and [examples](https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples) for answers to frequently asked questions.

PR #11993, opened 2026-03-06
[#11694][feat] AutoDeploy: Improve the piecewise CG memory usage

nvchenghaoz:
This PR is to improve the memory usage during the piecewise Cudagraph capturing. 

Several main updates:
1. Use the output buffer as a parameter for the attention operators so that during the piecewise cudagraph, there is no eager attention output allocation. Another advantage for this is the output tensor can be managed by the memory pool, which will hugely improve the memory usage. 
2. Update the piecewise cudagraph control code for the output buffer implementation. And wire the input with the output between dynamic (the ops that cannot be CG captured) and static (can be CG captured) regions. 
3. Some bug fixes and the updates. 

With the change, the memory usage of the piecewise cudagraph is under control and can be used for most cases. 

Also the perf improves a bit due to the changes - 
For Qwen 3.5 isl 1k osl 1k, H100 - 
Screenshot 2026-03-09 at 11 44 13 AM

The next step for piecewise CG is, the final gemm for the prefill is really big and AutoDeploy should enable the gather_logits transform by default. This region is marked as dynamic due to the unknown shape of the gemm. This will be addressed by a follow up task. My plan to fix this is to pad the final gemm to the max_batch_size. That would need some further investigation. #12049 to track the work. 

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added optional output buffer support across custom operations for more efficient memory usage.
  * Implemented dynamic submodule classification utilities for improved graph compilation.
  * Added batch capacity-based pruning for piecewise compilation buckets.

* **Bug Fixes**
  * Reduced default free GPU memory fraction to prevent OOM errors.

  * Enabled piecewise compilation by default for faster inference.
  * Implemented three-phase warmup/capture/replay pipeline with weak reference memory recycling.


taylor-yb-lee (approved):

nvchenghaoz:

PR #12060, opened 2026-03-10
[TRTLLM-11037][bug] Fix MoE DeepEP hang caused by non-deterministic GC

xxi-nv:
## Summary
- Fix MoE DeepEP hang caused by non-deterministic garbage collection timing across ranks
- DeepEP `Buffer.__del__` calls `intranode::barrier` (a collective op). Without explicit synchronous release, GC timing differences across ranks cause some ranks to block in the barrier indefinitely
- Add `destroy()` method to `Communication` base class and `DeepEP`/`DeepEPLowLatency` subclasses for explicit buffer release
- Call `destroy()` in `ConfigurableMoE` when falling back from DeepEP to AllGatherReduceScatter
- Make `ConfigurableMoE` a context manager so resources are released on scope exit
- Re-enable DEEPEPLOWLATENCY multi-GPU tests on H100 that were previously skipped due to this hang

## Test plan
- [x] Unit tests updated to use context manager pattern (`with create_moe(...) as fused_moe`)
- [ ] Multi-GPU MoE tests with DEEPEPLOWLATENCY on H100 (previously hanging, now expected to pass)
- [ ] CI validation

## Summary by CodeRabbit

* **New Features**
  * MoE communication components now support context manager syntax (with statement) for streamlined resource management.
  * Added explicit lifecycle control methods to communication layer components.

* **Tests**
  * Enhanced test coverage for communication subsystem reliability and low-latency scenarios.

### DIFF
--- a/tensorrt_llm/_torch/modules/fused_moe/communication/base.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/communication/base.py
@@ -186,6 +186,10 @@ class Communication(ABC):
         """
         raise NotImplementedError
 
+    def destroy(self):
+        """Synchronously release resources. Must be called on ALL ranks
+        before the object is discarded."""
+
     def get_eplb_gathered_statistics(self) -> Optional[torch.Tensor]:
         """
         Return gathered EPLB statistics from the last dispatch, if available.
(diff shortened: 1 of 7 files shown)


leslie-fang25 (approved):

QiJune (approved):
LGTM

PR #12062, opened 2026-03-10
[TRTLLM-11540][feat] Add EAGLE3 dynamic tree speculative decoding support

sunnyqgg:
## Summary
- Add dynamic tree speculative decoding support for EAGLE3 (both one-model and two-model flows)
- Implement `Eagle3OneModelDynamicTreeWorker` and `Eagle3OneModelDynamicTreeSampler` for one-model dynamic tree inference
- Add CUDA kernels for dynamic tree operations (expand, gather, update)
- Support growing context in dynamic tree mode for improved accept rates

## Changes
- **New**: `tensorrt_llm/_torch/speculative/eagle3_dynamic_tree.py` — One-model dynamic tree worker and sampler
- **New**: `tensorrt_llm/_torch/speculative/dynamic_tree_ops.py` — Python wrappers for dynamic tree CUDA ops
- **New**: `cpp/tensorrt_llm/kernels/speculativeDecoding/dynamicTreeKernels.cu/.h` — CUDA kernels
- **New**: `cpp/tensorrt_llm/thop/dynamicTreeOp.cpp` — Torch custom op bindings
- **Modified**: `tensorrt_llm/_torch/speculative/eagle3.py` — Refactored `Eagle3OneModelWorker` with dispatch pattern for linear vs dynamic tree
- **Modified**: `tensorrt_llm/_torch/speculative/utils.py` — Route to dynamic tree components when `use_dynamic_tree=True`
- **Modified**: `tensorrt_llm/_torch/speculative/drafting_loops.py` — Two-model dynamic tree drafting loop
- **Modified**: `tensorrt_llm/_torch/speculative/model_drafter.py` — Dynamic tree spec tree manager integration
- **Modified**: `tensorrt_llm/_torch/speculative/spec_tree_manager.py` — Support dynamic tree token organization
- **Modified**: `tensorrt_llm/_torch/pyexecutor/model_engine.py` — Dynamic tree detection for target model
- **Modified**: `tensorrt_llm/_torch/pyexecutor/sampler.py` — Dynamic tree batch verification
- **Modified**: `tensorrt_llm/_torch/models/modeling_speculative.py` — Hidden states handling for dynamic tree
- **Modified**: `tensorrt_llm/llmapi/llm_args.py` — Configuration validation for dynamic tree
- **Modified**: Attention backend files for dynamic tree metadata support

## Test plan
- [x] Unit tests pass (`tests/unittest/_torch/speculative/`)
- [x] One-model dynamic tree EAGLE3 inference matches two-model accept rates
- [x] Two-model dynamic tree EAGLE3 inference works correctly
- [x] No regression in standard (non-dynamic-tree) EAGLE3 flows

## Summary by CodeRabbit

* **New Features**
  * Added dynamic tree speculative decoding support for Eagle3, enabling efficient parallel token prediction with flexible tree structures
  * Added streaming generation support with configurable max draft tokens and speculative decoding parameters
  * Enhanced token verification with CUDA-accelerated kernels for improved inference performance

* **Configuration**
  * Added `--max_total_draft_tokens` parameter for controlling total draft token budget
  * Added `--streaming` flag for real-time token streaming output


mikeiovine (approved):

laikhtewari (dismissed):
There should be documentation on how to use this in the speculative decoding feature page

The feature combination matrix should also be updated

eopXD (commented):
1. Regarding `_relocate_kv_eagerly`, I think we should still keep KV cache management workflow concentrated under resource_manager.py. This should be the same goal as `_update_kv_cache_draft_token_location` as you have removed, but for 2D KV which you have implemented. Is it possible for you to add new methods under `resource_manager.py` and update how we deal with draft token locations under `update_resources`? We also need to make sure that all existing cases are covered with the removal of `_update_kv_cache_draft_token_location`.

2. Under `_update_kv_cache_draft_token_location`, thank you for the fix of specifying the correct number of kv_heads for TP. The fix has an implicit assumption that TP shards are uniform. We can resolve this by adding a function that returns the correct TP number. This gives us the flexibility and avoids future bug.

3. We should add unit test for the 2D KV relocation kernel.

venkywonka (approved):

eopXD (approved):
I see the tension between kv cache control / data plane manipulation inside the spec-dec.

The MR looks good to me.

A non-blocking sugggestion from me is that the method can still be under resource_manager. We can have something like:

```
  class KVCacheManager:
      # Control plane (called by executor hooks)
      def prepare_resources(self, ...): ...   # alloc blocks
      def update_resources(self, ...): ...    # rewind + store for reuse

      # Data plane (called by worker at GPU-appropriate time)
      def relocate_accepted_kv_2d(
          self,
          accepted_indices_2d: torch.Tensor,   # [batch, max_draft_len], GPU
          num_accepted: torch.Tensor,           # [batch], GPU
          kv_lens: torch.Tensor,                # [batch], GPU
          block_offsets: torch.Tensor,          # from attn_metadata
      ) -> None:
          """Compact accepted draft KV entries to contiguous positions.

          Must be called on the GPU execution path (CUDA-graph-safe).
          All inputs are GPU tensors — no GPU→CPU sync.
          """
```

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -969,6 +968,26 @@ def validate_eagle_config(self) -> 'EagleDecodingConfig':
...
+                logger.warning(
+                    f"max_total_draft_tokens is not provided, use the default value {default_max_total_draft_tokens} (default_max_total_draft_tokens = dynamic_tree_max_topK * max_draft_len)"
+                )
+            else:
+                assert self.max_total_draft_tokens <= total_history_draft_tokens and self.max_total_draft_tokens >= default_max_total_draft_tokens, f"max_total_draft_tokens should be between {default_max_total_draft_tokens} and {total_history_draft_tokens}"
+
+        # Linear tree
+        if self.max_total_draft_tokens is None:
+            self.max_total_draft_tokens = self.max_draft_len
+
```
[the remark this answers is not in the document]
sunnyqgg: Done

PerkzZheng (approved):

tests/unittest/_torch/speculative/test_eagle3.py
```diff
@@ -815,5 +816,95 @@ def test_eagle3_cdl_sampling(disable_overlap_scheduler: bool):
     llm_spec.shutdown()
 
 
+@pytest.mark.parametrize("disable_overlap_scheduler", [False, True])
+@pytest.mark.parametrize("use_cuda_graph", [False, True])
+@pytest.mark.high_cuda_memory
+@skip_blackwell
```
jhaotingc, line 822: Hi @sunnyqgg is this PR only for pre-blackwell? thanks a lot!!

sunnyqgg: Yes, another PR: https://github.com/NVIDIA/TensorRT-LLM/pull/12958 will enable it for blackwell

PR #12117, opened 2026-03-11
[#12116][fix] prevent KVBM hang when requests abort during KV cache transfer

zyang-Modular:
## Release Notes

* **New Features**
  * Added request abort capability to improve resource management and cleanup.

* **Bug Fixes**
  * Enhanced error handling to prevent execution stalls during worker operations.
  * Improved request cancellation with proper system resource cleanup.

* **Tests**
  * Added validation tests for request cleanup and error recovery scenarios.

## Description
This PR addresses the root cause of #12116. While it ca be prevented by upstream (fixed by this [PR](https://github.com/ai-dynamo/dynamo/pull/6635)), the engine should not hang if a request fails during an active KV transfer, regardless of the failure reason. This change improves robustness by ensuring failed requests are always cleaned up before worker.get_finished() is called, avoiding the collective stall entirely.

When a request fails (e.g. prompt exceeds `max_seq_len`, GPU OOM, or a KVBM Rust panic), its in-flight async KV transfer state is not cleaned up.
On the next iteration, `worker.get_finished()` on rank 0 blocks indefinitely waiting for a transfer that will never complete. This prevents rank 0 from reaching the `mpi_allgather` barrier, causing a collective stall across all TP ranks. The `HangDetector` fires after 300s and shuts down all ranks.

## Test Coverage

I put 2 new test cases

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/kv_cache_connector.py
+++ b/tensorrt_llm/_torch/pyexecutor/kv_cache_connector.py
@@ -519,6 +519,23 @@ class KvCacheConnectorManager(KvCacheConnectorManagerCpp):
 
         return saving_async
 
+    def abort_request(self, req: LlmRequest) -> None:
+        """Remove a request from all async tracking structures.
+
+        Must be called when a request is cancelled or fails to prevent
+        worker.get_finished() from blocking on transfers that will never complete.
+
+        Args:
+            req: The request to abort.
+        """
+        req_id = req.request_id
+        for store in (self.new_async_requests, self.pending_async_requests,
+                      self.local_finished_async_requests):
+            store.loading.pop(req_id, None)
+            store.saving.pop(req_id, None)
+        self.scheduler_output_manager.requests.pop(req_id, None)
+        self.scheduler_output_manager.external_loads.pop(req_id, None)
+
     def get_finished(self) -> List[LlmRequest]:
         """
         Process requests that have finished loading and saving.
@@ -533,9 +550,18 @@ class KvCacheConnectorManager(KvCacheConnectorManagerCpp):
         self.pending_async_requests.add_from(self.new_async_requests)
 
         # Pass these newly finished requests into get_finished, and get the list of requests that have finished saving and loading.
-        (finished_saving,
-         finished_loading) = self.worker.get_finished(finished_gen_req_ids,
-                                                      started_loading_req_ids)
+        # Wrap in try/except so that a worker failure still reaches the mpi_allgather barrier,
+        # preventing a collective stall where other ranks wait indefinitely for this rank.
+        try:
+            (finished_saving, finished_loading) = self.worker.get_finished(
+                finished_gen_req_ids, started_loading_req_ids)
+        except Exception as e:
+            logger.error(
+                f"KV connector worker get_finished() raised an exception: {e}. "
+                "Reporting no completions this iteration to unblock mpi_allgather."
+            )
+            finished_saving = []
+            finished_loading = []
 
         # Remove the requests from our pending list that have finished locally.
         new_local_finished_async_requests = self.pending_async_requests.extract_by_id(
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -3175,6 +3175,8 @@ class PyExecutor:
             ]
         self._enqueue_responses(list(error_responses.items()))
         for request in failed_requests:
+            if self.kv_connector_manager is not None:
+                self.kv_connector_manager.abort_request(request)
             self._terminate_request(request)
 
     def _terminate_request(self, request: LlmRequest):
@@ -3202,6 +3204,10 @@ class PyExecutor:
         Returns:
             bool: True if the request can be canceled (either successfully cancelled or doesn't need cancellation).
         """
+        if self.kv_connector_manager is not None:
+            self.kv_connector_manager.abort_request(request)
+            return True
+
         if self.kv_cache_transceiver is None:
             return True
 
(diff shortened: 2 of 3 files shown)


jthomson04 (changes requested):
Discussed this offline with @zyang-Modular. Will hold off on this MR for now.

bGOATnote:
I reviewed the reproduction in #12116 and the current approach in #12117. My read is that the current MR is directionally useful, but it fixes this too close to the `worker.get_finished()` call site rather than at the KV transfer lifecycle itself.

The root cause appears to be:

- A request fails while an async KV transfer is still in flight
- Transfer bookkeeping remains live
- `worker.get_finished()` later waits on a completion that will never arrive
- Rank 0 never reaches `mpi_allgather`
- All TP ranks stall until HangDetector kills the job

That means the real invariant violation is not that `get_finished()` handled one exception poorly. It is:

**An async KV transfer can outlive its request without being driven to a terminal state.**

Because of that, I think the safer fix is to make the KV transfer lifecycle explicitly abort-safe.

## Proposed direction

Instead of relying on cleanup "before `worker.get_finished()` is called," enforce a transfer-state invariant in the KV connector / transfer bookkeeping layer:

Every transfer must eventually reach exactly one terminal state:

- `COMPLETED`
- `FAILED`
- `ABORTED`

No transfer should be allowed to remain indefinitely `IN_FLIGHT` once its request has failed, been cancelled, or otherwise stopped making progress.

## Why this matters

The current reproduction uses prompt length overflow, but the same orphaned-transfer pattern can happen with any abort source:

- Prompt exceeds `max_seq_len`
- GPU OOM
- Runtime cancellation
- Connector-side exception
- KVBM-side failure or panic

So this is a lifecycle bug, not a single validation-path bug.

## Concrete implementation suggestion

1. Track transfer state explicitly in connector bookkeeping.
2. Add one idempotent "finalize transfer" helper that:
   - Marks the transfer terminal
   - Wakes any waiter
   - Removes stale bookkeeping
3. Call that helper from every abort/failure path, not just the current repro path.
4. Harden `worker.get_finished()` so it never waits forever on a transfer whose request is already gone.

The fix should live at the transfer-record layer, not only in the iteration loop around `get_finished()`.

### Pseudocode sketch

```python
def finalize_transfer(request_id, terminal_state, error=None):
    rec = transfer_records.get(request_id)
    if rec is None:
        return

    if rec.state != IN_FLIGHT:
        return  # idempotent

    rec.state = terminal_state
    rec.error = error
    rec.done_event.set()

    transfer_records.pop(request_id, None)
```

Then all abort/error paths converge on:

```python
finalize_transfer(req_id, ABORTED)
```

or

```python
finalize_transfer(req_id, FAILED, error=e)
```

And `worker.get_finished()` should defensively handle the case where a transfer record still exists but the corresponding request no longer does.

## Why I think this is stronger than the current MR

#12117 says it "ensures failed requests are always cleaned up before `worker.get_finished()` is called." That helps, but it is still tied to one control-flow assumption. It can leave holes if failure happens:

- After transfer registration but before the cleanup path runs
- Concurrently with transfer completion
- Via a different abort source than the repro
- In a future code path that forgets to mirror the same cleanup

A transfer-lifecycle invariant is harder to regress because every request failure path converges on the same terminal cleanup primitive.

## Tests needed to make this robust

The two tests in the current MR are a good start, but I would want coverage for the state machine rather than only the repro.

At minimum:

1. **Repro from #12116** — prompt exceeds `max_seq_len`, request fails, next iteration does not hang, all TP ranks continue.

2. **Abort during in-flight transfer** — start async transfer, cancel request mid-flight, assert terminal state is `ABORTED`, assert no stale transfer record remains.

3. **Transfer exception path** — inject connector failure during transfer, assert terminal state is `FAILED`, assert `get_finished()` does not block forever.

4. **Race / idempotency** — completion and abort happen close together, finalization may be entered twice, no double cleanup, no leaked state, no deadlock.

5. **Stress regression** — repeated oversize prompts / repeated aborts, verify no unbounded growth in transfer bookkeeping, verify no barrier stall after many iterations.

## Red-team concerns with this direction

A few failure modes I would watch for:

- **Double-finalize races**: if completion and abort both try to finalize, the helper must be idempotent.
- **Partial cleanup**: if one rank finalizes locally but peer ranks still believe the transfer is in flight, the deadlock can reappear in a different form.
- **Wrong ownership source of truth**: if request liveness is checked in a place that can lag behind connector state, `get_finished()` may still see stale entries.
- **Terminal-state wakeup without dequeue**: waking the waiter is not enough if the stale record remains discoverable in the next iteration.
- **Fixing only Python-side bookkeeping**: if the Rust / backend layer also tracks transfer lifecycle, terminal-state transition may need to be mirrored there to avoid future inconsistencies.

For that reason, I would strongly prefer a patch that makes terminal-state cleanup the single source of truth for transfer lifecycle, rather than layering more special-case handling around `get_finished()`.

## Bottom line

I would reframe this bug as:

**"Orphaned async KV transfer state can block progress after request abort"**

and fix it by enforcing:

**"Every transfer must terminate as `COMPLETED`, `FAILED`, or `ABORTED`, and terminal cleanup must be idempotent and centralized at the transfer lifecycle layer."**

brnguyen2:
👋 As part of an effort to reduce the TensorRT-LLM open-PR backlog, we're checking in on PRs with no activity in over 120 days. This one qualifies.

Could you let us know whether you still plan to land it?

- **If yes:** please rebase onto the latest `main`, resolve any merge conflicts, and re-request review (or just leave a comment with where it stands).
- **If it's no longer needed:** you can close it, or let us know and we'll close it for you.

If we don't hear back within 14 days, we'll close this PR to keep the review queue manageable. Closing isn't a rejection: the branch, commits, and discussion are all preserved, and you can reopen or resubmit at any time.

Thanks for the contribution!

brnguyen2:
👋 Closing this as part of the TensorRT-LLM open-PR backlog cleanup. We pinged on 2026-08-10 about 120+ days of inactivity and didn't hear back within the 14-day window.

This isn't a rejection — your branch, commits, and discussion are all preserved. If it's still relevant, just **reopen the PR** any time and we'll pick it back up; please rebase onto the latest `main` when you do. Thanks again for the contribution!

PR #12126, opened 2026-03-11
[TRTLLM-11091][feat] Add tunable nvfp4 quantize with additional FlashInfer backend

chang-l:
## Summary
- Add `Fp4QuantKernelRunner` and `trtllm::tunable_fp4_quantize` custom op that autoselects between TRTLLM CUDA and FlashInfer FP4 quantization kernels based on profiled performance
- Update `NVFP4LinearMethod._input_prepare` to use the tunable op for activation quantization
- Add unit tests verifying bitwise-identical output between backends and correct behavior under autotune context

## Details

The design follows the existing `Fp8QuantKernelRunner` pattern:
- During autotune warmup, both TRTLLM and FlashInfer FP4 quantize kernels are profiled for actual input shapes
- The AutoTuner caches the fastest backend per shape with zero inference-time overhead
- Falls back to TRTLLM when FlashInfer is unavailable or outside autotune context
- Includes `register_fake` for `torch.compile` compatibility

The visual_gen pipeline benefits automatically since its warmup already runs within `autotune()` context (pipeline_loader.py L229), which will profile the FP4 quantize kernel for NVFP4-quantized diffusion models.

## Test plan
- [x] `pytest tests/unittest/_torch/thop/parallel/test_fp4_quantize_flashinfer.py -v` — bitwise output comparison + tunable op + autotune context tests
- [x] Existing NVFP4 tests should pass unchanged (backward compatible — defaults to TRTLLM without autotune)
- [x] Visual gen NVFP4 pipeline warmup exercises the new tunable path

## Release Notes

* **New Features**
  * FP4 quantization now supports automatic backend selection between TRTLLM and FlashInfer backends for optimized performance.
  * Introduced tunable FP4 quantization framework that dynamically selects the best-performing backend at runtime.

* **Tests**
  * Added comprehensive test suite validating FP4 quantization across multiple shapes, data types, and backends.


hyukn (commented):
My major concern is the perf impact of introducing the tuning op. It might be safe to do some measurements to avoid unexpected perf regression.

tensorrt_llm/_torch/modules/linear.py
```diff
@@ -1289,7 +1289,7 @@ def _input_prepare(self, module: Linear, input: torch.Tensor):
                 input_scale = module.input_scale
                 alpha = module.alpha
 
-            act_fp4, act_sf = torch.ops.trtllm.fp4_quantize(
+            act_fp4, act_sf = torch.ops.trtllm.tunable_fp4_quantize(
```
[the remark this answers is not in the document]
hyukn: Thanks for the perf data. I doubt that for small input length and low batch sizes, this will has signification impacts on the host side, because there are many quantize kernels invoked. I suggest testing on some cases that are already CPU-bounded to measure the overall impacts.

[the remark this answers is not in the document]
luyiyun1021: We only enable tunable fp4 for visual gen models right now, since these models have larger shapes and may benefit more from this pr. We may enable it for all paths after the overhead can be reduced later.

hyukn (approved):

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -841,6 +841,13 @@ def from_pretrained(
                     cls.load_diffusion_quant_config(quant_dict)
                 )
 
+        # Enable tunable FP4 quantize for visual gen: larger activation
+        # tensors (full image/video latents) amortize the AutoTuner overhead.
+        if quant_config.quant_algo == QuantAlgo.NVFP4:
+            from tensorrt_llm._torch.modules.linear import NVFP4LinearMethod
+
+            NVFP4LinearMethod.use_tunable_quantize = True
```
zhenhuaw-me, line 849: Overwriting a class-level attribute is dangerous.

luyiyun1021: This flag is a workaround. Filed a jira to track this issue. May remove it in the future. https://jirasw.nvidia.com/browse/TRTLLM-11958

tensorrt_llm/_torch/modules/linear.py
```diff
@@ -1193,6 +1193,11 @@ def post_load_weights(self, module: Linear):
 
 class NVFP4LinearMethod(LinearMethodBase):
 
+    # When True, use tunable_fp4_quantize (AutoTuner selects TRTLLM vs
+    # FlashInfer). Visual gen pipelines set this to True before model
+    # construction; LLM paths leave it False to avoid host overhead.
```
zhenhuaw-me, line 1199:
```suggestion
    # Temporary workwround which will be resolved by TRTLLM-11958
    # When True, use tunable_fp4_quantize (AutoTuner selects TRTLLM vs
    # FlashInfer). Visual gen pipelines set this to True before model
    # construction; LLM paths leave it False to avoid host overhead.
```

zhenhuaw-me (approved):

QiJune (approved):

PR #12136, opened 2026-03-12
[None][feat] Add DWDP (Distributed Weight Data Parallelism) support for MoE inference

tianyuz-nv:
## Release Notes

* **New Features**
  * Added DWDP (Double-Weighted Distributed Prefetching) support for disaggregated MoE inference, enabling efficient cross-rank weight prefetching with ping-pong buffering for improved serving performance.
  * Extended MoE kernels to support multiple weight tensor configurations, improving flexibility for advanced quantization and distributed weight management scenarios.

* **Bug Fixes**
  * Fixed rank synchronization in disaggregated MoE initialization to use session-scoped barriers for improved compatibility.

* **Tests**
  * Added DWDP accuracy test for DeepSeek-V3-Lite model with GSM8K benchmark.

## Description

In the context phase of LLM inference, workload imbalances and communication bottlenecks often lead to excessive synchronization overhead, limiting GPU utilization. To resolve this, we propose **Distributed Weight Data Parallelism (DWDP)**, a strategy that leverages Data Parallelism combined with NVLink-based weight offloading to enable fully asynchronous execution across ranks.

Key properties of DWDP:
- **Context-phase acceleration**: DWDP is designed specifically for the context (prefill) phase of disaggregated serving.
- **Hardware-aware**: DWDP performance gains rely on the GB200 NVL72 architecture for high-bandwidth NVLink weight transfer.
- **Fully asynchronous execution**: DWDP eliminates synchronization barriers across ranks, providing significant performance advantages under imbalanced workloads.
- **Fine-grained resource management**: DWDP enables flexible expert-to-worker assignment, allowing more efficient GPU memory utilization.

A detailed technical report on DWDP internals and optimizations will be published separately. We welcome discussions and feedback.

### Changes

**Core DWDP runtime** (`tensorrt_llm/_torch/pyexecutor/dwdp.py` — new file):
- `DwdpManager`: Orchestrates IPC handle exchange across MPI ranks for zero-copy expert weight sharing
- `DwdpHandleCollector`: Per-layer collector that gathers CUDA IPC handles for weight/scale/bias tensors
- Expert weight prefetching with double-buffering via CUDA streams

**MoE integration** (`configurable_moe.py`, `fused_moe_cute_dsl.py`, `interface.py`):
- Integrate DWDP into `ConfigurableMoE` with automatic detection of compatible backends (CuteDSL + NVFP4)
- Add `NvFp4WeightView` dataclass for clean separation of DWDP vs non-DWDP weight access patterns
- Add contiguous gather/scatter grouped GEMM kernels for DWDP expert subset computation

**CuteDSL kernel extensions** (`cute_dsl_custom_ops.py`, `blockscaled_contiguous_*_fusion.py`):
- Extend blockscaled contiguous grouped GEMM kernels to support DWDP's gather-based expert selection
- Unify single-B and multi-B kernel wrappers into a single entry point

**Executor integration** (`py_executor.py`, `py_executor_creator.py`, `_util.py`, `llm_args.py`):
- Add `DwdpConfig` dataclass to `LlmArgs` for YAML-based DWDP configuration
- Initialize `DwdpManager` during executor creation when DWDP is enabled
- Wire DWDP expert prefetching into the per-step execution loop

**Disaggregated serving scripts** (`examples/disaggregated/slurm/benchmark/`):
- Add `start_worker_dwdp.sh` for launching DWDP workers via `mpirun`
- Extend `submit.py` with DWDP-specific configuration (dwdp_size, num_group, experts_per_worker, etc.)

## Test Coverage

- `tests/unittest/_torch/thop/parallel/test_cute_dsl_moe.py`: Unit tests for DWDP-specific CuteDSL MoE kernels (contiguous gather grouped GEMM with SwiGLU fusion, finalize fusion)
- `tests/scripts/cute_dsl_kernels/run_blockscaled_contiguous_gather_grouped_gemm_swiglu_fusion.py`: Standalone kernel correctness tests for DWDP gather GEMM
- `tests/scripts/cute_dsl_kernels/run_blockscaled_contiguous_grouped_gemm_finalize_fusion.py`: Standalone kernel correctness tests for DWDP finalize GEMM
- `tests/integration/defs/accuracy/test_disaggregated_serving.py::TestDeepSeekV3Lite::test_dwdp_accuracy`: End-to-end DWDP disaggregated serving accuracy test with DeepSeek-V3-Lite (NVFP4, 4 GPUs, GSM8K)

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/pyexecutor/dwdp.py
wanqian-nv: 3 is hard-coded, I think it's better to do something?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/blockscaled_contiguous_gather_grouped_gemm_swiglu_fusion.py
```diff
@@ -821,51 +870,82 @@ def __call__(
...
+            tma_atoms_b.append(atom_b_2)
+            tma_tensors_b.append(tensor_b_2)
+            tma_atoms_sfb.append(atom_sfb_2)
+            tma_tensors_sfb.append(tensor_sfb_2)
+        if cutlass.const_expr(self.num_b_tensors >= 4):
+            atom_b_3, tensor_b_3, atom_sfb_3, tensor_sfb_3 = _make_tma_b(b_tuple[3], sfb_tuple[3])
+            tma_atoms_b.append(atom_b_3)
+            tma_tensors_b.append(tensor_b_3)
+            tma_atoms_sfb.append(atom_sfb_3)
+            tma_tensors_sfb.append(tensor_sfb_3)
```
syuoni: So currently, the kernel supports up to 4 B tensors, right?

Does `cutlass.range_constexpr(self.num_b_tensors)` work here?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/blockscaled_contiguous_gather_grouped_gemm_swiglu_fusion.py
```diff
@@ -1872,35 +2048,247 @@ def kernel(
...
+                                    mcast_mask=b_full_mcast_mask,
+                                )
+                                cute.copy(
+                                    tma_atoms_sfb[0],
+                                    tBgSFB_0[(None, slice_n, b_producer_state.count, local_l_0)],
+                                    tBsSFB_pipe,
+                                    tma_bar_ptr=tma_bar,
+                                    mcast_mask=sfb_full_mcast_mask,
+                                )
+                            elif expert_idx < self.b_tensor_l_offsets[2]:
```
syuoni: Can we make a vector that maps `expert_idx` to B tensor index? So that we can avoid these runtime condition checks here.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_cute_dsl.py
```diff
@@ -759,16 +783,64 @@ def run_moe(
...
+                        fc2_global_list=buffer_data["fc2_alpha"],
+                        expert_size_per_partition=self.num_slots,
+                        slot_start=0,
+                    )
+                    nvfp4_args["w3_w1_weight_list"][dwdp_rank] = self.w3_w1_weight
+                    nvfp4_args["fc1_weight_block_list"][dwdp_rank] = self.quant_scales.fc1_weight_block
+                    nvfp4_args["fc1_global_list"][dwdp_rank] = self.quant_scales.fc1_global
+                    nvfp4_args["w2_weight_list"][dwdp_rank] = self.w2_weight
+                    nvfp4_args["fc2_weight_block_list"][dwdp_rank] = self.quant_scales.fc2_weight_block
+                    nvfp4_args["fc2_global_list"][dwdp_rank] = self.quant_scales.fc2_global
```
syuoni: Since these are NVFP4-specific, how about placing them in `run_moe_nvfp4`?

tensorrt_llm/_torch/modules/fused_moe/fused_moe_cute_dsl.py
```diff
@@ -759,16 +783,64 @@ def run_moe(
         Returns:
             final_hidden_states tensor.
         """
+        # Execute MoE computation
         if self.has_nvfp4:
-            return self.run_moe_nvfp4(
+            nvfp4_args = {
```
xxi-nv: The DWDP integration adds 8 weight-related parameters to run_moe_nvfp4 / run_moe_nvfp4_impl (6 → 14 params), and the same weights are named differently across layers (fc1_weight_block_list vs w3_w1_weight_scale vs b_sf_list).

Consider introducing a NvFp4WeightView dataclass to bundle all weight tensors into a single object with consistent naming:


@dataclass
class NvFp4WeightView:
    w3_w1_weight: List[torch.Tensor]       # FC1 weights
    fc1_weight_scale: List[torch.Tensor]   # FC1 block scales
    fc1_global_scale: List[torch.Tensor]   # FC1 global scales
    w2_weight: List[torch.Tensor]          # FC2 weights
    fc2_weight_scale: List[torch.Tensor]   # FC2 block scales
    fc2_global_scale: List[torch.Tensor]   # FC2 global scales
    expert_size_per_partition: int
    slot_start: int
Non-DWDP: CuteDslFusedMoE.get_local_weight_view() wraps self attributes into single-element lists.
DWDP: DwdpManager.build_weight_view() assembles local + prefetched weights (the ~40-line buffer assembly logic currently in run_moe moves here).
run_moe_nvfp4_impl takes one weight_view param instead of eight.
This keeps the non-DWDP path virtually unchanged, gives type safety, and avoids spreading DWDP assembly logic into the MoE compute path.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
```diff
@@ -1315,22 +1370,23 @@ def forward(self, inputs: List[torch.Tensor],
                 0] == self.tile_size, f"Tactic ({tactic}) is incompatible with tile size ({self.tile_size})"
 
             cache_key = (self.scaling_vector_size, self.tile_size, mma_tiler_mn,
-                         cluster_shape_mn, raster_along_m)
+                         cluster_shape_mn, raster_along_m, b_tensor_l_sizes)
```
syuoni: I'm wondering how many possible values `b_tensor_l_sizes` could have in a real scenario? In an EP4 instance, the four values are fixed or not?

tensorrt_llm/_torch/modules/fused_moe/fused_moe_cute_dsl.py
```diff
@@ -530,7 +544,10 @@ def run_moe_nvfp4(
         )
 
         inputs = [
-            x, token_selected_experts, token_final_scales, x_sf, moe_output
+            x, token_selected_experts, token_final_scales, x_sf, moe_output,
+            w3_w1_weight_list, fc1_weight_block_list, fc1_global_list,
+            w2_weight_list, fc2_weight_block_list, fc2_global_list,
+            expert_size_per_partition, slot_start,
         ]
```
syuoni: Should we update cache key for the new inputs?

tensorrt_llm/_torch/modules/fused_moe/fused_moe_cute_dsl.py
```diff
@@ -499,6 +503,16 @@ def run_moe_nvfp4(
...
+        # DWDP: weight parameters as list[tensor] (one per dwdp rank)
+        # list[dwdp_rank] = local weight, list[other_rank] = prefetched weight
+        w3_w1_weight_list: Optional[List[torch.Tensor]] = None,
+        fc1_weight_block_list: Optional[List[torch.Tensor]] = None,
+        fc1_global_list: Optional[List[torch.Tensor]] = None,
+        w2_weight_list: Optional[List[torch.Tensor]] = None,
+        fc2_weight_block_list: Optional[List[torch.Tensor]] = None,
+        fc2_global_list: Optional[List[torch.Tensor]] = None,
+        expert_size_per_partition: Optional[int] = None,
+        slot_start: Optional[int] = None,
```
syuoni: I suppose these tensors are not allowed to be `None` in the current implementation.

In fact I would prefer we have a compatible interface for both DWDP and normal paths. We can handle the two different cases inside this `run_moe_nvfp4` method. In addition, doc string is needed to explain this; otherwise, other users/developers might easily get confused.

I'm also wondering if the code is clearer if we have two `run_moe_nvfp4_impl` and `run_moe_nvfp4_impl_dwdp`. But this will cause some code duplication.

syuoni (commented):
This is a nice feature, great to see it's going to production.

Regarding the CuTeDSL MoE interface, overall I would suggest:
* Use two different ops for single-b and multi-b cases.
* Above the op level, we use two different code paths; this avoids confusion to users -- most users would use the single-b op only.
* Below the op level, we unify them by multi-b implementation; this simplifies implementation and avoids code duplication.

Thanks!

tensorrt_llm/_torch/modules/fused_moe/fused_moe_cute_dsl.py
```diff
@@ -622,25 +646,24 @@ def run_moe_nvfp4_impl(
...
                 top_k=effective_top_k,
-                num_local_experts=self.expert_size_per_partition,
-                local_expert_offset=self.slot_start,
+                num_local_experts=expert_size_per_partition,
+                local_expert_offset=slot_start,
                 tile_size=tile_size,
                 output_dtype=output_dtype,
             )
         else:
             x = torch.ops.trtllm.cute_dsl_nvfp4_grouped_gemm_blackwell(
```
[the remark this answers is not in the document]
tianyuz-nv: No, `cute_dsl_nvfp4_grouped_gemm_blackwell` (the non-fused-finalize FC2 op) does not support multiple B weights, and it does not need to.

This op is only called in `run_moe_nvfp4_impl` (the non-DWDP path). The DWDP path uses a separate function `run_moe_nvfp4_impl_dwdp`, which enforces `use_fused_finalize=True` via an assertion at the top:

```python
assert self.use_fused_finalize, (
    "DWDP requires fused finalize (cute_dsl_nvfp4_grouped_gemm_blackwell "
    "does not support multiple B weight tensors)"
)
```

So DWDP always uses `cute_dsl_nvfp4_grouped_gemm_finalize_inplace_blackwell` (which does support multi-B), and this non-fused op is never reached in the DWDP code path.

cpp/tensorrt_llm/thop/moeAlltoAllOp.cpp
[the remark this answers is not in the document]
tianyuz-nv: Agreed — will double-check with Chuang on this.

Context for the change: in DWDP mode, all workers (CTX + GEN) run within a single MPI world launched via `srun --mpi=pmix`. Using `MpiComm::world().barrier()` would block across all ranks (including different server types), which could cause hangs since MoE AlltoAll only operates within same-type workers. `MpiComm::session().barrier()` scopes the barrier to the current session group.

Will confirm with Chuang whether this change is correct/necessary and update accordingly.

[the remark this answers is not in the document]
wanqian-nv: Hi @chuangz0 Could you please take a look into this change?

xinhe-nv (approved):

syuoni (approved):

tests/integration/defs/accuracy/test_disaggregated_serving.py
```diff
@@ -1813,3 +1918,145 @@ def test_nixl_backend(self):
         with launch_disaggregated_llm(disagg_cfg, ctx_cfg, gen_cfg,
                                       self.MODEL_PATH) as llm:
             run_accuracy_test(llm, self.MODEL_NAME, ["GSM8K"])
+
+
+@contextlib.contextmanager
+def launch_dwdp_disaggregated_llm(
```
Shixiaowei02: Please also move this into a separate file.

Shixiaowei02 (changes requested):

hyukn (approved):
LGTM. But be aware of the potential overhead inside the custom op forward.

xxi-nv (approved):

liji-nv (approved):

chuangz0 (approved):

Shixiaowei02 (approved):

QiJune (approved):
LGTM

PR #12188, opened 2026-03-13
[None][fix] Fix KV cache V2 OOM with separate draft KV cache (EAGLE3/MTP)

yizhang-nv:
@coderabbitai summary

## Description

Fix OOM failures when KV cache manager V2 is used with one-model speculative decoding (EAGLE3/MTP) that requires a separate draft KV cache.

**Root cause:** When V2 creates both a target and a draft KV cache manager, both managers independently claim the full `max_gpu_total_bytes` budget, leading to OOM during `prepare_resources`.

**Changes:**
- **Budget splitting** (`_util.py`): Split `max_gpu_total_bytes` proportionally between target and draft KV cache managers based on per-token KV cache sizes. Only applies to final manager creation (not estimation).
- **Layer count fix** (`_util.py`, `resource_manager.py`): Add `num_layers` parameter to `get_cache_size_per_token` for draft models where the HF config layer count differs from runtime (e.g. EAGLE3: config says 1 layer, runtime uses 4). Prevents `pp_layers()` from giving wrong results for draft layers.
- **PP rank handling** (`_util.py`): Only count draft KV cache cost on the last PP rank where EAGLE3/MTP draft layers reside.
- **Sparse attention config** (`_util.py`): Pass `sparse_attention_config` from `effective_draft_config` to draft KV cache manager for MoE models (e.g. DeepSeek V3 with DSA).
- **BAD_PAGE_INDEX fix** (`kvCacheManagerV2Utils.cu`): Use 0 instead of `BAD_PAGE_INDEX` for padding in `copyBatchBlockOffsetsToDeviceKernel` to avoid out-of-bounds memory access in attention kernels.
- **Graceful resize failure** (`resource_manager.py`): Handle resize failures in dummy request preparation by returning None instead of raising ValueError, letting the scheduler retry.

## Test Coverage

TBD — V2 is not yet enabled by default; will be covered by the enable-v2-by-default PR.

## PR Checklist

### DIFF
--- a/cpp/tensorrt_llm/batch_manager/kvCacheManagerV2Utils.cu
+++ b/cpp/tensorrt_llm/batch_manager/kvCacheManagerV2Utils.cu
@@ -228,8 +228,8 @@ __global__ void copyBatchBlockOffsetsToDeviceKernel(SizeType32 const* __restrict
                 for (uint32_t j = 0; j < elemPerAccess; j++)
                 {
                     auto const val = src.unpacked[j];
-                    dstK.unpacked[j] = (val == BAD_PAGE_INDEX) ? val : (indexScales[poolIdx] * val);
-                    dstV.unpacked[j] = (val == BAD_PAGE_INDEX) ? val : (indexScales[poolIdx] * val + kvOffset[poolIdx]);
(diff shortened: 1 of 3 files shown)


HuiGao-NV (approved):
LGTM

lowsfer (approved):

PR #12189, opened 2026-03-13
[None][fix] replace busy-poll sleep in get_async_noblock with zmq async poller

edenfunf:
## Background

`ZeroMqQueue.get_async_noblock` is used by `GenerationExecutor.aget_stats`
and the RPC server/client paths to receive messages with a timeout. When the
socket has no data ready, the method enters a polling loop that catches
`zmq.Again` and calls `asyncio.sleep(0.01)` before retrying. This fixed
10 ms sleep means every iteration that finds no data incurs at least a 10 ms
delay, causing `aget_stats` to show >10 ms random latency spikes even when
stats are available moments later. The regression was first observed between
`1.2.0rc4` and `1.2.0rc5` when `zmq.NOBLOCK` was introduced (issue #10955).

## Summary

- Replace `asyncio.sleep(0.01)` in the `zmq.Again` handler with a
  `zmq.asyncio.Poller` that blocks until the socket becomes readable or the
  remaining timeout elapses.
- The poller integrates with asyncio's event loop: the coroutine yields to the
  scheduler and is resumed as soon as data arrives, with no artificial minimum
  latency.
- The `zmq.NOBLOCK` flag on the actual `recv` calls is preserved. The original
  motivation for using NOBLOCK was to avoid `asyncio.wait_for` cancelling a
  blocking recv (which can drop messages). This fix only replaces the *wait
  between attempts*—the receive itself is still non-blocking—so message-drop
  safety is maintained.

## Performance Impact

Before: minimum observable latency = N × 10 ms (N = number of empty polls).  
After: coroutine wakes within microseconds of data availability; no fixed-sleep
overhead.

## Verification

Manually tested with a standalone script that:
1. Confirms `TimeoutError` is raised in ≈ the requested timeout duration.
2. Confirms data sent after a 50 ms delay is received in ≈ 53 ms end-to-end
   (previously would have required rounding up to the next 10 ms boundary).

Existing unit tests in `tests/unittest/executor/test_ipc.py`
(`test_get_async_noblock_timeout`, `test_get_async_noblock_success`,
`test_get_async_noblock_router`) cover the corrected code path.

Fixes #10955

## Summary by CodeRabbit

* **Refactor**
  * Enhanced async data reception with improved timeout management and more efficient handling of incoming data.


Superjomn (approved):
LGTM

PR #12208, opened 2026-03-13
[TRTLLM-11492][fix] Fix benchmark disagg deadlock by eliminating blocking fill loop

chienchunhung:
## Summary by CodeRabbit

* **Refactor**
  * Enhanced benchmark disaggregated execution with improved gating logic for KV cache transfer synchronization and more efficient control flow across execution loops.

* **Tests**
  * Added comprehensive test suite for benchmark disaggregated mode, covering fill-completion detection, batch preparation, and allgather handling.

## Description

Fix a deadlock in benchmark disaggregated serving mode and improve code quality.

**Problem:** In benchmark disagg mode, the GEN executor's `_prepare_and_schedule_batch` contained a blocking fill loop that held control until a batch of requests was fetched. During this blocking wait, the executor could not service KV transfers, check timeouts, detect errors, or handle control requests.

The severity depends on the transport backend:
- **MPI:** Sends can block until the receiver posts a matching receive. Since `_check_disagg_gen_transfer_status` never runs inside the fill loop, GEN never receives, CTX's sends block, and CTX cannot free KV cache blocks — causing a **hard deadlock** when CTX capacity < `batch_size`.
- **RDMA (NIXL/UCX):** CTX frees blocks based on send completion independently, so the deadlock is less likely. However, the fill loop still starves all other executor work (timeout handling, error detection, hang detector, control requests) for the duration of the wait.

**Solution:** Replace the blocking fill loop with a non-blocking `can_forward` gate in the executor loops. Each main-loop iteration now performs a full processing cycle — fetch, service transfers, check timeouts, handle errors — then checks readiness via `_is_benchmark_disagg_fill_complete`. This eliminates the deadlock for all backends and ensures the executor remains responsive during the fill phase.

### Key changes

**Core fix:**
- Remove blocking fill loop from `_prepare_and_schedule_batch`, making it fully non-blocking (single fetch per call)
- Add `_is_benchmark_disagg_fill_complete` helper that counts generation-ready requests (excludes ADP dummy requests) and uses TP allgather for ADP, local counting otherwise
- Add `_check_benchmark_disagg_gate` helper that consolidates the gate logic shared by both `_executor_loop` and `_executor_loop_overlap`
- Add `can_forward` gate to `_executor_loop` (previously only existed in `_executor_loop_overlap`)
- Align gate retry sleep to 0.1s consistent with PR #12640

**ADP dummy request fix:**
- Refactor `_pad_attention_dp_dummy_request` into focused helpers: `_count_schedulable_active_requests` (counts requests past KV transfer) and `_should_skip_dummy_for_benchmark_disagg` (encapsulates skip decision)
- Add `_benchmark_fill_phase_active` flag to distinguish the fill phase (dummies skipped — gate blocks collectives, stuck dummies waste KV slots) from taper-down (dummies allowed — normal add-forward-terminate lifecycle handles ranks emptying at different rates due to e.g. speculative decoding acceptance variance)
- Flag starts `True` in benchmark disagg mode and is cleared when `can_forward` gate opens
- Exclude `is_attention_dp_dummy` requests from the generation count in the fill-complete check, preventing premature gate opening

**Code quality:**
- Extract `is_benchmark_disagg` attribute for consistent condition checks (replaces repeated `benchmark_req_queues_size > 0 and kv_cache_transceiver is not None`)
- Use `RuntimeError` for precondition violation in `_is_benchmark_disagg_fill_complete`
- Use `tuple[bool, bool]` return type annotation on `_check_benchmark_disagg_gate`
- Use `logger.debug` for fill-progress messages (avoids log spam at 0.1s intervals)

This PR is co-worked with Cursor agent.

## Test Coverage

Added `tests/unittest/_torch/executor/test_benchmark_disagg.py` with 40 tests across 8 test classes:

| Test class | # | What it covers |
|---|---|---|
| `TestFillCompleteNonADP` | 7 | Threshold (meets/exceeds/below/zero), no-allgather path, rank-0 debug logging, no log on non-zero rank |
| `TestFillCompleteADP` | 7 | Allgather path with threshold variants (meets/exceeds/below/uneven), allgather receives correct local count, dummy exclusion from allgather, logging |
| `TestFillCompleteADPDummyExclusion` | 5 | Regression: dummies alone don't trigger threshold, mixed real+dummy only counts real, below-threshold with dummies, non-ADP dummy exclusion, real-only meets threshold |
| `TestCanForwardGating` | 7 | Gate initialization for all 4 (benchmark\_size, transceiver) combinations, state transitions (complete/incomplete fill), latching (stays True once set) |
| `TestCheckBenchmarkDisaggGate` | 4 | Consolidated gate helper: opens when fill complete (clears fill phase flag), retries with short sleep when incomplete, warmup bypasses gate, already-forwarding skips check |
| `TestPadAttentionDpDummyBenchmarkDisagg` | 7 | Skips during fill phase, skips during fill with requests in transfer, allows dummy after fill phase (taper-down), allows during warmup, allows when not benchmark disagg, no dummy needed when active requests ready, skips when ADP disabled |
| `TestIncrementalFillScenario` | 3 | Multi-iteration convergence with limited CTX capacity, KV transfer lag (one iteration behind), worst-case single-request-at-a-time |
| `TestPrepareAndScheduleBatchNoBlock` | 1 | Verifies `_prepare_and_schedule_batch` calls fetch exactly once per invocation (non-blocking) |
| **Total** | **40** (originally 44, refined after review) | |

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -1951,6 +1951,10 @@ class PyExecutor:
         consolidates the check used by both ``_executor_loop`` and
         ``_executor_loop_overlap``.
 
+        A short sleep (0.1s) yields the CPU between retries while
+        keeping the polling interval short enough to avoid KV transfer
+        backpressure on the CTX server.
+
         Args:
             scheduled_batch: The current scheduled batch.
             can_forward: Current gate state.
@@ -1963,7 +1967,7 @@ class PyExecutor:
             can_forward = self._is_benchmark_disagg_fill_complete(
                 scheduled_batch)
             if not can_forward:
-                time.sleep(1)
+                time.sleep(0.1)
                 return can_forward, True
         return can_forward, False
 
@@ -2944,6 +2948,69 @@ class PyExecutor:
         self.kv_cache_transceiver.prepare_context_requests(
             gen_first_ctx_requests)
 
+    def _count_schedulable_active_requests(self) -> int:
+        """Count active requests that are ready for scheduling.
+
+        In non-disaggregated mode, all active requests are schedulable.
+        In disaggregated mode, requests still waiting for KV cache
+        transfer (in INIT or transmission-in-progress state) are
+        excluded because they cannot participate in the forward pass
+        until transfer completes.
+
+        Returns:
+            The number of active requests eligible for scheduling.
+        """
+        if self.kv_cache_transceiver is None:
+            return len(self.active_requests)
+
+        def _is_awaiting_kv_transfer(req) -> bool:
+            return (req.is_disagg_generation_init_state
+                    or req.is_disagg_generation_transmission_in_progress)
+
+        return sum(1 for req in self.active_requests
+                   if not _is_awaiting_kv_transfer(req))
+
+    def _should_skip_dummy_for_benchmark_disagg(
+            self, num_schedulable_requests: int) -> bool:
+        """Decide whether to skip ADP dummy insertion during benchmark disagg fill.
+
+        Dummies are permanent — once added they are never removed.  In
+        benchmark disagg mode we skip dummy insertion in almost all cases
+        because the ``can_forward`` gate prevents forward-pass collectives
+        during the fill phase, so temporarily-empty ranks are safe.
+
+        The only exception is the terminal case where all benchmark
+        requests have been fetched but there are fewer requests than TP
+        ranks (``benchmark_req_queues_size < tp_size``).  In that case
+        some ranks will **never** receive a real request and need a
+        permanent dummy for forward-pass collectives once the gate opens.
+
+        Args:
+            num_schedulable_requests: Number of active requests that have
+                completed KV transfer and are ready for the forward pass.
+
+        Returns:
+            True if dummy insertion should be skipped for this iteration.
+        """
+        if not self.is_benchmark_disagg or self.is_warmup:
+            return False
+        if num_schedulable_requests > 0:
+            return False
+
+        fill_phase_complete = (self.num_fetch_requests
+                               >= self.benchmark_req_queues_size)
+        some_ranks_permanently_empty = (self.enable_attention_dp
+                                        and self.benchmark_req_queues_size
+                                        < self.dist.tp_size)
+
+        if fill_phase_complete and some_ranks_permanently_empty:
+            return False  # allow dummy — rank will never get a real request
+
+        logger.info(f"Skipped adding dummy requests: "
+                    f"num_fetch_requests={self.num_fetch_requests}, "
+                    f"num_schedulable_requests={num_schedulable_requests}")
+        return True
+
     @nvtx_range("_pad_attention_dp_dummy_request")
     def _pad_attention_dp_dummy_request(self):
         """
@@ -2953,27 +3020,14 @@ class PyExecutor:
             return
 
         assert self.expected_num_active_requests >= len(self.active_requests)
-        if self.kv_cache_transceiver is None:
-            num_active_request = len(self.active_requests)
-        else:
-            num_active_request = len([
-                req for req in self.active_requests
-                if not (req.is_disagg_generation_init_state
-                        or req.is_disagg_generation_transmission_in_progress)
-            ])
+        num_active_request = self._count_schedulable_active_requests()
 
-        # In benchmark disagg mode, skip dummy addition while requests are
-        # still in INIT/transfer state (waiting for KV cache transfer).
-        # We also skip when active_requests is empty (early fill phase) to
-        # avoid occupying a KV cache slot that would block real requests.
-        if (self.is_benchmark_disagg and not self.is_warmup
-                and num_active_request == 0):
-            logger.info(
-                f"Skipped adding dummy requests: num_fetch_requests={self.num_fetch_requests}, {num_active_request=}"
-            )
+        if self._should_skip_dummy_for_benchmark_disagg(num_active_request):
             return
 
-        if self.expected_num_active_requests - num_active_request > 0 and num_active_request == 0:
+        # Other ranks have work but this rank is idle — insert a dummy so
+        # it can participate in collective operations during the forward pass.
+        if num_active_request == 0 and self.expected_num_active_requests > 0:
             llm_request = self.kv_cache_manager.add_dummy_requests(
                 request_ids=[0],
                 is_gen=True,
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -521,6 +521,14 @@ class PyExecutor:
         self.kv_cache_transceiver = kv_cache_transceiver
         self.is_benchmark_disagg = (self.benchmark_req_queues_size > 0
                                     and self.kv_cache_transceiver is not None)
+        # True while the benchmark disagg fill phase is in progress (waiting
+        # for all benchmark requests to complete KV transfer before the first
+        # forward pass).  Cleared by _check_benchmark_disagg_gate when the
+        # can_forward gate opens.  Used by _should_skip_dummy_for_benchmark_disagg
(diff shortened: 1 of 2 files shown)


Tabrizian (commented):
Did you also try the e2e benchmarks to make sure it still works?

chienchunhung:
/bot run --stage-list "DGX_H100-4_GPUs-PyTorch-Others-1" --extra-stage "DGX_B200-8_GPUs-PyTorch-PerfSanity-Post-Merge-1, DGX_B200-8_GPUs-PyTorch-PerfSanity-Post-Merge-2, DGX_B200-8_GPUs-PyTorch-PerfSanity-Post-Merge-3, DGX_B200-8_GPUs-PyTorch-PerfSanity-Post-Merge-4"

pcastonguay (commented):
lgtm except for 2 comments. Will leave it to @Tabrizian to give approval given he's made changes to fix gen-only hang issue lately.

Tabrizian (commented):
Is it possible to also make sure that in this new way each rank only has the same number of requests when attention DP is enabled? Thank you for simplifying this logic!

Tabrizian (approved):

peihu-nv (commented):
I am looking at this PR. Looks to me that there are three changes: 

1. sleep timer is reduced from 10s to 1s. As I mentioned in the other comment, that should help fixing the CTX KV back pressure issue. Recommend to sleep by 0.1s to be more safe based on my testing
2. Some refactoring by moving logics into separate functions are all good to me. 
3. The request fill loop is removed from the main loop to rely on the main loop alone. 

I think 3 is a bigger change: 
Previously with fill loop, it fill a batch of requests so naturally we don't need adding dummy requests as it fills all ranks. With the new change without fill loop, we try to skip adding dummy requests even it does not fill all ranks. Hope my understanding is correct. LMK if I missed anything @chienchunhung

I think overall the change looks good to me. The new logic for 3 should also work fine. Have we run some disagg tests especially the recent problematic ones to confirm it all works? I am just not sure if there is an absolute need for changing the fill logic in 3 so want to make sure it is not causing regressions in some edge cases.

chienchunhung:
Thanks for the review, @peihu-nv!

**On the sleep timer:** Reducing sleep does help with the common RDMA case (NIXL/UCX), where CTX frees blocks based on send completion without waiting for GEN's acknowledgment. For most production configs, 0.1s might be sufficient.

**On why removing the fill loop is still the right fix:** The improvement isn't just about sleep duration — it's structural. The fill loop calls only `_fetch_and_activate_new_requests` in its inner loop, without calling `_check_disagg_gen_transfer_status`. Whether this causes a true deadlock depends on the transport backend (MPI can block on send, RDMA typically doesn't). But even with RDMA, the fill loop prevents the GEN executor from doing any other work (scheduling, transfer status checks, timeout handling) while it waits. Removing the loop makes `_prepare_and_schedule_batch` fully non-blocking, so every main-loop iteration processes transfers, checks timeouts, and fetches — regardless of backend.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2927,6 +2948,69 @@ def _check_disagg_ctx_schedulable_status(self,
...
+        """Decide whether to skip ADP dummy insertion during benchmark disagg fill.
+
+        Dummies are permanent — once added they are never removed.  In
+        benchmark disagg mode we skip dummy insertion in almost all cases
+        because the ``can_forward`` gate prevents forward-pass collectives
+        during the fill phase, so temporarily-empty ranks are safe.
+
+        The only exception is the terminal case where all benchmark
+        requests have been fetched but there are fewer requests than TP
+        ranks (``benchmark_req_queues_size < tp_size``).  In that case
```
brb-nv: Non-blocking comment. Should the degenerate case rather be this?
```
(self.num_fetch_requests >= benchmark_req_queues_size && benchmark_req_queues_size % tp_size != 0)
```

Pls feel free to resolve after sharing your thoughts.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2927,6 +2948,69 @@ def _check_disagg_ctx_schedulable_status(self,
...
+        def _is_awaiting_kv_transfer(req) -> bool:
+            return (req.is_disagg_generation_init_state
+                    or req.is_disagg_generation_transmission_in_progress)
+
+        return sum(1 for req in self.active_requests
+                   if not _is_awaiting_kv_transfer(req))
+
+    def _should_skip_dummy_for_benchmark_disagg(
+            self, num_schedulable_requests: int) -> bool:
+        """Decide whether to skip ADP dummy insertion during benchmark disagg fill.
```
brb-nv, line 2985: Non-blocking comment. Do we skip appending dummy requests even during taper down? i.e. requests got filled to saturation first, but requests might complete at varied pace (say, due to varied acceptance rates with spec dec), so there's a taper down period.

Pls feel free to resolve after sharing your thoughts.

brb-nv: I see you have a change related to this but that's reached after this function.
"Other ranks have work but this rank is idle — insert a dummy"

brb-nv: Maybe it makes sense to `self.is_benchmark_disagg=False` once saturation is achieved the first time?

pcastonguay (approved):

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: The condition intentionally checks < tp_size (not % tp_size != 0) because dummies are only needed when a rank has zero requests. With benchmark_size=5, tp_size=4, the ADP router distributes 2-1-1-1 — every rank has at least one real request, so no dummy would be needed in this case.

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: During taper-down, num_schedulable_requests > 0 on most iterations (real requests are still in generation), so `_should_skip_dummy_for_benchmark_disagg` returns False and the normal dummy lifecycle applies. If a rank temporarily empties, `fill_phase_complete` is True but `some_ranks_permanently_empty` is False, so we still skip — and that's correct, since the `can_forward` gate is open and the dummy will be terminated after the forward pass normally.

On setting `is_benchmark_disagg=False` after saturation: the `can_forward` latch in the executor loop already achieves the same effect — once True, `_check_benchmark_disagg_gate` never re-enters the fill check. Resetting the flag would additionally disable the RuntimeError guard in `_is_benchmark_disagg_fill_complete` if the gate were ever accidentally re-entered. I'd prefer to keep the flag stable for defensive correctness. Happy to revisit if you see a concrete benefit.

chienchunhung: Great catch, @brb-nv. 

After further looking into the potential issue (missing ADP dummies), you're right that during taper-down (e.g., with speculative decoding causing varied acceptance rates), ranks can empty out at different rates. The previous guard was too aggressive — it skipped dummies any time `num_schedulable_requests == 0 with `is_benchmark_disagg == True`, which also catches the taper-down case where dummies are legitimately needed.

I've added a `_benchmark_fill_phase_active` flag that starts `True` and is set to `False` when the can_forward gate opens. `_should_skip_dummy_for_benchmark_disagg()` now only skips when the fill phase is active. Once the benchmark is running (fill phase inactive), the normal dummy add-forward-terminate lifecycle handles taper-down correctly — including the benchmark_size < tp_size terminal case, which no longer needs special handling since the flag cleanly separates the two phases.

This is cleaner than resetting `is_benchmark_disagg = False` because `is_benchmark_disagg` is a configuration fact (we are in benchmark disagg mode), while `_benchmark_fill_phase_active` is runtime state (which phase of that mode we're in).

chienchunhung: Verified that thee2e tests can complete successfully with the updated branch, but not before.

PR #12251, opened 2026-03-16
[TRTLLM-11513][fix] eliminate dangling reference and TOCTOU races in KVCacheManager

thorjohnsen:
## Summary

- **Race 1 — dangling reference in `getSequence`**: `getSequence()` released `mSequencesMtx` before returning a raw reference into `mSequences`, leaving every caller exposed to use-after-free if a concurrent `removeSequence()` extracted the node. Fixed by storing `GenerationRequest` as `shared_ptr` in `mSequences`; `getSequence()` now returns `shared_ptr<GenerationRequest>` (nullptr on miss), decoupling object lifetime from map membership.
- **Race 2 — TOCTOU in `storeContextBlocks`**: removed the double-lock check-then-act pattern (check under lock, release lock, re-acquire in `getSequence`); replaced with a single `getSequence` call and null check.
- **Race 3 — TOCTOU in `rewindKVCache`**: the old existence-check under lock did not cover the `removeToken` loop body. A `shared_ptr` is now held for the loop duration; `removeToken` null-guards its own `getSequence` call.
- **Race 4 — iterator invalidation in `addSequence`**: the old code used a `try_emplace` iterator after releasing `mSequencesMtx`, opening an `unordered_map` rehash window. The `shared_ptr` is now constructed before the lock and the local ptr used after lock release.
- **Race 5 — per-sequence lock acquired and released every iteration**: rewindKVCache called removeToken once per iteration, releasing and re-acquiring the per-sequence lock each time. removeSequence could slip in between iterations, set Removed=true, and cause all subsequent removeToken calls to exit early, leaving the cache partially rewound with no error reported. Fix is to extract removeTokenLocked(GenerationRequest&) which assumes the caller already holds the per-sequence lock, then have rewindKVCache acquire the lock once and call removeTokenLocked in the loop. removeToken continues to acquire the lock itself via the existing path.

## Changed files

| File | What changed |
|---|---|
| `cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h` | `mSequences` value type `GenerationRequest` → `shared_ptr<GenerationRequest>`; two `getSequence` overloads (const + non-const, returning raw refs) replaced by single `const` overload returning `shared_ptr` in both `BaseKVCacheManager` and `KVCacheManager` |
| `cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp` | `getSequence`, `addSequence`, `removeSequence`, `storeContextBlocks`, `rewindKVCache`, `removeToken`, `addToken`, `storeNewBlock`, `storeBlocksForReuse`, `pinBlocks`, `copyBlockOffsets`, `getNeededBlocksOneStep`, `getCacheBlockIds`, `getBatchCacheBlockIds`, `getLastBlockId`, `getRemainingBlocksToCompletion` |
| `cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp` | 14 `getSequence` call sites updated (`auto& seq = *manager.getSequence(...)`) — all downstream `.member()` access preserved unchanged |

## Test plan

- [x] All existing tests in `cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp` pass without modification to test logic
- [x] Any other `BaseKVCacheManager` subclasses (mocks, V2 variants) that implement the old two-overload `getSequence` virtual will need the same single-overload update
- [x] Stress-test with overlap scheduling / MTP mode to exercise the concurrent `removeSequence` + `rewindKVCache` path called out in the original comment

## Notes

Races 5–8 from the same analysis (non-atomic `KVCacheBlock` ref-counts, unprotected `BlockManager`, statistics counters, misleading mutex naming) are left for follow-on work to keep this PR reviewable.

## Release Notes

* **Refactor**
  * Enhanced robustness of KV cache sequence management with improved error handling.
  * Increased reliability of cache operations in edge cases and concurrent scenarios.


eopXD (commented):
Overall looks good to me with only a few comments.

PR #12256, opened 2026-03-16
[None][feat] Add PyTorchWorker backend for scaffolding framework (#3333)

### LINKED ISSUE #3333: Support scaffolding pytorch  worker
There are many papers on inference time calculation that use PyTorch to implement some components, such  reward module. We need a native PyTorch backend to reduce the difficulty of introducing new work.

nikJ13:
## Description

Closes #3333
Closes #3334

Adds a native PyTorch worker (`PyTorchWorker`) to the scaffolding framework, enabling inference with
HuggingFace models without TensorRT compilation. This supports inference-time compute patterns
(e.g. reward models for Best-of-N selection) that research papers commonly implement directly in PyTorch,
and it lowers the barrier for contributors who do not have a GPU development environment handy.

### What's included

- **`tensorrt_llm/scaffolding/pytorch_worker.py`** — new worker with three handlers:
  - `generation_handler`: generation via `model.generate()` with temperature, top-k/top-p, stop sequences,
    logprobs, and `return_context_logits`
  - `stream_generation_handler`: token-by-token decoding with KV cache, supporting pause/resume/cancel
  - `reward_handler`: scoring via sequence-classification models
- **`tensorrt_llm/scaffolding/__init__.py`** — export `PyTorchWorker` in the public API
- **`tensorrt_llm/scaffolding/controller.py`** — one-line fix so `BestOfNController` returns the selected
  candidate's output (details below)
- **`tests/unittest/scaffolding/test_pytorch_worker.py`** — unit tests for all handlers

### Design

- Subclasses the existing `Worker` ABC and registers handlers through the `task_handlers` class attribute,
  matching `OpenaiWorker` / `TRTLLMWorker`.
- Factory methods: `from_pretrained()` for causal LMs, `from_pretrained_reward_model()` for classifiers.
- Accepts both `input_str` and `input_tokens` for all task types.
- Works with the existing controllers (`NativeGenerationController`, `BestOfNController`,
  `MajorityVoteController`, …).

## Changes since the previous CI run

The previous run (`PR_Github #53348` / `L0_MergeRequest_PR #42529`) failed. This revision addresses the
following and rebases onto current `main`:

1. **Reverted the repo-wide `TaskStatus.WORKER_EXECEPTION` → `WORKER_EXCEPTION` rename.** The rename was
   authored against a February merge base and left dangling references on `main`
   (`worker.py:403,447,461,512,517,977` and `contrib/mcp/chat_handler.py:67` still use the original spelling),
   which would raise `AttributeError` on those error paths. It was also a breaking change to a publicly
   exported enum and unrelated to this PR's concern. `pytorch_worker.py` now uses the existing spelling; a
   non-breaking alias can be added separately if desired.
2. **Tests no longer download `gpt2` from the HuggingFace Hub.** The model now resolves through
   `llm_models_root()`, matching `tests/unittest/scaffolding/test_worker.py`, with a `pytest.skip` when the
   model cache is unavailable.
3. **Narrowed exception handling** in the three handlers from bare `except Exception` to
   `except (RuntimeError, ValueError)`, and routed the error through `tensorrt_llm.logger` instead of `print`,
   per `CODING_GUIDELINES.md` "Error Handling".
4. **Rebased onto `main`** and squashed into a single signed-off commit (no merge commit), as requested.
5. Removed the `examples/scaffolding/run_pytorch_backend.py` example per review feedback — the unit tests
   cover the same ground.
6. Issue links: this PR implements **#3333** (native PyTorch worker) and also resolves **#3334**
   (CPU inference for scaffolding) — `PyTorchWorker` runs on CPU via `device="cpu"`, and the unit tests
   exercise that path. Linking both was confirmed with @WeiHaocheng and @amemov on #3334.

## Reward path: what works today

Reward scoring has two entry points, and they behave differently:

- **`PRMController` (works directly).** It sends a `GenerationTask` with `return_context_logits=True` and
  scores from `task.context_logits`. `generation_handler` now honours that flag, so `PyTorchWorker` can act
  as the reward worker for `PRMController` with no extra work. Covered by
  `TestPyTorchWorkerContextLogits::test_drives_prm_controller`.
- **`reward_handler` requires a `RewardTask`-producing controller.** `Worker.run_task` dispatches on the exact
  task type, while `NativeRewardController` forwards the `GenerationTask` objects it is given (its
  `task = GenerationTask()` line is dead code, shadowed by the loop variable). A worker on the reward tag
  therefore receives a `GenerationTask` and lands in `generation_handler`. Until that contract changes, using
  `reward_handler` means supplying a controller that emits `RewardTask` — as
  `TestPyTorchWorkerBestOfN` does. Changing `NativeRewardController` to emit `RewardTask` would make
  `TRTLLMWorker` return `WORKER_NOT_SUPPORTED` on the reward tag, since no existing worker registers a
  `RewardTask` handler, so that is left to a follow-up PR (link to be added here once opened).

### Incidental fix in `BestOfNController`

`BestOfNController.process` ended with `task.result = best_task.result`. No generation task type defines
`result` (only `MCPCallTask` has `result_str` and friends), so any end-to-end Best-of-N run raised
`AttributeError` and produced an empty `ScaffoldingOutput` — `examples/scaffolding/run_best_of_n_with_reward.py`
included. It now propagates `output_str` and `output_tokens` from the winning candidate, matching what
`MajorityVoteController` already does. Nothing could have depended on the old line, since it always raised.

## Test Coverage

`tests/unittest/scaffolding/test_pytorch_worker.py`, picked up by the existing `unittest/scaffolding`
entry in `tests/integration/test_lists/test-db/l0_h100.yml`:

- **Initialization**: `from_pretrained()`, explicit device selection
- **Generation**: basic, `input_tokens` path, logprobs, stop sequences, deterministic output at `temperature=0`
- **Streaming**: completion, cancel, `input_tokens` path
- **Reward**: scoring through the `input_tokens` fallback; end-to-end Best-of-N through `ScaffoldingLlm`
  with a sequence-classification reward model
- **Context logits**: shape, absence unless requested, and scoring driven through the real `PRMController`
- **Integration**: single-prompt and batch generation through `ScaffoldingLlm`
- **Shutdown**: resource cleanup

## PR Checklist

## GitHub Bot Help

## Dev Engineer Review

- Added and exported `PyTorchWorker` for HuggingFace causal-language-model generation and sequence-classification reward scoring.
- Added batch generation, streaming generation, sampling controls, stop sequences, logprobs, context logits, CPU execution, token inputs, pause/resume/cancel state, and shutdown handling.
- Fixed stop-sequence truncation to use the earliest matching stop string.
- Updated reward handling to reject tasks without text or token inputs.
- Updated `BestOfNController` to propagate `output_str` and `output_tokens` from the selected generation task.
- The built-in reward path still dispatches `GenerationTask` objects. `reward_handler` therefore requires a custom controller that produces `RewardTask`.
- The deferred `NativeRewardController` and `RewardTask` contract change remains a follow-up item.
- No configuration or test-list files changed.

## QA Engineer Review

- Added `tests/unittest/scaffolding/test_pytorch_worker.py`.
- Coverage includes worker initialization, device selection, string and token generation, invalid input handling, logprobs, stop sequences, deterministic generation, streaming completion and cancellation, reward routing, context logits, PRM scoring, Best-of-N generation, `ScaffoldingLlm` single-request and batch generation, shutdown, and resource cleanup.
- Added regression coverage for stop-sequence ordering and reward-input validation.
- Tests skip when the cached GPT-2 model lacks the required configuration or weight files.
- The test functions are not listed in `tests/integration/test_lists/` files. They are unit tests and do not require test-list entries.
- Verdict: sufficient.

### DIFF
--- a/tensorrt_llm/scaffolding/pytorch_worker.py
+++ b/tensorrt_llm/scaffolding/pytorch_worker.py
@@ -1,0 +1,470 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""PyTorch backend worker for TensorRT-LLM Scaffolding.
+
+This module provides a worker implementation that uses native PyTorch and
+HuggingFace transformers for inference, enabling easy integration of research
+components without requiring TensorRT compilation.
+"""
+
+import copy
+import traceback
+from typing import List, Optional, Union
+
+import torch
+from transformers import (
+    AutoModelForCausalLM,
+    AutoModelForSequenceClassification,
+    AutoTokenizer,
+    GenerationConfig,
+    PreTrainedModel,
+    PreTrainedTokenizer,
+    StoppingCriteria,
+    StoppingCriteriaList,
+)
+
+from tensorrt_llm.executor.result import Logprob
+from tensorrt_llm.logger import logger
+
+from .result import ScaffoldingOutput
+from .task import GenerationTask, RewardTask, StreamGenerationTask, TaskStatus
+from .worker import Worker
+
+
+class _StopStringCriteria(StoppingCriteria):
+    """Custom stopping criteria that checks for stop strings in decoded output."""
+
+    def __init__(self, tokenizer: PreTrainedTokenizer, stop_strings: List[str], input_length: int):
+        self.tokenizer = tokenizer
+        self.stop_strings = stop_strings
+        self.input_length = input_length
+
+    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs) -> bool:
+        generated = self.tokenizer.decode(
+            input_ids[0, self.input_length :], skip_special_tokens=True
+        )
+        return any(s in generated for s in self.stop_strings)
+
+
+class PyTorchWorker(Worker):
+    """Worker that executes tasks using PyTorch and HuggingFace models.
+
+    This worker enables running inference with pure PyTorch models, which is
+    useful for:
+    - Rapid prototyping without TensorRT compilation
+    - Custom research models (reward models, verifiers, etc.)
+    - Models with architectures not yet supported by TensorRT
+    - Debugging and analysis requiring PyTorch-native tools
+
+    The worker supports generation tasks (using causal LMs), streaming
+    generation tasks (with pause/resume/cancel), and reward tasks (using
+    sequence classification models).
+
+    Example:
+        ```python
+        # Create worker with a HuggingFace model
+        worker = PyTorchWorker.from_pretrained("gpt2")
+
+        # Use in scaffolding
+        llm = ScaffoldingLlm(controller, {"generation": worker})
+        ```
+    """
+
+    def __init__(
+        self,
+        model: PreTrainedModel,
+        tokenizer: PreTrainedTokenizer,
+        device: Optional[Union[str, torch.device]] = None,
+        max_batch_size: int = 1,
+    ):
+        self.model = model
+        self.tokenizer = tokenizer
+
+        if device is None:
+            self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        else:
+            self.device = torch.device(device) if isinstance(device, str) else device
+
+        self.model.to(self.device)
+        self.model.eval()
+
+        self.max_batch_size = max_batch_size
+
+    @classmethod
+    def from_pretrained(
+        cls,
+        model_name_or_path: str,
+        device: Optional[Union[str, torch.device]] = None,
+        torch_dtype: Optional[torch.dtype] = torch.float16,
+        trust_remote_code: bool = False,
+        **model_kwargs,
+    ) -> "PyTorchWorker":
+        """Create a PyTorchWorker from a HuggingFace model name or path.
+
+        Args:
+            model_name_or_path: HuggingFace model identifier or local path
+            device: Device to run on (default: auto-detect CUDA)
+            torch_dtype: Data type for model weights (default: float16)
+            trust_remote_code: Whether to trust remote code in model
+            **model_kwargs: Additional arguments passed to model loading
+
+        Returns:
+            PyTorchWorker instance with loaded model
+        """
+        tokenizer = AutoTokenizer.from_pretrained(
+            model_name_or_path,
+            trust_remote_code=trust_remote_code,
+            padding_side="left",
+        )
+
+        if tokenizer.pad_token is None:
+            tokenizer.pad_token = tokenizer.eos_token
+
+        model = AutoModelForCausalLM.from_pretrained(
+            model_name_or_path,
+            torch_dtype=torch_dtype,
+            trust_remote_code=trust_remote_code,
+            **model_kwargs,
+        )
+
+        return cls(model, tokenizer, device=device)
+
+    @classmethod
+    def from_pretrained_reward_model(
+        cls,
+        model_name_or_path: str,
+        device: Optional[Union[str, torch.device]] = None,
+        torch_dtype: Optional[torch.dtype] = torch.float16,
+        trust_remote_code: bool = False,
+        **model_kwargs,
+    ) -> "PyTorchWorker":
+        """Create a PyTorchWorker for reward modeling tasks.
+
+        Loads a sequence classification model suitable for scoring text.
+        """
+        tokenizer = AutoTokenizer.from_pretrained(
+            model_name_or_path,
+            trust_remote_code=trust_remote_code,
+        )
+
+        if tokenizer.pad_token is None:
+            tokenizer.pad_token = tokenizer.eos_token
+
+        model = AutoModelForSequenceClassification.from_pretrained(
+            model_name_or_path,
+            torch_dtype=torch_dtype,
+            trust_remote_code=trust_remote_code,
+            **model_kwargs,
+        )
+
+        return cls(model, tokenizer, device=device)
+
+    def _tokenize_input(self, input_str: Optional[str], input_tokens: Optional[list]):
+        """Tokenize from input_str or input_tokens, returning model inputs dict.
+
+        Returns None if neither input_str nor input_tokens is provided.
+        """
+        if input_str is not None:
+            return self.tokenizer(input_str, return_tensors="pt", padding=True, truncation=True).to(
+                self.device
+            )
+        elif input_tokens is not None:
+            input_ids = torch.tensor([input_tokens], device=self.device)
+            return {
+                "input_ids": input_ids,
+                "attention_mask": torch.ones_like(input_ids),
+            }
+        return None
+
+    def convert_task_params(self, task: GenerationTask) -> GenerationConfig:
+        """Convert task sampling parameters to HuggingFace GenerationConfig."""
+        return GenerationConfig(
+            max_new_tokens=task.max_tokens if task.max_tokens is not None else 100,
+            temperature=task.temperature if task.temperature is not None else 1.0,
+            top_p=task.top_p if task.top_p is not None else 1.0,
+            top_k=task.top_k if task.top_k is not None else 50,
+            do_sample=(task.temperature is not None and task.temperature > 0),
+            pad_token_id=self.tokenizer.pad_token_id,
+            eos_token_id=self.tokenizer.eos_token_id,
+        )
+
+    def _build_stopping_criteria(self, task: GenerationTask, input_length: int):
+        """Build stopping criteria list from task stop sequences."""
+        if not task.stop:
+            return None
+        stop_strings = [task.stop] if isinstance(task.stop, str) else task.stop
+        if not stop_strings:
+            return None
+        return StoppingCriteriaList(
+            [_StopStringCriteria(self.tokenizer, stop_strings, input_length)]
+        )
+
+    def _strip_stop_sequence(self, text: str, task: GenerationTask) -> str:
+        """Remove trailing stop sequence from generated text."""
+        if not task.stop:
+            return text
+        stop_strings = [task.stop] if isinstance(task.stop, str) else task.stop
+        for s in stop_strings:
+            idx = text.find(s)
+            if idx != -1:
+                return text[:idx]
+        return text
+
+    def _extract_logprobs(self, scores: tuple, generated_tokens: List[int], num_logprobs: int):
+        """Extract top-k logprobs from generation scores.
+
+        Args:
+            scores: Tuple of [batch_size, vocab_size] tensors per step.
+            generated_tokens: List of token IDs that were generated.
+            num_logprobs: Number of top logprobs to return per token.
+
+        Returns:
+            TokenLogprobs (list[dict[int, Logprob]])
+        """
+        logprobs_list = []
+        for step_idx, step_scores in enumerate(scores):
+            log_probs = torch.log_softmax(step_scores[0].float(), dim=-1)
+            token_dict = {}
+
+            if num_logprobs > 0:
+                k = min(num_logprobs, log_probs.shape[0])
+                topk_vals, topk_indices = torch.topk(log_probs, k=k)
+                for rank, (val, idx) in enumerate(zip(topk_vals, topk_indices)):
+                    token_dict[idx.item()] = Logprob(logprob=val.item(), rank=rank + 1)
+
+            # Always include the actually generated token
+            if step_idx < len(generated_tokens):
+                gen_token_id = generated_tokens[step_idx]
+                if gen_token_id not in token_dict:
+                    gen_logprob = log_probs[gen_token_id].item()
+                    gen_rank = (log_probs > gen_logprob).sum().item() + 1
+                    token_dict[gen_token_id] = Logprob(logprob=gen_logprob, rank=gen_rank)
+
+            logprobs_list.append(token_dict)
+        return logprobs_list
+
+    def _sample_token(self, logits: torch.Tensor, gen_config: GenerationConfig) -> torch.Tensor:
+        """Sample a single token from logits using the generation config."""
+        if not gen_config.do_sample:
+            return torch.argmax(logits, dim=-1)
+
+        logits = logits / max(gen_config.temperature, 1e-7)
+
+        # Top-k filtering
+        if gen_config.top_k > 0:
+            topk_vals = torch.topk(logits, gen_config.top_k)[0]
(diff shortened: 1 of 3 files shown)


nikJ13 (commented):
Please review the changes. I have also removed run_pytorch_backend.py

WeiHaocheng (approved):

nikJ13:
Hi @WeiHaocheng could you merge my changes

WeiHaocheng:
> Hi @WeiHaocheng could you merge my changes

Sure, let us push it. You need to rebase the code from the main branch and then I will run the CI.

karljang:
@WeiHaocheng, could you please continue your review? Thanks!

karljang:
@WeiHaocheng, ah never mind. I'll run the CI since you already approved. Thanks

nikJ13:
@karljang @WeiHaocheng — I've rebased onto current `main` and pushed a single signed-off commit (no merge commit), plus fixes for what I believe broke the last run. Since the `L0_MergeRequest_PR #42529` report is on internal infrastructure I can't reach, I worked back from the tree the run was triggered on. Two concrete defects:

1. **Dangling `TaskStatus` references.** My earlier commit renamed `WORKER_EXECEPTION` → `WORKER_EXCEPTION` repo-wide, but it was written against a February merge base. Against current `main` that leaves the original spelling in `worker.py:403,447,461,512,517,977` and `contrib/mcp/chat_handler.py:67` pointing at an enum member that no longer exists, so those error paths would raise `AttributeError`. It was also a breaking change to a publicly exported enum and unrelated to this PR, so I've dropped the rename entirely — `pytorch_worker.py` now uses the existing spelling. Happy to add a non-breaking alias in a separate PR if you'd like the correct spelling available.

2. **The test pulled `gpt2` from the HuggingFace Hub.** It now resolves through `llm_models_root()`, matching `tests/unittest/scaffolding/test_worker.py`, and skips when the model cache isn't mounted.

I also narrowed the three handlers from bare `except Exception` to `except (RuntimeError, ValueError)` and routed errors through `tensorrt_llm.logger`, per the Error Handling section of `CODING_GUIDELINES.md`.

The diff is now exactly three files (+800/−0): the new worker, its unit test, and a two-line export.

Verification I was able to do locally: `pre-commit run --files` on all three paths passes with no hook modifications, and I confirmed the worker's full API surface still exists on current `main`. I could not run the scaffolding tests locally — my local C++ build is stale relative to the checkout — so the unit tests are unverified on my side.

Could one of you re-trigger CI when you get a chance? Thanks for the patience on this one.

nikJ13:
Correction to my previous comment: I said the unit tests were unverified on my side. I've since been able to run them, and they pass.

My local build was too stale to import `tensorrt_llm`, so I installed the `tensorrt_llm 1.3.0rc9` cp310 wheel from `pypi.nvidia.com` into a throwaway venv on a GPU node, dropped `pytorch_worker.py` onto the installed `scaffolding` package, and ran this PR's test file from `upstream/main`'s `tests/` tree:

```
15 passed, 2 warnings in 19.52s
```

All 15 pass — initialization, generation (basic, `input_tokens`, logprobs, stop sequences, deterministic at `temperature=0`), streaming (completion, cancel, `input_tokens`), reward, the two `ScaffoldingLlm` integration tests, and shutdown. Same result with `--noconftest`.

Worth noting for the model-download issue: the run had `HF_HUB_OFFLINE=1` set with `LLM_MODELS_ROOT` pointing at a local cache, and **nothing was skipped** — so `llm_models_root()` genuinely resolved the model and no HuggingFace download was attempted.

One caveat I want to be straight about: 1.3.0rc9 is a release wheel, not current `main`, so this exercises the worker against that version's scaffolding rather than the exact tree the PR targets. I separately checked every scaffolding API the worker touches against current `main` — `Worker.task_handlers`, `ScaffoldingOutput(text, token_ids)`, `GenerationTask.create_from_prompt`, the `StreamGenerationTask` fields (`cancel_flag`, `streaming_step`, `request_handle`, `end_flag`), and `custom_output_params` / `streaming_output_*` inherited from base `Task`. All present and unchanged.

nikJ13:
@karljang thanks for running it. `L0_MergeRequest_PR #51295` came back FAILURE on `8253bc7` — could you paste the failed test names? The report hosts (`nv/trt-llm-cicd`, `pbss.s8k.io`, `tensorrt-llm-ci-report...nvidia.com`) aren't reachable from outside NVIDIA, so I'm flying blind on which stage broke.

Most useful would be whether any failure is under `unittest/scaffolding`, or whether they're all elsewhere in the pipeline. The run used `--disable-fail-fast`, so I'd expect unrelated stages to be included too.

For context on what I can rule out from my side:

- This PR touches three files and adds no import to any existing module beyond a two-line export in `tensorrt_llm/scaffolding/__init__.py`. It does not modify `task.py`, `worker.py`, or anything outside `scaffolding`.
- `tests/unittest/scaffolding/test_pytorch_worker.py` resolves its model through `llm_models_root() / "gpt2"` and calls `pytest.skip` if the cache or the model is missing — so on a runner without that model it should skip rather than fail. If it is failing rather than skipping, the message would be very helpful.
- I ran the file locally against current `main`'s scaffolding package: 15 passed, 0 failed, with `HF_HUB_OFFLINE=1` set and nothing skipped, so no HuggingFace download is attempted.

If the failures turn out to be unrelated to scaffolding, happy to just rebase and have it re-run. If they are mine, the test names alone should be enough for me to fix them quickly.

karljang:
the failures appear unrelated to the PR, I'll rerun it.

nikJ13:
hi @karljang the CI failed again.

karljang:
Interestingly, the ci didn't run :(

nikJ13:
@karljang thanks for chasing the trigger — glad it went through this time. `L0_MergeRequest_PR #51474` came back FAILURE on the same commit `8253bc7`.

Since I can't reach the report from outside NVIDIA, could you check whether any of the failures are under `unittest/scaffolding`? If they look like the same unrelated set as `#51295`, I'm happy to just wait for another run rather than take up more of your time.

In the meantime I'm running a differential check on my side — the full `unittest/scaffolding` suite against current `main` with and without this PR's three files, so I can tell whether the change introduces any failure of its own. I'll report the outcome here either way.

nikJ13:
Found and fixed a real bug in this PR — I believe it explains the repeated failures. Pushed as `cc56bcf` on top of your merge, @karljang.

**The bug.** The test fixture resolved its model as `llm_models_root() / "gpt2"` and only checked that the directory *exists*. On `main`, that path is referenced in exactly one place — `tests/unittest/llmapi/test_llm.py`, as a **`tokenizer_dir`**:

```python
(get_model_path('gpt2'), False, 0.95),  # BPE
```

So the cache entry may hold only tokenizer files. In that case the guard passes, `AutoModelForCausalLM.from_pretrained()` raises, and the tests fail rather than skipping — deterministically, on every run, in `l0_h100`. That matches the pattern: five straight failures here while other PRs pass intermittently.

**Reproduced and verified.** I built a tokenizer-only cache directory and ran the file three ways against the same wheel + current `main` scaffolding:

| model dir | old guard (`8253bc7`) | new guard (`cc56bcf`) |
|---|---|---|
| full checkout | 15 passed | 15 passed |
| **tokenizer-only** | **3 failed, 12 errors** | **15 skipped** |
| absent | 15 skipped | 15 skipped |

The fix requires `config.json` plus at least one weight file (`*.safetensors` / `pytorch_model*.bin` / `model*.bin`) before using the directory, so a tokenizer-only or partial entry skips cleanly instead of erroring. It is one commit touching only the test file, +16/-3.

**What I can and can't claim.** I have not seen your CI report, so I can't confirm the `gpt2` entry on your runners is tokenizer-only. What I can say is that this was a genuine defect that produces exactly the observed symptom, it's now reproduced and fixed, and the test is robust to either cache layout.

Sorry for the runs this cost. Whenever you have a moment, a fresh trigger would tell us whether that was it.

karljang:
@nikJ13, thank you for your investigation! I’ve reviewed the failed tests, but they seem irrelevant once again. 😄

nikJ13:
Hi @karljang the ci has failed again. Is there any particular reason for it? Maybe I could make some changes in my commit?

karljang:
No worries, the CI has been unstable lately, and the failures still look unrelated to your PR. I'll leave a comment here if anything needs your attention. Thank you!

However, could you please take a look at this concern raised by my agent? 

> The existing reward controllers yield GenerationTasks, while Worker.run_task() dispatches by task type. This means a PyTorchWorker registered as the reward worker would call generation_handler() instead of reward_handler(). Could you wire RewardTask into the controller path and add an end-to-end Best-of-N test with a sequence-classification model?

nikJ13:
Thanks — and the agent is right. I traced it, and it's actually a bit worse than described.

**Confirmed.** `Worker.run_task()` dispatches on the exact `type(task)`. `BestOfNController.process()` builds `generation_tasks = [copy.deepcopy(task) ...]`, i.e. `GenerationTask`s, and hands those to `reward_controller.process()`. `NativeRewardController.process()` only stamps `worker_tag = REWARD` on the objects it receives and yields them unchanged — its `task = GenerationTask()` line is dead, immediately shadowed by the loop variable. So a `PyTorchWorker` on the reward tag receives a `GenerationTask` and lands in `generation_handler()`. `reward_handler()` is unreachable through that path.

**The wider point.** Nothing on `main` uses `RewardTask` today: no controller constructs one, and no worker registers a handler for it (`TRTLLMWorker.task_handlers` covers only `GenerationTask` and `StreamGenerationTask`). The reward path that actually works is `PRMController`, which sends a `GenerationTask` with `return_context_logits=True` and `max_tokens=1`, then computes scores from `task.context_logits`. My worker doesn't honour `return_context_logits` either — so it can't currently serve `PRMController` any more than it can serve `reward_handler`.

**What I'd like to do inside this PR**, since both are confined to `pytorch_worker.py` and its test:

1. Honour `return_context_logits` in `generation_handler` and populate `task.context_logits`, so `PyTorchWorker` works with `PRMController` — the existing, functioning reward path.
2. Add the end-to-end Best-of-N test you asked for, driving it through `ScaffoldingLlm` with a sequence-classification model, using a small test-local reward controller that yields a `RewardTask`. That exercises `reward_handler` end to end without changing library code.

**The one part I don't want to decide unilaterally.** Making `NativeRewardController` yield `RewardTask` would fix the dispatch mismatch properly, but since no existing worker registers `RewardTask`, `TRTLLMWorker` on the reward tag would start returning `WORKER_NOT_SUPPORTED` — a behaviour change outside this PR's scope, on a PR you've already approved. Options as I see them:

- **(i)** I do 1 and 2 here, and open a separate PR switching `NativeRewardController` to `RewardTask` together with a `RewardTask` handler for `TRTLLMWorker`, so the two land atomically.
- **(ii)** I do all of it in this PR and you re-review the larger diff.
- **(iii)** I do 1 and 2 here and just document the `RewardTask` limitation, leaving the controller contract alone.

I'd lean towards (i): it keeps this PR reviewable and avoids a cross-cutting change to controller semantics buried in a worker PR. Happy to go with whichever you prefer — just say which and I'll get it done.

karljang:
Thanks for the detailed investigation. Option (i) sounds best to me. Please implement 1 and 2 in this PR, and address the built-in `NativeRewardController`/`RewardTask` contract in a follow-up.

Could you also test `return_context_logits` through the actual `PRMController`? For the sequence-classification path, the Best-of-N test with a small custom controller sounds sufficient for this PR.  Could you also clarify in the description that a custom RewardTask-producing controller is currently required, and link the follow-up once it is created.

nikJ13:
@karljang all of the above is in, on `e96ed15`.

`return_context_logits` is honoured and tested through the real `PRMController`; Best-of-N runs end to end through `ScaffoldingLlm` with a sequence-classification reward model; and the description now spells out that `reward_handler` needs a `RewardTask`-producing controller, with the contract change left for the follow-up.

Two things worth a look while reviewing: `BestOfNController.process` ended with `task.result = best_task.result`, but no generation task type defines `result`, so end-to-end Best-of-N raised `AttributeError` and returned empty output — it now propagates `output_str`/`output_tokens` like `MajorityVoteController` does. I also fixed a stop-sequence bug found on the way, where `_strip_stop_sequence` cut at the first listed stop string rather than the earliest match.

22 tests pass locally and pre-commit is clean. Could you trigger CI when you get a chance?

nikJ13:
hi @karljang , wanted to know if the changes can be merged?

tensorrt_llm/scaffolding/pytorch_worker.py
```diff
@@ -0,0 +1,482 @@
...
+            gen_config = self.convert_task_params(task)
+            input_length = inputs["input_ids"].shape[1]
+
+            # Prompt-position logits, matching TRTLLMWorker's context_logits. Reward
+            # controllers such as PRMController score from these rather than from the
+            # generated text. generate() only exposes scores for generated steps, so
+            # this needs its own forward pass over the prompt.
+            if task.return_context_logits:
+                with torch.no_grad():
+                    task.context_logits = self.model(**inputs).logits[0]
```
karljang, line 305: I may be missing something here, but are these not causal-LM logits shaped [seq_len, vocab_size]? PRMController seems to treat them as [seq_len, 2] reward logits and reads label 1. With GPT-2, the test would therefore be scoring vocabulary token 1 rather than an actual reward label. Could you take another look at whether this test is exercising a real PRM model?

tensorrt_llm/scaffolding/pytorch_worker.py
```diff
@@ -0,0 +1,482 @@
...
+            generate_kwargs = {
+                "generation_config": gen_config,
+                "return_dict_in_generate": True,
+                "output_scores": task.num_logprobs is not None,
+            }
+            if stopping_criteria is not None:
+                generate_kwargs["stopping_criteria"] = stopping_criteria
+
+            with torch.no_grad():
+                outputs = self.model.generate(**inputs, **generate_kwargs)
```
karljang, line 318: One concern here: since model.generate() is synchronous, will this not block the scaffolding event loop even though the handler is async? In particular, Best-of-N tasks appear to be scheduled concurrently but would actually run one at a time. Do we need to move the model execution off the event loop or explicitly queue or batch these requests?

tensorrt_llm/scaffolding/pytorch_worker.py
```diff
@@ -0,0 +1,482 @@
...
+        self.model.eval()
+
+        self.max_batch_size = max_batch_size
+
+    @classmethod
+    def from_pretrained(
+        cls,
+        model_name_or_path: str,
+        device: Optional[Union[str, torch.device]] = None,
+        torch_dtype: Optional[torch.dtype] = torch.float16,
```
karljang, line 110: Since CPU inference is one of the intended use cases, should the default dtype be float32 when the resolved device is CPU? The tests explicitly pass float32, so I am not sure the public default path using float16 on CPU has been validated.

tensorrt_llm/scaffolding/pytorch_worker.py
```diff
@@ -0,0 +1,482 @@
...
+
+            with torch.no_grad():
+                outputs = self.model.generate(**inputs, **generate_kwargs)
+
+            generated_tokens = outputs.sequences[0, input_length:].tolist()
+            output_text = self.tokenizer.decode(generated_tokens, skip_special_tokens=True)
+            output_text = self._strip_stop_sequence(output_text, task)
+
+            task.output_str = output_text
+            task.output_tokens = generated_tokens
```
karljang, line 325: It looks like the stop sequence is removed from output_str, but output_tokens and logprobs still contain the original generated sequence. Would that leave the returned text and token IDs inconsistent? Could we truncate them together and add a small consistency check in the test?

PR #12297, opened 2026-03-17
[TRTLLM-11513][fix] fix iterator undefined behavior in WindowBlockManager::getFreeBlock offload path

thorjohnsen:
## Summary

- In `WindowBlockManager::getFreeBlock`, the KV-cache offload path called `swapMemoryPoolBlockOffset()` before calling `claimBlock()` on either block. After the swap, `getCacheLevel()` returns the post-swap cache level, but `mFreeBlockIterators[id]` still points into the pre-swap queue. `claimBlock()` then calls `.erase()` on the wrong `std::list` — undefined behaviour per C++17 §26.3.10.4.
- In practice this silently corrupts `mNumFreeBlocksPerLevel` and the free-list structure in `LRUEvictionPolicy`, causing block counts to diverge from reality and eventually producing "No free block found" aborts.
- Fix: claim both `block` (primary) and `offloadBlock` (secondary) **before** `swapMemoryPoolBlockOffset()`, so each `claimBlock()` call sees the cache level that matches the iterator. This is identical to the ordering already used correctly by `WindowBlockManager::offloadBlock()` (line 1121).

## Root cause (detailed)

`LRUEvictionPolicy::claimBlock()` (`evictionPolicy.cpp`):
```cpp
SizeType32 const cacheLevel = getCacheLevel(block);  // reads isPrimary() — reflects post-swap level
mFreeQueues[cacheLevel][getPriorityIdx(...)].erase(*mFreeBlockIterators[id]);  // iterator is from pre-swap level
```

Both `block` and `offloadBlock` are obtained from the eviction policy's free queues and therefore have valid `mFreeBlockIterators`. After the swap the iterator/level mismatch affects both of them.

## API cleanup

I have also updated the API so that it no longer matters whether claimBlock is called before or after swapMemoryPoolBlockOffset. With the cleaned up API, the above fix is no longer necessary, but I think it is the correct way of doing things. We should claim the blocks before changing their state with swapMemoryPoolBlockOffset.

## Test plan

- [x] Enable KV-cache offloading (`--kv_cache_offload_gpu_to_cpu_frac` or equivalent) and run multi-request inference — previously this path would eventually corrupt the free-list counters leading to a crash or incorrect "no free blocks" error.
- [x] Run existing `tests/unittest/batch_manager/test_kv_cache_manager.*` suite.
- [x] Verify with ASan/UBSan that no `std::list` iterator misuse is reported in `LRUEvictionPolicy::claimBlock`.

## Summary by CodeRabbit

* **Refactor**
  * Improved KV cache memory block management efficiency during offloading and swapping operations, enhancing resource allocation and ownership tracking.


nvpohanh:
@eopXD could you review this? thanks

eopXD (changes requested):
Looks good, but some test coverage can help illustrate the change and extra security.

eopXD (approved):

nvpohanh (approved):

thorjohnsen:
Closing: this fix landed on main independently via #12834 (commit ca41d8f7f5), which carries the identical change (claim both blocks before swapMemoryPoolBlockOffset) with co-authorship. Nothing in this PR remains unmerged.

PR #12306, opened 2026-03-18
[None][fix] KVCacheManagerV2 bug fixes (V2 remains default OFF)

yizhang-nv:
@coderabbitai summary

## Description

Port comprehensive bug fixes for KVCacheManagerV2 while keeping V2 **default OFF** (`use_kv_cache_manager_v2=False`). The fallback mechanism now warns instead of silently switching to V1.

### Fixes

**KV Cache Core**
- **Partial block rebase corruption**: Fixed condition in `_kv_cache.py` where a full tree block could corrupt shared pages during partial rebase (added `is_full` guard).
- **`get_batch_cache_indices` default `layer_id`**: Made `layer_id` optional (default pool 0) to avoid mandatory argument when layer is irrelevant.
- **`max_blocks_per_seq` undercount**: Computed after `max_seq_len` clamping and now accounts for `num_extra_kv_tokens + max_total_draft_tokens` so the host page-index buffer is large enough.
- **Multimodal block reuse**: Different images/videos sharing the same placeholder token ID were incorrectly matched in the radix tree. Added `_augment_tokens_for_block_reuse` using content digest (Blake3 hash) for distinct tree entries.

**KV Cache Manager**
- **Auto-provision host cache tier**: V2 `MAX_UTILIZATION` scheduler relies on suspend/resume; without a host tier, suspended pages have nowhere to go, causing deadlock. Now auto-provisions host tier matching GPU quota (capped at 50% available host memory).
- **`extend_capacity_for_tokens`**: Added to both `KVCacheManager` and `KVCacheManagerV2` for CUDA graph padding token capacity extension.
- **Float type in token estimation**: Wrapped `max_num_tokens_in_memory` computation with `int()` to prevent float propagation.
- **Draft KV cache stream**: Fixed `resume()` to use `draft_kv_cache_manager._stream` instead of `torch.cuda.current_stream()`.
- **Draft resource release**: Fixed `release_resources` to also free draft KV cache on context scheduling failure.
- **Context update resize**: Simplified resize to use `None` capacity for context phase, removing fragile draft OOM workaround.

**Scheduler V2**
- **Two-phase scheduling**: Context requests are now deferred to phase 2 so generation requests are fully budgeted first, preventing PEFT adapter eviction failures.
- **PEFT pre-claim for `GENERATION_TO_COMPLETE`**: Pre-claims PEFT pages for requests whose adapters haven't been released yet (overlap executor timing).
- **Deadlock detection**: Raises `RuntimeError` when generation requests exist but none can be scheduled or evicted (KV cache exhausted with no host tier).

**CUDA Graph**
- **Prevent on-the-fly capture**: Added `_capture_allowed` flag and `allow_capture()` context manager to prevent workspace tensor reallocation from invalidating existing graphs.
- **Disable graph replay during general warmup**: Avoids replaying graphs with stale KV cache block offsets from capture time.
- **Eager fallback for uncaptured keys**: Falls back to eager execution when a key was not captured during warmup instead of crashing.

**Speculative Decoding**
- **CUDA graph padding KV cache rewind mismatch**: `pad_draft_tokens_for_cuda_graph` now extends KV cache capacity to match padded length, preventing rewind underflow. Implemented in `drafter.py`, `model_drafter.py` (two-model), and `ngram.py`.

**Fallback**
- **Warn-only fallback**: Refactored V2 incompatibility check (beam search, kv_connector, event buffer, cache transceiver) to warn instead of silently switching to V1.

## Test Coverage

- Existing unit tests in `tests/unittest/`
- Integration tests for multimodal models
- CI pre-merge pipeline

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -134,8 +134,9 @@ class KvCacheCreator:
                         self._cache_transceiver_config is not None
                         and self._cache_transceiver_config.backend is not None):
                 logger.warning(
-                    "KVCacheManagerV2 is not supported with kv_connector_manager or beam width > 1 or event buffer max size > 0."
-                )
+                    "KVCacheManagerV2 is not supported with kv_connector_manager or beam width > 1 or event buffer max size > 0. "
+                    "Falling back to KVCacheManager.")
+                cls = KVCacheManager
         return cls
 
     def _get_kv_size_per_token(self):
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -118,22 +118,25 @@ class KvCacheCreator:
         self._net_max_seq_len = net_max_seq_len
         self._dummy_reqs = None
         self._profiling_stage_data = profiling_stage_data
-        self._kv_cache_manager_cls = get_kv_cache_manager_cls(
-            model_engine.model.model_config, kv_cache_config)
         self._execution_stream = execution_stream
-        if self._kv_cache_manager_cls == KVCacheManagerV2:
-            if kv_connector_manager is not None or (
-                    max_beam_width is not None and max_beam_width
-                    > 1) or self._kv_cache_config.event_buffer_max_size > 0:
-                logger.warning(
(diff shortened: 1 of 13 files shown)


nvpohanh:
@lowsfer could you review this? thanks

lowsfer (approved):

mikeiovine (approved):

PR #12358, opened 2026-03-19
[#12324][fix] Fixed wrong token suppressed with ignore_eos in torch sampler

MrGeva:
The model was hitting EOS before 1024 tokens. min_tokens was supposed to prevent that but didn't work due to two bugs:

  1. Fixed  in [PR already](https://github.com/NVIDIA/TensorRT-LLM/pull/12166): Prompt length counted toward min_tokens: The check compared min_tokens against total tokens (prompt + generated). With ISL=1024 and min_tokens=1024, the condition was already satisfied before the first
   generated token — so EOS was never suppressed.
  2. Wrong token suppressed with ignore_eos: ignore_eos sets end_id to -1. The penalty used that value, so logits[..., -1] suppressed the last vocab token (a random Chinese character) instead of the actual
  EOS token (id 11).

  The model generated EOS early for some requests, producing fewer than 1024 output tokens. aiperf 0.6.0 added detection for this and reported the mismatch warnings. The bug was always there — aiperf just started catching it.

Why did The model generated EOS early for some requests?

Because the model decided it was done answering. The prompts are synthetic Shakespeare text — at some point the model produces a natural completion and emits the EOS token. That's normal model behavior. min_tokens exists precisely to override that for benchmarking, but it wasn't working.

TLDR:
  What is the Sampler?

  When an LLM generates text, it produces logits — a score for every token in the vocabulary — at each step.
   The sampler picks the next token from those scores. Before picking, it can apply penalties that
  manipulate logits to enforce constraints like "generate at least N tokens."

  What is min_tokens (min_length)?

  A user-facing parameter: "I want at least N generated tokens (not counting the prompt)." The sampler
  enforces this by setting the EOS (end-of-sequence) token's logit to -infinity so the model can never pick
  "stop" until it has generated enough tokens.

  The Request Flow (and where the bugs lived)

  When a user sends a request, it passes through three layers before the sampler sees it:

  1. BaseWorker (base_worker.py) — creates an ExecutorRequest, sets end_id. If ignore_eos=True, it
  overwrites end_id to -1 (meaning "no EOS").
  2. LlmRequest (llm_request.py) — converts the executor request into the internal request object. Copies
  end_id → py_end_id.
  3. TorchSampler (sampler.py) — at each generation step, calls _apply_min_length_penalty() to suppress EOS
  if the request hasn't generated enough tokens yet.

  The Two Bugs

  Bug 1 — Prompt length included in comparison:

  The code compared min_tokens against total sequence length (prompt + generated), not just generated
  tokens. With a prompt of 1024 tokens and min_tokens=1024, the check 1024 < 1024 was immediately False, so
  the penalty was never applied and the model could stop after generating just 1 token.

  Bug 2 — Wrong token suppressed with ignore_eos:

  When ignore_eos=True, end_id gets overwritten to -1. The penalty code did logits[..., py_end_id] which
  became logits[..., -1] — Python negative indexing! This suppressed the last vocabulary token (some random
  token) instead of the actual EOS token.

In more details:
_apply_min_length_penalty() in sampler.py had two bugs that made min_tokens ineffective:

  1. Prompt length included in comparison: The guard condition compared min_tokens against total sequence length (prompt + generated tokens) instead of just generated tokens. When ISL >= min_tokens, the
  penalty was never applied — e.g., with ISL=1024 and min_tokens=1024, the check 1024 < 1024 was always False.
  2. Wrong token suppressed with ignore_eos: When ignore_eos=true, end_id is overwritten to -1. The penalty used this overridden value, so logits[..., -1] suppressed the last token in the vocabulary (Python negative indexing) instead of the actual EOS token.

  Fix: Three files changed:

  - sampler.py: Subtract py_orig_prompt_len from token counts so the comparison is against generated tokens only. Use py_original_end_id (the real EOS token) instead of py_end_id (which may be -1).
  - base_worker.py: Stash the original end_id on the executor request before the ignore_eos override.
  - llm_request.py: Propagate py_original_end_id to the LlmRequest so the sampler can access it.

  Testing: Verified with trtllm-serve + aiperf benchmarks on Nemotron-3-Nano-30B (AutoDeploy backend, ISL=1024, OSL=1024, concurrency 16/32/64). Before: 5-10% of requests terminated early with OSL mismatch
  warnings. After: zero mismatches across all concurrency levels.

Key Takeaway

  The fix ensures that min_tokens actually means "generate at least N tokens" regardless of prompt length,
  and that the correct EOS token is suppressed even when ignore_eos is enabled. Without the fix, 5–10% of
  requests terminated early; after the fix, zero early terminations.

sequence diagram:

Untitled diagram-2026-03-22-163510

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected minimum length penalty calculation to properly account for prompt length during token generation.
  * Fixed end-token handling in sampling logic to preserve and use the original token configuration when applying penalties.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/llm_request.py
+++ b/tensorrt_llm/_torch/pyexecutor/llm_request.py
@@ -1033,6 +1033,9 @@ def executor_request_to_llm_request(
                               LogprobMode.RAW),
     )
 
+    llm_request.py_original_end_id = getattr(executor_request,
+                                             "py_original_end_id",
+                                             llm_request.py_end_id)
     llm_request.py_disaggregated_params = getattr(executor_request,
                                                   "py_disaggregated_params",
                                                   None)
--- a/tensorrt_llm/_torch/pyexecutor/sampler.py
+++ b/tensorrt_llm/_torch/pyexecutor/sampler.py
@@ -3932,19 +3932,25 @@ class TorchSampler(Sampler[SampleStateTorch], AsyncWorkerMixin):
             current_offset = 0
             for index, r in enumerate(requests):
                 if r.py_min_length:
-                    for beam_idx in range(num_beams[index]):
-                        for step in range(num_steps[index]):
-                            if (
-                                r.get_num_tokens(beam_idx) - r.py_orig_prompt_len
-                            ) + step < r.py_min_length[0]:
-                                # NOTE(jthomson04): We can NOT just assign logits[...] = float("-inf").
-                                # This introduces a pageable HtoD transfer, which wreaks havoc on TPOT (up to ~20%)
-                                # Instead, we create a little tensor on device, then assign to that.
-                                # This way, we avoid the pageable transfer.
-                                neg_inf_tensor = torch.full((), float("-inf"), device=logits.device)
-                                logits[
-                                    current_offset + num_steps[index] * beam_idx + step, r.py_end_id
-                                ] = neg_inf_tensor
+                    # Use the original end_id (before ignore_eos override)
+                    # so we suppress the real EOS token, not token -1.
+                    end_id = getattr(r, "py_original_end_id", r.py_end_id)
+                    if end_id is not None and end_id > -1:
+                        for beam_idx in range(num_beams[index]):
+                            for step in range(num_steps[index]):
+                                if (
+                                    r.get_num_tokens(beam_idx) - r.py_orig_prompt_len
+                                ) + step < r.py_min_length[0]:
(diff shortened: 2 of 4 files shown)


JunyiXu-nv (approved):
LGTM

Funatiq (approved):

stnie (approved):

venkywonka (approved):

MrGeva:

MrGeva:

MrGeva:

nvchenghaoz (approved):

PR #12471, opened 2026-03-24
[None][fix] Add bounded timeout to gen-side KV transfer in Python transceiver

yifjiang:
## Summary
- `RxSession.wait_complete` calls `future.result()` without a timeout, causing the decode worker to hang indefinitely when a KV transfer stalls (e.g. cancelled request, prefill failure). The context (send) path already passes `_timeout_s` to `future.result()` and returns `WaitResult.TIMEOUT` — the receive path does not.
- This PR applies the same timeout pattern to `RxSession.wait_complete` and handles `WaitResult.TIMEOUT` in `check_gen_transfer_status` by keeping the request for retry on the next iteration.

### Root cause
In disaggregated serving, `check_gen_transfer_status` is called with `at_least_request_num=None` when all active decode requests are waiting for KV cache transfer. `RxSession.wait_complete` calls `task.future.result()` without passing `self._timeout_s`, so it blocks indefinitely if the transfer stalls. The Python-level `kv_transfer_timeout_ms` check never runs because the executor loop is stuck in the blocking wait.

The context (Tx) path already handles this correctly — `TxSession.wait_complete` passes `timeout=self._timeout_s` and returns `WaitResult.TIMEOUT`, and `check_context_transfer_status` logs a warning and keeps timed-out requests for retry.

### Changes

| File | Change |
|------|--------|
| `tensorrt_llm/_torch/disaggregation/native/transfer.py` | Pass `self._timeout_s` to `future.result()` in `RxSession.wait_complete`; return `WaitResult.TIMEOUT` instead of `WaitResult.FAILED` on `TimeoutError` |
| `tensorrt_llm/_torch/disaggregation/transceiver.py` | Handle `WaitResult.TIMEOUT` in `check_gen_transfer_status` — log warning and keep request for retry |
| `tests/unittest/disaggregated/test_check_gen_transfer_status.py` | 10 unit tests covering completed, timed-out, failed, and mixed transfer scenarios |

## Test plan
- [x] Unit tests for `check_gen_transfer_status` timeout/retry behavior (10 tests in `test_check_gen_transfer_status.py`)
- [x] Existing disaggregated serving unit tests pass (`tests/unittest/disaggregated/test_py_cache_transceiver_mp.py`)
- [x] Verify with disaggregated serving integration test that decode workers no longer hang when prefill-side transfers stall

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12476, opened 2026-03-24
[None][fix] Add bounded timeout to gen-side KV transfer in C++ CacheTransceiver

yifjiang:
## Summary
- `CacheTransceiver::checkGenTransferStatus` and `checkContextTransferStatus` call `future.get()` without a timeout when `blockAll=true`, causing an unbounded block when a KV transfer never completes. This leads to decode/prefill worker hangs in disaggregated serving.
- The fix replaces the unbounded `future.get()` with `future.wait_for(timeout)` in **both** context (send) and generation (receive) paths, unconditionally — regardless of whether `blockAll` is true or false.
- On timeout, a warning is logged and the request is skipped for retry on the next scheduler iteration, keeping the executor loop alive.
- Default timeout is 1000ms when `kv_transfer_sender_future_timeout_ms` is not configured.

### Root cause
In disaggregated serving, both `checkContextTransferStatus` and `checkGenTransferStatus` have a `blockAll` flag (set when `atLeastRequestNum` is `nullopt`). When `blockAll=true`, the original code skipped the timeout configuration entirely:

```cpp
// BEFORE (broken): timeout only applied when !blockAll
if (!blockAll && mCacheTransceiverConfig.has_value())
    receiverFutureTimeoutMs = mCacheTransceiverConfig->getKvTransferSenderFutureTimeoutMs();
// ...
// When receiverFutureTimeoutMs is nullopt, this falls through to future.get()
auto status = it->second.wait_for(std::chrono::milliseconds(receiverFutureTimeoutMs.value_or(0)));
if (status == std::future_status::ready || !receiverFutureTimeoutMs.has_value())
    it->second.get();  // BLOCKS FOREVER when transfer is stalled
```

The `!receiverFutureTimeoutMs.has_value()` condition meant that when the timeout was not set (blockAll path), the code always fell through to `future.get()` regardless of the `wait_for` result — making the `wait_for` call a no-op.

The Python-level `kv_transfer_timeout_ms` cannot help because the executor loop is stuck inside this C++ blocking call and never reaches the Python timeout check.

### How the timeout + retry works

The scheduler calls `checkGenTransferStatus` repeatedly in its main loop. With this fix:

1. `wait_for(timeoutMs)` checks if the KV transfer future is ready within the timeout
2. If **ready**: call `.get()`, mark transfer complete, proceed normally
3. If **timeout**: log a warning, `++it` (skip this future), return to the scheduler
4. The scheduler loops back and calls `checkGenTransferStatus` again on the next iteration
5. By then, the async NIXL/UCX transfer has likely completed, so `wait_for` returns `ready`

This means **requests still complete successfully even when timeouts fire** — the timeout just prevents the thread from blocking indefinitely on a single future. In E2E testing with a 1ms timeout (far shorter than actual KV transfer time), every request triggered multiple timeout warnings but all completed with full token output, because the transfer finished between scheduler iterations.

If a transfer is **genuinely stalled** (not just slow), the C++ timeout keeps the scheduler loop alive so the Python-level `kv_transfer_timeout_ms` can eventually cancel the request — something that was impossible before when `future.get()` blocked the thread.

### Changes
- **Remove `!blockAll` guard**: Always read `kv_transfer_sender_future_timeout_ms` from config, regardless of `blockAll`
- **Fix both paths**: Applied consistently to `checkContextTransferStatus` (send/prefill) and `checkGenTransferStatus` (receive/decode)
- **Use `value_or(1000)` default**: Ensures a bounded wait even when no config is set (previously `value_or(0)` which returned immediately and fell through to `.get()`)
- **Check `future_status::ready` directly**: Remove `!receiverFutureTimeoutMs.has_value()` fallthrough that bypassed the timeout
- On timeout: log warning, skip request (`++it`), retry next iteration
- On unexpected status: mark `kDISAGG_TRANS_ERROR`, erase future
- On exception: mark `kDISAGG_TRANS_ERROR`, erase future

## Test plan
- [x] Existing disaggregated serving unit tests pass
- [x] Verify decode workers no longer hang when prefill-side transfers stall
- [x] Verify timed-out gen transfers are retried on subsequent iterations — confirmed by setting `kv_transfer_sender_future_timeout_ms: 1` in config: decode worker logs show `[TensorRT-LLM][WARNING] Timed out waiting for generation KV cache transfer after 1 milliseconds.` and requests still complete successfully (transfers complete between scheduler iterations)
- [x] Companion Python transceiver fix: #12471 (tracked separately)

### E2E test results
- Built custom dynamo-trtllm image from source with this PR's TRT-LLM wheel
- Tested on GB200 NVL nodes (ARM64, 2-node disagg 1P:1D, NIXL + NVSwitch)
- Model: Qwen3-4B, TP=1

| Test | Result |
|---|---|
| Binary verification (`strings libtensorrt_llm.so`) | Timeout warning string present in compiled binary |
| With `kv_transfer_sender_future_timeout_ms: 1` | Timeout warnings fire on every request; all 5 requests still complete (100 tokens each) |
| Same 1ms setting on **previous** code (with `!blockAll` guard) | Zero timeout warnings — timeout was silently skipped |
| With default timeout (1000ms) | 5/5 basic requests pass, 20/20 concurrent stress test passes, no warnings, no hangs |


cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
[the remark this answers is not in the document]
yifjiang: Repro is documented in ~~https://docs.google.com/document/d/1GAOMcc2_C6X-DcATe00hX0e1qGvlISpWnC41YnJY3TY/edit?tab=t.0~~    https://docs.google.com/document/d/1ZfCgd-gZuSsc9GR4F7Ral_2PkmlVWwuym-ndnJr_Gwk/edit?tab=t.myfct0z2ax7v#heading=h.2jld0b32shal

pcastonguay (approved):

nv-yna:
Can we make the default value 10000ms? 1000ms seems like a value where it can easily pass under load, very long context, bad network, etc.

pcastonguay:
> Can we make the default value 10000ms? 1000ms seems like a value where it can easily pass under load, very long context, bad network, etc.

The 1000ms timeout value won't cause the request to be cancelled. It simply means that we will continue with the next iteration (to schedule new requests if needed for example), and we will check again on transfer status at the next iteration. The timeout value for requests that fail to complete the KV cache transfer is 60s. See https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/llmapi/llm_args.py#L2463-L2473

chuangz0 (commented):
Have you encountered a bug caused by this issue? I searched our codebase and found that we don’t use `checkContextTransferStatus(None)` or `checkGenTransferStatus(None)`. So, if you’re experiencing hang, it’s likely unrelated to the `blockAll` behavior.

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
```diff
@@ -551,8 +552,9 @@ RequestStatuses CacheTransceiver::checkContextTransferStatus(
             try
             {
                 // Wait for up to a specified timeout
-                auto status = future.wait_for(std::chrono::milliseconds(senderFutureTimeoutMs.value_or(0)));
-                if (status == std::future_status::ready || !senderFutureTimeoutMs.has_value())
+                auto const timeoutMs = senderFutureTimeoutMs.value_or(1000);
```
chuangz0, line 555: The `sender_future_time_out` parameter has a default value in `cache_transceiver_config`. If a user explicitly sets it to `none` intending for the operation to be blocking, but the actual behavior turns out to be 1000 ms, that would be  confusing.

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
```diff
@@ -599,6 +601,14 @@ RequestStatuses CacheTransceiver::checkContextTransferStatus(
 void CacheTransceiver::checkGenTransferStatus(std::optional<int> const& atLeastRequestNum)
 {
     bool blockAll = !atLeastRequestNum.has_value();
+    std::optional<int> receiverFutureTimeoutMs = std::nullopt;
+    // Always use a bounded timeout to prevent unbounded blocking.
+    // The caller (scheduler) loops, so timed-out transfers retry on next iteration.
+    if (mCacheTransceiverConfig.has_value())
+    {
+        receiverFutureTimeoutMs = mCacheTransceiverConfig->getKvTransferSenderFutureTimeoutMs();
```
chuangz0, line 609: Both the sender and receiver use the configured `SenderFutureTimeout`. We may need to rename `SenderFutureTimeout` to something like `kv_transfer_time_out_per_iteration` to make it easier to understand.

pcastonguay: I agree. Let's do this in a separate PR though.

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
pcastonguay: I think tha's a valid comment. @yifjiang could you address please? Thx.

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
```diff
@@ -485,8 +485,9 @@ RequestStatuses CacheTransceiver::checkContextTransferStatus(
 {
     bool blockAll = !atLeastRequestNum.has_value();
     std::optional<int> senderFutureTimeoutMs = std::nullopt;
-    // If blockAll is true, we want to block and not use a timeout
-    if (!blockAll && mCacheTransceiverConfig.has_value())
+    // Always use a bounded timeout to prevent unbounded blocking.
+    // The caller (scheduler) loops, so timed-out transfers retry on next iteration.
+    if (mCacheTransceiverConfig.has_value())
```
pcastonguay, line 490: I'm not sure we should change this. If the caller explicitly sets `atLeastRequestNum` to None, it expects this function to be blocking. From what I can tell, in `py_executor.py`, `atLeastRequestNum` is always set.

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
```diff
@@ -599,6 +601,14 @@ RequestStatuses CacheTransceiver::checkContextTransferStatus(
 void CacheTransceiver::checkGenTransferStatus(std::optional<int> const& atLeastRequestNum)
 {
     bool blockAll = !atLeastRequestNum.has_value();
+    std::optional<int> receiverFutureTimeoutMs = std::nullopt;
+    // Always use a bounded timeout to prevent unbounded blocking.
+    // The caller (scheduler) loops, so timed-out transfers retry on next iteration.
+    if (mCacheTransceiverConfig.has_value())
```
pcastonguay, line 607: I believe the logic should match the logic in `checkContextTransferStatus` and respect the `blockAll` parameter.

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
```diff
@@ -709,41 +719,68 @@ void CacheTransceiver::checkGenTransferStatus(std::optional<int> const& atLeastR
...
             {
-                it->second.get();
-                it->first->setState(LlmRequestState::kDISAGG_GENERATION_TRANS_COMPLETE);
-
-                // Gather the kv cache transfer time from all workers and update to leader rank
-                if (!common::getEnvKVCacheTimeOutputPath().empty())
+                // Wait for up to a specified timeout
+                auto const timeoutMs = receiverFutureTimeoutMs.value_or(1000);
+                auto status = it->second.wait_for(std::chrono::milliseconds(timeoutMs));
+                if (status == std::future_status::ready)
```
pcastonguay, line 732: Similarly here, if `receiverFutureTimeoutMs` is None, user would expect to block, so I don't think we should set to 1000ms. Can we match the logic from `checkContextTransferStatus`? In `py_executor.py`, we would always have a default value of 1000ms from the cache_transceiver config so we wouldn't block.

pcastonguay:
I opened https://github.com/NVIDIA/TensorRT-LLM/pull/13083 with the changes that I would merge instead. @yifjiang the behavior should be identical to your PR since the cache transceiver config has a default value of 1000ms, and we always set `atLeastRequestNum` in `py_executor.py`.

pcastonguay:
@yifjiang @chienchunhung should this be closed now?

chienchunhung:
> Member

The inflight #15356 covers the bounded polling logic. We can close this one to avoid duplication.

pcastonguay:
Closing, addressed in other PRs

PR #12491, opened 2026-03-24
[TRTLLM-9644][infra] Update isolation test

EmmaQiaoCh:
## Dev Engineer Review

* **Refactor**
  * Unified regular, isolated, sharded, rerun, and result-merging logic in `run_tests.py`.
  * Updated Groovy and Slurm workflows to use the unified runner.
  * Added rerun XML transfer and report upload handling.
  * Preserved multi-node performance checks and Slurm argument handling.

* **Follow-up required**
  * Fail when `pytest --collect-only` returns a non-zero exit code, even when collection produces output.
  * Ensure rerun report upload and timeout-result generation run when rerun tests fail.
  * Enforce the `--max-rerun-tests` limit.
  * Avoid single-quoted shell arguments for commands or failure signatures that contain apostrophes.
  * Account for the approximately 450 minutes of additional GB200 multi-node CI time from five isolation performance-sanity cases.

## QA Engineer Review

* Modified `tests/integration/test_lists/test-db/l0_gb200_multi_nodes_perf_sanity_ctx1_node1_gpu1_gen1_node1_gpu2.yml`.
* Added five GB200 multi-node isolation performance-sanity entries for GPT-OSS and Qwen3 configurations.
* CBTS coverage data is unavailable.

**Verdict: needs follow-up.**

## PR Checklist

## GitHub Bot Help


EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

tests/integration/defs/accuracy/test_llm_api_pytorch.py
```diff
@@ -2675,6 +2675,10 @@ def test_nvfp4_multi_gpus(self, tp_size, pp_size, ep_size, mtp_nextn, fp8kv,
                               attention_dp, enable_lm_head_tp_in_adp,
                               cuda_graph, overlap_scheduler, max_batch_size,
                               moe_backend):
+        # TODO: Remove this forced failure — injected to test isolation rerun logic in CI
+        if enable_lm_head_tp_in_adp and ep_size == 4 and attention_dp and moe_backend == "CUTLASS":
+            assert False, "Intentional failure to test isolation rerun logic"
```
mzweilz: Is this for testing? Leave a comment here as we should remove this before merge.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,817 @@
...
+
+    Returns:
+        Modified pytest command string.
+    """
+    # Remove args that shouldn't be in rerun commands
+    no_need_patterns = ["--splitting-algorithm", "--splits", "--group", "--cov"]
+    need_to_change_patterns = ["--test-list", "--csv", "--periodic-junit-xmlpath"]
+
+    parts = []
+    tokens = base_cmd.split()
```
mzweilz: Just a heads up - This breaks if any argument values contain spaces (e.g., paths, quoted strings). Consider using shlex.split(base_cmd) instead.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,817 @@
...
+        return f"{h}h {m}m {s}s"
+    if m > 0:
+        return f"{m}m {s}s"
+    return f"{s}s"
+
+
+# ---------------------------------------------------------------------------
+# render_test_list: replaces Groovy processShardTestList()
+# ---------------------------------------------------------------------------
+def render_test_list(test_db_list, working_dir, splits, group, perf_mode):
```
mzweilz: `render_test_list` is doing too much. ~130 lines mixing 4 concerns. Consider spliting this into small functions, for example
```
- parse_isolation_markers(lines) -> (cleaned_lines, isolation_set) — pure data transformation, easy to unit test
- collect_shard_tests(cleaned_file, splits, group, working_dir) -> list[str] — isolates the subprocess call and output parsing
- classify_tests(shard_tests, isolation_set, cleaned_lines) -> (regular, isolate) — matching logic in one place, easier to fix the substring matching issue
```
Each becomes independently testable and the main render_test_list reduces to a ~20-line orchestrator calling these three steps.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,817 @@
...
+
+        if rc != 0:
+            if not os.path.exists(xml_file):
+                print(f"The {rerun_tag} tests crashed during rerun attempt (no XML produced).")
+                raise RuntimeError(f"Rerun crashed for {rerun_tag}, no XML produced")
+            print(f"The {rerun_tag} tests still failed after rerun attempt.")
+            is_rerun_failed = True
+
+    print(f"is_rerun_failed for {rerun_tag}: {is_rerun_failed}")
+    return is_rerun_failed, rerun_xml_files
```
mzweilz, line 539: The return value of the `check_and_rerun` function is confusing. it returns True to mean "failed" in some paths but the semantics are inconsistent. When valid_count == 0 (no tests to rerun), it returns True (line 635), same as when there are too many failures (line 632). But "all tests passed and nothing needs rerun" is a different situation than "tests failed but can't be rerun." Consider returning an enum or named status instead of a bare bool.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,817 @@
...
+    """
+    print(f"Running pytest: {pytest_cmd}")
+    t0 = time.monotonic()
+    result = subprocess.run(pytest_cmd, shell=True, cwd=working_dir)
+    elapsed = time.monotonic() - t0
+    print(f"Pytest finished with exit code {result.returncode} ({format_duration(elapsed)})")
+    return result.returncode, elapsed
+
+
+def build_rerun_command(base_cmd, test_list, xml_path, csv_path, reruns):
```
mzweilz: build_rerun_command and build_isolated_command are near-duplicates. Both do the same token-stripping loop, just with different pattern lists. Could be a single `rebuild_pytest_command(base_cmd, remove_patterns, replace_args)`.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,817 @@
...
+        collect_output_dir = os.path.join(os.path.dirname(test_db_list), "collect_output")
+        os.makedirs(collect_output_dir, exist_ok=True)
+        collect_cmd = (
+            f"pytest --collect-only --splitting-algorithm least_duration "
+            f"--test-list={cleaned_file} --quiet "
+            f"--splits {splits} --group {group} "
+            f"--output-dir={collect_output_dir}"
+        )
+        print(f"Running: {collect_cmd}")
+        result = subprocess.run(
```
mzweilz: Consider setting a timeout on `pytest --collect-only`. `subprocess.run `with `capture_output=True` and no timeout. If collection hangs (e.g., import-time deadlock with MPI), this blocks forever with no output. The actual test execution at line 493 at least streams output, but collection is silent until it finishes.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,817 @@
...
+        output_dir: Output directory for the stage.
+        stage_name: Name of the test stage.
+        all_xml_files: List of all result XML file paths.
+    """
+    # Fix testsuite names in all XMLs
+    for xml_file in all_xml_files:
+        if os.path.exists(xml_file):
+            try:
+                content = Path(xml_file).read_text()
+                content = content.replace(
```
mzweilz: This does string replacement on XML, which is a fragile XML manipulation. If the XML contains this string in a test name or output, it'll get corrupted. Using xml.etree.ElementTree would be safer.

jenkins/scripts/slurm_run.sh
```diff
@@ -109,28 +109,41 @@ pytest_exit_code=0
...
-    echo "DEBUG: Check importability of conftest.py"
-    python3 -c "import sys; sys.path.insert(0, '.'); import conftest; print('DEBUG: conftest imported successfully')"
+# Build the run_tests.py arguments array for proper quoting
+runTestsArgs=(
+    python3 "$runTestsScript"
+    --render
+    --test-db-list "$testListPathNode"
+    --splits "${testSplits:-1}"
+    --group "${testGroup:-1}"
+    $perfModeFlag
```
mzweilz: $perfModeFlag is intentionally unquoted so it vanishes when empty, but this relies on bash word-splitting behavior inside array construction. Using a conditional append pattern would be more explicit:
```
  runTestsArgs=(...)
  if [ "$perfMode" = "true" ]; then
      runTestsArgs+=(--perf-mode)
  fi
```

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,817 @@
...
+            print(f"No failed {rerun_tag} tests need to be rerun {times} time(s)")
+            continue
+
+        print(f"Rerun test list ({times}):")
+        print(Path(rerun_list).read_text())
+
+        xml_file = os.path.join(rerun_dir, f"rerun_results_{times}.xml")
+        csv_file = os.path.join(rerun_dir, f"rerun_report_{times}.csv")
+
+        rerun_cmd = build_rerun_command(base_cmd, rerun_list, xml_file, csv_file, times - 1)
```
mzweilz, line 526: When `times=1`, this passes reruns=0 to pytest, which means no rerun flag is added. This is correct if the intent is "just run the test again once," but it might be confusing naming. The rerun_1.txt file implies 1 rerun, but `--reruns 0` means pytest won't retry internally. We should verify this matches the old Groovy behavior.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,817 @@
...
+            isolation_match = next((t for t in isolation_tests if trimmed in t), None)
+            if isolation_match:
+                isolate_tests.append(isolation_match)
+            else:
+                cleaned_match = next((t for t in cleaned_lines if trimmed in t), None)
+                if cleaned_match:
+                    regular_tests.append(cleaned_match)
+
+    # Write regular and isolate list files
+    regular_file = test_db_list.replace(".txt", "_regular.txt")
```
mzweilz: If the file path doesn't end in .txt, this silently produces a wrong path. A safer pattern: `Path(test_db_list).with_stem(Path(test_db_list).stem + "_regular")`. The same for the `_cleaned.txt`.

StanleySun639 (approved):

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

crazydemo (approved):
no conflict with cbts

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-1, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-2, GB200-8_GPUs-2_Nodes-PyTorch-Post-Merge-3, A100X-PyTorch-1" --disable-fail-fast

jenkins/scripts/run_tests.py
EmmaQiaoCh: Keep the most logic in this function as the old one.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,869 @@
...
+        unfinished_test_file,
+    )
+
+    # Step 2: If rerun_0.txt exists, some tests can't be rerun
+    rerun_0_file = os.path.join(rerun_dir, "rerun_0.txt")
+    if os.path.exists(rerun_0_file):
+        print(f"Contents of {rerun_0_file}:")
+        print(Path(rerun_0_file).read_text())
+        print("There are some failed tests that cannot be rerun, skip the rerun step.")
+        return True, []
```
yiqingy0: This part should be removed.

jenkins/scripts/run_tests.py
```diff
@@ -0,0 +1,869 @@
...
+            valid_count += count
+
+    if valid_count > max_rerun_tests:
+        print(
+            f"There are more than {max_rerun_tests} failed {rerun_tag} tests, skip the rerun step."
+        )
+        return True, []
+    elif valid_count == 0:
+        print(f"No failed {rerun_tag} tests need to be rerun.")
+        return True, []
```
yiqingy0: This part should be removed.

EmmaQiaoCh:
/bot run --stage-list "B200_PCIe-PackageSanityCheck-PY312-DLFW, A10-PyTorch-2" --disable-fail-fast

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

BowenFu:
Reviewed the full change (groovy delta, `slurm_run.sh`, and all 979 lines of the new `run_tests.py`). The consolidation is a good direction — one runner instead of orchestration split across Groovy try/catch — but two behaviours that exist on `main` do not survive the port.

**1. A collection error no longer fails the shard; it silently runs a partial one.**

`run_tests.py:173`
```python
if result.returncode != 0 and not output:
    print(f"Error: pytest --collect-only failed with exit code {result.returncode}")
    sys.exit(1)
```

On `main`, `processShardTestList` runs the same collect through `sh(..., returnStdout: true)`, which throws on *any* non-zero exit, and the catch is unconditional:

```groovy
} catch (Exception e) {
    error "Test collection failed for shard ${splitId}/${splits}. Cannot proceed without valid test list."
}
```

`pytest --collect-only --quiet` prints the items it did collect *before* the error summary, so a collection error in one module (import error, bad parametrization) gives exit code 2 **with non-empty stdout**. `not output` is then false, the guard is skipped, the truncated shard is parsed at :180 and executed, and the stage exits 0. That is a shard reporting PASSED with tests silently missing — the one CI failure mode with no downstream signal. Suggest `if result.returncode != 0:` and drop the `and not output`, keeping the printed output for diagnosis.

**2. On the failure path the rerun report is no longer produced or uploaded.**

`run_tests.py` exits 1 when `rerun_failed`, and the `sh` at `L0_Test.groovy:4290` is not wrapped in `try`/`catchError`, so it throws immediately. Everything after it in the stage body is skipped:

- the CBTS coverage liveness step (:4310);
- **the `rerun_results.html` upload and its printed URL (:4319-4326)**;
- the `results-timeout.xml` / `generateTimeoutTestResultXml` check (:4328).

On `main` the failure was caught inside the try/catch, so `stage("Generate Report") { generateRerunReport(...) }` and the timeout-XML check both ran before the terminal `if (rerunFailed) error`. Net effect: the rerun report is now generated only when nothing failed, which is when nobody needs it. Wrapping the `sh` in `catchError(buildResult: 'FAILURE', stageResult: 'FAILURE')`, or moving the upload into `run_tests.py` before `sys.exit(1)`, restores it.

Also worth noting the timeout case specifically: `main` distinguished "rerun passed but the first run timed out" (stage FAILURE, build SUCCESS) from a hard failure. That distinction now depends entirely on reaching :4328, which the above prevents.

**3. `--max-rerun-tests 5` is passed by both call sites but never enforced.**

`valid_count` is accumulated at `run_tests.py:494-501` and never read; `max_rerun_tests` is threaded through `main` → `run_regular_tests`/`run_isolated_tests` → `check_and_rerun` and never compared to anything. This is a faithful port — `validLineCount` on `main` is dead in exactly the same way, and the comment above it ("If the stage has more than 5 failed tests, skip the rerun step") already didn't describe the code. But the port promotes it into a documented CLI flag wired up at two call sites, which reads like a working knob. Either enforce it (`if valid_count > max_rerun_tests: skip`) or drop the flag rather than carrying the fiction forward.

**4. The whole pytest command now round-trips through single-quoted shell.**

`--pytest-base-cmd '${pytestCommand.join(" ")}'` (:4299) and `--fail-signatures '${failSignaturesList}'` (:4303) are single-quoted inside a Groovy-interpolated shell string, so one `'` anywhere in either value breaks the invocation. That constraint is why this PR has to change `--unittest-markexpr='...'` to `"..."` at :1398 — a correct fix, but it makes the quoting rule implicit and unenforced for the next argument someone adds. `getFailSignaturesList()` lives in the shared Jenkins library rather than in-tree, so I can't check its contents from the repo; if any signature contains an apostrophe this breaks today. A here-doc or passing the command via a file would remove the class of problem.

Minor: the sbatch path escapes `\` and `"` for the `export pytestCommand="..."` embedding (:1797-1800) and then re-quotes correctly in `slurm_run.sh` via the `runTestsArgs` array — that part is right, and the array form is a clear improvement over `eval $pytestCommand`.

**Scope:** the fifth file adds 5 `ISOLATION` perf-sanity cases at `TIMEOUT (90)` each to `l0_gb200_multi_nodes_perf_sanity_ctx1_node1_gpu1_gen1_node1_gpu2.yml`. Reasonable as the vehicle that actually exercises the new isolation path on multi-node, but it is up to ~450 minutes of additional GB200 multi-node time arriving inside a CI-refactor PR; worth stating in the description so it is a deliberate decision rather than a side effect.

Not blocking on 3, 4, or the scope note. 1 and 2 are behaviour that exists on `main` today and would be lost.

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

jenkins/L0_Test.groovy
```diff
@@ -1527,8 +1535,11 @@ def getPytestBaseCommandLine(
     if (stageName.contains("-Ray-")) {
         testCmdLine += ["--run-ray"]
     }
-    def unittestMarkExpr = (stageName.startsWith("CPU-")) ? "cpu_only" : "not cpu_only"
-    testCmdLine += ["--unittest-markexpr='${unittestMarkExpr}'"]
+    def unittestMarkExpr = (stageName.startsWith("CPU-")) ? "cpu_only and not disabled" : "not cpu_only"
+    testCmdLine += ["--unittest-markexpr=\"${unittestMarkExpr}\""]
+    if (ENABLE_UPLOAD_TEST_RESULTS) {
+        testCmdLine += ["-o console_output_style=progress-even-when-capture-no"]
+    }
```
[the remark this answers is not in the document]
EmmaQiaoCh: disable marker is not in use.

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

EmmaQiaoCh:
/bot run --stage-list "GB200-8_GPUs-2_Nodes-PyTorch-Disagg-PerfSanity-CTX1-NODE1-GPU1-GEN1-NODE1-GPU2-Post-Merge-1"

PR #12530, opened 2026-03-25
[https://nvbugs/5879577][fix] Fix KeyError in DeepSeekV3Lite FP8 MTP weight loading

sunnyqgg:
## Summary
- Fix `KeyError: 'model.layers.30.self_attn.kv_a_proj_with_mqa.weight'` when loading DeepSeekV3Lite FP8 with vanilla MTP (`num_nextn_predict_layers=2`)
- Root cause: `ConsumableWeightsDict.mark_consumed()` deletes checkpoint weights after the first MTP layer loads them, but when `model_nextn > ckpt_nextn`, subsequent MTP layers share the same checkpoint weights via modulo remapping and fail with KeyError
- Fix: Skip `mark_consumed` for MTP layers when checkpoint weights are shared across multiple model MTP layers
- Remove corresponding test waiver from `waives.txt`

## Test plan
- [x] Run `accuracy/test_llm_api_pytorch.py::TestDeepSeekV3Lite::test_fp8_block_scales[mtp=vanilla-fp8kv=False-attention_dp=False-cuda_graph=False-overlap_scheduler=False-torch_compile=False]` on H100
- [x] Verify no regression on other DeepSeekV3Lite FP8 tests (eagle MTP, non-MTP)
- [x] `/bot run --extra-stage "DGX_H100_PCIe-PyTorch-Post-Merge-2"`

## Summary by CodeRabbit

* **New Features**
  * Implemented MTP checkpoint-weight sharing detection for DeepSeek V3 models, optimizing weight loading during inference.

* **Tests**
  * Removed skipped test entry from DeepSeek V3 fp8 precision test suite, allowing previously excluded test variations to execute.


tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -409,7 +425,7 @@ def split_kv_b_proj(kv_b_proj: torch.Tensor,
                                 ).view(*attn_module.v_b_proj_dequant.shape).to(
                                     attn_module.v_b_proj_dequant.dtype))
                     # Mark consumed kv_b_proj weights
-                    if can_mark_consumed:
+                    if can_mark_consumed and not is_shared_mtp_layer:
```
2ez4bz: Nit: all these `if can_mark_consumed and not is_shared_mtp_layer` lines could reuse a variable

```python
mark_consumed = can_mark_consumed and not is_shared_mtp_layer
```

2ez4bz (approved):

tensorrt_llm/_torch/models/modeling_deepseekv3.py
sunnyqgg: Done, thanks a lot

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -333,6 +333,20 @@ def split_kv_b_proj(kv_b_proj: torch.Tensor,
...
+        # mark_consumed must be skipped to avoid deleting weights that later
+        # MTP layers still need.
+        ckpt_nextn = self.config.num_nextn_predict_layers or 0
+        model_nextn = 0
+        spec_config = self.model_config.spec_config
+        if spec_config is not None and hasattr(
+                spec_config, 'spec_dec_mode'
+        ) and spec_config.spec_dec_mode.is_mtp_one_model():
+            model_nextn = spec_config.num_nextn_predict_layers
+        has_shared_mtp_weights = model_nextn > ckpt_nextn > 0
```
Wanli-Jiang, line 348: can we make it as a helper function, to make the load_weight function more readable?

tensorrt_llm/_torch/models/modeling_deepseekv3.py
```diff
@@ -342,8 +356,10 @@ def split_kv_b_proj(kv_b_proj: torch.Tensor,
             else:
                 names = name.split('.')
                 parent_module_name = '.'.join(names[:-1])
+                is_shared_mtp_layer = False
                 if "model.layers" in name and int(
                         names[2]) >= self.config.num_hidden_layers:
+                    is_shared_mtp_layer = has_shared_mtp_weights
                     mtp_layer_idx = int(
                         names[2]) - self.config.num_hidden_layers
                     names[2] = str(mtp_layer_idx %
```
Wanli-Jiang, line 365: should we check `self.config.num_nextn_predict_layers` cannot be 0 somewhere?

Wanli-Jiang (approved):

tensorrt_llm/_torch/models/modeling_deepseekv3.py
sunnyqgg: Thanks for raising this. Looking at the code, `num_nextn_predict_layers = 0` is actually a valid and intentional state — the Kimi-K2.5 backbone explicitly sets it to 0 to indicate MTP is not applicable (see the note at line 1917). In that case, the `range(0)` would correctly produce an empty loop with no MTP layers constructed.

A guard of `> 0` would be wrong here; we should only add a positive check in the path where MTP is *expected* to be active (i.e., when spec decoding is actually configured). Happy to add that scoped check if it would help clarity, but a blanket "must be > 0" validation would break Kimi-K2.5.

tensorrt_llm/_torch/models/modeling_deepseekv3.py
sunnyqgg: Good suggestion , done

PR #12571, opened 2026-03-29
[None][fix] Fix DSACacheManager and RocketCacheManager KV cache estimation ignoring num_layers for draft models

lancelly:
Summary                                                                                                                                             
                                                                                                                                                    
  - DSACacheManager.get_cache_size_per_token accepted num_layers via **kwargs but never used it, causing draft KV cache estimation to use the full    
  model layer count (61) instead of actual draft layers (3) for MTP speculative decoding                                                            
  - This inflated kv_size_per_token from ~45K to ~86K bytes/token, halving max_tokens and wasting ~47% of available KV cache VRAM (~42 GiB on B200)   
  - Added explicit num_layers parameter to match parent class KVCacheManager.get_cache_size_per_token signature                                       
                                                                                     
                                                                                                                                                                                                                                            
  Only affects DSA + one-model speculative decoding (MTP/EAGLE3). Non-DSA and non-spec-decoding paths are unaffected.   

<img width="787" height="827" alt="截屏2026-03-29 21 21 10" src="https://github.com/user-attachments/assets/eea72e57-ca9e-4681-ae49-73958a8f1113" />


liji-nv (approved):

pengbowang-nv (approved):
LGTM. Just a small question: why Non-DSA and non-spec-decoding paths are unaffected by this bug? Will Non-DSA spec-decode path encounter similar issue?

mikeiovine (approved):

eopXD (approved):
Looks good.

PR #12648, opened 2026-04-01
[#12560][fix] Fix storeBlocks pinning wrong block during disagg KV cache transfer

achartier:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed reference counting for pinned KV cache blocks to ensure accurate block reuse tracking in batch operations.

## Description

storeBlocks pinned searchRoot (the matched reuse tree block) instead of block (the physical block from the sequence's block list). When a request's tokens match an existing tree entry backed by a different physical block, the pin/unpin cycle targets the wrong block:

1. storeBlocksForReuse pins the tree block (incRefCount)
2. unpinBlocksById unpins the tree block (decRefCount → 0 → releaseBlock)
3. The tree block is spuriously added to the free queue
4. If already in the queue from a prior release, it now appears twice
5. Subsequent getFreeBlock calls return the same block repeatedly, producing duplicate block IDs in the next sequence

This caused assertion failures in disaggregated serving when enable_partial_reuse=True (default) and requests share token prefixes with previously completed requests.

Fix: pin the physical block (block) instead of the tree block (searchRoot). The physical block is the one whose GPU memory holds the KV data being transferred and must stay alive during the async send.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
+++ b/cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
@@ -1688,8 +1688,8 @@ std::pair<SizeType32, std::vector<KVCacheBlock::IdType>> WindowBlockManager::sto
             }
             if (pinBlocks)
             {
-                searchRoot->incRefCount();
-                pinnedBlockIds.push_back(searchRoot->getBlockId());
+                block->incRefCount();
+                pinnedBlockIds.push_back(block->getBlockId());
             }
         }
         catch (std::out_of_range const& ex)


cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1688,8 +1688,8 @@ std::pair<SizeType32, std::vector<KVCacheBlock::IdType>> WindowBlockManager::sto
             }
             if (pinBlocks)
             {
-                searchRoot->incRefCount();
-                pinnedBlockIds.push_back(searchRoot->getBlockId());
+                block->incRefCount();
```
Tabrizian: I think we want to increase the refCount for the block that is inserted in the prefix tree. The request's block may not be inserted in the tree at all and since we have terminated the request earlier I think it won't be available for KVCache transmission. Please let me know if I'm misunderstanding.

PR #12698, opened 2026-04-02
[#12632][feat] Initial prototype for AutoDeploy compile cache

nvchenghaoz:
## Summary by CodeRabbit

* **New Features**
  * AutoDeploy now supports pipeline caching, enabling reuse of pre-weight-loading graph snapshots across multiple runs to improve build efficiency.
  * Configurable cache boundaries and storage location via `PipelineCacheConfig` settings (defaults: sharding boundary, `~/.cache/tensorrt_llm/auto_deploy_pipeline`).

* **Documentation**
  * Added design documentation for pipeline cache architecture, including cache key generation, artifact format, and validation.

* **Tests**
  * Added comprehensive unit tests for pipeline cache functionality, restore/save behavior, and edge cases.


tensorrt_llm/_torch/auto_deploy/config/default.yaml
```diff
@@ -1,5 +1,9 @@
 # This is the set of transforms running in "graph" mode. In this mode, we capture the full graph
 # of the model and optimize it for inference.
+pipeline_cache:
```
lucaslie, line 3: I would love for this to just be another transform. The way I think about it is you can insert it at the point where you check for the cache, and then we have a shared config in `interface.py`. You can use that shared config and modify that transform/interface.py file to then check that shared config and just skip over transforms. If there was a cache hit, what do you think?

tensorrt_llm/_torch/auto_deploy/config/default.yaml
lucaslie, line 26: We can deprecate this transform altogether since we are now removing the input constraint.

tensorrt_llm/_torch/auto_deploy/export/export.py
```diff
@@ -723,6 +738,11 @@ def _capture_fn(model, args, kwargs):
     # clean up checks --> generally the sanity checks are overly conservative and we can remove them
     _clean_up_assertions_and_guards(egm)
 
+    # Remove input constraint hooks added by torch.export — the AutoDeploy pipeline
+    # manages input shapes dynamically and these hooks would reject valid inputs
+    # during resize_kv_cache and other forward passes with different batch sizes.
+    _remove_export_input_constraint_hooks(egm)
```
lucaslie, line 744: It's great. It's a good time to just get rid of the input constraints alltogether

tensorrt_llm/_torch/auto_deploy/export/export.py
```diff
@@ -780,6 +800,7 @@ def export_onnx(ad_config: "LlmArgs") -> nn.Module:
     inference_optimizer = InferenceOptimizer(
         factory=factory,
         config=ad_config.transforms,
+        pipeline_cache_config=ad_config.pipeline_cache,
```
lucaslie, line 803: see my other comment: ideally, we don't need this.

tensorrt_llm/_torch/auto_deploy/models/factory.py
lucaslie, line 1: We need to at least also add `model_kwargs`, right?

tensorrt_llm/_torch/auto_deploy/transform/ad_ir.py
lucaslie, line 1: How stable is this infra? There is no freaking way we can in any way maintain or comprehend that as a human. What's wrong with just using standard off-the-shelf PyTorch export utilities?

tensorrt_llm/_torch/auto_deploy/transform/optimizer.py
lucaslie, line 1: Seems the changes here should be minimal if we just make pipeline snapshots part of the transforms.

tensorrt_llm/_torch/auto_deploy/transform/pipeline_cache.py
lucaslie, line 1: Same command as for the file ad_ir.py. We will not be able to maintain this, so we have to go full on agentic and test-driven. Are we confident?

tensorrt_llm/_torch/auto_deploy/llm_args.py
```diff
@@ -55,6 +57,24 @@ def _shortcut_description(description: str, shortcut: str) -> str:
     return f"{description} Alias for: {long_names_str}."
 
 
+class PipelineCacheConfig(StrictBaseModel):
```
lucaslie, line 60: See my other comments: could just be another transform.

tensorrt_llm/_torch/auto_deploy/transform/ad_ir.py
nvchenghaoz: When I first started the prototype, I tried the pytorch export utilities, but it turns out that some pytorch functions are not handling well on our use case. And seems they were the hard blocker for the work. 

>  1. ExportedProgram-based (torch.export.save/load, torch._export.serde) — requires an ExportedProgram, which has no slot for:                              
    - _load_state_dict_pre_hooks (AD sharding/aliasing/quant hooks)
    - GM-level Pydantic attrs                                                                                                                               
    - Non-aten custom ops (fail ExportedProgram.verifiers after pattern matching)
>  2. Pickle-based (torch.save, torch.package, torch.export.save payload, to_folder's state_dict.pt) — RCE vector for a shared cache root.                   
  torch.export.load's own docstring: "uses pickle under the hood... Never load data from an untrusted source."

tensorrt_llm/_torch/auto_deploy/transform/optimizer.py
nvchenghaoz: good idea and since we decided to remove the boundary set, this is doable. Will  update it later.

tensorrt_llm/_torch/auto_deploy/transform/pipeline_cache.py
nvchenghaoz: Fair.. Here is my thought - Most of the changes in the PR are general type handling, and they mostly depend on public FX APIs. The symmetric encode/decode is what roughly doubles the code size. Since there is no per-op hack, no model-specific handling, and no per-optimization hack, I have relatively high confidence this will stay stable as long as PyTorch does not introduce a breaking change. And if they do, it is fail-closed — cache miss, not a correctness bug, and agentic update :)

tensorrt_llm/_torch/auto_deploy/config/default.yaml
nvchenghaoz: good idea, will update.

tensorrt_llm/_torch/auto_deploy/transform/pipeline_cache.py
lucaslie: Sounds good, as discussed on Slack.

tensorrt_llm/_torch/auto_deploy/transform/ad_ir.py
lucaslie: Per our Slack discussion, we can keep it as is.

suyoggupta:
@nvchenghaoz : can you note the size (in KB) of the cached pipeline for qwen3.5 400B model?

nvchenghaoz:
Close this PR as we have a newer version of implementation.

PR #12733, opened 2026-04-03
[None][refactor] Unify sparse attention framework with clean backend interfaces

lfr-0531:
## Description

TensorRT-LLM already provides a sparse-attention framework for DSA, DeepSeek-V4,
RocketKV, and SkipSoftmax, but algorithm-specific module, backend, prediction,
metadata, cache, and kernel logic had become mixed across the common Attention
and MLA paths. That made the extension boundary inconsistent and required
changes to shared modules when adding or maintaining an algorithm.

This PR refactors sparse attention around explicit module and backend contracts:

- Attention and MLA remain the shared modules. They resolve a validated set of
  optional sparse-attention hooks for module initialization, weights, forward,
  custom-op execution, and output projection.
- Algorithm-specific module hooks and implementations live under
  `_torch/attention_backend/sparse/<algorithm>/`.
- `SparseBackendForwardArgs` carries sparse inputs from Attention/MLA to the
  selected backend, while `SparseRuntimeParams` carries prediction and
  SkipSoftmax runtime inputs from the backend to the attention op.
- Backend prediction dispatch is shared, while each algorithm implements only
  the hooks it needs.
- DSA, DeepSeek-V4, and RocketKV are organized by responsibility into
  `module.py`, `backend.py`, `params.py`, `metadata.py`, cache/indexer helpers,
  and algorithm-owned kernels. SkipSoftmax keeps its existing FMHA integration
  through the shared runtime parameters.

The refactor preserves dense MLA fallback behavior, DSA cross-layer indexer
sharing and piecewise CUDA graph support, DeepSeek-V4 compressed-cache paths,
RocketKV, MiniMax-M3 integration, and SkipSoftmax.

## Test Coverage

- [x] No C/C++ changes relative to `main`; a native rebuild is not required for
  this refactor.
- [x] `pre-commit` on all files changed by this PR.
- [x] Sparse Attention/MLA forward coverage, including DSA and DeepSeek-V4:
  70 passed.
- [x] Sparse LLM argument coverage: 7 passed, 245 deselected.
- [x] DSA pinned block-table staging smoke test on B200.
- [x] Verified that tests imported Python sources and native bindings from this
  worktree.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/sparse/deepseek_v4/metadata.py
+++ b/tensorrt_llm/_torch/attention_backend/sparse/deepseek_v4/metadata.py
@@ -697,7 +697,7 @@ class DeepseekV4TrtllmAttentionMetadata(DSAtrtllmAttentionMetadata):
         cu_seq_lens_buf: torch.Tensor,
         req_idx_per_token_buf: torch.Tensor,
     ) -> torch.Tensor:
-        """Compute cu_seq_lens, req_idx_per_token, and token_positions (eager)."""
+        """Compute cu_seq_lens and token_positions (eager)."""
         device = seq_lens.device
 
         # cu_seq_lens
@@ -705,14 +705,11 @@ class DeepseekV4TrtllmAttentionMetadata(DSAtrtllmAttentionMetadata):
             torch.cumsum(seq_lens.to(torch.int), dim=0), (1, 0)
         )
 
-        # req_idx_per_token via searchsorted
-        token_idx = torch.arange(num_tokens, dtype=torch.int32, device=device)
-        req_idx = torch.searchsorted(
-            cu_seq_lens_buf[1 : batch_size + 1].to(torch.int32), token_idx, right=True
-        )
-        req_idx_per_token_buf[:num_tokens] = req_idx
+        # DSA metadata rebuilt this map for the current seq_lens.
+        req_idx = req_idx_per_token_buf[:num_tokens].to(torch.int64)
 
         # token positions
+        token_idx = torch.arange(num_tokens, dtype=torch.int32, device=device)
         base_pos = cached_tokens[req_idx].to(torch.int32)
         offsets = token_idx - cu_seq_lens_buf[req_idx].to(torch.int32)
         return base_pos + offsets
--- a/tensorrt_llm/_torch/attention_backend/sparse/dsa/__init__.py
+++ b/tensorrt_llm/_torch/attention_backend/sparse/dsa/__init__.py
@@ -22,7 +22,7 @@ from .indexer import (
     transform_local_topk_and_prepare_pool_view,
     warmup_heuristic_topk_decode,
 )
-from .metadata import DSAtrtllmAttentionMetadata
+from .metadata import DSAtrtllmAttentionMetadata, build_req_idx_per_token
 from .params import DSABackendForwardArgs, DSAMetadataParams, DSAParams
 
 __all__ = [
@@ -42,6 +42,7 @@ __all__ = [
     "_effective_compress_ratio_divisor",
     "_pick_dsl_expand",
     "_select_indexer_compress_ratio",
+    "build_req_idx_per_token",
     "compute_cu_seqlen_kv_bounds_with_cache",
     "rotate_activation",
     "split_prefill_chunks",
--- a/tensorrt_llm/_torch/attention_backend/sparse/dsa/indexer.py
+++ b/tensorrt_llm/_torch/attention_backend/sparse/dsa/indexer.py
@@ -1590,6 +1590,7 @@ class Indexer(nn.Module):
 
         reuse_topk = (
             self.mtp_index_share
+            and metadata.in_mtp_draft_loop
             and metadata.indexer_skip_topk
             and metadata.shared_topk_indices is not None
         )
--- a/tensorrt_llm/_torch/attention_backend/sparse/dsa/metadata.py
+++ b/tensorrt_llm/_torch/attention_backend/sparse/dsa/metadata.py
@@ -45,6 +45,13 @@ if TYPE_CHECKING:
     from tensorrt_llm._torch.speculative.spec_tree_manager import SpecTreeManager
 
 
+def build_req_idx_per_token(seq_lens: torch.Tensor, num_tokens: int) -> torch.Tensor:
+    """Map each flattened-batch token to its request index."""
+    cu_seq_lens = torch.cumsum(seq_lens, dim=0, dtype=torch.int32)
+    token_idx = torch.arange(num_tokens, device=seq_lens.device, dtype=torch.int32)
+    return torch.searchsorted(cu_seq_lens, token_idx, right=True)
+
+
 @dataclass(init=False)
 class DSAtrtllmAttentionMetadata(TrtllmAttentionMetadata):
     """Attention metadata for DSA (Dense Sparse Attention) with indexer state."""
@@ -287,13 +294,18 @@ class DSAtrtllmAttentionMetadata(TrtllmAttentionMetadata):
         # pool_view cache here so it is recomputed on the next
         # transform_local_topk_and_prepare_pool_view() call.
         self._invalidate_pool_view_cache()
+
+        # The draft loop can rewrite seq_lens after prepare() built this map.
+        if self.num_tokens > 0:
+            self.req_idx_per_token[: self.num_tokens] = build_req_idx_per_token(
+                self.seq_lens_cuda[: self.num_seqs], self.num_tokens
+            ).to(self.req_idx_per_token.dtype)
+
         if self.kv_cache_manager is not None and self.num_tokens > 0:
             seq_lens = self.seq_lens_cuda[: self.num_seqs]
             # Runtime cached lengths after overlap/spec-dec correction.
             start_positions = self.kv_lens_cuda[: self.num_seqs] - seq_lens
 
-            # Reuse request-per-token mapping prepared in metadata.prepare().
-            # This avoids repeat_interleave in graph-capture mode.
             req_indices = self.req_idx_per_token[: self.num_tokens].to(dtype=torch.int64)
             seq_starts = torch.cumsum(seq_lens, dim=0, dtype=torch.int64) - seq_lens.to(torch.int64)
             token_offsets = (


tensorrt_llm/_torch/attention_backend/sparse/skip_softmax/backend.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+
+from __future__ import annotations
+
+from tensorrt_llm._torch.attention_backend.trtllm import TrtllmAttention, TrtllmAttentionMetadata
+
+from ..params import SparseParams
+
+
+class SkipSoftmaxTrtllmAttention(TrtllmAttention):
```
bobboli, line 28: Quick question: With this refactoring, do we need to specify `SKIPSOFTMAXTRTLLM` attention backend in LLM args/YAML to be able to use Skip Softmax Attention?

lfr-0531: We don't need that. Only set the sparse attention config as before. We can find the SkipSoftmaxTrtllmAttention from [here](https://github.com/lfr-0531/TensorRT-LLM/blob/887d21e722ef5b5f678c5e2f653acc601719ecbf/tensorrt_llm/_torch/attention_backend/sparse/registry.py#L52).

Already discussed on Slack.

karljang (approved):
LGTM from the VisualGen perspective. The existing TRTLLM SkipSoftmax configuration and runtime behavior appear to be preserved by this lower-level refactor.

Shixiaowei02 (approved):

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
```diff
@@ -0,0 +1,259 @@
...
+    "get_sparse_attn_hooks",
+    "prepare_sparse_runtime_params",
+]
+
+_SPARSE_ATTN_HOOK_MODULE_PATHS = {
+    "rocket": ".rocket.module",
+    "dsa": ".dsa.module",
+    "deepseek_v4": ".deepseek_v4.module",
+}
+_HOOK_PARAMETER_NAMES = {
```
heyuhhh: Can we add more comments for these hooks?

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
```diff
@@ -0,0 +1,259 @@
...
+    ),
+    "project_sparse_attn_output": (
+        "module",
+        "attn_output",
+        "position_ids",
+        "attn_metadata",
+        "all_reduce_params",
+    ),
+}
+_ALTERNATE_HOOK_PARAMETER_NAMES = {
```
heyuhhh: How to understand this? Seems that there are same function name but different parameters. If we have more than two different function signatures how to solve that? One way i can think is that remove this dict and move the hooks into the `_HOOK_PARAMETER_NAMES` and put hooks with same function name together

tensorrt_llm/_torch/attention_backend/sparse/minimax_m3/msa_backend.py
heyuhhh, line 1: Seems that there is no refactor for `minimax_m3` ?

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
lfr-0531: Added. Thanks, Yuhang!

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
lfr-0531: They have been replaced by one `_HOOK_SIGNATURES` table.

tensorrt_llm/_torch/attention_backend/sparse/minimax_m3/msa_backend.py
lfr-0531: For this PR, let's focus on the refactoring of the `MLA`/`Attention` module and sparse attention. Since `MiniMax M3` uses its own model-specific `MiniMaxM3Attention` and a specialized Python path FMHA, we'll skip modifying the MiniMax M3 attention part in this PR.

heyuhhh: Got it! Let's keep that in mind and add a TODO or NOTE there. Not only for us to further refactor later, also for coding agents not to follow this pattern :)

heyuhhh (approved):

tensorrt_llm/_torch/attention_backend/sparse/minimax_m3/msa_backend.py
lfr-0531: Added a NOTE to the `MiniMaxM3Attention` docstring.

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
```diff
@@ -0,0 +1,270 @@
...
+_SPARSE_ATTN_HOOK_MODULE_PATHS = {
+    "rocket": ".rocket.module",
+    "dsa": ".dsa.module",
+    "deepseek_v4": ".deepseek_v4.module",
+}
+# Each entry describes one optional function exported by an algorithm's
+# ``module.py``. A hook may have separate MLA and Attention call signatures;
+# keeping every accepted signature under the same hook name makes additional
+# module call sites explicit.
+_HOOK_SIGNATURES: dict[str, tuple[tuple[str, ...], ...]] = {
```
yihwang-nv: SparseAttnHooks.from_module() validates hooks by inspecting parameter names at runtime. This is brittle, detects interface errors late, and prevents static type checking. Consider defining separate typed adapters for Attention and MLA.

```python
class MLASparseHooks(Protocol):
    def forward(
        self,
        mla: "MLA",
        hidden_states: torch.Tensor,
        metadata: AttentionMetadata,
        outputs: list[torch.Tensor],
    ) -> None: ...

class DSAHooks:
    def forward(self, mla, hidden_states, metadata, outputs):
        forward_dsa(mla, hidden_states, metadata, outputs)
```

Algorithms could register these adapters explicitly, removing the need for inspect.signature() and alternate parameter-nametables.

What do you think?

lfr-0531: Sounds good to me. Let me refactor this hook implementation.

lfr-0531: Done. Thanks, Yihan!

bobboli:
Changes related to Skip Softmax Attention look nice.
Maybe also worth mentioning `AttentionForwardArgs` in sparse attention developer guide?

bobboli (approved):

tensorrt_llm/_torch/modules/mla.py
```diff
@@ -755,57 +452,113 @@ def yarn_get_mscale(scale=1, mscale=1):
...
 
-        self.has_dsv4_indexer = (
-            self.is_deepseek_v4
-            and layer_idx is not None
-            and sparse_params is not None
-            and sparse_params.compress_ratios[layer_idx] == 4
+        self.aux_stream = aux_stream
+        self.ln_events = [torch.cuda.Event(), torch.cuda.Event()]
+        self.kv_a_layernorm = RMSNorm(hidden_size=self.kv_lora_rank, dtype=dtype, eps=rms_norm_eps)
+        self.kv_b_proj = Linear(
```
mingyangHao: Blocking: this unconditionally constructs `kv_b_proj` (and later `v_b_proj`, `o_proj`, and the
  dense MHA backend) before the sparse hook is initialized. The DeepSeek-V4 hook then deletes
  these modules.

  This is not only an abstraction issue: `skip_create_weights_in_init` defaults to false, so the
  unused weights are actually allocated. With the dimensions used by the DeepSeek-V4 test, the
  three deleted BF16 parameters alone account for roughly 84.5 MiB of transient allocation per
  layer; larger production configurations can be substantially worse. Depending on the default
  device used during model construction, this can increase host memory or GPU memory peaks and
  potentially cause initialization OOMs.

  The pre-refactor implementation explicitly avoided constructing these modules for DeepSeek-V4.
  Please resolve the algorithm-specific build requirements before constructing the common MLA
  modules, e.g. through a declarative build spec/capability object or an algorithm-specific
  component factory, rather than constructing and deleting unused modules.

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+def register_attention_sparse_hooks(algorithm: str, hooks: type[AttentionSparseHooks]) -> None:
+    """Register the typed Attention adapter for ``algorithm``."""
+    if algorithm in _ATTENTION_HOOKS:
+        raise ValueError(f"Attention sparse hooks are already registered for {algorithm!r}")
+    _ATTENTION_HOOKS[algorithm] = hooks
+
+
+@lru_cache(maxsize=None)
```
mingyangHao: Major: this makes the public registration API ineffective for a new algorithm. Even after
  `register_mla_sparse_hooks("new_algo", NewHooks)` succeeds, this lookup returns `None` unless
  `"new_algo"` is also added to the private `_MLA_HOOK_MODULE_PATHS` map.

  The same issue exists in the Attention hook getter. The getter should first consult the hook
  registry and use the module-path map only to lazily import built-in algorithms. Please also
  add a test that registers an arbitrary algorithm name and retrieves its hook without modifying
  a central built-in map.

  If external registration is intentionally unsupported, the `register_*` functions should not
  be exported as public extension points.

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
```diff
@@ -0,0 +1,262 @@
...
+def _get_sparse_mla_hooks_for_algorithm(algorithm: str) -> Optional[MLASparseHooks]:
```
mingyangHao: Major: this cache stores the instantiated hook object, so every layer and every model using
  the same algorithm shares one global adapter instance.

  The current adapters happen to be stateless, but neither the interface nor the type system
  enforces that constraint. A future hook that stores initialization or layer-specific state on
  `self` would silently leak state across layers or model instances.

  Please cache only the lazy-import or hook class resolution and instantiate the adapter per
  owning MLA/Attention module. A state-isolation test with two module instances would help
  protect this contract.

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
```diff
@@ -0,0 +1,262 @@
...
+]
+
+
+class MLASparseHooks(ABC):
+    """Typed module-layer adapter for a sparse MLA algorithm."""
+
+    @abstractmethod
+    def initialize(
+        self,
+        mla: "MLA",
```
mingyangHao: Major architectural concern: although this adapter is typed, passing the concrete `MLA` object
  gives each sparse algorithm unrestricted access to the entire module implementation. The
  current implementations directly inspect, replace, and delete MLA attributes, so any internal
  MLA refactor can break an algorithm at runtime.

  Moreover, most of the actual algorithm logic is delegated to free functions whose `self`
  parameter is untyped, which means the typed adapter does not provide meaningful static
  checking for those internal dependencies.

  Please consider modeling the sparse implementation as a composed `SparseMLAComponent` that
  owns its algorithm-specific weights and state, and pass it a narrow, stable context/services
  interface instead of the full MLA object. At minimum, the set of MLA fields available to hooks
  should be explicitly defined and contract-tested.

tensorrt_llm/_torch/attention_backend/sparse/params.py
```diff
@@ -12,49 +12,46 @@
...
-@dataclass
-class SkipSoftmaxKernelParams:
-    """Skip-softmax thresholds passed to attention backend kernels."""
+    sparse_kv_indices: Optional[torch.Tensor] = None
+    sparse_kv_offsets: Optional[torch.Tensor] = None
+    sparse_attn_indices: Optional[torch.Tensor] = None
+    sparse_attn_offsets: Optional[torch.Tensor] = None
+    sparse_attn_indices_block_size: int = 0
+    # DeepSeek-V4 compressed-cache inputs.
+    sparse_mla_topk_lens: Optional[torch.Tensor] = None
```
mingyangHao: Major extensibility concern: this supposedly shared runtime contract already mixes generic
  sparse-index fields with DeepSeek-V4-specific cache inputs and SkipSoftmax-specific
  thresholds.

  With this design, every new algorithm-specific runtime value requires modifying the common
  dataclass and likely the shared THOP/C++ attention ABI, causing unrelated algorithms and
  backends to depend on fields they never use.

  Please keep only genuinely algorithm-independent indexing data in `SparseRuntimeParams`.
  Algorithm-specific values should remain in a typed/discriminated payload and be lowered by the
  selected backend at the THOP boundary. This would prevent the common carrier from growing into
  a sparse-attention parameter bag.

tensorrt_llm/_torch/modules/mla.py
```diff
@@ -755,57 +452,113 @@ def yarn_get_mscale(scale=1, mscale=1):
...
-        self.aux_stream = aux_stream
-        self.ln_events = [torch.cuda.Event(), torch.cuda.Event()]
-        self.dsv4_overlap_start_event = torch.cuda.Event()
-        self.dsv4_compressor_start_event = torch.cuda.Event()
-        self.dsv4_compressor_event = torch.cuda.Event()
-        self.dsv4_indexer_event = torch.cuda.Event()
+        if self.sparse_attn_hooks is not None:
+            self.sparse_attn_hooks.initialize(
+                self,
+                config=config,
```
yuxianq: Instead of passing many arguments, how about making all of them MLA fields and only pass self to `initialize`?

tensorrt_llm/_torch/modules/mla.py
```diff
@@ -881,41 +586,34 @@ def yarn_get_mscale(scale=1, mscale=1):
...
-        # models when the short-seq MHA optimization is active.
         if self.mha is not None:
             self.mha.update_quant_config(self.quant_config)
         self.mqa.update_quant_config(self.quant_config)
 
         # Although we use FP8 MLA for context/generation phase, the output is still in BF16
         self.out_scale = None
+        if self.sparse_attn_hooks is not None and self.sparse_attn_hooks.create_weights(self):
+            self._weights_transformed = False
+            return
```
yuxianq, line 653: If we implement `sparse_attn_hooks` and `sparse_attn_hooks.create_weights`, the normal `create_weights` is skipped, which is unexpected. For deepseek v4, it is correct, but not for other cases. Since we don't return True for all other cases, it happens to be correct.
Consider that all these code is about absorption and dsv4 does not need it, we can add a `need_absorption: bool` field, set it to False in dsv4 case, and only run these code if True.

tensorrt_llm/_torch/modules/mla.py
lfr-0531: Fixed. Thanks, Mingyang! `MLASparseHooks` now declares `need_absorption`, `need_dense_mha`, and `need_default_o_proj`, and `MLA` checks them before constructing optional modules. DeepSeek-V4 disables all three, while DSA creates dense MHA only for a supported short-sequence fallback.

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
lfr-0531: Fixed. Both hook getters now consult the public registry first and use the built-in map only for lazy imports.

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
lfr-0531: Fixed. Lookup now resolves the hook class and creates a fresh instance for each MLA or Attention module.

tensorrt_llm/_torch/attention_backend/sparse/hooks.py
lfr-0531: I agree that a composed component would provide a stronger ownership boundary. But the current algorithms still need to change some submodules in `MLA`, so that change requires a broader redesign. Let's keep the current implementation for this PR and see if there's a better design later. I'm thinking that if we don't modify the module in the algorithm part, we might need to add extra conditionals and branches in the mla/attention module code.

tensorrt_llm/_torch/attention_backend/sparse/params.py
lfr-0531: I kept `SparseRuntimeParams` as the flat backend-to-`AttentionOp` interface because the thop/C++ API is also flat. Module-to-backend inputs remain typed through `SparseBackendForwardArgs`; a Python discriminated payload would still need to be flattened at the native boundary. I renamed the reusable fields to `sparse_attn_kv_lens` and `aux_kv_cache_pool_ptr`. If other sparse attention methods use a similar design in the future, we can continue to reuse these two interfaces.

tensorrt_llm/_torch/modules/mla.py
lfr-0531: Fixed. Thanks, Yuxian! Required construction settings are now stored on `MLA`, and the hook contract is simply `initialize(self, mla)`.

tensorrt_llm/_torch/modules/mla.py
lfr-0531: Fixed. `need_absorption` now controls both weight paths, with DeepSeek-V4 setting it to `False`. Also added `need_dense_mha` and `need_default_o_proj`, which separately control the other optional modules.

zhaoyangwang-nvidia (approved):

yihwang-nv (approved):
Thanks, LGTM!

WeiHaocheng (approved):
Approve as the modeling part LGTM.

govind-ramnarayan (approved):

yizhang-nv (approved):

PR #12745, opened 2026-04-03
[#12633][feat] AutoDeploy: Support torch-cudagraph for Eagle

### LINKED ISSUE #12633: [Feature]:  [AutoDeploy]: TRTLLM Attention + CUDA Graph for Eagle

get performance for MTP

### Alternatives

_No response_

### Additional context

_No response_

govind-ramnarayan:
torch-cudagraph mainly tested on Llama-3.1-8B-Instruct + Eagle3, as well as SuperV3 + MTP.

Fixes: #12633 

Note: There are some TODOs left from this PR to investigate some lingering issues. They are here: 

#13100 
#13133 
#13134 
#13135
#13143 

## Summary by CodeRabbit

* **New Features**
  * Enhanced speculative decoding support for Eagle and MTP models with improved batch handling for extend-only sequences.
  * Extended maximum sequence length to 65K tokens.
  * Added TensorRT-LLM as a supported attention backend with CUDA graph capture optimization.
  * Improved distributed synchronization for speculative decoding in multi-GPU environments.

* **Bug Fixes**
  * Fixed attention backend compatibility issues with speculative decoding configurations.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
lucaslie, line 1: Diff here should be minimal based on first principle. We would just now capture graphs for shapes `(bs, 1+max_draft_len)` instead of `(bs, 1)`. Right?

lucaslie: This utility doesn't even need to know about spec dec or not

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1539,6 +1666,23 @@ def __eq__(self, other: Optional[ResourceHandler]) -> bool:
             and self.kv_layout == other.kv_layout
         )
 
+    def is_management_compatible(self, other: Optional[ResourceHandler]) -> bool:
```
lucaslie: probably can be deprecated based on our discussions, right?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -926,6 +927,74 @@ def set_generate_only_batch(self, batch_size: Optional[int] = None) -> None:
         batch_size = batch_size or self.max_batch_size
         self.set_example_sequence(torch.ones(batch_size, 1, dtype=torch.int))
 
+    def set_eagle_extend_batch(self, batch_size: int, max_draft_len: int) -> None:
```
lucaslie, line 942: we probably just want to adopt the `set_generate_only_batch` function, right?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1158,6 +1228,20 @@ def nest_sequences(
...
+        # Used by TRTLLM attention as context_lengths (GPU) arg to thop.attention.
+        # HACK: if not provided, fall back to seq_len (correct for prefill, wrong for
+        # extend/decode). The executor should be updated to always provide prompt_lens.
+        if prompt_lens is not None:
+            if isinstance(prompt_lens, (list, tuple)):
+                prompt_lens = torch.tensor(prompt_lens, dtype=torch.int)
+            self._stage_arg("prompt_lens", prompt_lens)
+        elif not self._is_active("prompt_lens"):
+            # First call — use seq_len as the initial prompt length
+            self._stage_arg("prompt_lens", sl_host)
```
lucaslie: why is this computed in `_prepare_inputs`? This way we don't need any hacks.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1283,7 +1367,10 @@ def register_host_prepare_for_attention_forward(
     @nvtx_range("ad_host_prepare_for_attention_forward")
     def run_host_prepare_for_attention_forward(self) -> None:
         for host_function, args in self._host_prepare_functions:
-            host_function(**{arg: self.get_arg(arg) for arg in args})
+            with nvtx_range("ad_host_prepare_get_args"):
+                kwargs = {arg: self.get_arg(arg) for arg in args}
+            with nvtx_range("ad_host_prepare_call"):
+                host_function(**kwargs)
```
lucaslie: nit: probably we don't need such a detailed information in the nsys trace unless you found that you need that level of debugging information

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1302,63 +1389,96 @@ def offset_pos_and_cache_(self, offset: torch.Tensor) -> None:
...
-            for k in _REQUIRES_UPDATE
-            if self._is_active(k + self._host_suffix, check_both=False)
-        ]
-        sync_to_host = any(needs_d2h_sync)
-        if sync_to_host:
-            ad_logger.debug(f"d2h sync required in offset_pos_and_cache_ for {needs_d2h_sync}")
-
-        num_sequences = self.num_sequences
-        assert offset.shape[0] == num_sequences, f"{offset.shape[0]=} != {num_sequences=}"
+        with nvtx_range("ad_opc_needs_d2h_sync"):
```
lucaslie: same here

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1371,7 +1491,14 @@ def offset_with_new_lens_(self, new_lens_ungathered: torch.Tensor) -> None:
...
         gather_slot_idx = self.get_arg("_gather_slot_idx", truncate=True)
-        increment[num_prefill:] = new_lens_ungathered[gather_slot_idx] - 1
+        num_overlap = gather_slot_idx.numel()
+        # AD overlap mode packs real overlap-carried non-prefill sequences first. CUDA-graph dummy
+        # requests are appended later and therefore remain in the trailing zero-increment region.
+        # This keeps the padded dummy requests from consuming a new_tokens_lens entry while still
+        # preserving the existing ordering-based contract used throughout SequenceInfo.
+        increment[num_prefill : num_prefill + num_overlap] = (
+            new_lens_ungathered[gather_slot_idx] - 1
+        )
```
lucaslie, line 1446: please help me understand this a bit more. Ideally, with an actual numerical example

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -1014,22 +1068,15 @@ def _forward_with_kv_cache(self, csi: CachedSequenceInterface):
...
-            if kv_cache_manager.is_speculative():
-                _ctx = SimpleNamespace(num_seqs=num_sequences, num_contexts=num_prefill)
-                kv_cache_manager.update_mamba_states(
-                    attn_metadata=_ctx,
-                    num_accepted_tokens=new_tokens_lens,
-                )
+        # NOTE: MTP state promotion (update_mamba_states) for Mamba hybrid models is
+        # handled by the executor after this forward returns. It was moved out of the
+        # forward to avoid baking Python ints into the CUDA graph capture, which caused
+        # incorrect state updates during graph replay.
```
lucaslie: is this still the case? Otherwise, I would love for this to move back here

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -745,6 +745,15 @@ def __init__(self, config: EagleWrapperConfig, target_model: nn.Module, draft_mo
...
         self.normalize_target_hidden_state = config.normalize_target_hidden_state
+        self.sync_before_hidden_state_capture = config.sync_before_hidden_state_capture
+        self._draft_hidden_size = draft_model.config.hidden_size
+        self._draft_num_capture_layers = getattr(draft_model.config, "num_capture_layers", 1)
+        self._buffers_initialized = False
+        self._buf_new_tokens_2d: Optional[torch.Tensor] = None
+        self._buf_next_new_tokens: Optional[torch.Tensor] = None
+        self._buf_new_tokens_lens: Optional[torch.Tensor] = None
+        self._buf_c_offset_ones: Optional[torch.Tensor] = None
+        self._buf_hidden_states: Optional[torch.Tensor] = None
```
lucaslie: why do we need to have an explicit separate buffer here?

tensorrt_llm/_torch/auto_deploy/llm_args.py
```diff
@@ -368,6 +383,7 @@ def create_factory(self) -> ModelFactory:
             tokenizer_kwargs=self.tokenizer_kwargs,
             skip_loading_weights=self.skip_loading_weights,
             max_seq_len=self.max_seq_len,
+            attn_backend=self.attn_backend,
```
lucaslie: why does the factory need to know about the attention backend?

lucaslie: if this is for the information regarding the extra sync let's just define a property that can be passed in that describes this and resolve it here

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -350,14 +461,19 @@ def trtllm_mha_with_cache(
     pool_pointers encodes kv_cache.data_ptr() (layer-specific), and
     pool_mapping is all zeros. See module docstring for details.
     """
+    # Get batch dimensions and model-level constants from host tensors (no device sync)
```
lucaslie: is this correctly updated now and defines num_extend and num_generate seperately? We should also use this in the attention_interface to correctly shape a batch for `(bs, 1+max_seq_len)` instead of `(1, total_num_tokens)` when we have extend requests

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
lucaslie, line 1: would love to spend some time walking through these changes with you to understand them better

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
govind-ramnarayan: Correct, agreed

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: need to check if I can set draft dtype to target dtype. Todo

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: wil remove - it's debugging cruft

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: No, this is a wrong comment - I moved it out to simplify the forward while debugging but I will move it back in after finishing

tensorrt_llm/_torch/auto_deploy/llm_args.py
govind-ramnarayan: Yep, that's it. That makes sense

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: Let me revisit this as part of the cleanup - I thought these explicit buffers would be needed for cuda graph compatibility so that we're not returning locally-allocated tensors as output. I'll revisit this at the end.

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
```diff
@@ -221,40 +227,36 @@ def capture_graph(self, get_args_kwargs: GetArgsKwargsForBatchSize, batch_sizes:
...
 
             # capture graph for truncated inputs
             combined_shape = sum((tuple(input.shape) for input in inputs_truncated), start=())
-            self.cudagraphs[combined_shape] = self._capture_one_graph(*args, **kwargs)
+            self.cudagraphs[combined_shape] = self._capture_one_graph(
+                *args, refresh_inputs_fn=refresh_inputs_fn, **kwargs
+            )
 
     def forward(self, *args, **kwargs) -> Any:
         """Run the compiled graph."""
```
govind-ramnarayan, line 269: Should there be any protection against context sequences that happen to have length 1+max_draft_len? We can do an if statement like 

```
    if csi is not None and not csi.info.is_extend_only:
            return self.model(*args, **kwargs)
```

but maybe want to avoid an extra host if statement. @lucaslie

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
govind-ramnarayan: Can you remind me what this was originally about? I think this is fine - the thing is that the TRTLLM Attention backend only has "context" and "generation" requests. Extend requests are effectively generation requests with speculative decoding mode on. So a lot of this code will do stuff like use `num_decode + num_extend `or only use `prefill` and total num tokens, etc.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
```diff
@@ -598,41 +595,43 @@ def test_accuracy(self, model_id, world_size, enable_attention_dp,
 
         print_memory_usage("after evaluation")
 
-    @pytest.mark.skip_less_device_memory(180000)
+    # TODO: re-instate skip_less_device_memory once CI runs on nodes with >180GB
+    # @pytest.mark.skip_less_device_memory(180000)
```
govind-ramnarayan, line 602: Reminder to self to un-comment this; but for now it is useful so I can test on H100s...

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: The issue was this was the combination of CUDA graph dummy requests with the overlap scheduler correction. 


Let's say there were 15 prefill requests, 15 extend (non-dummy) requests, and 2 dummy requests in this batch. Before, without dummy requests, the logic would do 

`increment[15:32] = new_lens_ungathered[gather_slot_idx] - 1`

but len(gather_slot_idx) == 15 (num extend non-dummy) due to the logic here: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py#L795

so we would get an error.

We know we want to update the real extend requests, as those are the ones that have an update to apply from the previous iteration. So we want this to instead be 

`increment[15:30] = new_lens_ungathered[gather_slot_idx] - 1`

and we compute 30 as 15+15.

The reason this works is that dummy requests are always at the end by construction: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py#L400

This fact is actually a little more subtle than it seems - we split scheduled_requests.generation_requests into extend and decode requests here: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py#L720

but we have an invariant that either `len(extend_requests) == 0` or `len(decode_requests) == 0` here at all times (either we have spec dec or not). I did not put an assert to keep asserts off the hot path.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -1101,6 +1189,12 @@ def _forward_with_kv_cache(self, csi: CachedSequenceInterface):
                 csi.info.offset_pos_and_cache_(c_offset)
 
         # ---- Phase 6: Package output ----
+        # For extend-only batches, reshape logits to (bs, mdl+1, vocab) so CUDA graph
```
govind-ramnarayan: @lucaslie should I keep this? These logits are only used for debugging, but if it comes for free it's nice to keep them properly formatted...

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
lucaslie: Context sequences are going to be shaped as `(1, total_num_tokens)`. The only issue here could be that multiple sequences exactly add up to 1+max_draft_len, in which case we then incorrectly trigger a CUDAGraph. I'd say the chances of these are fairly low; however, it is an important limitation of this entire module, and this is something we may have to change down the line anyway. 

I don't think we have to do it as part of this PR, though. For now, let's stick with the plan that we discussed, which is that batches with extend-only requests get shaped as (batch_size, 1 + max_draft_length). We then just do the same matching shape matching as we do right now.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -233,6 +275,36 @@ def clear_trtllm_attention_fp8_input_scale(attn_node: Node) -> None:
     attn_node.meta.pop(_TRTLLM_ATTN_FP8_INPUT_SCALE_KEY, None)
 
 
+@functools.cache
```
lucaslie: What's up with the decorator here?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -233,6 +275,36 @@ def clear_trtllm_attention_fp8_input_scale(attn_node: Node) -> None:
...
+def _generate_spec_decoding_position_offsets(max_num_requests: int, draft_len: int) -> torch.Tensor:
+    width = draft_len + 1
+    row = torch.arange(width, dtype=torch.int, device="cuda")
+    return row.unsqueeze(0).expand(max_num_requests, -1).contiguous()
+
+
+@functools.cache
```
lucaslie: Same here.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -243,26 +315,38 @@ def prepare_trtllm_metadata_host(
...
+        batch_info.get_max_blocks_per_seq(),
     )
+
+    _GlobalTrtllmPlanner.init_spec_decoding(batch_info.get_max_batch_size(), spec_config)
+
+    # Extend (spec decoding) requests are expected as decode requests by the kernel.
     _GlobalTrtllmPlanner.plan_host(
         num_prefill=num_prefill,
-        num_decode=num_decode,
+        num_decode=num_decode + num_extend,
```
lucaslie: Please correctly update the plan_host function now that we separately handle decode and extend requests.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
lucaslie: Our batch_info object defines decode and extend requests separately, so I would like for our wrappers for the kernel to also handle these separately. Of course, once you call thop.attention, you have to correctly resolve it to whatever the kernel wants.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -420,15 +506,46 @@ def trtllm_mha_with_cache(
...
     sm_version = get_sm_version()
-    if sm_version >= 89:  # Ada/Hopper
+    if not (sm_version < 100 or sm_version in (120, 121)):
         spec_decoding_tensor_params.extend([None, None, None])
 
     mla_tensor_params = [None, None]
 
+    # Defensive copy of block_offsets into a pre-allocated scratch buffer.
+    # Only needed for speculative decoding where the Eagle draft loop's
+    # repeated graph replays can race on the shared block_offsets singleton.
```
lucaslie: What does this comment even mean?

lucaslie: From our discussion yesterday, it seems you only need that in a very particular circumstance, so why not just document that here? As far as I remember, this is llama+eagle, and otherwise you don't need this, right?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
lucaslie, line 403: nit: Can you remove some of this diff so it's cleaner to review, please?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -420,15 +506,46 @@ def trtllm_mha_with_cache(
...
+
+    num_extend = batch_info.get_num_sequences()[1]
+    is_spec_decoding_enabled = _GlobalTrtllmPlanner.is_spec_decoding_enabled
+    use_spec_decoding = num_extend > 0
+    spec_decoding_bool_params = [is_spec_decoding_enabled, use_spec_decoding, False]
+    spec_decoding_tensor_params = [
+        _GlobalTrtllmPlanner.spec_decoding_generation_lengths,
+        _GlobalTrtllmPlanner.spec_decoding_position_offsets,
+        _GlobalTrtllmPlanner.spec_decoding_packed_mask,
+    ]
```
lucaslie: Why is this not part of some of the prepare functions?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -358,14 +442,19 @@ def trtllm_mha_with_cache(
     pool_pointers encodes kv_cache.data_ptr() (layer-specific), and
     pool_mapping is all zeros. See module docstring for details.
     """
+    # Get batch dimensions and model-level constants from host tensors (no device sync)
+    batch_info = BatchInfo(batch_info_host)
+    num_tokens = batch_info.get_total_num_tokens()
+    max_context_length = batch_info.get_max_context_length()
+    max_num_requests = batch_info.get_max_batch_size()
+    _GlobalTrtllmPlanner.update_host_request_types(batch_info)
```
lucaslie: Why is this one needed here in every call of the attention function?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -233,6 +275,36 @@ def clear_trtllm_attention_fp8_input_scale(attn_node: Node) -> None:
...
 
+@functools.cache
+def _generate_spec_decoding_position_offsets(max_num_requests: int, draft_len: int) -> torch.Tensor:
+    width = draft_len + 1
+    row = torch.arange(width, dtype=torch.int, device="cuda")
+    return row.unsqueeze(0).expand(max_num_requests, -1).contiguous()
+
+
+@functools.cache
+def _generate_spec_decoding_packed_mask(max_num_requests: int, draft_len: int) -> torch.Tensor:
```
lucaslie, line 323: So this is entirely static, right?

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
```diff
@@ -123,12 +122,22 @@ def _process_metadata_host(self, cm: CachedSequenceInterface):
...
 
-        # Register the host-side prepare metadata function with SequenceInfo.
-        # Arg availability is validated by require_copy() inside register_host_prepare.
         sig = inspect.signature(prep_meta_host_op)
-        cm.info.register_host_prepare_for_attention_forward(
-            prep_meta_host_op, list(sig.parameters.keys())
-        )
+        params = list(sig.parameters.keys())
+
+        # If the host-prepare function accepts spec_config (e.g., TRTLLM attention
```
lucaslie, line 128: what's up with this? why do you need spec_config as part of the host prepare function? If there is some information that is missing right now it should be added to `BatchInfo`

tensorrt_llm/_torch/auto_deploy/utils/cuda_graph.py
```diff
@@ -35,8 +35,11 @@ def in_warm_up() -> bool:
 
 @contextmanager
 def CudaGraphWarmUpPhase():
```
lucaslie, line 37: Please tell me more about this diff

tensorrt_llm/_torch/auto_deploy/utils/node_utils.py
lucaslie, line 1: Please help me better understand the changes here.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -759,6 +760,7 @@ def _prepare_inputs(
             input_ids.extend(prompt_tokens)
             cu_seqlen.append(len(input_ids))
             input_pos.append(begin_compute)
+            prompt_lens.append(request.context_chunk_size)
```
lucaslie, line 763: Is this context_chunk_size or end_compute? I am not quite sure of this notion.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
lucaslie: Like I mentioned before, everything should be formatted as (num_sequences, 1 + self.max_draft_len) when we have an extend-only batch

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -810,6 +845,14 @@ def apply_lm_head(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
     def sample_greedy(self, logits: torch.Tensor) -> torch.Tensor:
         ret = torch.argmax(logits, dim=-1)
+        # Broadcast rank 0's sampled tokens to all ranks. Without TP
+        # sharding we have observed slight differences in logits across
+        # ranks. When this leads to different tokens getting sampled,
+        # it can cause different acceptance patterns across ranks,
+        # inconsistent input_pos, and a hang. This synchronizes every
+        # validation step to prevent this.
+        if dist.is_initialized() and dist.get_world_size() > 1:
+            broadcast(ret, src=0)
```
lucaslie: Please make sure to use the correct wrappers for the dist operations.

lucaslie: Don't use the plain pytorch ones.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
govind-ramnarayan: It needs to update between the target attention and the draft attentions. It could happen once per submodule forward - I guess that means it can go in preparing the device metadata?

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
govind-ramnarayan: I need it just to initialize the spec_config fields in host metadata prepare

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: It's supposed to be chunk size when we are doing chunked prefill, and it's supposed to be the full prompt length when we do decode.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
govind-ramnarayan: These utilities are copied from https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/attention_backend/trtllm.py#L784 , along with the memoizing decorator.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
govind-ramnarayan: yup

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
govind-ramnarayan: yeah thats better ill do that

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
govind-ramnarayan: So it's just capturing the different shape, and then also handling the fact that EagleWrapper.forward() modifies metadata (it effectively changes the input request from all-extend to all-decode, for instance) so we need to re-initialize the inputs afterwards.

I think it makes sense to just completely re-initialize the inputs since it also changes various things about caches, but we could maybe also get away with just changing back from all-decode to all-extend.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: Refactored to use the same helper `set_example_sequence().` Since Eagle wrapper forward does stuff with cache metadata, I think it's good to make sure some of the relevant page info is set properly to get a representative run? So I set the size of the buffer as well as some metadata about pages.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: You could imagine that end_compute would be the right setting here ("prompt length so far" during prefill) but this does not seem to be what the PyTorch backend does, so I just went with what it does [here](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/model_engine.py#L2301C39-L2301C52). 

I suspect the idea is that the kernel does not know about chunked prefill, and assumes any chunk of context requests it gets is the entire context. We could consider handling this closer to the kernel to make it less confusing if we have the info about chunk size handy.

tensorrt_llm/_torch/auto_deploy/utils/node_utils.py
govind-ramnarayan: The way the sharder / utilities found the embedding width was to look for the first linear node in the graph and look at its input size. It then assumed that this embedding width can be used to identify the beginning and the end of the layer. This is not true with MTP / Eagle draft models, because these draft models take as input previous tokens AND hidden states, both of which are representing by the embedding dimension, so the input to their first linear is dimension 2h, where h is the embedding dim.

So if you apply the existing logic, you discover `2h` instead of `h` as the embedding dimension, and then get lost when you look for the end of the layer because it outputs dimension `h` instead of `2h`. This could be worked around by assuming there is only one layer to the drafter I guess, but it breaks the current assumptions, so we might as well properly find the embedding dimension.

This is handled slightly differently by different architectures - for Nemotron MTP, this is handled by doing a linear projection from 2h to h [here](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_nemotron_h.py#L684). For Llama + Eagle3, this is handled by making the attention inputs double the size [here](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py#L330). We want to detect both these cases and extract the correct embedding dimension in both cases.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
govind-ramnarayan: Not needed for us now that we've moved these to the planner initialization. - removed

tensorrt_llm/_torch/auto_deploy/custom_ops/utils/torch_gather_logits.py
govind-ramnarayan, line 27: Note: Changes to this file are from upstream/main. They were useful for me on this branch when I was cleaning before rebasing.

xinhe-nv:
@coderabbitai Act as a QA engineer reviewing test changes for TensorRT-LLM.

        QA test list hygiene (integration / release runs):
        - If the change adds or materially alters an integration test under
          tests/integration/defs/ (or otherwise affects what QA should run on a
          schedule), call out whether an entry is needed under
          tests/integration/test_lists/qa/. See tests/integration/test_lists/qa/README.md
          for which file to use (e.g. llm_function_core.txt for primary single-node
          multi-GPU functional cases; llm_function_multinode.txt for multi-node;
          llm_perf_*.yml for perf; llm_triton_integration.txt for Triton).
        - If the PR only touches unittest/ or narrow unit scope, say explicitly
          whether QA list updates are unnecessary or optional.

        Coverage expectations:
        - Assess whether new/changed tests cover happy path, important edge cases,
          and failure modes relevant to the feature or fix (skips, guards, env
          vars like LLM_MODELS_ROOT, GPU count, backend-specific paths).
        - Flag tests that assert overly brittle behavior (e.g. exact token match
          across speculative vs non-speculative paths) unless the product contract
          requires it.
        - Note missing negative tests, missing parametrization where multiple
          backends or dtypes apply, or tests that cannot fail when behavior regresses.

        Performance test coverage:
        - If the PR touches performance-sensitive paths (attention kernels, MoE
          routing/dispatch, KV cache management, scheduler, batching logic,
          CUDA graph capture, speculative decoding, or quantization kernels),
          check whether a perf test entry is present or updated in:
          (a) tests/integration/test_lists/test-db/l0_perf.yml or the
              appropriate per-GPU l0_*.yml (primary requirement — this is how
              tests enter pre-merge and scheduled CI); and
          (b) tests/integration/test_lists/qa/llm_perf_*.yml (single-node) or
              tests/integration/test_lists/qa/llm_perf_multinode.txt (multi-node)
              for QA scheduled runs.
          Flag if (a) is missing even when (b) is present, since test-db is
          the authoritative source for CI execution.
        - Flag if a test in tests/integration/defs/perf/test_perf_sanity.py is
          missing for a new model or serving configuration that warrants a
          latency/throughput baseline (examples/benchmark/ is deprecated).
        - Note if only functional correctness is tested for a change where a
          performance regression would not be caught (e.g. a kernel rewrite with
          no perf test, or a new scheduling policy with no throughput assertion).
        - Do not require perf tests for doc-only, infra, style, or pure refactor
          changes where no runtime behavior changes.

        Keep feedback actionable: suggest concrete list file names and whether
        coverage is sufficient, insufficient, or needs follow-up outside the PR.

xinhe-nv (approved):

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
lucaslie: Please elaborate. There is one CUDA graph that combines the target model with verification, changing the input and running the drifter in a loop. This function triggers replaying that CUDA graph. How is it relevant that the wrapper modifies metadata?

lucaslie: There is already a function `get_args_kwargs_bs` that will correctly reassign the inputs. How is this not sufficient for what you are describing?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
lucaslie: I think my question remains: How is this function different from just generalizing `set_generate_only_batch` to handle `1+max_draft_len` instead of just `1`?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1356,8 +1404,29 @@ def offset_pos_and_cache_(self, offset: torch.Tensor) -> None:
...
+            offset_for_pos_ids = torch.zeros(
+                num_tokens,
+                dtype=offset.dtype,
+                device=offset.device,
+            )
+            if num_extend > 0:
+                tokens_per_seq = num_extend_tokens // num_extend
+                offset_for_pos_ids[num_prefill_tokens:] = offset[
+                    num_prefill : num_prefill + num_extend
+                ].repeat_interleave(tokens_per_seq)
```
lucaslie: That logic just doesn't look correct to me. You're not handling mixed pre-fill and extend here. 

I think a cleaner split would be to handle generate only, then extend only, and then just keep the general-purpose utility for mixed batches.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
lucaslie: So it seems to me we just need to correctly update dummy requests so that we can account for them being at the end, right? It seems like if we do drafting, those dummy requests at the end should be extend requests and not plain generate requests. 

If I had to guess, this might actually also be the reason that you see an accuracy drop.

lucaslie: One way to test the theory I just mentioned is you can actually just capture a CUDA graph for every batch size and disable padding so that it falls back to eager execution if you can't have a CUDA graph that perfectly matches the batch size.

tensorrt_llm/_torch/auto_deploy/utils/node_utils.py
lucaslie: That logic is just bound to break at any given moment. I think we need to think about that and do it somewhere else. Happy to chat about it.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
lucaslie: I'm confused. Why can't you test this on h100? Why do you need to have this pytest marker?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: I'll give that a try; I wanted to make the call as close to what happens in the actual forward run as possible (reasonable cache_loc, extra_page_per_seq, etc.) The current form is pretty close to what you said but I can try to make it just that with extra_args.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: Okay, yeah I got rid of this one while trying to make chunked prefill work. I can put it back. 

The reasoning here was that the offset for prefill positions is 0 because they have not run before so they have no offset to gain (also see the logic before for populating increment - it populates num_prefill:), so by 0-initializing the vector and only putting things into the extend_token slots we get the correct result.

govind-ramnarayan: > Okay, yeah I got rid of this one while trying to make chunked prefill work. I can put it back. But I think it does work; here is my reasoning why:

> 
> The reasoning here was that the offset for prefill positions is 0 because they have not run before so they have no offset to gain (also see the logic before for populating increment - it populates num_prefill:), so by 0-initializing the vector and only putting things into the extend_token slots we get the correct result.

But let me remove it and see if it helps with things.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
govind-ramnarayan: Okay this is fair. I was just mimicking the test_accuracy() earlier because I figured this takes even more memory. But maybe that intends to skip pre_blackwell due to NVFP4 instead of gating on memory.

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
govind-ramnarayan: The main reason why I needed this was because the loop changes the metadata to go from prefill/extend to all-generation, including the metadata. Then when you run again, you hit this:

https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py#L933

I'll see if there is something less that will work - for instance just resetting the batch_info should work.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
lucaslie: Let's go with the two special cases and then fall back to the generic option later on if you want to remove that host sync. If we know that pre-fill is zero always, we can pass a flag to that function, and that gets controlled by whoever is calling the function.

tensorrt_llm/_torch/auto_deploy/utils/node_utils.py
govind-ramnarayan: I'm not sure what else to do at the moment - maybe I can chat with you and Greg about it. I can also put a task and a TODO so we can revisit it after. Maybe not worth delaying if sharding is changing soon anyway?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: The accuracy improves a tiny bit but does not substantially close on `torch-simple` accuracy with disabled padding. I've audited the logic related to overlap scheduler + padding and it looks correct. I think we'll need a dedicated debug pass after this merges to resolve the accuracy drop issue here.

tensorrt_llm/_torch/auto_deploy/utils/node_utils.py
govind-ramnarayan: I've updated the logic to detect the embd size for draft models using the output dimension, which is a lot cleaner. 

I still need something to find the input to eagle hidden layers properly in the current drafting setup, but "MTP" is basically clean with this change since MTP just does a concatenation followed by a normal attention, so existing sharding utils find the draft layer just fine.

I've made a task to clean it up - I think the need for this will disappear with the combination of new sharding IR + multiple KV cache managers, we can prioritize MTP / Eagle models and then deprecate this special-casing.

ruodil (approved):

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
```diff
@@ -217,6 +228,11 @@ def capture_graph(self, get_args_kwargs: GetArgsKwargsForBatchSize, batch_sizes:
             out = self.model(*args, **kwargs)
         self._out_buffer_flat, self._out_spec = tree_flatten(out)
 
+        # model.forward() sometimes modifies the cache_seq_interface directly
+        # Example: Eagle switches the batch from extend-only to generate-only mode during the
+        # draft loop. We want to restore the args to satisfy invariants needed by model.forward()
+        needs_refresh = "cache_seq_interface" in kwargs
```
lucaslie: Let's just always do a refresh to keep things simple.

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
```diff
@@ -249,9 +265,19 @@ def capture_graph(self, get_args_kwargs: GetArgsKwargsForBatchSize, batch_sizes:
             ]
             args, kwargs = self._in_spec.unflatten(inputs_truncated + args_static)
 
+            refresh_args_kwargs = None
+            if needs_refresh:
+
+                def refresh_args_kwargs(_bs: int = bs) -> None:
+                    get_args_kwargs(_bs)
```
lucaslie: Okay, so the way it's written right now, it's really just going to refresh the static buffers. That's fine because that's all we need, but we should clearly indicate that. Let's call the function something like `refresh_args_static`.

tensorrt_llm/_torch/auto_deploy/compile/backends/torch_cudagraph.py
```diff
@@ -249,9 +265,19 @@ def capture_graph(self, get_args_kwargs: GetArgsKwargsForBatchSize, batch_sizes:
...
+
             # capture graph for truncated inputs
             combined_shape = sum((tuple(input.shape) for input in inputs_truncated), start=())
-            self.cudagraphs[combined_shape] = self._capture_one_graph(*args, **kwargs)
+            self.cudagraphs[combined_shape] = self._capture_one_graph(
+                *args,
+                refresh_args_kwargs=refresh_args_kwargs,
+                **kwargs,
+            )
```
lucaslie, line 276: Can you change the signature of this function? It's a little confusing now. Something like:
```
args=args,
kwargs=kwargs,
refresh_args_static=refresh_args_static
```

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -502,6 +511,20 @@ def get_num_tokens_to_gather(self) -> int:
...
+        self._batch_info[12] = int(is_speculative)
+        self._batch_info[13] = max_draft_len
+
+    # --- spec-decoding info (slots 12-13) readers ---
+
+    def is_speculative(self) -> bool:
+        return bool(self._batch_info[12])
+
+    def get_max_draft_len(self) -> int:
+        return int(self._batch_info[13])
```
lucaslie: I don't see any of this being used. Can you remove it?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -935,18 +964,32 @@ def set_max_num_tokens_sample(self) -> None:
         input_ids = torch.ones(self.max_batch_size, seq_len, dtype=torch.int)
         self.set_example_sequence(input_ids)
 
-    def set_generate_only_batch(self, batch_size: Optional[int] = None) -> None:
-        """Set an example sequence for generate-only batch."""
+    def set_capture_batch(self, seq_len: int, batch_size: Optional[int] = None) -> None:
```
lucaslie: Wouldn't it be more appropriate to make the argument max_draft_len and set it to be zero by default?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1355,9 +1408,18 @@ def offset_pos_and_cache_(self, offset: torch.Tensor) -> None:
         # position_ids is per-token while offset is per-sequence; expand if needed
         if self.is_generate_only:
             offset_for_pos_ids = offset
+        elif self.is_extend_only:
+            _, num_extend, _ = self.batch_info.get_num_sequences()
+            _, num_extend_tokens, _ = self.batch_info.get_num_tokens()
+            if num_extend > 0:
+                tokens_per_seq = num_extend_tokens // num_extend
+                offset_for_pos_ids = offset[:num_extend].repeat_interleave(tokens_per_seq)
```
lucaslie:
```suggestion
            tokens_per_seq = num_extend_tokens // num_extend
            offset_for_pos_ids = offset[:num_extend].repeat_interleave(tokens_per_seq)
```
Don't need that extra if by definition of self.is_extend_only

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1385,7 +1447,14 @@ def offset_with_new_lens_(self, new_lens_ungathered: torch.Tensor) -> None:
...
         gather_slot_idx = self.get_arg("_gather_slot_idx", truncate=True)
-        increment[num_prefill:] = new_lens_ungathered[gather_slot_idx] - 1
+        num_overlap = gather_slot_idx.numel()
+        # AD overlap mode packs real overlap-carried non-prefill sequences first. CUDA-graph dummy
+        # requests are appended later and therefore remain in the trailing zero-increment region.
+        # We deliberately only write gather_slot_idx entries for non-dummy requests, so that we
+        # can apply them to the prefix of "increment" here.
+        increment[num_prefill : num_prefill + num_overlap] = (
+            new_lens_ungathered[gather_slot_idx] - 1
+        )
```
lucaslie, line 1452: Why is there still a diff here compared to the previous version? Shouldn't that diff be gone per our discussion?

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -155,6 +162,21 @@ def __init__(self, config: PretrainedConfig, model_type: str):
...
+            or config_dict.get("torch_dtype")
+        )
+        if target_dtype is not None and current_dtype != target_dtype:
+            ad_logger.info(
+                f"Setting AutoDeploy Eagle draft torch_dtype to target dtype {target_dtype} for shared KV "
+                f"cache handling. Draft dtype before change: {current_dtype}",
+            )
+            target_dtype_str = torch_dtype_to_str(target_dtype)
+            config_dict["dtype"] = target_dtype_str
+            config_dict["torch_dtype"] = target_dtype_str
```
lucaslie: Why do you have to hard code this here? Why can't we just use the existing mechanism of just setting the `dtype` in the `model_kwargs`?

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -810,6 +835,11 @@ def apply_lm_head(self, hidden_states: torch.Tensor) -> torch.Tensor:
 
     def sample_greedy(self, logits: torch.Tensor) -> torch.Tensor:
         ret = torch.argmax(logits, dim=-1)
+        # Broadcast rank 0's sampled tokens to all ranks. We have observed slight
+        # differences in logits across ranks. This can cause different acceptance patterns,
+        # inconsistent input_pos, and a hang. This synchronizes every
+        # validation step to prevent this. See: Issue #13134 in nvidia/TensorRT-LLM.
```
lucaslie: Please provide the link instead of just a comment about the issue number.

tensorrt_llm/_torch/auto_deploy/models/eagle.py
```diff
@@ -56,6 +56,10 @@ class EagleDrafterFactory(AutoModelForCausalLMFactory):
     (e.g., "llama") along with Eagle-specific fields like draft_vocab_size.
     """
 
+    def __init__(self, *args, target_dtype: Optional[torch.dtype] = None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.target_dtype = target_dtype
```
lucaslie: See my other comment. Ideally, we can avoid these hard-coded dtype hacks

tensorrt_llm/_torch/auto_deploy/utils/node_utils.py
lucaslie: I see. I don't think this code is maintainable or usable in the long run, but let's keep it for now. You should see if this code can be deprecated as part of the transition to the new sharding infrastructure. Please do check that.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: That was not my understanding. The dummies are already extend requests - they're just not in gather_slot_idx and this is intentional. We explicitly leave them out here, and this logic is the same across extend and decode: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py#L799

I think this logic is good, it makes sense that we don't need to gather dummy tokens from the previous run, and so we also should not need to adjust their offsets.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: I think either is fine, I can do this.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: Used here: https://github.com/NVIDIA/TensorRT-LLM/pull/12745/changes/BASE..73b096e0378a8f4a3b88a3c9367549ef13e024b8#diff-d686e5e3887e7b895ec01bc8f55b8bb80f9dc138a14ea620f19eff67224c34abR399

Just a readable accessor.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: I thought it was actually a good correctness condition to have - the issue with differing dtypes in the same KV cache manager causing inconsistencies was fairly painful to find. I can just put the check in model validation, though.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
lucaslie: looks good to me

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -268,6 +275,21 @@ def _identify_managed_kv_resources(
                 kv_ref = handler
             if handler == kv_ref:
                 kv_managed[name] = handler
+            else:
+                ad_logger.warning(
```
lucaslie: To get rid of the assertion and function argument passing of the dtype for draft versus target model, let's just do an assertion here instead. We check if we have trtllm attention backend in which case we don't allow unmanaged kv resources.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
lucaslie: Looks good to me. let's keep it like is

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -62,6 +62,7 @@ def __init__(
         kv_cache_config: Optional[KvCacheConfig] = None,
         vocab_size_padded: Optional[int] = None,
         spec_config=None,
+        attn_backend: Optional[str] = None,
```
lucaslie, line 65: How about we just give it an argument, like `allow_unmanaged_kv_caches=True`? We can set this argument when initializing the class. It seems cleaner than passing in attention backend. attn backend does not have a meaning and should not have a meaning within this class.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -453,6 +453,7 @@ def build_from_config(
             max_num_tokens=ad_config.max_num_tokens,
             vocab_size_padded=factory.vocab_size_padded,
             spec_config=ad_config.speculative_config,
+            attn_backend=ad_config.attn_backend,
```
lucaslie, line 456: Related to my other comment, we can have something like `ad_config.allow_unmanaged_kv_caches`, which is a property of the config class that we just set in `llm_args.py`. This way we keep the configuration clean.

lucaslie (approved):

lucaslie:
Looks great, awesome work, congratulations. Just left one final comment for you to please address.

tensorrt_llm/_torch/auto_deploy/utils/node_utils.py
lucaslie, line 1: @greg-kwasniewski1, please take a look

tensorrt_llm/_torch/auto_deploy/transform/library/sharding.py
lucaslie, line 1: @greg-kwasniewski1 , please take a look

tensorrt_llm/_torch/auto_deploy/shim/interface.py
govind-ramnarayan: Yeah agreed! Will do

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: ok I named it `ad_config.requires_uniform_kv_caches`. It covers both the fact that the KV caches all need to be managed (they will not be if they are not uniform) and that managed KV caches share the same block offset multiplier (if they don't, not uniform). This is the strong condition we want to make sure that our block offsets are consistent across layers for TRTLLM Attention.

govind-ramnarayan:

govind-ramnarayan:

govind-ramnarayan:

govind-ramnarayan:

govind-ramnarayan:

PR #12746, opened 2026-04-03
[None][feat] support multiple model names in --served_model_name

nvyutwu:
## Summary
- Allow `--served_model_name` to be specified multiple times, matching vLLM-style explicit served-name override semantics
- The first name is the primary served model name; additional names are accepted request aliases
- `/v1/models` returns all registered names; response `model` fields use the primary served model name

### Served model name semantics
`--served_model_name` follows vLLM-style override semantics:

- If `--served_model_name` is not specified, the launch model path is used as the served model name fallback. For local model directories, HTTP serving normalizes the basename as the primary name while preserving the original path as an accepted alias.
- If `--served_model_name` is specified, only the provided name(s) are valid request model names. The original model path is not implicitly added as an alias.
- When multiple names are provided, the first name is primary and is used in response metadata. Additional names are accepted request aliases, but responses still report the primary name.

### Changes
- **`serve.py`**: `--served_model_name` gains `multiple=True` (click); `launch_server`/`launch_grpc_server` accept `Sequence[str]`; normalizes to list internally
- **`openai_server.py`**: `__init__` accepts `Union[str, Sequence[str]]`; stores `self.model` (primary) and `self.served_model_names` (all); `get_model()` returns all names; `_check_model()` accepts requests addressed to any registered alias and rejects unknown explicit model names

### Usage
```bash
# Default fallback: served model name comes from the launch model path.
trtllm-serve /models/foo

# Explicit override with repeated flags.
trtllm-serve /models/foo \
  --served_model_name my-model \
  --served_model_name alias1

# Explicit override with comma-separated names.
trtllm-serve /models/foo --served_model_name my-model,alias1,alias2

# Quote comma-separated values if spaces are included after commas.
trtllm-serve /models/foo --served_model_name "my-model, alias1, alias2"
```

## Test plan
- [x] Unit tests: `pytest tests/unittest/llmapi/apps/test_served_model_name.py -v` passes 30 tests
- [x] Unit coverage includes:
  - Dedup and order preservation across multiple `--served_model_name` values
  - Alias request validation for known aliases, unknown aliases, `None`, and empty string
  - Directory-path → basename primary normalization while preserving original path as an alias
  - Single-string / empty input fallback for backward compatibility
  - Comma-separated and repeated-flag CLI parsing
- [x] Smoke test: `trtllm-serve <model> --served_model_name foo --served_model_name bar`
  - `GET /v1/models` lists both `foo` and `bar`
  - `POST /v1/chat/completions` with `model: "bar"` returns HTTP 200 and generates content; response `model` is `foo` because the first served name is primary
  - `POST /v1/chat/completions` with `model: "unknown"` returns HTTP 404 with `NotFoundError`

### DIFF
--- a/tensorrt_llm/serve/openai_server.py
+++ b/tensorrt_llm/serve/openai_server.py
@@ -765,20 +765,24 @@ class OpenAIServer:
             names: Sequence[str]) -> tuple[str, list[str]]:
         """Return (primary, aliases) from a sequence of names.
 
-        If the primary name points to an existing directory, its basename is used.
-        Duplicate aliases are removed while preserving order.
+        If the first name points to an existing directory, its basename
+        becomes the primary and the original path is kept as an alias so
+        clients that launched the server by path can still address it.
+        Duplicates are removed while preserving order.
         """
-        primary = names[0]
-        model_dir = Path(primary)
+        first = names[0]
+        model_dir = Path(first)
         if model_dir.exists() and model_dir.is_dir():
-            primary = model_dir.name
-        seen = {primary}
-        aliases = [primary]
-        for n in names[1:]:
-            if n not in seen:
+            candidates = [model_dir.name, first, *names[1:]]
+        else:
+            candidates = list(names)
+        seen: set[str] = set()
+        aliases: list[str] = []
+        for n in candidates:
+            if n and n not in seen:
                 seen.add(n)
                 aliases.append(n)
-        return primary, aliases
+        return aliases[0], aliases
 
     def _is_model_supported(self, model_name: Optional[str]) -> bool:
         """Return True if ``model_name`` is unset or matches a registered name."""
@@ -1823,6 +1827,9 @@ class OpenAIServer:
 
         Follows the OpenAI Images API specification for image generation.
         """
+        error_check_ret = self._check_model(request)
+        if error_check_ret is not None:
+            return error_check_ret
         try:
             image_id = f"image_{uuid.uuid4().hex}"
             params = parse_visual_gen_params(request, image_id)
@@ -1888,6 +1895,9 @@ class OpenAIServer:
         Follows the OpenAI Images API specification for image editing.
         Creates an edited or extended image given an original image and a prompt.
         """
+        error_check_ret = self._check_model(request)
+        if error_check_ret is not None:
+            return error_check_ret
         try:
             image_id = f"image_{uuid.uuid4().hex}"
             params = parse_visual_gen_params(request, image_id)
@@ -1951,6 +1961,9 @@ class OpenAIServer:
         try:
             # Parse request based on content-type
             request = await self._parse_video_generation_request(raw_request)
+            error_check_ret = self._check_model(request)
+            if error_check_ret is not None:
+                return error_check_ret
 
             # Resolve the video encode format (mp4/avi/auto)
             resolved_fmt, resolved_ext = resolve_video_format(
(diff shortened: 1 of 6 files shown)


hchings:
Hi @nvyutwu, could you resolve conflicts? I'll review within this week. Thanks!

nvyutwu:
@hchings Conflicts resolved — rebased onto latest `main`. Thanks for reviewing!

tensorrt_llm/serve/openai_server.py
```diff
@@ -764,8 +760,63 @@ async def version(self) -> JSONResponse:
...
+
+        Falsy values (``None`` and ``""``) are treated as "unspecified" and
+        accepted — clients that omit the field still get a response. Only
+        explicit unknown names are rejected by :meth:`_check_model`.
+        """
+        if not model_name:
+            return True
+        return model_name in self.served_model_names
+
+    def _check_model(self, request: Any) -> Optional[Response]:
```
hchings, line 806: Other than aligning trtllm's server behavior with vllm, is there a customer-facing reason that we are now making this strict check? Previously, `model` name in request is ignored and always return `200`. 

But with this MR, we still allow users to not pass `--served_model_name` during serving (fall back to `falls back to [llm_args["model"]]`), but then we now have strict check for model name in requests. This behavior can be quite implicit to users. But I might be missing some context on the motiation or customer needs.

tensorrt_llm/serve/openai_server.py
```diff
@@ -764,8 +760,63 @@ async def version(self) -> JSONResponse:
...
+
+        A bare ``str`` is accepted as a convenience for callers that still
+        pass a single name; it is treated as ``[name]``.
+        """
+        if isinstance(names, str):
+            names = [names]
+        first = names[0] if names else ""
+        model_dir = Path(first) if first else None
+        if model_dir is not None and model_dir.exists() and model_dir.is_dir():
+            candidates = [model_dir.name, first, *names[1:]]
```
hchings, line 781: nit: Is this the right design? 

if the user explicitly typed --served_model_name xx, maybe they wanted xx to be the primary. But in this case, the first user specified name will be silently dropped as alias. And now users must know to pay attention to the "Registered served model name" logging to discover this.

tests/unittest/llmapi/apps/_test_openai_metrics.py
```diff
@@ -26,7 +26,7 @@ def llm():
 @pytest.fixture(scope="module")
 def client(llm):
     app_instance = OpenAIServer(llm,
-                                model=llama_model_path,
+                                model=[llama_model_path],
```
hchings, line 29: Since you use `Union[str, Sequence[str]]`, we don't need to change this.

tests/unittest/_torch/visual_gen/test_trtllm_serve_endpoints.py
```diff
@@ -161,7 +161,7 @@ def _create_server(generator: MockVisualGen, model_name: str = "test-model") ->
     with patch("tensorrt_llm.serve.openai_server.VisualGen", MockVisualGen):
         server = OpenAIServer(
             generator=generator,
-            model=model_name,
+            model=[model_name],
```
hchings, line 164: Since you use `Union[str, Sequence[str]]`, we don't need to change this.

zhenhuaw-me:
I am not sure about the motivation of this PR. @JunyiXu-nv can decided whether we will take this or now. Thanks!

Issue #12617, opened 2026-03-31
[Bug]: Ngram speculative decoding is incompatible with LoRA


AlessioNetti:
### System Info

Nvidia A40 GPUs, AMD Epyc 7352 CPU, NGC 26.01 PyTorch image (CUDA 13.1.1, PyTorch 2.10.0a0), TensorRT-LLM 1.3.0rc9.

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

I created a small reproducer script [here](https://gist.github.com/AlessioNetti/353abd3c5a2dd7d8664ff03fa39844ac), which attempts to enable LoRA (using a single adapter) and Ngram speculative decoding at the same time. It can be used as follows:

```
python torch_lora_spec_dec_reproducer.py --model_path <BASE_MODEL_PATH> --lora_path <LORA_PATH> --tp_size 1 --pp_size 1
```

To reproduce the issue, we used the Qwen3 8B model coupled with [this](https://huggingface.co/ldsjmdy/Qwen3-8B-FreeLM-LoRA) LoRA adapter from HuggingFace.

### Expected behavior

We expect inference to work as normal when combining LoRA with NGram speculative decoding.

### actual behavior

The following error is produced:

```
  File "/virtualenv/lib/python3.12/site-packages/tensorrt_llm/_torch/models/modeling_qwen3.py", line 166, in forward
    hidden_states = self.self_attn(
  ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1787, in _call_impl
    return forward_call(*args, **kwargs)
^^^^^^^^^^^^^^^
  File "/virtualenv/lib/python3.12/site-packages/tensorrt_llm/_torch/modules/attention.py", line 853, in forward
    qkv_lora = self.splitted_qkv_lora(hidden_states, lora_params,
^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
[...]

PR #12771, opened 2026-04-06
[#12617][fix] fix LoRA for Qwen2 and Mistral decoder layers

edenfunf:
## What

Two PyTorch-backend decoder layers never forward `layer_idx` to their `GatedMLP`,
and one of them also drops `lora_params` before calling the MLP:

| Model | `GatedMLP(layer_idx=...)` | `lora_params` reaches MLP | Symptom with LoRA |
|---|---|---|---|
| Qwen2 (`QwenDecoderLayer`) | missing | yes | crashes |
| Mistral (`MistralDecoderLayer`) | missing | **no** | silently wrong |

`GatedMLP.forward_lora()` guards on the index it needs to look up per-layer adapter
weights:

```python
assert self.layer_idx is not None, "layer_idx is required for lora"
```

**Qwen2** reaches that assert and dies:

```
tensorrt_llm.executor.utils.RequestError: layer_idx is required for lora
```

**Mistral** never gets there. `MistralModel.forward()` passes `lora_params` down to
the decoder layer, which forwards it to `self_attn`, but the MLP call dropped it:

```python
hidden_states = self.mlp(hidden_states)      # **kwargs missing
```

So `GatedMLP.forward()` sees `lora_params=None`, `bool(lora_params)` is false,
`forward_lora()` is never entered, and the assert never fires. Attention LoRA still
applies, so generation looks fine — **the MLP adapter weights are silently ignored**.
That failure mode produces no error at all, which is why it went unnoticed.

## Why

Reported by @AlessioNetti in #12617 as a secondary issue. Qwen3 had the same
`layer_idx` bug and was fixed independently in #12785; that PR also added
`test_qwen3_sanity.py`. Qwen2 and Mistral were left behind. `LlamaDecoderLayer`,
`Phi3DecoderLayer`, and `Gemma3DecoderLayer` already pass `layer_idx`.

## How

- `modeling_qwen.py` — `QwenDecoderLayer`: pass `layer_idx=layer_idx` to `GatedMLP`
- `modeling_mistral.py` — `MistralDecoderLayer`: pass `layer_idx=layer_idx` to `GatedMLP`
- `modeling_mistral.py` — `MistralDecoderLayer.forward()`: `self.mlp(hidden_states, **kwargs)`

The `**kwargs` change matters: without it the `layer_idx` fix is dead code on Mistral,
because `lora_params` still never reaches the MLP. This mirrors the `**kwargs` half of
#12785 (which applied it to `modeling_qwen3_moe.py`).

## Test

New `tests/unittest/_torch/modules/tests_lora_modules/test_qwen2_mistral_sanity.py`,
covering Qwen2-0.5B-Instruct and mistral-7b-v0.1. The architecture-agnostic helpers
from `test_qwen3_sanity.py` are extracted into `lora_sanity_utils.py` and shared by
both files rather than duplicated; `test_qwen3_sanity.py` is otherwise unchanged in
behaviour.

**The adapters deliberately target only the MLP modules.** Attention LoRA works on
both models either way, so an adapter that also covered attention would still change
the output and the tests would pass with the MLP path broken — which is exactly the
Mistral bug. With MLP-only adapters, `lora_params` that never reaches the MLP means no
LoRA is applied at all, and the output comparison catches it.

The directory is already listed in `l0_b200.yml`, `l0_b300.yml`, `l0_h100.yml` and
`l0_gb300_multi_gpus.yml`, so the new file is picked up automatically.

Local red/green verification (RTX 5070, TensorRT-LLM 1.2.1) is attached in a separate
comment, along with an explicit statement of what that run does and does not cover.

## Scope

Seven other `GatedMLP` call sites also omit `layer_idx`, but none of them are affected:
HunYuan (dense and MoE) explicitly rejects LoRA (`assert lora_params is None, "LORA is
not supported for HunYuanAttention"`); GLM, Mllama and Eagle3 have no PEFT LoRA path at
all; DeepSeekV3's `lora` references are `kv_lora_rank`, i.e. MLA's low-rank projection,
unrelated to adapters; and Laguna only carries `lora_params` in an attention `forward`
signature that the model never feeds. Qwen2 and Mistral were the only two models where
LoRA is wired up but the MLP does not receive it.

## Dev Engineer Review

- `QwenDecoderLayer` and `MistralDecoderLayer` now pass `layer_idx` to `GatedMLP`.
- Mistral now forwards MLP call keyword arguments, including `lora_params`.
- These changes address Qwen2 LoRA initialization errors and Mistral MLP adapter omission.
- The current diff contains only two source files. No public API declarations or configuration files changed.
- Test execution status is not independently available.

## QA Engineer Review

No test changes.

## Per-File QA Perspective

- `tensorrt_llm/_torch/models/modeling_mistral.py`: Verify Mistral decoder-layer construction and MLP LoRA parameter forwarding. Also verify that the broader 64-line change does not alter unrelated decoder behavior.
- `tensorrt_llm/_torch/models/modeling_qwen.py`: Verify that each Qwen2 decoder MLP receives its layer index and that LoRA inference no longer raises the missing-`layer_idx` error.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_mistral.py
+++ b/tensorrt_llm/_torch/models/modeling_mistral.py
@@ -1,3 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 import copy
 import dataclasses
 from typing import Any, Dict, List, Tuple
@@ -134,6 +148,7 @@ class MistralDecoderLayer(DecoderLayer):
             bias=False,
             dtype=config.torch_dtype,
             config=model_config,
+            layer_idx=layer_idx,
         )
         self.input_layernorm = RMSNorm(
             hidden_size=config.hidden_size,
--- a/tensorrt_llm/_torch/models/modeling_qwen.py
+++ b/tensorrt_llm/_torch/models/modeling_qwen.py
@@ -1,3 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 from typing import Optional, Tuple
 
 import torch
@@ -134,6 +148,7 @@ class QwenDecoderLayer(DecoderLayer):
             bias=config.mlp_bias if hasattr(config, 'mlp_bias') else False,
             dtype=config.torch_dtype,
             config=model_config,
+            layer_idx=layer_idx,
         )
         self.input_layernorm = RMSNorm(hidden_size=config.hidden_size,
                                        eps=config.rms_norm_eps,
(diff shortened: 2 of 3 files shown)


SimengLiu-nv:
Please add corresponding tests.

SimengLiu-nv (changes requested):
Please add corresponding tests.

edenfunf:
> 請增加相應的測試。

Thanks for the reminder. I added the tests in `a2a23a5`.

`tests/unittest/_torch/modules/tests_lora_modules/test_qwen2_mistral_sanity.py`

I also extracted the shared helpers from `test_qwen3_sanity.py` into `lora_sanity_utils.py`.

Tested on TRT-LLM 1.2.1 + RTX 5070, with the fix applied to the installed package:

| test | before fix | after fix |
| --- | --- | --- |
| `test_qwen2_bf16_lora` | FAILED: `RequestError: layer_idx is required for lora` | **PASSED** |
| `test_mistral_bf16_lora` | FAILED: `AssertionError: LoRA outputs identical to base model` | **PASSED** |

`2 failed in 21.39s` → `2 passed in 13.48s`

The two failures cover the two separate bugs. Qwen2 reaches the assert, while Mistral never does because `lora_params` never reaches the MLP, so it produces the same output as the base model.

A couple of notes:

- The adapters target **MLP modules only**. If attention modules are included, the output changes anyway, so the Mistral test would pass even with the bug still present.
- I used a small stand-in Mistral checkpoint locally since the real 7B model doesn't fit in 12GB VRAM. CI on H100/B200 clears the 40GB guard and runs the real model.

The directory is already included in `l0_b200/b300/h100/gb300`, so no test-list change is needed.

litaotju:
These models are pretty old. 

Any one asking LORA support for these? 

If not, I intend to not to merge these.

PR #10188, opened 2025-12-22
[None][feat] Support VLM part for Mistral Large 3

byshiue:
## Release Notes

* **New Features**
  * Added multimodal (vision) support for Mistral models, enabling image processing alongside text inputs.
  * Implemented improved tokenizer with enhanced compatibility and special token handling.
  * Extended multimodal input processing capabilities for Mistral3 model type.

* **Documentation**
  * Updated example guide with multimodal usage instructions and commands.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

PR #12788, opened 2026-04-07
[TRTLLM-11375][feat] Add Kimi K2.5 multimodal vision support

tianyuxbear:
## Summary by CodeRabbit

* **New Features**
  * Native support for Kimi K2.5 multimodal model with image and video preprocessing, vision encoding, and fusion into text generation.
  * Multimodal input processor handling media placeholders and expanded tokenization.

* **Enhancements**
  * Improved handling of “thinking” / reasoning excerpts in generation outputs (conditional stripping/logging).

* **Tests**
  * Added unit, integration, and accuracy tests plus test-list entries and benchmarks for Kimi K2.5.

## Description

Add full vision-language support for Kimi K2.5 (`moonshotai/Kimi-K2.5`) in the PyTorch backend.

Previously only the text backbone was supported (via a thin wrapper in `modeling_deepseekv3.py`). This PR adds the complete multimodal pipeline: native MoonViT3d vision encoder, PatchMergerMLP projector, image/video input processor, and the composite `KimiK25ForConditionalGeneration` model that wires vision encoding into the DeepSeek-V3 text backbone.

### Model support
- **`tensorrt_llm/_torch/models/modeling_kimi_k25.py`** (new): MoonViT3d encoder, PatchMergerMLP projector, `KimiK25InputProcessor` (image + video), `KimiK25ForConditionalGeneration`
- **`tensorrt_llm/_torch/models/__init__.py`**: Register new model
- **`tensorrt_llm/_torch/models/modeling_deepseekv3.py`**: Remove old text-only stub

### Serving / evaluation
- **`tensorrt_llm/llmapi/reasoning_parser.py`**: Register `kimi_k25` reasoning parser
- **`tensorrt_llm/evaluate/lm_eval.py`**: `_strip_thinking()` to strip `<think>` tags for accuracy benchmarks

### Limitations
- NVFP4 quantization only (BF16 not yet validated)
- Multimodal disaggregated serving not yet supported (TODO)
- Video input requires `decord` or `cv2` for frame decoding

## Test Coverage

- **Unit tests**: `tests/unittest/_torch/modeling/test_modeling_kimi_k25.py` — structure, config parsing, input processor, E2E smoke test (4x B200)
- **MMMU accuracy**: `tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py::TestKimi_K25` — NVFP4, tp8/ep8/attn_dp/dep8, reference score 82.00
- **CI entries**: Added to `l0_dgx_b200.yml` (pre-merge) and `llm_function_core.txt` (QA regression)
- Existing text-only K2.5 tests unaffected

## PR Checklist

## GitHub Bot Help


xinhe-nv (approved):

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+        sample_fps: Target sampling frame rate.
+        timestamp_mode: Format for timestamp strings.
+
+    Returns:
+        Tuple of (chunks, video_prompt) where chunks is a list of dicts
+        ``{"type": "video_chunk", "video_chunk": [PIL.Image, ...],
+        "prompt": "<timestamp><|media_begin|>video<|media_content|><|media_pad|><|media_end|>"}``
+        and video_prompt is the concatenated prompt for all chunks.
+    """
+    from datetime import datetime, timezone
```
yechank-nvidia: nit: Can you please move the import to the top?

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+    """
+    freqs_cis = freqs_cis.unsqueeze(-2)
+    xq_ = torch.view_as_complex(xq.float().view(*xq.shape[:-1], -1, 2))
+    xk_ = torch.view_as_complex(xk.float().view(*xk.shape[:-1], -1, 2))
+    xq_out = torch.view_as_real(xq_ * freqs_cis).flatten(-2)
+    xk_out = torch.view_as_real(xk_ * freqs_cis).flatten(-2)
+    return xq_out.type_as(xq), xk_out.type_as(xk)
+
+
+def _vision_flash_attention(
```
yechank-nvidia: Is there any specific reason you tried not using TRT-LLM's Attention backend of FlashInfer's backend? 

AFAIK, we can not support TP > 1 or other parallelisms on ViT if we are using FlashAttention.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+        text_hidden_size: int,
+        merge_kernel_size: Tuple[int, int] = (2, 2),
+        ln_eps: float = 1e-5,
+    ) -> None:
+        super().__init__()
+        kh, kw = merge_kernel_size
+        self.merged_dim = mm_hidden_size * kh * kw
+        self.pre_norm = nn.LayerNorm(mm_hidden_size, eps=ln_eps)
+        self.proj = nn.Sequential(
+            nn.Linear(self.merged_dim, self.merged_dim),
```
yechank-nvidia: We prefer to use TRT-LLM's Linear for compatibility

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+        return torch.cat(
+            [
+                self.freqs_cis[:h, :w].reshape(-1, self.dim // 2).repeat(t, 1)
+                for t, h, w in grid_thws.tolist()
+            ],
+            dim=0,
+        )
+
+
+class _VisionMLP(nn.Module):
```
yechank-nvidia: I think we already have [MLP](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/modules/mlp.py)

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+        self.width = width
+        self.num_frames = num_frames
+        self.dim = dim
+        self.weight = nn.Parameter(torch.empty(height, width, dim))
+        self.register_buffer(
+            "time_weight",
+            torch.from_numpy(_get_1d_sincos_pos_embed(dim, num_frames)).float().unsqueeze(1),
+            persistent=False,
+        )
+        nn.init.normal_(self.weight)
```
yechank-nvidia, line 416: Can I ask why we need `init`?

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+
+    def get_num_tokens_per_image(self, *, image, **kwargs) -> int:
+        """Calculate the number of tokens for a single image.
+
+        Delegates to the HF KimiK25VisionProcessor's
+        ``media_tokens_calculator`` which accounts for Moon-Patch dynamic
+        tiling, spatial merge kernel, and patch size.
+
+        Args:
+            image: PIL Image to compute the token count for.
```
yechank-nvidia: nit) It can be PIL Image or Torch.Tensor

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+        # defer creation to load_weights() where meta init is complete.
+        self.mm_encoder = None
+        if not DISAGG:
+            try:
+                self.mm_encoder = KimiK25VisionModel(model_config)
+            except MetaInitException:
+                logger.info("Vision encoder deferred to load_weights() (MetaInitMode active)")
+
+        # --- LLM backbone (DeepSeek-V3, created directly) ---
+        from .modeling_deepseekv3 import DeepseekV3ForCausalLM
```
yechank-nvidia: Can you move import to the top of the file?

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+                    "found in input_ids — embeddings discarded."
+                )
+                inputs_embeds = None
+        else:
+            input_ids, inputs_embeds = fuse_input_embeds(
+                self.llm.model.embed_tokens,
+                input_ids,
+                mm_embeds,
+                **kwargs,
+            )
```
yechank-nvidia:
```suggestion
  mm_token_ids = None
  fuse_kwargs = kwargs

  if len(mm_embeds) > 0:
      placeholder_id = self._media_placeholder_token_id
      num_mm_in_ids = int((input_ids == placeholder_id).sum().item())
      if num_mm_in_ids == 0:
          logger.warning(
              "Vision embeddings computed but no placeholder tokens "
              "found in input_ids — embeddings discarded."
          )
          mm_embeds = []
      else:
          total_embed = sum(e.shape[0] for e in mm_embeds)
          if num_mm_in_ids < total_embed:
              logger.warning(
                  "Truncating %d vision embeddings to %d placeholder tokens",
                  total_embed, num_mm_in_ids,
              )
              mm_embeds = [torch.cat(mm_embeds, dim=0)[:num_mm_in_ids]]
          mm_token_ids = torch.tensor(
              [placeholder_id], dtype=input_ids.dtype, device=input_ids.device,
          )
          fuse_kwargs = {
              k: v for k, v in kwargs.items()
              if k not in ("mm_token_indices", "text_token_indices")
          }

  input_ids, inputs_embeds = fuse_input_embeds(
      self.llm.model.embed_tokens, input_ids, mm_embeds,
      mm_token_ids=mm_token_ids, **fuse_kwargs,
  )
```

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1486 @@
...
+                    device=input_ids.device,
+                )
+                total_embed = sum(e.shape[0] for e in mm_embeds)
+                if num_mm_in_ids < total_embed:
+                    logger.warning(
+                        "Truncating %d vision embeddings to %d placeholder tokens",
+                        total_embed,
+                        num_mm_in_ids,
+                    )
+                    mm_embeds = [torch.cat(mm_embeds, dim=0)[:num_mm_in_ids]]
```
yechank-nvidia: I'm not sure if this is expected or not. Normally, we think if num_mm_in_ids < total_embed, we consider it as a bug.

Is this kimi-k2.5's characteristics?

yechank-nvidia (commented):
Hi, thanks for the contribution.

Overall, I think the golden rule is re-use what TRT-LLM already have. I suggest to change Normalization, Linear and Attention layers to use our module for code maintainability.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: Done. Moved `from datetime import datetime, timezone`, `from PIL import Image`, and `import tempfile` to the top of the file. The local imports inside `_decode_video_to_chunks` and `_frames_to_chunks` have been removed.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: Good catch — switched the ViT attention to TRT-LLM's `Attention` backend. The new `_KimiK25VisionAttention` class subclasses `Attention` and applies Kimi's 2D RoPE before calling `forward_impl`. Also removed the `flash_attn_varlen_func` import and the `_vision_flash_attention` helper.

The encoder now constructs an `attn_metadata` per forward (built from `grid_thws` → cumulative seq lengths) and reuses the configured `attn_backend`, so TP/EP/attention-DP all work correctly on the ViT side.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: Done — `_PatchMergerMLP` now uses TRT-LLM's `Linear` with `TensorParallelMode.COLUMN` (first projection) and `ROW` (second projection), passing through `quant_config`, `mapping`, `skip_create_weights_in_init`, and `allreduce_strategy`. Under attention-DP we override `mapping` to TP=1 (via `_get_vision_tp_mapping`) so the projector stays per-rank.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: Done — `_VisionMLP` now subclasses `tensorrt_llm._torch.modules.mlp.MLP`, with `activation=_gelu_tanh` (HF uses GELU tanh approx) and `bias=True`. Weight key remapping is applied during load (`fc0/fc1` → `up_proj/down_proj`).

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: This mirrors the HF reference `Learnable2DInterpPosEmbDivided_fixed.reset_parameters()`:

```python
# HF modeling_kimi_k25.py L271-274
self.weight = nn.Parameter(torch.empty(height, width, dim))
...
self.reset_parameters()

def reset_parameters(self):
    nn.init.normal_(self.weight)
```

Since `torch.empty` doesn't initialize memory, `nn.init.normal_` is needed to avoid garbage values being read on any code path before checkpoint weights are loaded (e.g. dummy/meta-init paths). I left it in to keep parity with the reference; happy to remove if you prefer the more aggressive cleanup.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: Updated. Changed the type annotation to `Union[Image.Image, torch.Tensor]` and the docstring from "PIL Image" to "PIL image or Torch tensor".

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: Done. Moved `import copy`, `from .modeling_deepseekv3 import DeepseekV3ForCausalLM`, `from .checkpoints.base_weight_loader import ConsumableWeightsDict`, and `from .modeling_utils import filter_weights` to the top of the file.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: Adopted the single-call structure as suggested. Note that I dropped the `num_mm_in_ids < total_embed` truncation branch as a follow-up to your Comment 9 — agreed it's a bug, not K2.5-specific behavior.

The mismatch case is now handled by `fuse_input_embeds`'s built-in check at `modeling_multimodal_utils.py:382-387` which raises `ValueError`, matching the HF reference's `_merge_input_ids_with_image_features` behavior at HF L1001-1005 of `modeling_kimi_k25.py`.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: You're right — this is not a K2.5 characteristic. Removed the silent-truncation branch.

Mismatches between placeholder count and embedding count now fail loudly via `fuse_input_embeds`'s existing assertion (`raise ValueError("Multimodal token count mismatch: ...")`), which is consistent with the HF reference (`raise ValueError` at L1001-1005 of HF `modeling_kimi_k25.py`).

tensorrt_llm/evaluate/lm_eval.py
```diff
@@ -51,6 +51,114 @@
 LM_EVAL_DEFAULT_IMAGE_PLACEHOLDER = "<image>"
 
 
+def _strip_thinking(text: str) -> str:
```
2ez4bz: This kind of post-processing could very well be model-specific. Is there any way this could be encapsulated elsewhere than the base `LmEvalWrapper`?

tianyuxbear: Refactored — moved to `tensorrt_llm/evaluate/post_processing.py` and exposed as an opt-in `post_process_fn` hook on MultimodalLmEvalWrapper. 
Base LmEvalWrapper is untouched.

EmmaQiaoCh (approved):

tensorrt_llm/_torch/models/modeling_kimi_k25.py
```diff
@@ -0,0 +1,1687 @@
...
+    def forward(self, x: torch.Tensor, grid_thws: torch.Tensor) -> torch.Tensor:
+        freqs_cis = self.rope_2d.get_freqs_cis(grid_thws, x.device)
+        seq_lens = (grid_thws[:, 0] * grid_thws[:, 1] * grid_thws[:, 2]).tolist()
+        attn_metadata = self.prepare_attn_metadata(seq_lens, self.attn_metadata)
+        for block in self.blocks:
+            x = block(x, attn_metadata, freqs_cis)
+        return self.final_layernorm(x)
+
+
+class _PatchMergerMLP(nn.Module):
```
yechank-nvidia: nit) Is there any particular reason you added '_' for all Multimodal encoder part?

yechank-nvidia: As this is not helper, or private, I think it would be good to remove '_' on the class name.

tests/unittest/_torch/modeling/test_modeling_kimi_k25.py
```diff
@@ -0,0 +1,888 @@
...
+                add_generation_prompt=True,
+                tokenize=False,
+            )
+            inp = prompt_inputs(
+                {
+                    "prompt": text,
+                    "multi_modal_data": {"video": [video_tmp]},
+                }
+            )
+            # The framework's multimodal hashing (find_mm_token_lengths)
```
yechank-nvidia: Can you give me more context? We support video hashing.

tests/unittest/_torch/modeling/test_modeling_kimi_k25.py
```diff
@@ -0,0 +1,888 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
yechank-nvidia, line 1: Thanks for adding the test. But seems this is newly added by you with e2e smoke test. For multimodal test, we have the [skeleton](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/_torch/modeling/test_modeling_multimodal.py), it would be good if you can use it.

But, if there is specific reason you stick to this, I'm okay with this.

tensorrt_llm/_torch/models/modeling_kimi_k25.py
tianyuxbear: Done in 88f058b — removed `_` from all 8 vision encoder classes. Aligns with HF naming.

tests/unittest/_torch/modeling/test_modeling_kimi_k25.py
tianyuxbear: Framework hashing works for `VideoData` / `List[PIL.Image]`, but not raw file paths (`assert isinstance(item, list)` at multimodal.py:785). This test passes a file path on purpose to exercise `_decode_video_to_chunks` — the path used when trtllm-serve receives raw video uploads.

Resetting `multimodal_hashing_supported = None` triggers the first-try fallback at registry.py:929, which calls our input processor directly. In-code comment expanded in aae4940.

Happy to switch to pre-decoded frames if you'd prefer.

tests/unittest/_torch/modeling/test_modeling_kimi_k25.py
tianyuxbear: Sticking with the dedicated file:
1. It combines GPU-free input-processor unit tests with e2e smoke; the skeleton is e2e-oriented.
2. MMMU baseline (82.56) is already locked in via this test surface.
3. K2.5-specific paths (file-path video decoding, MoonViT3d 3D pos-emb) don't map cleanly to the skeleton's abstractions.

yechank-nvidia: Thanks for the detailed explanation! Just wanted to share a bit of context on our side — we try to enforce HF logits parity on every new multimodal model, and `test_modeling_multimodal.py `is the surface where that lives (the `TestModelingMultimodal` ABC does the HF forward-pass comparison with `assert_close`). MMMU is great for catching coarse regressions, but it can miss silent numerical drift in a new vision stack, which is why we lean on the unit-level parity check too.

Re-reading your three points with that skeleton in mind:
 - It's actually unit-level HF parity rather than e2e — so plugging in here would give us logits coverage we don't have yet for K2.5.
 - The MMMU lock-in lives in the integration accuracy harness, which is a separate surface from this skeleton, so the two aren't really in tension.
 - On the K2.5-specific paths (file-path video decoding, MoonViT3d 3D pos-emb) — totally fair concern, and I'd love to understand it better. Could you point to the specific abstract methods or scenarios in TestModelingMultimodal that feel awkward? Qwen2.5-VL and Qwen3-VL plug in cleanly today, but if there's a real gap, extending the skeleton is very welcome — that way the next model with similar needs benefits too. If it's just a small impedance mismatch, a thin adapter might do.

And of course, feel free to keep your dedicated file on top for anything the skeleton doesn't cover (config nesting, registration, etc.) — they can coexist nicely. Mainly hoping we can get the HF logits parity piece landed somewhere in the skeleton's flow. Appreciate you working through this with me!

tianyuxbear: Looking at the existing patterns:

1. All skeleton tests aggressively reduce config for memory (qwen2_5vl: 28→2 layers, qwen3vl_moe: 48→2 layers, llava_next: 4096→256 hidden_size, mllama: dynamic reduce_llama_config). Skeleton-style parity is only practical when the model fits a single GPU after reduction — HF doesn't support TP/EP. Mistral-Large-3-675B (#10188, merged Dec 2025) was onboarded with integration-only coverage and no entry in test_modeling_multimodal.py for this reason; K2.5 (~1T MoE) is in the same regime.

2. K2.5 has no transformers built-in class — HF side needs either `trust_remote_code=True` or a vendored copy of K2.5's modeling.py (pixtral/qwen2.5vl don't have this issue, transformers ships them natively).

3. All unit tests run in bfloat16 with random or HF-init weights, not the actual quantized checkpoint. K2.5's upstream weights are NVFP4-only — so any parity test would use random bf16 weights, meaning it verifies architecture equivalence, not real-weight accuracy. That's an inherent gap vs. e.g. Qwen2.5-VL where bf16 weights are publicly available.

yechank-nvidia: I see. Thanks for the explanation. Then, we can go with your new test files.

yechank-nvidia (approved):
Thanks for the work! LGTM

Superjomn (approved):
LGTM

tburt-nv (approved):
No dependency changes, approved

PR #12795, opened 2026-04-07
[https://nvbugs/5945047][fix] Fix Eagle3 one-model hang on SM120 via extend_ctx

ziyixiong-nv:
On SM120/121, XQA spec-dec cubins are not available for non-MLA models, causing Eagle3 one-model to hang. The previous fix (extend_ctx=True for one-model on SM120) correctly routes multi-token generation queries through FMHA instead of XQA, but broke the Eagle3 spec worker because extend_ctx inflates attn_metadata.num_contexts and num_ctx_tokens to include extended generation requests.

Fix the Eagle3 one-model spec worker to use spec_metadata.num_generations (the real generation count from the scheduler) instead of deriving it from attn_metadata.num_contexts. This decouples the spec worker's ctx/gen split from the attention layer's view, which may differ under extend_ctx mode.

Changes:
- model_engine: Set spec_metadata.num_generations in the incremental update path (_apply_incremental_update_target)
- eagle3: Use spec_metadata.num_generations for the real gen/ctx split in forward(), sample_and_accept_draft_tokens(), and prepare_1st_drafter_inputs()
- eagle3: Compute real num_ctx_tokens from _seq_lens[:num_contexts] instead of using the inflated attn_metadata.num_ctx_tokens

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected generation count metadata tracking in speculative decoding operations for improved accuracy.
  * Enhanced speculative decoding batch partitioning logic to use authoritative metadata sources.
  * Added optimized support for NVIDIA SM120/121 GPUs in one-engine speculative decoding mode.

* **Tests**
  * Restored test coverage by removing outdated skip directive.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/speculative/interface.py
```diff
@@ -262,7 +262,13 @@ def extend_ctx(self, attention_backend: Type[AttentionBackend]):
         """
 
         if self.use_one_engine():
-            # 1-model has separate logic for handling draft tokens
+            # SM120/121 lacks XQA spec-dec cubins for non-MLA models, so
+            # one-model must fall back to treating draft tokens as context.
+            # The C++ attention layer will still use FMHA fallback for the
+            # draft model's multi-token queries (spec-dec mode remains on).
+            if get_sm_version() in (120, 121) and issubclass(
```
cascade812, line 271: Should we return true for non-MLA only to align with above comment?

ziyixiong-nv: Good question. The broad condition is intentional — we haven't confirmed that MLA models' XQA path has working spec-dec cubins on SM120/121 either. Since `extend_ctx=True` is a safe fallback (correct but potentially slower), I'd prefer to keep the conservative approach for this hang-fix and narrow it in a follow-up once MLA + one-engine spec-dec is validated on SM120.

Updated the comment to better reflect the actual scope (31d635a).

cascade812: I see. In this case, should we apply same logic to MTP which also uses attn_metadata.num_contexts?

cascade812: @ziyixiong-nv kindly ping on above in case you didn't see it.

ziyixiong-nv: Sorry for didn't noticing the message.
MTP is already covered — one-model MTP is included in `self.use_one_engine()` in line 264. two-model spec dec is being deprecated.

```
  def use_one_engine(self):                                                                                                                                                                                                                      
      return self.is_eagle3_one_model() or self.is_mtp_one_model(
      ) or self.is_external_drafter() or self.is_sa()
```

cascade812: @ziyixiong-nv Apologies for the confusion, what I was asking is whether we need to apply the same logic (i.e., the changes you made to eagle3.py) to the MTP single-model path as well. My understanding is that after your change, MTP would also use the FMHA fallback, which means using attn_metadata.num_contexts as the number of context requests would be incorrect there too.

ziyixiong-nv: You are right. The bug is reported for test_eagle3, but it's indeed a problem for other spec dec algorithms like MTP. I'll apply the same fix to other one-model spec dec algorithms and add some tests to verify.

ziyixiong-nv:
The test hangs in ToT main and I found some deeper issue. Will close this PR.

PR #12820, opened 2026-04-07
[None][fix] Add bounded timeout to gen-side KV transfer in C++ CacheTransceiver

yifjiang:
## Summary
- Replace unbounded `future.get()` in `checkGenTransferStatus` with `future.wait_for()` that branches on ready/timeout/error status
- Fix non-blocking path that could still block indefinitely when `kv_transfer_sender_future_timeout_ms` is unset (`std::nullopt`) — uses `blockAll` flag instead of `!receiverFutureTimeoutMs.has_value()` to decide blocking behavior
- Guard the `updateKVCacheTransferBW` timing collective so it only runs when all ranks block together (`blockAll`) or the request was confirmed ready on every rank in the initial poll, preventing `allgather` hangs when a peer timed out and skipped the request

Supersedes #12476 — addresses CodeRabbit review feedback from that PR.

## Test plan
- [ ] Existing disaggregated serving unit tests pass
- [ ] Manual test with `kv_transfer_sender_future_timeout_ms` set and unset
- [ ] Multi-rank test with `TRTLLM_KVCACHE_TIME_OUTPUT_PATH` enabled to verify no allgather hang on timeout

### DIFF
--- a/cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
+++ b/cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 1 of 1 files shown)


yifjiang:
Closing as stale — superseded by #13083 (gen-side bounded timeout) and the gen-side enforcement in #13056.

PR #12835, opened 2026-04-08
[None][feat] optimize sm120 mxfp8 gemm kernel

CarstyYou:
## Description
                                                                                                                                                                                                                                                    
  Optimize FP8 block-scale GEMM kernels for SM120 (Blackwell RTX PRO 6000) by integrating the 6KD kernel library. Two logical commits:                                                                                                              
  
  1. **GEMM optimization** (`bda9f4e8`): replace TRT-LLM's sm120 FP8 block-scale GEMM kernel with the 6KD-optimized version.                                                                                                                        
     - Add `TileK` template parameter to `SM120BlockScaledBuilder` (supports K=64/128)
     - Add 128x128k64 and 64x128k64 tile configs for higher arithmetic intensity                                                                                                                                                                    
     - Implement subtile epilogue with multi-stage R2S/S2G pipelining                                                                                                                                                                               
     - Replace union `TensorStorage` with struct (independent load/store SMEM)
     - Utilization-driven tile dispatch in `gemm_dispatch_sm120` and `strided_batch_gemm_dispatch_sm120`                                                                                                                                            
                                                                                                                                                                                                                                                    
  2. **SwapAB kernel** (`a0f4da5f`): new path for small-M (M≤8) GEMV-like workloads where the standard 32×128 tile wastes 75-94% of rows to padding. Swaps A↔B so TileM=64 covers N (large) and TileN=8 covers M (small), with STG scatter store    
  doing the transpose on writeback. `k_slices=1` (no sliced-K; workspace-based extension reserved for future PR).                                                                                                                                   
                                                                                                                                                                                                                                                    
  ## Test Coverage                                             

  Covered by existing parametrized tests in `tests/unittest/_torch/thop/parallel/test_fp8_block_scale_gemm.py`:                                                                                                                                     
  
  - `test_fp8_block_scale_gemm` — dense GEMM                                                                                                                                                                                                        
  - `test_fp8_block_scale_bmm` — batched GEMM                  
  - `test_fp8_block_scale_moe_gemm` — MoE GEMM                                                                                                                                                                                                      
                                                                                                                                                                                                                                                    
  Locally verified with extended M ∈ {2, 4, 8, 16, 32, 64, 128, 256} covering both SwapAB (M≤8) and standard (M>8) paths on RTX PRO 6000:                                                                                                           
                                                                                                                                                                                                                                                    
  | Test | Total | SwapAB cases | Standard cases |                                                                                                                                                                                                  
  |------|-------|-------------|----------------|              
  | `test_fp8_block_scale_gemm` | 56/56 PASS | 21 | 35 |                                                                                                                                                                                            
  | `test_fp8_block_scale_bmm` | 96/96 PASS | 36 | 60 |                                                                                                                                                                                             
                                                                                                                                                                                                                                                    
  End-to-end: Qwen3-32B-FP8 inference validated on RTX PRO 6000.                                                                                                                                                                                    
                                                                                                                                                                                                                                                    
  ## PR Checklist                                              

  - [x] Code adheres to TRT-LLM coding guidelines (clang-format via pre-commit)                                                                                                                                                                     
  - [x] New code paths covered by existing parametrized tests
  - [x] No new third-party dependencies                                                                                                                                                                                                             
  - [x] No CODEOWNERS changes                                                                                                                                                                                                                       
  - [x] No documentation changes
  - [x] No architecture diagram changes (kernel-internal only)                                                                                                                                                                                      
                                                                                                                                                                                                                                                    
  ---


cpp/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/sm120_blockwise_gemm/sm120_fp8_gemm_1d1d.cuh
```diff
@@ -581,10 +571,14 @@ struct SM120BlockScaledKernel
         {
             uint32_t sf_phase = 0;
             uint32_t ab_phase = 0;
+            uint32_t se_phase[KT::StagesD];
+            for (int i = 0; i < KT::StagesD; ++i)
+                se_phase[i] = 1;
+            int epi_stage = 0;
```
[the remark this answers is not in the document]
CarstyYou: Thanks for the analysis. This is **not a deadlock** — the reasoning above assumes `wait(phase)` blocks *until* the barrier reaches `phase`, but CUTLASS's `ClusterTransactionBarrier::wait(parity)` uses **parity semantics**: it blocks *while*  
  `current_parity == parity`, and exits when the parity flips.                                                                                                                                                                                      
                                                                                                                                                                                                                                                    
  After `store_empty_mbar[i].init(1)`, the barrier's current parity is `0`. Initializing `se_phase[i] = 1` means "`wait(parity=1)` — pass immediately while current is still 0", which is the correct initial condition: on iteration 0, stage `i`'s
   SMEM slot hasn't been written yet, so the math thread should not wait for a store to consume it.                                                                                                                                                 
                                                                                                                                                                                                                                                    
  After the first math-produces / store-consumes cycle, the arrival flips parity 0 → 1, and `se_phase[i] ^= 1` toggles to 0 for the next wait on that stage — now correctly waiting for the consumer.                                               
  
  The symmetric producer-side `sf_phase[] = {}` (initialized to 0) works because those barriers follow the same parity convention from the opposite side of the pipeline.                                                                           
                                                            
  **Empirical evidence:**                                                                                                                                                                                                                           
  - 82/82 kernel unit tests pass on RTX PRO 6000 (`tests/unittest/_torch/thop/parallel/test_fp8_block_scale_gemm.py` — GEMM / BMM / MoE)
  - Qwen3-32B-FP8 end-to-end inference passes on RTX PRO 6000                                                                                                                                                                                       
                                                                                                                                                                                                                                                    
  Happy to keep discussing if I'm reading the convention wrong — a pointer to the opposite example in CUTLASS would help.

cpp/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/sm120_blockwise_gemm/sm120_fp8_swapab_1d1d.cuh
```diff
@@ -0,0 +1,483 @@
...
+                    }
+                });
+        }
+        else
+        {
+            // No slicing: direct STG scatter BF16
+            int l_coord = blockIdx.z;
+            auto mD = make_tensor(make_gmem_ptr(params.ptr_D),
+                make_layout(
+                    make_shape(M_k, N_k, L_k), make_stride(Int<1>{}, params.ld_D, (int64_t) N_k * params.ld_D)));
```
Barry-Delaney, line 311: The direct (non-sliced) epilogue rebuilds the BMM batch stride here as `N_k * params.ld_D`, but the user's `stride_d` reaches the kernel via `make_stride(ld_d, Int<1>{}, stride_d)` in `fp8_blockscale_gemm_kernel.cuh::launch_sm120_swapab_kernel` and is then **dropped** in `to_underlying_arguments` (line 100, `int64_t ld_D = get<0>(args.dD);` — `get<2>` is never stored on `Params`, which only carries `ld_D` at line 57).

The formula `N_k * params.ld_D` happens to equal `stride_d` whenever the output is tightly packed, which is why the new `test_fp8_block_scale_bmm_strided_out` cannot catch this — it pads only in `n`, so `stride(0) == M_real * ld_d`. For an `_out` BMM with extra L-direction padding, or any `stride_d != M_real * ld_d`, the kernel writes to wrong addresses. The sliced-K convert path is fine because it honors `stride_dst = stride_d`; this bug is direct-only.

On RTX PRO 6000 (~188 SMs) with `n = 2112` and `num_groups >= 6`, the dispatch hits `k_slices = 1` and therefore the buggy path.

Fix:

```cpp
// Params:
int64_t ld_D;
int64_t stride_D;   // add

// to_underlying_arguments:
int64_t ld_D     = get<0>(args.dD);
int64_t stride_D = get<2>(args.dD);
return {..., args.ptr_D, ld_D, stride_D, ps, nullptr, 1};

// direct branch (this line):
make_stride(Int<1>{}, params.ld_D, params.stride_D)
```

cpp/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/sm120_blockwise_gemm/sm120_fp8_swapab_1d1d.cuh
```diff
@@ -0,0 +1,483 @@
...
+        int lane_predicate = cute::elect_one_sync();
+        bool is_tma_warp = (warp_idx >= KT::kNumMathWarps);
+
+        auto [M, N, K, L] = params.problem_shape;
+        // Sliced-K: blockIdx.z encodes (l_coord, k_slice_idx) laid out as l_coord * k_slices + slice.
+        // When k_slices==1, blockIdx.z == l_coord (single-slice fast path preserved).
+        int k_slice_idx = (params.k_slices > 1) ? (int(blockIdx.z) % params.k_slices) : 0;
+        int l_coord = (params.k_slices > 1) ? (int(blockIdx.z) / params.k_slices) : int(blockIdx.z);
+        int K_per_slice = K / params.k_slices;
+        int32_t sf_tile_count = K_per_slice / 512;
```
Barry-Delaney, line 345: `sf_tile_count = K_per_slice / 512` and the host-side `sf_groups = shape_k / 512` in `fp8_blockscale_gemm_kernel.cuh:784` both round down, so any K that isn't a multiple of 512 silently drops the K-tail.

Worse, when `K < 512`, `sf_tile_count == 0`: the mainloop `for (int32_t sf_tile_idx = 0; sf_tile_idx < sf_tile_count - 1; ++sf_tile_idx)` is `< -1` which is false, so the loop body is skipped. The tail iteration then unconditionally executes `sf_full_mbar[0].wait(sf_phase)`. But the SF load warp's `for (int32_t i = 0; i < sf_tile_count; ++i)` also iterates 0 times and never arrives at `sf_full_mbar[0]`. Result: **deadlock.**

The existing non-SwapAB SM120 path uses `(K + 511) / 512`, so the new `M <= 8` dispatch is a support regression unless `K % 512 == 0` is an enforced contract.

Please pick one: (a) add `TLLM_CHECK_WITH_INFO(shape_k > 0 && shape_k % 512 == 0, ...)` in `launch_sm120_swapab_kernel` and document the contract, (b) make the slicing/mainloop cover `ceil_div(K, 512)` and the K-tail correctly, or (c) route shapes that don't meet the K contract to the non-SwapAB kernel.

tests/unittest/_torch/thop/parallel/test_fp8_block_scale_gemm.py
```diff
@@ -217,6 +217,68 @@ def test_fp8_block_scale_bmm(dtype, m, k, n, num_groups):
...
+)
+@pytest.mark.parametrize(
+    "m",
+    [7, 8],
+)
+@pytest.mark.parametrize(
+    "num_groups",
+    [2, 4],
+)
+def test_fp8_block_scale_bmm_strided_out(m, k, n, num_groups):
```
Barry-Delaney, line 238: Thanks for adding this regression test — but the four parameter combinations all hit the sliced-K (convert-kernel) path, not the direct path. With `n = 2112` and `n_blocks = 33` on ~188 SMs:

| `(k, n)` | `num_groups` | `n_blocks * groups` | `sf_groups` | `k_slices` |
|----------|--------------|---------------------|-------------|------------|
| (5120, 2112) | 2 | 66  | 10 | 5 |
| (5120, 2112) | 4 | 132 | 10 | 2 |
| (7168, 2112) | 2 | 66  | 14 | 7 |
| (7168, 2112) | 4 | 132 | 14 | 2 |

All combos yield `k_slices > 1`. Additionally the strided pattern `output_padded.narrow(2, 0, n)` has `stride(0) = m * padded_n == M_real * ld_d` — exactly the formula the buggy direct path uses, so even if the test routed through direct it wouldn't detect it.

To actually cover the direct path and catch M1: force `n_blocks * num_groups >= sm_count` (e.g. `num_groups = 8, n = 2112` → `n_blocks * groups = 264 >= 188`), and make `stride_d != M_real * ld_d` (e.g. construct the output via `torch.empty((num_groups + 1, m, n)).narrow(0, 0, num_groups)` so `stride(0) > m * n`). Also worth adding `m = 1` for the extreme small-M edge.

cpp/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/fp8_blockscale_gemm.cu
```diff
@@ -263,7 +266,29 @@ size_t CutlassFp8BlockScaleGemmRunner<ElementA, ElementB, ElementD>::getWorkspac
...
+    // and only reserve bytes when sliced-K will actually activate, so most DSV3
+    // shapes (N_blocks * L >= sm_count) incur zero allocation.
+    //
+    // For BMM (num_problems > 1), each batch's N-blocks count independently toward
+    // SM occupancy, so the effective n_blocks for the heuristic is n_blocks * L.
+    // Kernel's atomicAdd indexing includes an L offset (see sm120_fp8_swapab_1d1d.cuh).
+    if (tensorrt_llm::common::getSMVersion() == 120 && shape_m <= 8)
+    {
+        int sm_count = tensorrt_llm::common::getMultiProcessorCount();
+        int n_blocks = static_cast<int>((shape_n + 63) / 64); // TileM = 64
```
Barry-Delaney, line 281: `compute_k_slices_sm120_swapab` is shared between the host predictor here and the device launcher (`fp8_blockscale_gemm_kernel.cuh:866`). Suggest factoring `TileM` into a named constant near the helper (e.g. `kSwapAB_TileM = 64`) and referencing it from both call sites and the dispatch template arguments.

Barry-Delaney (commented):
Nice cleanup and a substantive optimization. Minor comments left.

PR #12843, opened 2026-04-08
[#12840][fix] propagate allreduce_strategy to Qwen3 embed_tokens

edenfunf:
## What

`Qwen3Model.embed_tokens` was constructed without passing `allreduce_strategy`, causing the `AllReduce` object inside the embedding layer to always use `AllReduceStrategy.AUTO` regardless of the user's configuration.

**Before:** `allreduce_strategy: NCCL` (or any explicit strategy) in `ModelConfig` was silently ignored for the Qwen3 embedding all-reduce path.

**After:** `embed_tokens` respects the configured `allreduce_strategy`, consistent with how all other TP linear layers in the model are constructed.

## Why

`LMHead` and `Embedding` inherit from `Linear`, which accepts `allreduce_strategy` and uses it to create the `AllReduce` object. However, neither `LMHead` nor `Embedding` exposed this parameter, so they always passed the default `AUTO` to `Linear`.

When a user sets `allreduce_strategy: NCCL` for TP=2 Qwen3 inference, initialization can fail with:
```
RuntimeError: CUDA Runtime API error: <cudaError_t.cudaErrorInvalidDevice: 101>
```
because `AUTO` attempts a different communication primitive than the user-configured one.

Confirmed by reading the code: `Embedding.forward()` calls `self.all_reduce(output)` which uses the `AllReduce` object created in `Linear.__init__` — always with `AUTO` before this fix.

## How

1. Added `allreduce_strategy: AllReduceStrategy = AllReduceStrategy.AUTO` to `LMHead.__init__` and passed it through to `Linear.__init__`.
2. Added the same parameter to `Embedding.__init__` and passed it to `LMHead.__init__`.
3. In `Qwen3Model.__init__`, passed `allreduce_strategy=config.allreduce_strategy` when constructing `embed_tokens`.

The default is `AUTO`, so all existing code that does not pass `allreduce_strategy` is unaffected.

The fix follows the same pattern used throughout the model (e.g., `modeling_cohere2.py` passes `allreduce_strategy=model_config.allreduce_strategy` to every `Linear` layer).

## Test

- Verified the bug root cause by reading `Embedding.forward()` → `self.all_reduce` → `Linear.__init__` → `AllReduce(strategy=allreduce_strategy)`.
- `pre-commit run` passes on all relevant hooks for both modified files.
- The change is backward-compatible: the `allreduce_strategy` parameter defaults to `AUTO` in all three layers, so no callers that omit it are affected.
- Multi-GPU regression test requires TP=2 hardware; existing `tests/unittest/_torch/multi_gpu/test_embedding.py` covers the `Embedding` forward path.

Fixes #12840

## Summary

* Propagated `allreduce_strategy` from `Qwen3Model` configuration to `embed_tokens`.
* Added optional `allreduce_strategy` parameters to `Embedding` and `LMHead`.
* Forwarded the parameter to `Linear`.
* Preserved `AllReduceStrategy.AUTO` as the default.

## Dev Engineer Review

* The change correctly applies configured all-reduce strategies, including `NCCL` and `NCCL_SYMMETRIC`, to the Qwen3 embedding path.
* The default preserves existing caller behavior.
* No test, configuration, or test-list changes were included.

## QA Engineer Review

No test changes.


byshiue (approved):

edenfunf:
@byshiue Hi! When you have a chance, could you help run the CI tests? Thanks!

moraxu (approved):

asfiyab-nvidia (approved):

edenfunf:
@moraxu Could you re-run CI, or share which tests failed? I can't see the CI report from outside NVIDIA. Thanks!

PR #12844, opened 2026-04-08
Fix: Beam Search Requests Never Terminate When `max_tokens` is Reached

Doloxetine:
### Problem

When using beam search (e.g., `beam_width=128`), requests fail to terminate even after reaching `max_tokens`. For example, with `max_tokens=8`, a request continues generating indefinitely — token counts grow from hundreds to thousands or even tens of thousands, exhausting KV cache blocks, blocking the entire batch, and causing QPS to drop to near zero.

### Root Cause

TRT-LLM has two termination mechanisms, and **both fail under beam search**:

#### Termination Path 1: `finishedSum == reqBeamWidth` (GPU-side)

The GPU decoder marks each beam as `finished` when it produces an EOS token or hits a length limit. `finishedSum` counts how many beams are finished. The request terminates only when `finishedSum == reqBeamWidth` (i.e., all beams are finished).

**Why it fails:** Beam search performs **beam reordering** after every step — it selects the top-k beams by cumulative log probability and physically rearranges beam data (KV cache, token sequences, and **finished states**). A beam that was marked `finished` in step N can be **overwritten** by an unfinished beam in step N+1:

| Step | finishedSum | What happened |
|------|-------------|---------------|
| 5 | 1 | beam[3] generated EOS |
| 6 | 0 | beam reordering replaced beam[3] with an unfinished beam |
| 7 | 2 | beam[1], beam[5] generated EOS |
| 8 | 1 | beam reordering overwrote some again |
| ... | ... | never reaches `reqBeamWidth` |

With `beam_width=128`, it is virtually impossible for all 128 beams to be simultaneously finished, because reordering keeps replacing finished beams with unfinished ones. `finishedSum` oscillates but never converges.

#### Termination Path 2: `willCompleteNextIteration()` (CPU-side)

The CPU-side check uses `mTokens` length:

```cpp
SizeType32 getMaxNumGeneratedTokens() const {
    return getMaxBeamNumTokens() - mPromptLen;
}

bool willCompleteNextIteration() const {
    return getMaxNumGeneratedTokens() + mNumTokensPerIteration >= mMaxNewTokens;
}
```

**Why it fails:** In `decoderSync`, new tokens are added to `mTokens` based on:

```cpp
auto const numNewOutputTokens = seqLen - llmReq->getNumTokens(beam);
numNewTokens[beam] = std::min(numGeneratedTokens, numNewOutputTokens);
```

After beam reordering, the GPU-side `sequenceLengths[beam]` may correspond to a **different** (shorter) beam than what the CPU-side `mTokens[beam]` tracked in the previous step. This can produce `numNewOutputTokens ≤ 0`, meaning **no new tokens are appended** to `mTokens` for that beam. As a result, `getMaxNumGeneratedTokens()` stalls at a small value and never reaches `mMaxNewTokens`, so `willCompleteNextIteration()` never returns `true`.

#### Cascading Failures

The infinite generation eventually triggers `hardMaxTokensCutoff` (if present) or other resource limits, which leads to two additional crashes:

1. **`cannot create std::vector larger than max_size()`**: When beam search is force-terminated before convergence, `gatherTree`/`finalize` may produce invalid `sequenceLengths` for unfinished beams. `postProcessRequest` computes `generatedLength = sequenceLengths[beam] - promptLen`, which can become an astronomically large value, causing `std::vector` allocation failure.

2. **`decoderFinishedEvent must be nullopt` assertion failure**: The above exception propagates out of `decoderSync`, skipping `decoderWaitEvent.reset()`. On the next `forwardAsync` call, the event still has a value, triggering the assertion and causing **all subsequent requests** to fail.

#### 1. Hard cutoff using `decodingIter` (root fix)

In `updateRequests` / `decoderSync`, add a reliable termination condition based on `decodingIter` — a pure iteration counter that increments unconditionally every decoder step, unaffected by beam reordering:

```cpp
auto const decodingIter = llmReq->getDecodingIter();
bool const hardMaxTokensCutoff = (decodingIter >= llmReq->mMaxNewTokens);
```

When `decodingIter >= maxNewTokens`, the request is force-terminated regardless of `finishedSum`. Any beams still in `NOT_FINISHED` state are set to `FinishReason::kLENGTH` to prevent Python-layer errors.

#### 2. `decoderFinishedEvent` assertion → graceful recovery

In `forwardAsync`, replace the hard assertion:

```cpp
TLLM_CHECK(!decoderFinishedEvent.has_value());
```

with a graceful recovery: if the event still has a value (indicating a previous `decoderSync` failed to reset it due to an exception), synchronize and reset it before proceeding. This prevents a single exception from cascading into permanent failure for all subsequent requests.

#### 3. `generatedLength` safety clamp in `postProcessRequest`

Add bounds checking for `generatedLength` computed from `sequenceLengths`:

```cpp
auto generatedLength = sequenceLengthsHostData[beam] - llmReq.mPromptLen;
if (generatedLength < 0 || generatedLength > maxGeneratedLength)
{
    generatedLength = std::clamp(generatedLength, 0, maxGeneratedLength);
}
```

This prevents `std::vector` allocation failure when `gatherTree` produces invalid sequence lengths for unconverged beams.

## Release Notes

* **Bug Fixes**
  * Improved error recovery in batch processing to prevent cascading failures after exceptions.
  * Enhanced validation of generated token lengths in beam search with automatic clamping to valid ranges.
  * Added safeguards to enforce maximum token generation limits and prevent runaway generation scenarios.

* **Style**
  * Minor code formatting adjustments.


Doloxetine (commented):
update

PR #12873, opened 2026-04-09
[None][feat] EXAONE-4.5 Support

yechank-nvidia:
This PR is adding Day-0 support of LG AI's new VLM model, [EXAONE-4.5](https://huggingface.co/LGAI-EXAONE/EXAONE-4.5-33B). 

This PR includes Text + Multimodal support. 

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/yechank/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/yechank/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">

ISL1000 OSL1000 IMG(512,512) BF16 | H200 | CONC | C8 | C16 | C32 | C64 | C128
-- | -- | -- | -- | -- | -- | -- | --
  |   | Output Token Throughput Per User (tokens/sec/user) | 130.85 | 116.83 | 96.95 | 82.09 | 62.76
  |   | Output Token Throughput (tokens/sec) | 988.6 | 1739.74 | 2838.79 | 4560.93 | 6427.07

ISL1000 OSL1000 IMG(512,512) BF16 | B200 | CONC | C8 | C16 | C32 | C64 | C128
-- | -- | -- | -- | -- | -- | -- | --
  |   | Output Token Throughput Per User (tokens/sec/user) | 194.82 | 176.8 | 148.64 | 123.08 | 89.24
  |   | Output Token Throughput (tokens/sec) | 1451.54 | 2580.11 | 4211.01 | 6443.06 | 8644.52

</body>

</html>

**Prerequisite**
```
pip install git+https://github.com/nuxlear/transformers.git@add-exaone4_5
```

**Sample command**
```
# Server run
trtllm-serve LGAI-EXAONE/EXAONE-4.5-33B --tp_size 2 --port 8000 --reasoning_parser qwen3

# Quickstart
python3 examples/llm-api/quickstart_multimodal.py --model_dir LGAI-EXAONE/EXAONE-4.5-33B
```

@codderabbit summary

### DIFF
--- a/examples/models/core/exaone/README.md
+++ b/examples/models/core/exaone/README.md
@@ -14,9 +14,11 @@ This document shows how to build and run [EXAONE](https://huggingface.co/LGAI-EX
     - [EXAONE-3.0](#exaone-30)
     - [EXAONE-Deep](#exaone-deep)
     - [EXAONE-4.0](#exaone-40)
+    - [EXAONE-4.5](#exaone-45)
     - [K-EXAONE](#k-exaone)
   - [PyTorch flow](#pytorch-flow)
     - [Running EXAONE-4.0](#running-exaone-40)
+    - [Running EXAONE-4.5](#running-exaone-45)
     - [Running K-EXAONE](#running-k-exaone)
       - [MoE Backend Options](#moe-backend-options)
     - [PyTorch flow Quantization](#pytorch-flow-quantization)
@@ -45,6 +47,7 @@ This document shows how to build and run [EXAONE](https://huggingface.co/LGAI-EX
   * FP16
   * BF16
   * Tensor Parallel (TP)
+  * Multimodal (EXAONE-4.5 only)
   * Expert Parallel (EP) (K-EXAONE only)
   * Attention Data Parallel (ADP) (K-EXAONE only)
   * Disaggregated Serving
@@ -59,7 +62,7 @@ This document shows how to build and run [EXAONE](https://huggingface.co/LGAI-EX
 
 **Note:**
 - **EXAONE-3.0** & **EXAONE-Deep** are supported using the [TRT Flow](#trt-flow).
-- **EXAONE-4.0** & **K-EXAONE** are supported using the [PyTorch flow](#pytorch-flow).
+- **EXAONE-4.0**, **EXAONE-4.5**, & **K-EXAONE** are supported using the [PyTorch flow](#pytorch-flow).
 
 Please refer to the corresponding sections below for usage instructions and examples for each model.
 
@@ -90,6 +93,17 @@ export HF_MODEL_DIR=hf_models/exaone4
 git clone https://huggingface.co/LGAI-EXAONE/EXAONE-4.0-32B $HF_MODEL_DIR
 ```
 
+### EXAONE-4.5
+
+EXAONE-4.5 is a multimodal model. It is supported only via the [PyTorch flow](#pytorch-flow).
+
+Download the HuggingFace checkpoint for your EXAONE-4.5 variant from the [LGAI-EXAONE](https://huggingface.co/LGAI-EXAONE) organization. The example below uses `EXAONE-4.5-33B`; replace it with the variant you want to run.
+
+```bash
+export HF_MODEL_DIR=hf_models/exaone4_5
+git clone https://huggingface.co/LGAI-EXAONE/EXAONE-4.5-33B $HF_MODEL_DIR
+```
+
 ### K-EXAONE
 
 K-EXAONE is a Mixture of Experts (MoE) model based on the EXAONE architecture. It features a hybrid architecture with both dense and MoE layers, sliding window attention, and supports FP8 and NVFP4 quantization for efficient inference.
@@ -115,6 +129,21 @@ The output will be like:
 [0] Prompt: 'Hello, my name is', Generated text: " [Your Name], and I'm a [Your Profession]. I'm here to learn and share with you.\n\nBest regards,\n[Your Name]\n\nThis letter is concise, professional, and clearly states who you are and what you're here for. It's a good starting point"
 [1] Prompt: 'The capital of France is', Generated text: ' Paris.\n\nThe capital of France is Paris.\n\nThe capital of France is Paris.\n\nThe capital of France is Paris.\n\nThe capital of France is Paris.\n\nThe capital of France is Paris.\n\nThe capital of France is Paris.\n\nThe capital of France is Paris'
 [2] Prompt: 'The future of AI is', Generated text: ' not just about technology but also about how we choose to use it. We must ensure that AI is developed and deployed in a way that benefits all of humanity, not just a select few. This means prioritizing ethical considerations, transparency, and accountability in AI development. It also means involving diverse stakeholders in the conversation about AI'
+```
+
+### Running EXAONE-4.5
+
+To quickly run EXAONE-4.5 models, you can use [examples/llm-api/quickstart_multimodal.py](../../../llm-api/quickstart_multimodal.py):
+
+```bash
+python ../../../llm-api/quickstart_multimodal.py --model_dir $HF_MODEL_DIR
+```
+
+The output will be like:
+```bash
+[0] Prompt: 'Describe the natural environment in the image.', Generated text: 'Okay, the user asked me to describe the natural environment in the image. But wait, there's no image provided here. Hmm, that's a problem. How can I describe something I can't see?\n\nFirst, I need to check if there's any image attached. The user mentioned "the image," but in the current context, there's no image data. Maybe they forgot'
+[1] Prompt: 'Describe the object and the weather condition in the image.', Generated text: 'Okay, the user asked me to describe the object and the weather condition in the image. But wait, there's no image provided here. Hmm, that's a problem. How can I describe something I can't see?\n\nFirst, I need to check if there's any image attached. The user mentioned "the image," but in the current context, there's no image data. Maybe'
+[2] Prompt: 'Describe the traffic condition on the road in the image.', Generated text: 'Okay, the user is asking me to describe the traffic condition on the road in the image. But wait, there's a problem here—I don't actually see any image. The user mentioned "the image," but in this text-based interface, there's no visual content provided. \n\nHmm, I need to handle this carefully. The user might have forgotten to attach the image or assumed I could see it'
 ```
 
 ### Running K-EXAONE
--- a/examples/models/core/exaone/README.md
+++ b/examples/models/core/exaone/README.md
@@ -101,7 +101,7 @@ Download the HuggingFace checkpoint for your EXAONE-4.5 variant from the [LGAI-E
 
 ```bash
 export HF_MODEL_DIR=hf_models/exaone4_5
-git clone https://huggingface.co/LGAI-EXAONE/EXAONE-4.5-33B $HF_MODEL_DIR
+huggingface-cli download LGAI-EXAONE/EXAONE-4.5-33B --local-dir $HF_MODEL_DIR
 ```
 
 ### K-EXAONE
@@ -112,7 +112,7 @@ Download the HuggingFace checkpoints of the K-EXAONE model:
 
 ```bash
 export HF_MODEL_DIR=hf_models/kexaone
-git clone https://huggingface.co/LGAI-EXAONE/K-EXAONE-236B-A23B $HF_MODEL_DIR
+huggingface-cli download LGAI-EXAONE/K-EXAONE-236B-A23B --local-dir $HF_MODEL_DIR
 ```
 
 ## PyTorch flow
--- a/tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
+++ b/tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
@@ -1,3 +1,18 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 from tensorrt_llm._torch.models.checkpoints.base_weight_loader import ConsumableWeightsDict
 from tensorrt_llm._torch.models.checkpoints.hf.weight_mapper import HfWeightMapper
 from tensorrt_llm._torch.models.modeling_utils import register_mapper
--- a/tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
+++ b/tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
@@ -1,12 +1,20 @@
+from typing import Union
+
+from torch import nn
+
+from tensorrt_llm._torch.model_config import ModelConfig
 from tensorrt_llm._torch.models.checkpoints.base_weight_loader import ConsumableWeightsDict
 from tensorrt_llm._torch.models.checkpoints.hf.weight_mapper import HfWeightMapper
-from tensorrt_llm._torch.models.modeling_utils import register_mapper
+from tensorrt_llm._torch.models.modeling_utils import DecoderModelForCausalLM, register_mapper
 
 
 @register_mapper("HF", "Exaone4_5_ForConditionalGeneration")
 class Exaone4_5HfWeightMapper(HfWeightMapper):
-    def __init__(self):
-        super().__init__()
+    def init_model_and_config(
+        self, model: Union[nn.Module, DecoderModelForCausalLM], config: ModelConfig
+    ):
+        super().init_model_and_config(model, config)
+        self.model.config.tie_word_embeddings = False
 
     def preprocess_weights(self, weights: dict):
         """Rename HF checkpoint prefixes; supports plain dict and ConsumableWeightsDict."""
--- a/tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
+++ b/tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
@@ -1,21 +1,10 @@
-from typing import Union
-
-from torch import nn
-
-from tensorrt_llm._torch.model_config import ModelConfig
 from tensorrt_llm._torch.models.checkpoints.base_weight_loader import ConsumableWeightsDict
 from tensorrt_llm._torch.models.checkpoints.hf.weight_mapper import HfWeightMapper
-from tensorrt_llm._torch.models.modeling_utils import DecoderModelForCausalLM, register_mapper
+from tensorrt_llm._torch.models.modeling_utils import register_mapper
 
 
 @register_mapper("HF", "Exaone4_5_ForConditionalGeneration")
 class Exaone4_5HfWeightMapper(HfWeightMapper):
-    def init_model_and_config(
-        self, model: Union[nn.Module, DecoderModelForCausalLM], config: ModelConfig
-    ):
-        super().init_model_and_config(model, config)
-        self.model.config.tie_word_embeddings = False
-
     def preprocess_weights(self, weights: dict):
         """Rename HF checkpoint prefixes; supports plain dict and ConsumableWeightsDict."""
         is_consumable = isinstance(weights, ConsumableWeightsDict)
--- a/tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
+++ b/tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
@@ -1,0 +1,26 @@
+from tensorrt_llm._torch.models.checkpoints.base_weight_loader import ConsumableWeightsDict
+from tensorrt_llm._torch.models.checkpoints.hf.weight_mapper import HfWeightMapper
+from tensorrt_llm._torch.models.modeling_utils import register_mapper
+
+
+@register_mapper("HF", "Exaone4_5_ForConditionalGeneration")
+class Exaone4_5HfWeightMapper(HfWeightMapper):
+    def __init__(self):
+        super().__init__()
+
+    def preprocess_weights(self, weights: dict):
+        """Rename HF checkpoint prefixes; supports plain dict and ConsumableWeightsDict."""
+        is_consumable = isinstance(weights, ConsumableWeightsDict)
+        renamed = {}
+        for key, value in weights.items():
+            if key.startswith("model.visual."):
+                new_key = key.replace("model.visual.", "visual.")
+                renamed[new_key] = value
+            elif key.startswith("model.language_model."):
+                new_key = key.replace("model.language_model.", "model.")
+                renamed[new_key] = value
+            else:
+                renamed[key] = value
+        if is_consumable:
+            return ConsumableWeightsDict(renamed)
+        return renamed
--- a/tensorrt_llm/_torch/models/modeling_exaone4_5.py
+++ b/tensorrt_llm/_torch/models/modeling_exaone4_5.py
@@ -15,10 +15,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import copy
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 
 import torch
-from transformers import AutoTokenizer, PretrainedConfig, PreTrainedModel
+from transformers import AutoConfig, AutoTokenizer, PretrainedConfig, PreTrainedModel
 
 from tensorrt_llm._torch.models.checkpoints.base_weight_mapper import BaseWeightMapper
 from tensorrt_llm._torch.models.modeling_multimodal_utils import _is_disagg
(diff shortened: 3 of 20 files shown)


examples/models/core/exaone/README.md
```diff
@@ -90,6 +93,17 @@ export HF_MODEL_DIR=hf_models/exaone4
...
 
+### EXAONE-4.5
+
+EXAONE-4.5 is a multimodal model. It is supported only via the [PyTorch flow](#pytorch-flow).
+
+Download the HuggingFace checkpoint for your EXAONE-4.5 variant from the [LGAI-EXAONE](https://huggingface.co/LGAI-EXAONE) organization. The example below uses `EXAONE-4.5-33B`; replace it with the variant you want to run.
+
+```bash
+export HF_MODEL_DIR=hf_models/exaone4_5
+git clone https://huggingface.co/LGAI-EXAONE/EXAONE-4.5-33B $HF_MODEL_DIR
```
2ez4bz: Does this download large files too? Wouldn't it be better to recommend using the `hf` CLI?

tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
```diff
@@ -0,0 +1,34 @@
...
+from tensorrt_llm._torch.model_config import ModelConfig
+from tensorrt_llm._torch.models.checkpoints.base_weight_loader import ConsumableWeightsDict
+from tensorrt_llm._torch.models.checkpoints.hf.weight_mapper import HfWeightMapper
+from tensorrt_llm._torch.models.modeling_utils import DecoderModelForCausalLM, register_mapper
+
+
+@register_mapper("HF", "Exaone4_5_ForConditionalGeneration")
+class Exaone4_5HfWeightMapper(HfWeightMapper):
+    def init_model_and_config(
+        self, model: Union[nn.Module, DecoderModelForCausalLM], config: ModelConfig
```
2ez4bz: Out of curiosity, is this ever a non-`DecoderModelForCausalLM`?

tensorrt_llm/_torch/models/modeling_exaone4_5.py
```diff
@@ -0,0 +1,261 @@
...
+except ImportError:
+
+    class Exaone4_5_VisionConfig(PretrainedConfig):
+        model_type = "exaone4_5_vision"
+        base_config_key = "vision_config"
+
+    AutoConfig.register(Exaone4_5_VisionConfig.model_type, Exaone4_5_VisionConfig, exist_ok=True)
+
+    class Exaone4_5Config(PretrainedConfig):
+        """VLM config: nested ``text_config`` / ``vision_config`` from JSON become real sub-configs."""
```
2ez4bz: Nit:

```
sed -i 's/``/`/g'
```

tensorrt_llm/_torch/models/modeling_exaone4_5.py
```diff
@@ -0,0 +1,261 @@
...
+                "image_grid_thw": processed_inputs.get("image_grid_thw"),
+            }
+
+        pixel_values_videos = processed_inputs.get("pixel_values_videos", None)
+        if pixel_values_videos is not None:
+            multimodal_data["video"] = {
+                "pixel_values_videos": pixel_values_videos.to(self.dtype),
+                "video_grid_thw": processed_inputs.get("video_grid_thw"),
+            }
+        fused_input_ids = processed_inputs["input_ids"][0]
```
2ez4bz, line 140: This seems to be where the implementation diverges from the base `Qwen2VLInputProcessorBase.__call__` - just checking:

1. Is it intentionally leaving out `mrope_config`?
2. If so, should we override `get_mrope_config` with `raise NotImplementedError` to prevent accidental misuse?

tensorrt_llm/_torch/models/modeling_exaone4_5.py
```diff
@@ -0,0 +1,261 @@
...
+            "multimodal_data": multimodal_data,
+        }
+
+
+class Exaone4_5_VisionModel(Qwen2VisionModelBase):
+    def __init__(
+        self, model_config: ModelConfig[PretrainedConfig], model_class: type[Qwen2_5_VisionModel]
+    ):
+        super().__init__(model_config, model_class=model_class)
+        self.config.tie_word_embeddings = False
```
2ez4bz: Could you leave a comment for why this line is necessary?

tensorrt_llm/_torch/models/modeling_exaone4_5.py
```diff
@@ -0,0 +1,261 @@
...
+        model_config: ModelConfig[PretrainedConfig],
+        *args,
+        **kwargs,
+    ) -> None:
+        self.original_arch = model_config.pretrained_config.architectures[0]
+
+        config = model_config.pretrained_config
+
+        self._supports_sdpa = True
+        PreTrainedModel.__init__(self, config)
```
2ez4bz, line 189: Hm, the fact that we're calling `PreTrainedModel.__init__(self, ...)` instead of `super().__init__(self, ...)` when we inherit from `Qwen2VLModelBase` seems conceptually a bit odd. We even override the `forward` method below.

Do we need to inherit from `Qwen2VLModelBase`?

**EDIT**: I guess you are using inheritance for `_get_requests_with_mm_data` and `load_weights`? The fact that we added 300 lines of code to `modeling_qwen2vl.py` to (presumably) only handle Exaone 4.5's specificities makes me think that code should be this file instead. WDYT?

tensorrt_llm/_torch/models/modeling_exaone4_5.py
```diff
@@ -0,0 +1,261 @@
...
+        content_format=ContentFormat.STRING,
+    ),
+)
+class Exaone4_5_ForConditionalGeneration(Exaone4_5_VLModel):
+    def __init__(self, model_config: ModelConfig[PretrainedConfig], *args, **kwargs):
+        kwargs["vision_model_class"] = Qwen2_5_VisionModel
+        super().__init__(model_config, *args, **kwargs)
+
+    @property
+    def multimodal_data_device_paths(self) -> List[str]:
```
2ez4bz: Should this be in `Exaone4_5_VLModel`?

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -305,9 +312,11 @@ def _preprocess(self, text: Dict[str, any], mm_data: Dict[str, any],
                                   **mm_processor_kwargs)
 
     def _postprocess(self, input_ids: torch.IntTensor) -> torch.IntTensor:
-        masks = (input_ids == self.config.image_token_id) | (
-            input_ids == self.config.vision_token_id) | (
-                input_ids == self.config.video_token_id)
+        masks = torch.zeros_like(input_ids, dtype=torch.bool)
+        for attr in ("image_token_id", "vision_token_id", "video_token_id"):
```
2ez4bz: Would `torch.isin` work here?

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -537,40 +601,81 @@ def __init__(self,
...
+                                                       torch.Tensor]] = None):
+        seq_len, _ = q.size()
+        cos, sin = position_embeddings
+
+        # FlashInfer fused RoPE assumes head_size is a multiple of 64 (see
+        # auto_deploy custom op rope docs / flashinfer tests). Qwen2.5-VL vision
+        # uses head_dim=80 (e.g. 1280 hidden / 16 heads), so use PyTorch RoPE.
+        if IS_FLASHINFER_AVAILABLE and self.head_dim % 64 == 0 and position_ids is not None:
+            try:
+                from ..custom_ops import \
```
2ez4bz: Out of curiosity, why is this import not at the module-level? If there is a specific reason, could you leave a comment for it?

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -694,8 +799,18 @@ def __init__(self, model_config: ModelConfig[PretrainedConfig]):
             embed_dim=self.config.hidden_size,
         )
 
-        head_dim = self.config.hidden_size // self.config.num_heads
-        self.rotary_pos_emb = Qwen2_5_VisionRotaryEmbedding(head_dim // 2)
+        self.config.max_position_embeddings = 8192
```
2ez4bz: Could you leave a comment for where these values come from?

tensorrt_llm/serve/chat_utils.py
```diff
@@ -344,7 +344,7 @@ def parse_chat_messages_coroutines(
     conversation = []
     mm_placeholder_counts = []
     mm_data_tracker = MultimodalDataTracker(
-        model_config.model_type,
+        type(model_config).model_type,
```
2ez4bz: Just checking: which changes in this PR made it a class-attribute that might not exist in an instance?

tests/unittest/_torch/modeling/test_modeling_exaone4_5.py
```diff
@@ -0,0 +1,220 @@
...
+    "vocab_size": 153600,
+    # Source of tokenizer / image processor / video processor files at runtime.
+    # Resolved against ``LLM_MODELS_ROOT`` (defaults to ``/code/llm-models``).
+    "_name_or_path": str(os.path.join(llm_models_root(), "EXAONE-4.5-33B")),
+}
+
+
+@dataclass(repr=False)
+class TestExaone4_5Scenario(MultimodalScenario):
+    """Scenario config (name avoids pytest collecting as Test* class)."""
```
2ez4bz: Is this still true with the current name?

tests/unittest/_torch/helpers.py
```diff
@@ -268,10 +268,12 @@ def make_hf_hybrid_cache_for_tests(
...
+    accuracy when comparing against TRT-LLM's per-layer KV management. Use
+    ``DynamicCache`` so HF's hybrid / sliding-window models manage cache state
+    layer-by-layer through their own ``layer_types`` metadata.
     """
-    del max_batch_size, device, dtype  # StaticCache doesn't accept these kwargs.
-    from transformers.cache_utils import StaticCache
+    from transformers.cache_utils import DynamicCache
 
-    return StaticCache(config=config, max_cache_len=max_cache_len)
+    return DynamicCache(config=config)
```
2ez4bz: Out of curiosity, what necessitated this change?

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: I think it is just legacy pattern from how we used in FLASHINFER's operation. Maybe it was to avoid the circular import. Let me verify.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
yechank-nvidia: 1. Correct. EXAONE does not require mrope metadata from inputprocessor.
2. Let me add it.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
yechank-nvidia: > Do we need to inherit from Qwen2VLModelBase?

The model architecture is the same as Qwen2VLModelBase. So, it would be easier for me to re-use some functions like `load_weights`.

> The fact that we added 300 lines of code to modeling_qwen2vl.py to (presumably) only handle Exaone 4.5's specificities makes me think that code should be this file instead. WDYT?

It is not. The reason for the change in  `modeling_qwen2vl.py` was to optimize the ViT's flow in Qwen2/2.5's ViT. It would also help for Qwen2.5 series. It is nothing to do with EXAONE's adoption.

tests/unittest/_torch/modeling/test_modeling_exaone4_5.py
yechank-nvidia: Thanks for the catch. Let me fix.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
yechank-nvidia: `multimodal_data_device_paths` is consumed via `getattr(self.model, ...)` on the registered auto-model instance, so functionally either placement works. 

We keep it on the registered leaf class to match the convention used by every sibling VL model (Qwen2_5_VLModel, Qwen3VLModel, Qwen3MoeVLModel, …) — the unregistered base (Qwen2VLModelBase, Qwen3VLModelBase, Exaone4_5_VLModel) is a shared scaffold and intentionally doesn't declare per-architecture tensor keys. Leaving it on Exaone4_5_ForConditionalGeneration.

yechank-nvidia: Actually, you are kind of right on this point. We don't need `Exaone4_5_VLModel` and `Exaone4_5_ForConditionalGeneration` at the same time for Exaone. Let me unify it.

tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
yechank-nvidia: Good catch — you're right. For this mapper, model is never a DecoderModelForCausalLM.

tensorrt_llm/serve/chat_utils.py
yechank-nvidia: Will add a note on it.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: It works. Will change.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
yechank-nvidia: It was needed for transformers upgrade, verified that it was not needed now.

tests/unittest/_torch/helpers.py
yechank-nvidia: Removed. Was a legacy from older version transformers compatibility.

tensorrt_llm/serve/chat_utils.py
yechank-nvidia: Added a Note.

examples/models/core/exaone/README.md
yechank-nvidia: Addressed. Thx.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
yechank-nvidia: Did. thx.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Noted.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Moved to module level.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
yechank-nvidia: Removed.

tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py
```diff
@@ -94,6 +94,30 @@ def test_nvfp4(self):
             task.evaluate(llm, sampling_params=self.sampling_params)
 
 
+class TestExaone4_5_33B(LlmapiAccuracyTestHarness):
```
crazydemo, line 103: please add this case to llm_function_core.txt for QA testing

xinhe-nv:
@coderabbitai review this code as a QA engineer, if function test and perf test is covered, both aggregated and disaggregated tests are necessary.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
```diff
@@ -0,0 +1,260 @@
...
+        return fused_input_ids.to(torch.int32).tolist(), {
+            "multimodal_data": multimodal_data,
+        }
+
+
+class Exaone4_5_VisionModel(Qwen2VisionModelBase):
+    def __init__(
+        self, model_config: ModelConfig[PretrainedConfig], model_class: type[Qwen2_5_VisionModel]
+    ):
+        super().__init__(model_config, model_class=model_class)
```
2ez4bz: Do we even still need the `__init__`? It looks like it's the same as the parent's.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
2ez4bz: Ok but it still looks odd here that we're calling `PreTrainedModel.__init__` instead of `super()`. Could you leave a comment explaining why?

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -518,12 +584,23 @@ def __init__(self,
                  reduce_output: bool = True) -> None:
 
         config = model_config.pretrained_config.vision_config
+        # Composite VLM configs (transformers 5.x strict mode) keep
+        # ``max_position_embeddings`` inside ``text_config`` rather than at
```
2ez4bz: Nit: probably good to just run this on all changed files:

```
sed -i 's/``/`/g'
```

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -518,12 +584,23 @@ def __init__(self,
...
+            max_position_embeddings = getattr(text_config,
+                                              "max_position_embeddings", None)
         super().__init__(
             hidden_size=config.hidden_size,
             num_attention_heads=config.num_heads,
-            num_key_value_heads=config.num_heads,
-            max_position_embeddings=model_config.pretrained_config.text_config.
-            max_position_embeddings,
+            num_key_value_heads=config.num_key_value_heads if getattr(
+                config, "num_key_value_heads", None) is not None else
```
2ez4bz: Nit: could just be

```
getattr(config, "num_key_value_heads", config.num_heads)
```

tensorrt_llm/_torch/models/modeling_qwen2vl.py
2ez4bz: Actually, for correctness, shouldn't we be matching what the old `Qwen2_5_VisionRotaryEmbedding` was doing? Is 8192 that value?

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -716,136 +854,187 @@ def __init__(self, model_config: ModelConfig[PretrainedConfig]):
...
-            window_seq_lens, self.window_attn_metadata)
-
-        # From this point, pure GPU operation
         hidden_states = self.patch_embed(pixel_values)
+
         seq_len, _ = hidden_states.size()
+        rope_position_ids = torch.arange(seq_len,
+                                         dtype=torch.int32,
+                                         pin_memory=prefer_pinned())
+        grid_rows = grid_thw.detach().cpu().tolist()
```
2ez4bz: Does this usually come in as a GPU tensor? If so, is there any way to exclude it from being sent to the GPU in the first place?

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -609,6 +608,77 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+    m_size: int,
+    dtype: torch.dtype,
+) -> torch.Tensor:
+    """Eager PyTorch bilinear position-embedding interpolation.
+
+    Returns a tensor of shape ``(t * h * w, hidden_dim)`` with the
+    bilinearly-interpolated position embeddings in spatial-merge order.
+    """
+    assert h % m_size == 0 and w % m_size == 0, (
+        f"h={h} and w={w} must be divisible by m_size={m_size}"
```
2ez4bz: Nit: f-strings allow using `f"{x=}"` instead of `f"x={x}"`.

tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py
```diff
@@ -94,6 +94,30 @@ def test_nvfp4(self):
...
+        stop="<|endofturn|>",
+    )
+
+    kv_cache_config = KvCacheConfig(free_gpu_memory_fraction=0.6)
+
+    def test_auto_dtype(self):
+        with LLM(
+            self.MODEL_PATH,
+            max_num_tokens=self.MAX_NUM_TOKENS,
+            kv_cache_config=self.kv_cache_config,
```
2ez4bz, line 130: Do you want functional coverage with chunked prefill as well? (Doesn't have to be a separate test case to save on CI time).

tests/unittest/_torch/modeling/test_modeling_exaone4_5.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def get_hf_inputs(self, modality: str, prompt: List[str], media: List[str]):
+        # On transformers < 5.8 there is no native EXAONE 4.5 processor, so
+        # ``AutoProcessor.from_pretrained`` returns a ``TokenizersBackend``
+        # (itself a ``PreTrainedTokenizerBase`` subclass) without a
+        # ``.tokenizer`` attribute. Monkey-patch ``AutoProcessor.from_pretrained``
+        # for the duration of the base implementation so it surfaces the
+        # backend as its own tokenizer. On transformers >= 5.8 the native
+        # processor already exposes ``.tokenizer`` and this is a no-op.
+        from transformers import AutoProcessor
```
2ez4bz: Nit: module-level.

tests/unittest/_torch/modeling/test_modeling_exaone4_5.py
```diff
@@ -0,0 +1,244 @@
...
+    releases the HF reference model is unavailable and HF-vs-TRT-LLM
+    comparison is skipped.
+    """
+
+    @property
+    def skip_hf_inference(self) -> bool:
+        return HFExaone4_5ForConditionalGeneration is None
+
+    @property
+    def skip_test(self) -> bool:
```
2ez4bz: Sorry but for consistency reasons I'm not a big fan of these `unittest.TestCase`. I believe we invoke unit tests with `pytest` anyway, so would it work if we just had a `@pytest.mark.skipif(...)` on top of this class instead of adding these new properties + modifying the base `TestModelingMultimodal`?

2ez4bz (approved):

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: No it comes as CPU tensor as it was removed from `multimodal_device_paths`. I think I was make sure to make it as CPU Tensor to List object.

tests/unittest/_torch/modeling/test_modeling_exaone4_5.py
yechank-nvidia: Done.

tests/unittest/_torch/modeling/test_modeling_exaone4_5.py
yechank-nvidia: Done.

tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py
yechank-nvidia: Added. Thx for letting me know.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
yechank-nvidia: Done.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Done. Thx.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Done

tensorrt_llm/_torch/models/modeling_exaone4_5.py
yechank-nvidia: Removed.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Converted back to LLM's value.

The reason for the previous change was vllm set to 8192, but couldn't find the reason behind that.

tensorrt_llm/_torch/models/modeling_exaone4_5.py
yechank-nvidia: Noted.

Superjomn (approved):
LGTM on the llmapi changes

fredricz-20070104 (approved):

tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
venkywonka, line 1: pls to add copyright header

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -626,8 +694,19 @@ def __init__(self, model_config: ModelConfig[PretrainedConfig]):
         self.pos_embed = nn.Embedding(self.config.num_position_embeddings, self.config.hidden_size)
         self.num_grid_per_side = int(self.config.num_position_embeddings**0.5)
 
-        head_dim = self.config.hidden_size // self.config.num_heads
-        self.rotary_pos_emb = HFQwen3VLVisionRotaryEmbedding(head_dim // 2)
+        self.config.max_position_embeddings = 8192
```
venkywonka: is this hardcode intentional?

yechank-nvidia: Thx for the catch! It somehow reverted to previous version. Let me fix it.

yechank-nvidia: Changed.

tensorrt_llm/_torch/models/checkpoints/hf/exaone4_5_weight_mapper.py
yechank-nvidia: Added. Thx.

juney-nvidia (approved):

PR #12893, opened 2026-04-09
[None][fix] Fix missing disagg_request_id fallback in context responses

pich4ya:
## Summary

This change makes the disaggregated proxy tolerant of context-phase responses that include `ctx_request_id` but omit `disagg_request_id`.

Instead of failing hard in `_verify_ctx_response`, the proxy now falls back to `ctx_request_id` and logs a warning.

## Why

On a live 2-node DGX Spark disaggregated setup (`gpt-oss-120b`, UCX KV-cache transfer over RoCE/RDMA), the context server returned:
- `disaggregated_params.ctx_request_id != None`
- `disaggregated_params.disagg_request_id == None`

The proxy then raised:

```text
Invalid disaggregated params in context phase response. disagg_request_id is None
```

That prevented an otherwise healthy context/generation/proxy stack from serving requests.

## Rationale for the fallback

For context-first disaggregated flow, `ctx_request_id` is already the request identifier that the generation side uses to continue the request. When `disagg_request_id` is absent but `ctx_request_id` is present, using `ctx_request_id` preserves the existing request identity instead of aborting the request.

This is intentionally conservative:
- still errors if `disaggregated_params` is missing
- still errors if `ctx_request_id` is missing
- only fills `disagg_request_id` when it is the sole missing field
- emits a warning so the underlying worker-side omission remains visible

## Test

Added a regression test to verify `_verify_ctx_response` accepts a context-phase response with:
- `ctx_request_id` set
- `disagg_request_id` unset

and backfills `disagg_request_id` from `ctx_request_id`.

## Validation performed

Local validation:
- `python3 -m py_compile tensorrt_llm/serve/openai_disagg_service.py tests/unittest/disaggregated/test_openai_disagg_service.py`

I could not run the full pytest target in this lightweight clone because the local environment does not include TensorRT-LLM's Python test dependencies (for example `transformers`).

## Reproduction environment

Observed on:
- 2-node DGX Spark setup
- TensorRT-LLM `1.3.0rc10`
- disaggregated serving
- context server on node 1
- generation server on node 2
- proxy on node 1
- UCX KV-cache transfer over RoCE/RDMA

## Dev Engineer Review

- Added a fallback for missing `disagg_request_id` in context responses.
- Applies the fallback only when `ctx_request_id` matches the expected request ID.
- Rejects missing or mismatched context IDs.
- Emits the warning once with a stable key.
- Preserves retry-updated request IDs.
- No configuration or public API changes were introduced.

## QA Engineer Review

- Updated the context retry test for missing `disagg_request_id`.
- Added coverage for matching fallback, mismatched IDs, response mutation, null IDs, and bounded warning calls.
- No corresponding `tests/integration/test_lists/`, `test-db/`, or `qa/` entries were changed.
- Verdict: needs follow-up for CI test-list coverage.

### DIFF
--- a/tensorrt_llm/serve/openai_disagg_service.py
+++ b/tensorrt_llm/serve/openai_disagg_service.py
@@ -372,10 +372,14 @@ class OpenAIDisaggregatedService(OpenAIService):
                             f" disagg_request_id={choice.disaggregated_params.disagg_request_id!r}"
                         )
                     if choice.disaggregated_params.disagg_request_id is None:
-                        raise ValueError(
-                            f"Invalid disaggregated params: disagg_request_id is None for choice {idx}."
-                            f" finish_reason={choice.finish_reason!r},"
-                            f" ctx_request_id={choice.disaggregated_params.ctx_request_id!r}"
+                        logger.warning(
+                            "Context server choice %d is missing disagg_request_id; "
+                            "falling back to ctx_request_id=%s.",
+                            idx,
+                            choice.disaggregated_params.ctx_request_id,
+                        )
+                        choice.disaggregated_params.disagg_request_id = (
+                            choice.disaggregated_params.ctx_request_id
                         )
             return ctx_response
 
--- a/tensorrt_llm/serve/openai_disagg_service.py
+++ b/tensorrt_llm/serve/openai_disagg_service.py
@@ -146,7 +146,9 @@ class OpenAIDisaggregatedService(OpenAIService):
                 ctx_response = await self._ctx_client.send_request(
                     ctx_req, server=ctx_server, hooks=hooks, req_id=disagg_request_id
                 )
-                await self._verify_ctx_response(ctx_response)
+                await self._verify_ctx_response(
+                    ctx_response, ctx_req.disaggregated_params.disagg_request_id
+                )
                 ctx_response_disagg_params = ctx_response.choices[0].disaggregated_params
                 if ctx_response_disagg_params.disagg_request_id is not None:
                     disagg_request_id = ctx_response_disagg_params.disagg_request_id
@@ -352,7 +354,11 @@ class OpenAIDisaggregatedService(OpenAIService):
         await self._gen_client.shutdown()
         await self._coordinator.stop()
 
-    async def _verify_ctx_response(self, ctx_response: UCompletionResponse) -> None:
+    async def _verify_ctx_response(
+        self,
+        ctx_response: UCompletionResponse,
+        expected_disagg_request_id: Optional[int],
+    ) -> None:
         if ctx_response:
             for idx, choice in enumerate(ctx_response.choices):
                 if choice.disaggregated_params is None:
@@ -372,11 +378,18 @@ class OpenAIDisaggregatedService(OpenAIService):
                             f" disagg_request_id={choice.disaggregated_params.disagg_request_id!r}"
                         )
                     if choice.disaggregated_params.disagg_request_id is None:
-                        logger.warning(
-                            "Context server choice %d is missing disagg_request_id; "
-                            "falling back to ctx_request_id=%s.",
-                            idx,
-                            choice.disaggregated_params.ctx_request_id,
+                        if choice.disaggregated_params.ctx_request_id != expected_disagg_request_id:
+                            raise ValueError(
+                                f"Invalid disaggregated params: ctx_request_id="
+                                f"{choice.disaggregated_params.ctx_request_id!r} for choice {idx}"
+                                " does not match the successful context attempt's"
+                                f" disagg_request_id={expected_disagg_request_id!r}."
+                            )
+                        logger.warning_once(
+                            f"Context server choice {idx} is missing disagg_request_id; "
+                            f"falling back to verified ctx_request_id="
+                            f"{choice.disaggregated_params.ctx_request_id}.",
+                            key="missing_disagg_request_id_fallback",
                         )
                         choice.disaggregated_params.disagg_request_id = (
                             choice.disaggregated_params.ctx_request_id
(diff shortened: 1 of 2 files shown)


hchings (approved):
LGTM

karljang:
@pich4ya , 
We’ve received approval, but could you please address the failing DCO check to proceed? For more details, refer to [CONTRIBUTING.md#signing-your-work](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CONTRIBUTING.md#signing-your-work).

Thanks!

karljang:
Oops, we’ve encountered some conflicts because the disagg is hot these days. Could you please resolve these conflicts as well?
Thanks!

pich4ya:
Rebased onto current `main` and force-pushed commit `53085f7b52` with a DCO sign-off. The previously reported merge conflicts are resolved; the fallback warning and regression test are retained. The current DCO check is passing. @karljang, could you please re-check when convenient?

pich4ya:
@karljang, thanks for triggering CI. The L0 run on `53085f7` reports `9538 passed, 12 failed, 901 skipped`, but I can’t access the internal Jenkins/CI report from my current network to see the failed-test details.

Could you please share the failed test names and relevant logs (or the Jenkins test report)? I’ll investigate and address any PR-related regression promptly.

karljang:
the failures appear unrelated to the PR, I'll rerun it.

Tabrizian:
@pich4ya Can you share a reproducer for the issue you are seeing? I think if you set `disagg_request_id` from the client side you won't run into this issue.

tensorrt_llm/serve/openai_disagg_service.py
```diff
@@ -372,10 +372,14 @@ async def _verify_ctx_response(self, ctx_response: UCompletionResponse) -> None:
                             f" disagg_request_id={choice.disaggregated_params.disagg_request_id!r}"
                         )
                     if choice.disaggregated_params.disagg_request_id is None:
-                        raise ValueError(
-                            f"Invalid disaggregated params: disagg_request_id is None for choice {idx}."
-                            f" finish_reason={choice.finish_reason!r},"
-                            f" ctx_request_id={choice.disaggregated_params.ctx_request_id!r}"
+                        logger.warning(
```
chienchunhung: `logger.warning` is TensorRT-LLM’s variadic wrapper, so it joins arguments instead of interpolating `%d` and `%s`. This also warns on every affected response, creating unbounded log traffic on the serving hot path. Please pre-format one message and use `logger.warning_once(..., key=...)` or another bounded mechanism

tensorrt_llm/serve/openai_disagg_service.py
```diff
@@ -372,10 +372,14 @@ async def _verify_ctx_response(self, ctx_response: UCompletionResponse) -> None:
...
+                            "Context server choice %d is missing disagg_request_id; "
+                            "falling back to ctx_request_id=%s.",
+                            idx,
+                            choice.disaggregated_params.ctx_request_id,
+                        )
+                        choice.disaggregated_params.disagg_request_id = (
+                            choice.disaggregated_params.ctx_request_id
```
chienchunhung, line 395: We should avoid promoting an unverified `ctx_request_id` to `disagg_request_id`. 

The latter is the globally unique ctx↔gen transfer key, but the former can be worker-local when the context worker did not consume the orchestrator-supplied ID. Different context workers can therefore produce duplicate values, turning the previous fail-fast into KV-session collisions. 

Please carry the ID used by the successful context attempt into `_verify_ctx_response`, require `ctx_request_id` to match it, and backfill only then. 

The test should cover a retry-updated matching ID and reject a mismatched/local ID.

Tabrizian: @chienchunhung Not sure I understand this comment. In the past ctx_request_id was used as the global transfer key

karljang:
@pich4ya, could you please review the reviewer’s comments at your convenience?

tensorrt_llm/serve/openai_disagg_service.py
pich4ya: Good catch. Updated in 65acb8f19c to pre-format the message and use warning_once with a stable key, so the wrapper emits the intended text and the hot-path warning is bounded. The regression test now asserts the exact warning_once call.

tensorrt_llm/serve/openai_disagg_service.py
pich4ya: Agreed. Updated in 65acb8f19c so the caller passes the disagg_request_id from ctx_req after the context send succeeds, which captures any retry-time ID replacement. The fallback now requires the returned ctx_request_id to match that successful-attempt ID; a mismatch raises and leaves disagg_request_id unset. Tests cover both a retry-updated matching ID and a mismatched/local ID.

pich4ya:
@chienchunhung @Tabrizian Thanks for the review. I addressed both points in commit `65acb8f19c`:

- The fallback warning is now pre-formatted and emitted through `warning_once` with a stable key.
- The verifier receives the `disagg_request_id` from the successful context attempt after `send_request` returns, so retry-time ID replacement is included. It backfills only when the returned `ctx_request_id` matches that value; a mismatched/local ID now fails closed.
- Tests cover the retry-updated matching case, the mismatch rejection, the mutation, and the bounded warning call.

The original failure was observed on TensorRT-LLM 1.3.0rc10 in a two-node DGX Spark context-first deployment using UCX over RoCE/RDMA. I no longer have a minimal standalone reproducer or the original request/response ID trace from that deployment. The stricter equality check keeps the compatibility fallback without treating an unverified worker-local ID as a global transfer key.

Repository pre-commit hooks and static checks pass. The focused pytest target could not be collected in this local checkout because its Python environment lacks `torch` and `transformers`, so this new commit still needs CI validation. @karljang, could you please rerun CI when convenient? A re-review would also be appreciated.

karljang:
@chienchunhung, could you please review this PR again? Thanks.

pcastonguay:
@chienchunhung can you resolve the open thread if you're ok with the changes?

PR #12896, opened 2026-04-09
[TRTLLM-12026][feat] Support MTP with block reuse enabled for hybrid models

VALLIS-NERIA:
## Summary
- Extend KVCacheManager C++ layer with mutable cache block ID accessors and logging for linear cache memory budget calculation
- Refactor MambaCacheManager to support block-level cache operations (add/remove/reuse) instead of flat tensor management
- Update scheduler to handle linear attention cache reuse alongside KV cache reuse
- Wire through `enable_cache_reuse` flag for mamba cache in executor and resource manager
- Add integration test for mamba2 hybrid model cache reuse
- Add unit tests for KV cache manager with linear attention metadata

## Test plan
- [x] Integration test for mamba2 hybrid model with cache reuse
- [x] Existing KV cache reuse tests still pass

## Release Notes

* **New Features**
  * Added speculative decoding support for Mamba models with improved cache management.

* **Improvements**
  * Enhanced KV cache estimation for linear-attention layers.
  * Better error reporting and validation messaging for cache block operations.
  * Optimized cache block handling for recurrent state scenarios.

* **Tests**
  * Added MTP speculative decoding test configurations and block reuse validation.


tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -418,6 +418,7 @@ def append_to_kv_heads_per_layer(num_kv_heads_per_layer: List[int],
             # max_tokens under _util.py::try_prepare_estimation
             # Since this is a dry run, assigning the same max_tokens capacity
             # to all window sizes as they are full attentions is enough.
+
```
eopXD: Unnecessary diff here.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -2673,6 +2675,35 @@ void BlockManager::releaseLastBlock(GenerationRequest& sequence, SizeType32 wind
 
 void WindowBlockManager::releaseLastBlock(GenerationRequest& sequence)
 {
+    if (isRecurrentState())
+    {
+        // In recurrent state, the last block always contains the current state and should not be released.
+        // Since the only caller of releaseLastBlock is speculative decoding rewinding, it only happens in decoding
```
eopXD, line 2699: Guarantee `Since the only caller of releaseLastBlock is speculative decoding rewinding` with an assertion to fool-proof the use of this function.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -433,13 +434,51 @@ def append_to_kv_heads_per_layer(num_kv_heads_per_layer: List[int],
...
+                    )
+                    # self.max_attention_window_vec is a pattern, repeat it to match num_layers
+                    self.max_attention_window_vec = (
+                        self.max_attention_window_vec *
+                        (self.num_layers // len(self.max_attention_window_vec) +
+                         1))[:self.num_layers]
+                    logger.info(
+                        f"Adjusted max_attention_window_vec for linear attention: {self.max_attention_window_vec}"
+                    )
+                # _util.py::try_prepare_estimation can't estimate linear attentions properly
```
eopXD: Instead of patching here, extending `_get_token_num_for_estimation` (or adding a helper called from it) to account for linear-attention overhead would centralize the correction at its source.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -433,13 +434,48 @@ def append_to_kv_heads_per_layer(num_kv_heads_per_layer: List[int],
...
+                total_bytes_per_token += bytes_per_linear_block * self.max_batch_size // kv_cache_config.max_tokens
+                max_snapshots = self.max_batch_size
+                if kv_cache_config.enable_block_reuse:
+                    total_bytes_per_token += bytes_per_linear_block // linear_attention_metadata.states_snapshot_interval
+
+                expand_factor = total_bytes_per_token / self.get_cache_bytes_per_token(
+                )
+
+                kv_cache_config.max_tokens = int(kv_cache_config.max_tokens //
+                                                 expand_factor)
```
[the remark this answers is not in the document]
eopXD: +1, a valid concern

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -709,8 +749,29 @@ def prepare_resources(self, scheduled_batch: ScheduledRequests):
...
+            # (mContextChunkSizeTarget / mContextChunkSizeDraft).  The
+            # scheduler only sets the *target* chunk size, so the draft
+            # field keeps its default value (= promptLen).  When
+            # reset_context_requests reads is_last_context_chunk under
+            # use_draft_model=True it sees the unmodified draft chunk size,
+            # which covers the full prompt and therefore evaluates to True,
+            # incorrectly promoting the request to context_requests_last_chunk.
+            # That reclassification persists after the context manager
+            # restores use_draft_model=False, breaking downstream sampling
+            # and MTP verification.
```
eopXD: If so, why don't we just call this `scheduled_batch.reset_context_requests()` outside of draft context?

VALLIS-NERIA: fixed

tensorrt_llm/_torch/pyexecutor/resource_manager.py
VALLIS-NERIA: Pushed another commit to refactor the memory estimation calculation

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
VALLIS-NERIA: This is hard to implement. I don't think we want stacktrace in a function that is frequently called.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
VALLIS-NERIA: fixed

tensorrt_llm/_torch/pyexecutor/resource_manager.py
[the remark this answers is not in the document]
VALLIS-NERIA: refactored and now the upstream runtime calculates the correct memory

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -2673,6 +2675,35 @@ void BlockManager::releaseLastBlock(GenerationRequest& sequence, SizeType32 wind
...
+        allocatedBlocks.erase((++it).base());
+        // Remove stored block ids in sequence
+        auto beamWidth = sequence.getBeamWidth();
+        TLLM_CHECK_WITH_INFO(beamWidth == 1, "Spec decoding with linear attention does not support beam search");
+        for (auto beamIdx = 0; beamIdx < beamWidth; ++beamIdx)
+        {
+            sequence.getCacheBlockIds(mWindowSize)[beamIdx].erase(
+                sequence.getCacheBlockIds(mWindowSize)[beamIdx].end() - 2);
+        }
+        return;
```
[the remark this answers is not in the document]
VALLIS-NERIA: This is not true. In decoding phase only the last block is real and all previous blocks are placeholders. Spec decoding rewind never rewinds to context phase.

tests/integration/test_lists/test-db/l0_gb200_multi_gpus.yml
```diff
@@ -57,6 +57,7 @@ l0_gb200_multi_gpus:
   - accuracy/test_llm_api_pytorch.py::TestNemotronV3Super::test_nvfp4_4gpus_online_eplb[moe_backend=CUTLASS]
   - accuracy/test_llm_api_pytorch.py::TestNemotronV3Super::test_nvfp4_4gpus_block_reuse[TEP4]
   - accuracy/test_llm_api_pytorch.py::TestNemotronV3Super::test_nvfp4_4gpus_block_reuse[TEP4_ADP]
+  - accuracy/test_llm_api_pytorch.py::TestNemotronV3Super::test_nvfp4_4gpus_block_reuse[TEP4_ADP_MTP]
```
xinhe-nv, line 60: please add this case into qa test list, too.

xinhe-nv (approved):

xinhe-nv:
@coderabbitai review the code and test list as a QA engineer.

nvpohanh:
@eopXD could you review this? thanks!

tests/integration/test_lists/test-db/l0_gb200_multi_gpus.yml
VALLIS-NERIA: done

tensorrt_llm/_torch/pyexecutor/config_utils.py
```diff
@@ -99,6 +101,146 @@ def get_qwen3_hybrid_num_attention_layers(config):
...
+@dataclasses.dataclass
+class MambaKVCacheParams:
+    """Normalized mamba-related inputs for kv_cache_manager_cls.
+
+    Field naming is consistent across the Nemotron-hybrid and Qwen3-hybrid
+    families so call sites don't special-case per family.
+    """
+    # Mamba / linear-attention state dimensions
+    state_size: int  # ssm_state_size      | linear_key_head_dim
+    conv_kernel: int  # conv_kernel         | linear_conv_kernel_dim
```
hnover-nv, line 123: So this is how historical the kernel is?  This would be 4 today (maybe 3)?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -780,7 +780,17 @@ def update_mamba_states(self, attn_metadata: "AttentionMetadata",
                                        state_indices)
 
 
-class MixedMambaHybridCacheManager(KVCacheManager, MambaCacheManager):
+class MambaHybridCacheManager:
+    """Marker base class for hybrid mamba cache manager implementations.
```
hnover-nv, line 805: Thank you!
But does it not need the method signatures, for the type checkers to be happy?  Otherwise wont' the checkers say MambaHybridCacheManager doesn't have method foo, when we hint something is a MambaHybridCacheManager?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -1062,20 +1099,103 @@ def __init__(
...
+        # Attention slope from the parent's existing formula.
+        attention_slope = KVCacheManager.get_cache_size_per_token(
+            model_config, mapping, num_layers=num_layers, **kwargs)
+
+        params = extract_mamba_kv_cache_params(
+            model_config.pretrained_config,
+            quant_config=model_config.quant_config,
+        )
+
+        tp_size = mapping.tp_size if not mapping.enable_attention_dp else 1
```
hnover-nv: I find it confusing what is divided by tp and what isn't.  Could we use foo for overall and then foo_local for divided by TP, to avoid any confusion?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv, line 313: This, vs identical d_inner_local above, snuck through a rebase in my change.  Could you clean it up while you're here?  My preference (I said in anothe comment) is we adopt foo_local for all "divided by TP" quantities, leave 'foo" to mean the whole model, to avoid ever getting confused.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -1062,20 +1099,103 @@ def __init__(
...
+        conv_bytes = conv_dim * (params.conv_kernel - 1) * conv_dtype.itemsize
+        ssm_bytes = (nheads * params.head_dim * params.state_size *
+                     ssm_dtype.itemsize)
+        state_bytes_per_layer = conv_bytes + ssm_bytes
+
+        # Per-rank mamba layer count, mirroring how the parent slices PP for
+        # attention (len(mapping.pp_layers(global_attn_count))).
+        num_mamba_layers_per_rank = len(
+            mapping.pp_layers(params.num_mamba_layers))
+        state_bytes_per_rank = num_mamba_layers_per_rank * state_bytes_per_layer
```
hnover-nv, line 1176: this formula is missing replay,and I think also legacy path?  legacy path is a little more complicated, because in theory we allocate that to max batch size, not max # of requests (e.g. memory for 1024 requets, but batch size could be 512). But I think the plumbing today sizes both slots and intermediate at max batch size.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -1062,20 +1099,103 @@ def __init__(
...
+
+        # Per-request fixed cost. STATIC_SLOTS_PER_REQUEST = 1 today (the
+        # live mamba state); fixed-position snapshots are not yet
+        # implemented and would simply increment this constant.
+        STATIC_SLOTS_PER_REQUEST = 1
```
hnover-nv, line 1181: So there are both fixed position snapshots and periodic snapshots?  We don't yet do fixed position, butwe do intervals/. Is fixed position basically just irregular intervals?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -1125,34 +1246,112 @@ def _prepare_resources(self, scheduled_batch: ScheduledRequests):
...
+        # Reset replay double-buffer state for fresh context blocks. A reused
+        # block (prefix-cache hit or block recycled across requests) may carry
+        # stale prev_num_accepted_tokens / cache_buf_idx values from a prior
+        # owner; the replay kernel reads these on the first decode step.
+        if self._use_replay_state_update and self.prev_num_accepted_tokens is not None:
+            num_contexts = len(scheduled_batch.context_requests)
+            if num_contexts > 0:
+                ctx_slots = self.cuda_state_indices[:num_contexts].long()
+                self.prev_num_accepted_tokens[ctx_slots] = 0
+                self.cache_buf_idx[ctx_slots] = 0
```
hnover-nv: Nit: you don't need to reset teh cache_buf_idx, that just points into a double buffer and you don't care which half you start writing into.  prev_num_accepted_tokens is the only required reset.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -1247,84 +1446,111 @@ def _setup_states_views(self) -> None:
...
-            return issubclass(
-                subclass,
-                (MixedMambaHybridCacheManager, CppMambaHybridCacheManager))
-        return super().__subclasscheck__(subclass)
-
-    def __getattr__(cls, name):
-        """Forward class-level attribute access (e.g. static methods) to
-        KVCacheManager. Add attributes here as needed."""
-        return getattr(KVCacheManager, name)
+    def _setup_mtp_intermediate_states(self, spec_config,
```
hnover-nv, line 1481: In replay they're not really called intermediate states.  Maybe _setup_mtp_state or _setup_mtp_buffers or _setup_mtp_cache?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -1247,84 +1446,111 @@ def _setup_states_views(self) -> None:
...
+        device = self.all_ssm_states.device
+
+        # Shared across layers (consumed by the replay kernel via slot index).
+        self.prev_num_accepted_tokens = torch.zeros(cache_size,
+                                                    dtype=torch.int32,
+                                                    device=device)
+        self.cache_buf_idx = torch.zeros(cache_size,
+                                         dtype=torch.int32,
+                                         device=device)
+        # Per-layer double-buffered caches.
```
hnover-nv: x is not double buffered, but it is per layer.  The others are double buffered.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -1247,84 +1446,111 @@ def _setup_states_views(self) -> None:
...
-            mamba_cache_dtype,
-            mamba_ssm_cache_dtype,
-            kv_cache_config,
-            kv_cache_type,
-        )
+        T = spec_config.max_draft_len + 1
+        num_local_mamba_layers = self.local_num_mamba_layers
+        # all_ssm_states: [num_local_mamba_layers, num_blocks_in_pool, ...]
+        cache_size = self.all_ssm_states.shape[1]
+        nheads, head_dim, d_state = self.ssm_state_shape
```
hnover-nv, line 1541: nheads is after TP here?

hnover-nv (commented):
Sorry, I'm a bit wiped tonight, will pick this up tomorrow.  But sending what I have now in case you want to look.

Question, I am working on a follow-on that would tweak the sizes of the old_* buffers (T would be something else) and change the update_mamba_state logic.  It looks like that is completey within python, so if I just update all the implementations in mamba cache manager, that would work ok?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: Everything here are local. We don't ever need the global dim right?

tensorrt_llm/_torch/pyexecutor/config_utils.py
VALLIS-NERIA: This is the convolution window. 3 in Qwen.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: added BaseResourceManager, BaseMambaCacheManager as base classes

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: Everything here are local. We don't ever need the global dim right?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: It only counts memory to be managed by the C++ part. Python buffers are not included.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: It is the active state that is used in decoding phase.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: commented out it and left a comment as hint.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: moved the replay part to _setup_replay_buffers

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: fixed

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
VALLIS-NERIA: yes

tensorrt_llm/_torch/pyexecutor/config_utils.py
hnover-nv: conv_kernel_width might be clearer then, if this isn't an inherited name from elsewhere in the code base.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv: ty!

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv: we don't, but reading it through is hard to keep straight without some more suffixes or other naming convention, given we get passed/extract the global values.
Fundamentally we divide the heads up, so if you just do n_heads/TP, n_groups/TP, then everythign falls out.  But the code doesn't do that, we calculate conv_dim from non-TP values, then divide nheads and n_groups later, and conv_dim.    But then we calculate d_inner from the non-TP values and divide that too.
So when reading the derivations, it's hard to tell, when it's a global value vs a post-TP value.  So when you divide TP again, is that the right thing to do, or are you combining two TP values, etc.
Just for clarify if correctness, I feel like we should refactor them to at least have local on the names appropriately, or fine do it in-place, but divide by TP first thing nheads and ngroups and then only use TP-adjusted values thereafter, no need to divide again.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv: Right, see comment above for what I was going for in my prior PR, but apparently messed up in a rebase.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv: Right, that's what it is when it's 1.  I was just wondering about the comment, when will it be >1, vs the approach  below about incremental snapshots.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv: In my diff view it still whows as live?  If you could fix the comment and remove this.  I see the python version has it too, sorry about that, if you could remove there.  Not a big deal but might as well save the write.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv: Hmm, maybe you just didnt' push?  I hae relaoded github a few times and the version I'm seeing still shows the old code.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv: Ah, ok.  I don't understand the design choice behind what is managed in C++ vs python.  If it's what comes over from disagg prefill, then sure, that's fine, but we'll need to revisit when we implement decode reuse?

hnover-nv (approved):
Thanks.  Just the nits.  Separately I need to undersatnd how the kv cache reuse / state caching works for mamba, but that can be follow-up.

StanleySun639 (approved):

nv-guomingz (approved):

Wanli-Jiang (approved):

tensorrt_llm/_torch/pyexecutor/config_utils.py
VALLIS-NERIA: It comes from the original `MambaForCausalLM`

pcastonguay (approved):

eopXD (approved):

mikeiovine (approved):

PR #12911, opened 2026-04-10
[TRTLLM-11794][feat] Optimize ViT Attention kernel on Nemotron

yechank-nvidia:
## Description

This PR introduces new Flashinfer Backend `BatchPrefillWithRaggedKVCacheWrapper` for the case where kv_cache_manger is None.(e.g. Vision Attention case)

For the case of seq_len > 1024, FLASHINFER's cudnn backend seems promising result. And the image data is on average over seq_len > 1024, so it is convincing that changing to FLASHINFER's cudnn is good. 

Below is the data from microbenchmark.
```
batch=1 seq_lens=[128, 256, 512, 1024, 2048, 4096, 8192] dim=1280 heads=16 dtype=bfloat16 warmup=10 iters=50

===================================================================================================
seq_len | FLASHINFER_ms |  TRTLLM_ms  | FLASHINFER_µs/tok | TRTLLM_µs/tok | ratio_FLASHINFER/TRTLLM
---------------------------------------------------------------------------------------------------
    128 |       0.106 |       0.091 |       0.829 |       0.713 |          1.162
    256 |       0.111 |       0.089 |       0.433 |       0.349 |          1.240
    512 |       0.110 |       0.089 |       0.214 |       0.173 |          1.238
   1024 |       0.110 |       0.090 |       0.107 |       0.088 |          1.223
   2048 |       0.110 |       0.157 |       0.054 |       0.077 |          0.699
   4096 |       0.156 |       0.480 |       0.038 |       0.117 |          0.325
   8192 |       0.420 |       1.743 |       0.051 |       0.213 |          0.241
===================================================================================================

```

## Summary by CodeRabbit

* **New Features**
  * Added support for ragged KV cache prefill without explicit cache management, enabling more efficient context batch processing
  * Added configurable vision attention backend parameter for multimodal models

* **Tests**
  * Added test coverage for ragged prefill mode validation

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/flashinfer.py
+++ b/tensorrt_llm/_torch/attention_backend/flashinfer.py
@@ -236,10 +236,10 @@ class FlashInferAttentionMetadata(AttentionMetadata):
         assert self.seq_lens_cuda is not None
         assert self.seq_lens is not None
 
-        q_seqlens = self.seq_lens[:self.num_contexts].to(dtype=torch.int32)
-        kv_seqlens = (q_seqlens if not self.is_cross else
-                      self.seq_lens_kv[:self.num_contexts].to(
-                          dtype=torch.int32))
+        # NOTE: When kv_cache_manager is None (e.g. ViT), ragged prefill runs only for the context phase.
+        # Restrict seq_lens to the first num_contexts entries accordingly.
+        q_seqlens = self.seq_lens[:self.num_contexts]
+        kv_seqlens = q_seqlens
 
         max_query_tokens_per_sequence = int(
             self.seq_lens[:self.num_contexts].max().item())
@@ -262,22 +262,14 @@ class FlashInferAttentionMetadata(AttentionMetadata):
             num_key_value_heads = plan_params.num_kv_heads
             attention_head_dim = plan_params.head_dim
             query_output_element_indptr[1:].copy_(
-                torch.cumsum(
-                    q_seqlens.to(torch.int64) * int(num_query_output_heads) *
-                    int(attention_head_dim),
-                    dim=0,
-                ).to(torch.int32))
+                torch.cumsum(q_seqlens, dim=0).mul_(num_query_output_heads *
+                                                    attention_head_dim))
             key_value_element_indptr[1:].copy_(
-                torch.cumsum(
-                    kv_seqlens.to(torch.int64) * int(num_key_value_heads) *
-                    int(attention_head_dim),
-                    dim=0,
-                ).to(torch.int32))
+                torch.cumsum(kv_seqlens, dim=0).mul_(num_key_value_heads *
+                                                     attention_head_dim))
 
-        q_seqlens_cuda = self.seq_lens_cuda[:self.num_contexts].to(
-            dtype=torch.int32)
-        kv_seqlens_cuda = (self.seq_lens_kv_cuda[:self.num_contexts].to(
-            dtype=torch.int32) if self.is_cross else q_seqlens_cuda)
+        q_seqlens_cuda = self.seq_lens_cuda[:self.num_contexts]
+        kv_seqlens_cuda = q_seqlens_cuda[:self.num_contexts]
 
         ragged_prefill_wrapper.plan(
             qo_indptr=query_output_element_indptr,
--- a/tensorrt_llm/_torch/attention_backend/flashinfer.py
+++ b/tensorrt_llm/_torch/attention_backend/flashinfer.py
@@ -2,7 +2,7 @@ import math
 import os
 import weakref
 from dataclasses import dataclass, field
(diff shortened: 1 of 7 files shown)


tensorrt_llm/_torch/attention_backend/flashinfer.py
2ez4bz: Nit: since `num_query_output_heads` and `attention_head_dim` are scalars, should we just multiply the result of `cumsum` instead of its input? Similar question below.

Also, out of curiosity, why do we do a int32 -> int64 -> int32 roundtrip...?

2ez4bz, line 245: Just checking: is it a safe assumption the first `:self.num_contexts` entries are for context? Maybe leave a comment saying why?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
2ez4bz, line 1625: Just curious, why the deepcopy calls?

tensorrt_llm/_torch/models/modeling_radio.py
2ez4bz: This seems a bit overkill, no? Can't the default value for `vision_attn_backend` just be set to `FLASHINFER` directly? Children can still override if necessary.

2ez4bz (commented):
Thanks for this!

2ez4bz (approved):
Approving to unblock.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
yechank-nvidia: In the `vision_model_config`, when we initialize the VisionModel, I am modifying its Attention Backend to FLASHINFER, so not want to affect the LLM's element.

tensorrt_llm/_torch/attention_backend/flashinfer.py
yechank-nvidia: > Also, out of curiosity, why do we do a int32 -> int64 -> int32 roundtrip...?

This was removed. Thanks for pointing out.

>  since num_query_output_heads and attention_head_dim are scalars, should we just multiply the result of cumsum instead of its input? Similar question below.

Also, removed the redundant operation.

yechank-nvidia: Comments added. TLDR; This was added for ViT when kv_cache_manger is None, so I thought it was safe to assume to trim them into first `:self_num_contexts`.

tensorrt_llm/_torch/models/modeling_radio.py
yechank-nvidia: It was. Removed to just to set vision_attn_backend as default to "FLASHINFER".

tensorrt_llm/_torch/attention_backend/flashinfer.py
```diff
@@ -46,10 +49,13 @@ class PlanParams:
...
 class FlashInferWrappers:
-    decode_wrapper: flashinfer.BatchDecodeWithPagedKVCacheWrapper
-    prefill_wrapper: Optional[flashinfer.BatchPrefillWithPagedKVCacheWrapper]
-
     is_planned: bool
+    decode_wrapper: Optional[
+        flashinfer.BatchDecodeWithPagedKVCacheWrapper] = None
+    prefill_wrapper: Optional[
+        flashinfer.BatchPrefillWithPagedKVCacheWrapper] = None
+    ragged_prefill_wrapper: Optional[
```
brb-nv, line 57: You mentioned this is for the case where `KVCacheManager` is None. What about `BatchPrefillWithRaggedKVCacheWrapper` is suitable for this? Why can't we use the existing wrappers? Can you please let me know?

brb-nv (approved):
Just an informational question. LGTM.

tensorrt_llm/_torch/attention_backend/flashinfer.py
yechank-nvidia: AFAIK, `BatchPrefillWithRaggedKVCacheWrapper` requires kv_cache in [plan()](https://docs.flashinfer.ai/api/attention.html#flashinfer.decode.BatchDecodeWithPagedKVCacheWrapper.plan).

Also, `BatchPrefillWithRaggedKVCacheWrapper` stores KV as a flat contiguous tensor across sequences, while `BatchPrefillWithPagedKVCacheWrapper` expects KV split into fixed-size pages with a page table mapping logical to physical locations. 

In our no-cache case (e.g. ViT), KV is simply the current input laid out contiguously — there's no paging structure at all. Forcing it into the paged wrapper would require artificially constructing page tables and reshaping the flat KV into page-sized chunks, which adds unnecessary complexity. The ragged wrapper is the natural fit since it works directly with contiguous KV tensors.

PR #12943, opened 2026-04-11
[None][feat] Add Attention2D sequence parallelism for visual-gen models

venmugil:
<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Added Attention2DAttention, a 2D-mesh context parallelism strategy for visual-gen diffusion models (FLUX, Wan, LTX2), based on https://arxiv.org/pdf/2503.15758
  * New --dit_attn2d_row_size / --dit_attn2d_col_size CLI options; default 1 (disabled)
  * Enhanced log-sum-exp output in the Flash Attention 4 backend to support partial-output combining across the row process group

* **Improvements**
  * Unified Attention2D under cp_size alongside ring attention — both are context parallelism strategies, mutually exclusive, sharing the same cp_size slot in VisualGenMapping
  * Validated mutual exclusivity: Ulysses + Attention2D raises NotImplementedError; Attention2D + ring raises ValueError

* **Tests**
  * Added comprehensive distributed multi-GPU tests for Attention2D functionality and mesh variants
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

<!--
Please write the PR title by following this template:

**[JIRA ticket/NVBugs ID/GitHub issue/None][type] Summary**

Valid ticket formats:
  - JIRA ticket: [TRTLLM-1234] or [FOOBAR-123] for other FOOBAR project
  - NVBugs ID: [https://nvbugs/1234567]
  - GitHub issue: [#1234]
  - No ticket: [None]

Valid types (lowercase): [fix], [feat], [doc], [infra], [chore], etc.

Examples:
  - [TRTLLM-1234][feat] Add new feature
  - [https://nvbugs/1234567][fix] Fix some bugs
  - [#1234][doc] Update documentation
  - [None][chore] Minor clean-up

Alternative (faster) way using CodeRabbit AI:

**[JIRA ticket/NVBugs ID/GitHub issue/None] @coderabbitai title**

NOTE: "@coderabbitai title" will be replaced by the title generated by CodeRabbit AI, that includes the "[type]" and title.
For more info, see /.coderabbit.yaml.

-->

## Description

Adds `Attention2DAttention`, a 2D-mesh sequence parallelism strategy for visual-gen diffusion models (FLUX, Wan, LTX2), based on the approach described in [arxiv.org/pdf/2503.15758](https://arxiv.org/pdf/2503.15758).

Unlike Ulysses, which requires the number of attention heads to be divisible by the parallelism degree, Attention2D distributes the sequence across a `row_size × col_size` process group mesh with no head-divisibility constraint. It also has asymptotically lower communication cost: as the number of processors increases, each GPU handles a smaller sequence shard, so the per-GPU communication volume scales as O(1/P).

**Communication pattern:**
 - Q tokens: all-gathered across the row group
 - K/V tokens: fused into a single all-gather across the col group
 - Partial outputs: combined via flash_attn_combine (LSE-weighted reduction) across the row group

**Design:**
 - Attention2D is unified under cp_size in VisualGenMapping, the same slot used by ring attention — the two are mutually exclusive context parallelism strategies
 - Controlled by dit_attn2d_row_size / dit_attn2d_col_size (default 1, i.e., disabled)
 - Composable with CFG parallelism


## Test Coverage

  - New file: `tests/unittest/_torch/visual_gen/multi_gpu/test_attn2d_attention.py`                                                                                                                                                                                                                                                           
    - Multi-GPU tests covering 2×2, 1×4, and 4×1 mesh configurations with both vanilla and FlashAttention4 inner backends                                                                                                                                                                                                                                                                                         
    - Single-GPU smoke test for `FlashAttn4Attention.forward`                                                                                                                                                                                                                                                                                 
  - Extended: `tests/unittest/_torch/visual_gen/test_visual_gen_args.py`                                                                                                                                                                                                                                                                      
    - ParallelConfig attn2d fields, seq_parallel_size computation for attn2d and Ulysses modes       
  - Extended: `tests/unittest/_torch/visual_gen/multi_gpu/test_visual_gen_mapping.py`
    - attn2d size/rank/group construction tests; constructor validation tests (ulysses+attn2d, cp_size mismatch, ring+attn2d exclusivity); multi-GPU CP collective test verifying attn2d_mesh_rank == cp_rank across all 4 ranks                                                                                                                                                                                                                                                                                       
 

## PR Checklist

## GitHub Bot Help


chang-l:
Hi @NVShreyas , can you please help review this PR if possible?

tensorrt_llm/_torch/visual_gen/modules/attention.py
```diff
@@ -125,9 +125,15 @@ def __init__(
             ]
         )
 
+        # Check which parallelism strategy is active (Attention2D vs Ulysses, mutually exclusive)
```
chang-l: Can we add this to our TODO list? I expect we’ll need to support Ulysses + CP soon.

tests/unittest/_torch/visual_gen/test_visual_gen_args.py
```diff
@@ -161,6 +161,39 @@ def test_validate_world_size_fails(self):
...
+    def test_attn2d_validate_world_size_passes(self):
+        pc = ParallelConfig(dit_cfg_size=2, dit_attn2d_row_size=2, dit_attn2d_col_size=2)
+        pc.validate_world_size(8)
+
+    def test_attn2d_validate_world_size_fails(self):
+        pc = ParallelConfig(dit_cfg_size=2, dit_attn2d_row_size=2, dit_attn2d_col_size=2)
+        with pytest.raises(ValueError, match="exceeds world_size"):
+            pc.validate_world_size(4)
+
+    def test_attn2d_asymmetric_mesh(self):
```
chang-l, line 211: Would attn2D be more suitable when q_seq and kv_seq are highly asymmetric?

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
```diff
@@ -165,3 +175,185 @@ def preferred_layout(self) -> AttentionTensorLayout:
...
+            o_partial.float().contiguous(), lse_partial, out_dtype=out_dtype, return_lse=False
+        )
+
+    @property
+    def preferred_layout(self) -> AttentionTensorLayout:
+        return self._preferred_layout
+
+    @classmethod
+    def support_fused_qkv(cls) -> bool:
+        # FlashAttn4 (the only backend currently supporting the required LSE output)
```
chang-l, line 422: It seems cuDNN also supports LSE output since 9.20 (26.03 pytorch release).. Can you double check?

tensorrt_llm/_torch/visual_gen/models/flux/transformer_flux2.py
```diff
@@ -433,16 +433,34 @@ def __init__(self, model_config: "DiffusionModelConfig"):
...
+        elif use_ulysses:
+            self.use_seq_parallel = True
+            self.seq_parallel_size = ulysses_size
+            self.seq_parallel_pg = vgm.ulysses_group
+            self.seq_parallel_rank = vgm.ulysses_rank
+        else:
+            self.use_seq_parallel = False
+            self.seq_parallel_size = 1
+            self.seq_parallel_pg = None
+            self.seq_parallel_rank = 0
```
chang-l, line 462: Can we abstract or generalize these repeated code changes across models?

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -114,13 +121,26 @@ class ParallelConfig(StrictBaseModel):
...
+
     @property
     def n_workers(self) -> int:
-        return self.dit_cfg_size * self.dit_ulysses_size
+        return self.dit_cfg_size * self.seq_parallel_size
 
     @property
     def total_parallel_size(self) -> int:
-        return self.dit_cfg_size * self.dit_tp_size * self.dit_ring_size * self.dit_ulysses_size
+        return self.dit_cfg_size * self.dit_tp_size * self.dit_ring_size * self.seq_parallel_size
```
chang-l: why we still have dit_ring_size here?

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -114,13 +121,26 @@ class ParallelConfig(StrictBaseModel):
...
+        """Parallelism degree over the sequence/context axis.
+
+        Returns the active parallel degree: Attention2D total mesh size (context
+        parallelism) if enabled, otherwise Ulysses size (head-sharding). Exactly
+        one of these is active at a time; combining them is not yet implemented.
+        """
+        attn2d_size = self.dit_attn2d_row_size * self.dit_attn2d_col_size
+        if attn2d_size > 1:
+            return attn2d_size
+        return self.dit_ulysses_size
```
chang-l, line 142: we should include ring_size in this function

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -93,6 +98,8 @@ class ParallelConfig(StrictBaseModel):
     dit_ulysses_size: int = PydanticField(1, ge=1)  # Supported
     dit_ring_size: int = PydanticField(1, ge=1)  # Not yet supported
     dit_cp_size: int = PydanticField(1, ge=1)
+    dit_attn2d_row_size: int = PydanticField(1, ge=1)  # Supported
+    dit_attn2d_col_size: int = PydanticField(1, ge=1)  # Supported
```
chang-l, line 107: Do we need to define cp size here?

It seems we already define sp = cp * ulysses, where `cp` is implemented as either attn2d or ring..

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -20,20 +20,21 @@
...
 
 
 class VisualGenMapping(DeviceMeshTopologyImpl):
     """Multi-dimensional communicator mesh for visual generation models.
 
     Parallelism Strategy:
     - CFG Parallelism: Distributes positive/negative prompts across GPUs
-    - Ulysses Parallelism: Distributes sequence within each CFG group
+    - CP (Context Parallelism): Ring attention or Attention2D (sequence sharding)
+    - Ulysses Parallelism: Head sharding within each CFG group
```
chang-l, line 33: Can we improve our doc/comment to better clarify the relationship between cp / sp / ring / ulysses? e.g., 
```
cfg:
sp:
  cp:
    ring
    attn2d
  ulysses:
tp:
```
right now these notations/definitions are scattered and hard to follow.

tensorrt_llm/_torch/visual_gen/attention_backend/flash_attn4.py
```diff
@@ -119,26 +142,35 @@ def forward(
...
+        q, k, v, is_causal, origin_dtype = self._prepare_inputs(q, k, v, attention_mask)
+        output, lse = self._fwd(q, k, v, is_causal)
         if output.dtype != origin_dtype:
             output = output.to(origin_dtype)
+        return output, lse
 
-        return output
+    @classmethod
+    def support_lse(cls) -> bool:
+        return True
```
NVShreyas, line 173: could you add this in the interface as well so it can return false by default for other backends?

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
```diff
@@ -165,3 +175,185 @@ def preferred_layout(self) -> AttentionTensorLayout:
...
+        self._preferred_layout = AttentionTensorLayout.NHD
+
+        if _flash_attn_combine is None:
+            raise ImportError(
+                "flash_attn_combine is not available. Attention2DAttention requires "
+                "the Flash Attention JIT kernels to be built. "
+                f"Import error: {_flash_attn_combine_import_error}"
+            ) from _flash_attn_combine_import_error
+
+        if not getattr(inner_backend, "support_lse", lambda: False)():
```
NVShreyas: this can be simplified to self.inner_backend.support_lse after adding to interface

tensorrt_llm/_torch/visual_gen/attention_backend/flash_attn4.py
```diff
@@ -119,26 +142,35 @@ def forward(
...
 
-        output = self._fwd(q, k, v, is_causal)
+    def forward_with_lse(
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        attention_mask: PredefinedAttentionMask = PredefinedAttentionMask.FULL,
+        **kwargs,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
```
NVShreyas, line 155: could we use 1 forward method to return LSE maybe using kwargs? or we will have to force backends to implement this method if they support LSE.

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -103,6 +126,53 @@ def build_mesh(self):
...
+        col_size = self.attn2d_col_size
+        mesh_size = row_size * col_size
+        rank = self._rank
+
+        row_pg: Optional[ProcessGroup] = None
+        col_pg: Optional[ProcessGroup] = None
+
+        for cfg_id in range(self.cfg_size):
+            base = cfg_id * mesh_size
+
```
NVShreyas, line 178: we don't consider TP here? could you add a check so TP isn't enabled when attn2D is enabled.

tensorrt_llm/_torch/visual_gen/attention_backend/flash_attn4.py
venmugil: Good point. Added.

tensorrt_llm/_torch/visual_gen/attention_backend/flash_attn4.py
venmugil: I think it's possible, but the current design was intentional — backends can add LSE support incrementally by separately adding forward_with_lse() and support_lse(), and parallel wrappers that require LSE (e.g., Attention2DAttention) call forward_with_lse explicitly, making the dependency visible and less error-prone.

Also, a backend that doesn't support LSE would silently ignore return_lse=True and return just a Tensor — the caller expecting a tuple would either get a runtime unpack error immediately, or silently propagate a wrong value if it passes the result along without unpacking, making it harder to debug.
Adding an LSE kwarg might also complicate static type-checking: AttentionBackend.forward() (tensorrt_llm/_torch/visual_gen/attention_backend/interface.py:57) currently returns Tensor. Adding a return_lse kwarg would change the return type to Tensor | Tuple[Tensor, Tensor], which type checkers can't narrow based on the runtime value of the argument.

If we are worried about a backend missing to implement forward_with_lse function when it claims support_lse() -> True, should we instead add forward_with_lse to the interface with a default NotImplementedError making it easier to catch the error?

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
venmugil: Latest cuDNN does appear to support LSE output, but Attention2D works as a wrapper over an existing standalone backend, and we don't currently have a cuDNN backend in visual_gen. The available backends are FlashAttn4, TRTLLMAttention, and VanillaAttention.

Additionally, to merge partial attention outputs across ranks, we need a combine kernel. cuDNN doesn't expose a standalone combine API yet. If cuDNN's LSE output format is compatible with Flash Attention's CuTe JIT kernel, we may be able to reuse it for the combine step — but that would need to be verified.

tensorrt_llm/_torch/visual_gen/models/flux/transformer_flux2.py
venmugil: Makes sense. Should I leave it as is in this PR, and address it separately in @NVShreyas 's PR?

tensorrt_llm/_torch/visual_gen/config.py
venmugil: Good point. `dit_cp_size` is unused and probably not required. Removed it.

tensorrt_llm/_torch/visual_gen/config.py
venmugil: Makes sense. Added ring_size.

tensorrt_llm/_torch/visual_gen/mapping.py
venmugil: Updated the documentation.

In the current design `cfg-tp-cp-ulysses`, tp, cp, ulysses are treated as fully orthogonal dimensions.
TP shards weights along head dim, Ulysses shards activations along head dim, while CP shards along sequence dimension.
Currently only tp_size=1 is used, but I believe once tp is supported, TP and Ulysses will need to share the same process group (as done in Megatron-LM), since both operate on the head dimension - CP can remain a truly orthogonal dimension.

tensorrt_llm/_torch/visual_gen/mapping.py
venmugil: Good catch. Added a check above at line 85.

tests/unittest/_torch/visual_gen/test_visual_gen_args.py
venmugil: Yes, attn2d will be much more useful when kv_seq > q_seq. 
Since ring-attn only parallelizes computation along q_seq, ring's applicability diminishes as q_seq gets smaller and attn2d will be more suitable in this case.

tensorrt_llm/_torch/visual_gen/modules/attention.py
venmugil: Updated the comment with todo.

tensorrt_llm/_torch/visual_gen/config.py
venmugil: Removed it, now that dit_ring_size has been added to `def seq_parallel_size`

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
venmugil: Done, thanks.

tensorrt_llm/_torch/visual_gen/attention_backend/flash_attn4.py
NVShreyas: yes makes sense, lets do that

NVShreyas:
could you add a test similar to this for one of the models?
https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/_torch/visual_gen/multi_gpu/test_flux_ulysses.py

tensorrt_llm/_torch/visual_gen/mapping.py
NVShreyas: I don't think TP and Ulysses should share the same process group but thats not related to this PR and we can cover it when we have the PR for TP.

tensorrt_llm/_torch/visual_gen/attention_backend/flash_attn4.py
venmugil: Done.

venmugil:
> could you add a test similar to this for one of the models? https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/_torch/visual_gen/multi_gpu/test_flux_ulysses.py

Good point. Added a new test `tests/unittest/_torch/visual_gen/multi_gpu/test_wan_attn2d.py`

NVShreyas:
could you rebase on main so we can run CI?

venmugil:
> could you rebase on main so we can run CI?

Done.

tensorrt_llm/_torch/visual_gen/models/flux/transformer_flux2.py
NVShreyas: I will add this in my follow up PR

NVShreyas (approved):

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
```diff
@@ -165,3 +175,185 @@ def preferred_layout(self) -> AttentionTensorLayout:
...
+
+
+class Attention2DAttention(AttentionBackend):
+    """
+    Attention2D Context Parallelism wrapper.
+
+    Based on the Attention2D parallelism scheme from:
+        "Attention2D: Communication Efficient Distributed Self-Attention Mechanism"
+        https://arxiv.org/pdf/2503.15758
+
```
chang-l, line 187: Can we improve on this docstring field for `Attention2DAttention`, including purpose, mesh layout, supported inner backends, communication pattern (Q row-AG, KV col-AG, LSE combine), etc?

chang-l (commented):
Can we also update `examples/visual_gen/README.md`: add --attn2d_row_size/--attn2d_col_size to the Common Arguments table; add an "Attention2D (2D context parallelism)" subsection alongside Ulysses in Multi-GPU Parallelism section; rewrite the "ulysses_size must divide head count" troubleshooting line to point to Attention2D as the alternative?

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
venmugil: I've updated this docstring, and also the readme file `examples/visual_gen/README.md`. Thanks.

chang-l (approved):
Thanks for the contribution!

PR #12956, opened 2026-04-12
[None][fix] Fall back to local cache when loading tokenizer for gated models

ssam18:
When serving a gated model (e.g. Llama 4) in an environment without an HF token, `load_hf_tokenizer` was failing and returning `None` even though the files were already downloaded locally. This adds a retry with `local_files_only=True` so the cached tokenizer gets picked up instead of bailing out. Fixes #12805.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced tokenizer loading with automatic fallback mechanism for improved reliability.
  * Improved error handling and logging for better diagnostics during tokenizer loading failures.


1MrazorT1:
Hi, I want to flag a process concern here. I opened issue #12805, identified the root cause, and explicitly stated in the issue that I had a fix ready and was waiting for maintainer validation before submitting a PR, as required by the contributing guidelines:

> "The issue request must be reviewed by TensorRT-LLM engineers and approved prior to code review."

This PR was opened before that validation occurred, which bypasses the required process. I had independently arrived at the same fix. I'd appreciate if the maintainers could weigh in on how to handle this.

ssam18:
Fair point. Thanks

PR #13015, opened 2026-04-13
[feat][None] Fused moe all-reduce routed scaling factor + quant support

murphymatt:
## Release Notes

* **New Features**
  * Added FP4 quantization output support for MoE all-reduce fusion operations
  * Introduced routed scale factor parameter to control expert reduction scaling in MoE operations

* **Enhancements**
  * Extended API to support optional quantization outputs with improved validation
  * Enhanced MoE configuration flexibility with new optional parameters

## Description

**PORT FROM FLASHINFER https://github.com/flashinfer-ai/flashinfer/pull/2966**

In some cases like kimi k2.5, glm, we require an extra scale factor applied over all routed experts. While we can add a constant multiplier over the experted scaling factors, it is more efficient to fuse this multiplication inside the kernel.

e.g. https://huggingface.co/moonshotai/Kimi-K2.5/blob/main/config.json#L143

Furthermore, we want to also support quantized outputs for this operator.

Note, since all relevant MoE paths already apply routed scaling factor within the moe kernel our router impl themselves, we actually avoid using this input parameter in fused paths within TRT-LLM. The purpose of this enablement is to maintain API parity between TRT-LLM and Flashinfer.

## Test Coverage

```
pytest tests/unittest/_torch/modules/moe/test_moe_backend.py
pytest tests/unittest/_torch/multi_gpu/test_allreduce.py
```

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/thop/allreduceOp.cpp
+++ b/cpp/tensorrt_llm/thop/allreduceOp.cpp
@@ -1473,7 +1473,7 @@ private:
 #endif // ENABLE_MULTI_DEVICE
 
 void preallocateNCCLWindowBuffer(
-    torch::Tensor const& input, torch::List<int64_t> const& group, const int64_t buffersPerSize)
+    torch::Tensor const& input, torch::List<int64_t> const& group, int64_t const buffersPerSize)
 {
 #if ENABLE_MULTI_DEVICE
     if (buffersPerSize <= 0 || group.size() == 0 || input.numel() == 0 || input.size(0) == 0)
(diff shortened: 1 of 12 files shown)


cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -1877,10 +1935,13 @@ TORCH_LIBRARY_FRAGMENT(trtllm, m)
         "moe_finalize_allreduce("
         "Tensor input,"
         "Tensor? residual,"
+        "Tensor? quant_out,"
```
liji-nv: Because you are not really using the quant_out and scale_out in model, I suggest that do not add them as mutated input but generate the output inside the custom op. Especially you are returning the mutated output again and the input is optional. You cannot even mark which output is the mutated input. 

Torch compile path cannot correctly handle such usage.

liji-nv (changes requested):

cpp/tensorrt_llm/thop/allreduceOp.cpp
murphymatt: Thanks for the feedback. Wanted to keep the API consistent with what we landed in flashinfer. Updated here, and can submit a similar PR to update there as well.

murphymatt:
@liji-nv could you take another look?

PR #13034, opened 2026-04-14
[TRTLLM-11872][perf] Multi-threading async media loading and optimizing video frame decoding in trtllm-serve

yechank-nvidia:
## Summary

  Two performance fixes in `tensorrt_llm/inputs/utils.py` for `trtllm-serve` multimodal API requests.

  ### Fix 1: Truly Async Media Loading

  - `async_load_image` / `async_load_video` / `async_load_audio` were blocking the event loop despite being `async`. CPU-bound work (PIL,cv2, soundfile) now runs in a thread pool via `run_in_executor`. Additional improvements:

  - Global `aiohttp.ClientSession` singleton — reuses TCP connections instead of creating a new session per request
  - `retrieve_all_async` now launches a single `asyncio.gather` across all modalities simultaneously instead of gathering per-modality sequentially

  ### Fix 2: Faster Video Frame Decoding

  **Before:** `CAP_PROP_POS_FRAMES` seek per frame — each seek decodes from the nearest keyframe to the target frame. Gets worse as GOP size grows (real H.264 videos: 1–8s keyframe intervals).

  **After:**
  - Single sequential forward scan with `grab()` — no per-frame seek
  - PIL `Image.fromarray` + `ToTensor` replaced with direct `numpy.transpose` → `torch.from_numpy` (7.3× faster for tensor conversion)

  ## Results

  Baseline: original per-frame seek + PIL conversion, 32 frames extracted, H.264.

  **Single request latency**

  | Scenario | Before | After | Speedup |
  |---|---|---|---|
  | 360p, 5s | 501 ms | 79 ms | **6.3×** |
  | 720p, 5s | 628 ms | 116 ms | **5.4×** |
  | 1080p, 5s | 843 ms | 222 ms | **3.8×** |
  | 1080p, 60s | 878 ms | 560 ms | 1.6× |

  Short videos benefit most — seek overhead dominates decode time for short clips and is eliminated entirely by the sequential scan.

  ## Files Changed

  - `tensorrt_llm/inputs/utils.py`
  - `tests/unittest/inputs/test_async_media_loading.py` (13 new tests)

## Summary by CodeRabbit

* **New Features**
  * Implemented efficient HTTP session reuse for media downloads.
  * Added concurrent gathering of multimodal data for improved performance.

* **Bug Fixes**
  * Optimized video frame extraction logic with improved frame sampling and processing.

* **Tests**
  * Added comprehensive test coverage for async media loading, session management, and concurrent data retrieval operations.


tensorrt_llm/inputs/utils.py
```diff
@@ -159,19 +172,24 @@ async def async_load_image(
...
     if parsed_url.scheme in ["http", "https"]:
-        async with aiohttp.ClientSession() as session:
-            async with session.get(image) as response:
-                content = await response.read()
-                image = _load_and_convert_image(BytesIO(content))
+        session = await _get_aiohttp_session()
+        async with session.get(image) as response:
+            content = await response.read()
+        # Offload CPU-bound PIL decoding to thread pool to avoid blocking event loop
+        image = await loop.run_in_executor(None, _load_and_convert_image,
```
2ez4bz: I might be missing something, but is there some advantage here to using the `loop.run_in_executor`? Since we're using the default executor anyway, could we instead just do `asyncio.to_thread(_load_and_convert_image, BytesIO(content))`?

tensorrt_llm/inputs/utils.py
```diff
@@ -216,38 +228,48 @@ def _load_video_by_cv2(video: str,
     num_frames_to_sample = max(1, num_frames_to_sample)  # at least one sample
 
     if num_frames_to_sample == frame_count:
```
2ez4bz: Nit: do we even need the `if / elif` anymore? I think when these 2 values match, what's in the `else` block should return the same as `list(range(...))`. Slightly less code to read if we can combine the 2 branches.

tensorrt_llm/inputs/utils.py
```diff
@@ -298,27 +319,27 @@ async def async_load_video(video: str,
     assert format in ["pt", "pil"], "format must be either Pytorch or PIL"
 
     parsed_url = urlparse(video)
+    loop = asyncio.get_running_loop()
```
2ez4bz: Similar comment here re using the event loop vs just `to_thread`.

tensorrt_llm/inputs/utils.py
```diff
@@ -480,19 +505,39 @@ def __init__(
...
     ) -> tuple[Optional[Dict[str, List[Any]]], Optional[Dict[str, List[Any]]]]:
-        """Retrieve all collected multimodal data and embeddings."""
+        """Retrieve all collected multimodal data and embeddings.
+
+        All coroutines across all modalities (and across _data/_embeddings) are
+        gathered concurrently in a single asyncio.gather call, so e.g. image
+        downloads and video downloads overlap instead of running sequentially.
+        """
 
         async def _retrieve(
```
2ez4bz, line 500: Would the following work?

```python
async def _retrieve(
        data: Optional[dict[str, list]]) -> Optional[Dict[str, List[Any]]]:
    if not data:
        return None
    pairs = [(modality, item)
             for modality, items in data.items() if items
             for item in items]
    if not pairs:
        return None
    modality_keys, coroutines = zip(*pairs)
    results = await asyncio.gather(*coroutines)
    out: dict[str, list] = defaultdict(list)
    for modality, result in zip(modality_keys, results):
        out[modality].append(result)
    return dict(out)
```

tests/unittest/inputs/test_async_media_loading.py
```diff
@@ -0,0 +1,276 @@
...
+    @pytest.mark.asyncio
+    async def test_load_image_from_data_url_pil(self):
+        url = _make_image_data_url()
+        result = await async_load_image(url, format="pil")
+        assert isinstance(result, Image.Image)
+        assert result.mode == "RGB"
+
+    @pytest.mark.asyncio
+    async def test_load_image_from_data_url_pt(self):
+        import torch
```
2ez4bz: Nit: please move this to be a module-level import.

tests/unittest/inputs/test_async_media_loading.py
```diff
@@ -0,0 +1,276 @@
...
+
+class TestAsyncLoadAudio:
+    @pytest.mark.asyncio
+    async def test_load_audio_from_file(self):
+        with tempfile.NamedTemporaryFile(suffix=".wav", delete=False) as f:
+            path = _make_audio_file(f.name)
+        result = await async_load_audio(path)
+        audio_array, sample_rate = result
+        assert isinstance(audio_array, np.ndarray)
+        assert sample_rate == 16000
```
2ez4bz: Nit: can you leave a comment for why this is the expected value?

tests/unittest/inputs/test_async_media_loading.py
```diff
@@ -0,0 +1,276 @@
...
+    @pytest.mark.asyncio
+    async def test_new_session_created_after_close(self):
+        utils_module._global_aiohttp_session = None
+        try:
+            s1 = await _get_aiohttp_session()
+            await s1.close()
+            s2 = await _get_aiohttp_session()
+            assert s1 is not s2, "Expected a fresh session after the old one was closed"
+            assert not s2.closed
+        finally:
```
2ez4bz: Nit: maybe this block of code can be made a fixture for this class?

```python
@pytest.fixture(autouse=True)
def close_global_aiohttp_session(self):
    yield
    if utils_module._global_aiohttp_session is not None:
        await utils_module._global_aiohttp_session.close()
        utils_module._global_aiohttp_session = None
```

The block after yield serves the same purpose as the `finally` in both test functions, and also executes if exceptions are raised.

tests/unittest/inputs/test_async_media_loading.py
```diff
@@ -0,0 +1,276 @@
...
+
+            target[modality].append(_coro())
+
+    @pytest.mark.asyncio
+    async def test_single_modality_returns_correct_data(self):
+        tracker = self._make_tracker()
+        self._inject(tracker, "image", [1, 2, 3])
+
+        data, embeddings = await tracker.retrieve_all_async()
+        assert data == {"image": [1, 2, 3]}
```
2ez4bz: Nit: this could be assigned to a variable so that lines 213 and 216 use the same variable?

Similar comment for the test functions below.

tests/unittest/inputs/test_async_media_loading.py
```diff
@@ -0,0 +1,276 @@
...
+        tracker._data["image"].extend([_slow("img1"), _slow("img2")])
+        tracker._data["video"].extend([_slow("vid1"), _slow("vid2")])
+
+        start = time.perf_counter()
+        data, _ = await tracker.retrieve_all_async()
+        elapsed = time.perf_counter() - start
+
+        assert data["image"] == ["img1", "img2"]
+        assert data["video"] == ["vid1", "vid2"]
+        assert elapsed < DELAY + TOLERANCE, (
```
2ez4bz: I'm a bit hesitant to have such timing tests in CI - from past experience, even the simplest tests can turn out flaky under load. I guess only one way to find out...

2ez4bz (approved):
Approving to unblock, but please consider the comments 🙏

tensorrt_llm/inputs/utils.py
yechank-nvidia: You are correct. I referenced from vllm where they set different # of threads per feature, which is not the case for TRT-LLM.

Will remove it and switch to `to_thread()`

tensorrt_llm/inputs/utils.py
yechank-nvidia: You are right. Those two are doing the same thing. It was leftover from legacy code. Will remove it.

tests/unittest/inputs/test_async_media_loading.py
yechank-nvidia: Agreed. Removed. Thx.

tests/unittest/inputs/test_async_media_loading.py
yechank-nvidia: Done. Thanks.

tests/unittest/inputs/test_async_media_loading.py
yechank-nvidia: Done. Thanks.

tests/unittest/inputs/test_async_media_loading.py
yechank-nvidia: Added. Thx.

tests/unittest/inputs/test_async_media_loading.py
yechank-nvidia: Removed. Thx.

tensorrt_llm/inputs/utils.py
yechank-nvidia: Changed. Thx.

tensorrt_llm/inputs/utils.py
yechank-nvidia: Changed. Thx.

tensorrt_llm/inputs/utils.py
yechank-nvidia: Removed. Thx.

2ez4bz:
first try

PR #12898, opened 2026-04-09
feat: Add MX checkpoint loader for ModelExpress P2P RDMA weight transfer

KavinKrishnan:
## Summary

Add `@register_checkpoint_loader("MX")` for GPU-to-GPU weight transfer via [ModelExpress](https://github.com/ai-dynamo/modelexpress). Uses TRT-LLM's existing `BaseCheckpointLoader` architecture — no changes to `model_loader.py`, `llm_args.py`, or `linear.py`.

### Architecture

Following the two-axis model:
- **Weight source axis** (`checkpoint_format`): `MX` registered via `@register_checkpoint_loader("MX")`
- This is orthogonal to memory management (GMS), enabling clean composition of all modes

### How auto-detect works

`MxWeightLoader.load_weights()` probes the MX server:
1. **Sources found** → target mode: delegates to `MxLiveWeightLoader` for RDMA receive
2. **No sources** → source mode: falls back to `HfWeightLoader` for disk load, then `publish_from_worker()` publishes weights to MX

### New files

```
tensorrt_llm/_torch/models/checkpoints/mx/
├── __init__.py
├── checkpoint_loader.py    # @register_checkpoint_loader("MX") — MxCheckpointLoader
├── config_loader.py        # Delegates to HfConfigLoader
└── weight_loader.py        # MxWeightLoader — auto-detect + RDMA/disk fallback
```

### Modified files

- `checkpoints/__init__.py`: Import `MxCheckpointLoader`
- `executor/worker.py`: `publish_from_worker()` hook — auto-publishes when `MODEL_EXPRESS_URL` is set and `MODEL_EXPRESS_TARGET` is not

### Usage

```python
llm = LLM(model="model-name", checkpoint_format="MX")
# Or via Dynamo: python3 -m dynamo.trtllm --model-express-url mx-server:8001
```

### Validated

Kimi K2.5 (TP=8, MoE, nvfp4) on GCP GB200: 90.75 GB/rank at 365-509 Gbps, end-to-end disaggregated inference verified.

### Dependencies

- [ModelExpress](https://github.com/ai-dynamo/modelexpress) (merged to main): `MxLiveWeightLoader`, `publish_from_worker`
- [Dynamo PR #8037](https://github.com/ai-dynamo/dynamo/pull/8037): `--model-express-url` engine integration

PR #13045, opened 2026-04-14
[TRTLLM-11851][feat] MX and GMS integration MVP for Dynamo weight sharing

chienchunhung:
## Summary by CodeRabbit

* **New Features**
  * GMS (GPU Memory Service) backend for disaggregated GPU weight memory
  * MX P2P weight-transfer support with automatic fallback and presharded-weight handling for Linear modules
  * New LoadFormat.GMS and user-facing config options: mx_server_url, mx_preshard_strategy, gms_socket_path, gms_mode, gms_tag

* **Tests**
  * Added comprehensive unit tests for GMS backend, MX checkpoint loader, and MX/GMS argument validation

* **Chores**
  * Packaging metadata updated (SPDX year) and clarifying install guidance added to setup notes

# [TRTLLM-11851][feat] MX and GMS integration prototype for Dynamo weight sharing

### Summary

Prototype implementation of the **two-axis integration model** for MX (ModelExpress P2P transfer) and GMS (GPU Memory Service) in TRT-LLM's PyTorch backend. Enables fast cold-start via GPU-to-GPU weight streaming (MX) and zero-copy weight sharing with crash-resilient failover (GMS).

MX and GMS map onto two **independent axes** in TRT-LLM's existing loading pipeline:

- **`checkpoint_format`** (weight source axis): `"MX"` — P2P via upstream `MxLiveWeightLoader.load_weights()`, with automatic HF disk fallback (inherited from `HfCheckpointLoader`)
- **`LoadFormat`** (memory management axis): `GMS` — out-of-process GPU memory pool with RW/RO dual paths via upstream `gpu_memory_service` class-based API

These compose independently, giving four modes with no combinatorial config explosion:

| Mode | `checkpoint_format` | `LoadFormat` | Use case |
|---|---|---|---|
| Pure TRT-LLM | `"HF"` (default) | `AUTO` (default) | Current behavior, unchanged |
| MX only | `"MX"` | `AUTO` | Cross-node P2P weight transfer |
| GMS only | `"HF"` | `GMS` | Within-node weight sharing + crash resilience |
| MX + GMS | `"MX"` | `GMS` | Cross-node P2P + within-node sharing |

Aligned against the actually-merged upstream APIs: `gpu-memory-service` 0.9.0 (from `ai-dynamo/dynamo`) and `modelexpress` 0.3.0 (from `ai-dynamo/modelexpress`). TRT-LLM owns the integration policy (Layer 3) and calls the upstream libraries' stable per-call primitives (Layer 2); we never duplicate the wire-protocol or runtime mechanics (Layer 1 — gRPC, NIXL RDMA, CUDA VMM).

#### What's included

**MXCheckpointLoader** (`_torch/models/checkpoints/mx/checkpoint_loader.py`, ~380 lines)

- Subclasses `HfCheckpointLoader` — HF disk fallback is inherited, no separate fallback loader needed
- Delegates the actual NIXL/RDMA transfer to upstream `MxLiveWeightLoader(mx_server=url).load_weights(checkpoint_dir, mapping=, model=)` (handles agent setup, source matching, dtype-cast handling, PVC fallback for size-mismatched tensors)
- Delegates the publish side to upstream `publish_model_params(torch_model)` via an env-var dance over `MODEL_EXPRESS_URL` and `MODEL_NAME`
- Exposes `p2p_succeeded` property — consumed by `ModelLoader` to set `_weights_presharded` on `Linear` modules
- Pre-`post_load_weights()` MX source publish timing so targets receive raw loaded state and run their own transforms
- `model_name` constructor parameter (plumbed from `llm_args.model`) so publish identity comes from the user-supplied Hub ID with HF-snapshot path unmangling, not the upstream `"unknown"` sentinel
- Conservative mixed-success: if MX returns size-mismatched fallback weights, falls through to a full disk load to avoid mixing presharded and non-presharded weights
- Defensive `MX_SOURCE_QUERY_TIMEOUT=30` setdefault when an MX URL is configured, capping upstream `MxLiveWeightLoader._query_source` polling at 30 s instead of the 1-hour default. The correct upstream fix is to adopt `RdmaStrategy`'s immediate-fallback pattern (single `list_sources` call, no polling) — tracked as [MX-4](https://github.com/chienchunhung/TensorRT-LLM/blob/docs-and-plans/docs/design/mx-gms-integration/15-prototype-validation-plan.md#mx-4-adopt-rdmastrategys-immediate-fallback-source-discovery-in-mxliveweightloader)

**GMSBackend** (`_torch/memory/gpu_memory_backend.py`, ~440 lines)

- `GPUMemoryBackend` protocol + concrete `GMSBackend` implementation wrapping the upstream `GMSClientMemoryManager` and the `gms_use_mem_pool` context manager
- Mode resolved at `connect()` time via `get_or_create_gms_client_memory_manager(socket, device, mode=RequestedLockType.RW_OR_RO)` and inspected via `granted_lock_type`
- **RW path:** caller uses `gms_backend.mem_pool_scope(device)` context manager; after load, `move_untracked_params()` migrates stray buffers into the pool (inside the scope, matching upstream `_load_rw` order), then `finalize_write()` delegates to upstream `finalize_gms_write()` (register + sync + commit + RO reconnect + remap)
- **RO path:** `post_load_weights()` runs *before* `materialize_module()`, guard prevents double-execution; `materialize_module_from_gms(mgr, model, device_index=N)` does the zero-copy import
- Default tag is `"weights"` matching the GMS library convention
- Default socket path resolves via `get_socket_path(device, tag)` (UUID-based, stable across `CUDA_VISIBLE_DEVICES`)
- VMM safety: `connect()` applies `patch_empty_cache()` to prevent segfaults on VMM-backed allocations

**ModelLoader changes** (`_torch/pyexecutor/model_loader.py`, +255/-19)

- New `LoadFormat.GMS` branch with full RW/RO dual paths
- GMS RW: respects `model.llm_checkpoint_dir` redirect (matches AUTO branch); loads speculative-decoding draft weights inside the same `mem_pool_scope` so a single `finalize_write(model)` covers both main and draft via the model-tree walk
- MX source publish fires for `LoadFormat.AUTO` and GMS-RW modes, before `post_load_weights()`
- `_weights_presharded` flag set per-path (MX P2P in AUTO, GMS materialize in RO)
- Draft model modules excluded from presharded flag

**Linear module** (`_torch/modules/linear.py`)

- `_weights_presharded = False` declared in `__init__` — TP slicing skipped when True
- `_effective_tp_coords(module)` helper centralized TP-coordinate logic: when `_weights_presharded` is True, returns `(tp_size=1, tp_rank=0)` for *all* `load_weight_shard(...)` calls — both the main weight/bias tensors AND their quantization metadata (`weight_scale`, `input_scale`, `pre_quant_scale`, etc.) across all quant-scheme loaders (FP8, NVFP4, MXFP4, W4A16-AWQ, W4A8-AWQ, etc.)

**Configuration** (`llmapi/llm_args.py`)

- `mx_server_url`, `mx_preshard_strategy`, `gms_socket_path`, `gms_mode`, `gms_tag` — all `status="prototype"`
- Validators: `validate_mx_config()`, `validate_gms_config()` — honor upstream `MODEL_EXPRESS_URL` env var as fallback for `mx_server_url`; reject empty/whitespace `gms_tag` values
- `mx_preshard_strategy: per_module | global` (default `per_module`); `global` raises `NotImplementedError` until `LoadFormat.PRESHARDED` lands upstream
- API stability YAML updated

**Optional dependencies** (`setup.py`)

- The MX (`modelexpress`) and GMS (`gpu-memory-service`) Python packages are intentionally **NOT** bundled as `[mx]` / `[gms]` / `[dynamo]` extras while the integration is at prototype status. Users install manually:

  ```bash
  # MX — Apache-2.0, on PyPI
  pip install "modelexpress>=0.3.0,<0.4.0"

  # GMS — Apache-2.0, source-only (not yet on PyPI)
  git clone https://github.com/ai-dynamo/dynamo
  pip install ./dynamo/lib/gpu_memory_service
  ```

  Once installed, the MX/GMS code paths auto-detect them via `import`; without them, the loaders fall back to plain HF disk loading.

- **Why no extras:** (1) `gpu-memory-service` is not published to PyPI (GMS-6), so NVIDIA's Blossom-CI vulnerability scanner can't resolve it; (2) `modelexpress` is brand-new on PyPI and not yet in the internal OSS allowlist (MX-7). Both blockers are tracked in [§15 of the design doc](https://github.com/chienchunhung/TensorRT-LLM/blob/docs-and-plans/docs/design/mx-gms-integration/15-prototype-validation-plan.md#upstream-alignment-requests). Restoring `pip install tensorrt_llm[dynamo]` is a single-hunk revert once both are cleared.

### Running with MX or GMS enabled

The MX/GMS knobs are set via a YAML config passed to `trtllm-serve --config`, or as keyword arguments when using the Python `LLM` API directly.

| Field | Type | Purpose |
|---|---|---|
| `checkpoint_format` | str | `"MX"` to enable MX P2P loading. Default `None` → `"HF"`. |
| `mx_server_url` | str | URL of the ModelExpress server. Falls back to `MODEL_EXPRESS_URL` env var. |
| `load_format` | str / `LoadFormat` | `"GMS"` to enable GMS-backed loading. Default `"AUTO"`. |
| `gms_socket_path` | str | Unix socket of the GMS daemon. Default `None` → UUID-keyed auto-resolve. |
| `gms_mode` | str | `auto` (default), `rw`, `ro`. |
| `gms_tag` | str | GMS pool tag. Default `"weights"`. |

**MX only:**
```yaml
# config_mx.yaml
checkpoint_format: "MX"
mx_server_url: "http://mx-server:8001"
```
```bash
trtllm-serve <model> --config config_mx.yaml
```

**GMS only:**
```yaml
# config_gms.yaml
load_format: "GMS"
```
```bash
# First worker: gets RW, loads from disk, commits to GMS pool
trtllm-serve <model> --config config_gms.yaml
# Subsequent workers: get RO, zero-copy import
trtllm-serve <model> --config config_gms.yaml
```

**Python API:**
```python
from tensorrt_llm import LLM
llm = LLM(model="<model>", checkpoint_format="MX", mx_server_url="http://mx:8001")
# or
from tensorrt_llm.llmapi import LoadFormat
llm = LLM(model="<model>", load_format=LoadFormat.GMS)
```

### Design decisions

- **No monkey-patching.** The upstream `setup_gms()` entry point patches `ModelLoader.load` from outside. We use `GMSBackend` as the explicit, reviewable boundary instead.
- **MX P2P is NOT used in GMS RW mode.** Model params are meta tensors at that point — no CUDA buffers for P2P to write into. Disk loading under the GMS pool is the correct behavior.
- **Per-module `_weights_presharded` instead of `LoadFormat.PRESHARDED`.** Composes more cleanly with `LoadFormat.GMS` and lets us mark only MX-delivered modules in mixed-success scenarios.
- **Quantization-metadata-aware TP bypass.** `_effective_tp_coords(module)` ensures `_weights_presharded` skips TP slicing for weights AND their quantization metadata (`weight_scale`, `input_scale`, `pre_quant_scale`) — preventing shape mismatches on quantized models.

### Relationship to existing work

- **PR [#12898](https://github.com/NVIDIA/TensorRT-LLM/pull/12898)** (MX team prototype, closed): we adopt `_weights_presharded` concept and pre-`post_load_weights()` publish timing, using two-axis model instead of `LoadFormat.PRESHARDED`.
- **dynamo PR [#7575](https://github.com/ai-dynamo/dynamo/pull/7575)** (TRT-LLM sleep/wake with GMS): we use the same Layer 2 primitives but invoke them directly, not via monkey-patch.
- **MX team's downstream PR [chienchunhung/TensorRT-LLM #1](https://github.com/chienchunhung/TensorRT-LLM/pull/1)**: drove the `MODEL_EXPRESS_URL` env-var fallback, the `MX_SOURCE_QUERY_TIMEOUT` defensive default (MX-4), and the `model_name` plumbing.
- **Design doc:** [`docs/design/mx-gms-integration/`](https://github.com/chienchunhung/TensorRT-LLM/tree/docs-and-plans/docs/design/mx-gms-integration) on the `docs-and-plans` branch. [§15](https://github.com/chienchunhung/TensorRT-LLM/blob/docs-and-plans/docs/design/mx-gms-integration/15-prototype-validation-plan.md) has the validation plan, upstream alignment requests, and test matrix.

### Upstream alignment requests

Workarounds in the prototype that could be removed if the corresponding upstream change lands. Full details in [§15](https://github.com/chienchunhung/TensorRT-LLM/blob/docs-and-plans/docs/design/mx-gms-integration/15-prototype-validation-plan.md#upstream-alignment-requests).

| ID | To | Title | Workaround in this PR | Blocks merge? |
|---|---|---|---|---|
| **MX-1** | MX | `LoadFormat.PRESHARDED` vs per-module flag | `mx_preshard_strategy='global'` raises until upstream lands | No |
| **MX-2** | MX | Promote `_build_trtllm_identity` to public API | env-var dance in `publish_as_source` | No |
| ~~MX-3~~ | MX | ~~Per-rank addressing~~ | **Resolved** — upstream uses global MPI rank on both sides | — |
| **MX-4** | MX | Adopt `RdmaStrategy` immediate-fallback in `MxLiveWeightLoader` | `MX_SOURCE_QUERY_TIMEOUT=30` (30 s poll vs correct try-once) | No |
| **MX-5** | MX | NIXL ownership for MX+GMS composition | MX P2P bypassed in GMS-RW path | **Yes for MX+GMS** |
| **MX-7** | MX/TRT-LLM | Onboard `modelexpress` into NVIDIA OSS allowlist | extras removed from `setup.py` | No |
| **GMS-1** | GMS | Non-monkey-patch integration path | `GMSBackend` owns the integration | No |
| **GMS-2** | GMS | Promote `_move_untracked_params` to public | Re-implemented in `GMSBackend` | No |
| **GMS-6** | GMS | Publish `gpu-memory-service` to PyPI | extras removed from `setup.py`; install from source | No |

**MX-1 and MX-5** are the design items most likely to come up in review. **MX-7 and GMS-6** block restoring `pip install tensorrt_llm[dynamo]` ergonomics.

## Test Coverage

148 CPU-only unit tests in `tests/unittest/`, runs in <1 s, no MX/GMS deps required:

- `tests/unittest/llmapi/test_mx_gms_args.py` — config validation, defaults, cross-field warnings, `MODEL_EXPRESS_URL` env-var fallback, `gms_tag` empty-string rejection, `LoadFormat` int-conversion regression
- `tests/unittest/_torch/models/checkpoints/mx/test_mx_checkpoint_loader.py` — construction + registry, parametrized fallback paths, MX-success integration contract, `MX_SOURCE_QUERY_TIMEOUT` defensive default, `model_name` plumbing, `_normalize_model_identity` (HF-snapshot unmangling), `_resolve_mx_model_name` priority order, `publish_as_source` `MODEL_NAME` end-to-end with env restoration
- `tests/unittest/_torch/memory/test_gms_backend.py` — construction, pre-connect state, `connect()` graceful failure, RW-only method gating, `_ptr_in_gms` half-open intervals + zero-sentinel guards, `_storage_nbytes`, protocol conformance. **All tests run on CPU CI** (CUDA dependency monkeypatched via autouse fixture).

Tests do NOT exercise the upstream libraries — fallback paths use `sys.modules`-injection to force `ImportError`, success paths use fake module trees.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/memory/gpu_memory_backend.py
+++ b/tensorrt_llm/_torch/memory/gpu_memory_backend.py
@@ -1,0 +1,439 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""GPU Memory Backend protocol and GMS implementation.
+
+Defines a thin protocol that adapts the upstream GPU Memory Service
+(``gpu_memory_service`` from ``ai-dynamo/dynamo``) into TRT-LLM's
+``ModelLoader`` pipeline. The protocol intentionally exposes only the
+operations TRT-LLM needs at specific points in the loading lifecycle —
+all heavy lifting (CUDA VMM, FD passing, zero-copy tensor construction)
+is delegated to the GMS library's stable public Python primitives.
+
+Design notes:
+- We deliberately avoid the upstream ``setup_gms()`` monkey-patch entry
+  point. TRT-LLM owns the integration policy in ``ModelLoader``; this
+  backend is a thin adapter over the GMS library's primitive API.
+- The protocol has a single point of contact per concern, so when the
+  upstream GMS Python API drifts, only the methods on ``GMSBackend``
+  need to change — the call sites in ``model_loader.py`` are stable.
+
+Operating modes:
+- **RW (Read-Write)**: First worker loads weights via the normal
+  checkpoint loader pipeline, with allocations captured by a GMS-managed
+  CUDA memory pool. After loading, weights are committed for read-only
+  access by other workers and the client transitions to RO mode in place.
+- **RO (Read-Only)**: Subsequent workers zero-copy import already-committed
+  weights from the GMS pool. ``post_load_weights()`` must run BEFORE
+  materialization so that module aliases are set up correctly.
+"""
+
+from contextlib import contextmanager
+from typing import Iterator, Optional, Protocol, runtime_checkable
+
+import torch
+from torch import nn
+
+from tensorrt_llm.logger import logger
+from tensorrt_llm.mapping import Mapping
+
+
+@runtime_checkable
+class GPUMemoryBackend(Protocol):
+    """Thin abstraction over GPU memory services for API stability."""
+
+    def connect(self) -> bool:
+        """Establish a session with the backend. Returns True on success."""
+        ...
+
+    @property
+    def is_rw(self) -> Optional[bool]:
+        """Whether this backend was granted RW (vs RO). None pre-connect."""
+        ...
+
+    def has_committed_weights(self) -> bool:
+        """Whether committed weights are ready to be imported (RO-ready)."""
+        ...
+
+    def mem_pool_scope(
+        self,
+        device: Optional[torch.device] = None,
+    ) -> "Iterator[None]":
+        """Return a context manager scoping CUDA allocations to the backend pool."""
+        ...
+
+    def materialize_module(self, model: nn.Module) -> None:
+        """Zero-copy import committed weights into model params (RO path)."""
+        ...
+
+    def finalize_write(self, model: nn.Module) -> int:
+        """Register, commit, and transition to RO. Returns bytes committed."""
+        ...
+
+    def move_untracked_params(self, model: nn.Module) -> None:
+        """Move stray params not in the backend pool into shared memory."""
+        ...
+
+    def cleanup(self) -> None:
+        """Release all resources and disconnect."""
+        ...
+
+
+# String mode -> upstream RequestedLockType. Resolved lazily inside connect()
+# so importing this module does not require the GMS library to be installed.
+_MODE_ALIASES = ("rw", "ro", "auto")
+
+
+class GMSBackend:
+    """Concrete ``GPUMemoryBackend`` using ``gpu_memory_service`` (GMS).
+
+    GMS is a multi-process GPU memory manager (out-of-process server,
+    in-process clients) that lets multiple inference instances share weight
+    bytes via CUDA VMM mappings and Unix-socket FD passing.
+
+    This adapter calls the GMS library's stable per-call primitives
+    (``GMSClientMemoryManager`` + the helpers in
+    ``gpu_memory_service.client.torch.allocator`` and ``.module``) and
+    composes them into the methods TRT-LLM's ``ModelLoader`` invokes for
+    the ``LoadFormat.GMS`` branch. We intentionally do **not** use the
+    upstream ``setup_gms()`` monkey-patch — TRT-LLM owns the integration
+    policy and we keep the dependency surface narrow and explicit.
+    """
+
+    DEFAULT_TAG = "weights"
+
+    def __init__(
+        self,
+        socket_path: Optional[str],
+        mapping: Mapping,
+        mode: str = "auto",
+        tag: str = DEFAULT_TAG,
+    ) -> None:
+        """Initialize the GMS backend.
+
+        Args:
+            socket_path: Unix domain socket path for the per-GPU GMS daemon.
+                When ``None``, the default per-GPU UUID-keyed path from
+                ``gpu_memory_service.common.utils.get_socket_path`` is used
+                (resolved lazily inside :meth:`connect`).
+            mapping: TRT-LLM distributed ``Mapping`` for TP/PP rank info.
+            mode: Operating mode — ``"auto"`` (RW first, RO after committed),
+                ``"rw"`` (require RW), or ``"ro"`` (require RO).
+            tag: Logical tag identifying this weight set in the GMS server.
+                Default ``"weights"`` matches the GMS library convention.
+        """
+        if mode not in _MODE_ALIASES:
+            raise ValueError(f"GMS mode must be one of {_MODE_ALIASES}, got {mode!r}")
+
+        self._socket_path = socket_path
+        self._mapping = mapping
+        self._mode = mode
+        self._tag = tag
+        self._device_index = torch.cuda.current_device()
+
+        # Late-bound state, populated by connect()
+        self._client = None  # GMSClientMemoryManager (lazy-typed)
+        self._is_rw: Optional[bool] = None
+
+    # ------------------------------------------------------------------
+    # Lifecycle
+    # ------------------------------------------------------------------
+
+    def connect(self) -> bool:
+        """Connect to the per-GPU GMS daemon and acquire a session.
+
+        Returns ``True`` on success. On import failure (library not
+        installed) or socket failure, returns ``False`` and logs a warning
+        — callers are expected to surface a useful error to the user.
+        """
+        try:
+            from gpu_memory_service.client.torch.allocator import (
+                get_or_create_gms_client_memory_manager,
+            )
+            from gpu_memory_service.common.locks import GrantedLockType, RequestedLockType
+            from gpu_memory_service.common.utils import get_socket_path
+            from gpu_memory_service.integrations.common.patches import patch_empty_cache
+        except ImportError:
+            logger.warning(
+                "gpu_memory_service library not installed; cannot use "
+                "LoadFormat.GMS. Install from "
+                "https://github.com/ai-dynamo/dynamo/tree/main/lib/gpu_memory_service"
+            )
+            return False
+
+        mode_map = {
+            "rw": RequestedLockType.RW,
+            "ro": RequestedLockType.RO,
+            "auto": RequestedLockType.RW_OR_RO,
+        }
+
+        # Resolve default socket path against the upstream convention so
+        # multiple TRT-LLM instances on the same node automatically share
+        # the same per-GPU daemon.
+        socket_path = self._socket_path
+        if socket_path is None:
+            socket_path = get_socket_path(self._device_index, self._tag)
+        self._socket_path = socket_path
+
+        try:
+            self._client = get_or_create_gms_client_memory_manager(
+                socket_path,
+                self._device_index,
+                mode=mode_map[self._mode],
+                tag=self._tag,
+            )
+        except Exception as e:
+            logger.warning(
+                "Failed to connect to GMS at %s (mode=%s, tag=%s): %s",
+                socket_path,
+                self._mode,
+                self._tag,
+                e,
+            )
+            self._client = None
+            return False
+
+        self._is_rw = self._client.granted_lock_type == GrantedLockType.RW
+
+        # GMS-backed VMM tensors can segfault if torch.cuda.empty_cache()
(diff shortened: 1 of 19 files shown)


tensorrt_llm/_torch/memory/gpu_memory_backend.py
```diff
@@ -0,0 +1,437 @@
...
+        from gpu_memory_service.client.torch.allocator import gms_use_mem_pool
+
+        target_device = device
+        if target_device is None:
+            target_device = torch.device("cuda", self._device_index)
+
+        with gms_use_mem_pool(self._tag, target_device):
+            yield
+
+    def move_untracked_params(self, model: nn.Module) -> None:
```
galletas1712, line 275: are we sure this won't cause an OOM? typically when you migrate something into GMS you will have to request a new allocation for the same piece of memory, then copy it in. simply remapping will not store it in GMS.

tensorrt_llm/_torch/memory/gpu_memory_backend.py
```diff
@@ -0,0 +1,437 @@
...
+        finalize_gms_write``.
+        """
+        if self._client is None:
+            raise RuntimeError("GMS client not connected. Call connect() first.")
+        if self._is_rw is False:
+            raise RuntimeError("GMS finalize_write() is only valid in RW mode.")
+
+        from gpu_memory_service.integrations.common.utils import finalize_gms_write
+
+        bytes_committed = int(finalize_gms_write(self._client, model))
```
galletas1712, line 341: finalize_gms_write is not quite the right abstraction if we are going to want to support speculative decoding models in this path as well, if I remember correctly. we would need to defer commit until draft model weights are loaded into GMS.

chienchunhung: Right — the original RW branch only loaded the main model and committed immediately, which would have left model.draft_model either uncommitted or out-of-pool. Just pushed **c6c6e570c** which loads draft weights inside the same `mem_pool_scope` (mirroring the AUTO branch's if self.spec_config is not None and ...need_load_draft_weights(): flow) so they land in the GMS pool. `finalize_gms_write(model)` then picks them up via its existing model-tree walk (`finalize_gms_write()` walks `named_parameters`, so anything reachable from model — including `model.draft_model` — gets registered in the same commit).

tensorrt_llm/_torch/memory/gpu_memory_backend.py
chienchunhung: Good catch. To clarify: `GMSBackend.move_untracked_params` is not a remap — it walks the model's parameters, calls `gms_client.create_mapping(size=nbytes, tag=self._tag)` to allocate fresh GMS-pool memory, then `replacement.copy_(tensor)` and reassigns `tensor.data = replacement`. Mirrors upstream `gpu_memory_service.integrations.trtllm.model_loader._move_untracked_params` 1:1. So the bytes do end up in GMS.

The transient peak you're worried about is real though: for the duration of each per-tensor `replacement.copy_(tensor)`, both the old (out-of-pool) buffer and the new (in-pool) buffer exist before the old one is GC'd. In practice this only fires for stray params that landed outside mem_pool_scope (post-load transforms, etc.), which is a small fraction of total weights — but for very large models on tight budgets it could spike.

Also, in this round I moved `move_untracked_params` inside mem_pool_scope (commit **c6c6e570c**) to match the upstream `_load_rw` reference path exactly. That doesn't change the per-tensor peak, but it does ensure any intermediate buffers `tensor.copy_()` may allocate via PyTorch's caching allocator end up in the GMS pool and get drained by the subsequent `torch.cuda.empty_cache()`.

Worth tracking as an upstream ask: a streaming/in-place migrator that rebinds an existing physical alloc to a GMS-mapped VA without a copy. What do you think?

chienchunhung:
@coderabbitai review

chienchunhung:
Opened another PR (https://github.com/NVIDIA/TensorRT-LLM/pull/13531) to extract the MX-only part.

Changing this PR back to the draft mode to keep it as a reference; no need for code review for this PR.

PR #13055, opened 2026-04-14
[TRTLLM-9920][feat] Add support for arbitrary KVCache transfer

Tabrizian:
## Dev Engineer Review

- Adds request-free arbitrary KV-cache transfer using serialized transceiver state, reuse-tree lookup, block pinning, and cleanup on success/failure.
- Extends Python, RPC, worker, LLM, and HTTP APIs with `get_data_transceiver_state()`.
- Adds serialization support and explicit arbitrary-transfer metadata.
- Reuse-tree locking and rollback improve consistency, though the broader locking changes may affect contention/performance and need validation.
- Partial-prefix transfer and avoiding blocks already cached by the receiver remain unsupported.
- Mamba cache support is tracked separately in `TRTLLM-13970`.
- CI merge-request pipelines consistently reported failure; investigation and a successful rerun are required before approval.

## QA Engineer Review

Added:
- `test_arbitrary_kv_cache_transfer`
- `test_arbitrary_kv_cache_transfer_missing_blocks`

Both are included in `tests/integration/test_lists/test-db/l0_a10.yml` for CI coverage. The tests cover successful transfer, cache reuse, missing sender blocks, error handling, and continued worker operation.

**Verdict: sufficient**

## Description

Adds support for **arbitrary KV cache transfer** between TRT-LLM workers: a generation worker can request KV cache blocks directly from another worker's reuse tree, without requiring a preceding `context_only` request on the sender.

Use cases:
- Pull cached prompts across workers to amortize prefill cost
- Re-route generation to a different worker after routing decisions change
- Move KV cache during load balancing or failover

```python
def arbitrary_cache_transfer():
     """Test KV cache transfer from the reuse tree.

      Flow:
      1. Worker 0 runs a normal generate to fill the KV cache reuse tree.
      2. Retrieve the data_transceiver_state from worker 0.
      3. Worker 1 sends a generation_only request using that state.
         Sender (worker 0) serves blocks directly from its reuse tree.
      """
      kv_cache_config = KvCacheConfig(max_tokens=2048 * 8, enable_block_reuse=True)
      cache_transceiver_config = CacheTransceiverConfig(backend="DEFAULT")

      worker0 = LLM(model=model_path(model),
                    tensor_parallel_size=1,
                    disable_overlap_scheduler=True,
                    cuda_graph_config=CudaGraphConfig(),
                    kv_cache_config=kv_cache_config,
                    cache_transceiver_config=cache_transceiver_config)

      worker1 = LLM(model=model_path(model),
                    tensor_parallel_size=1,
                    disable_overlap_scheduler=not generation_overlap,
                    cuda_graph_config=CudaGraphConfig(),
                    kv_cache_config=kv_cache_config,
                    cache_transceiver_config=cache_transceiver_config)

      prompt = "What is the capital of Germany?"

      # 1. Fill worker 0's reuse tree with a normal generate
      worker0.generate(prompt, SamplingParams(max_tokens=10, ignore_eos=True))

      # 2. Grab the transceiver state (opaque bytes) from worker 0
      state = worker0.get_data_transceiver_state()
      assert isinstance(state, bytes) and len(state) > 0

      # 3. generation_only on worker 1 using state from worker 0.
      #    max_tokens=1 — only one decode step, so KV transfer must
      #    deliver all prompt tokens.
      disagg_params = DisaggregatedParams(
          request_type="generation_only",
          opaque_state=state,
          first_gen_tokens=[0],
          disagg_request_id=42,
      )
      output = worker1.generate(prompt,
                                SamplingParams(max_tokens=1, ignore_eos=True),
                                disaggregated_params=disagg_params)
      assert len(output.token_ids) > 0

      # 4. Normal request on worker 1 still works after the transfer
      output2 = worker1.generate(prompt,
                                 SamplingParams(max_tokens=10, ignore_eos=True))
      assert len(output2.token_ids) > 0
      cached1 = output2.cached_tokens

      # 5. Unrelated prompt on worker 1 → no reuse match
      output3 = worker1.generate("List three primes greater than one hundred",
                                 SamplingParams(max_tokens=5, ignore_eos=True))
      cached2 = output3.cached_tokens
      assert cached2 <= 1            # at most BOS overlap
      assert cached2 < cached1       # transferred prompt cached more
```
### Process overview

```mermaid
sequenceDiagram
    participant Client
    participant W0 as Worker 0 (Sender)
    participant W1 as Worker 1 (Receiver)

    Note over W0: Normal request fills<br/>KV cache reuse tree
    Client->>W0: generate("What is...?")
    W0-->>Client: output tokens + KV stored in reuse tree

    Note over Client: Client queries<br/>DataTransceiverState from W0
    Client->>W0: llm.get_data_transceiver_state()
    W0-->>Client: opaque_state (bytes)

    Note over Client,W1: Arbitrary KV transfer via generation_only
    Client->>W1: generate(prompt, generation_only,<br/>opaque_state=state)
    W1->>W0: requestAndReceive (RequestInfo)
    W0->>W0: BlockRange::fromReuseTree(<br/>pinBlocks=true)
    W0->>W0: pinBlocksById (incRef + claimBlock)
    W0->>W1: send KV cache blocks
    W0->>W0: unpinBlocksById (decRef)
    W1-->>Client: generated tokens
```

### Key implementation points

- **`LLM.get_data_transceiver_state()`** — new API that returns the serialized `CacheState + CommState` (connection + layout info) needed by the receiver.
- **`CacheFormatter::getBlockRangeForSending()`** — when no `LlmRequest` is provided, the blocks are looked up via `BlockRange::fromReuseTree()` using the token hashes the receiver has declared.
- **Block pinning during transfer** — `fromReuseTree(pinBlocks=true)` increments the ref count (and claims the block from the eviction policy if it had no refs) to prevent the sender's reuse tree from evicting the blocks mid-transfer. `sendSyncFromReuseTree()` unpins after the send completes.

## Test Coverage

Integration test in `tests/integration/defs/disaggregated/test_disaggregated_single_gpu.py::test_arbitrary_kv_cache_transfer` (registered in `l0_a10.yml`):

1. Worker 0: normal generate → fills reuse tree
2. Query `data_transceiver_state` from worker 0
3. Worker 1: `generation_only` with `opaque_state` + `max_tokens=1` → asserts KV cache was delivered (no prefill recompute path)
4. Post-transfer: normal request on worker 1 works (`cached_tokens > 0`)
5. Different prompt on worker 1 → `cached_tokens ≤ 1` (only BOS matches)

## PR Checklist

- [x] PR follows TRT-LLM coding guidelines.
- [x] Test cases provided for new code paths.
- [x] No new dependencies.
- [x] Reviewers assigned.

## GitHub Bot Help


Tabrizian:
Just an FYI, this is not completed yet but wanted to open it for reviews if there are any early comments or feedback.

nvpohanh:
@eopXD @thorjohnsen @SimengLiu-nv could you review this? thanks

tests/integration/test_lists/test-db/l0_a10.yml
```diff
@@ -80,6 +79,8 @@ l0_a10:
   - disaggregated/test_disaggregated_single_gpu.py::test_disaggregated_simple_llama[False-True-TinyLlama-1.1B-Chat-v1.0]
   - disaggregated/test_disaggregated_single_gpu.py::test_disaggregated_simple_llama[True-False-TinyLlama-1.1B-Chat-v1.0]
   - disaggregated/test_disaggregated_single_gpu.py::test_disaggregated_simple_llama[True-True-TinyLlama-1.1B-Chat-v1.0]
+  - disaggregated/test_disaggregated_single_gpu.py::test_arbitrary_kv_cache_transfer[False-TinyLlama-1.1B-Chat-v1.0]
```
StanleySun639, line 107: Could you please add the new feature test cases into QA test[ lists](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/test_lists/qa/llm_function_core.txt) to gain test coverage on more GPUs, thanks!

StanleySun639 (approved):

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1902,9 +1940,25 @@ std::shared_ptr<KVCacheBlock> WindowBlockManager::findBlocksInReuseTreeByBlockKe
 
         if (matchingBlock == nullptr)
         {
+            // Roll back any pins taken during this partial walk so callers see a
+            // clean miss with no refcount side-effects.
```
eopXD: Should also clean `pinnedBlockIds` so the function is correct by contract. No incorrectness for now though according to your code logic but this hides a potential hazard.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -78,8 +78,9 @@ void TransferSession::send(size_t idx, void const* data, size_t size)
     }
     catch (std::exception const& e)
     {
+        auto requestId = mRequest ? mRequest->mRequestId : 0;
```
eopXD: I don't think its a good convention to have 0 here. `0` is also a valid request ID. The state here should be representing a "no llmRequest (llmRequest-agnostic) send".

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -91,15 +92,19 @@ void TransferSession::recv(size_t idx, void* data, size_t size)
...
```
eopXD: Same comment under `TransferSession::send`.

cpp/tensorrt_llm/batch_manager/mlaCacheFormatter.cpp
```diff
@@ -124,17 +124,20 @@ void MLACacheFormatter::format(tensorrt_llm::batch_manager::TransferSession& ses
 {
     NVTX3_SCOPED_RANGE(MLACacheFormatter_format);
     session.setTime(TransferSession::kTimeFormatter);
-    auto const& llmRequest = session.getLlmRequest();
-    TLLM_LOG_DEBUG(
-        mpi::MpiComm::world().getRank(), "Start sending KV cache for request ID: %ld.", llmRequest.mRequestId);
+    auto llmRequest = session.getLlmRequest();
+    if (llmRequest.has_value())
+    {
+        TLLM_LOG_DEBUG(
```
eopXD, line 130: If the goal is a "request agnostic" transfer, is this debug message still desirable? Same comment goes to the other guards you have added under this file.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -593,72 +608,116 @@ class CacheSender::Impl
...
+                        }
+                        else
                         {
-                            break;
+                            sendReadySignal(reqId, true);
+                            std::promise<void> promise;
+                            Response resp{nullptr, reqId, std::move(promise), std::move(pinnedIds)};
+                            if (dynamic_cast<executor::kv_cache::AgentConnectionManager*>(mManager) != nullptr)
+                            {
+                                sendAndRemoveResponse(reqId, std::move(resp));
```
eopXD, line 955: Double `sendReadySignal` called. Maybe delete the one under `sendSyncFromReuseTree`?

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -593,72 +608,116 @@ class CacheSender::Impl
...
+                            }
+                            else
+                            {
+                                asyncSendAndRemoveResponse(reqId, std::move(resp));
+                            }
```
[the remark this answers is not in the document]
eopXD: Coderabbit pointed the problem out but the proposed fix is not correct.
Measurement report should be re-considered and not tied to llmRequest.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -1902,9 +1940,25 @@ std::shared_ptr<KVCacheBlock> WindowBlockManager::findBlocksInReuseTreeByBlockKe
...
+            for (auto const& block : pinnedInScope)
+            {
+                block->decRefCount();
+            }
             return nullptr;
         }
 
+        if (pinBlocks)
+        {
+            matchingBlock->incRefCount();
```
eopXD: This pinning is half done.

Check the pinning under `WindowBlockManager::storeBlocks`. You should claim the block as well.
Unify such pinning action and allow callers to leverage the same function to avoid logics scattered.

eopXD (commented):
Thank you for the merge request Iman. There is still some correctness issues that needs to be resolved before merging the MR. Please checkout the comments.

Moreover, I think the ultimate goal should be a robust "llmRequest-free" transfer. The current blocker should be the RNN caches. The logic should be a cache locator where for KV caches, we seek for block IDs (for v1) and slot IDs (for v2). For RNN caches, we should seek for the slot ID. May you share with me how you plan to follow-up this to allow  this to happen?

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
Tabrizian: Addressed in a8a597a797.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
Tabrizian: Addressed in a52c9f5e41.

Tabrizian:
@eopXD Thanks for the review. I have created a ticket here for mamba cache: https://jirasw.nvidia.com/browse/TRTLLM-13970

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
Tabrizian: Addressed in a52c9f5e41

cpp/tensorrt_llm/batch_manager/mlaCacheFormatter.cpp
Tabrizian: We can't get rid of the path that requires request completely so we still need to keep the logic that relates to older path.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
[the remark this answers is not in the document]
Tabrizian: Addressed in 861091120e.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -593,72 +608,116 @@ class CacheSender::Impl
...
+        {
+            std::unique_lock<std::mutex> lk(mMtxForMap);
+            auto it = mRequestToSession.find(requestId);
+            TLLM_CHECK(it != mRequestToSession.end());
+            session = std::addressof(it->second);
+        }
+        // format() without LlmRequest — uses reuse tree path
+        mCacheTransferLayer.format(*session);
+        sendReadySignal(requestId, true);
+    }
```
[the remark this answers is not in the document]
Tabrizian: Fixed.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
Tabrizian: Good catch, fixed.

cpp/tensorrt_llm/batch_manager/dataTransceiver.h
```diff
@@ -119,7 +119,7 @@ class TransferSession
 
     void recv(size_t idx, void* data, size_t size);
 
-    [[nodiscard]] LlmRequest const& getLlmRequest() const;
+    [[nodiscard]] std::optional<LlmRequest const*> getLlmRequest() const;
```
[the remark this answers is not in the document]
Tabrizian: Addressed in 3ae8176929.

tensorrt_llm/executor/base_worker.py
```diff
@@ -659,6 +659,16 @@ def get_disaggregated_params(self) -> dict:
             return {}
         return self.engine.kv_cache_transceiver.get_disaggregated_params()
 
+    def get_cache_transceiver(self):
+        if self.engine is None:
+            return None
+        return self.engine.kv_cache_transceiver
```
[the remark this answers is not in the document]
Tabrizian: Addressed in a3c9e3a689.

tensorrt_llm/executor/executor.py
```diff
@@ -367,6 +367,12 @@ def aget_kv_events(self, timeout=None) -> IterationResult:
     def get_disaggregated_params(self) -> dict:
         return {}
 
+    def get_cache_transceiver(self):
+        return None
```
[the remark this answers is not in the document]
Tabrizian: Addressed in a3c9e3a689.

tensorrt_llm/executor/proxy.py
```diff
@@ -400,6 +400,17 @@ def get_disaggregated_params(self) -> dict:
...
+    def get_data_transceiver_state(self) -> bytes:
+        """Get serialized DataTransceiverState from worker runtime via RPC."""
+        if self.rpc_client is None:
+            return b""
+        try:
+            return self.rpc_client.get_data_transceiver_state().remote()
+        except RPCError as e:
+            logger.warning(
+                f"Error fetching data transceiver state via RPC: {e}")
+            return b""
```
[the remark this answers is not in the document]
Tabrizian: Addressed in 83c037d811.

tests/integration/defs/disaggregated/test_disaggregated_single_gpu.py
```diff
@@ -938,5 +953,280 @@ def test_disaggregated_logits(model, generation_overlap):
...
+            assert len(responses) == 1
+            intercomm.send("GET_LAST_CACHED_TOKENS", dest=1, tag=MPI_REQUEST)
+            cached2 = intercomm.recv(source=1, tag=MPI_RESULT)
+            print(f"Worker 1 different prompt cached_tokens: {cached2}",
+                  flush=True)
+            # BOS token may be shared, so at most 1 cached token is expected
+            assert cached2 <= 1, \
+                f"Expected at most 1 cached token for unrelated prompt, got {cached2}"
+            assert cached2 < cached1, \
+                f"Unrelated prompt should have fewer cached tokens than transferred prompt"
```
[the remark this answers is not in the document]
Tabrizian: Addressed in 168fd990ae.

tests/integration/test_lists/test-db/l0_a10.yml
Tabrizian: Addressed in 4db0be2172.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
Tabrizian: Addressed in a8a597a797.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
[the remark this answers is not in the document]
Tabrizian: Addressed in 9f30754cef — request-free transfers are now excluded from KV cache time capture.

jieli-matrix (approved):

cpp/tensorrt_llm/executor/cache_transmission/ucx_utils/ucxCacheCommunicator.cpp
```diff
@@ -620,7 +620,16 @@ Connection const* UcxConnectionManager::recvConnect(DataContext const& ctx, void
         buffer.data(), buffer.size(), ucxx::Tag(ctx.getTag()), ucxx::TagMask(0xFFFFFFFF), false, completionCallback);
     if (!req->isCompleted())
     {
-        future.get();
+        // Poll with timeout to allow checking the terminate flag
+        auto const& terminate = ctx.getTransferTerminate();
+        while (future.wait_for(std::chrono::milliseconds(100)) != std::future_status::ready)
+        {
+            if (terminate.load())
```
eopXD, line 636: We will need to handle worker teardown here correctly.
The completion callback will be fired accessing `promise.set_value` where the stack-framed value will no longer be feasible if the UCX worker is no longer available.

To safely handle this, decouple the object lifetime from the stack frame and have it heap-allocated.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -2142,7 +2155,36 @@ std::shared_ptr<KVCacheBlock> WindowBlockManager::findBlocksInReuseTreeByBlockKe
         blockKeys.emplace_back(blockKey.usesExtraIds, blockKey.loraTaskId, blockedUniqueTokensList, blockKey.extraKeys,
             blockKey.cacheSalt);
     }
-    return searchReuseTree(blockKeys);
+    auto searchRoot = mCachedBlocksRoot;
+    std::vector<BlockPtr> pinnedInScope;
+    for (auto const& blockKey : blockKeys)
+    {
+        auto [partialMatch, numMatched, matchingBlock] = searchRoot != nullptr
```
eopXD, line 2169: To clarify further, are partial-matched blocks also feasible for this arbitrary transfer?
Do we distinguish them from full-matched blocks?

eopXD (commented):
Currently the arbitrary transfer has its limitations.

We throw for
- RNN / hybrid (mamba-attention) sender
- VSWA model (pool > 1)
- ppSize > 1
- layerwise transfer config

From a top-down point-of-view, should we recognize the inability for arbitrary transfer before we attempt to do so? This avoids condition checking and throw and leave us with simple assertions.

eopXD (changes requested):
Thinking further, we are now trying to support an "arbitrary" transfer theme.

The pinBlock/unpinBlock action under WindowBlockManager touches refCount, now we are not doing this under the py_executor loop, that is assumed to be single-threaded. The dis-aggregated setting, with this "arbitrray transfer" we are trying to support in this merge request, we may be trying to pin a block that has no ref count for trasnfer. Such pinning may race to block-obtaining for a request of block since the ref count of an unclaimed block is zero and it is inside the queue for getFreeBlock.

This leads to corruption in data or worse, an undetected double use which is probably very hard to debug. This goes similar to `mEvictionPolicy` and the structure it touches.

Please check again on the code and consider the concurrenting schme to make sure bookkeeping is well-protected.

nafis271:
Thank you so much for working on this!

One more API that would be really helpful: for the same request, different workers have very different prefix-cache hit rates; and often the best-cached worker is also the busy one (say A at ~70% hit but very busy) while an idle replica (B at ~10%) is free. We would like to run the request on the idle B but first pull the cached blocks from the busy A into B, so B serves at A's hit rate and the heavy compute lands on the idle worker while still reusing the KV cached on the loaded one.  Since A typically holds only a prefix of the prompt (the 70%, not all of it), this would need a partial transfer- basically pull whatever prefix the sender has and let the receiver prefill the remaining tail and generate.

Tabrizian:
@nafis271 thank for your feedback. Just to make sure I understand your request correctly, is it possible to open a github issue that illustrates the API you need?

If your request is when transferring from worker A to B to not transfer the blocks that worker B already has this is definitely something that we plan to work on next.

cpp/tensorrt_llm/executor/cache_transmission/ucx_utils/ucxCacheCommunicator.cpp
Tabrizian: I fixed it.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
Tabrizian: We don't distinguish but only the portion of partial block that has correct info will be stored in destination reuse tree.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
```diff
@@ -414,12 +425,13 @@ void CacheFormatter::format(tensorrt_llm::batch_manager::TransferSession& sessio
     SizeType32 const numKvPools = static_cast<SizeType32>(kvWindowSizes.size());
 
     TLLM_LOG_DEBUG("CacheFormatter::format: allWindowSizes=%zu, kvWindowSizes=%d, numPools=%d, requestId=%lu",
-        allWindowSizes.size(), numKvPools, numPools, llmRequest.mRequestId);
+        allWindowSizes.size(), numKvPools, numPools, llmRequest.has_value() ? (*llmRequest)->mRequestId : 0);
```
eopXD: requestID `0` is a valid requestID.

eopXD (approved):
Change looks good to me with one nit. Thank you for patiently addressing the comments.

juney-nvidia (approved):
Approved from API perspective, with leaving a small comment.

tests/unittest/api_stability/api_stability_core.py
```diff
@@ -157,9 +157,18 @@ def from_inspect(cls, method: MethodType):
...
+        """Strip the :tag:`...` prefix added by @set_api_status decorator."""
+        if docstring and docstring.startswith(":tag:"):
+            import re
+            docstring = re.sub(r'^:tag:`[^`]*`\s*', '', docstring)
+        return docstring
+
     @classmethod
     def from_docstring(cls, method: MethodType):
-        doc = docstring_parser.parse(method.__doc__)
+        docstring = cls._strip_api_status_tag(method.__doc__)
```
juney-nvidia, line 170: Out of curiosity, why do we need to add this change? No need to block this PR though.
@Tabrizian

Tabrizian: There was a bug in parsing the docstrings for APIs that didn't have `Args:` in the docstring section.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -2130,6 +2136,19 @@ bool WindowBlockManager::blockInRadixTree(BlockPtr const& block)
...
+    return findBlocksInReuseTreeByBlockKeyImpl(blockKey, /*pinBlocks=*/false, unusedPinnedBlockIds);
+}
+
+std::shared_ptr<KVCacheBlock> WindowBlockManager::findBlocksInReuseTreeByBlockKey(
+    BlockKey const& blockKey, std::vector<KVCacheBlock::IdType>& pinnedBlockIds)
+{
+    return findBlocksInReuseTreeByBlockKeyImpl(blockKey, /*pinBlocks=*/true, pinnedBlockIds);
+}
+
+std::shared_ptr<KVCacheBlock> WindowBlockManager::findBlocksInReuseTreeByBlockKeyImpl(
```
Shixiaowei02, line 2151: It replaces the original path with block traversal, which bypasses the existing SWA logic and stops the chain on nullptr. Since this overload is used by `BlockRange::fromReuseTree()` in the standard ctx->gen path, the impact should be confirmed again.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -815,16 +905,14 @@ class CacheSender::Impl
...
-                    {
-                        break;
-                    }
+                    break;
                 }
 
+                // Always listen for incoming transfer requests: arbitrary (llmRequest-agnostic)
+                // transfers arrive without a pre-registered response, so the loop cannot gate on
+                // mReadyResponses being non-empty.
                 auto requestInfo = recvRequestInfo();
```
Shixiaowei02, line 912: Please check for the potential deadlock regression risks here.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
```diff
@@ -197,9 +197,11 @@ BlockRange getBlockRangeForSending(BaseKVCacheManager* cacheManager, LlmRequest
         || lastBlockKey.uniqueTokens.size() == 0 || recvSideHasCP || ppSize > 1)
     {
         // disable reuse path, and vwsa don't support reuse.
+        // This path requires an LlmRequest with an active sequence in the cache manager.
+        TLLM_CHECK_WITH_INFO(llmRequest.has_value(), "LlmRequest required for non-reuse-tree transfer path");
```
Shixiaowei02, line 200: This could introduce a new timeout when exception is thrown here.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -700,6 +749,12 @@ class CacheSender::Impl
             discardTransferState(id);
             failResponse(resp, exception);
         }
+        // Release any reuse-tree blocks we pinned for this transfer. Safe on every exit
+        // path; the LlmRequest-backed branch never populates mPinnedBlockIds.
+        if (!resp.mPinnedBlockIds.empty())
```
Shixiaowei02: This conflicts with the `noexcept` behavior in `sendAndRemoveResponse` and could potentially lead to `std::terminate`.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -837,22 +925,76 @@ class CacheSender::Impl
...
+                        std::scoped_lock lock(mSenderMutex);
+                        mCurrentRequest = reqId;
+                    }
+                    auto countIt = mRemainSendCount.find(reqId);
+                    auto const count = --countIt->second;
+                    TLLM_CHECK(count >= 0);
+                    if (count == 0)
+                    {
+                        mRemainSendCount.erase(countIt);
+                        auto pinnedIds = pinReuseTreeBlocks(reqId);
```
Shixiaowei02, line 937: Please ensure the logic remains consistent across different ranks when reaching this path, to avoid introducing deadlocks in the future.

tensorrt_llm/serve/openai_server.py
```diff
@@ -1097,6 +1100,19 @@ def register_visual_gen_routes(self):
                                self.delete_video,
                                methods=["DELETE"])
 
+    async def data_transceiver_state(self) -> JSONResponse:
+        """Return the serialized DataTransceiverState as base64-encoded JSON."""
+        state = self.generator.get_data_transceiver_state()
```
Shixiaowei02, line 1105: Please confirm the perf impact here, and ensure it does not conflict with the ongoing refactor by @reasonsolo .

Shixiaowei02:
Hi @Tabrizian , thanks for the efforts. I've completed a review of several code-level details. More importantly, I'd like to better understand whether, given our current infrastructure, we still need the direct D2D transfer from gen->ctx. With the current constraint that the transfer can only begin after the ctx request has completed, would leveraging **router affinity** to keep cache locality be able to achieve a similar effect in a more direct way, without requiring the additional D2D transfer path? I'd appreciate your thoughts. Thanks!

tensorrt_llm/serve/openai_server.py
Tabrizian: @Shixiaowei02 this doesn't have any perf impact. This is not in the hot path for disagg transfer.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -2142,7 +2161,36 @@ std::shared_ptr<KVCacheBlock> WindowBlockManager::findBlocksInReuseTreeByBlockKe
...
-    return searchReuseTree(blockKeys);
+    auto searchRoot = mCachedBlocksRoot;
+    std::vector<BlockPtr> pinnedInScope;
+    for (auto const& blockKey : blockKeys)
+    {
+        auto [partialMatch, numMatched, matchingBlock] = searchRoot != nullptr
+            ? searchRoot->findMatchingBlock(blockKey, true, true)
+            : std::make_tuple(false, 0, nullptr);
+
+        if (matchingBlock == nullptr)
```
chienchunhung: Is it sufficient to check only `matchingBlock` without checking `numMatched`?

IIUC A non-null `matchingBlock` does not prove that all requested tokens matched. For example, if the cached final key is `[A, B, C]` and the requested key is `[A, B, C, D, E]`, `findMatchingBlock()` can return the cached block with `partialMatch=true` and `numMatched=3`. The implementation logic here does not check the matched length and would treat that result as a **complete** transfer. Pinning the matched prefix is fine, but given that the current arbitrary-transfer request is `generation_only`, the receiver has no context phase to compute the missing suffix. 

Shall we require `numMatched == blockKey.uniqueTokens.size()` for every requested chunk? 

PS: Exact non-full final blocks (i.e., matching suffix) where all requested tokens match should remain valid.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
chienchunhung: I have the same question. 

IIUC For SWA, `findReusableBlockMatches()` can traverse value-less anchors and only accept the chain after the missing anchor falls outside the attention window; this walk instead returns a miss at the first null block and cannot represent `isTraversalOnly`. That can regress standard context-to-generation transfers for SWA. 

Shall we either preserve searchReuseTree for the existing non-pinning overload, or make the pinning path reuse the same SWA-aware match plan? PS: Adding a test to cover for SWA could also be usful.

chienchunhung (commented):
Thanks for the PR!

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
```diff
@@ -806,6 +861,41 @@ class CacheSender::Impl
...
+    // empty vector means no matching chain was found.
+    std::vector<kv_cache_manager::KVCacheBlock::IdType> pinReuseTreeBlocks(RequestIdType requestId)
+    {
+        std::unique_lock<std::mutex> lk(mMtxForMap);
+        auto it = mRequestToSession.find(requestId);
+        auto const& lastBlockKey = it->second.getLastBlockKey();
+        auto* cacheManager = mCacheTransferLayer.getCacheManager();
+        auto windowSize = cacheManager->getBlockManager().getWindowSizesMetadata().begin()->first;
+        std::vector<kv_cache_manager::KVCacheBlock::IdType> pinnedIds;
+        auto lastBlock = cacheManager->findBlocksInReuseTreeByBlockKey(lastBlockKey, windowSize, pinnedIds);
```
chuangz0, line 888: The code seems couldn't handle worker with a secondary (host) pool enabled, offloaded blocks stay in the reuse tree, so pinReuseTreeBlocks() still matches and pins them, the lookup never checks isPrimary() and doesn't onboard.
But may need check whether  the blocks is primary ,and send not_ready to requester if not

chuangz0:
I think unit tests like `CacheTransceiverTest` are required.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
Tabrizian: @Shixiaowei02 This is assertion so if the could reaches here it means that there is a bug that need to be fixed.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
Tabrizian: @chienchunhung This is not supported today for SWA. There is follow up ticket for SWA support.

@Shixiaowei02 This is mostly a small refractor of previous logic. Let me know if there is a specific part where we need to quantify the perf impact.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
Tabrizian: Good catch! Fixed.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
Tabrizian: Thanks for the feedback, it is fixed now.

cpp/tensorrt_llm/batch_manager/dataTransceiver.cpp
Tabrizian: Good catch, fixed.

Shixiaowei02 (approved):
Today, we effectively have five tools available:

1. Session affinity
2. Conditional disaggregation
3. P2P direct KV transfer (this PR)
4. KV offload
5. Recomputation (re-prefill)


Roughly speaking:

- Short follow-up turns: (1)
- Tool-oriented agentic workflows (short generation + long tool output): (3)
- Shared system prefixes: (1) + (4)
- Heavy agentic workloads (long generation + large inputs): (2)
- Multi-user shared generation: (4)
- KV not yet committed: (4) or (5)


As the framework evolves, we now have a broader set of tools to address different requirements. For the agentic workflow scenario above, P2P KV transfer seems like a viable approach. While extending the current transceiver model could enable additional use cases, it may also be worth balancing that flexibility with keeping the underlying design simple and maintainable. We can certainly explore this direction in v1. For v2, my current view is that it may fit better as a dedicated abstraction or module rather than an extension of the existing transfer model.

Thanks for your efforts!

PR #13059, opened 2026-04-15
[TRTLLM-11421][fix] fix data races in speculative decoding fast-logits handoff

eopXD:
## Summary

- `draftModelSendLogitsThread` called `terminateRequest()` directly from a background thread, racing with the main thread on `KVCacheManager::removeSequence()` and other shared batch manager state (sequence slot manager, PEFT cache manager).
- Fix: after sending logits via MPI, the background thread now pushes completed requests into a handoff queue (`mDraftRequestsDoneSendingLogits`) drained by the main thread in `forwardSync()`, where `terminateRequest()` is called.
- The same two threads also shared `mDraftRequestsWaitingToSendLogits` (main thread push_back, background thread iterate+erase) without synchronization. Fix: extend the existing mutex (renamed `mDraftRequestsDoneMtx` → `mDraftRequestsMtx`) to guard both queues. MPI sends remain outside the lock to avoid serializing network I/O.
- This also removes the logits thread dependency on `SequenceSlotManager`, `BaseKVCacheManager`, and `BasePeftCacheManager`, simplifying its interface.

## Related

- Complements #12251 (KVCacheManager TOCTOU race fixes) -- that PR fixes races within the KV cache manager itself; this PR fixes the upstream race where `terminateRequest` was called from the wrong thread entirely.

## Test plan

- [x] Existing speculative decoding fast-logits orchestration tests pass
- [x] Stress test with overlap scheduling / MTP mode to exercise the concurrent send-logits + terminate path
- [x] CI: `/bot run`

## Summary by CodeRabbit

* **Refactor**
  * Refactored speculative decoding thread coordination: request termination is now deferred from the background logits thread to the main thread via a shared synchronization queue, improving inter-thread synchronization and simplifying background thread dependencies.

### DIFF
--- a/cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
+++ b/cpp/tensorrt_llm/batch_manager/trtGptModelInflightBatching.cpp
@@ -487,7 +487,7 @@ TrtGptModelInflightBatching::TrtGptModelInflightBatching(std::shared_ptr<nvinfer
     {
         mDraftModelSendLogitsThread
             = std::make_unique<std::thread>(&utils::draftModelSendLogitsThread, mDevice, &mDraftModelThreadShouldExit,
-                &mDraftRequestsWaitingToSendLogits, &mDraftRequestsDoneSendingLogits, &mDraftRequestsDoneMtx);
+                &mDraftRequestsWaitingToSendLogits, &mDraftRequestsDoneSendingLogits, &mDraftRequestsMtx);
     }
 
     mCreateNewDecoderRequests = std::make_unique<CreateNewDecoderRequests>(
@@ -908,7 +908,7 @@ void TrtGptModelInflightBatching::forwardSync()
         {
             RequestVector doneSending;
             {
-                std::lock_guard<std::mutex> lk(mDraftRequestsDoneMtx);
+                std::lock_guard<std::mutex> lk(mDraftRequestsMtx);
                 doneSending.swap(mDraftRequestsDoneSendingLogits);
             }
             for (auto const& llmReq : doneSending)
@@ -2537,6 +2537,7 @@ void TrtGptModelInflightBatching::updateRequests(ScheduledRequests const& schedu
             {
                 if (llmReq->getReturnGenerationLogits() && mSpeculativeDecodingFastLogits && mIsLeaderInOrchMode)
                 {
+                    std::lock_guard<std::mutex> lk(mDraftRequestsMtx);
                     mDraftRequestsWaitingToSendLogits.push_back(llmReq);
                 }
                 else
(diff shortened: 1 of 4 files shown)


SimengLiu-nv:
Codex review points out a potential race condition:

The race is between two threads mutating the same std::vector without a happens-before relationship.

In the main batch-manager thread, updateRequests() appends completed draft requests here:

``` c++
// trtGptModelInflightBatching.cpp (line 2527)

mDraftRequestsWaitingToSendLogits.push_back(llmReq);
```

At the same time, the background MPI sender thread scans and erases from that same vector here:

``` c++
// logitsThread.cpp (line 90)

for (auto it = draftRequestsWaitingToSendLogits->begin();
     it != draftRequestsWaitingToSendLogits->end(); ++it)
{
    ...
    draftRequestsWaitingToSendLogits->erase(it);
}
```
A concrete interleaving:

Draft request A finishes on the draft model. The main thread enters updateRequests() and is about to append A to mDraftRequestsWaitingToSendLogits.

The target model concurrently calls targetModelReceiveLogits() and sends an MPI request asking the draft model for logits for an earlier request B.

draftModelSendLogitsThread() receives that MPI request and starts iterating mDraftRequestsWaitingToSendLogits to find B.

While the sender thread holds an iterator into the vector, the main thread executes push_back(A).

If push_back() reallocates, the vector moves its storage and invalidates the sender thread’s iterator. The sender can then dereference freed/moved storage or erase using an invalid iterator.

Even if no reallocation happens, push_back() still mutates vector size and constructs a new element while the sender thread is reading begin(), end(), and elements. That is still a C++ data race and therefore undefined behavior.

There is also the reverse mutation case: the sender thread may erase() an element while the main thread is appending. erase() shifts elements and updates vector size; push_back() may update size/capacity at the same time. That can corrupt vector metadata, drop a request, duplicate a request pointer, or make the later TLLM_CHECK(draftRequest != nullptr) fail spuriously.

The PR fixes the later “done sending, now terminate” handoff with a mutex, but the earlier “waiting to send logits” handoff remains unsynchronized.

SimengLiu-nv (changes requested):
Need to verify potential race conditions.

eopXD:
Addressed with latest commit. Now the four access sites, all now under mDraftRequestsMtx:

- trtGptModelInflightBatching.cpp:2540 — push_back to waiting queue (main thread, new lock)
- logitsThread.cpp:100 — iterate+erase on waiting queue (sender thread, new lock)
- logitsThread.cpp:119 — push_back to done queue (sender thread, existing lock, mutex renamed)
- trtGptModelInflightBatching.cpp:911 — swap of done queue (main thread, existing lock, mutex renamed)

SimengLiu-nv (approved):
LGTM.

eopXD:
/bot --skip "We had earned success before. x86 Single GPU has passed. Failure in the latest CI/CD is not related to our code change. Let us merge the MR without further delay."

eopXD:
/bot skip --comment "We had earned CI/CD success before. In the latest CI/CD, x86 Single GPU has passed. The failure is not related to our code change. Let us merge the MR without further delay."

PR #11708, opened 2026-02-25
[TRTLLM-11163][feat] Introduce a fast path (token IDs + MM) for VLMs instead of de-tokenizing already encoded prompt

moraxu:
## Release Notes

* **New Features**
  * Added fast-path support for processing pre-tokenized prompts with multimodal data, eliminating unnecessary detokenization steps for improved performance.
  * Enhanced image placeholder expansion and handling for multimodal inputs.

* **Refactor**
  * Improved input processing pipeline to seamlessly support both text-based and pre-tokenized prompt paths with unified multimodal handling.

## Description

Process token IDs + MM data without de-tokenizing, that is, by instead:

1. Processing multi-modal inputs with dummy inputs (for media placeholder tokens to match the number of multi-modal inputs),
2. Replacing the placeholder token IDs for the multi-modal inputs with the actual feature token IDs.

See: https://docs.vllm.ai/en/latest/design/mm_processing/#multi-modal-data-processing

Currently implemented only for `tensorrt_llm/_torch/models/modeling_llava_next.py`

## Test Coverage

Tested for:

- `pytest -sv tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py::TestLlava_V1_6_Mistral_7B::test_auto_dtype`
- `pytest -sv tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py::TestQwen3VL::test_auto_dtype`
- `llava-v1.6-mistral-7b-hf` in Dynamo with https://github.com/ai-dynamo/dynamo/pull/6567/:
    - image URL:
        - AGG, E/P/D
    - embeddings:
        - AGG, E/P/D, E/PD
    - text only:
        - AGG, E/P/D
- `Qwen/Qwen3-VL-2B-Instruct` in Dynamo with https://github.com/ai-dynamo/dynamo/pull/6567/:
    - image URL:
        - E/P/D

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
[...]

PR #13069, opened 2026-04-15
[TRTLLM-11974][feat] Handle multimodal placeholder expansion in token space for Nemotron Nano

moraxu:
## Summary by CodeRabbit

* **New Features**
  * Faster tokenized multimodal path with improved placeholder expansion for images, video, and audio.
  * More efficient video token pruning and EVS-aware retained-token computation.

* **Bug Fixes**
  * More consistent placeholder expansion results and clearer error cases for mismatched multimodal input.

* **Tests**
  * Extensive new unit and integration tests covering multimodal preprocessing, placeholder variants, and EVS behaviors.

## Description

Adds the tokenized-prompt + multimodal-data fast path (the same pattern PR #11708 introduced for LLaVA-Next) to Nemotron Nano V3 OMNI, covering all three modalities — image, video, audio — including EVS-enabled video. Also fixes a pre-existing bug in `NanoV2VLInputProcessor.get_num_tokens_per_video`'s EVS branch where the per-tubelet retained token count was inflated by routing through `get_num_tokens_per_image`'s dynamic-tiler logic, causing `find_mm_token_positions` to disagree with the vision encoder's real post-EVS output.

- `tensorrt_llm/_torch/models/modeling_nemotron_nano.py` — Adds `get_text_with_mm_placeholders`, `get_num_tokens_per_audio`, `expand_prompt_token_ids_for_mm` (+ per-modality helpers) to `NanoV2VLInputProcessor`. Video expansion uses a two-tier strategy mirroring vLLM's `_apply_prompt_updates` (token-level subsequence match with a text-level decode/split/re-encode fallback). EVS support included via a parallel `evs_ids` stream consumed by `merge_evs_mm_embeds`. Fixes `get_num_tokens_per_video`'s EVS branch to derive `video_size` the same way `_process_videos_frames` does.

- `tensorrt_llm/inputs/evs.py` - Extracts `compute_retained_tokens_from_tubelet_budget` as the shared retention-count helper; `compute_retained_tokens_count` now delegates to it.

- `tensorrt_llm/inputs/multimodal.py` — Adds audio branch to `find_mm_token_lengths`; fixes `find_mm_token_positions` no-match early return from `[]` to `([], [])` to match its declared 2-tuple signature.

- `tensorrt_llm/inputs/registry.py` - `tokenized_multimodal_process` threads `multi_modal_data` through to the input processor (needed by video expansion) and merges any returned `mm_data_updates` (e.g. `evs_ids`) into `extra_processed_inputs["multimodal_data"]`. Adds "audio" to the multimodal-hashing modality whitelist.

- `tensorrt_llm/_torch/models/modeling_llava_next.py` - `expand_prompt_token_ids_for_mm` signature updated to return `Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]`; always returns `(expanded, None)`.

## Test Coverage

`tests/unittest/_torch/modeling/test_nemotron_nano_preprocessing.py` - 8 new test classes covering each fast-path method, multi-token BPE patterns, EVS token-level + text-level tiers, dispatch integration, and audio.

Tested with VLMEvalKitMcore for the following datasets InfoVQA_VAL, OCRBench, VoxPopuli_ASR_test, TedLium_ASR_Test, Video-MME (EVS off), Video-MME (EVS of 0.5) by temporarily tokenizing the text prompt to trigger the fast path. The scores match the default, non-fast path where the text prompt and MM data is handled directly by the HF processor.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_llava_next.py
+++ b/tensorrt_llm/_torch/models/modeling_llava_next.py
@@ -155,6 +155,7 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
         prompt_token_ids: List[int],
         num_mm_tokens_per_placeholder: List[int],
         hf_processor_mm_kwargs: Optional[Dict[str, Any]] = None,
+        mm_data: Optional[Dict[str, Any]] = None,
     ) -> Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
         """
         Expands MM placeholder tokens in `prompt_token_ids` so that each single placeholder
@@ -172,6 +173,10 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
                 specifies the number of MM feature tokens to expand/repeat for that placeholder.
             hf_processor_mm_kwargs (Optional[Dict[str, Any]]): Optional dictionary of arguments
                 to pass to the HuggingFace processor, if needed for token expansion.
+            mm_data (Optional[Dict[str, Any]]): The original `multi_modal_data` dict
+                (e.g. `{"image": [...], "video": [...]}`). Subclasses whose expansion
+                depends on the raw MM content (e.g. per-video frame metadata) use this;
+                LLaVA-Next ignores it because its expansion is a pure token-ID rewrite.
 
         Returns:
             Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
@@ -180,6 +185,7 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
                 `extra_processed_inputs["multimodal_data"]`. Always `None` for LLaVA-Next
                 (no auxiliary streams needed).
         """
+        del mm_data  # LLaVA-Next's image-only expansion doesn't need raw MM data
         expanded, _, _ = self._expand_image_placeholders_in_token_ids(
             prompt_token_ids, num_mm_tokens_per_placeholder)
         return expanded, None
--- a/tensorrt_llm/_torch/models/modeling_llava_next.py
+++ b/tensorrt_llm/_torch/models/modeling_llava_next.py
@@ -155,6 +155,7 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
         prompt_token_ids: List[int],
         num_mm_tokens_per_placeholder: List[int],
         hf_processor_mm_kwargs: Optional[Dict[str, Any]] = None,
+        **kwargs,
     ) -> Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
         """
         Expands MM placeholder tokens in `prompt_token_ids` so that each single placeholder
--- a/tensorrt_llm/_torch/models/modeling_llava_next.py
+++ b/tensorrt_llm/_torch/models/modeling_llava_next.py
@@ -155,7 +155,6 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
         prompt_token_ids: List[int],
         num_mm_tokens_per_placeholder: List[int],
         hf_processor_mm_kwargs: Optional[Dict[str, Any]] = None,
-        mm_data: Optional[Dict[str, Any]] = None,
     ) -> Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
         """
         Expands MM placeholder tokens in `prompt_token_ids` so that each single placeholder
@@ -173,10 +172,6 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
                 specifies the number of MM feature tokens to expand/repeat for that placeholder.
             hf_processor_mm_kwargs (Optional[Dict[str, Any]]): Optional dictionary of arguments
                 to pass to the HuggingFace processor, if needed for token expansion.
-            mm_data (Optional[Dict[str, Any]]): The original `multi_modal_data` dict
-                (e.g. `{"image": [...], "video": [...]}`). Subclasses whose expansion
-                depends on the raw MM content (e.g. per-video frame metadata) use this;
-                LLaVA-Next ignores it because its expansion is a pure token-ID rewrite.
 
         Returns:
             Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
@@ -185,7 +180,6 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
                 `extra_processed_inputs["multimodal_data"]`. Always `None` for LLaVA-Next
                 (no auxiliary streams needed).
         """
-        del mm_data  # LLaVA-Next's image-only expansion doesn't need raw MM data
         expanded, _, _ = self._expand_image_placeholders_in_token_ids(
             prompt_token_ids, num_mm_tokens_per_placeholder)
         return expanded, None
--- a/tensorrt_llm/_torch/models/modeling_llava_next.py
+++ b/tensorrt_llm/_torch/models/modeling_llava_next.py
@@ -155,7 +155,7 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
         prompt_token_ids: List[int],
         num_mm_tokens_per_placeholder: List[int],
         hf_processor_mm_kwargs: Optional[Dict[str, Any]] = None,
-    ) -> List[int]:
+    ) -> Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
         """
         Expands MM placeholder tokens in `prompt_token_ids` so that each single placeholder
         is replaced by the corresponding number of multimodal feature tokens.
@@ -174,12 +174,15 @@ class LlavaNextInputProcessor(BaseMultimodalInputProcessor,
                 to pass to the HuggingFace processor, if needed for token expansion.
 
         Returns:
-            List[int]: The prompt token IDs where each MM placeholder token has been
-                replaced/expanded with the appropriate number of MM feature tokens.
+            Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
+                `expanded_ids` (List[int]) — prompt token IDs with each MM placeholder expanded.
+                `mm_data_updates` (Optional[Dict]) — extra fields to merge into
+                `extra_processed_inputs["multimodal_data"]`. Always `None` for LLaVA-Next
+                (no auxiliary streams needed).
         """
         expanded, _, _ = self._expand_image_placeholders_in_token_ids(
             prompt_token_ids, num_mm_tokens_per_placeholder)
-        return expanded
+        return expanded, None
 
     def _postprocess(
         self, input_ids: torch.Tensor, mm_features: Union[torch.Tensor,
--- a/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
@@ -994,10 +994,10 @@ class NanoV2VLInputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyIn
 
     def get_mm_special_token_ids(self) -> torch.Tensor:
         "Return multimodal special token ids for NanoV2VL."
-        ids = [self.image_start_token_id, self.image_end_token_id]
+        ids = list(self._img_start_token_ids) + list(self._img_end_token_ids)
         if self._sound_start_token_id is not None:
             ids.extend([self._sound_start_token_id, self._sound_end_token_id])
-        return torch.tensor(ids)
+        return torch.tensor(sorted(set(ids)))
 
     def get_mm_token_ids(self):
         ids = [self.img_context_token_id]
--- a/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
@@ -44,6 +44,16 @@ from .modeling_utils import register_auto_model
 
 # Set max_num_tiles to 1 for video modality, to match the training behavior.
 VIDEO_MAX_NUM_TILES = 1
+# Several video code paths (`_process_videos_frames`'s HF-processor fallback,
+# `_get_video_tokens_per_frame`'s fallback, `_compute_video_size_lightweight`,
+# and `get_num_tokens_per_video`'s EVS branch) all assume a single tile per
+# frame. Increasing this constant would require updating those paths
+# accordingly — multi-tile video expansion isn't currently supported and the
+# prompt-side MM counts would diverge from the vision encoder's output.
+# Guard against accidental change.
+assert VIDEO_MAX_NUM_TILES == 1, (
+    "NanoV2VL video paths assume a single tile per frame; see comment above."
+)
 IMAGE_PLACEHOLDER = "<image>"
 VIDEO_PLACEHOLDER = "<video>"
 AUDIO_PLACEHOLDER = "<so_embedding>"
--- a/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
@@ -997,7 +997,7 @@ class NanoV2VLInputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyIn
         ids = list(self._img_start_token_ids) + list(self._img_end_token_ids)
         if self._sound_start_token_id is not None:
             ids.extend([self._sound_start_token_id, self._sound_end_token_id])
-        return torch.tensor(sorted(set(ids)))
+        return torch.tensor(ids)
 
     def get_mm_token_ids(self):
         ids = [self.img_context_token_id]
@@ -1009,9 +1009,6 @@ class NanoV2VLInputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyIn
         """Return minimal placeholder text for the given multimodal item counts,
         so that the HF processor can be called with (dummy_text, mm_data)
         without error. Used when processing tokenized prompt + MM data.
-
-        NanoV2VL currently accepts only one modality per request, so at most
-        one of `image`, `video`, or `audio` will be non-zero in `mm_counts`.
 
         Args:
             mm_counts (Dict[str, int]): A mapping of each multimodal modality
--- a/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
@@ -1038,9 +1038,9 @@ class NanoV2VLInputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyIn
         num_videos = mm_counts.get("video", 0)
         num_audios = mm_counts.get("audio", 0)
         parts: List[str] = []
-        parts.extend([IMAGE_PLACEHOLDER] * num_images)
-        parts.extend([VIDEO_PLACEHOLDER] * num_videos)
-        parts.extend([AUDIO_PLACEHOLDER] * num_audios)
+        parts.extend([self.img_context_token] * num_images)
+        parts.extend([self.video_context_token] * num_videos)
+        parts.extend([self._sound_context_token] * num_audios)
         return "".join(parts)
 
     def get_num_tokens_per_image(
@@ -1057,10 +1057,7 @@ class NanoV2VLInputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyIn
             budget = self.dynamic_tiler._max_num_patches
             params, _ = self.dynamic_tiler.process_media(image, budget)
             num_image_tokens = params.num_embeddings
-            # Add only image-specific special tokens (img_start, img_end),
-            # not all multimodal special tokens (which also includes
-            # sound_start, sound_end when audio is supported).
-            num_image_tokens += 2  # <img> and </img>
+            num_image_tokens += len(self._img_start_token_ids) + len(self._img_end_token_ids)
             return num_image_tokens
 
         # The logic is copied and modified from HuggingFace ImageProcessor.
@@ -1125,10 +1122,7 @@ class NanoV2VLInputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyIn
         if self.processor.use_thumbnail and blocks != 1:
             blocks += 1
         num_image_tokens = self.num_image_token * blocks
-        # Add only image-specific special tokens (img_start, img_end),
-        # not all multimodal special tokens (which also includes
-        # sound_start, sound_end when audio is supported).
-        num_image_tokens += 2  # <img> and </img>
+        num_image_tokens += len(self._img_start_token_ids) + len(self._img_end_token_ids)
         return num_image_tokens
 
     def _get_video_tokens_per_frame(
@@ -1567,7 +1561,7 @@ class NanoV2VLInputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyIn
                 expansion.extend(prefix_ids)
                 if evs_expansion is not None:
                     evs_expansion.extend(prefix_ids)
-            for frame_sep, num_tokens in zip(frame_seps, tokens_per_frame):
+            for frame_sep, num_tokens in zip(frame_seps, tokens_per_frame, strict=True):
                 sep_ids = self.tokenizer.encode(frame_sep, add_special_tokens=False)
                 expansion.extend(sep_ids)
                 expansion.extend(self._img_start_token_ids)
--- a/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
@@ -24,6 +24,7 @@ from ...inputs import (
     MultimodalPlaceholderPlacement,
     TextPrompt,
     compute_retained_tokens_count,
+    compute_retained_tokens_from_tubelet_budget,
     compute_retention_mask,
     register_input_processor,
 )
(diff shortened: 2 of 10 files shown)


moraxu:
@coderabbitai full review

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu, line 1130: cc @Wanli-Jiang about the changes we discussed

Wanli-Jiang: should we add a guard check between base codes and the changed codes in unittests?

Also would it break the assumption that VIDEO_MAX_NUN_TILES is not 1? if so better to add an assertion or check in the `compute_retained_tokens_from_tubelet_budget`?

moraxu: Good point, addressed in https://github.com/NVIDIA/TensorRT-LLM/pull/13069/changes/f5ac9d0d6f7338308f68f0b729f8a3584daee624

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -43,6 +44,16 @@
...
 # Set max_num_tiles to 1 for video modality, to match the training behavior.
 VIDEO_MAX_NUM_TILES = 1
+# Several video code paths (`_process_videos_frames`'s HF-processor fallback,
+# `_get_video_tokens_per_frame`'s fallback, `_compute_video_size_lightweight`,
+# and `get_num_tokens_per_video`'s EVS branch) all assume a single tile per
+# frame. Increasing this constant would require updating those paths
+# accordingly — multi-tile video expansion isn't currently supported and the
+# prompt-side MM counts would diverge from the vision encoder's output.
+# Guard against accidental change.
+assert VIDEO_MAX_NUM_TILES == 1, (
```
2ez4bz: Hm... we set `VIDEO_MAX_NUM_TILES = 1` just above. What is the point of this assert?

tensorrt_llm/inputs/registry.py
```diff
@@ -785,10 +785,24 @@ def tokenized_multimodal_process(
                 "tokenized_multimodal_process: find_mm_token_lengths returned "
                 "no token lengths for the provided multi_modal_data.")
 
-        expanded_ids = input_processor.expand_prompt_token_ids_for_mm(
+        # Build kwargs for expand, injecting mm_data so model-specific
+        # expansion (e.g. video frame separators) can access it.
+        expand_kwargs = dict(inputs.get("mm_processor_kwargs") or {})
+        expand_kwargs["_multi_modal_data"] = mm_data
```
2ez4bz: Just curious, why the leading underscore? Also, these don't seem like `hf_processor_mm_kwargs` anyway (in the sense that they aren't passed to the `hf_processor` when calling it).

Should we add a new `mm_data` argument to the `expand_prompt_token_ids_for_mm` interface signature?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,529 @@ def get_num_tokens_per_video(
...
+    ) -> Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
+        """Expand MM placeholder tokens in `prompt_token_ids` so that each
+        single placeholder is replaced by the corresponding number of
+        multimodal feature tokens.
+
+        This is used when processing a tokenized prompt plus multimodal data,
+        without calling the full HuggingFace processor. Detects which modality
+        is present by scanning for placeholder token IDs and dispatches to the
+        matching per-modality expansion helper. NanoV2VL allows only one
+        modality per request, so this raises `ValueError` if placeholders for
```
2ez4bz, line 1225: What happens when we send videos with audio? For example, if we spin the server up with:

```
--media_io_kwargs '{"video": {"extract_audio": true}}'
```

Here's a test video I've been using: https://qianwen-res.oss-cn-beijing.aliyuncs.com/Qwen2-VL/space.mp4

tensorrt_llm/inputs/evs.py
```diff
@@ -4,11 +4,35 @@
 import torch
 
 
+def compute_retained_tokens_from_tubelet_budget(num_tubelets: int,
```
2ez4bz, line 7: Nit: should we maybe move this to nano instead and revert the changes to `compute_retained_tokens_count`? We can commonize if / when necessary. For example, I'm not sure which version Qwen3 VL would want to call if we ended up having to implement EVS for it.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,529 @@ def get_num_tokens_per_video(
...
+        return expanded, evs
+
+    def _text_level_video_replace(
+        self,
+        prompt_token_ids: List[int],
+        video_expansions: List[List[int]],
+        video_evs_expansions: Optional[List[List[int]]] = None,
+    ) -> Tuple[List[int], Optional[List[int]]]:
+        """Fallback used when token-level subsequence matching fails because
+        BPE merged the `<video>` placeholder with its trailing context.
```
2ez4bz, line 1702: 🫠  What a weird decision not to have `<video>` be its own special token.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,529 @@ def get_num_tokens_per_video(
...
+                expanded.append(tok)
+
+        if audio_idx != len(num_mm_tokens_per_placeholder):
+            raise ValueError(
+                f"Expected {len(num_mm_tokens_per_placeholder)} audio "
+                f"placeholders, found {audio_idx}."
+            )
+        return expanded
+
+    def _compute_video_size_lightweight(
```
2ez4bz: Nit: we can drop the `_lightweight`.

tests/integration/test_lists/test-db/l0_l40s.yml
```diff
@@ -18,6 +18,7 @@ l0_l40s:
   - unittest/_torch/modeling -k "modeling_siglip"
   - unittest/_torch/modeling -k "modeling_vila"
   - unittest/_torch/modeling -k "modeling_nemotron_nano_v2_vl"
+  - unittest/_torch/modeling/test_nemotron_nano_preprocessing.py -k "TestGetTextWithMMPlaceholders or TestExpandImagePlaceholders or TestExpandAudioPlaceholders or TestExpandVideoPlaceholders or TestExpandVideoPlaceholdersMultiToken or TestExpandVideoPlaceholdersEVS or TestExpandPromptTokenIdsForMM or TestGetNumTokensPerAudio or TestGetNumTokensPerVideoInvariants"
```
2ez4bz: Isn't the entire preprocessing file covered in [this test list](https://github.com/NVIDIA/TensorRT-LLM/blob/659d5fb506953ab66b33086e773ab8ae743c11a6/tests/integration/test_lists/test-db/l0_a10.yml#L24)?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Currently, `VIDEO_MAX_NUM_TILES > 1` would introduce silent errors in other asserts, just like the original issue of video token calculations not being in sync with each other in different methods caused some assertion to fail in `find_mm_token_positions`. I just wanted to have an official assert for the tile number, if that makes sense? As per @Wanli-Jiang comment. I realize this might look bizarre given the assignment above, I'm open to other suggestions.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Yes, it's only used internally on a per-tubelet/frame basis for the purpose of EVS. I mean the one in the model config that maps to a single ID

tensorrt_llm/inputs/evs.py
moraxu: I see, my preference would be to leave it for now and only move it to nano if Qwen3 VL ends up requiring a different definition for those calculations? That way, we're slightly minimizing code duplication..

tests/integration/test_lists/test-db/l0_l40s.yml
moraxu: Good catch, removed.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Ah, right, I haven't tested out that PR that got merged in the meantime. I'd have to make some modifications in `_expand_video_placeholders_in_token_ids`, `get_text_with_mm_placeholders` and `expand_prompt_token_ids_for_mm` for this to work - feel free to add other comments (the detailed ones) first, unless you'd prefer for me to implement these tweaks first?

tensorrt_llm/inputs/registry.py
moraxu: why the leading underscore?

Yeah, it is to distinguish it from a regular hf_processor_mm_kwargs arg - you're right that a separate argument would be cleaner. Addressed in https://github.com/NVIDIA/TensorRT-LLM/pull/13069/commits/8b54134c5305b5bb00413e87f987680efcf3128d

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: How about at least `_compute_video_size_descriptor` given the multiple methods of a similar name that compute tokens per video/frame etc. etc.?

tensorrt_llm/_torch/models/modeling_llava_next.py
```diff
@@ -155,7 +155,8 @@ def expand_prompt_token_ids_for_mm(
         prompt_token_ids: List[int],
         num_mm_tokens_per_placeholder: List[int],
         hf_processor_mm_kwargs: Optional[Dict[str, Any]] = None,
-    ) -> List[int]:
+        mm_data: Optional[Dict[str, Any]] = None,
```
2ez4bz: Nit: can just be a blanket `**kwargs` if it's going to be unused.

tensorrt_llm/_torch/models/modeling_llava_next.py
```diff
@@ -172,14 +173,22 @@ def expand_prompt_token_ids_for_mm(
...
         Returns:
-            List[int]: The prompt token IDs where each MM placeholder token has been
-                replaced/expanded with the appropriate number of MM feature tokens.
+            Tuple[List[int], Optional[Dict[str, Dict[str, Any]]]]:
+                `expanded_ids` (List[int]) — prompt token IDs with each MM placeholder expanded.
+                `mm_data_updates` (Optional[Dict]) — extra fields to merge into
+                `extra_processed_inputs["multimodal_data"]`. Always `None` for LLaVA-Next
+                (no auxiliary streams needed).
         """
+        del mm_data  # LLaVA-Next's image-only expansion doesn't need raw MM data
```
2ez4bz: I don't know why claude like to do this. This is at best confusing, and should be removed.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
2ez4bz: I think it would make more sense to have it at line 1818, if there's any place to have it.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,533 @@ def get_num_tokens_per_video(
...
+        #      (e.g. when `<video>` sits at a token boundary).
+        #
+        #   2. **Text-level fallback**: when token-level matching doesn't
+        #      locate all expected occurrences (because BPE merged the
+        #      placeholder's last token with the following character — e.g.
+        #      `>` + `\n` fuses into a single different token ID), decode
+        #      `prompt_token_ids` back to text, split on the `<video>`
+        #      string, and re-encode each segment. This is the same "search
+        #      across token boundaries" escape hatch vLLM uses.
+        tier1 = self._try_token_level_video_replace(
```
2ez4bz: Nits:

1. `_try` -> `_maybe`? `try` has a particular connotation in python.
2. `try / except TypeError` instead of `tier1 is not None`?

*EDIT*: Thinking about this more, should we even bother trying when `len(self._video_token_ids) > 1`? maybe we can just choose which of these to use depending on the length?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,533 @@ def get_num_tokens_per_video(
...
+                expanded.append(tok)
+
+        if image_idx != len(num_mm_tokens_per_placeholder):
+            raise ValueError(
+                f"Expected {len(num_mm_tokens_per_placeholder)} image "
+                f"placeholders, found {image_idx}."
+            )
+        return expanded
+
+    def _expand_audio_placeholders_in_token_ids(
```
2ez4bz: Nit: could the image helper and this one leverage a shared helper? I was thinking we could pass in the `start_ids` and `end_ids` as variables; for images, they would be `self._img_start_token_ids` and `self._img_end_token_ids`, whereas for audio, `[self._sound_start_token_id]` and `[self._sound_end_token_id]`.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
2ez4bz: Sure. Could consider `_shape` to instead of `_size`, to borrow a tensor / array term.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,533 @@ def get_num_tokens_per_video(
...
+        #
+        # Per-tubelet img_context counts come from
+        # `_compute_token_numbers_per_video`, which matches what the
+        # vision encoder actually produces post-EVS (single-tile
+        # per-tubelet spatial grid). `get_num_tokens_per_video` — used by
+        # `find_mm_token_lengths` — must agree with this.
+        video_expansions: List[List[int]] = []
+        video_evs_expansions: Optional[List[List[int]]] = [] if evs_enabled else None
+        for video_data in videos:
+            if hasattr(video_data, "frames"):
```
2ez4bz: Nit:

```python
frames = getattr(video_data, "frames", video_data)
metadata = getattr(video_data, "metadata", None)
```.

Btw, could you remind me why they could appear in 2 different ways? Maybe leave a comment?

File: tensorrt_llm/_torch/models/modeling_nemotron_nano.py [hunk 13]
```diff
@@ -1127,26 +1172,533 @@ def get_num_tokens_per_video(
...
+                frames = video_data.frames
+                metadata = video_data.metadata
+            else:
+                frames = video_data
+                metadata = None
+
+            video_size = self._compute_video_size_lightweight(frames)
+            # When EVS is enabled, `_compute_token_numbers_per_video`
+            # returns the dummy pattern [K_pre_evs, 0, 0, ...] — same shape
+            # the non-fast-path uses to size `input_ids` pre-EVS-merge;
```
2ez4bz: Nit: `non-fast-path` -> `str-replacement-path`?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,533 @@ def get_num_tokens_per_video(
...
+            return None
+
+        pattern_len = len(pattern)
+        expanded: List[int] = []
+        evs: Optional[List[int]] = [] if video_evs_expansions is not None else None
+        video_idx = 0
+        i = 0
+        while i < len(prompt_token_ids):
+            if (
+                i + pattern_len <= len(prompt_token_ids)
```
2ez4bz: Nit: this check is unnecessary. While indexing with indices that are out of range will raise an `IndexError`, slicing does not. See e.g.

```
In [1]: a = []

In [2]: a[5:8]
Out[2]: []
```

So comparing `prompt_token_ids[i:i + pattern_len] == pattern` suffices.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,533 @@ def get_num_tokens_per_video(
...
+                and prompt_token_ids[i : i + pattern_len] == pattern
+            ):
+                if video_idx >= len(video_expansions):
+                    # More token-level matches than videos — ambiguous; bail.
+                    return None
+                expanded.extend(video_expansions[video_idx])
+                if evs is not None and video_evs_expansions is not None:
```
2ez4bz: Nit: based on line 1631 where `evs` is defined, only one of these checks is necessary.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1172,533 @@ def get_num_tokens_per_video(
...
+                i += pattern_len
+            else:
+                tok = prompt_token_ids[i]
+                expanded.append(tok)
+                if evs is not None:
+                    evs.append(tok)
+                i += 1
+
+        if video_idx != len(video_expansions):
+            return None  # At least one <video> not found as a subsequence.
```
2ez4bz: Note: if you decide to choose between `_try_token_level_video_replace` and `_text_level_video_replace` based on the length of the video IDs like I suggested, then this line and line 1641 can be replaced with runtime errors.

tensorrt_llm/inputs/evs.py
2ez4bz: Ok. Maybe leave a comment indicating this is a nano-specificity at the moment.

tensorrt_llm/inputs/multimodal.py
```diff
@@ -862,7 +862,9 @@ def find_mm_token_positions(
 
     # If no multimodal tokens found, return empty list
     if not torch.any(mm_mask):
-        return []
+        # Preserve the declared 2-tuple return type so callers can unpack
```
2ez4bz: Sorry, even with the comment, I do not follow why this change was necessary. Can you elaborate?

tests/unittest/_torch/modeling/test_modeling_llava_next.py
```diff
@@ -118,10 +118,13 @@ def test_llava_next_expand_prompt_token_ids_for_mm():
     prompt_token_ids = [1, 2, image_token_id, 3, image_token_id, 4]
     num_mm_tokens_per_placeholder = [10, 20]
 
-    expanded = input_processor.expand_prompt_token_ids_for_mm(
+    expanded, mm_data_updates = input_processor.expand_prompt_token_ids_for_mm(
         prompt_token_ids, num_mm_tokens_per_placeholder
     )
 
+    # LLaVA-Next has no auxiliary streams; mm_data_updates must be None.
```
2ez4bz: Sorry, I find the term `stream` a bit undescriptive here. Is there a better term we can come up with?

tests/unittest/_torch/modeling/test_nemotron_nano_preprocessing.py
```diff
@@ -999,3 +999,592 @@ def test_prediction_matches_actual(self, orig_sr, target_sr, audio_length):
...
+    proc.image_end_token_id = 501
+    # Audio special token IDs.
+    proc._sound_context_token_id = 30
+    proc._sound_start_token_id = 200
+    proc._sound_end_token_id = 201
+    return proc
+
+
+class TestGetTextWithMMPlaceholders:
+    def test_single_image(self):
```
2ez4bz: Can we parameterize these using `pytest.mark.parametrize`?

venkywonka:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Good point, applied that change.

tensorrt_llm/inputs/multimodal.py
moraxu: That's what the only call site of this method looks like:
```
        start_positions, start_special_token_positions = find_mm_token_positions(
            input_ids=prompt_token_ids,  # token sequence
            num_mm_tokens=
            num_mm_tokens,  # list of lengths of each chunk of visual tokens
            vocab_size=vocab_size,
            mm_token_ids=mm_ids,
            mm_special_token_ids=mm_special_token_ids,
        )
```

tensorrt_llm/inputs/registry.py
```diff
@@ -785,10 +785,22 @@ def tokenized_multimodal_process(
...
-        expanded_ids = input_processor.expand_prompt_token_ids_for_mm(
+        expanded_ids, mm_data_updates = input_processor.expand_prompt_token_ids_for_mm(
             prompt_token_ids,
             num_mm_tokens,
-            hf_processor_mm_kwargs=inputs.get("mm_processor_kwargs"))
+            hf_processor_mm_kwargs=inputs.get("mm_processor_kwargs"),
+            mm_data=mm_data,
+        )
+
+        # Merge any model-specific auxiliary streams (e.g. video evs_ids for
```
moraxu, line 875: TODO: Add comment here that this is because evs_ids would be computed based off the incorrect dummy prompt

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Added in https://github.com/NVIDIA/TensorRT-LLM/pull/13069/changes/e84b0c52bbd053ee7c96776c1873f3400a886a27.

I'll add unit tests and run all the datasets again later this week/on Monday

tensorrt_llm/_torch/models/modeling_llava_next.py
```diff
@@ -155,7 +155,7 @@ def expand_prompt_token_ids_for_mm(
         prompt_token_ids: List[int],
         num_mm_tokens_per_placeholder: List[int],
         hf_processor_mm_kwargs: Optional[Dict[str, Any]] = None,
```
venkywonka, line 157: this signature seems to conflict with how its called in registry.py (missing `mm_data=...`)

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1127,26 +1163,579 @@ def get_num_tokens_per_video(
...
+            and "audio_samples" in video_metadata
+            and self._audio_extractor is not None
+        ):
+            num_total_tokens += self.get_num_tokens_per_audio(
+                audio=(
+                    video_metadata["audio_samples"],
+                    video_metadata["audio_sample_rate"],
+                )
+            )
+
```
venkywonka, line 1203: codex identified repo-wide gap in audio support: 

```
tensorrt_llm/inputs/multimodal.py:634 hashes VideoData frames but ignores metadata["audio_samples"],
    while this PR makes extracted audio affect token counts and embeddings at tensorrt_llm/_torch/models/
    modeling_nemotron_nano.py:1192 and tensorrt_llm/_torch/models/modeling_nemotron_nano.py:2245. Two videos
    with identical sampled frames but different audio can share a multimodal hash, so KV-cache reuse can
    return stale audio-conditioned states.
```

 So these two requests can produce the same multimodal hash:

  request A: frames = F, audio = "turn left"
  request B: frames = F, audio = "turn right"

  From the cache’s point of view, both are “the same video” because only F is hashed. From the
  model’s point of view, they are different inputs because the audio embeddings are different. If KV-
  cache reuse is enabled, request B can reuse cached KV states from request A even though those
  states were computed with request A’s audio.


This is probably beyond scope of this pr but wanted to note here - maybe a todo somewhere to track this?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -1003,10 +1044,7 @@ def get_num_tokens_per_image(
             budget = self.dynamic_tiler._max_num_patches
             params, _ = self.dynamic_tiler.process_media(image, budget)
             num_image_tokens = params.num_embeddings
-            # Add only image-specific special tokens (img_start, img_end),
-            # not all multimodal special tokens (which also includes
-            # sound_start, sound_end when audio is supported).
-            num_image_tokens += 2  # <img> and </img>
+            num_image_tokens += len(self._img_start_token_ids) + len(self._img_end_token_ids)
```
venkywonka, line 1047: does this mean we need to update `get_mm_special_token_ids()` too? that func assumes we don't BPE-split the framing tokens

tensorrt_llm/_torch/models/modeling_llava_next.py
moraxu: Good catch, I forgot to apply William's suggestion.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Good catch, will fix in the next commit.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Yup, this is indeed an issue. I'll add a TODO in `multimodal.py` with a JIRA to track this. cc @2ez4bz

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -981,17 +994,45 @@ def get_vocab_size(self):
 
     def get_mm_special_token_ids(self) -> torch.Tensor:
         "Return multimodal special token ids for NanoV2VL."
-        ids = [self.image_start_token_id, self.image_end_token_id]
+        ids = list(self._img_start_token_ids) + list(self._img_end_token_ids)
         if self._sound_start_token_id is not None:
             ids.extend([self._sound_start_token_id, self._sound_end_token_id])
-        return torch.tensor(ids)
+        return torch.tensor(sorted(set(ids)))
```
2ez4bz: Out of curiosity, does this need to be sorted? If so, should it be documented in the interface? If not, can we remove the `sorted`?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -981,17 +994,45 @@ def get_vocab_size(self):
...
         if self._sound_context_token_id is not None:
             ids.append(self._sound_context_token_id)
         return torch.tensor(ids, dtype=torch.int32)
 
+    def get_text_with_mm_placeholders(self, mm_counts: Dict[str, int]) -> str:
+        """Return minimal placeholder text for the given multimodal item counts,
+        so that the HF processor can be called with (dummy_text, mm_data)
+        without error. Used when processing tokenized prompt + MM data.
+
+        NanoV2VL currently accepts only one modality per request, so at most
```
2ez4bz: Nit: is this necessary to mention in this docstring? At a glance, it doesn't look like this limitation has any bearing on this function's code.

2ez4bz (approved):

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Reverted locally, I'll reuse the previous CI result if the pending one happens to pass

Wanli-Jiang (approved):

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
2ez4bz: @moraxu if this is all addressed, please resolve.

PR #13108, opened 2026-04-16
[None][fix] Enforce NCCL >= 2.28 at CMake configure time

eopXD:
## Summary
- Bumps the CMake `find_package(NCCL ...)` minimum version from `2` to `2.28`
- `allreduceOp.cpp` uses `NCCLWindowAllocator` and `createNCCLWindowTensor` unconditionally, but these symbols are guarded by `#if NCCL_VERSION_CODE >= NCCL_VERSION(2, 28, 0)` in `ncclUtils.h`
- Without this fix, building with NCCL < 2.28 fails deep into compilation with cryptic undeclared symbol errors. With this fix, the build fails immediately at CMake configure time with a clear version mismatch message

## Test plan
- [x] Build with NCCL >= 2.28 succeeds (no change in behavior)

## Summary by CodeRabbit

* **Chores**
  * Updated minimum NCCL dependency version requirement to 2.28 for multi-device support builds.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -226,8 +226,9 @@ set(CMAKE_CUDA_USE_RESPONSE_FILE_FOR_INCLUDES 0)
 find_library(RT_LIB rt)
 
 if(ENABLE_MULTI_DEVICE)
-  # NCCL dependencies
-  find_package(NCCL 2 REQUIRED)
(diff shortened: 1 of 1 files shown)


bobboli:
Hi @eopXD , could the version be further pinned to >= 2.29 ? There are some device APIs added.
TRTLLM image has also bumped to 2.29.2 https://github.com/NVIDIA/TensorRT-LLM/pull/11227

atrifex (approved):

PR #13116, opened 2026-04-16
[TRTLLM-10288][perf] Reduce AutoTuner host overhead in inference hot path

hyukn:
## Description

Reduces AutoTuner host overhead in the inference hot path. Profiling on B200 showed the `choose_one()` → `search_cache()` → `get_cache_key()` path adds significant Python overhead on every custom op call. This PR eliminates the bulk of it.

### Changes

**Phase 1: WAR Optimizations**
- Cached key prefix in `get_cache_key()` (avoids recomputing `__class__.__name__` + `str(unique_id())` per call)
- Singleton runner caches for `nvfp4_gemm` and `quantize_e4m3_per_tensor`

**Phase 2: `_choose_one_cache` fast path in `choose_one()`**
- `_choose_one_cache`: maps `(custom_op, runner_ids, *bucketed_dims)` → `(runner_id, tactic)`. On cache hit: one `_make_fast_key()` + one `dict.get()`, then resolve runner from caller's list.
- `_make_fast_key()`: lightweight key using only dynamic dimensions + runner identity (`_cache_key_prefix`).
- `dispatch()`: simplified to thin wrapper over `choose_one()` — delegates all caching to `_choose_one_cache`.
- `TunableRunner.bind_tactic()`: returns callable pre-bound to a specific tactic. Subclasses pre-resolve sub-runners.
- Converted 21 callsites to `dispatch()` across `torch_custom_ops.py` (12) and `cute_dsl_custom_ops.py` (9).

**Cache collision fix (runner identity in key)**
- Cache key includes `runner_ids` (tuple of each runner's `_cache_key_prefix`) so that different runner configurations for the same `custom_op` + dimensions don't collide (e.g. int8 vs int4 `weight_only_quant_gemm` runners).
- Cache stores `(runner_id, tactic)` instead of runner instances — resolution always uses caller's `runners` list.

## Profiling Results

Clean wall-clock measurements on B200, nvfp4_gemm (M=128, K=4096, N=4096, CUTLASS, 500 iters, no nsys/NVTX overhead):

| Path | Total/call | Overhead | vs Original |
|---|---|---|---|
| **Original** (choose_one + runner ctor) | ~30.1us | +12.7us | baseline |
| **This PR** (choose_one cache hit, via @custom_op) | 25.2us | +7.8us | **saves 4.9us (1.20x)** |
| Raw kernel floor (dispatch_fn direct) | 17.4us | — | — |

### Overhead breakdown

| Component | Original | This PR |
|---|---|---|
| Python dispatch logic | 6.2us | **1.3us** (79% reduction) |
| torch.ops @custom_op framework tax | 6.5us | 6.5us (unchanged) |
| **Total overhead** | **12.7us** | **7.8us** |

The \`_choose_one_cache\` reduces Python overhead from 6.2us to 1.3us by replacing per-call runner construction + \`choose_one()\` tactic search with \`_make_fast_key()\` + \`dict.get()\`.

The remaining 6.5us is the \`@torch.library.custom_op\` framework dispatch cost (DispatchKeySet traversal, auto-functionalization, schema validation). This is addressable separately via \`fast_custom_op\` (PR #13149), which reduces it to 2.3us.

### Per-step impact (40 tunable ops)

| Config | Overhead/step |
|---|---|
| Original | 1.21ms |
| This PR | 1.01ms (**-16%**) |
| This PR + fast_custom_op (#13149) | 0.84ms (**-30%**) |

### Cache collision fix
Single-GPU collision repro (B200, \`weight_only_quant_gemm\`, same dims 4096x7168x2112):
- **Before fix**: int8 runner cached, int4 call returns wrong runner → \`RuntimeError: size of tensor a (1056) must match size of tensor b (2112)\`
- **After fix**: Both int8 and int4 pass — runner identity in cache key prevents collision

### Unit tests
| Test Suite | Result |
|---|---|
| Core AutoTuner (test_autotuner.py) | All passed |
| FP4 Linear (test_fp4_linear.py) | 8 passed |
| Fused MoE (test_fused_moe.py) | Passed |

## Files Modified

| File | Change |
|---|---|
| \`tensorrt_llm/_torch/autotuner.py\` | \`_choose_one_cache\` in choose_one(), simplified dispatch(), \`_make_fast_key()\` with runner identity, \`bind_tactic()\` |
| \`tensorrt_llm/_torch/custom_ops/torch_custom_ops.py\` | Runner caches, 12 callsites → dispatch() |
| \`tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py\` | 9 callsites → dispatch() |
| \`tests/unittest/_torch/misc/test_autotuner.py\` | +2 tests: cache hit/invalidation, collision prevention |

## PR Checklist

### DIFF
--- a/tensorrt_llm/_torch/autotuner.py
+++ b/tensorrt_llm/_torch/autotuner.py
@@ -253,6 +253,21 @@ class TunableRunner(ABC):
         """
         return tuple(self.__dict__.values())
 
+    @property
+    def _cache_key_prefix(self) -> Tuple[str, str]:
+        """Cached (class_name, unique_id_str) tuple — computed once per runner instance.
+
+        This avoids re-computing runner.__class__.__name__ and str(runner.unique_id())
+        on every get_cache_key() call during inference. These values are constant for a
+        given runner instance.
+        """
+        try:
+            return self.__cache_key_prefix
+        except AttributeError:
+            self.__cache_key_prefix = (self.__class__.__name__,
+                                       str(self.unique_id()))
+            return self.__cache_key_prefix
+
 
 @contextlib.contextmanager
 def autotune(tune_mode: bool = True,
@@ -298,6 +313,9 @@ def autotune(tune_mode: bool = True,
         autotuner.skip_dynamic_tuning_buckets = old_skip
         if autotune_enabled:
             logger.info("[Autotuner] Autotuning process ends")
+            # Invalidate the choose_one cache since the profiling cache
+            # has been updated with new entries.
+            autotuner._choose_one_cache.clear()
 
         # save cache
         if cache_path is not None:
@@ -447,11 +465,15 @@ class AutoTunerProfilingCache:
         tuning_config: TuningConfig,
         apply_map_to_tuning_buckets: bool = True,
     ) -> Tuple:
+        # Use cached (class_name, unique_id_str) prefix to avoid
+        # recomputing str(runner.unique_id()) on every call (~1-2us saving).
+        prefix = runner._cache_key_prefix
         return (
             custom_op,
-            runner.__class__.__name__,
-            str(runner.unique_id()),
-            AutoTuner.get()._find_nearest_profile(
+            prefix[0],
+            prefix[1],
+            # Call classmethod directly — avoids AutoTuner.get() singleton lookup.
+            AutoTuner._find_nearest_profile(
                 input_shapes,
                 tuning_config.dynamic_tensor_specs,
                 tuning_config.constraint_specs,
@@ -760,6 +782,12 @@ class AutoTuner:
         # Last captured choose_one() contexts
         self._last_capture: Optional['AutoTuner.TacticsCapture'] = None
 
+        # Fast-path cache for choose_one() during inference.
+        # Maps (custom_op, *bucketed_dynamic_dim_values) -> (runner_id, tactic).
+        # Avoids the full search_cache → get_cache_key → _find_nearest_profile
+        # chain on repeated calls with the same bucketed input shapes.
+        self._choose_one_cache: dict = {}
+
         # Dsitributed tuning state
         self._dist: Optional[Distributed] = None
         self._has_received_cache: bool = False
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/autotuner.py
```diff
@@ -934,28 +981,37 @@ def choose_one(
...
-            # Log the cache miss. Expect no cache miss in inference.
+            fast_key = self._make_fast_key(custom_op, runners, tuning_config,
+                                           inputs)
+            cached = self._choose_one_cache.get(fast_key)
+            if cached is not None:
+                best_runner_id, best_tactic = cached
+                return (runners[best_runner_id], best_tactic)
+
+            # Cache miss — resolve via search_cache, populate cache
+            input_shapes = tuple(self._get_input_sizes(inputs))
```
liji-nv: Duplicate code for the tuning mode and non tuning mode.

hyukn: The duplication is intentional — the two paths serve different purposes and have different lifetimes:

1. **Non-tuning path** (the new cache): `search_cache` here is only hit once per unique shape (on `_choose_one_cache` miss). After warmup, this code is effectively dead — all subsequent calls hit the fast `dict.get()` above and return immediately. Storing the result in `_choose_one_cache` is what makes this a one-time cost.

2. **Tuning path** (existing code): `search_cache` here runs every iteration during active profiling and feeds into the "should we re-profile?" decision logic below it. It cannot use `_choose_one_cache` because the cache is cleared/invalid during tuning.

I considered extracting a shared helper, but it would need to return different things for each path (the non-tuning path just needs `(runner_id, tactic)` to cache; the tuning path also needs `is_cache_hit` and `min_time` for the profiling decision). The 4-line `search_cache` call is straightforward enough that a shared helper would add indirection without meaningful deduplication.

Happy to add a brief comment in the code clarifying why both exist if that would help readability.

tensorrt_llm/_torch/autotuner.py
```diff
@@ -888,6 +907,26 @@ def choose_one(
...
+                runners,
+                input_shapes,
+                tuning_config,
+                apply_map_to_tuning_buckets=True)
+            best_runner = runners[best_runner_id]
+            if not is_cache_hit:
+                logger.warning_once(
+                    f"[AutoTuner] {custom_op} using the fallback tactic, due to cache miss on input shapes={input_shapes}",
+                    key=(custom_op, "warning_autotuning_cache_miss_fallback"))
+            return (best_runner, best_tactic)
```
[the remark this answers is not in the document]
hyukn: This has been addressed — `_get_input_sizes()` is now used in the cache-miss path (line 995 in the current revision). The fast path uses `_make_fast_key()` which works on bucketed dimensions rather than raw input sizes, so the two paths are intentionally different.

For the type annotation suggestions: will add `_cache_key_prefix: str` annotation to the `KernelRunner` protocol in a follow-up cleanup.

hyukn:
The optimization will affect some of the tuning process and may hurt the overall inference perf.

PR #13151, opened 2026-04-17
[None][fix] Fix Mamba cache correctness under MTP + CUDA-graph padding

Wanli-Jiang:
# Features

- Reserve max_draft_len + 1 extra Mamba slots in
  MambaHybridCacheManager so real requests and CUDA-graph padding
  dummies both fit.
- Allocate a permanent slot for the CUDA-graph sentinel; padding
  reuses it via direct mamba_cache_index lookup and no longer aliases
  live requests parked under the overlap scheduler.
- update_mamba_states scatters into the caller's state_indices
  (mamba_metadata.state_indices under MTP), removing the stale-tail
  read.
- Relax mamba2_mtp_ssm_cache_update's intermediate_states.size(0)
  check to ">= bs"; it's indexed by batch position, not slot.
- Release Phase-1 CUDA-graph pools before final KV allocation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified Mamba state index allocation logic, removing redundant slot management.
  * Relaxed validation for Mamba2 intermediate states to permit larger cache pools.
  * Improved padding slot handling in Mamba cache manager by reusing pre-allocated sentinel slots.
  * Adjusted CUDA graph pool release timing in memory estimation path.

* **Tests**
  * Added comprehensive Mamba cache manager regression tests covering padding and state update behavior.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/thop/mamba2MTPSSMCacheOp.cpp
+++ b/cpp/tensorrt_llm/thop/mamba2MTPSSMCacheOp.cpp
@@ -55,7 +55,11 @@ void mamba2_mtp_ssm_cache_update(th::Tensor ssm, th::Tensor x, th::Tensor dt, th
     int const head_dim = ssm.size(2);
     int const ssm_dim = ssm.size(3);
 
-    TORCH_CHECK(intermediate_states.dim() == 5 && intermediate_states.size(0) == ssm.size(0)
+    // intermediate_states is indexed by intermediate_states_indices (batch
+    // position 0..bs-1), independent of ssm's slot pool; it only needs to
+    // fit the forward batch, which may be smaller than ssm.size(0) when
+    // the cache pool reserves extra slots (e.g. CUDA-graph padding dummies).
+    TORCH_CHECK(intermediate_states.dim() == 5 && intermediate_states.size(0) >= bs
             && intermediate_states.size(1) == cache_steps && intermediate_states.size(2) == nheads
             && intermediate_states.size(3) == head_dim && intermediate_states.size(4) == ssm_dim,
         "intermediate_states shape check failed");
--- a/cpp/tensorrt_llm/thop/mamba2MTPSSMCacheOp.cpp
+++ b/cpp/tensorrt_llm/thop/mamba2MTPSSMCacheOp.cpp
@@ -55,10 +55,11 @@ void mamba2_mtp_ssm_cache_update(th::Tensor ssm, th::Tensor x, th::Tensor dt, th
     int const head_dim = ssm.size(2);
     int const ssm_dim = ssm.size(3);
 
-    // intermediate_states is indexed by intermediate_states_indices (batch
-    // position 0..bs-1), independent of ssm's slot pool; it only needs to
-    // fit the forward batch, which may be smaller than ssm.size(0) when
-    // the cache pool reserves extra slots (e.g. CUDA-graph padding dummies).
+    // ssm.size(0) is the Mamba cache capacity — independent of the
+    // current batch (may include parked requests or reserved dummy
+    // slots). intermediate_states is per-step scratch indexed by
+    // intermediate_states_indices in [0, bs), so it only needs to
+    // fit the forward batch.
     TORCH_CHECK(intermediate_states.dim() == 5 && intermediate_states.size(0) >= bs
             && intermediate_states.size(1) == cache_steps && intermediate_states.size(2) == nheads
             && intermediate_states.size(3) == head_dim && intermediate_states.size(4) == ssm_dim,
--- a/tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
+++ b/tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
@@ -32,8 +32,7 @@ from tensorrt_llm._torch.pyexecutor.llm_request import LlmRequest
 from tensorrt_llm._torch.pyexecutor.resource_manager import (
     BaseResourceManager, CacheTypeCpp, DataType, KVCacheManager, get_pp_layers)
 from tensorrt_llm._torch.pyexecutor.scheduler import ScheduledRequests
-from tensorrt_llm._utils import (nvtx_range, prefer_pinned,
-                                 torch_dtype_to_binding)
+from tensorrt_llm._utils import nvtx_range, torch_dtype_to_binding
 from tensorrt_llm.bindings.internal.batch_manager import (
     KvCacheConnectorManager, LinearAttentionMetadata, LinearCacheType)
 from tensorrt_llm.llmapi.llm_args import KvCacheConfig
@@ -373,12 +372,6 @@ class PythonMambaCacheManager(BaseResourceManager):
         # mamba cache index, maps request_id -> state indices
         self.mamba_cache_index: Dict[int, int] = {}
 
-        # mamba cache state indices
-        self.state_indices: torch.Tensor = torch.arange(max_batch_size,
-                                                        device=device,
-                                                        dtype=torch.int32)
-        # save mamba state indices for requests
-        self.state_indices_list: List[int] = []
         # save intermediate state indices for requests
         self.intermediate_state_indices = torch.arange(max_batch_size,
                                                        dtype=torch.int32,
@@ -397,23 +390,13 @@ class PythonMambaCacheManager(BaseResourceManager):
 
     @torch.inference_mode()
     def _prepare_mamba_cache_blocks(self, request_ids: List[int]):
-        self.state_indices_list.clear()
         for r in request_ids:
-            # cache hit
             if r in self.mamba_cache_index:
-                self.state_indices_list.append(self.mamba_cache_index[r])
-            # cache miss
-            else:
-                if len(self.mamba_cache_free_blocks) == 0:
-                    raise RuntimeError("run out of mamba cache blocks")
-                block = self.mamba_cache_free_blocks.pop()
-                self.mamba_cache_index[r] = block
-                self.state_indices_list.append(block)
-        self.state_indices[:len(self.state_indices_list)].copy_(
-            torch.tensor(self.state_indices_list,
-                         dtype=torch.int32,
-                         pin_memory=prefer_pinned()),
-            non_blocking=True)
+                continue
+            if len(self.mamba_cache_free_blocks) == 0:
+                raise RuntimeError("run out of mamba cache blocks")
+            block = self.mamba_cache_free_blocks.pop()
+            self.mamba_cache_index[r] = block
 
     def prepare_resources(self, scheduled_batch: ScheduledRequests):
         context_ids = [
@@ -494,9 +477,6 @@ class PythonMambaCacheManager(BaseResourceManager):
 
     def shutdown(self):
         """Release tensor memory."""
-        # Clear state indices
-        self.state_indices = torch.tensor([])
-
         # Clear mamba cache states
         if isinstance(self.mamba_cache, self.SpeculativeState):
             self.mamba_cache = self.SpeculativeState(
@@ -668,15 +648,17 @@ class MambaCacheManager(BaseResourceManager, BaseMambaCacheManager):
     def shutdown(self):
         self._impl.shutdown()
 
-    def update_mamba_states(self,
-                            attn_metadata: "AttentionMetadata",
+    def update_mamba_states(self, attn_metadata: "AttentionMetadata",
                             num_accepted_tokens: torch.Tensor,
-                            state_indices: Optional[torch.Tensor] = None):
+                            state_indices: torch.Tensor):
+        # Non-speculative configs don't allocate intermediate state; the
+        # promotion is a clean no-op.
+        if not self._impl.is_speculative():
+            return
+        # Belt-and-suspenders: C++ is non-speculative today so this is
+        # unreachable. Fires if C++ ever grows speculative support
+        # without also implementing the scatter there.
         assert not self._use_cpp, "update_mamba_states is not supported in CppMambaCacheManager"
-        # Resolve the forward-path fallback outside the @torch.compile body
-        # so Dynamo only specializes on a concrete Tensor.
-        if state_indices is None:
-            state_indices = self._impl.state_indices
         self._impl.update_mamba_states(attn_metadata, num_accepted_tokens,
                                        state_indices)
 
@@ -793,14 +775,6 @@ class MixedMambaHybridCacheManager(KVCacheManager, MambaCacheManager):
                          kv_cache_dtype_byte_size: float = None):
         KVCacheManager.update_resources(self, scheduled_batch, attn_metadata,
                                         kv_cache_dtype_byte_size)
-
-    def update_mamba_states(self,
-                            attn_metadata: "AttentionMetadata",
-                            num_accepted_tokens: torch.Tensor,
-                            state_indices: Optional[torch.Tensor] = None):
-        MambaCacheManager.update_mamba_states(self, attn_metadata,
-                                              num_accepted_tokens,
-                                              state_indices)
 
 
 def calc_context_stop_positions(prompt_len: int,
--- a/tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
+++ b/tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
@@ -191,12 +191,10 @@ class CppMambaCacheManager(BaseResourceManager):
         self.mamba_impl.free_cache_block(request.py_request_id)
 
     def add_dummy_requests(self, request_ids: List[int], **kwargs):
-        # For CUDA graph dummy requests, the blocks will be allocated
-        # when get_state_indices is called.
-        from .cuda_graph_runner import CUDA_GRAPH_DUMMY_REQUEST_ID
-        request_ids = [
-            rid for rid in request_ids if rid != CUDA_GRAPH_DUMMY_REQUEST_ID
-        ]
+        # Allocate a permanent slot for every id, including CUDA-graph
+        # padding sentinels (matches PythonMambaCacheManager). Padding
+        # entries in get_state_indices then resolve via mCacheIndex to
+        # the sentinel's reserved slot and never alias a live request.
         if request_ids:
             self.mamba_impl.allocate_cache_blocks(request_ids)
 
@@ -428,10 +426,16 @@ class PythonMambaCacheManager(BaseResourceManager):
         self._prepare_mamba_cache_blocks(request_ids)
 
     def add_dummy_requests(self, request_ids: List[int], **kwargs):
-        from .cuda_graph_runner import CUDA_GRAPH_DUMMY_REQUEST_ID
-        request_ids = [
-            rid for rid in request_ids if rid != CUDA_GRAPH_DUMMY_REQUEST_ID
-        ]
+        # Allocate a permanent slot for every dummy request ID, including
+        # the CUDA-graph padding sentinel. Padding entries in a batch all
+        # reference the same dummy request ID, so they share one slot via
+        # mamba_cache_index lookup in get_state_indices. This mirrors how
+        # MTP's per-draft-len padding dummies already behave (they use
+        # CUDA_GRAPH_DUMMY_REQUEST_ID - draft_len, which was never
+        # filtered here) and keeps padding writes off every live
+        # request's slot, even under the overlap scheduler where a prior
+        # batch's completed requests linger in mamba_cache_index until
+        # _process_previous_batch runs.
         if request_ids:
             for r in request_ids:
                 if r not in self.mamba_cache_index:
@@ -448,29 +452,10 @@ class PythonMambaCacheManager(BaseResourceManager):
 
     def get_state_indices(self, request_ids: List[int],
                           is_padding: List[bool]) -> List[int]:
-        assert len(request_ids) == len(is_padding), (
-            "request_ids and is_padding must have the same size")
-
-        used_slots = {
-            self.mamba_cache_index[req_id]
-            for req_id, pad in zip(request_ids, is_padding) if not pad
-        }
-        available_slots = iter(
-            sorted(set(range(self.state_indices.numel())) - used_slots))
-
-        def slot_for(req_id: int, pad: bool):
-            if pad:
-                try:
-                    return next(available_slots)
-                except StopIteration:
-                    raise RuntimeError(
-                        "Run out of available slots for padding") from None
-            return self.mamba_cache_index[req_id]
-
-        result = [
-            slot_for(rid, pad) for rid, pad in zip(request_ids, is_padding)
-        ]
-        return result
+        # Padding entries reuse the slot pre-allocated by their dummy
+        # request in add_dummy_requests; see that method for the
+        # overlap-scheduler rationale.
+        return [self.mamba_cache_index[rid] for rid in request_ids]
 
     def get_conv_states(self, layer_idx: int) -> torch.Tensor:
         layer_offset = self.mamba_layer_offsets[layer_idx]
@@ -530,14 +515,14 @@ class PythonMambaCacheManager(BaseResourceManager):
 
     @torch.compile(options={"max-autotune": True})
     def update_mamba_states(self, attn_metadata: "AttentionMetadata",
-                            num_accepted_tokens: torch.Tensor):
+                            num_accepted_tokens: torch.Tensor,
+                            state_indices: torch.Tensor):
         batch_size = attn_metadata.num_seqs
         num_contexts = attn_metadata.num_contexts
         num_gens = batch_size - num_contexts
         num_accepted_draft_tokens = num_accepted_tokens[
             num_contexts:num_contexts + num_gens] - 1
-        state_indices_d = self.state_indices[num_contexts:num_contexts +
-                                             num_gens]
+        state_indices_d = state_indices[num_contexts:num_contexts + num_gens]
 
         conv_states = self.mamba_cache.conv
         ssm_states = self.mamba_cache.temporal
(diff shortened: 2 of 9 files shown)


cpp/tensorrt_llm/thop/mamba2MTPSSMCacheOp.cpp
```diff
@@ -55,7 +55,11 @@ void mamba2_mtp_ssm_cache_update(th::Tensor ssm, th::Tensor x, th::Tensor dt, th
     int const head_dim = ssm.size(2);
     int const ssm_dim = ssm.size(3);
 
-    TORCH_CHECK(intermediate_states.dim() == 5 && intermediate_states.size(0) == ssm.size(0)
+    // intermediate_states is indexed by intermediate_states_indices (batch
+    // position 0..bs-1), independent of ssm's slot pool; it only needs to
+    // fit the forward batch, which may be smaller than ssm.size(0) when
+    // the cache pool reserves extra slots (e.g. CUDA-graph padding dummies).
```
hnover-nv: I think I agree with the weakeend check but not the comment.
In theory, ssm.size(0) is the number of requests we have memory for, independent of concurrency or what we're working on now.  Maybe we have more active requests than concurrency and rotate them for fairness, or maybe we finished the request but keep the memory around for the equivalent of kv cache re-use in case we get some new request.  (I have no idea what we actually support here).
Meanwhile, we need intermediate state space for each entry in the batch, including dummies, because all the intermediate work has to go somewhere.  But we'll clear it out when update_mamba_states is called.
Except, actually, if there are padding dummies,then in theory the dummies can share a slot (in the intermediate states buffer).  so it doesn't HAVE to be bigger than the batch size, at any given step.
But, presumably we want to support full concurrency (why would w assume there's >1 dummy) and so in general, if we dont' know this requests's composition, the max we'd ever need (if we assign the intermediate slots efficiently) is batch size.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -138,12 +138,10 @@ def free_resources(self, request: LlmRequest):
         self.mamba_impl.free_cache_block(request.py_request_id)
 
     def add_dummy_requests(self, request_ids: List[int], **kwargs):
-        # For CUDA graph dummy requests, the blocks will be allocated
-        # when get_state_indices is called.
-        from .cuda_graph_runner import CUDA_GRAPH_DUMMY_REQUEST_ID
-        request_ids = [
-            rid for rid in request_ids if rid != CUDA_GRAPH_DUMMY_REQUEST_ID
-        ]
+        # Allocate a permanent slot for every id, including CUDA-graph
```
hnover-nv, line 193: maybe outside the scope of your PR, but why do we have different request IDs for different draft lengths?  and yet use the same request ID for the same draft length?  My understanding is if we have two dummies with DL=5, they both get request id CONSTANT-5.  But why, why can't they all get CONSTANT, if they're allowed to share?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -621,10 +606,17 @@ def mamba_layer_cache(
...
+    def update_mamba_states(self,
+                            attn_metadata: "AttentionMetadata",
+                            num_accepted_tokens: torch.Tensor,
+                            state_indices: Optional[torch.Tensor] = None):
         assert not self._use_cpp, "update_mamba_states is not supported in CppMambaCacheManager"
-        self._impl.update_mamba_states(attn_metadata, num_accepted_tokens)
+        # Resolve the forward-path fallback outside the @torch.compile body
+        # so Dynamo only specializes on a concrete Tensor.
+        if state_indices is None:
+            state_indices = self._impl.state_indices
```
hnover-nv: I don't understand this one

tensorrt_llm/_torch/speculative/mtp.py
```diff
@@ -1167,9 +1167,15 @@ def forward(
...
+            # Use the forward-path state_indices so the scatter lines up
+            # with the mixer's reads (covers the full padded batch).
+            mamba_metadata = getattr(attn_metadata, 'mamba_metadata', None)
+            mamba_state_indices = (mamba_metadata.state_indices
+                                   if mamba_metadata is not None else None)
             attn_metadata.kv_cache_manager.update_mamba_states(
                 attn_metadata=attn_metadata,
-                num_accepted_tokens=num_accepted_tokens)
+                num_accepted_tokens=num_accepted_tokens,
+                state_indices=mamba_state_indices)
```
hnover-nv: Is this right?
mixer is assigning slots based on the attn metadata.
so we need update mamba states  to use that same mapping, so we provide the order here.
We don't use the internal state indices any more, unless this is None.  Internal state indices is just used to allocate permanent dummy slots?
Actually, do we use internal state_indices anywhere any more?
And do we ever go down the None path?

hnover-nv (commented):
Some basic questions for you.

cpp/tensorrt_llm/thop/mamba2MTPSSMCacheOp.cpp
Wanli-Jiang: Agreed — rewrote the comment to reflect that framing. New wording:

```cpp
// ssm.size(0) is the Mamba cache capacity — independent of the
// current batch (may include parked requests or reserved dummy
// slots). intermediate_states is per-step scratch indexed by
// intermediate_states_indices in [0, bs), so it only needs to
// fit the forward batch.
```

This matches your points: `ssm.size(0)` is total cache capacity (parked reals, CUDA-graph padding reservations, or any future reuse scheme — not a concurrency number), while `intermediate_states` is per-step scratch that only needs to cover the forward batch.

On the "can dummies share a slot?" side point — yes, in principle padding dummies could share a single intermediate slot since their output is discarded post-verification. We don't rely on that here: the `>= bs` check allows the buffer to be sized for full-concurrency without assuming any dummy-sharing scheme, so the `>= bs` upper bound is the safe contract regardless of how we assign intermediate slots.

Concrete sizing in this PR: `MambaHybridCacheManager` sizes the `ssm` pool to `max_batch_size + max_draft_len + 1` (reserves permanent slots for CUDA-graph padding dummies), while `intermediate_ssm_states` stays sized to `max_batch_size` via `spec_state_size`. So at runtime `ssm.size(0) > intermediate_states.size(0) >= bs` and the old `== ssm.size(0)` check no longer holds — hence the relaxation to `>= bs`.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
Wanli-Jiang: the answer splits into "why same id within a DL" and "why different ids across DLs":

**Same DL → same id falls out automatically.** `cuda_graph_runner.py:481` does:

```python
batch.generation_requests.extend([padding_dummy_request] * padding_size)
```

That's Python list replication: all `N` padding positions hold the same `LlmRequest` object, so they trivially share one `request_id`. We didn't design "they get the same id"; we designed "there's one dummy object per DL, referenced `N` times."

**Different DLs → different ids because the LlmRequest objects must differ.** Each dummy is built in `resource_manager.py:2231` with DL-specific fields that must match what the CUDA graph was captured against:

* KV cache capacity: `token_num + max_num_draft_tokens + 1`
* `py_draft_tokens = [1] * max_num_draft_tokens` (length = DL)
* `prompt_len = token_num - 1` where `token_num = 1 + max_num_draft_tokens`

So a DL=5 captured graph needs an `LlmRequest` with a length-5 `py_draft_tokens`, KV sized for 5 drafts, etc. You can't use a DL=3 dummy in its place — the shape-affecting fields would mismatch what the graph was captured with. The natural cache is `self.padding_dummy_requests[runtime_draft_len]` at `cuda_graph_runner.py:450` — one object per DL, created on first use, reused forever.

**"Couldn't they all be CONSTANT?"** `request_id` is the system-wide key into every per-request resource map: paged KV blocks, `SlotManager.slot_mapping`, `mamba_cache_index`, spec-slot allocator. Two distinct `LlmRequest` objects sharing one id would collide — e.g. `SlotManager.add_slot` at `resource_manager.py:2689-2701` treats an already-mapped id as idempotent re-registration and returns the existing slot, so the second dummy would alias the first's slot. Same story for `mamba_cache_index`. So once you've accepted "one `LlmRequest` per DL, shaped for that DL", you also need one id per DL. `CONSTANT - runtime_draft_len` is just a convenient encoding of the DL into a reserved-sentinel range.

The only alternative that keeps a single `CONSTANT` id would be: size one dummy to `max_draft_len` and reuse it for smaller DLs. That wastes KV for smaller DLs and still requires separate captured graphs per shape, so it's strictly worse.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
Wanli-Jiang: that comment was trying to explain a structural hack that's now gone. The fallback has been removed entirely in commit `5239b11ac`.

What it used to do: `MambaCacheManager.update_mamba_states` took `state_indices: Optional[Tensor] = None`, and when `None` it fell back to `self._impl.state_indices` (a tensor the Python cache manager kept in sync with `mamba_cache_index`). The resolution was done outside the `@torch.compile`-decorated `PythonMambaCacheManager.update_mamba_states` inner function because:

1. Dynamo specializes the compiled graph on input types. Passing `Optional[Tensor]` in would create a separate specialization for the `None` branch, which itself traces through `self._impl.state_indices` (an attribute read that Dynamo guards on tensor identity — fragile if the tensor ever gets reassigned). Resolving outside kept the compiled inner's arg a concrete `Tensor`, one specialization, no attribute guards.
2. AutoDeploy was the only user of the `None` branch — it didn't have a `Mamba2Metadata.state_indices` tensor, so it relied on the cache manager's mirror.

What changed in the follow-up: both callers now pass an explicit `state_indices` tensor.
* MTP path: `attn_metadata.mamba_metadata.state_indices` (already `int32`, fixed shape).
* AutoDeploy eagle: `csi.get_arg("slot_idx", truncate=True)` (`int64`, AutoDeploy's per-step slot tensor).

With both callers covered, I deleted the `Optional[Tensor]` default, the `None` branch, and the `PythonMambaCacheManager.state_indices` tensor itself — it had no remaining readers and its per-step `copy_(torch.tensor(..., pin_memory=True), non_blocking=True)` write in `_prepare_mamba_cache_blocks` was pure overhead. The `mamba_cache_index` dict is now the sole source of truth; per-step `state_indices` tensors are built by the callers from it.

tensorrt_llm/_torch/speculative/mtp.py
Wanli-Jiang: Yes, your understanding is right — mixer assigns slots via `mamba_metadata.state_indices` during forward; `update_mamba_states` must scatter into the same slots by batch position, so we pass that tensor here. Concrete answers to each question:

**"We don't use the internal state indices any more, unless this is None. Internal state indices is just used to allocate permanent dummy slots?"**
Slot allocation itself runs on `self.mamba_cache_index` (the dict), not on `self.state_indices` (the tensor). The tensor was just a mirror that `_prepare_mamba_cache_blocks` kept in sync with the dict for the `None` fallback. Dummies go into the dict like every other id.

**"Do we use internal state_indices anywhere any more?"**
No. After removing the `None` branch, the tensor had no remaining readers. Commit `5239b11ac` deletes it end-to-end: the attribute allocation in `__init__`, the `_prepare_mamba_cache_blocks` write (an `torch.tensor(..., pin_memory=True) + copy_(..., non_blocking=True)` on every scheduler step), the `shutdown()` clear, and the `state_indices_list` helper. Net win: one pinned-host allocation and one H2D copy removed per step. `mamba_cache_index` (the dict) is now the sole source of truth.

**"Do we ever go down the None path?"**
Used to: AutoDeploy at `modeling_eagle.py:1029` called `update_mamba_states` without a `state_indices` kwarg, hitting the fallback. In the follow-up, AutoDeploy now passes `csi.get_arg("slot_idx", truncate=True)` explicitly — the same `int64` slot tensor its

tests/unittest/_torch/executor/test_mamba_cache_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    CppMambaCacheManager.add_dummy_requests(stub, [100, CUDA_GRAPH_DUMMY_REQUEST_ID, 101])
+    stub.mamba_impl.allocate_cache_blocks.assert_called_once_with(
+        [100, CUDA_GRAPH_DUMMY_REQUEST_ID, 101]
+    )
+
+
+def test_cpp_add_dummy_requests_noop_on_empty_list():
+    stub = SimpleNamespace(mamba_impl=MagicMock())
+    CppMambaCacheManager.add_dummy_requests(stub, [])
+    stub.mamba_impl.allocate_cache_blocks.assert_not_called()
```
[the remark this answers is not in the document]
Wanli-Jiang: added

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
hnover-nv: Ah, ok.  I got the sense from your PR before it included the slot fix that we care a lot about wasted slots though.  Here we're now allocating DL+1 extras mamba state slots (but I think, at least, not (DL+ 1 for dummies ) * (DL+1 for MTP) by also expanding the intermediate ssm?  we'd have to trace the constructor to be sure).

But I don't understand the argument against reusing a single constant.  we're ok with different dummies aliasing the same slots.  If it's idempotent we won't duplicate resources.  Having just one dummy that supports up to DL vs one for no DL, one for DL+1, one for DL+2, etc., that seems better?  Especially as the DL we pass in is (for now) fixed for the run, and even when we vary it in upcoming work we'd expect to use that maxed value sometimes.  Or is the problem we'd end up trying to force that max DL dummy in with lesser DL requests?  I think our code doesn't even support mixing DL lengths.  

But if we had multiple dummy requests but they shared a request_id, then that seems like it could be ok, with the idempotency?  Or are you worried we run it through as DL=0 first, and the returned resources are too small, so when we rerun with DL=5, we don't get bigger resources but instead dont' have enough and crash or write out of memory or something?

Even if that's all true, we could special case the dummy requests for mamba and reserve a specific mamba SLOT, that all the dummy requests map to.  so we only ever waste one mamba slot, which is the expensive thing.

Not that we have to fold that all in to this PR, but if we're trying to minimize wasted wasted slots and also we might push DL in the future, it could matter soon.

tensorrt_llm/_torch/speculative/mtp.py
hnover-nv: Ok, so to be really clear, we've got two mappings to different types of slots.
There's a request_id to mamba cache slot, which is the dict in mamba_cache_index.  Persistent until we drop requests.
Then there's a map from _something_ to intermediate ssm state.  That's currently done by state_indices, positionally, which says intermediate[i] is associated to slot state_indices[i].  Which is also probably the position in the batch, for some canonical "requests in the batch" tensor?

hnover-nv (approved):

govind-ramnarayan (approved):
LGTM from the AutoDeploy side, thanks!

Tabrizian (approved):

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
Wanli-Jiang: Good points. Three pieces:

**On `intermediate_ssm` sizing** — your guess is correct, no compounding. `intermediate_ssm_states.shape = (num_local_layers, spec_state_size, max_draft_len+1, nheads, head_dim, d_state)` where `spec_state_size = max_batch_size` (the user-facing one, not `pool_size`). See `mamba_cache_manager.py:270` and the 8th positional arg to `MambaCacheManager.__init__` at line 659. So this PR's `+ max_draft_len + 1` bump only applies to `ssm_states` / `conv_states`, not intermediate buffers.

**On "why not a single `CONSTANT` id?"** — you pinned the exact issue in your last paragraph: "we run it through as DL=0 first, resources too small, then DL=5 …". Yes, that's the trap. Every call to `add_dummy_requests` creates a new `LlmRequest` with DL-specific sizing (`resource_manager.py:2345-2398`):
* `input_tokens = [1] * (1 + max_num_draft_tokens)`
* `py_draft_tokens = [1] * max_num_draft_tokens`
* KV cache sized to `token_num + num_extra_kv_tokens + num_extra_decoding_steps`.

If two calls share `request_id=CONSTANT`, the second one's `_create_kv_cache(CONSTANT, ...)` collides with the first's already-registered cache — the second caller inherits the first's wrong-size resources. For a DL=5 graph that expects length-5 `py_draft_tokens` and DL=5-sized KV, getting back the DL=0 resources → OOB reads / writes. So `request_id` has to encode DL because it's the key into per-request resource maps (paged KV blocks, `SlotManager.slot_mapping`, spec-slot allocator, `mamba_cache_index`), and those maps need different entries for differently-sized resources.

You're right that today's code doesn't mix DLs within a single step, but `padding_dummy_requests` is a cache across calls (populated lazily, one entry per DL ever seen). Warmup / CUDA-graph capture can make multiple DLs show up over a run, and the dict retains them.

**On your "special-case Mamba" idea** — this is the good path, and I think it's worth doing (as a follow-up). Keep distinct `LlmRequest` objects per DL (so KV / spec resources stay correctly sized with their own keys), but make `mamba_cache_index` map all `CONSTANT - k` ids to one shared reserved slot. Writes from any DL's padding position land on that shared slot — harmless since padding outputs are discarded.

Rough shape of the change:

```python
# PythonMambaCacheManager.add_dummy_requests
for r in request_ids:
    if r in self.mamba_cache_index:
        continue
    if self._is_padding_sentinel(r):                 # CONSTANT - k for k in 0..max_draft_len
        if self._padding_slot is None:
            self._padding_slot = self.mamba_cache_free_blocks.pop()
        self.mamba_cache_index[r] = self._padding_slot
    else:
        self.mamba_cache_index[r] = self.mamba_cache_free_blocks.pop()
```

Plus the mirror in `CppMambaCacheManager` / `RnnStateManager::allocateCacheBlocks`, and the pool bump drops from `+ max_draft_len + 1` to `+ 1` in `MambaHybridCacheManager.__init__`. For `max_draft_len=3`, that saves 3 Mamba slots per Mamba layer on every rank — small absolute, but scales with `max_draft_len` which is the direction we're heading.

Wanli-Jiang: filed a ticket to track it: https://jirasw.nvidia.com/browse/TRTLLM-12209

tensorrt_llm/_torch/speculative/mtp.py
Wanli-Jiang: Yes — two conceptually distinct mappings, linked by batch position. Let me walk through concrete source and a worked example.

### The two mappings

**(1) `mamba_cache_index`: `request_id` → persistent Mamba slot**

Dict keyed by `py_request_id`, populated on allocation, freed on request completion. The slot indexes into `conv_states` / `ssm_states` (shape dim 1 = `pool_size`).

```python
# mamba_cache_manager.py:316
self.mamba_cache_index: Dict[int, int] = {}

# mamba_cache_manager.py:337-342 (_prepare_mamba_cache_blocks)
if r in self.mamba_cache_index:
    continue
if len(self.mamba_cache_free_blocks) == 0:
    raise RuntimeError("run out of mamba cache blocks")
block = self.mamba_cache_free_blocks.pop()
self.mamba_cache_index[r] = block

# mamba_cache_manager.py:375-376 (free_resources)
if request_id in self.mamba_cache_index:
    block = self.mamba_cache_index.pop(request_id)
    self.mamba_cache_free_blocks.append(block)
```

Persistence: alive as long as the request is alive.

**(2) `intermediate_state_indices`: batch position → intermediate-buffer position**

A plain `arange`, no per-request logic:

```python
# mamba_cache_manager.py:319
self.intermediate_state_indices = torch.arange(max_batch_size,
                                               dtype=torch.int32,
                                               device=device)
```

The intermediate buffers themselves:

```python
# mamba_cache_manager.py:269-282 (speculative branch of __init__)
intermediate_ssm_states = torch.zeros(
    size=(num_local_layers, self.spec_state_size,
          speculative_num_draft_tokens + 1) + ssm_state_shape, ...)
intermediate_conv_window_cache = torch.zeros(
    size=(num_local_layers, self.spec_state_size,
          speculative_num_draft_tokens + 1) + conv_state_shape, ...)
```

Shape: `(num_layers, spec_state_size=max_batch_size, max_draft_len+1, ...)`. Dim 1 is batch-position-indexed, not slot-indexed. Re-used every step — no persistence.

### Who writes / who reads

**Mixer writes: positional**

```python
# mamba2_mixer.py:411-413 (is_target_verify branch)
intermediate_state_indices = _cached_arange(
    attn_metadata.kv_cache_manager.get_max_resource_count(),
    state_indices_d.device)[:num_decodes]  # [0, 1, ..., num_decodes-1]

# mamba2_mixer.py:428-429 — conv write
intermediate_conv_window=intermediate_conv_states,
intermediate_state_indices=intermediate_state_indices,

# mamba2_mixer.py:543-545 — ssm write
intermediate_states_buffer=intermediate_ssm_states,
...
intermediate_state_indices=intermediate_state_indices,
```

Mixer writes `intermediate_ssm[:, i, t, :]` for each gen batch position `i ∈ [0, num_decodes)` and each draft step `t ∈ [0, draft_token_num)`. The position `i` is within the gen partition of the captured batch.

**`update_mamba_states` reads positional, writes by slot**

```python
# mamba_cache_manager.py:440-465 (PythonMambaCacheManager.update_mamba_states)
batch_size = attn_metadata.num_seqs
num_contexts = attn_metadata.num_contexts
num_gens = batch_size - num_contexts
num_accepted_draft_tokens = num_accepted_tokens[
    num_contexts:num_contexts + num_gens] - 1               # per-request draft step
state_indices_d = state_indices[num_contexts:num_contexts + num_gens]  # per-position slot

src_state_indices = self.intermediate_state_indices[:num_gens]  # [0, 1, ..., num_gens-1]

accepted_ssm_state = intermediate_state_cache[:, src_state_indices,
                                              num_accepted_draft_tokens]
ssm_states[:, state_indices_d, :] = accepted_ssm_state

accepted_conv_state = intermediate_conv_window_cache[:,
                                                     src_state_indices,
                                                     num_accepted_draft_tokens]
conv_states[:, state_indices_d, :] = accepted_conv_state
```

For each batch position `i ∈ [0, num_gens)`:
* Read `intermediate_ssm[layer, i, accept_idx[i], :]` — partial state for batch-pos `i` at its accepted draft step
* Write `ssm_states[layer, state_indices_d[i], :]` — the permanent slot owned by that request

### Worked example

Setup: 3 real gens + 1 CUDA-graph padding dummy, `max_draft_len=3`, all IDs already allocated:

| Batch pos | `request_id` | `mamba_cache_index[req]` |
| :--- | :--- | :--- |
| 0 (ctx) | `100` | `3` |
| 1 (gen) | `101` | `5` |
| 2 (gen) | `102` | `7` |
| 3 (gen) | `SENTINEL` | `11` |

`Mamba2Metadata.state_indices = [3, 5, 7, 11]` (partition order, via `kv_cache_manager.get_state_indices(request_ids, is_padding)` — dict lookup).

**Forward (mixer) with `num_contexts=1`, `num_gens=3`:**
* `intermediate_state_indices = [0, 1, 2]` (just `arange(3)`)
* At gen batch pos 0 (= req 101): mixer writes `intermediate_ssm[:, 0, t, :]` for each draft step `t=0..3`
* At gen batch pos 1 (= req 102): writes `intermediate_ssm[:, 1, t, :]`
* At gen batch pos 2 (= padding): writes `intermediate_ssm[:, 2, t, :]` (discarded)

After verification, say `num_accepted_tokens = [1, 3, 2, 1]` (ctx=1, then per-gen accepted counts).
`num_accepted_draft_tokens = [2, 1, 0]` (= `num_accepted[1:4] - 1`).

**`update_mamba_states`:**
* `state_indices_d = state_indices[1:4] = [5, 7, 11]`
* `src_state_indices = [0, 1, 2]`
* **Read:** `intermediate_ssm[:, [0,1,2], [2,1,0], :]` → 3 partial states, one per gen position, each at its accepted draft step
* **Write:** `ssm_states[:, [5, 7, 11], :] = <those 3 states>`

**Result:**
* Slot 5 (= req 101's slot): updated with req 101's state at draft step 2
* Slot 7 (= req 102's slot): updated with state at draft step 1
* Slot 11 (padding): written with dummy's state at draft step 0 — harmless, dummy's output isn't used

### Why two layers of indirection

| | Intermediate buffer | Final ssm/conv |
| :--- | :--- | :--- |
| **Lifetime** | Step-scoped (overwritten each step) | Request-scoped (alive across steps) |
| **Indexing** | Positional (batch pos in gen partition) | Slot-keyed (from `mamba_cache_index`) |
| **Size (dim 1)** | `max_batch_size` | `pool_size = max_batch_size + max_draft_len + 1` |
| **Written by** | Mixer `is_target_verify` branch | `update_mamba_states` after verify |

The intermediate buffer is cheap because it's positional — no tracking. The final cache is request-keyed because conv/ssm states have temporal dependencies across steps (each step's output depends on the previous step's final state for that request), so we need a stable per-request slot.

### Bridging them: `state_indices`

`state_indices` (= `mamba_metadata.state_indices`) is the bridge — batch-position → slot mapping. Both the mixer (during forward) and `update_mamba_states` (after verify) read the same tensor, so positional writes to `intermediate_ssm` and slot-keyed writes to `ssm_states` agree on which request is at which batch position.

Constructed by `Mamba2Metadata.prepare` → `kv_cache_manager.get_state_indices(request_ids, is_padding)` which is just a dict lookup on `mamba_cache_index`:

```python
# mamba_cache_manager.py:384
return [self.mamba_cache_index[rid] for rid in request_ids]
```

`request_ids` here is partition-ordered (`[context..., extend_real, extend_dummy, first_draft, gen...]`), so `state_indices[i]` = slot owned by request at batch position `i`.

### Consistency contract

The only correctness requirement: the `state_indices` tensor the mixer reads during forward `==` the `state_indices` tensor `update_mamba_states` uses for the scatter. Before this PR, the mixer read `mamba_metadata.state_indices` but `update_mamba_states` read `self._impl.state_indices` (a different tensor, possibly with stale / differently-ordered entries). This PR unifies both reads on `mamba_metadata.state_indices` — see `mtp.py:1175` (`state_indices=attn_metadata.mamba_metadata.state_indices`) and `modeling_eagle.py:1040` (`csi.get_arg("slot_idx", truncate=True)`), which is AutoDeploy's equivalent-order tensor.

zheyuf (approved):

nv-guomingz (approved):

Wanli-Jiang:
/bot run --stage-list "DGX_B200-4_GPUs-PyTorch-2" --disable-fail-fast

Wanli-Jiang:
/bot run --stage-list "DGX_B200-4_GPUs-PyTorch-2" --disable-fail-fast

Wanli-Jiang:
/bot run --stage-list "DGX_B200-4_GPUs-PyTorch-2" --disable-fail-fast

Wanli-Jiang:
/bot skip --comment "tests are passed at this PR and PR13489"

Issue #11625, opened 2026-02-22
[Bug]: trtllm-serve crashes with std::bad_cast when eos_token_id is a list (Llama 3.1)


CristyNel:
### System Info

* GPU: NVIDIA GeForce RTX 5090 Blackwell (SM 120)
* CPU: Intel i9-14900K Raptor Lake-S Refresh (32) @ 5.700GHz
* RAM: TeamGroup Delta RGB 64GB DDR5-7600 7600MHz Dual Channel
* OS: Ubuntu 24.04.3 LTS x86_64
* Kernel: 6.14.0-37-generic
* Docker Image: nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc2
* Python Version: 3.12.3
* CUDA Version: 12.8

### Who can help?

@2ez4bz
@yuanjingx87
@karljang
@greg-kwasniewski1
@Wanli-Jiang
@kaiyux
@nv-guomingz
@nekorobov
@Funatiq
@byshiue

### Information

- [x] The official example scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)

### Reproduction

```bash
# Build the engine for Llama 3.1 (which uses a list for eos_token_id in generation_config.json)

trtllm-build \
  --checkpoint_dir /models/llama3_8b_legacy \
  --output_dir /models/llama3_8b_base_engine \
  --max_batch_size 1 \
  --max_seq_len 3000 \
  --max_num_tokens 6000

#  Start the FastAPI server using the C++ Executor backend

trtllm-serve /models/llama3_8b_base_engine \
  --backend tensorrt \
  --tokenizer /models/Llama-3.1-8B-Instruct \
  --host 0.0.0.0 \
  --port 8000

#  Send a basic completions request (in a separate terminal)

curl -sS <http://127.0.0.1:8000/v1/completions> \
  -H "Content-Type: application/json" \
  -d '{
    "model": "llama3_8b_base_engine",
    "prompt": "Say hello.",
    "max_tokens": 10
  }'
```

### Expected behavior

The `trtllm-serve` wrapper and underlying C++ Executor API should natively parse and accept an array/list of integers for `eos_token_id` (and `pad_token_id`) directly from the Hugging Face `generation_config.json`, correctly mapping them to multiple stop conditions without throwing a C++ exception.

### Actual behavior

[...]

PR #13178, opened 2026-04-18
[#11625][fix] handle list-typed eos_token_id in SamplingParams._setup

edenfunf:
## What

`SamplingParams._setup()` crashed `trtllm-serve` with `std::bad_cast` whenever
the tokenizer's `eos_token_id` is a list instead of a scalar int.

**Before:**
```python
# sampling_params.py – _setup()
self.end_id = tokenizer.eos_token_id   # could be a list → C++ crash
```

**After:**
```python
eos_token_id = tokenizer.eos_token_id
if isinstance(eos_token_id, list):
    self.end_id = eos_token_id[0] if eos_token_id else None
    # register remaining tokens so generation still halts on all of them
    for tok in eos_token_id[1:]:
        if tok not in (self.stop_token_ids or []):
            self.stop_token_ids = (self.stop_token_ids or []) + [tok]
else:
    self.end_id = eos_token_id
```

## Why

HuggingFace tokenizers for models like **Llama 3.1** expose
`eos_token_id = [128001, 128009]` (a `List[int]`).
`_setup()` blindly assigned this list to `self.end_id`, which is typed
`Optional[int]`.  Downstream, `base_worker.py` passes `end_id` directly into
`tllm.Request(end_id=...)` — a C++ nanobind binding that expects `int32_t`.
PyBind/nanobind cannot cast a Python `list` to `int32_t`, so it raises
`std::bad_cast`, killing the server for every inference request on
any model whose tokenizer advertises multiple EOS tokens.

### Why is this PR open if issue #11625 is closed?

Issue #11625 was closed by a maintainer who was **unable to reproduce** the
crash on their own hardware (`"I'm closing this issue. Please feel free to open
a new one if you need~"`).  It was **not** closed because a fix was merged.
The original reporter confirmed the crash persisted after the closure.
`upstream/main` still contains the unfixed assignment as of this PR.

The root cause is version-dependent: newer versions of `transformers`
(≥ 4.43) began returning `List[int]` from `tokenizer.eos_token_id` for
models whose `tokenizer_config.json` has a list-valued `eos_token_id`.
Earlier versions always returned a scalar, which is why the bug was not
universally reproducible.

## How

- Read `tokenizer.eos_token_id` into a local variable first.
- If it is a `list`, use `eos_token_id[0]` as the scalar `end_id` and
  append `eos_token_id[1:]` to `stop_token_ids` so that generation
  terminates on **all** configured EOS tokens, not just the first one.
- If it is already a scalar, behaviour is identical to before.
- The existing `generation_config.eos_token_id` list-normalisation logic
  (lines 436-447) is untouched; de-duplication between the two paths is
  handled by the existing `stop_token not in self.stop_token_ids` guard.

Change is **minimal and isolated** to the 14 new lines inside `_setup()`.
No other call sites, no architecture changes.

## Test

**Before fix (reproduction):**
```python
from tensorrt_llm.sampling_params import SamplingParams

class FakeTokenizer:
    eos_token_id = [128001, 128009]  # Llama 3.1 style
    pad_token_id = 0

params = SamplingParams()
params._setup(FakeTokenizer(), None, None)
# self.end_id is now [128001, 128009] → std::bad_cast when passed to C++
```

**After fix:** `params.end_id == 128001` (int), `params.stop_token_ids == [128009]`.

**Regression tests added** in `tests/unittest/llmapi/test_sampling_params.py`
(pure Python, no GPU or model weights required):

| Test | Covers |
|------|--------|
| `test_setup_scalar_eos_token_id` | scalar path unchanged |
| `test_setup_list_eos_token_id_single_element` | single-element list → scalar |
| `test_setup_list_eos_token_id_multiple_elements` | multi-element list, `end_id` is `int` |
| `test_setup_list_eos_token_id_no_duplicate_stop_tokens` | no duplicate stop tokens |
| `test_setup_end_id_already_set_ignores_tokenizer_eos` | explicit `end_id` not overwritten |
| `test_setup_list_eos_token_id_pad_fallback` | `pad_id` fallback still an int |

All pre-commit hooks passed (CRLF, ruff, ruff-format, codespell, DCO).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated `SamplingParams._setup()` to support scalar and list-valued `tokenizer.eos_token_id`.
  * Uses the first EOS token as `end_id`.
  * Adds remaining EOS tokens to `stop_token_ids` without duplicates.
  * Preserves explicit `end_id` values and falls back to `end_id` for `pad_id` when needed.
  * Prevents the `std::bad_cast` crash caused by list-valued EOS IDs.

* **Tests**
  * Added unit tests for scalar, single-element, and multi-element EOS token IDs.
  * Added tests for duplicate stop-token handling, explicit `end_id`, and `pad_id` fallback.

## Dev Engineer Review

* The `_setup()` change is isolated and preserves existing scalar behavior.
* List handling avoids duplicate `stop_token_ids` entries.
* Empty EOS lists set `end_id` to `None`.
* No public API declarations, configuration files, or test-list files changed.
* No performance or error-handling concerns were identified.

## QA Engineer Review

* Added test functions in `tests/unittest/llmapi/test_sampling_params.py` for:
  * Scalar EOS IDs.
  * Single-element EOS ID lists.
  * Multi-element EOS ID lists.
  * Duplicate stop-token handling.
  * Explicit `end_id`.
  * `pad_id` fallback.
* These tests are not listed in `tests/integration/test_lists/` for `test-db/` CI or `qa/` manual QA.
* Verdict: needs follow-up.


karljang:
Thanks for the thorough writeup!
However, I'd suggest holding off on this PR until the premise can be demonstrated, specifically, a case where `tokenizer.eos_token_id` is a list without a hand-written stub standing in for the tokenizer. If that turns up, I'm happy to revisit; the change itself is small and clean.

I have measured on unpatched main branch with `transformers==5.5.4` (the pin at `requirements.txt:35`):

```
tokenizer.eos_token_id           128009                      int
generation_config.eos_token_id   [128001, 128008, 128009]    list
--------------------------------------------------------------------
after _setup: end_id             128009
after _setup: stop_token_ids     [128001, 128008]
```

Additionally, the existing generation_config block at sampling_params.py:614-628 already produces the outcome this PR wants.

PR #13198, opened 2026-04-20
[None][feat] Add hit-rate gate and fair-share cap to KV-aware ADP router

lancelly:
## Description

Tunes `KVCacheAwareADPRouter` so cache affinity and load balance work together instead of fighting each other. Three behaviour changes, one config surface change, and a small amount of cleanup.

### 1. Hit-rate gate — `kv_cache_routing_match_rate_threshold` (default `0.1`)
For each request, `match_len` contributes to scoring only when `max(match_len) / request_tokens` across eligible ranks is strictly above the threshold; below it, `match_len` is forced to `0` and routing is driven purely by load. This prevents a small universal prefix (e.g. a shared system prompt) from pinning all traffic to the first warm ranks. Set to `0.0` to honour any nonzero match.

### 2. Fair-share cap — `kv_cache_routing_fair_share_multiplier` (default `2.0`)
Per-rank active-request cap expressed as a multiplier of the ceil fair-share, i.e. `fair_share_multiplier * ceil((total_active + new) / tp_size)`. Once a rank hits the cap within a scheduling batch it is dropped from the eligible set for the rest of that batch. `2.0` leaves enough slack for affinity to win in the common case while preventing runaway concentration; set to `1.0` for strict fair share.

### 3. Transfer-in-progress load accounting
Requests mid-KV-transfer to GEN are no longer visible in `active_requests`, so the router used to under-count load on the rank that is sending. The router now pulls `requests_in_transfer()` from the PyExecutor's `AsyncTransferManager` and folds those requests into both `num_active_requests` and `num_active_tokens` when building each rank's `RankState`.

## Release Notes

* **New Features**
  * Added two configuration parameters to control KV cache-aware routing behavior: `kv_cache_routing_match_rate_threshold` and `kv_cache_routing_fair_share_multiplier`.

* **Improvements**
  * Optimized request routing algorithm to better utilize KV cache affinity while maintaining balanced load distribution across distributed ranks.
  * Improved handling of asynchronous request transfers during routing initialization.

### DIFF
--- a/tensorrt_llm/llmapi/llm_args.py
+++ b/tensorrt_llm/llmapi/llm_args.py
@@ -621,6 +621,29 @@ class AttentionDpConfig(StrictBaseModel):
         "Weight (beta) for the load-balance term in KV cache-aware routing. "
         "Higher values prioritize load balance over cache affinity. "
         "Only used when enable_kv_cache_aware_routing is True.")
+    kv_cache_routing_match_rate_threshold: float = Field(
+        default=0.1,
+        description=
+        "Cache-affinity gate in KV cache-aware routing. For each request, "
+        "match_len contributes to scoring only when max(match_len) / "
+        "request_tokens across eligible ranks is strictly above this "
+        "threshold; otherwise match_len is forced to 0 so routing is driven "
+        "purely by load. Default 0.1 requires at least a 10% hit rate before "
+        "cache affinity kicks in, which prevents a small universal prefix "
+        "(e.g. a shared system prompt) from pinning all traffic to the "
+        "first warm ranks. Set to 0.0 to honour any nonzero match. "
+        "Only used when enable_kv_cache_aware_routing is True.")
+    kv_cache_routing_fair_share_multiplier: float = Field(
+        default=2.0,
+        description=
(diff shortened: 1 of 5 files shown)


lishicheng1996-nv (approved):
LGTM, it works great!

tensorrt_llm/llmapi/llm_args.py
QiJune, line 625: Should we set kv_cache_routing_match_rate_threshold to 0.0, and kv_cache_routing_fair_share_multiplier to 1.0? So, the main's routing policy is preserved.

QiJune (approved):
LGTM

tensorrt_llm/llmapi/llm_args.py
lancelly: This routing policy is default off, so its default parameter will not affect the main's routing policy?

lancelly: The values of kv_cache_routing_match_rate_threshold and kv_cache_routing_fair_share_multiplier are indeed hard to decide... I chose the default values that work for AA dataset.

PR #13221, opened 2026-04-20
[None][feat] Add per-rank iteration stats to /metrics endpoint

lishicheng1996-nv:
Previously, /metrics and get_stats_async only returned rank-0's iteration stats. Under attention-DP or any multi-rank PyTorch-executor deployment, each rank has its own KV cache / scheduler / batch state and their utilization can diverge, but only rank 0 was observable.

This change makes PyExecutor._append_iter_stats collectively gather each rank's IterationStats (+ RequestStats, KV cache iter stats) via a tp_allgather. Rank 0 serializes the gathered results to JSON dicts tagged with "rank": N and stores them alongside its local stats so the same existing /metrics transport (RPC -> _iter_stats_result queue -> JSON list) returns one entry per (iteration, rank).

The gather is opt-in via the `TLLM_METRICS_ALL_RANKS=1` environment variable and only runs when `tp_size > 1` and `enable_attention_dp=True`. Default behavior (env unset or `!=1`) is byte-identical to upstream: only rank 0's stats are exported. Non-leader ranks drop the gathered result (they don't export).

base_worker._stats_serializer gains a fast-path: if the buffer entry is the new ("per_rank_dict", {...}) tuple, emit its JSON directly instead of calling to_json_str() on the already-serialized dict.

## Usage

Opt in on every worker process (e.g. via `trtllm-serve` env) in an attention-DP deployment:

```bash
TLLM_METRICS_ALL_RANKS=1 trtllm-serve <model> --tp_size 8 --ep_size 8 \
    --extra_llm_api_options extra_llm_api_options.yaml ...
```

with `enable_attention_dp: true` and `enable_iter_perf_stats: true` in the server config. Then each `/metrics` JSON entry carries a new `"rank": N` field; a single iteration produces N entries (one per rank). Group by `iter` for cross-rank snapshots, or filter by `rank` for per-rank time series.

Under pure TP (no attn-DP) the gather is skipped regardless of the env var — every rank runs the same requests on the same iteration, so per-rank stats would be redundant and the allgather would only add a CPU-GPU sync on the hot path.

## Overhead verification

image

## Release Notes

* **Performance Monitoring**
  * Enhanced per-rank iteration performance statistics collection for distributed tensor parallel setups when multi-GPU deployment is enabled.
  * Updated statistics serialization format to support per-rank metrics collection and storage.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -1351,7 +1351,88 @@ class PyExecutor:
     def _append_iter_stats(self,
                            stats: IterationStats,
                            req_stats: Optional[List[RequestStats]] = None):
+        """Append one iteration's stats to the export buffer.
 
+        Under attention-DP with ``tp_size > 1`` each rank has diverging
+        scheduler/KV-cache state. When the environment variable
+        ``TLLM_METRICS_ALL_RANKS=1`` is set, this method collectively
+        gathers every rank's :class:`IterationStats` (+ optional per-request
+        stats and KV iteration stats) via :meth:`tp_allgather` and rank 0
+        stores one pre-serialized JSON dict per rank (tagged with
+        ``"rank"``) in ``self.stats`` so ``/metrics`` can export stats from
+        every rank. Non-leader ranks drop the gathered result.
+
+        The env var defaults to off so upstream behavior (rank-0-only
+        export) is preserved unless users explicitly opt in. Under pure TP
+        (no attention-DP) every rank runs the same requests on the same
+        iteration, so the gather would be redundant and only adds a CPU-GPU
+        sync on the hot path — we fall through to the legacy rank-0-only
+        append in that case regardless of the env var.
+
+        Args:
+            stats: Iteration-level stats from the local rank.
+            req_stats: Optional per-request stats from the local rank.
+        """
+
+        tp_size = getattr(self.dist, "tp_size", 1)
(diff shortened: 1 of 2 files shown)


eopXD (commented):
Overall looks good.

Some questions:
- We we need to guard the read of iteration stats here safely with a lock?
- Is there a way avoid enumerating all fields under `kvCacheIterationStats` here?

eopXD (approved):
No other blocker to me otherwise. The comments are nit. Please evaluate and see if you can address them.

lishicheng1996-nv:
Hi eop. Thanks for the review!
For the first question, Claude evaluate that the lock here is reasonable and correct. As quoted


> What does need a lock — and is already correct
> 
>   self.stats is the cross-thread hand-off buffer: the executor main-loop thread writes to it, and the RPC handler thread reads-and-clears it via get_stats_latest. That's why stats_lock exists.
>
>   Both code paths — the new per-rank gathered path and the legacy rank-0-only path — hold stats_lock while touching self.stats, which is what eop's "safe lock" concern is really about. The reads of stats / req_stats / _latest_kv_iter_stats outside that critical section are local / single-threaded and don't need additional guarding.

   
For the second question, we didn't come up with a lightweight method to de-duplicate the enum. It either need another enum in C++ binding, or need to use methods like X-macro.

lancelly (approved):
LGTM

venkywonka (approved):

PR #13246, opened 2026-04-20
[None][feat] add Gemma 4 reasoning and tool-call parsers

suyoggupta:
## Summary

- Register a `Gemma4ReasoningParser` that splits on the `<|channel>` / `<channel|>` special-token pair used by Gemma 4's thought channel. Streaming holds back partial delimiters and `finish()` flushes buffered trailing text.
- Add a `Gemma4ToolParser` at `tensorrt_llm/serve/tool_parser/gemma4_parser.py` that parses `<|tool_call>call:NAME{...}<tool_call|>` blocks. Arguments use the Gemma 4 custom format (`<|"|>`-delimited strings, bare identifier keys, standard JSON for numbers/bools/arrays/nested objects); a string-aware scanner converts them to standard JSON.
- Set `needs_raw_special_tokens = True` on the tool parser so the server disables `skip_special_tokens`, preserving the delimiter tokens in the decoded stream.
- Wire both parsers into the auto-resolve mappings (`MODEL_TYPE_TO_REASONING_PARSER`, `MODEL_TYPE_TO_TOOL_PARSER`) keyed on `model_type="gemma4"`.

## Test plan

- [x] `pytest tests/unittest/llmapi/test_reasoning_parser.py` — 10 new reasoning-parser cases (one-shot, streaming with split delimiters, `finish()` flush, interleaved channels, auto-detect from `config.json`).
- [x] `pytest tests/unittest/llmapi/apps/test_tool_parsers.py` — 14 new tool-parser cases covering one-shot parsing, mixed arg types, nested objects, multi-call blocks, streaming with partial bot-token, structural-tag info, and the args-to-JSON helper.
- [x] `pre-commit run --files ...` on all touched files.
- [ ] CI: post `/bot run --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"` once the PR is up.

## Known follow-up

The existing `needs_raw_special_tokens` hook on the tool parser only fires when `request.tools` is set. For requests that use the reasoning parser **without** tools, callers must pass `skip_special_tokens=false` themselves — otherwise `<|channel>` / `<channel|>` get stripped before the parser sees them. Adding a parallel flag on reasoning parsers would close this gap.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive Gemma 4 model support, including reasoning parsing and tool invocation capabilities.
  * Enabled sliding window attention with multi-pool KV cache management for improved memory efficiency in long-context sequences.

  * Optimized GPU synchronization in CUDA graph execution to reduce overhead and improve overall inference throughput.

### DIFF
--- a/cpp/tensorrt_llm/nanobind/batch_manager/kvCacheManager.cpp
+++ b/cpp/tensorrt_llm/nanobind/batch_manager/kvCacheManager.cpp
@@ -591,6 +591,11 @@ void tb::kv_cache_manager::KVCacheManagerBindings::initBindings(nb::module_& m)
         .def("count_reusable_blocks", &BaseKVCacheManager::countReusableBlocks, nb::arg("unique_tokens"),
             nb::arg("llm_request"), nb::arg("only_allocated") = false, nb::call_guard<nb::gil_scoped_release>())
         .def("get_cache_block_ids", &BaseKVCacheManager::getCacheBlockIds, nb::call_guard<nb::gil_scoped_release>())
+        .def(
+            "get_num_front_blocks_removed",
+            [](BaseKVCacheManager const& self, tb::LlmRequest::RequestIdType requestId)
+            { return self.getSequence(requestId).getNumFrontBlocksRemoved(); },
+            nb::call_guard<nb::gil_scoped_release>())
         .def("get_batch_cache_block_ids", &BaseKVCacheManager::getBatchCacheBlockIds,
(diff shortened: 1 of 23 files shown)


suyoggupta:
Superseded by #13248, which contains only the Gemma 4 reasoning/tool-call parser commit off a clean main branch.

PR #13275, opened 2026-04-21
[None][fix] initialize sampler state for ADP dummy requests

bobboli:
## Summary
- initialize sampler state for ADP dummy generation requests
- include ADP dummy generation requests in sampler setup request collection
- add unit coverage for the ADP dummy sampler path

## Problem
When attention DP is enabled, the executor can inject generation-phase dummy requests so an otherwise idle rank can still participate in collective operations during the forward pass.

Before this change, `setup_sampler_step()` only initialized sampler-side state for requests returned by `_filter_new_requests()`, which only covers newly activated context-last-chunk requests. ADP dummy generation requests were therefore skipped during sampler setup, even though they could still participate in later request grouping and sampling logic.

That left these dummy requests without the expected sampler slot/state initialization and could lead to incorrect sampler bookkeeping on the ADP dummy path.

## Fix
Use a broader request collection step for sampler setup:
- keep including the normal newly activated context-last-chunk requests
- additionally include ADP dummy generation requests (`is_attention_dp_dummy`)

This preserves the normal sampler setup path while ensuring ADP dummy requests get the same required initialization before later sampler processing.

## Testing
- Not run (not requested)

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/sampler.py
+++ b/tensorrt_llm/_torch/pyexecutor/sampler.py
@@ -2667,20 +2667,15 @@ class TorchSampler(Sampler[SampleStateTorch], AsyncWorkerMixin):
         return num_accepted_draft_tokens - 1
 
     @classmethod
-    def _filter_new_requests(cls, scheduled_requests: ScheduledRequests) -> list[LlmRequest]:
-        # list is faster than generator
-        return [
-            request
-            for request in scheduled_requests.context_requests_last_chunk
-            if not request.is_finished and not request.py_is_draft
-        ]
-
-    @classmethod
     def _collect_new_requests_for_setup(cls, scheduled_requests: ScheduledRequests) -> list[LlmRequest]:
         # ADP can inject generation-phase dummy requests after request activation.
         # Those still need sampler-side slot initialization before grouping/sampling.
         # The two source lists are disjoint by construction.
-        context_new_requests = cls._filter_new_requests(scheduled_requests)
+        context_new_requests = [
+            request
+            for request in scheduled_requests.context_requests_last_chunk
+            if not request.is_finished and not request.py_is_draft
+        ]
         adp_dummy_generation_requests = [
             request
             for request in scheduled_requests.generation_requests
--- a/tensorrt_llm/_torch/pyexecutor/sampler.py
+++ b/tensorrt_llm/_torch/pyexecutor/sampler.py
@@ -2676,22 +2676,17 @@ class TorchSampler(Sampler[SampleStateTorch], AsyncWorkerMixin):
         ]
 
     @classmethod
-    def _filter_adp_dummy_generation_requests(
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/pyexecutor/sampler.py
Funatiq: Finished generation requests are not filtered out in the sampler execution so we shouldn't filter them out here. Not sure if it's even possible that the dummy requests are finished or draft.

Funatiq: We should make it clear that the sets of requests selected by `_filter_new_requests` and `_filter_adp_dummy_generation_requests` are disjoint.
TBH I would prefer having a single function that creates both lists and concatenates them. Just to avoid passing scheduled_requests to both filters and have everything in one place.

bobboli: Updated this in 9d0a2e2be. I removed the extra filtering on ADP dummy generation requests here and now collect them based only on `is_attention_dp_dummy`, so this setup path no longer drops finished/draft dummy requests.

bobboli: Updated this in 9d0a2e2be. I collapsed the helper into a single `_collect_new_requests_for_setup()` implementation and added a comment that the two source lists are disjoint by construction, so the request selection logic now lives in one place.

tests/unittest/_torch/sampler/test_torch_sampler.py
```diff
@@ -68,6 +69,54 @@
...
+            adp_dummy_generation_request,
+            regular_generation_request,
+            finished_adp_dummy_generation_request,
+            draft_adp_dummy_generation_request,
+        ]
+
+        assert TorchSampler._collect_new_requests_for_setup(scheduled_requests) == [
+            context_request,
+            adp_dummy_generation_request,
+        ]
```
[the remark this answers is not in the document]
bobboli: Addressed in 9d0a2e2be. The test now asserts identity and length explicitly instead of relying on value-based `SimpleNamespace` equality.

tensorrt_llm/_torch/pyexecutor/sampler.py
Funatiq: You can also inline the list creation from _filter_new_requests and remove this function.

bobboli: Done in 806990cdd. I inlined the former `_filter_new_requests()` list construction directly into `_collect_new_requests_for_setup()` and removed the extra helper. I also reran the targeted unittest after the cleanup: `tests/unittest/_torch/sampler/test_torch_sampler.py::TestSetupSamplerStepRequestSelection::test_collect_new_requests_for_setup_includes_adp_dummy_generation_requests` passed.

Funatiq (approved):

PR #13285, opened 2026-04-21
[TRTLLM-11767][feat] LTX2 pipeline refactor part1 - auto upgrade to two stage pipeline

yibinl-nvidia:
## Summary by CodeRabbit

* **New Features**
  * The LTX2 visual generation example now automatically discovers spatial upsampler and distilled LoRA paths from checkpoint directories when CLI arguments are not explicitly provided, with informative logging for successful discoveries and warnings when multiple candidates are found.

## Description

This PR refactors LTX2 two-stage pipeline selection so it is handled in the core VisualGen config flow. spatial_upsampler_path and distilled_lora_path can now be provided through VisualGenArgs, and when omitted, TensorRT-LLM attempts to discover matching auxiliary checkpoints next to the LTX2 model checkpoint. When both auxiliary checkpoints are available, the base LTX2Pipeline is automatically promoted to LTX2TwoStagesPipeline.

This PR also adds an explicit one_stage_pipeline override for debugging, feature testing, and cases where users need to force one-stage execution even when two-stage assets are present. When enabled, TensorRT-LLM skips two-stage auxiliary checkpoint discovery and prevents promotion to LTX2TwoStagesPipeline. Unit coverage was added for default two-stage discovery/promotion, incomplete auxiliary checkpoint fallback, and the forced one-stage override.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/examples/visual_gen/visual_gen_ltx2.py
+++ b/examples/visual_gen/visual_gen_ltx2.py
@@ -3,6 +3,7 @@
 
 import argparse
 import time
+from pathlib import Path
 
 from tensorrt_llm import VisualGen, VisualGenArgs, VisualGenParams, logger
 from tensorrt_llm._torch.visual_gen.config import CacheDiTConfig
@@ -298,6 +299,32 @@ def parse_args():
     return parser.parse_args()
 
 
+def _discover_two_stage_paths(model_path: str) -> tuple[str, str]:
+    """Look up spatial upsampler and distilled LoRA in an LTX2 checkpoint directory.
+
+    Returns (upsampler_path, distilled_lora_path); empty strings if not a directory
+    or no unique match is found.
+    """
+    d = Path(model_path)
+    if not d.is_dir():
+        return "", ""
+
+    def _pick(pattern: str) -> str:
+        matches = sorted(d.glob(pattern))
+        if len(matches) == 1:
+            return str(matches[0])
+        if len(matches) > 1:
+            logger.warning(
+                f"Multiple files matching {pattern!r} under {model_path}: "
+                f"{[m.name for m in matches]}. Pass the path explicitly to disambiguate."
+            )
+        return ""
+
+    upsampler = _pick("*spatial-upscaler*.safetensors") or _pick("*upsampler*.safetensors")
+    distilled_lora = _pick("*distilled-lora*.safetensors") or _pick("*distilled*lora*.safetensors")
+    return upsampler, distilled_lora
+
+
 def _linear_type_to_quant_config(linear_type: str):
     """Map --linear_type CLI shortcut to quant_config dict for VisualGenArgs."""
     mapping = {
@@ -370,6 +397,15 @@ def _build_diffusion_args(args) -> VisualGenArgs:
 
 def main():
     args = parse_args()
+
+    if not args.spatial_upsampler_path or not args.distilled_lora_path:
+        discovered_upsampler, discovered_lora = _discover_two_stage_paths(args.model_path)
+        if not args.spatial_upsampler_path and discovered_upsampler:
+            args.spatial_upsampler_path = discovered_upsampler
+            logger.info(f"Discovered spatial upsampler: {discovered_upsampler}")
+        if not args.distilled_lora_path and discovered_lora:
+            args.distilled_lora_path = discovered_lora
+            logger.info(f"Discovered distilled LoRA: {discovered_lora}")
 
     if bool(args.spatial_upsampler_path) != bool(args.distilled_lora_path):
         missing = (
--- a/examples/visual_gen/visual_gen_ltx2.py
+++ b/examples/visual_gen/visual_gen_ltx2.py
@@ -1,11 +1,16 @@
 #!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
 """LTX2 Text/Image-to-Video generation using TensorRT-LLM Visual Generation."""
 
 import argparse
+import os
 import time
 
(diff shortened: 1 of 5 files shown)


examples/visual_gen/visual_gen_ltx2.py
```diff
@@ -281,6 +282,32 @@ def parse_args():
     return parser.parse_args()
 
 
+def _discover_two_stage_paths(model_path: str) -> tuple[str, str]:
```
zhenhuaw-me: If this is a LTX ckpt convention? Should we do this in core lib?

yibinl-nvidia: Good point, updated.

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -392,8 +393,20 @@ class VisualGenArgs(StrictBaseModel):
...
-            "the transformer for stage 2 denoising and un-merged afterwards."
+            "the stage 2 refinement pass. If omitted, TensorRT-LLM tries to "
+            "discover it in the checkpoint directory. The LoRA weights are "
+            "merged into the transformer for stage 2 denoising and un-merged afterwards."
+        ),
+    )
+
+    # LTX-2 specific option: force base one-stage pipeline even when two-stage
+    # auxiliary checkpoints are passed explicitly or discoverable next to the model.
+    one_stage_pipeline: bool = PydanticField(
```
zhenhuaw-me: This seems to be a debug knob, can we switch to use env var?

zhenhuaw-me (changes requested):

tensorrt_llm/_torch/visual_gen/config.py
yibinl-nvidia: That is good point, changed to env variable.

yibinl-nvidia:
@zhenhuaw-me could you please review again? I changed to env variable instead of CLI flag.

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -23,6 +24,9 @@
 
 CacheBackendName = Literal["teacache", "cache_dit"]
 
+LTX2_FORCE_ONE_STAGE_ENV = "TLLM_LTX2_FORCE_ONE_STAGE_PIPELINE"
```
zhenhuaw-me: I feel like this is over-engineered - I guess we can just query this knob when LtxPipeline init?

I guess someday we should have something like a env var registry in TRT-LLM, we can include this at that time.

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -562,6 +568,106 @@ def create_attention_metadata_state() -> Dict[str, Any]:
...
+        distilled_lora_path = _pick_unique_file(
+            search_dir, "*distilled-lora*.safetensors"
+        ) or _pick_unique_file(search_dir, "*distilled*lora*.safetensors")
+
+    return spatial_upsampler_path, distilled_lora_path
+
+
+def resolve_ltx2_pipeline_extra_attrs(
+    checkpoint_path: Path,
+    spatial_upsampler_path: str = "",
```
zhenhuaw-me: Does it make sense to put these utils impl in LTX folders?

yibinl-nvidia: good point, modified.

tensorrt_llm/_torch/visual_gen/config.py
yibinl-nvidia: I think it is fine in the current scope? We only check it at the time we are checking if to upgrade to two stage pipeline.

zhenhuaw-me: Sure, it's okay to implement like this. Meanwhile, I'd prefer to put the "model specific code" to the corresponding file, and the "general code" can call or use some of the per-model code/utils.

examples/visual_gen/visual_gen_ltx2.py
```diff
@@ -1,11 +1,19 @@
...
 """LTX2 Text/Image-to-Video generation using TensorRT-LLM Visual Generation."""
 
 import argparse
+import os
 import time
 
 from tensorrt_llm import VisualGen, VisualGenArgs, VisualGenParams, logger
 from tensorrt_llm._torch.visual_gen.config import CacheDiTConfig
+from tensorrt_llm._torch.visual_gen.models.ltx2.pipeline_ltx2 import (
+    LTX2_FORCE_ONE_STAGE_ENV,
```
zhenhuaw-me: I don't think it's a good idea to import such internal knobs to an example. Examples should avoid using internal components, otherwise we don't have a bar of what's committed to be stable and what's not.

examples/visual_gen/visual_gen_ltx2.py
```diff
@@ -1,11 +1,19 @@
...
+)
+from tensorrt_llm.serve.media_storage import MediaStorage
```
zhenhuaw-me: I think we don't have such interfaces anymore?

zhenhuaw-me (approved):
LGTM in general with 2 comments.

examples/visual_gen/visual_gen_ltx2.py
yibinl-nvidia: good point, this is from a merge conflict. Removed.

tensorrt_llm/_torch/visual_gen/config.py
yibinl-nvidia: Modified to move model specific code to LTX-2 pipeline file.

examples/visual_gen/visual_gen_ltx2.py
yibinl-nvidia: modified.

arysef (approved):

PR #13301, opened 2026-04-21
[NVBug 5969206][draft] Subset of PR 13056 cherry-picked onto 4e69c14f73 (rc11 base) — 15 C++/Python fixes

yifjiang:
## Intent

Draft PR. Subset of the commits in #13056, rebased/cherry-picked onto base commit `4e69c14f732a6e6afce4f71616db5b5cd2b10530` (the TRT-LLM commit the `tensorrt-llm==1.3.0rc11` wheel in NVIDIA NGC's `dynamo-trtllm:dynamo-main-pr12639-x86-b200` image is built on).

This branch is **not** meant to replace PR #13056 — it exists so we can build a rc11-based image with the shared_ptr / UAF fixes applied and test it against the d64 reproducer documented downstream. If the resulting image no longer crashes with `std::future_error: Broken promise`, that localises the remaining failure mode.

## What's included (15 commits from #13056, in cherry-pick order, + 1 follow-up fix for a resolution mistake)

All 15 fix commits come from PR #13056. One additional follow-up commit on this branch re-adds a guard that was accidentally dropped during conflict resolution (see "Conflict resolution audit" below).

1. `d7844fc6f` — Enforce `kv_transfer_timeout_ms` on prefill sender + guard against dangling pointers
2. `b2b0a820f` — Report timed-out transfers to Python for full cleanup
3. `9ad782231` — Apply `kv_transfer_timeout_ms` in blockAll mode
4. `daf70d403` — Close remaining UAF paths and enforce gen-side deadline
5. `b9450529b` — Hoist `kv_transfer_timeout_ms` check above readiness gate
6. `267e813c2` — Release receiver resources when gen-side deadline evicts
7. `738183086` — `sendRequestInfo` / `receiveReadySignal` / data-phase send honor `perRequestCancel`
8. `3736d1129` — Sender-side in-flight cancel parity + `notifySyncMessage` pre-notify check
9. `6cf21fd7c` — Tie sender cancel-flag lifetime to session + port `catch(...)` to sender worker
10. `149421e72` — Bound `assignBufferIndex` CV wait + plumb cancel + diagnostic `[buf]` logs
11. `30f9cb2ec` — RAII `BufferIndexHolder` for `requestSync` — stop leaking pool slots on non-happy exits
12. `28ef114db` — RAII `BufferIndexHolder` on sender formatters + reqId-tagged `[buf]` logs
13. `902fab728` — Send-side RAII: atomic `release()`, cancel parity, FUTURE_WAIT/JOIN diagnostics
14. `33262c5ba` — **Carry `std::shared_ptr<LlmRequest>` through transceiver** — close UAF, remove Path A/B guards, unwedge KV block leak
15. `f63172f7e` — Harden drain worker loop against non-`std::exception` escapes
16. _(follow-up on this branch, not from #13056)_ — Re-add `is_disagg_context_complete_state` guard to `_handle_responses` that was dropped during the `d7844fc6f` conflict resolution. See audit below.

## What's intentionally **not** here (vs PR #13056)

- Comment-only chores (`84499475b`, `568454fbd`, `b2dd08146`, `5779e5c53`, `4001ac5d7`)
- Regression test (`88987cc08`) — pulled out; test harness needs adaptation on the older base
- Diagnostic-only commits (`e61a5fa2b`, `5904b9e11`)
- Reverted pair (`80474ce30`, `d30767c69`)

## Conflict resolution audit

Four files needed manual resolution because the base at `4e69c14f73` predates refactors the PR assumed. Each resolution was re-audited after the initial pass — findings below.

### `py_executor.py` first hunk (`d7844fc6f`) — **had a bug, now fixed**

PR 13056 introduces three guard branches before the terminate paths:

```python
if request.is_disagg_context_complete_state:
    requests_finished_by_transfer.append(request)   # not in this subset's codepath
elif request.is_disagg_context_transmission_state:
    pass
elif force_terminate_for_partial_reuse:
    requests_to_terminate.append(request)
else:
    requests_to_terminate.append(request)
```

The `requests_finished_by_transfer` local variable is defined elsewhere in the PR 13056 base but **not** on `4e69c14f73`, so the first branch can't be ported verbatim. The initial resolution dropped the whole first branch.

Audit found this was unsafe: the `is_disagg_context_complete_state` attribute IS available on `4e69c14f73` (C++ `isDisaggContextCompleteState()` + nanobind binding). A COMPLETE-state request can remain in `active_requests` when `_end_transfer_and_maybe_terminate` has `end_transfer()` return False, and without the guard `_handle_responses` would append it to `requests_to_terminate` and call `_terminate_request` → `resource_manager.free_resources(request)` a second time (nvbug/5961736). This is a pre-existing latent issue on rc11 base that PR 13056 fixes; dropping the guard during cherry-pick left the subset failing to apply that fix.

The follow-up commit on this branch re-adds the guard as a `pass` branch — the subset doesn't need the `requests_finished_by_transfer` stats tracking, only the double-free avoidance.

### `tests/unittest/others/test_kv_cache_transceiver.py` (`b9450529b`) — test coverage loss, no runtime impact

The two new regression tests `test_context_transfer_times_out_on_stuck_sender` and `test_context_transfer_times_out_with_at_least_zero` depend on test helpers not available on `4e69c14f73`. Dropped entirely; the subset is C++-fixes-only and the tests should be ported separately.

### `dataTransceiver.cpp` (`738183086` / `30f9cb2ec`) — signature evolution, took the later form

The signature of `sendRequestInfo` evolves across commits 7 and 11:
- (7) `sendRequestInfo(llmRequest, perRequestCancel)`
- (11) `sendRequestInfo(llmRequest, perRequestCancel, std::move(cacheBufferIds))`

The cherry-pick for (7) conflicted because HEAD had the single-argument form; the cherry-pick for (11) then conflicted again on the same site. Both were resolved by taking the later commit's form (11), which is what the PR ultimately lands on. No divergence from PR 13056's final state.

### `cacheTransceiver.cpp` + `py_executor.py` (`33262c5ba`) — faithful to commit intent

This commit explicitly removes the Path A/B Python guards that earlier PR commits (notably `daf70d403`) added, because `mSenderFutures` / `mRequesterFutures` now hold `std::shared_ptr<LlmRequest>` and the UAF class the guards existed to prevent is gone. The conflict resolution accepts the removal — this is exactly what the commit message describes.

## Build / test instructions

This branch is for another agent to build into a Dynamo container image. The intended usage:

```bash
git clone https://github.com/yifjiang/TensorRT-LLM.git
cd TensorRT-LLM
git checkout pr13056-subset-on-4e69c14
# Build a tensorrt_llm wheel from this branch (replaces the 1.3.0rc11 wheel baked into the dynamo image)
# Layer the new wheel into a dynamo-trtllm image; see the downstream d64 reproducer doc for the expected harness.
```

Do not merge this draft PR into `main` — it is a cherry-pick against a known past base commit and will conflict with current `main`. It exists only to supply a clean patch series for an offline build.

## Verification plan (run downstream)

1. Build the patched `tensorrt_llm` wheel from this branch.
2. Overlay into the Dynamo container image currently used by the d64 reproducer.
3. Run the 1P1D loadgen at conc 16 / 48 / 128 with live worker-log tails.
4. Pass criteria: no `std::future_error: Broken promise`, no `bad optional access` in the event loop; all three recovery probes return `ok200`.

### DIFF
--- a/cpp/include/tensorrt_llm/batch_manager/cacheTransceiver.h
+++ b/cpp/include/tensorrt_llm/batch_manager/cacheTransceiver.h
@@ -204,13 +204,22 @@ class BaseCacheTransceiver
 {
 public:
     virtual ~BaseCacheTransceiver() = default;
-    virtual void respondAndSendAsync(LlmRequest* llmRequest) = 0;
+    // These methods take std::shared_ptr<LlmRequest> (not a raw pointer) so the
+    // transceiver and its async workers can hold a strong reference for the
+    // duration of the transfer. Previously they took LlmRequest*, which left
+    // the worker (via RequestAndPromise::mRequest) dereferencing a pybind
+    // shared_ptr that Python could drop via _terminate_request — a
+    // forensically-confirmed UAF pattern (mRequestId read as
+    // 0x5555555555555555 under MALLOC_PERTURB_). Passing the shared_ptr through
+    // keeps the LlmRequest alive until every holder (Python active_requests,
+    // C++ futures map, async worker) has released its reference.
+    virtual void respondAndSendAsync(std::shared_ptr<LlmRequest> llmRequest) = 0;
     virtual void respondAndSendLayerWise(
         RequestVector const& requests, std::shared_ptr<ContextProgress> const& progress)
         = 0;
 
-    virtual void requestAndReceiveSync(LlmRequest* llmRequest) = 0;
-    virtual void requestAndReceiveAsync(LlmRequest* llmRequest) = 0;
+    virtual void requestAndReceiveSync(std::shared_ptr<LlmRequest> llmRequest) = 0;
+    virtual void requestAndReceiveAsync(std::shared_ptr<LlmRequest> llmRequest) = 0;
 
     /// Check all requests transferring context, and return the requests that have completed or encountered an error.
     virtual RequestStatuses checkContextTransferStatus(
@@ -221,7 +230,7 @@ public:
 
     [[nodiscard]] virtual bool checkGenTransferComplete() const = 0;
 
-    virtual bool cancelRequest(LlmRequest* llmRequest) = 0;
+    virtual bool cancelRequest(std::shared_ptr<LlmRequest> llmRequest) = 0;
 };
 
 class CacheTransceiver : public BaseCacheTransceiver
@@ -252,13 +261,13 @@ public:
 
     virtual ~CacheTransceiver();
 
-    void respondAndSendAsync(LlmRequest* llmRequest) override;
+    void respondAndSendAsync(std::shared_ptr<LlmRequest> llmRequest) override;
 
     void respondAndSendLayerWise(
         RequestVector const& requests, std::shared_ptr<ContextProgress> const& progress) override;
 
-    void requestAndReceiveSync(LlmRequest* llmRequest) override;
-    void requestAndReceiveAsync(LlmRequest* llmRequest) override;
+    void requestAndReceiveSync(std::shared_ptr<LlmRequest> llmRequest) override;
+    void requestAndReceiveAsync(std::shared_ptr<LlmRequest> llmRequest) override;
 
     RequestStatuses checkContextTransferStatus(
         std::optional<int> const& atLeastRequestNum = std::nullopt, bool markComplete = false) override;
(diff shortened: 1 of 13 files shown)


brnguyen2:
👋 Thanks for this! As part of a one-time TensorRT-LLM open-PR backlog cleanup, we're closing this PR since it looks like a subset of #13056, already covered there. If that's not right, just reopen and let us know.

PR #13276, opened 2026-04-21
[TRI-966] [fix] Fix L0_backend_trtllm

pskiran1:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streaming input handling to ensure it's always properly coerced to a boolean value, with a default of `false` when omitted. This improves robustness in request processing.

NV bug: https://nvbugs/6089262

## Description

This PR updates the Triton `tensorrt_llm` model to pass the `streaming` argument as a boolean.
Previously, the model failed when building a `trtllm.Request(**inputs, ...)` because it passed `streaming=None` to the `pybind` constructor, which only accepts `streaming: bool` (not `Optional[bool]`).

Error:
`
File \"/opt/tritonserver/tensorrtllm_backend/ci/L0_backend_trtllm/triton_model_repo/tensorrt_llm_bls/1/lib/triton_decoder.py\", line 122, in _exec_triton_request_single\n    raise pb_utils.TritonModelException(responses.error().message())\nc_python_backend_utils.TritonModelException: An error occurred when processing the input values for request id d0b08d87-0515-4af6-aba8-adb07bf99666, the error was '__init__(): incompatible function arguments. The following argument types are supported:\n    1. __init__(self, input_token_ids: collections.abc.Sequence[int], max_tokens: int, *, streaming: bool = False, ...
`
## Test Coverage

L0_backend_trtllm--base

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [X] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

[...]

PR #13308, opened 2026-04-22
[TRI-978][fix] Fix streaming=None crash and CI test failures in L0_backend_trtllm

mc-nv:
## Summary

Fixes multiple CI failures in `L0_backend_trtllm` caused by the executor requiring an explicit `streaming` bool.

- **`model.py`**: coerce `streaming=None` → `False` when `get_input_scalar_by_name()` returns `None` (mirrors NVIDIA/TensorRT-LLM#13276)
- **`base_metrics_verification_tests.py`**: pass explicit `streaming=False` tensor; rename `"streaming"` → `"stream"` to match the ensemble model's declared external input
- **`custom_metrics_verification_tests.py`**: fix malformed `assertTrue` chained comparison; use `fromtimestamp` instead of `utcfromtimestamp` to avoid 8-hour offset on UTC-offset runners
- **`custom_metrics_reporter.cc`**: set `tm_isdst=-1` before `mktime()` so DST is auto-determined, matching `localtime()` used in `getCurrentTimestamp()`
- **`benchmark_core_model.py`**: pass explicit `streaming` tensor; switch to gRPC protocol to avoid gevent segfault on aarch64 during Python exit

Related: NVIDIA/TensorRT-LLM#13276

## Release Notes

* **Bug Fixes**
  * Improved streaming parameter handling with proper boolean normalization
  * Enhanced timestamp conversion accuracy for daylight saving time compatibility

* **Tests**
  * Updated test infrastructure to include streaming parameter support in inference requests
  * Added explicit protocol specification to benchmarking tools

### DIFF
--- a/triton_backend/ci/L0_backend_trtllm/base_metrics_verification_tests.py
+++ b/triton_backend/ci/L0_backend_trtllm/base_metrics_verification_tests.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python
-# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -81,6 +81,8 @@ class TRTLLMBaseMetricsTest(unittest.TestCase):
                 utils.prepare_tensor("max_tokens", output0_len, "http"),
                 utils.prepare_tensor("bad_words", bad_words_list, "http"),
                 utils.prepare_tensor("stop_words", stop_words_list, "http"),
+                utils.prepare_tensor("stream", np.array([[False]], dtype=bool),
+                                     "http"),
             ]
             if beam_width_value > 1:
                 beam_width = np.ones_like(input0).astype(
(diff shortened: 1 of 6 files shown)


triton_backend/inflight_batcher_llm/src/custom_metrics_reporter/custom_metrics_reporter.cc
```diff
@@ -75,6 +75,7 @@ uint64_t convertTimestampToMicroseconds(std::string const& ts)
     std::tm tm = {};
     std::stringstream ss(ts);
     ss >> std::get_time(&tm, "%m-%d-%Y %H:%M:%S");
+    tm.tm_isdst = -1; // Let mktime determine DST to match localtime() used in getCurrentTimestamp
```
SimengLiu-nv: C++ guidelines require nontrivial literals to be named constants. Use a small function-scope constant, for example `int const kAutoDetermineDst = -1`; then assign that here.

triton_backend/ci/L0_backend_trtllm/base_metrics_verification_tests.py
```diff
@@ -81,6 +81,8 @@ def _run_infer(self, client, prompts, output_lens, beam_width_value=1):
                 utils.prepare_tensor("max_tokens", output0_len, "http"),
                 utils.prepare_tensor("bad_words", bad_words_list, "http"),
                 utils.prepare_tensor("stop_words", stop_words_list, "http"),
+                utils.prepare_tensor("stream", np.array([[False]], dtype=bool),
```
SimengLiu-nv, line 84: Should this be `streaming`?

SimengLiu-nv (changes requested):

triton_backend/ci/L0_backend_trtllm/base_metrics_verification_tests.py
mc-nv: This value is accurate: 
https://github.com/nvidia/TensorRT-LLM/blob/main/triton_backend/all_models/inflight_batcher_llm/ensemble/config.pbtxt#L186

Issue #13320, opened 2026-04-22
[Bug]:  kv_cache_connector: decode-continuation blocks never allocated — GenerationRequest's mCacheBlockIds frozen while mNumTokens grows


GavinZhu-GMI:
### Host
  - Server: Dell PowerEdge XE9680 (one of three nodes in an RKE2 k8s cluster)
  - CPU architecture: x86_64 (128 vCPUs visible)
  - Host memory: 2 TiB (2113271052 KiB capacity)
  - GPU: 8 × NVIDIA H200, 140 GB each, compute capability 9.0
  - NVIDIA driver: 580.126.09
  - Host OS: Ubuntu 22.04.5 LTS (kernel 5.15.0-171-generic)
  - Container runtime: containerd 2.1.4-k3s2
  - Kubernetes: RKE2 v1.31.13+rke2r1

  ### Container (custom Dynamo build from `ai-dynamo/dynamo` `container/render.py --framework trtllm --cuda-version 13.1`)
  - Image: `customized build`
  - Dynamo submodule pinned at: `aea7e8e` (`[None][feat] Switch CP cache transmission from contiguous to round-robin (#13180)`)
  - Base OS in image: Ubuntu 24.04.3 LTS
  - Also reproduced against the stock TRT-LLM `1.3.0rc11` pip wheel installed into the same image; `cpp` source inspected is from main branch (which already contains #12976 and #13029) — bug
  reproduces identically on main as well, so the fix does not appear to be upstream yet at the time of filing.

  ### Libraries (inside the container)
  - Python: 3.12.3
  - tensorrt_llm: **1.3.0rc11**
  - tensorrt: shipped with wheel (verified via `pip show`)
  - torch: `2.11.0a0+eb65b36914.nv26.02` (NVIDIA PyTorch build)
  - transformers: 4.57.3
  - nvidia-modelopt: 0.37.0
  - nvcc: CUDA 13.1.115 (release 13.1, V13.1.115)
  - libcublas-13-1: 13.2.1.1-1
  - libcudnn9-cuda-13: 9.21.0.82-1
  - libnccl2: 2.29.2-1+cuda13.1

  ### Model under test
  - `zai-org/GLM-5.1-FP8` (`GlmMoeDsaForCausalLM`, 754 B params, 78 layers, 256 routed experts, block-wise FP8 KV)

  ### Backend selection (from engine config + defaults)
  - `backend: pytorch`
[...]

PR #12161, opened 2026-03-12
[None][perf] Cache FlashMLA tile-scheduler metadata across attention layers

bobboli:
## Description
Pre-compute FlashMLA tile-scheduler metadata once per forward pass in Python and pass it into the C++ attention op, instead of rebuilding it inside every attention layer.
This also fixes the metadata layout for mixed prefill + decode batches by computing metadata from the compacted generation sub-batch, so the generated num_splits and tile-scheduler metadata match the block_ids_per_seq layout consumed by FlashMLA.
### Why
The previous flow rebuilt FlashMLA metadata inside the C++ attention path for each layer. Besides the repeated overhead, the ported logic could use metadata derived from the wrong batch view when context and generation requests were mixed together.
That mismatch could leave FlashMLA reading metadata that did not correspond to the active generation sub-batch, which showed up as accuracy failures and, in some cases, illegal memory access.
### What changed
Added a Python/C++ entry point to compute FlashMLA metadata into pre-allocated tensors.
Extended the attention op plumbing to accept pre-computed flash_mla_tile_scheduler_metadata and flash_mla_num_splits.
Updated the FlashMLA generation path to require and consume the pre-computed metadata instead of generating it locally.
Allocated stable metadata buffers in TrtllmAttentionMetadata so they can be reused across layers and CUDA graph captures.
Recompute metadata from the generation slice only, and invalidate cached metadata when KV lengths change or speculative decoding updates the batch state.

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [ ] Please check this after reviewing the above items as appropriate for this PR.

[...]

PR #13323, opened 2026-04-22
[TRTLLM-13320][fix] py_executor_creator: propagate FlashMLA tokens_per_block override onto kv_cache_config

GavinZhu-GMI:
## Summary
This fixes issue #13320 
When `is_mla(config) && enable_flash_mla` is true in `py_executor_creator.py`, the MLA path rebinds the local `tokens_per_block` variable to 64 but leaves `kv_cache_config.tokens_per_block` at whatever the user set (default 32). `kv_cache_creator` is then passed the local (64) and the KVCacheManager is built with 64, while any `KvCacheConnectorScheduler` subclass instantiated below via `scheduler_cls(llm_args)` reads `llm_args.kv_cache_config.tokens_per_block` and sees the stale 32.

## The bug this produces

For the Dynamo KVBM connector (same pattern applies to LMCache) the desync produces a frozen `cache_block_ids` view:

- KVCacheManager internally uses 64-token blocks.
- Connector thinks `block_size = 32`, allocates its KV offload pools on that assumption, and asks for block deltas every iter.
- Decode in the KVCacheManager never crosses a 64-token boundary for short-to-medium generations, so `get_cache_block_ids()` keeps returning the single prefill block.
- From the connector's perspective the engine is writing past its registered blocks but never reporting new allocations. Offloads are skipped / misaligned. Output is plausible-looking but mathematically corrupted because decode KV is written past the connector's bookkeeping.

Reproduced on `tensorrt_llm==1.3.0rc11` with a block-FP8 MLA model (GLM-5.1-FP8, `GlmMoeDsaForCausalLM`) + Dynamo KVBM connector; `tokens_per_block=32` in user config; `impl.tokens_per_block=64` at runtime per nanobind property.

Evidence table (rank-0 probe of `KvCacheConnectorSchedulerOutputRequest.update_and_build_data`, block_size=64 in C++):

| iter | `impl.get_token_count` | `impl.tokens_per_block` | `get_cache_block_ids` |
|------|------------------------|-------------------------|------------------------|
| prefill | 14 | 64 | `[2117]` |
| decode 1 | 15 | 64 | `[2117]` |
| decode boundary | 33 | 64 | `[2117]` |
| decode past boundary | 61 | 64 | `[2117]` |

Meanwhile the Dynamo-side slot assertion at `block_manager/vllm/connector/leader/slot.rs:734` (`next_position <= device_blocks.len() * block_size`) is evaluated with connector `block_size=32` and fires at `current_position=32, num_scheduled_tokens=1` → `33 > 32`. The connector is correct relative to its (stale) view of the block size; the engine is correct relative to its (actual) block size. They disagree by 2x.

## Fix

One line. The neighboring overrides in the same `if is_mla(config):` block (e.g. `kv_cache_config.enable_block_reuse = False` at line 569) already write back onto the config object; the `tokens_per_block = 64` override forgot the same pattern. Add the dotted assignment:

```
if model_engine.model.model_config.enable_flash_mla:
tokens_per_block = 64
kv_cache_config.tokens_per_block = tokens_per_block # new
logger.info(...)
```

Both consumers — `kv_cache_creator(tokens_per_block=...)` and `scheduler_cls(llm_args)` — now resolve to the same value.

## Related

- Dynamo-side soft-degrade PR that turns the downstream assertion into a `kvbm_slot_desync_total` counter + warn log so the engine stays alive while this fix is in flight: https://github.com/ai-dynamo/dynamo/pull/8490.
- Affects any `KvCacheConnectorScheduler` (LMCache, KVBM, custom) paired with an MLA model that enables FlashMLA.

## Test plan

- [x] Confirmed reproducer: GLM-5.1-FP8, TP=8, FlashMLA on, KVBM connector. Decode past 32 tokens produces frozen `cache_block_ids`.
- [x] Existing TRT-LLM CI (`pytest tests/unittest/`).
- Optional follow-up: extend `tests/integration/defs/llmapi/test_llm_api_connector.py` to exercise MLA + FlashMLA + connector and assert `scheduler.update_state_after_alloc` + cached-delta invariants. Happy to add in this PR if reviewers prefer.


GavinZhu-GMI:
@leslie-fang25  friendly ping — this is a one-liner fix for FlashMLA scheduler/cache mismatch (matches `KVCacheManager.tokens_per_block = 64` already set above). Happy to address feedback at your convenience.

GavinZhu-GMI:
@bobboli — could you take a quick look when you have a moment? This is a one-line fix that propagates `KVCacheManager.tokens_per_block = 64` (set on the FlashMLA path above) onto `kv_cache_config` so scheduler components see the same value. Closely related to your #12161 work on FlashMLA tile-scheduler caching. Happy to address feedback.

bobboli (approved):

GavinZhu-GMI:
@bobboli thanks for the review! The PR is now blocked on (1) Jenkins CI never running (community-PR allowlist gate) and (2) the `trt-llm-torch-runtime-devs` team request still showing pending. Could you kick off `/bot run` ? Happy to be added to the allowlist if that's the path.

bobboli:
> @bobboli thanks for the review! The PR is now blocked on (1) Jenkins CI never running (community-PR allowlist gate) and (2) the `trt-llm-torch-runtime-devs` team request still showing pending. Could you kick off `/bot run` ? Happy to be added to the allowlist if that's the path.

Thanks for your contribution! The issue is fixed in https://github.com/NVIDIA/TensorRT-LLM/pull/13752 and now this PR is duplicated. Could you confirm and close this?

PR #13334, opened 2026-04-22
[None][infra] enable CUDA line info by default for Debug/RelWithDebInfo

bobboli:
## Summary
- Emit `--generate-line-info` by default for both `Debug` and `RelWithDebInfo` CUDA flags, so `nsys`/`ncu` can map samples back to source without needing to hand-tweak CMake.
- Emit `-G` by default for `Debug` so `cuda-gdb` can step through kernels.
- Because these flags inflate device-side section sizes enough that linking against every supported CUDA architecture overflows ELF section limits, `scripts/build_wheel.py` now rejects `Debug`/`RelWithDebInfo` builds that don't pass an explicit `--cuda_architectures`. `Release` builds are unaffected.

## Motivation
Previously the project shipped with `--generate-line-info` and `-G` commented out in `cpp/CMakeLists.txt` because enabling them caused link-time failures when building for all CUDA archs. In practice the only builds that compile `RelWithDebInfo`/`Debug` are developers working on specific GPUs — they don't need every arch, and they *do* need line info/device debug info. Trading a default-all-archs developer build (which almost no one actually uses with debug info) for usable debug/profile builds is a clear win.

## Test plan
- [ ] `python scripts/build_wheel.py -b Release` still succeeds with default `cuda_architectures=all`.
- [ ] `python scripts/build_wheel.py -b RelWithDebInfo` now fails fast with an informative error telling the developer to pass `--cuda_architectures`.
- [ ] `python scripts/build_wheel.py -b RelWithDebInfo --cuda_architectures 90-real` succeeds and produces a `libtensorrt_llm.so` with CUDA line info resolvable by `nsys`/`ncu`.
- [ ] `python scripts/build_wheel.py -b Debug --cuda_architectures 90-real` succeeds and produces a binary debuggable with `cuda-gdb`.

## Summary by CodeRabbit

* **Chores**
  * Updated CUDA compilation flags for Debug and RelWithDebInfo builds to include source line mappings and device debug information.
  * Build process now requires explicit CUDA architecture specification when using Debug or RelWithDebInfo configurations instead of using defaults.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -323,15 +323,16 @@ if(UNIX)
   set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -O0 -fno-inline")
 endif()
 
-# Note: The following are desirable settings that should be enabled if we
-# decrease shared library size. See e.g.
-# https://github.com/rapidsai/cudf/pull/6134 for a similar issue in another
-# project.
-
-# set(CMAKE_CUDA_FLAGS_RELWITHDEBINFO "${CMAKE_CUDA_FLAGS_RELWITHDEBINFO}
-# --generate-line-info")
-
-# set(CMAKE_CUDA_FLAGS_DEBUG "${CMAKE_CUDA_FLAGS_DEBUG} -G")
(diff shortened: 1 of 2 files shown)


yuxianq (approved):

juney-nvidia (approved):
Approved from OSS compliance perspective.

PR #13340, opened 2026-04-22
[None][feat] Integrate FP4 indexer for DSA on Blackwell

lfr-0531:
@coderabbitai summary

## Description

Add the FP4 indexer path for DSA on Blackwell (SM100+) and land it with a native CUDA fused-quantize op. Three commits on top of `main`:

1. **`[None][feat] Integrate FP4 indexer for DSA on Blackwell`** (`e9542afeb`)
   - Upgrade DeepGEMM to `Barry-Delaney/DeepGEMM@a97b74d7` (`user/jinshik/nv_dev_rebase`), which adds the FP4 MQA logits kernels and switches the paged MQA logits APIs to require 2D `context_lens`. All `get_paged_mqa_logits_metadata` / `fp8_paged_mqa_logits` call sites in `dsa.py` are migrated to 2D via a pre-allocated `kv_lens_cuda_2d` buffer (no capture-time allocations).
   - Add `DeepSeekSparseAttentionConfig.indexer_k_dtype: Literal["fp8","fp4"]`.
   - Thread an `indexerKCacheUseFp4` flag through the C++ KVCacheManager chain (`WindowBlockManager`, `BlockManager`, all 4 `KVCacheManager` overloads, the nanobind binding) and through the disagg `CacheState` (serialization + `operator==`, so prefill/decode refuse to pair when the flag disagrees). `createIndexerKCachePools` picks `data_size = index_head_dim/2` under FP4, shrinking per-token size from 132B to 68B.
   - Relax `IndexerKCacheScatterOp` / `IndexerKCacheGatherOp` (+ their kernels) to accept `head_dim` ∈ `{128 (FP8), 64 (FP4-packed)}`, pick `threads_per_block = head_dim / VEC_SIZE` at runtime, and forward `head_dim=64` under FP4 from the callers.
   - `Indexer._prep_q_or_k` dispatches to the FP4 quantizer when `use_fp4`; `pre_indexer_proj` returns a 5th `q_scale` output; weights under FP4 carry only `softmax_scale * n_heads^-0.5` because the kernel applies `q_scale` internally. New `_call_mqa_logits` / `_call_paged_mqa_logits` helpers reinterpret the FP8 gather output as int8 / int32 under FP4 and route to `fp8_fp4_mqa_logits` / `fp8_fp4_paged_mqa_logits`. `DSACacheManager` computes per-token size based on `use_fp4`.
   - Thread `q_scale` through the two-op CUDA-graph-split DSA path in `attention.py` and through `KVCacheManager.__init__` in `resource_manager.py`.
   - First-pass quantizer: `fp4_quantize_1x32_sf_transpose` — a Triton kernel that fuses amax, UE8M0 ceil, FP4 E2M1 quantize, nibble packing, and four-per-int32 scale packing, bit-identical to DeepGEMM's `testing.per_token_cast_to_fp4(..., gran_k=32, use_packed_ue8m0=True)` reference.

2. **`[None][perf] Replace Triton FP4 indexer quantizer with fused_cat_fp4 CUDA op`** (`c90c56e85`)
   - Mirror the existing `fused_cat_fp8` op with an FP4 E2M1 variant: `torch.ops.trtllm.fused_cat_fp4(pe, nope) -> (packed_int8, scale_int32)`. Fuses concat + per-block-32 quantize + UE8M0 scale packing into one CUDA kernel, removing the remaining Triton DSL dependency on the DSA CUDA-graph hot path and giving FP4/FP8 indexer branches a uniform native-CUDA call-site shape.
   - Kernel: grid `(ceil(M / 8),)`, block `256`. Each warp handles one 128-element row; thread `t` covers elements `[4t, 4t+3]`. Per-block-32 amax via 3-round `__shfl_xor_sync` (offsets 1, 2, 4 — stays inside each group of 8 lanes). UE8M0 scale via IEEE 754 bit trick; `MIN_AMAX = 1e-12f` keeps `ratio` normal so no `exp` clamp needed. FP4 E2M1 quantize via 7-way bucketize on `{0.25, 0.75, 1.25, 1.75, 2.5, 3.5, 5.0}`. IEEE `div.rn.f32` for the scaled value. Nibble pack (even → low, odd → high); lane 0 gathers the four UE8M0 exponent bytes from `{0, 8, 16, 24}` via `__shfl_sync` and writes one int32 per row (little-endian).
   - `Indexer._prep_q_or_k` switched to the new op; `fp4_quantize_1x32_sf_transpose` import removed.
   - `tensorrt_llm/quantization/utils/fp4_utils.py` reverted exactly to its pre-PR state (`git diff HEAD^` on this file is empty). `tests/unittest/_torch/quantization/{__init__.py,test_fp4_quantize.py}` removed; coverage moves to `test_cpp_custom_ops.py`. The `unittest/_torch/quantization` entry is dropped from `l0_b200.yml`.

3. **`[None][fix] Address review feedback on FP4 indexer PR`** (`397eafe06`) + **`[None][fix] Address round-2 review feedback on FP4 indexer PR`** (`eac155a4c`)
   - **Round-1 P0**: FP4 branch of `pre_indexer_proj` now runs `q_scale = q_scale.view(-1, self.n_heads, 1)` — `fused_cat_fp4` flattens to `M = N * n_heads`, so downstream token-axis slicing in `sparse_attn_indexer` (chunk / ctx / decode) needs the same reshape the FP8 branch already had. Without this, the FP4 path was silently slicing wrong per-token scales.
   - **Round-1 P1**: `DSACacheManager.get_cache_size_per_token` / `get_cache_bytes_per_token` now pick `index_head_dim // 2` under FP4, so the KV cache block budget sees the 68 B/token footprint (was hardcoded to 132 B). Without this, the block budget over-counted and the FP4 memory win never reached the allocator.
   - **Round-1 P1**: `@model_validator(mode="after")` on `DeepSeekSparseAttentionConfig` rejects `fp4` on SM<100 or `index_head_dim != 128`, degrades gracefully when CUDA is unavailable, and now includes recovery hints in the error message.
   - **Round-2 P0**: `ModelConfig.from_pretrained` (DSV3.2 / GlmMoeDsa rebuild branch) now forwards `indexer_k_dtype` — previously it was hand-copying every other sparse-attention field, so the user knob was silently reset to `"fp8"` before any downstream consumer saw it. Added a regression-guard test that exercises the rebuild and statically asserts the keyword is present in the source.
   - **Round-2 P1**: dropped the dead `scheduler_metadata_buffer_mtp3` path. DeepGEMM's upgraded paged MQA logits kernel picks `num_kv_multicast=1` for every `next_n` on SM100 (verified by the `_schedule_meta_size` assertion in `deepgemm-src/csrc/apis/attention.hpp:339` firing when the legacy layout is passed). Removed the buffer allocation, both population sites, and the production dispatch branch.
   - **Round-1 / Round-2 P2 + nits**: `CacheState::operator==` now also compares `mHasIndexerKCache`, `mIndexerDimPerHead`, `mIndexerKCacheQuantBlockSize` so prefill/decode refuse to pair on any incompatible indexer layout (not just FP8/FP4); `quant_block_size` assertion tightened to `== 128`; DeepGEMM-source reference comment on the FP4 magic constants; 8-byte base-pointer `TORCH_CHECK` (with pointer in the message) on `fused_cat_fp4Op`; refreshed `_call_mqa_logits` comment post q_scale reshape fix; copyright year bumps on `dataTransceiverState.h`, `cacheTransceiver.cpp`, `serialization.cpp`.

## Test Coverage

Bit-exactness of the new CUDA op:
- `tests/unittest/_torch/attention/sparse/test_cpp_custom_ops.py::test_fused_cat_fp4_matches_deepgemm` — parametrized over shapes `(4, 128)`, `(1, 32, 128)`, `(3, 7, 128)`, `(2, 5, 4, 128)` and seeds `{0, 42, 2026}`. Uses `torch.equal` (not `allclose`) on both packed bytes and scale int32 vs `tensorrt_llm.deep_gemm.utils.math.per_token_cast_to_fp4(..., use_ue8m0=True, gran_k=32, use_packed_ue8m0=True)` — all 12 parametrizations pass.
- `test_fused_cat_fp4_shape_dispatch[64-64 / 32-96 / 16-112 / 96-32]` — asymmetric `pe`/`nope` splits, shape and dtype sanity.
- `test_fused_cat_fp4_noncontiguous_split` — op accepts non-contiguous views from `torch.split()` and matches the contiguous baseline.

Config plumbing:
- `test_dsa_fp4_indexer.py::test_indexer_k_dtype_survives_model_config_rebuild` — new. Exercises the `ModelConfig.from_pretrained` DSV3.2 rebuild branch with a stub pretrained config, asserts `indexer_k_dtype="fp4"` round-trips, and statically greps `ModelConfig.from_pretrained`'s source for the forwarding keyword so a future edit that drops it fails fast.

FP4 indexer end-to-end on B200:
- `test_fp4_mqa_logits_shape_and_topk_intersection[32|64]` — FP4 vs FP8 top-k overlap.
- `test_fp4_quantize_roundtrip_matches_bf16_kv` — packing / scale recovery round-trip.
- `test_fp4_indexer_k_cache_per_token_size_drops_to_68_bytes` — per-token KV-cache shrinkage from 132B to 68B.

FP8 indexer regression:
- `test_dsa_indexer.py` — FP8 path unregressed; all parametrizations including `next_n ∈ {1,2,3,4}` now pass with the base scheduler metadata buffer.
- `test_cpp_custom_ops.py` existing gather / scatter / convert_req_index tests still pass.

Full attention sweep:
- `tests/unittest/_torch/attention` on B200 — no regressions from any of the four commits.

FP8 DSA smoke on a real model:
- `tests/integration/defs/accuracy/test_llm_api_pytorch.py::TestDeepSeekV32::test_fp8_blockscale[baseline]` launched end-to-end on 8 × B200 with `DeepSeek-V3.2-Exp-hf`; model loaded and forward passes proceeded without error, confirming the FP8 `fused_cat_fp8` path is unaffected (stopped after ramp-up to save time).

## PR Checklist

## GitHub Bot Help


Barry-Delaney:
The 2 yaml files need to be updated in https://github.com/NVIDIA/TensorRT-LLM/tree/main/scripts/attribution/data.
We can re-run the [script](https://github.com/NVIDIA/TensorRT-LLM/blob/main/scripts/attribution/README.md?plain=1#L17) to update it.

cpp/tensorrt_llm/deep_gemm/CMakeLists.txt
```diff
@@ -40,6 +40,9 @@ foreach(SOURCE_FILE ${DEEP_GEMM_ALL_FILES})
     file(READ ${SOURCE_FILE} _content)
     string(REPLACE "from . import _C" "import tensorrt_llm.deep_gemm_cpp_tllm"
                    _content "${_content}")
+    string(REPLACE "from .. import _C"
+                   "import tensorrt_llm.deep_gemm_cpp_tllm as _C" _content
```
Barry-Delaney: Should be replacing `from .. import _C` into `import tensorrt_llm.deep_gemm_cpp_tllm`.
The following part will call `tensorrt_llm.deep_gemm_cpp_tllm.foo()`, current replace will resulting in import errors.

cpp/tensorrt_llm/nanobind/batch_manager/kvCacheManager.cpp
```diff
@@ -633,7 +633,8 @@ void tb::kv_cache_manager::KVCacheManagerBindings::initBindings(nb::module_& m)
             nb::arg("enable_partial_reuse") = true, nb::arg("copy_on_partial_reuse") = true,
             nb::arg("kv_connector_manager") = nullptr, nb::arg("enable_indexer_k_cache") = false,
             nb::arg("indexer_k_cache_quant_block_size") = 128, nb::arg("indexer_k_cache_index_head_dim") = 0,
-            nb::arg("linear_attention_metadata").none() = std::nullopt, nb::call_guard<nb::gil_scoped_release>())
+            nb::arg("indexer_k_cache_use_fp4") = false, nb::arg("linear_attention_metadata").none() = std::nullopt,
```
reasonsolo: It's better to append the binding arg list only when there are default values , I have encountered some misaligned arg-vs-param in similar cases.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1043,17 +1047,34 @@ def _mla_dsa_proj_fake(
...
-        dtype=torch.float8_e4m3fn)
-    k_fp8 = hidden_states.new_empty([num_tokens, indexer.head_dim],
-                                    dtype=torch.float8_e4m3fn)
-    k_scale = hidden_states.new_empty([num_tokens, 1], dtype=torch.float32)
+    # Indexer intermediates: q_fp8, k_fp8, k_scale, weights, q_scale.
+    # Under FP4 q_fp8's trailing dim is head_dim // 2 (two E2M1 codes per
+    # byte) and q_scale carries one int32 per (token, head) packing four
+    # UE8M0 exponents; under FP8 q_fp8's trailing dim is head_dim and
+    # q_scale carries one float32 per (token, head).
+    if getattr(indexer, "use_fp4", False):
```
reasonsolo: I think it's better to use indexer.use_fp4 explicity than getattr and fallback to False silently. This makes me feel that we are hiding config/api changes in implementation details.

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
```diff
@@ -1105,6 +1142,12 @@ def __init__(self,
         self.softmax_scale = self.head_dim**-0.5
         # TODO: make it configurable from hf config
         self.scale_fmt = "ue8m0"
+        # indexer_k_dtype controls both Q and K precision. DeepGEMM's
+        # fp8_fp4_mqa_logits / fp8_fp4_paged_mqa_logits kernels only dispatch
+        # to FP4xFP4 or FP8xFP8 (no mixed-precision variant). The DeepGEMM
+        # kernel asserts SM100 + head_dim=128 at launch time under FP4.
+        self.use_fp4 = getattr(sparse_attention_config, "indexer_k_dtype",
```
reasonsolo: Same here, why not add the `indexer_k_dtype` with default `fp8` to the config itself?

tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
```diff
@@ -1151,8 +1161,9 @@ def _(req_id: torch.Tensor, block_table: torch.Tensor,
 
     @torch.library.register_fake("trtllm::indexer_k_cache_gather_op")
     def _(k_cache: torch.Tensor, slot_mapping_fp8: torch.Tensor,
-          slot_mapping_scale: torch.Tensor, k_token_start: int,
-          num_tokens: int) -> Tuple[torch.Tensor, torch.Tensor]:
-        k_fp8 = k_cache.new_empty([num_tokens, 128], dtype=torch.float8_e4m3fn)
+          slot_mapping_scale: torch.Tensor, k_token_start: int, num_tokens: int,
+          head_dim: int) -> Tuple[torch.Tensor, torch.Tensor]:
+        k_fp8 = k_cache.new_empty([num_tokens, head_dim],
```
hyukn, line 1186: NITS: It is a little confusing to use k_fp8 still for the fp4 path. And I guess the head dim means the same thing as "payload bytes per token", as described in the cpp.

hyukn (approved):

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -347,6 +347,42 @@ class DeepSeekSparseAttentionConfig(BaseSparseAttentionConfig):
         "Whether to reuse previous step's TopK indices as heuristic hints "
         "for the decode indexer TopK kernel, reducing threshold search iterations."
     )
+    indexer_k_dtype: Literal["fp8", "fp4"] = Field(
+        default="fp8",
+        description=
+        "Data type used for the indexer K cache. Use `fp4` on Blackwell+ "
+        "(SM>=100) to halve the indexer K cache per-token footprint from "
+        "132 B to 68 B at index_head_dim=128. Requires index_head_dim=128.",
```
yuxianq: "Requires index_head_dim=128." is ambiguous, only fp4 requires it, we can say "`fp4` requires Blackwell+ (SM>=100) and index_head_dim=128, it can halve the indexer K cache per-token footprint from 132 B to 68 B."

tensorrt_llm/_torch/modules/attention.py
dc3671: done and I also removed getattr for kvcache_manager.

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
dc3671: done and I also removed getattr for kvcache_manager.

tensorrt_llm/llmapi/llm_args.py
dc3671: reasonable for me. Done

tests/integration/defs/accuracy/test_llm_api_pytorch.py
```diff
@@ -3241,26 +3241,29 @@ def test_nvfp4_multi_gpus_piecewise_cuda_graph(self, tp_size, pp_size,
...
+            (8, 1, 8, 0, True, True, True, True, 24, "CUTLASS", False, False),
+            (8, 1, 8, 3, False, False, True, True, 1, "TRTLLM", False, False),
+            (8, 1, 8, 1, False, True, True, True, 24, "CUTLASS", True, False),
+            (1, 4, 1, 1, False, False, True, True, 24, "TRTLLM", False, False),
+            (8, 1, 8, 0, False, True, True, True, 24, "CUTLASS", False, True),
         ],
         ids=[
             "baseline", "baseline_mtp1", "baseline_fp8kv", "latency",
-            "disable_skip_indexer", "baseline_pp4_mtp1"
+            "disable_skip_indexer", "baseline_pp4_mtp1", "fp4_indexer"
```
yuxianq, line 3450: We do not add `accuracy/test_llm_api_pytorch.py::TestDeepSeekV32::test_nvfp4_multi_gpus[fp4_indexer]` to any test list, please add it to cover this test in CI.

cpp/tensorrt_llm/nanobind/batch_manager/kvCacheManager.cpp
dc3671: As dicussed offline, Lizhi recommended just append new arg to the end to avoid positional args mismatch. But it will always use `self.impl = KVCacheManagerCpp(**kwargs)` to avoid this problem. So just keep the order to make indexer related args together.

tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
dc3671: As discussed offline, k_fp8 always hold fp8 tensor in cpp side. And there are lots of `_fp8` vars in cpp side. Will keep the name with additional comment to demonstrate the usage.

reasonsolo (approved):

tests/integration/defs/accuracy/test_llm_api_pytorch.py
dc3671: done

juney-nvidia (approved):
Approved from oss compliance perspective.

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
```diff
@@ -1396,10 +1449,12 @@ def prepare(metadata: DSAtrtllmAttentionMetadata):
...
             head_dim = kv_cache_manager.index_head_dim
             quant_block_size = kv_cache_manager.quant_block_size
+            use_fp4 = kv_cache_manager.use_fp4
+            data_bytes_per_token = head_dim // 2 if use_fp4 else head_dim
             cached_tokens = metadata.kv_cache_params.num_cached_tokens_per_seq
             scale_size = head_dim // quant_block_size * 4
-            tokens_per_block * (head_dim + scale_size)
-            tokens_per_block * head_dim
+            tokens_per_block * (data_bytes_per_token + scale_size)
+            tokens_per_block * data_bytes_per_token
```
yuxianq: These two line does not store to variable, should we remove them?

tests/integration/test_lists/test-db/l0_dgx_b200.yml
```diff
@@ -183,6 +183,7 @@ l0_dgx_b200:
   - accuracy/test_llm_api_pytorch.py::TestDeepSeekV32::test_nvfp4_multi_gpus[baseline_fp8kv] TIMEOUT (60)
   - accuracy/test_llm_api_pytorch.py::TestDeepSeekV32::test_nvfp4_multi_gpus[latency] TIMEOUT (60)
   - accuracy/test_llm_api_pytorch.py::TestDeepSeekV32::test_nvfp4_multi_gpus[disable_skip_indexer] TIMEOUT (60)
+  - accuracy/test_llm_api_pytorch.py::TestDeepSeekV32::test_nvfp4_multi_gpus[fp4_indexer] TIMEOUT (60)
```
yuxianq, line 186: Please also add to tests/integration/test_lists/qa/llm_function_core.txt for QA.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -971,10 +975,15 @@ void WindowBlockManager::createIndexerKCachePools()
         {
             continue;
         }
+        // scaleSize evaluates to 4 at indexHeadDim=128 / quantBlockSize=128 for both
+        // FP8 (float32 scale) and FP4 (packed UE8M0 x4 stored as one int32). The
+        // data size is either indexHeadDim bytes (one FP8 byte per element) or
+        // indexHeadDim / 2 bytes (two packed FP4 E2M1 codes per byte).
         SizeType32 scaleSize = mIndexerKCacheIndexHeadDim / mIndexerKCacheQuantBlockSize * 4;
+        SizeType32 dataSize = mIndexerKCacheUseFp4 ? mIndexerKCacheIndexHeadDim / 2 : mIndexerKCacheIndexHeadDim;
```
yuxianq, line 1042: We should also do the same thing for cache transmission. see `computeDimsPerHead` in https://github.com/NVIDIA/TensorRT-LLM/blob/aea7e8e11306225a69e13f2da1a0c2a7f16be916/cpp/tensorrt_llm/executor/cache_transmission/cacheSplitConcat.cu#L56 and https://github.com/NVIDIA/TensorRT-LLM/blob/c69c956326cce44f8b070353ca675b05d83aca5b/cpp/tests/unit_tests/multi_gpu/cacheTransceiverTest.cpp#L1088 and https://github.com/NVIDIA/TensorRT-LLM/blob/c69c956326cce44f8b070353ca675b05d83aca5b/cpp/tests/unit_tests/multi_gpu/cacheTransceiverTest.cpp#L1182

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
yuxianq: The annotation should be `Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]` 5 tuples now.

tests/unittest/_torch/attention/sparse/test_dsa_indexer.py
yuxianq, line 1: Should we add a `indexer_k_cache_scatter_op` test for fp4 case?

tensorrt_llm/_torch/modules/attention.py
yuxianq: Update doc, we have 5 elements now.

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
```diff
@@ -965,13 +991,22 @@ def on_update_kv_lens(self):
...
             global_positions = start_positions[req_indices] + token_offsets
+            # Under FP4 the indexer cache stores two E2M1 codes per byte, so
+            # the per-token data footprint is head_dim // 2; otherwise it is
+            # head_dim (one FP8 byte per element). Feed the real byte count
+            # into _compute_slot_mappings so scatter/gather see offsets that
+            # match the pool layout produced by createIndexerKCachePools.
+            use_fp4 = self.kv_cache_manager.use_fp4
+            index_head_dim = self.kv_cache_manager.index_head_dim
+            data_bytes_per_token = index_head_dim // 2 if use_fp4 else index_head_dim
             fp8_indices, scale_indices = _compute_slot_mappings(
```
yuxianq, line 1020: Will we change the name of `fp8_indices` since it supports fp8/fp4 now? There are many places that still use fp8 prefix for fp8+fp4 variable name.

tests/unittest/_torch/attention/sparse/test_dsa_fp4_indexer.py
```diff
@@ -0,0 +1,304 @@
...
+import tensorrt_llm  # noqa: F401
+
+try:
+    from tensorrt_llm import deep_gemm
+except ImportError:
+    # Only skip on actual module-missing failures — any other error should
+    # surface rather than silently turn into a test-wide skip.
+    HAS_DEEP_GEMM = False
+else:
+    HAS_DEEP_GEMM = hasattr(deep_gemm, "fp8_fp4_mqa_logits")
```
yuxianq: We skip DeepGEMM tests when we cannot find `fp8_fp4_mqa_logits`, but this case indicates that we use a wrong version of DeepGEMM, should we fail the test instead of skipping it?

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
```diff
@@ -1538,12 +1649,17 @@ def sparse_attn_indexer(
                     global_q_start = chunk.token_start + chunk_q_start
                     global_q_end = chunk.token_start + chunk_q_end
 
-                    logits = fp8_mqa_logits(
+                    chunk_q_scale = q_scale[
+                        global_q_start:global_q_end,
+                        ...] if q_scale is not None else None
```
yuxianq: We always provide not-none `q_scale` in non-test cases, so we cannot skip `chunk_q_scale` in fp8 case, how about use `if self.use_fp4` as condition to skip it?

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
```diff
@@ -1588,13 +1704,16 @@ def sparse_attn_indexer(
                 cu_seqlen_ks = metadata.cu_seqlen_ks[:num_ctx_tokens]
                 cu_seqlen_ke = metadata.cu_seqlen_ke[:num_ctx_tokens]
 
-                logits = fp8_mqa_logits(
+                ctx_q_scale = q_scale[:num_ctx_tokens,
+                                      ...] if q_scale is not None else None
```
yuxianq: Same issue, how about use `if self.use_fp4` as condition?

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
```diff
@@ -1663,11 +1784,19 @@ def sparse_attn_indexer(
...
                 self.layer_idx)
 
-            logits_decode = fp8_paged_mqa_logits(q_decode, k_cache,
-                                                 weights_decode, context_lens,
-                                                 block_table,
-                                                 scheduler_metadata_buffer,
-                                                 max_seq_len)
+            decode_q_scale = q_scale[num_ctx_tokens:num_ctx_tokens +
+                                     num_gen_tokens,
+                                     ...] if q_scale is not None else None
```
yuxianq: Same issue, how about use `if self.use_fp4` as condition?

yuxianq (approved):

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
dc3671: As discussed offline and previous discussion with Yukun, we will not do this in this PR.

StanleySun639 (approved):

Barry-Delaney (approved):

PR #13347, opened 2026-04-22
[https://nvbugs/6093911][fix] Fix disagg gen-only benchmark hang under ADP router imbalance

chienchunhung:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved benchmark disaggregation fill gate to determine readiness based on executor state rather than request counts, ensuring more accurate synchronization across distributed ranks.
  * Enhanced diagnostic logging to report transfer status and request state information.

* **Tests**
  * Updated and expanded test coverage for state-based readiness validation, including new regression tests for distributed training scenarios.

## Summary

Fixes four ADP/NIXL regressions PR #12208 introduced when it removed the blocking fill loop, and corrects a long-latent KV oversubscription in the Kimi-2 8k1k perf-sanity test config that #12208's fill-phase changes exposed.

## Production fixes (2 commits)

PR #12208 replaced the blocking fill loop with a non-blocking gate to fix an MPI hard deadlock. The replacement was correct for MPI but introduced four ADP/NIXL-side regressions, all addressed here:

- **State-based fill gate** — count-based predicate is unsatisfiable under ADP per-rank rebalancing. Replaced with `(num_fetched ≥ queue_size) ∧ (no active req in INIT/TRANS) ∧ (no transceiver pending)`. Fires uniformly across ranks.
- **ADP router hard cap** at `max_num_active_requests` (= per-rank `max_batch_size`) on both `DefaultADPRouter` and `KVCacheAwareADPRouter` — composes with upstream's `fair_share_multiplier` loose cap. Fixes nvbug 6071070; without it, bulk admission can route past `max_batch_size` and leave excess requests stuck in `INIT`.
- **Fail-fast suppression during fill** — PR #12206's fail-fast was tripping on transient bulk-fetch state (some requests in INIT while others transfer). Resumes after the gate opens.
- **`tp_size`-per-iter admission throttle** in `_pop_from_waiting_queue` — restores the pre-#12208 conservative admission rate without reintroducing the blocking pattern. Scoped to non-warmup benchmark fill.

### Memory math (B200/GB200, 184 GiB HBM)

|  | 0.6 (upstream) | 0.78 (this PR) |
|---|---|---|
| KV pool | 61.4 GiB | 79.8 GiB |
| KV tokens / rank | 1,875,008 | 2,437,504 |
| Required at `bs=256 × max_seq_len=9256` | 2,369,536 | (same) |
| Sequences fit / 256 | **203** ❌ | **263** ✅ |
| Headroom for transients | ~40 GiB | ~22 GiB |

At 0.6, only 203 of 256 max-batch sequences fit per rank — 53 are perpetually paused, costing ~21 % of decode throughput across the entire run. 0.78 is the floor that fits the workload while preserving the ~22 GiB needed for NIXL receive buffers, NCCL workspace, EPLB layer churn, and CUDA caching-allocator fragmentation.

### Why the test was passing before PR #12208

Same YAML, same oversubscription — but timing differed:

- **Pre-#12208:** count-based fetch loop completed in ~26 s; decode then ran for ~85–88 min under the chronic preemption; total wall landed just inside the test's 5400 s `DEFAULT_TIMEOUT` with 2–5 min of cluster-variance-dependent margin.
- **Post-#12208 (and this PR):** state-based fill flag remains active for ~84 min (tracks CTX prefill completion, which is the dominant cost when `concurrency = capacity`). The flag adds only ~50 s of `all_gather` overhead, but it ate the cluster-variance margin that had been hiding the oversubscription.

The test was always running on luck. Bumping to 0.78 makes it do what it claims (`bs=256` steady-state) instead of depending on the right side of a 2–5 min margin.

### Why this YAML change is bundled into this PR

The bump is a one-line test-config correctness fix that stands on its own — it would have been correct at any point in the test's lifetime. It belongs here because (a) the gate fixes are necessary but not sufficient for the Kimi-2 CI to reach green, and (b) the two changes share the same failure narrative.

## Test coverage

Production fixes are covered by `tests/unittest/_torch/executor/test_benchmark_disagg.py` (10 test classes, 28 tests, no GPU required):

| Production fix | Coverage |
|---|---|
| State-based fill gate | `TestFillCompleteStateBased` (4 tests, parameterized over the predicate's truth table), `TestFillCompleteADP` (3 tests, cross-rank `allgather`), `TestFillCompleteADPRouterImbalance` (2 tests, per-rank rebalancing edge cases), `TestCheckBenchmarkDisaggGate` (3 tests, wait-loop behavior), `TestCanForwardGating` (parameterized initial state) |
| ADP router hard cap | `TestADPRouterPerRankCap` (3 tests, parameterized over both `DefaultADPRouter` and `KVCacheAwareADPRouter`: cap applied at max, no rank exceeds, no-op below max) |
| Fail-fast suppression during fill | `TestFailFastSuppressedDuringFill` (3 tests: suppressed during fill, resumes after, full `(fill_active × is_warmup)` matrix) |
| `tp_size`-per-iter admission throttle | `TestBenchmarkFillAdmissionFlowControl` (2 tests, parameterized over `tp_size ∈ {1, 4, 32}`: cap applied during fill, full capacity used after) |
| Adjacent invariants | `TestPadAttentionDpDummyBenchmarkDisagg` (4 tests, dummy-request skip during fill), `TestPrepareAndScheduleBatchNoBlock` (1 test, fetch called once per iter) |
| End-to-end lifecycle | `TestFillPhaseEndToEnd::test_full_lifecycle` (admission-cap → KV-transfer-complete → gate-opens → normal admission, exercising both production code paths together) |

The Kimi-2 perf-sanity test has been verified separately: https://nvbugspro.nvidia.com/bug/6093911?commentNumber=9

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -2220,7 +2220,14 @@ class PyExecutor:
             # scheduler could not allocate KV for any of them, the benchmark
             # will hang forever because in-progress generation requests won't
             # release their KV cache.
+            #
+            # Suppress during the fill phase: INIT requests are expected
+            # while KV transfers are still in progress.  The state-based
+            # gate (_is_benchmark_disagg_fill_complete) will open once all
+            # transfers complete; only after that does a stuck INIT request
+            # indicate genuine KV insufficiency.
             if (self.benchmark_req_queues_size > 0 and not self.is_warmup
+                    and not self._benchmark_fill_phase_active
                     and not fitting_disagg_gen_init_requests):
                 stuck_init_requests = [
                     req for req in self.active_requests
@@ -2273,42 +2280,88 @@ class PyExecutor:
 
     def _is_benchmark_disagg_fill_complete(
             self, scheduled_batch: ScheduledRequests) -> bool:
-        """Check whether all benchmark disagg requests have completed KV transfer.
+        """State-based fill-complete predicate for benchmark disagg mode.
 
-        With ADP, generation requests are distributed across TP ranks, so an
-        allgather is needed to obtain the global count.  Without ADP every
-        request is local, and we can compare directly.
+        The gate opens when all three conditions hold globally:
+
+        (A) The executor has fetched at least ``benchmark_req_queues_size``
+            requests cumulatively.
+        (B) Every request in ``active_requests`` on this rank is past the
+            KV-transfer phase (not in INIT or TRANS_IN_PROGRESS).
+        (C) The KV cache transceiver has no pending receive sessions
+            (no transfers about to complete that would change state).
+
+        For ADP, (B) and (C) are AND-ed across TP ranks via allgather.
+
+        This predicate is immune to ADP router distribution skew: it asks
+        "are all admitted requests ready for generation?" not "did the
+        router put >= threshold/tp_size on every rank?"
 
         This method must only be called when ``is_benchmark_disagg`` is True.
 
         Args:
-            scheduled_batch: The current iteration's scheduled requests,
-                used to count generation requests that have completed
-                KV transfer.
+            scheduled_batch: Passed for API compatibility with callers
+                but no longer used by this predicate.
 
         Returns:
-            True when the total number of generation-ready requests
-            reaches ``benchmark_req_queues_size``.
+            True when the fill phase is complete and the first forward
+            pass can proceed.
         """
         if not self.is_benchmark_disagg:
             raise RuntimeError(
-                "_is_benchmark_disagg_fill_complete() should not be called outside benchmark "
-                "disagg mode.  This is an unexpected error.")
-        local_gen_count = sum(1 for req in scheduled_batch.generation_requests
-                              if not req.is_attention_dp_dummy)
-        if self.enable_attention_dp:
-            total_gen_count = sum(self.dist.tp_allgather(local_gen_count))
-        else:
-            total_gen_count = local_gen_count
+                "_is_benchmark_disagg_fill_complete() should not be called "
+                "outside benchmark disagg mode.")
 
-        if total_gen_count >= self.benchmark_req_queues_size:
-            return True
-        if self.dist.rank == 0:
+        # (A) All benchmark requests have been fetched from the queue.
+        if self.num_fetch_requests < self.benchmark_req_queues_size:
+            if self.dist.rank == 0:
+                logger.debug(
+                    f"Benchmark disagg fill: fetching "
+                    f"{self.num_fetch_requests}/{self.benchmark_req_queues_size}"
+                )
+            return False
+
+        # (B) Every active request on this rank is past KV-transfer states.
+        local_all_past_transfer = not any(
+            req.is_disagg_generation_init_state
+            or req.is_disagg_generation_transmission_in_progress
+            for req in self.active_requests)
+
+        # (C) No pending receive sessions on the transceiver.
+        local_no_inflight = (
+            self.kv_cache_transceiver is None
+            or self.kv_cache_transceiver.check_gen_transfer_complete())
+
+        local_ok = int(local_all_past_transfer and local_no_inflight)
+
+        if self.enable_attention_dp:
+            all_ranks_ok = self.dist.tp_allgather(local_ok)
+            global_ok = min(all_ranks_ok) == 1
+        else:
+            all_ranks_ok = None
+            global_ok = bool(local_ok)
+
+        if not global_ok and self.dist.rank == 0:
+            blocked_ranks = [
+                rank for rank, ok in enumerate(all_ranks_ok or [local_ok])
+                if not ok
+            ]
+            num_init = sum(1 for req in self.active_requests
+                           if req.is_disagg_generation_init_state)
+            num_in_progress = sum(
+                1 for req in self.active_requests
+                if req.is_disagg_generation_transmission_in_progress)
             logger.debug(
-                f"Benchmark disagg fill in progress: "
-                f"num_fetched={self.num_fetch_requests}, "
-                f"total_gen_count={total_gen_count} (local={local_gen_count})")
-        return False
+                f"Benchmark disagg fill: blocked on ranks {blocked_ranks} "
+                f"(rank {self.dist.rank} local: {num_init} INIT, "
+                f"{num_in_progress} in-progress, "
+                f"inflight={not local_no_inflight})"
+            )
+        if global_ok and self.dist.rank == 0:
+            logger.info(f"Benchmark disagg fill complete: "
+                        f"{len(self.active_requests)} active requests ready, "
+                        f"gate opening.")
+        return global_ok
 
     def _check_benchmark_disagg_gate(self, scheduled_batch: ScheduledRequests,
                                      can_forward: bool) -> tuple[bool, bool]:
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -2355,8 +2355,7 @@ class PyExecutor:
                 f"Benchmark disagg fill: blocked on ranks {blocked_ranks} "
                 f"(rank {self.dist.rank} local: {num_init} INIT, "
                 f"{num_in_progress} in-progress, "
-                f"inflight={not local_no_inflight})"
-            )
+                f"inflight={not local_no_inflight})")
         if global_ok and self.dist.rank == 0:
             logger.info(f"Benchmark disagg fill complete: "
                         f"{len(self.active_requests)} active requests ready, "
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -3110,6 +3110,19 @@ class PyExecutor:
 
         max_new_requests = total_max - total_num_active_requests
 
+        # Benchmark disagg fill-phase admission control. Without this cap, the
+        # executor can admit up to `total_max` requests in a single iteration as
+        # soon as the benchmark queue is preloaded, which spikes peak KV-cache
+        # reservations + recv-buffer reservations and can OOM-kill the GEN
+        # server before any KV transfer drains.  The `tp_size` cap restores the
+        # conservative per-rank rate that PR #12091 had in its blocking fill
+        # loop: one new request per rank per executor iteration.  It is scoped
+        # to the initial fill only; once the gate fires, normal admission
+        # resumes.
+        if (self.is_benchmark_disagg and self._benchmark_fill_phase_active
+                and not self.is_warmup):
+            max_new_requests = min(max_new_requests, self.dist.tp_size)
+
         return get_from_waiting_queue(
             waiting_queue,
             max_new_requests,
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -3099,12 +3099,12 @@ class PyExecutor:
 
         max_new_requests = total_max - total_num_active_requests
 
-        # Benchmark disagg fill-phase admission throttle: without it, the
-        # executor admits `total_max` requests in one iteration when the
-        # benchmark queue is preloaded, putting that many KV recvs in
-        # flight at once.  The transient transfer-staging peak can OOM-
-        # kill the GEN server before any transfer drains.  Cap at
-        # `tp_size` (one per rank per iteration) until the fill gate opens.
+        # Benchmark disagg fill-phase admission throttle: a preloaded queue
+        # (concurrency == total_max) would otherwise pop all requests into
+        # DISAGG_GENERATION_INIT in one iter, tripping PR #12206's fail-fast
+        # under ADP-router imbalance and growing the recv-buffer fallback
+        # faster than transfers drain.  Cap at `tp_size` (≈ pre-#12208
+        # blocking fill rate).  Verified on Kimi-K2 8k1k ctx8/gen1 con=8192.
         if (self.is_benchmark_disagg and self._benchmark_fill_phase_active
                 and not self.is_warmup):
             max_new_requests = min(max_new_requests, self.dist.tp_size)
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -2216,18 +2216,12 @@ class PyExecutor:
                     self._check_disagg_ctx_cache_transfer_status(0)
 
             # In gen-only benchmark mode, all requests must fit in KV cache
-            # simultaneously. If some requests are stuck in INIT state and the
-            # scheduler could not allocate KV for any of them, the benchmark
-            # will hang forever because in-progress generation requests won't
-            # release their KV cache.
-            #
-            # Suppress during the fill phase: INIT requests are expected
-            # while KV transfers are still in progress.  The state-based
-            # gate (_is_benchmark_disagg_fill_complete) will open once all
-            # transfers complete; only after that does a stuck INIT request
-            # indicate genuine KV insufficiency.
+            # simultaneously because generation requests do not release their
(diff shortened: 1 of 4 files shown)


pcastonguay:
Some tests seem to fail due to your changes. For example:
```

[2026-04-30T01:31:33.474Z] FAILED unittest/DGX_B200-PyTorch-2/unittest/_torch/executor/test_benchmark_disagg.py::TestADPRouterPerRankCap::test_cap_not_applied_when_below_max[KVCacheAwareADPRouter] - assert 4 == 2
```
Also the release-check job is failing. Could you take a look?

chienchunhung:
> Some tests seem to fail due to your changes. For example:
> 
> ```
> 
> [2026-04-30T01:31:33.474Z] FAILED unittest/DGX_B200-PyTorch-2/unittest/_torch/executor/test_benchmark_disagg.py::TestADPRouterPerRankCap::test_cap_not_applied_when_below_max[KVCacheAwareADPRouter] - assert 4 == 2
> ```
>

Some of the new tests introduced by this PR failed after rebasing to the latest HEAD; updated the tests. 
 
> Also the release-check job is failing. Could you take a look?

Yup over the past few days I have been working closely with QA on the Kimi2 test.

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
```diff
@@ -536,15 +544,22 @@ def _sort_key(req_item):
 
         # Loose cap at fair_share_multiplier * ceil(total / tp_size): safety
         # net against runaway concentration, still loose enough that cache
-        # affinity wins in the common case.
+        # affinity wins in the common case.  Hard-cap additionally at
+        # max_num_active_requests — same guard as DefaultADPRouter
+        # (nvbug 6071070); critical for benchmark disagg fill where bulk
```
brb-nv: I would suggest that we leave out details specific to this bug. Won't make much sense outside the context of this MR.

brb-nv: Same with other mentions of the bug in this MR.

tests/integration/defs/perf/disagg/test_configs/wideep/perf/kimi-k2-thinking-fp4_8k1k_ctx8_gen1_dep32_bs256_eplb416_mtp0_ccb-NIXL.yaml
```diff
@@ -73,7 +73,22 @@ worker_config:
...
+      # GEN KV memory budget — DO NOT lower below 0.78 without also lowering
+      # max_batch_size or max_seq_len.
+      #
+      # Floor (0.78): minimum fraction that fits max_batch_size=256 at
+      # max_seq_len=9256 (~2.37M KV tokens/rank). Below this the scheduler
+      # cannot keep all 256 ADP-routed slots resident and chronically
+      # preempts (e.g. at 0.6 only ~203/256 fit, costing ~21% of decode
+      # throughput and pushing the 90-min wall over the test timeout).
+      #
+      # Ceiling (~0.85): leaves ~15 GiB headroom on B200/GB200 (184 GiB).
```
brb-nv: Nit: Please consider skipping adding this much detail (or any detail). Might not make much sense beyond the scope of this MR.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2935,6 +2987,19 @@ def _pop_from_waiting_queue(
 
         max_new_requests = total_max - total_num_active_requests
 
+        # Benchmark disagg fill-phase admission control. Without this cap, the
+        # executor can admit up to `total_max` requests in a single iteration as
+        # soon as the benchmark queue is preloaded, which spikes peak KV-cache
+        # reservations + recv-buffer reservations and can OOM-kill the GEN
+        # server before any KV transfer drains.  The `tp_size` cap restores the
+        # conservative per-rank rate that PR #12091 had in its blocking fill
```
brb-nv: PR details won't make much sense beyond the scope of this MR. Let's remove it and make the comments brief, if possible.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2190,42 +2197,87 @@ def _kv_connector_wait_for_save(self):
...
+        (A) The executor has fetched at least ``benchmark_req_queues_size``
+            requests cumulatively.
+        (B) Every request in ``active_requests`` on this rank is past the
+            KV-transfer phase (not in INIT or TRANS_IN_PROGRESS).
+        (C) The KV cache transceiver has no pending receive sessions
+            (no transfers about to complete that would change state).
+
+        For ADP, (B) and (C) are AND-ed across TP ranks via allgather.
+
+        This predicate is immune to ADP router distribution skew: it asks
```
brb-nv: Nit: Excess detail.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2190,42 +2197,87 @@ def _kv_connector_wait_for_save(self):
...
+                )
+            return False
+
+        # (B) Every active request on this rank is past KV-transfer states.
+        local_all_past_transfer = not any(
+            req.is_disagg_generation_init_state
+            or req.is_disagg_generation_transmission_in_progress
+            for req in self.active_requests)
+
+        # (C) No pending receive sessions on the transceiver.
```
brb-nv: Dumb question: Aren't B & C redundant? i.e. if all requests are past KV-transfer stage, is there a case where kv_cache_transceiver say otherwise?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2190,42 +2197,87 @@ def _kv_connector_wait_for_save(self):
...
 
-        if total_gen_count >= self.benchmark_req_queues_size:
-            return True
-        if self.dist.rank == 0:
+        if not global_ok and self.dist.rank == 0:
+            blocked_ranks = [
+                rank for rank, ok in enumerate(all_ranks_ok or [local_ok])
+                if not ok
+            ]
+            num_init = sum(1 for req in self.active_requests
```
brb-nv: Dumb question: Do we need to mention `else 0`? Or it just skips adding any entry in the else case and the math works out ok?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2190,42 +2197,87 @@ def _kv_connector_wait_for_save(self):
...
-            total_gen_count = local_gen_count
+            all_ranks_ok = None
+            global_ok = bool(local_ok)
 
-        if total_gen_count >= self.benchmark_req_queues_size:
-            return True
-        if self.dist.rank == 0:
+        if not global_ok and self.dist.rank == 0:
+            blocked_ranks = [
+                rank for rank, ok in enumerate(all_ranks_ok or [local_ok])
```
brb-nv: Can we just make `all_ranks_ok=[local_ok]` for TP above so that we can simply enumerate over `all_ranks_ok` here?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2190,42 +2197,87 @@ def _kv_connector_wait_for_save(self):
...
+            global_ok = min(all_ranks_ok) == 1
         else:
-            total_gen_count = local_gen_count
+            all_ranks_ok = None
+            global_ok = bool(local_ok)
 
-        if total_gen_count >= self.benchmark_req_queues_size:
-            return True
-        if self.dist.rank == 0:
+        if not global_ok and self.dist.rank == 0:
```
brb-nv: Can we merge the two if-conditionals here over `dist.rank==0` and then check for `global_ok` under that?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2935,6 +2987,19 @@ def _pop_from_waiting_queue(
...
+        # soon as the benchmark queue is preloaded, which spikes peak KV-cache
+        # reservations + recv-buffer reservations and can OOM-kill the GEN
+        # server before any KV transfer drains.  The `tp_size` cap restores the
+        # conservative per-rank rate that PR #12091 had in its blocking fill
+        # loop: one new request per rank per executor iteration.  It is scoped
+        # to the initial fill only; once the gate fires, normal admission
+        # resumes.
+        if (self.is_benchmark_disagg and self._benchmark_fill_phase_active
+                and not self.is_warmup):
+            max_new_requests = min(max_new_requests, self.dist.tp_size)
```
brb-nv, line 3110: Will this not significantly raise the wait time for large batch size?

Also, KV cache reservations (determined by num_admitted_requests) and recv-buffer reservations (determined by cache_transceiver_config.max_tokens_in_buffer) would be constant right - with out without this careful regulation? So, how does this regulation help? Very likely I'm missing something.

brb-nv (commented):
Left some comments.

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Good question — let me try to answer both halves.

**Wait time at large batch.** The throttle only fires during the initial fill phase; once the gate opens, `_benchmark_fill_phase_active` flips False and admission returns to normal `total_max`. During fill, no forward pass runs (i.e., the gate is closed waiting for KV transfers), so each iteration is short. For our worst case (Kimi-2: concurrency=8192, tp_size=32), fill needs ~256 additional admission iterations. Those iterations are lightweight and are dwarfed by CTX prefill time, which gates the same window anyway. We have not seen fill-phase wait time become a wall-time issue in any benchmark size we test.

**Why the throttle helps when KV reservations are constant.** You're right that the steady-state KV pool reservation (`256 * max_seq_len`) is the same with or without the throttle. The pressure that OOM-kills GEN is not the steady-state KV pool itself; it is the transient transfer-staging memory created by a burst of concurrent KV recv sessions. `max_tokens_in_buffer` sizes the preallocated transfer buffer pool, and that pool is not one buffer per admitted request (`mRecvBufferCount` defaults to 1). When concurrent receives exceed the preallocated coverage, the C++ formatter falls back to dynamic buffers, which can fail or spike memory with NIXL.

So the throttle's effect is on the transient dynamic-allocation peak:
- Without throttle: `_pop_from_waiting_queue()` can admit up to `total_max` in one executor iteration (`tp_size * max_num_active_requests`; for Kimi-2, 32 * 256 = 8192 globally / up to 256 per rank). Those fitting INIT requests then start KV recv sessions immediately, before any transfer has had time to drain.
- With throttle (`tp_size=32`): admission is limited to roughly one request per rank per executor iteration. This does not strictly cap total in-flight recvs at 32, but it prevents the pathological single-iteration burst and gives earlier transfers time to drain before the next batch of INIT requests is admitted.

In short: KV pool = constant, in-flight transfer-staging peak = not constant. The throttle reshapes the temporal distribution of transfer-staging pressure, not the steady-state KV total. I also tightened the inline comment to say "transient transfer-staging peak" instead of "KV/recv-buffer reservations", since the latter was ambiguous about what's actually scaling.

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Done. One outer `if self.dist.rank == 0:` with `if global_ok` / `else` inside. Same logic, no extra rank check.

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Done. Set `all_ranks_ok = [local_ok]` in the non-ADP branch and lifted `global_ok = min(all_ranks_ok) == 1` out of the if/else so it works uniformly for both. The `enumerate(all_ranks_ok or [local_ok])` fallback is gone — `enumerate(all_ranks_ok)` is enough.

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Yup `sum(1 for req in xs if cond)` is equivalent to `sum(1 if cond else 0 for req in xs)`.

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Good call-out: they usually move together today, but they are checking two different ownership boundaries:
- (B) checks the executor-visible request state in `active_requests`.
- (C) asks the transceiver whether its async receive-session bookkeeping is drained.

Today the executor thread owns the request-state transition and also checks completed recv sessions, while the transfer worker threads only advance the underlying transfer task statuses. So in the current common path, B and C generally flip together.

That said, I still think C is worth keeping as a separate barrier invariant. The fill gate is exactly the point where we want both views to agree before allowing the first forward. The check is cheap, and it avoids depending on the current coupling between request state and transceiver cleanup if that lifecycle becomes more asynchronous later.

I added a short comment to make this intent explicit: `Also require the transceiver's async receive bookkeeping to be drained before opening the fill gate.`

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Reduced the size a bit. However, the logic is not trivial. To save the mental load and improve code maintainability, it is best to call out all the conditions explicitly.

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Reduced the comment size and removed PR reference.

tests/integration/defs/perf/disagg/test_configs/wideep/perf/kimi-k2-thinking-fp4_8k1k_ctx8_gen1_dep32_bs256_eplb416_mtp0_ccb-NIXL.yaml
chienchunhung: I reduced the comment size but would like to keep the rational overall. This is test specific config updates and it is critical to record the rational so that future changes don't revert them by accident.

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
chienchunhung: Done. Trimmed the comments and removed the nvbug info.

brb-nv (approved):
LGTM.

pcastonguay (approved):

StanleySun639 (approved):

yingguo-trt (approved):

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -3097,6 +3099,16 @@ def _pop_from_waiting_queue(
...
         max_new_requests = total_max - total_num_active_requests
 
+        # Benchmark disagg fill-phase admission throttle: without it, the
+        # executor admits `total_max` requests in one iteration when the
+        # benchmark queue is preloaded, putting that many KV recvs in
+        # flight at once.  The transient transfer-staging peak can OOM-
+        # kill the GEN server before any transfer drains.  Cap at
+        # `tp_size` (one per rank per iteration) until the fill gate opens.
+        if (self.is_benchmark_disagg and self._benchmark_fill_phase_active
+                and not self.is_warmup):
```
Tabrizian, line 3109: @chienchunhung Are there any dynamic memory allocation in disagg KVCache transfer? Do you actually run into this issue when you ran the benchmark?

chienchunhung: IIUC dynamic alloc is from: cacheFormatter.cpp:868 falls back to `bufferManager.gpu(...)` when the pre-allocated recv pool (default `mRecvBufferCount`=1/rank) is exhausted. No OOM observed because PR-#12206's fast-fail catches the same bulk-admission burst at the scheduler before the dynamic fallback grows. Throttling to `tp_size/iter` prevents both at the source. Updated the comments to be more precise.

PR #13351, opened 2026-04-22
[Issue-13318][fix] Trim context chunks in _prepare_tp_inputs to prevent over-admission crash

### LINKED ISSUE #13318: [Bug]: Scheduler deadlock on main + #12976 + #13029: AssertionError total_num_tokens > max_num_tokens in _prepare_tp_inputs under KV offload + chunked prefill permanently hangs the event loop
### System Info

Host
- CPU architecture: x86_64
- CPU: AMD EPYC 7C13 64-Core Processor (61 vCPUs visible)
- Host memory: 197 GiB
- GPU: 1 × NVIDIA GeForce RTX 5090, 32 GB, compute capability 12.0
- NVIDIA driver: 580.126.09
- Host OS: Ubuntu 22.04.5 LTS (kernel 6.8.0-59-generic)

Container (official TRT-LLM image)
- Image: ssadds/saddss:main-pr12976
- Image ID: sha256:67ef443efe93046cfed60dc918783b7b3722168857616b46aab5432ab82990fd
- Image built: 2026-04-20 10:44:12 UTC
- Image labels:
    com.nvidia.build.ref         = fdfaeb27df34739d5d656e2d94755cd8f42fa80e
    com.nvidia.build.id          = 269443099
    com.nvidia.pytorch.version   = 2.11.0a0+eb65b36
    com.nvidia.tensorrt.version  = 10.15.1.29
    com.nvidia.cublas.version    = 13.2.1.1
    com.nvidia.cudnn.version     = 9.19.0.56
    com.nvidia.nccl.version      = 2.29.stable.20260109
    com.nvidia.nvvm.version      = 13.1.115
    org.opencontainers.image.ref.name / version = ubuntu / 24.04
- This image is built from TensorRT-LLM `main` @ fdfaeb27df34739d5d656e2d94755cd8f42fa80e,
  which already contains PR #12976 and PR #13029.
- Latest public `main` HEAD at time of filing: 0d2bea7c3c99b734a8e09c4c767820e03136a15b
  (2026-04-22 03:12:34 UTC) — bug still reproduces on that image as well.
- Also reproduced on: TensorRT-LLM 1.3.0rc12 (same failure mode, different overshoot values).

Libraries (inside the container)
- Python: 3.12.3
- tensorrt_llm: 1.3.0rc13
- tensorrt: 10.15.1.29
- torch: 2.11.0a0+eb65b36914.nv26.2 (NVIDIA PyTorch build)
- nvidia-modelopt: 0.37.0
- nvcc: CUDA 13.1.115 (release 13.1, V13.1.115)
- libcublas: 13.2.1.1
- libcudnn9-cuda-13: 9.19.0.56

Model under test
- A Mistral-Nemo-12B-architecture chat model (≈12.2 B params, based on
  mistralai/Mistral-Nemo-Instruct-2407), quantized via TensorRT Model Optimizer
  to NVFP4 weights + FP8 KV cache.

Backend selection (from engine_config + defaults)
- backend: pytorch
- V1 C++ MicroBatchScheduler + V1 KVCacheManager
- use_python_scheduler = False
- use_kv_cache_manager_v2 = False

nvidia-smi (host)

### Who can help?

_No response_

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [x] My own task or dataset (give details below)

### Reproduction

1. Start `trtllm-serve` (pytorch backend) with the KV-offload config below:
   `max_num_tokens=4096`, `host_cache_size=180 GiB`, `enable_chunked_prefill=true`,
   `enable_partial_reuse=true`, `enable_block_reuse=true`, `kv_cache dtype=fp8`.

2. Replay a realistic multi-turn chat trace (~3–5 k prompt tokens, ~170 output tokens,
   heavy prefix overlap) at `qps=5.2` for 10–12 rounds until prefix-cache hit-rate plateaus.

3. Bump the same replay to `qps=7.0`. Within 1–3 rounds the assertion fires in the
   event-loop thread, the scheduler stops advancing, and every subsequent
   `POST /v1/chat/completions` hangs indefinitely (`GET /v1/models` still returns 200,
   so naive liveness probes do not detect the outage). Only `docker kill` + restart recovers.

#### engine_config.yaml
max_num_tokens: 4096
max_seq_len: 8000
max_batch_size: 64
backend: pytorch
enable_chunked_prefill: true
cuda_graph_config: { enable_padding: true, max_batch_size: 64 }
kv_cache_config:
  free_gpu_memory_fraction: 0.9
  enable_block_reuse: true
  enable_partial_reuse: true
  host_cache_size: 193273528320   # 180 GiB
  dtype: fp8

#### Server
trtllm-serve serve <MODEL> --backend pytorch --host 0.0.0.0 --port 8080 \
  --config /workspace/engine_config.yaml

#### Client: warm up the reuse cache
python3 online_replay.py --input <replay.jsonl> --replay-mode qps \
  --target-qps 5.2 --sample-range 0.0 0.104 --round-duration 30

#### Client: trigger
python3 online_replay.py --input <replay.jsonl> --replay-mode qps \
  --target-qps 7.0 --sample-range 0.0 0.14 --round-duration 30

File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/_torch/pyexecutor/model_engine.py",
  line 2678, in _prepare_tp_inputs
    assert total_num_tokens <= self.max_num_tokens, (
AssertionError: total_num_tokens (6986) should be less than or equal to max_num_tokens (4096)

[TRT-LLM] [E] Encountered an error in forward function:
  total_num_tokens (6986) should be less than or equal to max_num_tokens (4096)
[TRT-LLM] [E] Error in event loop: Sampling failed
Exception in thread Thread-3 (_event_loop_wrapper):
  ...
  raise AssertionError("Sampling failed")

After this point:
   - POST /v1/chat/completions hangs forever (no 5xx, no response)
   - GET /v1/models still returns 200
   - Iter perf metrics stop advancing in the server log
   - Only `docker kill` + restart recovers

 Overshoot values observed across runs: +1, +4, +22, +2890 tokens.

### Expected behavior

When the offered load exceeds the server's sustainable capacity, the server should
degrade gracefully: queues grow, P50 latency rises, chunked prefill splits larger
prompts so that `total_num_tokens <= max_num_tokens` per iteration, the scheduler
defers or shrinks requests as needed — but the event loop stays alive and the
HTTP endpoints keep responding.
Concretely: a single batch whose token accounting would exceed `max_num_tokens`
by a handful of tokens should not be allowed to crash the background event-loop
thread and brick the whole process.

### actual behavior

With `max_num_tokens=4096`, KV offload enabled, `enable_partial_reuse=true`,
under a realistic replay at qps=7.0, `_prepare_tp_inputs` appends position_ids
whose length exceeds `max_num_tokens` by a small amount (overshoots observed:
+1, +4, +22, and once +2890 tokens). The assertion at `model_engine.py:2678`
fires inside the event-loop thread. That thread then dies; nothing re-spawns
it or drops the offending batch; the scheduler never advances another iteration.
All in-flight and every subsequent `/v1/chat/completions` request hangs with
no response, while `/v1/models` keeps returning 200. The process has to be
force-killed and restarted to recover.
Reproduction matrix (12-round windows, fresh container each run, same trace):
| #  | Config                       | QPS         | Stable P50 e2e  | First failure                                 |
|----|------------------------------|-------------|-----------------|-----------------------------------------------|
| 1  | mnt=4096, partial=on         | 5.2         | 5–6 s           | clean across 12 rounds                        |
| 2  | mnt=4096, partial=on         | 7.0         | —               | ~R2: `total_num_tokens=6986 > 4096` → hang    |
| 3  | mnt=8192, partial=on         | 6.0         | ~5 s            | clean across 12 rounds                        |
| 4  | mnt=8192, partial=on         | 40 (stress) | —               | clean across tens of thousands of requests    |
| 5  | mnt=4096, partial=off        | 5.2         | 69–81 s (15×!)  | R4: `total_num_tokens=4118 > 4096` → hang     |
| 6  | mnt=8192, partial=off        | 6.0         | 81–83 s         | R4: `total_num_tokens=8196 > 8192` → hang     |

Takeaways:
- Raising `max_num_tokens` from 4096 to 8192 hides the bug in practice
  (no failure up to 8× overload, >10 k requests) but is not a root-cause fix.
- `enable_partial_reuse=false` makes things strictly worse: the assertion fires
  at lower QPS and in earlier rounds, *and* P50 regresses 15–20×.
  So "disable partial reuse" is not a viable workaround.
- Reproduced on both 1.3.0rc12 and the main + PR #12976 + PR #13029 build.

### Hypothesis on root cause

`_prepare_tp_inputs` computes `total_num_tokens = len(position_ids)`, appending
`range(begin_compute, begin_compute + context_chunk_size)` per context request
and `1 + num_draft_tokens` per gen request. The C++ `MicroBatchScheduler` (V1)
admits a batch using a budget derived from
`reuse_adjusted_compute(chunk_size, reusable, context_remaining)`, which for
non-last chunks returns `chunk_size` and for last chunks returns
`max(0, context_remaining - reusable)`.

The observed overshoots (+1 … +22 near block boundaries; +2890 in extreme
alignment) are consistent with a small mismatch between this admission-side
budget and the actual `position_ids` length appended in `_prepare_tp_inputs`,
specifically when a last-chunk context request with `reusable > 0` is admitted
alongside near-full generation traffic. That `enable_partial_reuse=false`
increases (rather than decreases) failure frequency is consistent: without
partial skipping the actual `position_ids` extension is longer.

### Related PRs, status anchored to main HEAD 0d2bea7c3c99 (2026-04-22 03:12 UTC)

Already in the tested image, bug still reproduces:
- #12976 — [fix] Fix compute token accounting for KV cache reuse with context chunking
  (merged main 2026-04-18)
- #13029 — [feat] Batch addSequence with two-phase claim  (merged main 2026-04-18)

Merged to main after image build, verified not to fix this:
- #13095 — radix-tree walk consolidation, pure perf refactor, no semantic change
  on the budget path
- #13104 — touches V2 scheduler / V2 KV cache manager only; we run V1
  (`use_python_scheduler=False`, `use_kv_cache_manager_v2=False`)

Fixes that target this exact failure mode but are NOT on main:
- #12806 — [fix] Pre-subtract non-first-chunk context costs in reuse budget check.
  15-line Python fix. Merged only to `feat/bench_y` on 2026-04-07.
  Its umbrella PR #12865 (feat/bench_y → main) was *closed unmerged* on
  2026-04-21 16:44 UTC, so #12806 currently has no path to main.
- #12665 — [fix] Fix KV cache reuse crashes. Closed unmerged 2026-04-01. Same
  failure-mode description.
- #12658 — open draft with Python-side pre-validation, stale since 2026-04-01.

Updating to the latest main does not fix this bug.

### What I'd like from this issue

- Confirmation that this is the same bug #12806 / #12665 describe, or a
  different instance.

### Attachments I can provide

- Full `docker logs` from failing runs (assertion stack + the exact iteration
  where the scheduler stops advancing).
- Per-round client summary JSON (QPS, P50/P90, success rate) for all 6 rows
  in the table above.
- An anonymized replay trace that reliably triggers this in ~5 minutes.

yifjiang:
## Summary

Fixes #13318. Adds a Python-side post-scheduler guard in `_prepare_tp_inputs` that pre-computes the predicted `position_ids` growth and trims the trailing context chunk(s) if the batch would overshoot `max_num_tokens`. Mitigates the event-loop-killing `assert total_num_tokens <= self.max_num_tokens` crash observed in production with `enable_chunked_prefill: true` + `enable_block_reuse: true`.

## Root cause (quoting the issue)

- `MicroBatchScheduler` (V1, C++) admits batches using a reuse-discounted token budget. For a last-chunk context request with `reusable > 0` it charges only `max(0, context_remaining - reusable)` tokens — see `_reuse_adjusted_compute` in `tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py:321` and its C++ counterpart.
- `_prepare_tp_inputs` (this file, around line 2276) extends `position_ids` by the **full** `context_chunk_size` per context request regardless of reuse: the reusable prefix's position embeddings still need to be computed in the forward pass.
- When `max_num_tokens` is tight, the admission-vs-forward mismatch lets `len(position_ids)` overshoot `max_num_tokens` by up to `min(chunk_size, reusable)` per offending request. The assertion at `model_engine.py:2678` fires, the event-loop thread dies, and the server hangs until `docker kill` + restart (issue #13318 has the full reproduction).

The admission math in `_reuse_adjusted_compute` (and its C++ twin) is still wrong after this fix — it subtracts reusable tokens that the forward pass has to process anyway. A proper root-cause fix should align the admission budget with the actual forward-pass cost. Until that lands, this guard keeps the event loop alive and lets the server degrade gracefully instead of crashing.

## What the fix does

At the top of `_prepare_tp_inputs` (before `input_ids = []`), after the `_apply_incremental_update` early-return:

1. Compute `predicted_ctx_tokens = sum(req.context_chunk_size for req in context_requests)` — exact match for what the context loop will extend onto `position_ids`.
2. Compute `predicted_gen_tokens = sum(1 + max(get_draft_token_length(req), self.runtime_draft_len) for req in generation_requests)` — conservative upper bound covering both extend-request and plain-gen / first-draft paths.
3. If `predicted_total > self.max_num_tokens`, iterate the context requests in reverse and shrink `req.context_chunk_size` until the overshoot is cleared (leaving each chunk at least 1 token).
4. If the overshoot cannot be fully cleared (pathological config), log an error and let the original assertion fire so the condition stays observable.

Why mutating `context_chunk_size` mid-pipeline is safe:

- `isLastContextChunk()` (`cpp/include/tensorrt_llm/batch_manager/llmRequest.h:1636`) is **live-derived** from `getContextCurrentPosition() + getContextChunkSize() == mPromptLen`, so shrinking the chunk automatically de-promotes a "last" chunk to a "middle" chunk for this iteration. The next scheduler tick picks up the remaining tokens through the normal chunked-prefill progression.
- `context_chunk_size` is a read/write property in the nanobind binding (`cpp/tensorrt_llm/nanobind/batch_manager/bindings.cpp:128`), mapped to `setContextChunkSize` which clamps to `getContextRemainingLength()` and rejects negatives.
- KV cache already reserved by `prepare_resources` for the original chunk size stays in place and is reclaimed when the request finishes.
- The cost is O(N_requests) per iteration and only fires when overshoot is detected.

## Relationship to PR #12806

PR #12806 fixes the analogous bug on `feat/bench_y`, which carries a Python-side `remaining_budget` re-check in `prepare_resources` that doesn't exist on `main`. The fix here applies the same intent ("pre-account committed forward-pass cost; shrink admissions to fit") but at the layer that exists on `main` — the pre-loop guard in `_prepare_tp_inputs`. Once the C++ admission math is aligned with forward-pass cost (or a Python-side pre-scheduler budget lands on `main`), this guard becomes a redundant backstop and can be removed.

## Test plan

- [ ] Run the reproduction from issue #13318 (Mistral-Nemo-12B replay at qps=7.0 with `max_num_tokens=4096`, `enable_chunked_prefill=true`, `enable_partial_reuse=true`, `enable_block_reuse=true`): verify no `AssertionError`, no event-loop hang, `[issue-13318] trimming context chunk` warnings appear in the server log, and traffic continues to be served.
- [ ] Verify `max_num_tokens=8192` baseline (clean across 10k+ requests per issue's matrix) is unchanged.
- [ ] Unit test: construct a `ScheduledRequests` batch whose summed `context_chunk_size + gen_tokens` exceeds a mocked `max_num_tokens`; assert the trailing `context_chunk_size` is reduced to fit and a warning is logged.
- [ ] No-regression check on non-reuse / non-chunked configs: the guard only activates when `predicted_total > max_num_tokens`, which should not happen under a correct scheduler, so steady-state behavior should be unchanged.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -2237,6 +2237,74 @@ class PyTorchModelEngine(ModelEngine):
                 num_accepted_tokens_device, req_id_to_old_request,
                 resource_manager)
 
+        # Guard against scheduler over-admission (issue #13318).
+        #
+        # The MicroBatchScheduler admits a batch using a reuse-discounted
+        # token budget: for a last-chunk context request with `reusable > 0`
+        # it charges only `max(0, context_remaining - reusable)` tokens
+        # (see `_reuse_adjusted_compute` in scheduler/scheduler.py and the
+        # matching C++ logic). The forward pass below, however, extends
+        # `position_ids` by the full `context_chunk_size` per context request
+        # regardless of reuse (the reusable prefix still needs its position
+        # embeddings computed). When `max_num_tokens` is tight, this
+        # mismatch lets `len(position_ids)` overshoot `self.max_num_tokens`
+        # by up to `min(chunk_size, reusable)` per offending request, which
+        # trips the assertion a few lines below and kills the event loop
+        # thread (see issue #13318).
+        #
+        # Pre-compute the predicted `position_ids` growth using the same
+        # accounting the loops below will use, and shrink the trailing
+        # context chunk(s) to fit. `isLastContextChunk()` /
+        # `isFirstContextChunk()` are live-derived from
+        # `context_current_position`, `context_chunk_size`, and
+        # `prompt_len` — mutating `context_chunk_size` here automatically
+        # de-promotes a "last" chunk back to "middle" for this iteration,
+        # and the next scheduler tick picks up the remaining tokens via
+        # the normal chunked-prefill progression. KV cache allocated by
+        # `prepare_resources` for the original chunk size stays reserved
+        # and is reclaimed when the request eventually finishes.
+        if self.max_num_tokens is not None:
+            predicted_ctx_tokens = sum(
+                req.context_chunk_size
+                for req in scheduled_requests.context_requests)
+            # Each generation request extends `position_ids` by
+            # `1 + num_draft_tokens`. The gen loop below uses either
+            # `get_draft_token_length(req)` (extend path) or
+            # `self.runtime_draft_len` (plain gen / first-draft path), so
+            # take the larger of the two to stay conservative — a slight
+            # overestimate only causes us to trim a little earlier, never
+            # fails to catch a real overshoot.
+            predicted_gen_tokens = sum(
+                1 + max(get_draft_token_length(req), self.runtime_draft_len)
+                for req in scheduled_requests.generation_requests)
+            predicted_total = predicted_ctx_tokens + predicted_gen_tokens
+            if predicted_total > self.max_num_tokens:
+                overshoot = predicted_total - self.max_num_tokens
+                for req in reversed(scheduled_requests.context_requests):
+                    if overshoot <= 0:
+                        break
+                    if req.context_chunk_size <= 1:
+                        continue
+                    trim = min(overshoot, req.context_chunk_size - 1)
+                    new_size = req.context_chunk_size - trim
+                    logger.warning(
+                        "[issue-13318] Scheduler over-admitted by "
+                        "%d token(s); trimming context chunk for request "
+                        "%s: chunk_size %d -> %d",
+                        overshoot, req.py_request_id,
+                        req.context_chunk_size, new_size,
+                    )
+                    req.context_chunk_size = new_size
+                    overshoot -= trim
+                if overshoot > 0:
+                    # Could not trim enough (e.g. every chunk already at 1
+                    # and generation traffic alone exceeds max_num_tokens).
+                    # Let the assertion below fire so the condition stays
+                    # observable rather than silently corrupting state.
+                    logger.error(
+                        "[issue-13318] Residual overshoot of %d token(s) "
+                        "after trimming; assertion will fire.", overshoot)
+
         # if new_tensors_device exist, input_ids will only contain new context tokens
         input_ids = []  # per sequence
         sequence_lengths = []  # per sequence


yifjiang:
Closing in favor of porting the `prepare_resources` `remaining_budget` re-validation from `feat/bench_y` to `main`, then applying #12806's accounting fix on top. That keeps the fix at the same architectural layer the upstream solution was designed for, rather than introducing a parallel post-scheduler guard in `_prepare_tp_inputs`. Production unblock for the immediate qwen-coder deployment will be handled via a downstream-only image patch, not via this PR.

PR #13354, opened 2026-04-22
[#13321][fix] AutoDeploy: stabilize piecewise CUDA graph outputs with multi_stream_moe

### LINKED ISSUE #13321: [Bug]:  Re-enable piecewise cudagraph for Qwen3.5
### System Info

Piecewise cuda graph support breaks when multi-stream MoE is enabled for Qwen3.5 models. Currently we disabled piecewise cuda graph.

Steps to reproduce:
set `piecewise_enabled: true` in `examples/auto_deploy/model_registry/configs/qwen3.5_moe_35b.yaml`

Run:
python examples/auto_deploy/build_and_run_ad.py --model Qwen/Qwen3.5-35B-A3B --use-registry

### Who can help?

_No response_

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

In description

### Expected behavior

In description

### actual behavior

In description

### additional notes

In description

suyoggupta:
## Summary
Enables piecewise CUDA graph together with `multi_stream_moe` for the Nemotron-3-Nano-30B-A3B-FP8 and Qwen3.5-35B-A3B model families (closes #13321). Without this fix, piecewise + multi_stream_moe either produced gibberish tokens (nano_v3, `ADPiecewiseRunner ADDRESS MISMATCH`) or crashed during capture/inference (Qwen3.5, asynchronous IMA in `flashinfer_rope`, then a `size 16 vs 64` `aten.add` shape mismatch).

## Root causes
1. **Unstable output address of reclassified multi-stream partitions.** `piecewise_utils.split_graph_at_dynamic_ops` moves partitions containing `begin/end/wait/record_aux_stream_passthrough` to the dynamic set so they run eagerly (CPU-side `synchronize()` cannot be called during CUDA graph capture). `PiecewiseCapturedGraph.prepare()` left those partitions unwrapped. Their eager output got a fresh caching-allocator address each call, while the following static partition's captured graph had baked in the capture-time address — `ADPiecewiseRunner` logged `ADDRESS MISMATCH` on every layer following one of these partitions, and on Qwen3.5 the pool's address recycling surfaced as an asynchronous IMA in the next layer's RoPE kernel.
2. **Wrong narrow-size in `_copy_to_static_buffers` for the piecewise path.** Static input buffers (e.g. `inputs_embeds`) were narrowed to the real request size. Captured partitions tolerated this silently (padding zeros behind the view), but eager multi-stream submodules hit strict `aten.add` shape checks between narrowed inputs (size 16) and intermediates still at bucket size (size 64).

## Fix
- `MultiStreamWrapper` (`piecewise_runner.py`) wraps each reclassified partition and holds its own per-bucket stable buffer in the regular caching allocator. The submodule runs eagerly and the wrapper copies the result into that buffer before returning, so the next captured static partition always replays against a fixed address. Allocating outside the graph pool sidesteps the aggressive pool-recycling race that caused asynchronous IMAs on larger partitions.
- `_copy_to_static_buffers(kwargs, bucket=None)` (`backends/torch_cudagraph.py`) — when called from the piecewise path with `bucket=num_tokens`, narrows views to the bucket size and zeroes the padding tail. Monolithic path still narrows to actual size.
- `submod_has_stream_switch` (`piecewise_utils.py`) is promoted from private to public so the compile backend can call it.
- `nano_v3.yaml` now sets `compile_model.piecewise_enabled: true` (matches `qwen3.5_moe_35b.yaml`, which already had it on main).

## Test plan
- [x] `build_and_run_ad.py --model nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-FP8 --use-registry` — 0 `ADDRESS MISMATCH`, 10/10 prompts coherent.
- [x] `build_and_run_ad.py --model Qwen/Qwen3.5-35B-A3B` (default config, piecewise on + multi_stream_moe on) — 0 errors, 10/10 prompts coherent.
- [x] `pre-commit run` clean on all modified files.

## Known follow-up
A small number of `CUDA Graph is empty … wrong device or stream` warnings appear on the largest piecewise bucket capture (1 per rank). They do not affect correctness — all outputs are verified coherent — but are worth investigating separately.

## Summary by CodeRabbit

* **New Features**
  * Piecewise model compilation is now available and can be enabled via configuration settings.
  * Enhanced multi-stream execution handling for compiled models.

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/compile/piecewise_runner.py
+++ b/tensorrt_llm/_torch/auto_deploy/compile/piecewise_runner.py
@@ -467,3 +467,104 @@ class DynamicOpWrapper(nn.Module):
             )
             kwargs["out"] = out_buf
         return self.submodule(*args, **kwargs)
+
+
+class MultiStreamWrapper(nn.Module):
+    """Wraps a multi-stream partition to give its output(s) stable addresses.
+
+    Multi-stream partitions (those containing ``begin_aux_stream_passthrough``
+    and friends) are reclassified from static to dynamic because they rely
+    on a host-side ``caller_stream.synchronize()`` that cannot be called
+    during CUDA graph capture.  They therefore run eagerly between captured
+    static segments.
+
+    Their output tensors, however, feed directly into the *next* static
+    segment's captured graph.  The caching allocator hands out a fresh
+    address for each eager allocation, so the next segment replays against
+    a stale pointer and produces wrong outputs (observed as
+    ``ADPiecewiseRunner ADDRESS MISMATCH`` errors).
+
+    Fix: the wrapper holds its own per-bucket stable buffer allocated from
+    the default caching allocator (outside any graph pool).  The buffer is
+    strong-ref'd by the wrapper, so its address never moves.  On each call
+    the wrapper copies the freshly-computed eager output into this buffer
+    and returns the buffer; the following static segment's captured graph
+    therefore replays against a fixed address.
+
+    Why not the preceding runner's graph pool (as ``DynamicOpWrapper``
+    does)?  The graph pool aggressively recycles allocations whose Python
+    strong refs drop after the runner's ``forward`` returns.  With larger
+    partitions (e.g. Qwen3.5's MoE-plus-next-layer-RoPE MS regions), the
+    pool reuses those addresses for later captures, corrupting the MS
+    output and producing asynchronous ``illegal memory access`` errors.
+    Allocating outside the pool avoids the recycling race entirely.
+
+    Cost: one persistent buffer per (wrapper × bucket) in the caching
+    allocator.  Copy cost is one ``cudaMemcpyAsync`` per layer — negligible
+    next to the MoE compute — and runs on the main stream since
+    ``end_aux_stream_passthrough`` has already restored the caller stream.
+    """
+
+    def __init__(self, submodule: nn.Module):
+        super().__init__()
+        self.submodule = submodule
+        # Keyed by num_tokens bucket; value is a Tensor or a list of
+        # (Tensor-or-None) mirroring the submodule's output container.
+        self._stable_bufs: Dict[int, Any] = {}
+
+    def forward(self, *args, **kwargs) -> Any:
+        result = self.submodule(*args, **kwargs)
+
+        phase = ADPiecewiseRunner._current_phase
+        if phase == "warmup":
+            return result
+
+        nt = ADPiecewiseRunner._current_num_tokens
+        if nt is None:
+            return result
+
+        stable = self._stable_bufs.get(nt)
+        if stable is None:
+            stable = self._alloc_stable(result)
+            if stable is None:
+                return result
+            self._stable_bufs[nt] = stable
+
+        return _copy_into_stable(result, stable)
+
+    @staticmethod
+    def _alloc_stable(result: Any) -> Any:
+        if isinstance(result, torch.Tensor):
+            return torch.empty_like(result)
+        if isinstance(result, (tuple, list)):
+            return [torch.empty_like(r) if isinstance(r, torch.Tensor) else None for r in result]
+        return None
+
+
+def _copy_into_stable(result: Any, stable: Any) -> Any:
+    """Copy ``result`` into the stable buffer(s) and return the stable structure.
+
+    Mirrors ``result``'s container type (Tensor/tuple/list).  Non-tensor
+    items in a tuple/list are passed through unchanged.
+    """
+    if isinstance(result, torch.Tensor):
+        if not isinstance(stable, torch.Tensor):
+            return result
+        if result.data_ptr() != stable.data_ptr():
+            stable.copy_(result)
+        return stable
+
+    if isinstance(result, (tuple, list)):
+        if not isinstance(stable, list) or len(stable) != len(result):
+            return result
+        merged = []
+        for r, s in zip(result, stable):
+            if isinstance(r, torch.Tensor) and isinstance(s, torch.Tensor):
+                if r.data_ptr() != s.data_ptr():
+                    s.copy_(r)
+                merged.append(s)
+            else:
+                merged.append(r)
+        return type(result)(merged) if isinstance(result, tuple) else merged
+
+    return result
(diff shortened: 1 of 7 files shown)


suyoggupta:

tensorrt_llm/_torch/auto_deploy/compile/piecewise_runner.py
```diff
@@ -467,3 +467,104 @@ def forward(self, *args, **kwargs) -> Any:
             )
             kwargs["out"] = out_buf
         return self.submodule(*args, **kwargs)
+
+
+class MultiStreamWrapper(nn.Module):
```
bmarimuthu-nv, line 472: Can we add some unit tests to cover this functionality?

tensorrt_llm/_torch/auto_deploy/compile/piecewise_runner.py
```diff
@@ -467,3 +467,104 @@ def forward(self, *args, **kwargs) -> Any:
...
+    """
+
+    def __init__(self, submodule: nn.Module):
+        super().__init__()
+        self.submodule = submodule
+        # Keyed by num_tokens bucket; value is a Tensor or a list of
+        # (Tensor-or-None) mirroring the submodule's output container.
+        self._stable_bufs: Dict[int, Any] = {}
+
+    def forward(self, *args, **kwargs) -> Any:
```
bmarimuthu-nv, line 515: Nit: Idea for future refactor - we can probably create base that class defines the skeleton and subclasses customizing it as needed, e.g:


1. `PiecewiseWrapper` (base) — holds submodule, dispatches on phase/nt, passes through during warmup. All three wrappers inherit this.                     
2. `StableOutputWrapper` (intermediate, extends base) — run-then-copy pattern: execute submodule eagerly, copy result into a persistent per-bucket buffer dict, return stable address. Shared _copy_to_stable() handles Tensor/tuple/list containers. `MetadataWrapper` and `MultiStreamWrapper` inherit this, overriding only _alloc_stable() (Mamba padding vs empty_like) and optionally _copy_to_stable() (shape-mismatch tolerance vs simple copy).
3. `DynamicOpWrapper` (extends base directly) — different mechanism entirely: injects out= kwarg before running, no post-copy. Shares only the phase dispatch from the base.

bmarimuthu-nv (approved):
LGTM w/ unit tests added, thanks!

tensorrt_llm/_torch/auto_deploy/compile/piecewise_runner.py
```diff
@@ -467,3 +467,104 @@ def forward(self, *args, **kwargs) -> Any:
...
+    and friends) are reclassified from static to dynamic because they rely
+    on a host-side ``caller_stream.synchronize()`` that cannot be called
+    during CUDA graph capture.  They therefore run eagerly between captured
+    static segments.
+
+    Their output tensors, however, feed directly into the *next* static
+    segment's captured graph.  The caching allocator hands out a fresh
+    address for each eager allocation, so the next segment replays against
+    a stale pointer and produces wrong outputs (observed as
+    ``ADPiecewiseRunner ADDRESS MISMATCH`` errors).
```
nvchenghaoz, line 485: Two ways to solve this, one is the copy; one is the preallocated output buffer and directly write or copy to that buffer. If we have the output buffer approach, I think the code change might be cleaner from the piecewise perspective.. 

 Also please check if we can combine the metadata copy and this copy if a copy is required.

tensorrt_llm/_torch/auto_deploy/compile/piecewise_runner.py
```diff
@@ -467,3 +467,104 @@ def forward(self, *args, **kwargs) -> Any:
...
+    strong-ref'd by the wrapper, so its address never moves.  On each call
+    the wrapper copies the freshly-computed eager output into this buffer
+    and returns the buffer; the following static segment's captured graph
+    therefore replays against a fixed address.
+
+    Why not the preceding runner's graph pool (as ``DynamicOpWrapper``
+    does)?  The graph pool aggressively recycles allocations whose Python
+    strong refs drop after the runner's ``forward`` returns.  With larger
+    partitions (e.g. Qwen3.5's MoE-plus-next-layer-RoPE MS regions), the
+    pool reuses those addresses for later captures, corrupting the MS
```
nvchenghaoz, line 498: I think the pool only reuses addresses what have no refs. Since model will run the layers sequentially, where is the corrupting from? From two streams or from a later layer?  Are we sure the claim here is correct?

tensorrt_llm/_torch/auto_deploy/compile/piecewise_runner.py
```diff
@@ -467,3 +467,104 @@ def forward(self, *args, **kwargs) -> Any:
...
+    output and producing asynchronous ``illegal memory access`` errors.
+    Allocating outside the pool avoids the recycling race entirely.
+
+    Cost: one persistent buffer per (wrapper × bucket) in the caching
```
nvchenghaoz, line 502: What is the size it added for this persistent buffer? This is kind of conflict to the piecewise design as adding more persistent buffer will cause the in-accurate estimation. 

Ideally we want to make the piecewise cudagraph as the feature that we won't need to update the  free_gpu_memory_fraction when piecewise enabled.

tensorrt_llm/_torch/auto_deploy/compile/piecewise_utils.py
```diff
@@ -262,6 +276,10 @@ def _submod_has_stream_switch(submod: GraphModule) -> bool:
     return False
 
 
+# Backwards-compatible private alias (used within this module below).
+_submod_has_stream_switch = submod_has_stream_switch
```
nvchenghaoz, line 280: let's batch update them to the submod_has_stream_switch. Seems no value for the backward compatibility.

nvchenghaoz (changes requested):

suyoggupta:
closed in favor of #13396

PR #13357, opened 2026-04-23
[https://nvbugs/5839028][fix] Lower KV cache fraction for DeepSeek-R1 FP8 DEEPGEMM variants

xxi-nv:
## Summary
- Lower KV-cache `free_gpu_memory_fraction` from 0.6 to 0.5 for the DEEPGEMM MoE backend in `TestDeepSeekR1::test_fp8_blockscale` on Blackwell (SM100f).
- Unwaive `accuracy/test_llm_api_pytorch.py::TestDeepSeekR1::test_fp8_blockscale[throughput_mtp]` (nvbugs/5839028).

## Motivation

Per the pipeline log attached to [nvbugs/5839028](https://nvbugs/5839028), rank 7 OOMs during GSM8K generation right after MMLU completes:
```
CUDA out of memory. Tried to allocate 3.47 GiB. GPU 7 has a total capacity of 178.35 GiB of which 3.28 GiB is free.
Including non-PyTorch memory, this process has 175.06 GiB memory in use.
139.09 GiB allowed; Of the allocated memory 132.21 GiB is allocated by PyTorch, with 113.88 MiB allocated in private pools (e.g., CUDA Graphs), and 3.57 GiB is reserved by PyTorch but unallocated.
```

The DEEPGEMM MoE backend has a higher peak-memory profile than TRTLLM (on-the-fly shape fusion, per-tactic autotuner scratch). Combined with throughput batch sizes, MTP speculation, and CUDA Graph private pools, `free_gpu_memory_fraction=0.6` leaves no headroom.

Evidence that memory pressure is backend-specific, not param-specific:
- `throughput_mtp_trtllm` (same params, TRTLLM backend) passes at 0.6.
- Sister bug [nvbugs/6084764](https://nvbugs/6084764) waives `throughput` (no MTP, still DEEPGEMM) with the same OOM.

So the fix is keyed on the MoE backend: 0.5 for DEEPGEMM, 0.6 for TRTLLM.

## Verification

On `umbriel-b200-082` (8x B200, release/1.3.0rc11 image), the worktree with this fix ran the unwaived test to completion:
```
accuracy/test_llm_api_pytorch.py::TestDeepSeekR1::test_fp8_blockscale[throughput_mtp]
  MMLU weighted average accuracy: 87.43 (threshold 85.756, reference 87.573) PASS
  GSM8K evaluated accuracy:        95.15 (threshold 92.21,  reference 95.413) PASS
PASSED in 1147.15s (0:19:07)
```
No OOM, no hang.

## Test plan
- [x] Reproduce-then-verify on 8x B200 — passed (see above).
- [ ] Re-run the unwaived case in CI (DGX_B200-8_GPUs-PyTorch-1).

## Summary by CodeRabbit

* **Tests**
  * Re-enabled previously skipped test case for FP8 blockscale functionality.
  * Updated memory configuration parameters to enhance test stability.

### DIFF
--- a/tests/integration/defs/accuracy/test_llm_api_pytorch.py
+++ b/tests/integration/defs/accuracy/test_llm_api_pytorch.py
@@ -2973,7 +2973,15 @@ class TestDeepSeekR1(LlmapiAccuracyTestHarness):
         if is_sm_100f():
             moe_backend = "DEEPGEMM" if moe_backend == "_DEFAULT" else moe_backend
             moe_config = MoeConfig(backend=moe_backend, max_num_tokens=16384)
-            kv_cache_config = KvCacheConfig(free_gpu_memory_fraction=0.6)
+            # The DEEPGEMM MoE backend has a higher peak-memory profile than
+            # the TRTLLM backend on B200 (shapes fused on the fly, per-tactic
+            # autotuner scratch). With throughput batch sizes, MTP speculation
+            # and CUDA Graph private pools, the 0.6 fraction leaves no headroom
+            # and OOMs intermittently during GSM8K generation (nvbugs/5839028
+            # and nvbugs/6084764). Use 0.5 only for DEEPGEMM; keep 0.6 for
+            # TRTLLM which is already exercised successfully at that budget.
+            fraction = 0.5 if moe_backend == "DEEPGEMM" else 0.6
+            kv_cache_config = KvCacheConfig(free_gpu_memory_fraction=fraction)
         else:
             if moe_backend != "_DEFAULT":
                 pytest.skip("Not supported MoE backend!")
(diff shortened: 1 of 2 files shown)


eopXD (commented):
Simply fixing the fraction does not resolve the root causing problem and we will fail into similar cases on other models.

  Concrete next step I'd suggest:

  1. Add one-shot instrumentation inside the measured window and log, for the DEEPGEMM run on B200:
    - self.max_num_tokens, curr_max_num_tokens, num_extra_kv_tokens, original_max_draft_len
    - The per-call x.shape[0] / m_max observed in DeepGemmMoEOp during warmup vs during the GSM8K run that OOMs
    - torch_peak_memory and extra_cost at the end of the window
  2. If the observed warmup m_max is strictly less than inference m_max, force the autotuner warmup to use
  min(self.max_num_tokens, batch_size*(max_seq_len-1)) directly (dropping the get_num_available_tokens clamp) and ensure
  _get_token_num_for_estimation (_util.py:314-369) sizes the probe to fit that request — otherwise _create_warmup_request
  returns None and the warmup silently skips (if batch is None: continue), which would be strictly worse.
  3. Keep the 0.5 fraction for DEEPGEMM for now as a belt-and-suspenders margin; if step 2 lands and the peak captures
  true worst-case workspace, we can revisit whether 0.6 is recoverable.

Please spread with you agent on this. Thank you.

eopXD (changes requested):

StanleySun639 (approved):

PR #13368, opened 2026-04-23
[TRTLLM-12188][feat] Implement SWA prefill memory reuse (scratch slots)

lowsfer:
This commit introduces an opt-in memory saving feature for SWA (Sliding Window Attention) layers during prefill.

During prefill of a new request, out-of-window blocks' KV data is only needed during a single layer's attention computation, and can then be overwritten by the next layer. We leverage this by reinterpreting shared sub-pages within a coalesced slot to serve different blocks for the currently executing layer, rather than different layers for the same block.

Memory Savings:
For a 32 SWA layer model with prompt=1024, window=128, and tokens_per_block=32:
- Current peak: 32 coalesced slots (one per block, each storing all 32 layers).
- With scratch reuse: ceil(27/32) = 1 scratch slot + 5 normal slots = 6 slots.
- Total reduction in peak KV cache memory: ~81%.

Implementation details:
1. `KVCacheManagerConfig` introduces `enable_swa_scratch_reuse`.
2. `_KVCache.resize()` partitions new blocks into scratch and normal blocks.
3. Added `ScratchDesc` and `PageIndexMode` to handle the two-source index conversion logic. `PageIndexMode.PER_LAYER` implies the converted indices include the layer's position within the coalesced slot, while `PageIndexMode.SHARED` indicates that the base pointer holds the per-layer offset.
4. `PageIndexConverter.__call__` now supports processing base indices via scratch mode when configured.

Trade-off:
KV cache prefix reuse is degraded since scratch blocks have no preserved KV data after the step.

Made-with: Cursor

## Summary by CodeRabbit

* **New Features**
  * Added scratch memory reuse support for sliding window attention models, improving KV cache efficiency during prefill operations.
  * Extended public API with new page indexing modes and scratch descriptor utilities.

* **Documentation**
  * Added comprehensive subsystem documentation with build instructions and architectural guidance.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/runtime/kv_cache_manager_v2/CLAUDE.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+### Building Extensions
+
+```bash
+# Build rawref C extension
+cd rawref && python setup.py build_ext --inplace
+
+# Build mypyc-compiled version (run from runtime/ parent dir)
+cd .. && python kv_cache_manager_v2/setup_mypyc.py build_ext --inplace
```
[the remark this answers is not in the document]
lowsfer: Not a real issue. These are **separate instructions** (each with its own `#` comment header), not a sequential script. The second command assumes you start from `kv_cache_manager_v2/`, so `cd ..` correctly lands in `runtime/`, and `python kv_cache_manager_v2/setup_mypyc.py` works as intended.

lowsfer:
Responding to CodeRabbit nitpick comments from the review:

**Nitpick: `_utils.py` line 272-283 — `index_type` docstring**
Not a real issue. `NewType` constructors ARE callables — that's exactly how Python's `NewType` works. The annotation `Callable[[Any], Index]` matches the docstring description ("A type alias for the NewType index, e.g. `type(BlockOrdinal(0))`"). No change needed.

lowsfer:
**Nitpick: CLAUDE.md hardcoded `~/tekit` paths (lines 15-32)**
Good catch — fixed. Replaced hardcoded `~/tekit` paths with `REPO_ROOT="$(git rev-parse --show-toplevel)"` in all three command blocks (Fast mode, Single test, Production mode).

lowsfer:
**Nitpick: Unused `scratch_ranges` in `resume()` (_kv_cache.py line 754)**
Valid — renamed to `_`.

lowsfer:
**Nitpick: Unused `scratch_beg/end` in `resize()` (_kv_cache.py line 507)**
Valid — removed the dead unpack line entirely.

lowsfer:
**Nitpick: `zip(..., strict=True)` (_kv_cache.py line 779)**
Added `strict=True`.

lowsfer:
**Nitpick: Add perf test CI entries (test file lines 1867-2158)**
Not applicable. These are unit tests with `FakeEngine` (mock GPU operations), not end-to-end inference benchmarks. The perf CI configs (`l0_perf.yml`, `llm_perf_*.yml`) are for model-level perf tracking — adding a unit test class there is meaningless.

tests/unittest/kv_cache_manager_v2_tests/test_kv_cache_manager_v2.py
```diff
@@ -1864,5 +1864,299 @@ def test_typical_covers_constraint_ratio_unchanged(self):
...
+            # Commit all tokens — scratch slots persist for reuse by future resize
+            kv.commit(prompt)
+            kv.stop_committing()
+            self.assertTrue(kv.has_scratch_slots)
+
+        s.take_finish_event().synchronize()
+        kv.close()
+        # Scratch freed on close
+        self.assertFalse(kv.has_scratch_slots)
+        self.manager.clear_reusable_blocks()
```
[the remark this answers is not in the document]
lowsfer: Already covered. `scratch_desc.slot_ids` contains actual allocated slot IDs from storage (not just metadata), and `base_page_indices` reflects real allocation state. The test validates:
- `len(scratch_desc.slot_ids) == expected_scratch_slots` (line 1960) — actual slot count
- `expected_total < num_blocks` (line 1957) — proves memory reduction
- Per-block `BAD_PAGE_INDEX` vs valid index (lines 1964-1972) — verifies real allocation per block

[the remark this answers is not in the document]
lowsfer: Already addressed in an upcoming push. Added prefix-reuse regression tests:
1. **Exact match reuse** — `kv2` with identical prompt, verifies reuse works
2. **Prefix match reuse** — `kv3` with `prompt[:896]`, verifies scratch blocks are not reused (comment: "Since kv3 doesn't reuse out-of-window scratch blocks, input is prompt3")

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache_manager.py
```diff
@@ -101,20 +103,81 @@ class AggregatedPageDesc:
...
+    range: HalfOpenRange[BlockOrdinal]  # block ordinal range [beg, end)
+    slot_ids: Sequence[int]  # scratch slot IDs, length = ceil(num_scratch_blocks / scale)
+
+
 @dataclass(slots=True, frozen=True)
 class PageIndexConverter:
     scale: int
     expansion: int
+    layer_offset: int | None  # sub-page offset within coalesced slot
+    scratch_pages_per_block: int = 1
```
jiaganc, line 127: `scratch_pages_per_block` seems to be a per layer group data, so maybe better to put in in `ScratchDesc`.

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
```diff
@@ -387,6 +401,34 @@ def get_aggregated_page_indices(
...
+        )
+
+    @property
+    def has_scratch_slots(self) -> bool:
+        """True if this KV cache currently has scratch slots allocated."""
+        return any(len(s) > 0 for s in self._scratch_slots)
+
+    @property
+    def page_index_mode(self) -> PageIndexMode:
+        return PageIndexMode.PER_LAYER if self.has_scratch_slots else PageIndexMode.SHARED
```
jiaganc: When a batch has both context and generation requests, we expect requests have the same kv cache base address. But if context requests use `PRE_LAYER` mode and generation requests use `SHARED` mode, this expectation is broken. So I think use `PER_LAYER` only when scratch reuse is enabled.

lowsfer: OK, I think we can change it to `def supports_index_mode(self, index_mode) -> bool`. PER_LAYER is actually always supported. SHARED is allowed only when there is no scratch blocks. Does that sound good?

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache_manager.py
lowsfer: But ScratchDesc is for data that changes on every step of every request, while what is in this converter is always constant during lifetime of the kv cache manager.

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
jiaganc: After offline discussion, requests without scratch can use both `PER_LAYER` and `SHARED` mode, so this is not a blocker. Keep it as-if.

jiaganc (approved):
Approved with comments.

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache_manager.py
jiaganc: Learned that. It's better to have comments to note what data is constant acoss steps and what is not.

tensorrt_llm/runtime/kv_cache_manager_v2/_page.py
```diff
@@ -472,3 +475,72 @@ def batched_lock_to_gpu(
         page.lock(kv_cache, beam_index, ordinal, life_cycle, skip_wait=True)
         for page, beam_index, ordinal, life_cycle in tasks
     ]
+
+
+@dataclass(slots=True)
+class TempSlotLock:
```
eopXD: Is this dead code?

eopXD (commented):
Should we add some logging capabilities so we can observe the scratch slot actions?

tensorrt_llm/runtime/kv_cache_manager_v2/_page.py
lowsfer: Yes, I intend to use this instead of raw Slot for most usage, to simplify some code and make it safer. I'll remove it for now and add back later.

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache_manager.py
lowsfer: ScratchDesc has attribute `frozen=True` about the intent.

jiaganc: `frozen=True` means the object cannot be changed after creation, but it doesn't indicate that we need to get a new `ScratchDesc` in each step. Similarly, I cannot tell if I need to get a new `PageIndexConverter` in each step from the doc string of it or `get_page_index_converter` method.

tensorrt_llm/runtime/kv_cache_manager_v2/_page.py
eopXD: Got it. Thank you.

yizhang-nv (approved):

PR #13385, opened 2026-04-23
[None][feat] Add duration-based execution to benchmark

### LINKED ISSUE #13487: [None][feat] Add duration-based execution to trtllm-bench
## Feature Description
Add a duration-based execution feature to the TensorRT-LLM Python benchmark suite (`trtllm-bench`).
Currently, benchmarks rely on a fixed `num_requests`, which can lead to excessively long run times for scenarios with high Input Sequence Length (ISL) and Output Sequence Length (OSL).

We need a way to limit the benchmark by time (e.g., run for 10 minutes) regardless of the initial request count.

## Proposed Solution
Implement a wall-clock check in the worker loop to stop pulling new requests after the duration has elapsed, while allowing in-flight requests to drain to ensure valid statistics.

This should affect both `throughput` and `latency` commands.

weikuo0506:
## Description
This PR adds a duration-based execution feature to the TensorRT-LLM Python benchmark suite (`trtllm-bench`). Currently, benchmarks rely on a fixed `num_requests`, which can lead to excessively long run times for scenarios with high Input Sequence Length (ISL) and Output Sequence Length (OSL). 

The new `--duration` option (in seconds) allows users to limit the benchmark run time. The implementation uses a wall-clock check in the worker loop to stop pulling new requests after the duration has elapsed, while allowing in-flight requests to drain to ensure valid statistics.

This affects both `throughput` and `latency` commands as they share the same execution logic.

Fixes #13487

## Potential Impacts
- **Performance**: This change only affects the benchmarking tool and has no impact on inference performance.
- **Functional**: It adds a new optional CLI parameter and changes the stop condition of the benchmark loop when used.

## Test Coverage
- Added a unit test `test_bench_async.py` to verify `LlmManager` duration logic using mocks.
- Local verification was skipped due to missing dependencies (`torch`, `pytest`) in the host environment, but the test is designed to be run in the standard container.

## PR Checklist
[✓] PR description clearly explains what and why.
[✓] PR Follows TRT-LLM CODING GUIDELINES.
[✓] Test cases are provided for new code paths.


dc3671:
@FrankD412 Can you help review this?

FrankD412:
> @FrankD412 Can you help review this?

Can do -- will handle it asap 🙂

tensorrt_llm/bench/benchmark/utils/asynchronous.py
```diff
@@ -197,9 +219,20 @@ def _task_done_callback(self, task: asyncio.Task):
...
 
     async def worker(self) -> None:
+        """Worker task that pulls requests from inbox and processes them."""
         try:
             while not self._stop.is_set():
                 self._raise_for_failed_tasks()
+
+                if self.duration is not None and self.start_time and time.perf_counter() - self.start_time >= self.duration:
+                    logger.info("Duration reached. Stopping pulling requests.")
+                    while not self._inbox.empty():
```
FrankD412, line 265: Question here -- if the duration is over, do we want to accept new requests? Is this simply to make sure that we don't violate concurrency/keep the engine saturated? Couldn't we just set the stop event and break the loop?

weikuo0506: Hi @FrankD412, thanks for the review! The reason we don't set `self._stop.set()` and just break here is to support the **drain period** for in-flight requests:

1. If we set `self._stop.set()`, the `finally` block will detect it and proactively cancel all remaining active tasks (`for task in self._tasks: task.cancel()`). This would interrupt requests that are currently being processed and cause us to lose their performance statistics.
2. The reason we clear `_inbox` is that `async_benchmark` relies on `backend.busy` returning `False` to determine when it can wrap up. Since the dataset submission process eagerly enqueues all requests into the `_inbox` at the start, the inbox is full. If we break without clearing the inbox, `backend.busy` will stay `True`, causing the benchmark to hang indefinitely. Clearing the inbox allows us to skip all upcoming requests while letting the in-flight ones run to completion and be recorded.

FrankD412: Ok — yeah, looking over this it makes sense. For some reason it wasn’t clicking on my last review but that makes sense!

FrankD412 (approved):

weikuo0506:
Hi @dc3671 , can you help to review?  Thanks

karljang:
@weikuo0506,
It appears there are pre-commit failures. Could you please address them before we proceed?

weikuo0506:
Hi @karljang, thanks for checking this! The branch was quite a bit behind upstream `main`, which caused the legacy baseline pre-commit checks to fail in CI. I have synchronized the branch with the latest `main` to resolve the baseline configuration drift.

/bot run

karljang:
@weikuo0506 , could you please address the pre-commit check failures~ thank you!

weikuo0506:
Hi @karljang, I have addressed the pre-commit formatting and linting failures and pushed the updates. Thank you for the reminder!

karljang:
@weikuo0506,
Thanks for paying attention and keeping it up-to-date. However, it’s also a good practice to wait until we have a valid reason to update the branch. This is because the update will invalidate previous CI results, considering a few flaky tests. :) 
By the way, I asked the team to review this PR, please allow us some time for review~.
Thank you!

dc3671 (approved):

tensorrt_llm/bench/benchmark/utils/asynchronous.py
YihuiLu512: In this `while` loop body, all calls are synchronous, so it will fully drain all requests from the inbox and dispatch them until the queue is empty. Only then does the first scheduling point occur, at which time the first request gets scheduled and `self.start_time` is set in the `_process_single_request` function.

If the above logic is correct, **then regardless of how `duration` is configured, all requests will still be dispatched in full** (i.e., the control logic runs before the actual behavior, resulting in a priority inversion).

Given that dispatched requests are not actively killed at present, those requests will eventually complete as in-flight requests. **Therefore, `duration` cannot effectively control the benchmark run time.**

karljang:
@YihuiLu512 , thank you for your comment! 

@weikuo0506 , could you please take a look at the comment: https://github.com/NVIDIA/TensorRT-LLM/pull/13385#discussion_r3612009547
Thank you!

tensorrt_llm/bench/benchmark/utils/asynchronous.py
weikuo0506: Hi @YihuiLu512, thanks for catching this — you're right, and the analysis is exactly on point.

Since the dispatch path in the worker loop has no await point (`get_nowait()` and `create_task()` are both synchronous), the entire inbox is dispatched into tasks in one synchronous burst before the first request coroutine ever runs. `start_time` is therefore still `None` throughout dispatch, so the duration check could never fire, and the `finally` block then waited for every dispatched task to complete — `--duration` had no effect on runtime.

I've pushed a fix that enforces the deadline at execution time instead:

1. Each request coroutine now checks the deadline **after acquiring its concurrency slot** and returns without submitting (and without emitting a perf item) if the duration has elapsed. In-flight requests still drain to completion, preserving the statistics semantics discussed earlier in this thread with @FrankD412. The multi-turn path also stops issuing further turns once the deadline passes.
2. The worker-side check you quoted is kept, with a comment documenting precisely why it cannot fire during dispatch and what it does do: it lets the worker exit its idle loop cleanly after the deadline, and the inbox drain defensively guarantees `busy` can reach `False` (preventing a benchmark hang if request submission ever becomes paced).
3. Known limitation: with unbounded concurrency (`--concurrency` unset), all requests are submitted to the engine immediately and no client-side checkpoint remains — enforcing duration there would require aborting engine-side requests, which also raises a stats-censoring question (requests still running at the deadline skew long). I've added a warning log for this combination and propose handling engine-side abort in a follow-up issue to keep this PR's scope contained.
4. Tests: the previous unit test passed only due to a timing artifact (it sampled `outbox.qsize()` while the third request was still in flight). It now awaits full worker drain, and I've added a regression test for exactly the eager-dispatch scenario you described (20 requests, concurrency=2, duration=1s → ~10 processed, wall time bounded by duration plus one drain).

Please take another look when you get a chance. Thanks again for the careful review!

karljang: @weikuo0506 , thank you for the quick turnaround!
A quick comment, while waiting for @YihuiLu512 's reply.
Given the known limitation you mentioned, the `test_async_benchmark_duration` also doesn't provide `--concurrency` setting, and expecting only 2 of 3 requests to be processed.  I'm not sure if the test will pass.

weikuo0506: Hi @karljang, good catch — thank you!

You're right that `test_async_benchmark_duration` runs without a concurrency setting. Digging into it: without a concurrency limit, all three mock requests (0.6s each) start immediately and finish at ~t=0.6s, before the 1s deadline — so the `== 2` assertion couldn't hold. This was an oversight in the test when I first wrote it (its sibling tests `test_llm_manager_duration` and `test_llm_manager_duration_not_exceeded` both correctly pass `concurrency=1`, and its own comment — "only 2 should complete within 1s limit" — assumes serialized execution). The flaw predates yesterday's enforcement fix: with unbounded concurrency, every request starts before the deadline under both the old and new code.

I've pushed a commit that passes `concurrency=1` to `async_benchmark` in that test, with a comment explaining why the limit is required for duration to bound the run (matching the warning now emitted by `LlmManager` for the unbounded case).

karljang: @YihuiLu512 , could you please take a look at the updated code? Thank you!

tests/unittest/llmapi/test_bench_async.py
```diff
@@ -0,0 +1,267 @@
...
+
+    manager = LlmManager(
+        llm=mock_llm,
+        outbox=outbox,
+        streaming=False,
+        concurrency=concurrency,
+        duration=duration,
+    )
+
+    req = InferenceRequest(input_ids=[1, 2, 3], output_tokens=10)
```
YihuiLu512: InferenceRequest’s `task_id: int` is still required and has no default value at head `1dd8b9` (`tensorrt_llm/bench/dataclasses/general.py:22`). Can the current constructor still run successfully?

YihuiLu512: same problem for: L62/L134/L192/L241

tests/unittest/llmapi/test_bench_async.py
```diff
@@ -0,0 +1,267 @@
...
+            concurrency=1,
+            duration=1,  # 1 second limit
+        )
+
+    # With concurrency=1, requests run back-to-back (0.6s each): requests 1
+    # and 2 complete at 0.6s and 1.2s; request 3 acquires its slot past the
+    # 1s deadline and is skipped. Without a concurrency limit all three would
+    # start immediately and finish within 0.6s, before the deadline — duration
+    # cannot bound an unbounded-concurrency run (see LlmManager warning).
+    assert len(stats.requests) == 2
```
YihuiLu512: This assertion seems impossible to satisfy. `stats` is a dictionary deduplicated by `request_id`, and here `request_id` is always `1` (L231). Therefore, both items have `request_id == 1`, so they both hit key `1`, leaving only one key in the dictionary. As a result, `len(stats.requests) == 1`, while the assertion expects `== 2`, so the test will inevitably fail.

YihuiLu512 (changes requested):
No further issues are apparent in the functional logic. 

However, the test is unsound and is not included in the CI test list. In light of the ongoing code normalization effort (/cc @QiJune ), please evaluate whether to remove the invalid test or to correct it.

tests/unittest/llmapi/test_bench_async.py
YihuiLu512: Does this test pass on your local machine?

tests/unittest/llmapi/test_bench_async.py
weikuo0506: Confirmed and fixed — `InferenceRequest.task_id` is required with no default, so all four constructions (L62/L134/L192/L241) raised `ValidationError` and no test in the file could run. Each now passes a `task_id`; the three requests in `test_async_benchmark_duration` get distinct ids since they model distinct dataset entries.

tests/unittest/llmapi/test_bench_async.py
weikuo0506: Confirmed on both counts.

**The assertion**: `StatsKeeper.requests` is keyed by `request_id`, which comes from `response.id`, and the mock returned a single shared object with `id = 1`. Both completed requests therefore collapsed into one `RequestRecord` — which not only made `len(stats.requests) == 2` unsatisfiable but merged their events, skewing the recorded statistics. `generate_async` now returns a fresh output with a distinct id per call. (Worth noting `num_complete` increments without dedup, so it would have passed either way — `len(stats.requests)` was the only assertion that could catch this.)

**"Does this test pass on your local machine?"** — it did not, and I had not run it. I've since set up an environment that can (`tensorrt-llm==1.3.0rc21` matching the branch, CPU-only — the tests mock `LLM` entirely and need no GPU) and confirmed the before/after:

```
before:  4 failed in 17.88s
after:   4 passed in 22.83s   (stable across 3 consecutive runs)
```

Running it also surfaced a third defect neither of us had caught: `from tensorrt_llm._tensorrt_engine import LLM` — that module was removed in #15918 and reached this branch through a main merge, so the file failed at collection. It now imports `LLM` from `tensorrt_llm`, matching the other tests in this directory.

**Root cause**: `tests/unittest/llmapi/test_bench_async.py` is not referenced by any list under `tests/integration/test_lists/`, so CI has never collected it — which is how all three defects survived. I've registered it in `l0_cpu_x86.yml` and `l0_cpu_arm.yml` (both `system_gpu_count: 0`). I also replaced a fixed 1.5s sleep with awaiting the perf items, and widened the margin on the eager-dispatch wall-clock bound, since these now run in pre-merge. Happy to move it to a different stage if you'd prefer.

weikuo0506:
@karljang @YihuiLu512 — pushed fixes for both of YihuiLu512's findings, plus a third defect that running the tests surfaced (`_tensorrt_engine` was removed in #15918 and reached this branch via a main merge, breaking collection). Verified locally against `tensorrt-llm==1.3.0rc21`: 4 failed → 4 passed, stable across 3 runs.

One change worth flagging: this test file was not in any test list, so CI has never collected it — that's how the defects survived. I registered it in `l0_cpu_x86.yml` / `l0_cpu_arm.yml` (both `system_gpu_count: 0`, tests are fully mocked). Please let me know if a different stage suits better. Ready for CI whenever convenient.

tensorrt_llm/bench/benchmark/utils/asynchronous.py
```diff
@@ -216,9 +274,17 @@ async def worker(self) -> None:
...
-            for task in self._tasks:
-                task.cancel()
+            logger.debug("Worker task finishing...")
+            if self._stop.is_set():
+                logger.debug(
+                    "Worker task cancelling remaining requests due to stop signal..."
+                )
+                for task in self._tasks:
+                    task.cancel()
+            else:
```
YihuiLu512: When `_raise_for_failed_tasks` raises a `ValueError`, `self._stop` is not set (`stop` is only set by `backend.stop()` during normal shutdown). This causes execution to enter the `else` path, skip cancellation, and wait for all tasks to finish. As a result, **runtime on failure may be longer than the pre-change baseline**.

tensorrt_llm/bench/benchmark/throughput.py
```diff
@@ -194,6 +194,13 @@
     "Number of requests to cap benchmark run at. If not specified or set to 0, it will be the "
     "length of dataset.",
 )
+@optgroup.option(
+    "--duration",
+    type=int,
```
YihuiLu512: `0` and negative values are accepted. With `duration=0`, `_duration_exceeded()` becomes true immediately after `start_time` is first set for the initial request, effectively skipping almost all requests. Please evaluate whether this aligns with the intended design.

YihuiLu512 (approved):
LGTM

Please address the remaining corner cases as appropriate.

tensorrt_llm/bench/benchmark/utils/asynchronous.py
weikuo0506: Good catch — this was a real regression. `self._stop` is only set by an explicit shutdown, so the failure path took the drain branch and the worker waited for every outstanding request before surfacing the error. Draining is only correct on duration expiry, where the point is to record the in-flight requests' statistics; on failure the run is aborting anyway.

Fixed by tracking the duration-triggered exit explicitly and cancelling in every other case, so a stop signal and a failure both behave as they did before duration support was added. Also added a regression test for the failure path, which had no coverage — verified it fails against the previous code (worker blocks on the hung requests until the test times out) and passes with the fix.

tensorrt_llm/bench/benchmark/throughput.py
weikuo0506: Agreed — `--duration 0` silently produced a run that skips nearly everything while looking like it worked, which is worse than an error.

I considered treating `0` as "no limit" for consistency with `--num-requests`, but "run for zero seconds" has no useful meaning the way "cap at the dataset length" does, so a silent reinterpretation seemed more surprising than a rejection. The option is now `click.IntRange(min=1)` on both `throughput` and `latency`, which also rejects negatives and reports the mistake at the command line. Leaving it unset still means no limit.

weikuo0506:
Thanks @YihuiLu512 — both corner cases are pushed: the failure path now cancels in-flight requests (with a regression test), and `--duration` is `click.IntRange(min=1)` on both commands.

On your earlier point about the test not being in the CI list, I corrected it rather than removing it and registered it in `l0_cpu_x86.yml` / `l0_cpu_arm.yml`. Happy to move it if the normalization effort you mentioned prefers otherwise.

tensorrt_llm/bench/benchmark/utils/asynchronous.py
```diff
@@ -216,9 +278,15 @@ async def worker(self) -> None:
...
+                logger.debug(
+                    "Duration reached. Waiting for in-flight requests to complete..."
+                )
+            else:
+                logger.debug("Worker task cancelling remaining requests...")
+                for task in self._tasks:
+                    task.cancel()
             logger.debug("Waiting for requests...")
             if self._tasks:
                 await asyncio.wait(self._tasks)
```
karljang: @weikuo0506 , thank you for addressing the review comments! 
I believe this comment might be the last one :) 

There may still be a failure-after-duration corner case. If the duration expires first, `drain_in_flight` becomes True and the finally block waits for every task. If an in-flight request then fails while another remains slow or hung, the failure is not checked until all tasks finish.

Could the drain path use `asyncio.wait(..., return_when=asyncio.FIRST_EXCEPTION)` and cancel the pending tasks when any completed task has an exception? If all tasks succeed, it would retain the intended duration-drain behavior. A regression test where duration expires before one request fails and another hangs would cover this ordering.

Thank you.

xinhe-nv (approved):

tensorrt_llm/bench/benchmark/utils/asynchronous.py
weikuo0506: Good catch, and thanks for the concrete suggestion — implemented as you described. The drain path now waits with `FIRST_EXCEPTION` and cancels the remainder once a drained request raises; when every request succeeds the call still waits for all of them, so the duration-drain behaviour is unchanged.

Added the regression test with the ordering you asked for: duration expires first, then one in-flight request fails while another hangs for 30s. Verified it fails against the previous code (the drain blocks until the test times out) and passes with the fix. Full file: 6 passed.

karljang:
Marked all conversations as resolved, since all approved the PR.
I will trigger CI~

karljang (approved):

karljang:
~~The latest `/bot run` comment has not received a bot acknowledgement. Could an authorized maintainer please trigger CI for the current head (`6d74e9bc`)?~~

BowenFu:
The default path is clean and I checked it carefully, since the `finally` restructure runs on *every* benchmark, not just `--duration` ones: with `duration=None`, `_duration_exceeded()` is always False, the worker early-exit never fires, `drain_in_flight` stays False so every task is still cancelled and awaited, and `assert finished_requests == len(requests)` still runs. The `pending = set(self._tasks)` snapshot is safe — only the worker adds to `_tasks` and the done callback only discards — and it incidentally fixes the old `await asyncio.wait(self._tasks)` racing the callback that mutates that set. Good change.

Two things worth fixing before it lands:

**1. `--duration` is a no-op on the most common `throughput` invocation.** `throughput` defaults concurrency to `-1`, i.e. unlimited (`throughput.py:254-258`), and the enforcement point is *after* the concurrency slot is acquired. So `trtllm-bench throughput ... --duration 60` dispatches the whole dataset immediately, every request is already past the semaphore, and the run goes to completion — the user gets a warning and the full dataset. That contradicts the flag's own help text ("Maximum run time in seconds. Benchmark stops at whichever limit is hit first"). `latency` is fine because it defaults to concurrency 1 (`low_latency.py:179-183`).

A warning isn't enough for a flag whose entire purpose is bounding runtime on long ISL/OSL runs — that's precisely the scenario in your description. I'd make it a validation error when `duration` is set and concurrency is unlimited, or have `--duration` imply an enforceable concurrency.

**2. The exact-count assertions will flake in pre-merge CI.** Both list entries are `stage: pre_merge` (`l0_cpu_x86.yml:12`, `l0_cpu_arm.yml:12`), so these run on every PR, on x86 *and* ARM. `test_llm_manager_duration` asserts `outbox.qsize() == 2` from 0.6s sleeps against a 1s deadline: request 2 acquires its slot at t≈0.6 and only just makes the deadline. A ~0.4s scheduling stall on a loaded runner pushes it past 1.0s, one request completes, and the test fails. `test_async_benchmark_duration` (`assert len(stats.requests) == 2`) has the same shape.

The elapsed-time bounds elsewhere are generously slack and fine; it's the exact equalities on either side of the deadline that are the problem. Relaxing them to `>= 1 and < 3` would still catch "duration had no effect" without pinning the boundary.

Minor, non-blocking: the timer starts when the first request coroutine runs, not when the benchmark starts, so enqueue time is excluded from "maximum run time"; a multi-turn request truncated mid-conversation is emitted as an ordinary completed perf item, which mixes partial and complete conversations in duration-mode stats; and the worker logs "finishing" twice on normal exit.

**Heads-up on a collision, not a review point:** #16498 deletes both `l0_cpu_x86.yml` and `l0_cpu_arm.yml` and replaces them with a unified `l0_cpu.yml` that lists only the three existing tests. Whichever of the two lands second will hit a modify/delete conflict on those files, and the easy resolution silently drops `unittest/llmapi/test_bench_async.py` from CI entirely. Worth coordinating with @tongyuantongyu so the entry is carried into `l0_cpu.yml`.

fredricz-20070104 (approved):
## Review summary - Approve

Reviewed the full diff; no blocking or major issues found.

Minor, non-blocking notes:
- `tensorrt_llm/bench/benchmark/utils/asynchronous.py`: --duration silently ineffective when concurrency is unbounded (common default)
- `tests/unittest/llmapi/test_bench_async.py`: Exact-count assertion is timing-sensitive

_Automated review by NVCortex Lite, run by @fredricz-20070104._

weikuo0506:
@BowenFu Both points fixed; this also covers the same two flagged by the automated review against `eea40601`.

**Unbounded concurrency**: now a validation error on both commands, raised before the model loads.

I rejected the combination rather than having `--duration` imply a concurrency limit: any value we picked would silently change what a throughput benchmark measures, which seemed worse than refusing outright. The deadline is checked when a request acquires its concurrency slot, so with no limit there is no such checkpoint — every request is already inside the engine by the time it could apply. `--duration` with unbounded concurrency is therefore not supported, and now says so instead of running the full dataset behind a warning.

**Boundary assertions**: both now `1 <= n < 3`. A third request cannot complete either way, so the upper bound still catches a limit that has no effect.

**Minor points**: truncated multi-turn requests are no longer emitted as completed perf items, with a regression test; duplicate worker log removed. Left the timer start as is — measuring from the first request excludes model loading, which seems the more useful definition of "maximum run time".

**#16498**: @tongyuantongyu, this PR adds `unittest/llmapi/test_bench_async.py` to `l0_cpu_x86.yml` and `l0_cpu_arm.yml` — could you carry the entry into `l0_cpu.yml`? Happy to rebase onto yours if it lands first.

Verified on CPU against 1.3.0rc21: 7 passed, each new test confirmed to fail without its fix.

weikuo0506:
@karljang Review fixes pushed, including the boundary assertions that are my main suspect for the CI failures.

The pipeline and analysis links are internal, so I can't see which tests failed. If the next run is green we're done; if it fails again, could you share the failing cases?

BowenFu (approved):
Re-verified on `bcc4ab038`. With `--duration` unset every new branch is inert (`_duration_exceeded()` false, `drain_in_flight` false, the multi-turn `truncated` early-return unreachable), and the rewritten `worker()` `finally` is equivalent to main's cancel-then-wait — the `pending = set(self._tasks)` snapshot additionally fixes main's race where `_task_done_callback` mutates that set while `asyncio.wait` iterates it.

Both points I raised earlier are addressed, and one better than I asked: `--duration` without a concurrency limit is now a hard `click.UsageError` raised before model load on *both* commands rather than a warning followed by a full-dataset run, so the flag can no longer be a silent no-op on `trtllm-bench throughput`.

Two things I'm explicitly not blocking on: the duration drain's final `await asyncio.wait(pending)` has no timeout, so in duration mode a single non-terminating request can hang the run; and requests dropped at the deadline are excluded from `StatsKeeper.requests`, so a truncated run reports throughput/latency over completed requests only, not the submitted dataset — worth a line in the `--duration` help text.

karljang:
@weikuo0506, I regret to inform you that, as @BowenFu mentioned earlier, we’ve encountered conflicts. Could you kindly resolve them?

weikuo0506:
@karljang Conflict resolved and pushed (`69ac657`); the branch is level with main again. It was another append to the same line in `l0_cpu_x86.yml`, so both entries are kept.

@BowenFu On your two non-blocking notes: I documented the reporting scope in the `--duration` help text, since a truncated run reporting over completed requests only is easy to misread. I left the drain without a timeout — a bound there needs its own decision on what value is right for long ISL/OSL requests and what to do when it expires, and picking one arbitrarily would risk cancelling healthy requests and losing exactly the statistics the drain exists to preserve. Note the same exposure predates `--duration`: a non-terminating request already keeps `backend.busy` true and spins `async_benchmark`'s loop on main.

CI has now failed three times (`6d74e9b`, `eea4060`, `bcc4ab0`) and the pipeline links are internal, so I still can't see which tests failed or whether they relate to this PR. Could you share the failing cases? Verified locally against 1.3.0rc21 on CPU: 7 passed.

karljang:
The last failure looks not relevant to this PR. I will retrigger the CI, thank you!

karljang:
@weikuo0506, it’s finally merged! Thank you for your contribution and efforts 👍

PR #13404, opened 2026-04-24
[TRTLLM-12200][feat] WideEP FT: add active_rank_mask to NVLink AlltoAll kernels (1a.2)

chienchunhung:
## Release Notes

* **New Features**
  * Added active rank mask capability to optimize MoE all-to-all communication by efficiently skipping inactive ranks during dispatch and combine operations.
  * Extended maximum supported Expert Parallel (EP) size from 64 to 128 ranks.
  * New optional `active_rank_mask` parameter for MoE dispatch and combine operations, with default behavior unchanged when not specified.

## Description

Adds an `active_rank_mask` parameter (uint64[2] bitmask of currently-alive EP ranks) to the NVLink one-sided MoE AlltoAll dispatch and combine kernels. When the mask is omitted, behavior is bit-identical to before; when a bit is cleared, the kernel skips every interaction with that peer rank — most importantly, the completion-flag spin-wait loops that today hang forever when a peer dies. **No new behavior is exposed at the Python wrapper layer in this PR** — only the kernel and the `torch.ops.trtllm.moe_a2a_*` op signatures.

### Background

When a single GPU dies in a Wide-EP group, the surviving ranks today spin forever on `completion_flags` in symmetric memory: the dispatch and combine kernels poll a flag word per peer and have no timeout, no abort, and no fallback. The system limps along until the host-side `HangDetector` fires after 5 minutes, then a full executor restart adds 2-3 more minutes of warmup — so a single GPU failure costs ~7-8 minutes of full downtime. This PR is the kernel-level half of the fix: the kernels themselves now have a way to know which peers are alive, so the spin loops can skip dead ranks instead of waiting for them. The Python wrapper that actually feeds the mask in (sourced from `EPGroupHealth`) is a follow-up PR.

### What this PR does

The mask is plumbed through every layer that touches the kernels:

- **Kernel ABI:** `kRankMaskWords = 2` and `uint64_t active_rank_mask[kRankMaskWords]` added to `MoeA2ADispatchParams`, `MoeA2ACombineParams`, and the kernel-side pointer structs. `kMaxRanks` is bumped 64 → 128 to cover NVL72 (72 ranks) with headroom.
- **Dispatch kernel:** masked targets are skipped on (a) per-token routing — a token whose top-k expert lives on a dead rank collapses to the same `-1` sentinel that combine already uses for duplicates, (b) the `recv_counters` store loop, (c) the EPLB stats write loop, (d) the completion-flag write loop, and most importantly (e) the completion-flag wait loop where the hang lives.
- **Combine kernel:** masked peers are skipped on the completion-flag write and wait loops. The per-token reduction needs no explicit mask check because dispatch already set `topk_send_indices[k] = -1` for dead-targeted slots, and the existing `dst_idx < 0` guard handles them.
- **Torch op:** `moe_a2a_dispatch` and `moe_a2a_combine` gain an optional `Tensor? active_rank_mask=None` parameter (CPU `uint64[2]`). A small `resolveActiveRankMask` helper validates dtype/device/shape and defaults to all-ones when omitted, so existing call sites (including the unchanged `MoeAlltoAll` Python wrapper) work bit-identically.

Key design choices:

- **Single bit-test per peer iteration.** `is_rank_active(mask, rank)` is one word load + one shift + one mask + one branch. The dispatch and combine peer loops run `O(ep_size)` iterations once per kernel invocation, so the overhead at NVL72 is at most 72 extra branches per launch — well inside the <0.1% steady-state regression gate.
- **Mask lives in the launch-param struct, not as a separate kernel arg.** Keeps the change localised and means launch sites that zero-initialise their params (`MoeA2ADispatchParams params{};`) get the safe default automatically; `resolveActiveRankMask` then overwrites with the user mask or all-ones.
- **Local rank's bit must always be set.** Asserted at launch time and at the torch-op boundary; the kernel itself is running on the local rank, so a cleared self-bit indicates an upstream bug rather than a recoverable state.
- **Dead-target tokens are dropped rather than re-routed.** The kernel-layer responsibility is "make AlltoAll survive a rank failure"; choosing where else to send a dead-targeted token is an EPLB-layer concern (separate follow-up PR). Dropping is the same code path as the existing duplicate handling, so combine needs no new logic.

## Test Coverage

`tests/unittest/_torch/modules/moe/test_moe_comm.py` — MPI-driven multi-GPU test, follows the same `MnnvlMemory.supports_mnnvl()` skip pattern as the existing `test_moe_a2a.py`. Both tests bypass the `MoeAlltoAll` Python wrapper and call `torch.ops.trtllm.moe_a2a_dispatch` / `moe_a2a_combine` directly so they exercise the new C++ op signature without depending on the follow-up wrapper PR.

- **`test_all_active_mask_matches_no_mask`** — regression guard. Two consecutive dispatch + combine rounds on identical input on `ep_size = 4`, one with `mask = None` and one with `mask = all-ones`. Asserts bit-identical combined output and identical `topk_target_ranks` workspace state. Parametrized over `(local_num_tokens, top_k) ∈ {(16, 2), (32, 4)}`.
- **`test_one_rank_masked_completes`** — parametrized over `dead_rank ∈ {0, 2, 3}` to cover the lowest-numbered, mid, and highest-numbered ranks (so the bit-mask logic is exercised at both ends of word 0). The "dead" rank participates in workspace init (which has its own MPI barrier), then sits at `MPI.COMM_WORLD.barrier()`; the surviving ranks call dispatch + combine with the dead rank's bit cleared. Asserts (a) the test reaches its assertions at all (no hang), (b) on every surviving rank, every top-k slot whose expert routed to `dead_rank` was dropped to `-1`, (c) all other slots match what the contiguous-partition routing rule predicts, and (d) the combined output has the expected `(local_num_tokens, hidden_size)` shape.

Both tests need MNNVL hardware (GB200) and `ep_size` GPUs to actually run; they skip cleanly elsewhere. Run: `mpirun -np 4 pytest tests/unittest/_torch/modules/moe/test_moe_comm.py -k rank_mask -v`.

## PR Checklist

## GitHub Bot Help

## Post-merge scope clarification (2026-06-30)

This merged PR is the **launch-time / next-launch rank-mask primitive**. The two mask words are copied by value into the dispatch/combine launch structures. Therefore:

- a kernel that was launched while all peers were active cannot observe a watchdog update after a peer dies;
- the existing multi-GPU test masks a rank that remains alive, so it proves mask behavior but not destructive peer-death escape;
- the internal `-1` / zero-filled contribution for a masked target is never valid model output from a failed execution epoch.

Promoted MVP item **1a.8** owns a running-kernel-observable abort/mask generation and a recoverable return path that replaces the 300-second `trap;` backstop. Item **1c.4b** must discard the failed epoch and atomically commit placement, survivor communicators, mask, and generation before serving resumes.

Canonical plan: https://github.com/chienchunhung/TensorRT-LLM/blob/docs-and-plans/docs/design/wide-ep-fault-tolerance/pr-execution/08-implementation-plan.md


chienchunhung:
/bot run --disable-fail-fast --stage-list "DGX_B200-PyTorch-2,DGX_B200-4_GPUs-PyTorch-Ray-1"

tests/unittest/_torch/multi_gpu/test_moe_a2a_rank_mask.py
```diff
@@ -0,0 +1,407 @@
...
+    hidden_size = 1024
+    num_experts = 32
+    workspace_size_per_rank = 256 * 1024 * 1024
+
+    args = (ep_size, local_num_tokens, top_k, workspace_size_per_rank, num_experts, hidden_size)
+    results = list(
+        mpi_pool_executor.map(
+            _worker_all_active_matches_no_mask,
+            *zip(*[args] * ep_size),
+        )
```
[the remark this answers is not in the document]
chienchunhung: fixed

tests/unittest/_torch/multi_gpu/test_moe_a2a_rank_mask.py
```diff
@@ -0,0 +1,407 @@
...
+    ],
+    indirect=["mpi_pool_executor"],
+)
+def test_all_active_mask_matches_no_mask(mpi_pool_executor, local_num_tokens, top_k):
+    """An all-ones active_rank_mask must produce identical output to omitting it."""
+    try:
+        MnnvlMemory.initialize()
+        assert MnnvlMemory.supports_mnnvl()
+    except Exception:
+        pytest.skip("MNNVL not supported on this system")
```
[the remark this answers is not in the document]
chienchunhung: fixed

cpp/tensorrt_llm/thop/moeAlltoAllOp.cpp
```diff
@@ -42,6 +42,40 @@ inline size_t alignOffset(size_t offset, size_t alignment)
...
+    TORCH_CHECK(t.is_contiguous(), "active_rank_mask must be contiguous");
+    auto const* src = static_cast<uint64_t const*>(t.const_data_ptr());
+    for (int w = 0; w < kRankMaskWords; ++w)
+    {
+        out[w] = src[w];
+    }
+    // Local rank's bit must be set; otherwise the kernel would be running on a "dead" rank.
+    TORCH_CHECK((out[epRank >> 6] >> (epRank & 63)) & 1ULL, "active_rank_mask must mark the local ep_rank (", epRank,
+        ") as active");
+}
```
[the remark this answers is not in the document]
chienchunhung: fixed

xxi-nv:
@bobboli Could you help to review this PR?

tests/unittest/_torch/multi_gpu/test_moe_a2a_rank_mask.py
```diff
@@ -0,0 +1,397 @@
...
+def _make_payload(local_num_tokens: int, hidden_size: int, rank: int) -> torch.Tensor:
+    """Deterministic per-rank payload so we can assert exact bit-for-bit equality."""
+    base = torch.arange(local_num_tokens * hidden_size, dtype=torch.bfloat16, device="cuda").view(
+        local_num_tokens, hidden_size
+    )
+    # Encode rank into the payload so cross-rank mismatches are immediately visible.
+    return base + (rank * 1000.0)
+
+
+def _read_topk_target_ranks(moe_a2a: MoeAlltoAll, max_num_tokens: int, top_k: int) -> torch.Tensor:
```
xxi-nv: Given that most communication methods will support the common mask function, I suggest adding the test to the unified Communication test framework - test_moe_comm.py rather than creating a new test file

chienchunhung: Thanks for the feedback! Agreed that consolidating the test is better; applied as suggested. 

I moved the `active_rank_mask` coverage into the unified MoE communication test module (`tests/unittest/_torch/modules/moe/test_moe_comm.py`) as focused NVLinkOneSided rank-mask cases, and removed the standalone `test_moe_a2a_rank_mask.py` file.

chienchunhung:
/bot run --disable-fail-fast --stage-list "RTXPro6000D-PyTorch-1, DGX_B200-4_GPUs-PyTorch-1, GB200-4_GPUs-PyTorch-2"

tests/unittest/_torch/modules/moe/test_moe_comm.py
```diff
@@ -114,6 +114,10 @@
 # to avoid _WORKSPACE singleton assertion failures.
 NVLINK_WORKSPACE_MB = "512"
 
+# Must match kRankMaskWords in
+# cpp/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.h.
+EP_MASK_NUM_WORDS = 2
```
xxi-nv: Is there a way to make it a singleton or a reference? Otherwise, we have to maintain two places to ensure they are consistent.

xxi-nv (approved):

tests/unittest/_torch/modules/moe/test_moe_comm.py
chienchunhung: Addressed in baf8919e. The test no longer keeps its own EP_MASK_NUM_WORDS literal; it now builds active-rank mask words through EPGroupHealth, so the test references the production mask-word source instead of duplicating it.

chienchunhung:
/bot run --disable-fail-fast --stage-list "A30-AutoDeploy-1, DGX_B200-AutoDeploy-1, H100_PCIe-AutoDeploy-1, DGX_B200-4_GPUs-PyTorch-Ray-1, DGX_B200-8_GPUs-PyTorch-1, GB200-8_GPUs-2_Nodes-PyTorch-1"

liji-nv (approved):

chienchunhung:
/bot run --disable-fail-fast --stage-list "DGX_B200-PyTorch-5"

chienchunhung:
/bot run --disable-fail-fast --stage-list "DGX_B200-PyTorch-5"

pcastonguay (approved):

PR #13426, opened 2026-04-24
[None][perf] EAGLE3 dynamic tree kernel optimizations

sunnyqgg:
## Summary

EAGLE3 one-model **dynamic tree** speculative decoding performance optimizations + correctness fixes.

## Changes

**Perf**
- Pack dynamic tree retrieve into single `[B,N,3]` tensor + optimize slot ID HtoD copies
- `_get_compact_packed_mask_view()` directly writes packed mask in compact 1D layout (eliminates `_repack_mask_padded_to_packed()` overhead)
- Buffer dtype `int64 → int32` for `draft_tokens_buffer` / `position_ids_buffer` / `history_draft_tokens_buffer` / step0 input buffers (eliminates per-layer `.to(int32)` kernels inside CUDA graph)
- Fast path in `compute_spec_dec_packed_mask()` for `num_blocks==1` (dt ≤ 32)
- `torch.compile` fused ops for EAGLE3 dynamic tree
- Extract `DynamicTreeSlotStorage` and optimize slot-indexed read path

**Correctness**
- Fix `position_offsets_stride` for dynamic tree multi-batch correctness
- Fix `position_offsets_query_len` and packed mask width for Hopper vs Blackwell
- Unify packed mask layout for Hopper and Blackwell via `cumSeqLensQ`
- Decouple `position_offsets` query length from buffer stride and fix Hopper mask width
- Fix target model packed mask layout and position_ids for Hopper dynamic tree
- CUDA-graph-safe 1D→2D reshape of dynamic tree position offsets via `position_offsets_stride`
- Separate KV cache reservation from draft token count for dynamic tree (fixes CUDA illegal address during warmup)


tensorrt_llm/_torch/pyexecutor/model_engine.py
mikeiovine, line 763: Does this have to use `max_draft_loop_tokens`? Similar comment for other `get_num_available_tokens` callsites.

mikeiovine (approved):

tensorrt_llm/_torch/pyexecutor/model_engine.py
sunnyqgg: I think so.
```
            self.max_draft_loop_tokens = max(
                self.original_max_total_draft_tokens,
                spec_config.dynamic_tree_max_topK * spec_config.max_draft_len)
```
The reason for this is that the number of tokens the target model infers might be smaller than the draft model's. For example, the draft model infers 30 tokens and the target model only infers 6 tokens.

cpp/tensorrt_llm/kernels/speculativeDecoding/dynamicTreeKernels.cu
```diff
@@ -264,89 +264,71 @@ void invokeBuildDynamicTree(int64_t const* parentList, int64_t const* selectedIn
...
-//! \param targetPredict        [in]  target model prediction per position [bs * numDraftTokens]
-//! \param batchSize            batch size
-//! \param numSpeculativeTokens second dim of acceptIndex (>= max possible accepts + 1)
-//! \param numDraftTokens       total tree nodes per batch (including root)
-__global__ void verifyDynamicTreeGreedyKernel(int64_t* predicts, int64_t* acceptIndex, int64_t* acceptTokenNum,
-    int64_t* acceptToken, int64_t const* candidates, int32_t const* retrieveIndex, int32_t const* retrieveNextToken,
-    int32_t const* retrieveNextSibling, int64_t const* targetPredict, bool const* treeValid, uint32_t batchSize,
+//! retrievePacked layout [bs, numDraftTokens, 3] int32 row-major:
+//! [b,n,0]=retrieveIndex, [b,n,1]=retrieveNextToken, [b,n,2]=retrieveNextSibling
+__global__ void verifyDynamicTreeGreedyPackedKernel(int32_t* acceptIndex, int32_t* acceptTokenNum, int32_t* acceptToken,
```
eopXD: Why do we need to trim the Doxygen comments here? I think other existing parameters should still hold. Please recover them and add your new packed parameter among them.

cpp/tensorrt_llm/kernels/speculativeDecoding/dynamicTreeKernels.h
```diff
@@ -60,36 +60,8 @@ void invokeBuildDynamicTree(int64_t const* parentList, int64_t const* selectedIn
...
-//! \param targetPredict [batchSize, numDraftTokens], on GPU. int64.
-//! Target model predictions.
-//! \param batchSize runtime::SizeType32. Batch size.
-//! \param numDraftTokens runtime::SizeType32. Total number of draft tokens.
-//! \param numSpecStep runtime::SizeType32. Number of speculative steps.
-//! \param stream cuda stream
-void invokeVerifyDynamicTreeGreedy(int64_t* predicts, int64_t* acceptIndex, int64_t* acceptTokenNum,
-    int64_t* acceptToken, int64_t const* candidates, int32_t const* retrieveIndex, int32_t const* retrieveNextToken,
-    int32_t const* retrieveNextSibling, int64_t const* targetPredict, bool const* treeValid,
+void invokeVerifyDynamicTreeGreedyPacked(int32_t* acceptIndex, int32_t* acceptTokenNum, int32_t* acceptToken,
```
eopXD, line 88: Ditto for the deleted comments.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -406,6 +406,17 @@ def append_to_kv_heads_per_layer(num_kv_heads_per_layer: List[int],
                                        1) if spec_config is not None else 0
         self.linear_attention_metadata = linear_attention_metadata
 
+        # Dynamic-tree draft manager reserves K*max_draft_len KV slots (the draft
+        # loop can write that many even if max_total_draft_tokens is smaller).
+        # Target manager keeps max_total_draft_tokens exactly.
+        self._kv_reserve_draft_tokens = self.max_total_draft_tokens
```
eopXD, line 626: May you add test coverage for this fix?

eopXD (approved):
Regarding the resource manager parts, the changes looks good to me with minor comments that needs to be addressed.

cpp/tensorrt_llm/kernels/speculativeDecoding/dynamicTreeKernels.cu
sunnyqgg: Done,thanks

cpp/tensorrt_llm/kernels/speculativeDecoding/dynamicTreeKernels.h
sunnyqgg: Done,thanks

tensorrt_llm/_torch/pyexecutor/resource_manager.py
sunnyqgg: Thanks,done

nv-guomingz (approved):

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -1278,12 +1279,10 @@ def add_dummy_requests(
         num_extra_decoding_steps: int = 0,
         draft_kv_cache_manager: Optional[KVCacheManager] = None,
     ) -> List[LlmRequest]:
-        requests = super().add_dummy_requests(request_ids, token_nums, is_gen,
-                                              prepare_resource,
-                                              max_num_draft_tokens, use_mrope,
-                                              max_beam_width,
-                                              num_extra_decoding_steps,
-                                              draft_kv_cache_manager)
+        requests = super().add_dummy_requests(
```
2ez4bz, line 1307: Nit: could we consider a kwarg-style invocation here? i.e.

```python
requests = super().add_dummy_requests(
    x=x,
    y=y,
    ...,
)
```

2ez4bz (approved):

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
sunnyqgg: The pre-commit hook controls this format, right?

2ez4bz: I meant like:

```python
requests = super().add_dummy_requests(
    request_ids=request_ids,
    token_nums=token_nums,
   ...,
)
```

that way it's not sensitive to argument order.

sunnyqgg: Thanks , done

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -228,6 +228,18 @@ def __init__(
...
 
+        # Dynamic tree draft loop produces up to K * max_draft_len tokens,
+        # which may exceed max_total_draft_tokens. Use the larger value for
+        # KV cache reservation only; verify/tree output stays at max_total_draft_tokens.
+        if (spec_config is not None
+                and getattr(spec_config, 'use_dynamic_tree', False)
+                and getattr(spec_config, 'dynamic_tree_max_topK', 0) > 0):
+            self.max_draft_loop_tokens = max(
+                self.original_max_total_draft_tokens,
+                spec_config.dynamic_tree_max_topK * spec_config.max_draft_len)
```
PerkzZheng, line 239: what is the difference between original_max_total_draft_tokens and max_draft_len ?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1562,6 +1589,19 @@ def _run(
...
 
+    @staticmethod
+    def _reshape_position_offsets_for_cpp(metadata):
+        """Reshape 1D spec-dec position offsets to (num_requests, query_len) for C++."""
+        offsets = metadata.spec_decoding_position_offsets
+        if offsets is None or offsets.dim() != 1:
+            return offsets
+        num_requests = metadata.max_num_requests
+        query_len = metadata.position_offsets_query_len
+        if num_requests > 0 and query_len > 0:
```
PerkzZheng: just curious when they will be 0 ?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -850,25 +861,36 @@ def update_spec_dec_param(
...
+                                                 0, slot_ids)[:, :n_dt]
+                    pos_dst = self.spec_decoding_position_offsets[:num_gens *
+                                                                  n_dt].view(
+                                                                      num_gens,
+                                                                      n_dt)
+                    pos_dst.copy_(pos_src, non_blocking=True)
+                    self.position_offsets_query_len = n_dt
+
+                    actual_mask_width = math.ceil(n_dt / 32)
+                    mask_src = torch.index_select(
```
PerkzZheng: it might be passible to merge the above operations. Is the num_gen > 0 condition made in order to save unnecessary operations for pure context requests ?

PerkzZheng (approved):

tensorrt_llm/_torch/pyexecutor/model_engine.py
sunnyqgg: `max_draft_len` refers to the maximum depth of the tree while `original_max_total_draft_tokens` refers to the number of tokens target model will verify.

tensorrt_llm/_torch/attention_backend/trtllm.py
sunnyqgg: Yes, exactly. The `num_gen > 0` condition is there to avoid unnecessary operations when the batch contains only context (prefill) requests

tensorrt_llm/_torch/attention_backend/trtllm.py
sunnyqgg: This layer of guard is merely to prevent the silent reshape of tensors with incorrect shapes when metadata is abnormally initialized from being passed to the C++ kernel, serving as an additional boundary protection.

tests/unittest/_torch/speculative/test_eagle3.py
```diff
@@ -951,8 +951,8 @@ def test_eagle3_lora(use_cuda_graph: bool):
     llm_spec.shutdown()
 
 
-@pytest.mark.parametrize("disable_overlap_scheduler", [False, True])
-@pytest.mark.parametrize("use_cuda_graph", [False, True])
+@pytest.mark.parametrize("disable_overlap_scheduler", [False])
+@pytest.mark.parametrize("use_cuda_graph", [True])
```
yuxianq, line 955: Why we remove other 3 cases?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -746,6 +757,12 @@ def prepare_resources(self, scheduled_batch: ScheduledRequests):
                 self.impl.add_token(req.py_request_id)
                 for _ in range(draft_len):
                     self.impl.add_token(req.py_request_id)
+                # Top up to _kv_reserve_draft_tokens (see __init__): on
+                # dynamic-tree draft manager this exceeds draft_len; otherwise
+                # only nonzero when scheduler truncated py_draft_tokens.
+                reserve_slack = self._kv_reserve_draft_tokens - draft_len
+                for _ in range(max(0, reserve_slack)):
```
yuxianq: Don't need another loop, can merge with the previous loop and use `for _ in range(max(draft_len, self._kv_reserve_draft_tokens)):` instead.

tests/unittest/_torch/speculative/test_eagle3.py
sunnyqgg: Since this features are low priority, there's no need to waste CI resources on testing it

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -850,25 +861,36 @@ def update_spec_dec_param(
...
+                        slot_storage.packed_mask, 0,
+                        slot_ids)[:, :, :actual_mask_width]
+                    total = num_gens * n_dt * actual_mask_width
                     self.spec_decoding_packed_mask.view(-1)[:total].copy_(
-                        spec_tree_manager.spec_dec_packed_mask[:batch_size].
-                        reshape(-1),
-                        non_blocking=True)
+                        mask_src.reshape(-1), non_blocking=True)
+                else:
+                    self.position_offsets_query_len = n_dt
```
yuxianq: Both if/else case use `position_offsets_query_len = n_dt`, we can move it to outer scope.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -883,6 +905,7 @@ def update_spec_dec_param(
                 # Case 2: static tree and target model
                 if is_target_model:
                     # For the target model, we update the spec-dec parameters with the spec_tree_manager, which is prepared in advance.
+                    self.position_offsets_query_len = 0
```
yuxianq: We have multiple `position_offsets_query_len = 0`, how about set it to 0 for all cases first, and only update it if necessary?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1291,11 +1321,8 @@ def _run(
             metadata.is_spec_decoding_enabled, metadata.use_spec_decoding,
             metadata.is_spec_dec_tree
         ]
-        position_offsets_for_cpp = metadata.spec_decoding_position_offsets
-        if (metadata.spec_decoding_position_offsets is not None
-                and metadata.spec_decoding_position_offsets.dim() == 1):
-            position_offsets_for_cpp = metadata.spec_decoding_position_offsets.view(
-                metadata.max_num_requests, -1)
+        position_offsets_for_cpp = self._reshape_position_offsets_for_cpp(
```
yuxianq: Instead of reshaping it in forward, how about move it to `update_spec_dec_param`? We can immediately reshape it after we get `spec_decoding_position_offsets` and `position_offsets_query_len`. Add a paired `spec_decoding_position_offsets_cpp` to `TrtllmAttentionMetadata`, `position_offsets_query_len` can be a local variable now, and move `_reshape_position_offsets_for_cpp` to `TrtllmAttentionMetadata`.

tests/unittest/_torch/speculative/test_eagle3.py
yuxianq: Got it.

tensorrt_llm/_torch/attention_backend/trtllm.py
sunnyqgg: Make sense,done, thanks. I think the cpp view needs to be refreshed anywhere
  the effective query_len changes, not only in update_spec_dec_param(). In
  dynamic-tree mode _apply_spec_metadata() later updates the per-step query_len, so
  spec_decoding_position_offsets_cpp need to update there
  as well.

tensorrt_llm/_torch/attention_backend/trtllm.py
sunnyqgg: Thanks,done

tensorrt_llm/_torch/attention_backend/trtllm.py
sunnyqgg: Thanks, done

tensorrt_llm/_torch/pyexecutor/resource_manager.py
sunnyqgg: Thanks, done

yuxianq: Seems that it is unchanged.

yuxianq (approved):

tensorrt_llm/_torch/pyexecutor/resource_manager.py
sunnyqgg: Miss it, done,thanks a lot

PR #13444, opened 2026-04-24
[None][feat] Add Llama sequence classification / reward-model support

pst2154:
## Summary by CodeRabbit

* **New Features**
  * Added support for Llama sequence classification models, enabling classifier and reward model use cases. Models can now be loaded from Hugging Face and converted to TensorRT-LLM format.

* **Documentation**
  * Updated support matrix and model documentation to include the new classifier architecture.

* **Tests**
  * Added comprehensive unit tests for sequence classification model validation against Hugging Face implementations.

## Description

Add support for `LlamaForSequenceClassification` in the PyTorch backend so that Llama-based reward / classifier checkpoints (e.g. `Skywork/Skywork-Reward-Llama-3.1-8B-v0.2`) can be served via the TensorRT-LLM PyTorch runtime.

Implementation follows the existing `Qwen2ForRewardModel` pattern:

- Register `@register_auto_model("LlamaForSequenceClassification")` for a new class that inherits from `DecoderModelForCausalLM[LlamaModel, LlamaConfig]` and bypasses its `__init__` via `nn.Module.__init__(self)`, so that the vocab-sized `lm_head` is replaced with a small `score` head of size `num_labels`. The parent's `PostInitCaller` metaclass still triggers `__post_init__`, which handles `create_weights()` on leaf modules.
- Forward pass returns the per-request score read from the last token of each sequence: `logits[cumsum(seq_lens) - 1]`.
- `is_generation_model` returns `False` for this architecture (same as `Qwen2ForRewardModel`, `BertForSequenceClassification`, `LlamaForTextEmbedding`).
- Legacy TensorRT engine-build path in `tensorrt_llm/models/llama/{config,convert,model}.py` and `MODEL_MAP` in `tensorrt_llm/models/__init__.py` route `LlamaForSequenceClassification` through `LLaMAForCausalLM`, reusing the Llama weight converter with the `score` head mapped to `lm_head`.
- `LlamaConfig.num_labels` is plumbed through `from_hugging_face` and `to_dict` for checkpoint round-trip.

## Test Coverage

- `tests/unittest/_torch/modeling/test_modeling_llama.py::TestLlama::test_llama_sequence_classification_allclose_to_hf` builds a tiny random `LlamaForSequenceClassification` with 2 layers, `hidden_size=128`, `num_labels=3`, loads the same weights into the HF reference implementation, and asserts `torch.testing.assert_close(hf_logits, tllm_logits, rtol=1.5e-2, atol=1.5e-2)` on a batched forward through the VANILLA attention backend.

- Manually validated against the real `Skywork/Skywork-Reward-Llama-3.1-8B-v0.2` reward model on RTX PRO 6000 in `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc12` (bf16, TRTLLM attention backend):

  | prompt | HF `AutoModelForSequenceClassification` | TRT-LLM `LlamaForSequenceClassification` |
  | --- | ---: | ---: |
  | "What is 2+2?" &rarr; "2 + 2 = 4." (good) | 11.5625 | 11.75 |
  | "What is 2+2?" &rarr; "Banana." (bad)     | -22.5   | -22.625 |

  Max abs diff 0.1875 on scores of magnitude ~22 (within bf16 noise), and both implementations correctly rank the good answer above the bad.

## PR Checklist

### DIFF
--- a/docs/source/models/supported-models.md
+++ b/docs/source/models/supported-models.md
@@ -32,6 +32,7 @@ The following is a table of supported models for the PyTorch backend:
 | `KimiK25ForConditionalGeneration`    | Kimi-K2.5                          | `moonshotai/Kimi-K2.5`                       |
 | `LagunaForCausalLM`                  | Laguna-XS                          | `poolside/laguna-XS.2`                       |
 | `LlamaForCausalLM`                   | Llama 3.1, Llama 3, Llama 2, LLaMA | `meta-llama/Meta-Llama-3.1-70B`              |
+| `LlamaForSequenceClassification`     | Llama-based classifier/reward model | `Skywork/Skywork-Reward-Llama-3.1-8B-v0.2` |
 | `Llama4ForConditionalGeneration`     | Llama 4                            | `meta-llama/Llama-4-Scout-17B-16E-Instruct`  |
 | `MiniMaxM2ForCausalLM` [^5]          | MiniMax M2/M2.1/M2.7              | `MiniMaxAI/MiniMax-M2.7`                    |
 | `MiniMaxM3SparseForConditionalGeneration` [^12]| MiniMax-M3                       | `MiniMaxAI/MiniMax-M3`                      |
--- a/tensorrt_llm/_torch/model_config.py
+++ b/tensorrt_llm/_torch/model_config.py
@@ -324,7 +324,7 @@ class ModelConfig(Generic[TConfig]):
         return model_architectures[0] not in [
             "BertForSequenceClassification", "Qwen2ForProcessRewardModel",
             "Qwen2ForRewardModel", "LlamaForTextEmbedding",
-            "Qwen3ForTextEmbedding"
+            "LlamaForSequenceClassification", "Qwen3ForTextEmbedding"
         ]
         # TODO: should be 'not model_type == ModelType.ENCODER_ONLY'
         # once ModelType is used in pytorch flow.
--- a/tensorrt_llm/_torch/models/__init__.py
+++ b/tensorrt_llm/_torch/models/__init__.py
@@ -31,7 +31,7 @@ from .modeling_hunyuan_moe import HunYuanMoEV1ForCausalLM
 from .modeling_hyperclovax import HCXVisionForCausalLM
 from .modeling_kimi_k25 import KimiK25ForConditionalGeneration
 from .modeling_laguna import LagunaForCausalLM
-from .modeling_llama import LlamaForCausalLM
+from .modeling_llama import LlamaForCausalLM, LlamaForSequenceClassification
 from .modeling_llava_next import LlavaNextModel
 from .modeling_minimaxm2 import MiniMaxM2ForCausalLM
 from .modeling_minimaxm3 import (MiniMaxM3ForCausalLM,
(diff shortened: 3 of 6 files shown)


docs/source/legacy/reference/support-matrix.md
laikhtewari, line 1: Does this work out of the box on TRT backend? I think you would still need the modeling definition which I don't think we have upstream

If not, I'd suggest removing here and just documenting support on current support matrix i.e. for Pytorch backend

laikhtewari (changes requested):

docs/source/legacy/reference/support-matrix.md
pst2154: Removed this row from the legacy support matrix so we don't imply TensorRT backend support. It remains documented in the current PyTorch backend supported models table.

pst2154 (commented):
Addressed the review comments:

- Removed `LlamaForSequenceClassification` from the legacy support matrix and kept it documented in the current PyTorch backend supported models table.
- Added/updated NVIDIA copyright headers for modified files.
- Made HF architecture detection robust by checking membership in `hf_config.architectures`.
- De-duplicated `lm_head` construction while preserving the sequence-classification output size.
- Added a unit assertion that `LlamaForSequenceClassification` is not treated as a generation model.

laikhtewari (approved):

karljang:
Thanks @pst2154, clean implementation and great validation.

 A few things before this can move:

1. Rebase - the PR is currently conflicting with main.
2. DCO - please sign off your commits (git commit -s).

Additionally, this adds both the PyTorch path and the legacy TensorRT engine-build path (models/llama/...), which is being deprecated. Could you please drop the legacy half and keep this PyTorch-only? It removes ~5 files and the num_labels plumbing.

Thanks again, happy to take this once it's rebased and trimmed.

pst2154:
Addressed all three requests in `d86626ea3a`: rebuilt the PR cleanly on current main, reduced it to one DCO-signed commit, and removed the legacy TensorRT engine-build implementation entirely. The PR is now PyTorch-only and touches six files. The B300 Hugging Face parity check passed with max absolute error `0.00024414`; file-scoped pre-commit, byte compilation, and diff checks also pass.

BowenFu (approved):
LGTM. The only touch to existing code is the FULL-vs-CAUSAL mask in `LlamaDecoderLayer.__init__`, and it's gated on `LlamaForSequenceClassification` being in `architectures`, so `LlamaForCausalLM` / `LlamaForTextEmbedding` and the Qwen reward models are unchanged. Rest is additive (registry, `is_generation_model` list, serialization allowlist), and the allclose-to-HF test is a real check.

Non-blocking: the mask check tests membership in the whole `architectures` list while model resolution and `is_generation_model()` use `architectures[0]` — `architectures[0] == ...` would keep those consistent. Longer term this would be cleaner as an explicit causal-attention property on the model config rather than an architecture-name special case in the decoder layer, since more decoder-based reward models will land.

YihuiLu512 (approved):
LGTM

litaotju:
THE PR is inactive for more than 2 months. Closing. 

If you still want to add it, please rebase to latest main and re-open.

PR #13449, opened 2026-04-24
[None][feat] Add Qwen image support

pst2154:
## Summary
- Adds support for Qwen image workflows.

## Validation
- Validation was run before filing this PR.
- Added Qwen-Image parity coverage against diffusers for transformer components.
- Added registry smoke tests for Qwen-Image AutoPipeline detection.

## Summary by CodeRabbit

* **New Features**
  * Added support for Qwen-Image text-to-image generation model with full pipeline integration
  * Supports single and batch image generation with configurable parameters (image size, denoising steps, CFG guidance, negative prompts)
  * Added example with timing metrics and JSON output for batch inference

* **Documentation**
  * Updated feature matrix and serving documentation to include Qwen-Image with parameter specifications and supported features

* **Tests**
  * Added parity validation tests and registry integration tests for the new model


tensorrt_llm/_torch/visual_gen/models/qwen_image/pipeline_qwen_image.py
```diff
@@ -0,0 +1,538 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Qwen-Image text-to-image pipeline (Phase 1).
+
+Ports the denoise loop, Qwen2.5-VL text encoding, FlowMatchEuler
+sampling, and VAE decode from ``diffusers.QwenImagePipeline`` onto the
+TensorRT-LLM VisualGen executor. The transformer backbone is our port
+from ``transformer_qwen_image.py`` (see PHASE1_PLAN.md M2..M6).
```
chang-l: Please remove the `PHASE1*/PLAN`– the personal CC style comments.

tensorrt_llm/_torch/visual_gen/models/qwen_image/transformer_qwen_image.py
```diff
@@ -0,0 +1,885 @@
...
+"""Qwen-Image MMDiT transformer (Phase 1).
+
+Pure-PyTorch port of the reference implementation in
+``diffusers.models.transformers.transformer_qwenimage`` at commit
+matching diffusers 0.37.1. Module and attribute names mirror the
+diffusers reference exactly so the HuggingFace checkpoint's
+``transformer/*.safetensors`` state_dict can be loaded verbatim.
+
+Milestones completed so far (see
+``/home/scratch.asteiner/trtllm-qwen-image/PHASE1_PLAN.md``):
```
chang-l: Please update and take another look at your comments :)

tensorrt_llm/_torch/visual_gen/models/qwen_image/transformer_qwen_image.py
```diff
@@ -0,0 +1,885 @@
...
+        )
+        return freqs.clone().contiguous()
+
+
+# ===========================================================================
+# Joint self-attention for MMDiT (M5).
+# ===========================================================================
+
+
+class QwenJointAttention(nn.Module):
```
chang-l: This file needs significant rework. Please use the FLUX model implementation as a reference and leverage existing TRT-LLM modules instead of introducing plain PyTorch implementations for performance. For example, QwenJointAttention should derive from visual gen's Attention class, GELU should use the TRT-LLM implementation (within `_torch/module`), and the FFN should reuse the TRT-LLM MLP modules, etc.

chang-l (commented):
Does this PR also support Qwen-Image-2512 : https://huggingface.co/Qwen/Qwen-Image-2512?

tensorrt_llm/_torch/visual_gen/models/qwen_image/pipeline_qwen_image.py
pst2154: Cleaned this up in the latest commit by removing the `PHASE1`/plan-style wording from the pipeline module docstring.

tensorrt_llm/_torch/visual_gen/models/qwen_image/transformer_qwen_image.py
pst2154: Updated the stale implementation comments in the transformer/package docs and removed the phase/milestone wording. I also cleaned up the matching registry test docstring so it no longer describes the transformer as a stub.

tensorrt_llm/_torch/visual_gen/models/qwen_image/transformer_qwen_image.py
pst2154: Reworked this in `f85525e783` following the FLUX/VisualGen pattern:

- `QwenJointAttention` now derives from VisualGen `Attention` and uses the shared attention path for the unmasked case.
- Q/K/V, output projections, top-level projections, and modulation projections now use TRT-LLM `Linear` with the `DiffusionModelConfig` quantization/weight-loading hooks.
- The Qwen FFN now subclasses TRT-LLM `MLP`; the diffusers `net.0/net.2` checkpoint keys are remapped during load.
- Added `DynamicLinearWeightLoader`/`post_load_weights()` handling for the TRT-LLM modules.

Validated on a GPU node inside `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc12` with the focused Qwen Image tests:

```bash
QWEN_IMAGE_CKPT=/models/qwen-image python3 -m pytest /tmp/qwen_tests/test_qwen_image_registry.py /tmp/qwen_tests/test_qwen_image_parity.py -q
```

Result: `22 passed`.

pst2154:
Pushed `f85525e783` with the review fixes.

Summary:
- Addressed the CodeRabbit items: fixed the README table, prompt-file empty/comment handling, missing NVIDIA headers, negative prompt broadcasting, safe rank logging, explicit transformer argument validation, `assert` -> `ValueError`, raw regex string, and removed the developer-local checkpoint default.
- Reworked the Qwen Image transformer to reuse TRT-LLM/VisualGen modules per the reviewer request.
- Ran local checks:
  - `python -m py_compile ...`
  - `git diff --check`
- Ran focused GPU validation inside `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc12` on `alon-ts1-iec-05`:
  - `QWEN_IMAGE_CKPT=/models/qwen-image python3 -m pytest /tmp/qwen_tests/test_qwen_image_registry.py /tmp/qwen_tests/test_qwen_image_parity.py -q`
  - Result: `22 passed`.

On Qwen-Image-2512: its Hugging Face `transformer/config.json` uses the same `QwenImageTransformer2DModel` class and matching core transformer dimensions (`attention_head_dim=128`, `axes_dims_rope=[16,56,56]`, `in_channels=64`, `joint_attention_dim=3584`, `num_attention_heads=24`, `num_layers=60`, `out_channels=16`, `patch_size=2`), so the implementation is expected to be compatible through the same Qwen Image path. I have validated this commit against `Qwen/Qwen-Image`; I have not yet run an end-to-end validation with a local `Qwen/Qwen-Image-2512` checkpoint.

pst2154:
Update on Qwen-Image-2512 validation: I downloaded `Qwen/Qwen-Image-2512` to the GPU host and ran the same focused validation in `nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc12`.

Command shape:

```bash
QWEN_IMAGE_CKPT=/models/qwen-image-2512 python3 -m pytest /tmp/qwen_tests/test_qwen_image_registry.py /tmp/qwen_tests/test_qwen_image_parity.py -q
```

Result: `22 passed, 37 warnings in 273.51s`.

So the transformer/registry parity path is now validated for both `Qwen/Qwen-Image` and `Qwen/Qwen-Image-2512`.

docs/source/models/visual-generation.md
```diff
@@ -43,6 +44,9 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
 | **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
 | **Wan 2.2** | Yes | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes |
 | **LTX-2** | Yes | Yes | No | Yes | Yes | No | No | Yes | Yes |
+| **Qwen-Image** [^qi1] | No | No | No | No | No | No | No | No | Yes |
```
chang-l: Can you help make sure the Qwen-Image support matrix is aligned with FLUX.2, i.e.:
```
| **FLUX.2** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes |
```

(Note: here, FP8/NVFP4 does not require FP8/FP4 checkpoints. VisualGen supports dynamic quantization, which means the checkpoint can be BF16.)

I don’t think there’s a gap in your current implementation, though I’m not sure about Ulysses. We just need to test and verify that these features work out of the box. You can refer to FLUX.2 to see how each feature is enabled in the code and E2E tests.

examples/visual_gen/serve/README.md
```diff
@@ -62,6 +63,7 @@ Current supported & tested models:
 2. FLUX.1 for image generation (t2i)
 3. FLUX.2 for image generation (t2i)
 4. LTX-2 for video generation with audio (t2v, ti2v)
+5. Qwen-Image for image generation (t2i, BF16 reference; FP8/CUDA-graph/parallelism pending follow-up PRs)
```
chang-l: once resolved the previous comment, pls update here too

examples/visual_gen/README.md
```diff
@@ -226,31 +276,35 @@ python visual_gen_ltx2.py \
...
-| `--ulysses_size` | ✓ | ✓ | — | 1 | Sequence parallelism |
-| `--linear_type` | ✓ | ✓ | — | default | Quantization type |
-| `--enhance_prompt` | — | ✓ | False | Gemma3 prompt enhancement |
-| `--stg_scale` | — | ✓ | 0.0 | Spatiotemporal guidance scale |
-| `--modality_scale` | — | ✓ | 1.0 | Cross-modal guidance scale |
-| `--rescale_scale` | — | ✓ | 0.0 | Variance-preserving rescale factor |
+| Argument | FLUX | Qwen-Image | WAN | LTX2 | Default | Description |
+|----------|------|------------|-----|------|---------|-------------|
+| `--model_path` | ✓ | ✓ | ✓ | ✓ | — | Path to model checkpoint directory |
+| `--text_encoder_path` | — | — | ✓ | ✓ | — | Path to Gemma3 text encoder |
```
chang-l: here, `text_encoder_path` should be 
```
| `--text_encoder_path` | — | ✓ | — | ✓ | — | Path to Gemma3 text encoder |
```
Right?
Please double check your changes regarding this section

tensorrt_llm/_torch/visual_gen/models/qwen_image/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+
+This package exposes ``QwenImagePipeline`` and the implemented
+``QwenImageTransformer2DModel`` stack used by the VisualGen Qwen-Image
+integration.
+"""
+
+from .pipeline_qwen_image import QwenImagePipeline
+from .transformer_qwen_image import (
+    AdaLayerNormContinuous,
+    FeedForward,
```
chang-l: I dont think we need to define `FeedForward` in the transformer

tensorrt_llm/_torch/visual_gen/models/qwen_image/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    QwenEmbedRope,
+    QwenImageTransformer2DModel,
+    QwenImageTransformerBlock,
+    QwenJointAttention,
+    QwenTimestepProjEmbeddings,
+    RMSNorm,
```
chang-l: Can we avoid defining RMSNorm inside the transformer model, and instead use F.rms_norm or `import RMSNorm from _torch.modules.rms_norm`?

You can refer to the FLUX.2 handling for the RMSNorm handling

tensorrt_llm/_torch/visual_gen/models/qwen_image/transformer_qwen_image.py
```diff
@@ -0,0 +1,987 @@
...
+            # attention_mask is (B, Sjoint) bool or float. Expand to
+            # (B, 1, 1, Sjoint) so SDPA broadcasts over (H, Sq).
+            attn_mask = attention_mask[:, None, None, :]
+
+        if attn_mask is None:
+            out = self._attn_impl(joint_q.transpose(1, 2).flatten(2),
+                                  joint_k.transpose(1, 2).flatten(2),
+                                  joint_v.transpose(1, 2).flatten(2))
+        else:
+            out = F.scaled_dot_product_attention(joint_q,
```
chang-l: Can we remove `if` and just do:
```
out = self._attn_impl(...)
```
Or does Qwen-Image need to support a custom mask, rather than always using bidirectional attention?

tensorrt_llm/_torch/visual_gen/models/qwen_image/transformer_qwen_image.py
```diff
@@ -0,0 +1,987 @@
...
+        super().__init__(
+            hidden_size=dim,
+            num_attention_heads=num_attention_heads,
+            head_dim=attention_head_dim,
+            qkv_mode=QKVMode.SEPARATE_QKV,
+            qk_norm=True,
+            qk_norm_mode="per_head",
+            eps=eps,
+            bias=True,
+            fuse_qk_norm_rope=False,
```
chang-l, line 461: I think we should be able to enable `fuse_qk_norm_rope=True` using the same kernel, as FLUX.2. We may just need a small format adaptation to support Qwen-Image’s complex frequency representation. 
If indeed we can reuse the same kernel for Flux, it is okay to enable it in next PR with TODO comment here :) 
cc. @karljang

tensorrt_llm/_torch/visual_gen/models/qwen_image/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    Timesteps,
+    TimestepEmbedding,
+    apply_rotary_emb_qwen,
```
chang-l, line 21: Can you double-check to see if it is possible to reuse `apply_rotary_emb` from visual gen's common attn module [here](https://github.com/NVIDIA/TensorRT-LLM/blob/e903428ccde994a3ff4c394ea576d75f4bccef25/tensorrt_llm/_torch/visual_gen/models/flux/attention.py#L20)?

tests/unittest/_torch/visual_gen/test_qwen_image_parity.py
```diff
@@ -0,0 +1,473 @@
...
+
+# ===========================================================================
+# Timestep embedding.
+# ===========================================================================
+
+
+@requires_ckpt
+@requires_cuda
+@pytest.mark.parametrize("timestep_value", [0.001, 0.25, 0.5, 0.99])
+def test_qwen_timestep_proj_embedding_parity(
```
chang-l: Similar to Flux.2 model, can we add more tests, including some unit/integration tests, e.g., `tests/unittest/_torch/visual_gen/test_flux_transformer.py`/`test_flux_pipeline.py`? 
btw, @karljang do we have any E2E accuracy test for FLUX.2 that we can use as a reference here?

chang-l (commented):
I dropped a few other comments and thanks for your patience!
Also adding @zhenhuaw-me to help review as well

docs/source/models/visual-generation.md
pst2154: Updated in `1744abfb18`: Qwen-Image now matches the FLUX.2-style row for dynamic FP8/NVFP4, Ulysses, CUDA graph, torch.compile, and serve support, with TeaCache still marked `No`. I also updated the footnote to call out that FP8/NVFP4 use VisualGen dynamic quantization from BF16 checkpoints.

examples/visual_gen/serve/README.md
pst2154: Updated in `1744abfb18` to match the feature matrix wording: Qwen-Image is listed with BF16, dynamic FP8/NVFP4, CUDA graph, Ulysses, and torch.compile support.

examples/visual_gen/README.md
pst2154: Yes, you were right. Updated the row in `1744abfb18` to `— | ✓ | — | ✓ | —`.

tensorrt_llm/_torch/visual_gen/models/qwen_image/__init__.py
pst2154: Cleaned this up in `1744abfb18` by removing `FeedForward` from the package-level import/export list. The internal class remains only as the Qwen-specific MLP wrapper used by the transformer block/checkpoint key mapping.

tensorrt_llm/_torch/visual_gen/models/qwen_image/__init__.py
pst2154: Done in `1744abfb18`. The custom Qwen `RMSNorm` class is removed; the transformer now imports `RMSNorm` from `tensorrt_llm._torch.modules.rms_norm`. For the 4D per-head Q/K tensors I apply `F.rms_norm(...)` with those shared module weights, matching the FLUX-style handling and avoiding FlashInfer's 4D RMSNorm limitation.

tensorrt_llm/_torch/visual_gen/models/qwen_image/transformer_qwen_image.py
pst2154: I kept the `if` and added a clarifying comment in `1744abfb18`. Qwen-Image does need the custom mask path because prompt encoding pads text embeddings; after concatenating `[text | image]`, the mask is needed to ignore padded text tokens. The unmasked case still uses the shared VisualGen `_attn_impl(...)` path.

tensorrt_llm/_torch/visual_gen/models/qwen_image/transformer_qwen_image.py
pst2154: Added the TODO in `1744abfb18`. I left `fuse_qk_norm_rope=False` for this PR and noted that enabling the shared fused kernel should be a follow-up after adapting Qwen's complex RoPE frequency cache to the shared real cos/sin format.

tensorrt_llm/_torch/visual_gen/models/qwen_image/__init__.py
pst2154: Double-checked and documented this in `1744abfb18`. The shared VisualGen helper takes real cos/sin tensors, while Qwen-Image keeps the diffusers-compatible complex-valued frequency cache. I kept the Qwen adapter for now and added a note that we can share this once we add a common complex RoPE helper or normalize Qwen's cache format.

tests/unittest/_torch/visual_gen/test_qwen_image_parity.py
pst2154: Added a focused transformer forward sanity test in `1744abfb18` that runs a tiny Qwen-Image transformer on CUDA through both the unmasked shared-attention path and the text-masked SDPA path. The existing checkpoint-backed parity tests still cover per-module/full-transformer diffusers parity when `QWEN_IMAGE_CKPT` is available.

pst2154:
Pushed `1744abfb18` with the latest review fixes.

Summary:
- Aligned the Qwen-Image feature matrix / serve docs with FLUX.2-style VisualGen support and clarified dynamic FP8/NVFP4 from BF16 checkpoints.
- Fixed the common-args `text_encoder_path` row.
- Removed package-level `FeedForward` / `RMSNorm` exports and removed the custom Qwen `RMSNorm` class.
- Switched per-head Q/K normalization to `F.rms_norm(...)` using the shared TRT-LLM `RMSNorm` weights.
- Kept the masked attention branch and documented why Qwen needs it for padded text tokens.
- Added TODO/commentary for future fused qk-norm+RoPE once Qwen's complex RoPE cache is adapted to the shared cos/sin format.
- Added a CUDA tiny-transformer forward sanity test for both unmasked and text-masked paths.

Validation:
- `python -m py_compile ...` passed.
- `git diff --check` passed.
- Targeted local pytest is still blocked by missing `mpi4py` in `tests/unittest/conftest.py`.
- In the TRT-LLM `1.3.0rc12` release container, validated the modified Qwen source as an overlay on the container's built TensorRT-LLM package: syntax ok, Qwen module import ok, tiny transformer forward ok for both `mask=False` and `mask=True`.

docs/source/models/visual-generation.md
yibinl-nvidia: resolved.

examples/visual_gen/serve/README.md
yibinl-nvidia: resolved.

tests/unittest/_torch/visual_gen/test_qwen_image_parity.py
yibinl-nvidia: resolved.

examples/visual_gen/README.md
yibinl-nvidia: resolved.

docs/source/models/visual-generation.md
```diff
@@ -32,18 +32,22 @@ TensorRT-LLM **VisualGen** provides a unified inference stack for diffusion mode
...
 
 | Model | FP8 blockwise | NVFP4 | TeaCache | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve | Attention2D | Ring Attention |
-|---|---|---|---|---|---|---|---|---|---|--|--|
+|---|---|---|---|---|---|---|---|---|---|---|---|
 | **FLUX.1** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes |
 | **FLUX.2** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes |
 | **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
 | **Wan 2.2** | Yes | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
 | **LTX-2** | Yes | Yes | No | Yes | Yes | No | No | Yes | Yes | Yes | Yes |
+| **Qwen-Image** [^qi1] | Yes | Yes | No | No [^1] | Yes | No | Yes | Yes | Yes | - | - |
```
yibinl-nvidia: Need to support Attention2D and Ring Attention like https://github.com/NVIDIA/TensorRT-LLM/pull/13944/changes#diff09d8e9e30baf114e6bd70bd4f9752b2a93afffac47f3284510372b0b131ad91eR231 after PR13944 merged.

yibinl-nvidia (approved):

docs/source/models/visual-generation.md
```diff
@@ -33,6 +33,7 @@ TensorRT-LLM **VisualGen** provides a unified inference stack for diffusion mode
 | `Wan-AI/Wan2.2-I2V-A14B-Diffusers` | Image-to-Video |
 | `Wan-AI/Wan2.2-TI2V-5B-Diffusers` | Text-to-Video, Image-to-Video |
 | `Lightricks/LTX-2` | Text-to-Video (with Audio), Image-to-Video (with Audio) |
+| `Qwen/Qwen-Image` | Text-to-Image |
```
chang-l, line 36: pls also list `Qwen/Qwen-Image-2512` here, as it should also be supported in this PR

docs/source/models/visual-generation.md
```diff
@@ -45,9 +46,12 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
 | **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
 | **Wan 2.2** | Yes | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
 | **LTX-2** | Yes | Yes | No | Yes | Yes | No | No | Yes | Yes | Yes | Yes | No |
+| **Qwen-Image** [^2] | Yes | Yes | No | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes |
```
chang-l: CFG Parallelism `No [^1]` looks incorrect. `[^1]` is FLUX-specific ("embedded guidance, no separate negative-prompt path"), but this pipeline implements real CFG — `true_cfg_scale` + `negative_prompt` (`do_true_cfg` / `neg_noise_pred` in `pipeline_qwen_image.py`). Either mark `Yes`, or write a Qwen-specific footnote if CFG-parallel just isn't wired up yet rather than reusing the FLUX rationale.

docs/source/models/visual-generation.md
```diff
@@ -45,9 +46,12 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
...
 
 [^1]: FLUX models use embedded guidance and do not have a separate negative prompt path, so CFG parallelism is not applicable.
 
+[^2]: Qwen-Image ships a native BF16 implementation with per-module numerical parity vs `diffusers.QwenImagePipeline` (cosine >= 0.999 on the full 20B transformer) and `trtllm-serve` / `/v1/images/generations` support. FP8 blockwise and NVFP4 use VisualGen dynamic quantization from BF16 checkpoints; no pre-quantized checkpoint is required. TeaCache integration is scheduled for a follow-up PR.
```
chang-l: Heads-up: the duplicated Supported-Models table and Feature Matrix in `docs/source/models/supported-models.md` weren't updated in this PR — they still have no Qwen row. Please add Qwen-Image (+2512) there too so the two docs stay in sync.

docs/source/models/visual-generation.md
chang-l: Also, here it marked CUDA Graph, torch.compile, Attention2D, Ring, TP all Yes. Can you double check? 
In this PR, it only validates only per-module parity + registry smoke tests, without any of these features.. Can we add more feature tests?

tests/unittest/_torch/visual_gen/test_qwen_image_registry.py
```diff
@@ -0,0 +1,129 @@
...
+    """model_index.json with _class_name=QwenImagePipeline resolves."""
+    (tmp_path / "model_index.json").write_text(json.dumps({"_class_name": "QwenImagePipeline"}))
+    assert AutoPipeline._detect_from_checkpoint(str(tmp_path)) == "QwenImagePipeline"
+
+
+@pytest.mark.parametrize(
+    "variant_class_name",
+    [
+        "QwenImageImg2ImgPipeline",
+        "QwenImageEditPipeline",
```
chang-l: Do we support this?  would all these models route to QwenImage pipeline?

tensorrt_llm/_torch/visual_gen/pipeline_registry.py
```diff
@@ -181,6 +181,10 @@ def _detect_from_checkpoint(checkpoint_dir: str) -> str:
                 return "Flux2Pipeline"
             if "Flux" in class_name:
                 return "FluxPipeline"
+            # Qwen-Image family (QwenImagePipeline, and future variants
+            # like QwenImageEditPipeline / QwenImageImg2ImgPipeline).
+            if "QwenImage" in class_name:
+                return "QwenImagePipeline"
```
chang-l, line 185: So, do we plan to map every class name containing QwenImage to the text-to-image QwenImagePipeline? For example: `qwen-image2image`, `qwen-image-edit`, etc.

If not, can we raise a NotImplementedError and tighten this check?

docs/source/models/visual-generation.md
yibinl-nvidia: modified.

docs/source/models/visual-generation.md
yibinl-nvidia: modified

docs/source/models/visual-generation.md
yibinl-nvidia: Theoretically, Qwen Image support CFG parallelism only when `negative_prompt` is present, but `negative_prompt` is an optional parameter. I will add CFG support to QWEN image in the follow up PR.

tensorrt_llm/_torch/visual_gen/pipeline_registry.py
yibinl-nvidia: this is a typo. `QwenImageEditPipeline` and `QwenImageImg2ImgPipeline` are never implemented.

tests/unittest/_torch/visual_gen/test_qwen_image_registry.py
yibinl-nvidia: No, I am not sure why they are added in the first place...

docs/source/models/visual-generation.md
yibinl-nvidia: Added more tests for cuda graph, attention2d, and ring attention. Also updated supporting metric table that TP is not supported.

chang-l (approved):

PR #13454, opened 2026-04-25
[None][fix] Split TRT-LLM-only rope fusion out of standalone auto_deploy

lucaslie:
## Summary

Fixes 22 standalone test failures introduced by #12357.

`tensorrt_llm/_torch/auto_deploy/transform/library/rope.py` added a top-level
`from ...custom_ops.attention.trtllm_attention import _TRTLLM_ROPE_INFO_KEY`,
which transitively imports `tensorrt_llm.bindings.internal.thop` and
`tensorrt_llm.functional`. In the standalone auto_deploy package those are
unavailable, so `rope.py` itself fails to import — taking down every
backend-agnostic RoPE transform with it. As a result, 22 rope/qkv tests
that have nothing to do with TRT-LLM started failing in standalone mode.

The fix moves the TRT-LLM-specific code out into its own module so the
standalone import-guard mechanism can skip exactly the TRT-LLM-only piece:

- `FuseRopeIntoTrtllmAttention` and its 7 helpers (`_FUSIBLE_ROPE_OPS`,
  `_trace_to_rope`, `_extract_explicit_rope_info`, `_extract_complex_rope_info`,
  `_convert_to_thop_cos_sin`, `_build_rope_info`, `_get_real_tensor`,
  `_compute_cos_sin_from_inv_freq`, `_find_inv_freq_tensor`) →
  new `transform/library/fuse_rope_into_trtllm_attention.py`. The new module
  reuses `_trace_to_buffer_source` / `_get_nested_attr` / `_trace_back_index`
  from `rope.py`. `library/__init__.py` already swallows `ImportError` /
  `ModuleNotFoundError` for transform modules in standalone mode, so the new
  file is silently skipped there.
- The two TRT-LLM-only tests `test_fuse_qkv_with_trtllm_cache_insertion` /
  `test_fuse_qkv_gqa_with_trtllm_cache_insertion` (and the `QKVAttentionModel`
  they rely on for that pipeline) → new
  `tests/.../transformations/library/test_gemm_fusion_trtllm.py`. The
  remaining QKV/SwiGLU fusion tests stay in `test_gemm_fusion.py`.
- `examples/auto_deploy/create_standalone_package.py`: add
  `test_gemm_fusion_trtllm.py` to `EXCLUDE_TEST_FILES`.

No behavior change in the full TRT-LLM build — same transforms register, same
tests run.

## Test plan

- [x] `pytest tests/unittest/auto_deploy/standalone` — 12/12 passed (was 11/12 with 22 sub-test failures)
- [x] `pytest tests/unittest/auto_deploy/singlegpu/transformations/library/test_rope_transformation.py` — 20 passed, 2 xfailed
- [x] `pytest tests/unittest/auto_deploy/singlegpu/transformations/library/test_gemm_fusion.py tests/unittest/auto_deploy/singlegpu/transformations/library/test_gemm_fusion_trtllm.py` — 26 passed, 8 skipped, 1 xfailed
- [x] CI to verify AutoDeploy stages: `/bot run --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"`

## Release Notes

* **Refactor**
  * Reorganized RoPE fusion transformation into a dedicated module for improved code organization.

* **Tests**
  * Added comprehensive tests validating QKV GEMM fusion combined with TRT-LLM cached attention insertion for both dense and grouped query attention patterns.

* **Chores**
  * Updated standalone package generation to exclude specialized test files from distribution.


lucaslie:
/bot run --stage-list "A10-Build_Docs, A10-PackageSanityCheck-PY310-UB2204, A100X-PackageSanityCheck-PY312-UB2404, A30-AutoDeploy-1, H100_PCIe-AutoDeploy-1, DGX_B200-AutoDeploy-1, A100X-PyTorch-1"

bmarimuthu-nv (approved):
LGTM, thanks!

govind-ramnarayan:
I'm trying to understand why only the test needs to be excluded and not the transform fuse_rope_into_trtllm_attention.py. Claude suggests that we have a guard for transforms here: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/transform/library/__init__.py

This makes sense. Is this the main such guard?

One issue that comes to mind is that if this standalone package is integrated with another inference framework, this skip could swallow errors in the development of that integration by silently dropping some transforms. Is that correct?

Maybe premature and excessive, but we could add some gating on the name of the transform? E.g. it should have "trtllm" in the title?

govind-ramnarayan (approved):

lucaslie:
/bot skip --comment "AD tests are passing"

lucaslie:
> I'm trying to understand why only the test needs to be excluded and not the transform fuse_rope_into_trtllm_attention.py. Claude suggests that we have a guard for transforms here: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/transform/library/__init__.py
> 
> This makes sense. Is this the main such guard?
> 
> One issue that comes to mind is that if this standalone package is integrated with another inference framework, this skip could swallow errors in the development of that integration by silently dropping some transforms. Is that correct?
> 
> Maybe premature and excessive, but we could add some gating on the name of the transform? E.g. it should have "trtllm" in the title?

I think that makes sense to be a little more rigorous long term. Definitely, right now the split is based on minimal changes, and I think as LLMC is maturing, we can be more rigorous about how we separate it.

PR #13486, opened 2026-04-27
[None][fix] visual_gen UlyssesAttention: pass post-A2A seq_len to inner backend

karljang:
## Summary

`UlyssesAttention` in visual_gen forwards a sharded `seq_len` kwarg into
the inner attention backend even after its all-to-all has gathered the
sequence dim back to full length. The inner backend uses `seq_len` for
2D reshapes (Sage path and the regular fused-QKV path in `trtllm.py`),
so the row count mismatches the actual tensor — causing either a
`ValueError` or a `q_hidden_size` assertion inside the FMHA kernel
runner whenever Ulysses sequence parallelism is on (world_size > 1).

This PR overrides `kwargs["seq_len"]` (and `seq_len_kv` in the unfused
path) with the post-A2A tensor shape inside `UlyssesAttention.forward`,
so the inner backend sees a length that matches what it actually
received.

## Root cause trace

1. Caller (`tensorrt_llm/_torch/visual_gen/modules/attention.py:_attn_impl`)
   reads `seq_len = q.shape[1]` from the **pre-A2A** (sharded) tensor —
   e.g. `4096 / 8 = 512` under 8-way Ulysses — and passes it to
   `self.attn.forward(..., seq_len=512, seq_len_kv=512)`.
2. `UlyssesAttention.forward` runs `all_to_all_4d/5d`: tensors come back
   with full seq_len (4096). `kwargs["seq_len"]` is still 512.
3. Inner backend (`tensorrt_llm/_torch/visual_gen/attention_backend/trtllm.py`):
   - Sage path (lines 276–278): `q.reshape(batch_size * seq_len, -1)`
   - Fused path (line 292): `qkv.reshape(batch_size * seq_len, -1)`
   Both use the stale 512 against a 4096-row tensor → kernel runner
   asserts on `q_hidden_size`.

## Fix

Two-line addition in each of `_forward_fused` and `_forward_unfused`,
captured *before* any HND transpose so the seq axis is taken from the
correct dimension.

## Correctness

- `world_size == 1`: A2A is skipped; `q.shape[1]` already equals the
  caller's `seq_len`. Override is a no-op.
- Self-attention (fused QKV): `seq_len_kv = seq_len` (Q and K share
  length).
- Cross-attention (unfused): `kv_seq_len_full = k.shape[1]` set
  separately, so different Q/KV lengths are preserved.
- HND backends: shape values are captured before the transpose at
  line 132–138.

## Test plan

- [x] FLUX.1-dev with multi-rank Ulysses + Sage attention — verified
      locally: without this PR, fails at `q_hidden_size` assertion in
      the FMHA kernel runner; with the PR applied, runs cleanly.
- [x] FLUX.1-dev / FLUX.2-dev single-rank baseline (no `--ulysses_size`) —
      bit-exact regression vs main expected (override is a no-op when
      `world_size == 1`).

## Bug Fixes
- Fixed assertion failures in parallel attention backend kernel execution by ensuring sequence lengths are correctly handled during distributed tensor operations.

### DIFF
--- a/tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
+++ b/tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
@@ -104,6 +104,11 @@ class UlyssesAttention(AttentionBackend):
 
         B, seq_len, _, Hp, D = qkv.shape
 
+        # Caller passed pre-A2A (sharded) seq_len; the inner backend
+        # reshapes by it, so hand it the post-A2A length instead.
+        kwargs["seq_len"] = seq_len
+        kwargs["seq_len_kv"] = seq_len
+
         output = self.inner_backend.forward(q=qkv, k=None, v=None, **kwargs)
 
         return self._output_a2a(output, batch_size, seq_len)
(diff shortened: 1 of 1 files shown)


karljang:
Hi @NVShreyas, 
Could you please review this? I had to make these changes to resolve an issue I encountered while sweeping the performance of a DiT model.

NVShreyas (approved):
Thanks for fixing, LGTM

PR #13503, opened 2026-04-27
[TRTLLM-8997][feat] Add encoder_max_batch_size & encoder_max_num_tokens and deterministic dummy sizing

yechank-nvidia:
## Summary

  Make TensorRT-LLM's multimodal **encoder workspace user-sizable and accurate**. Two new `TorchLlmArgs` knobs declare the encoder budget at LLM init; a shared `MultimodalEncoderMixin` lets the engine drive `AttentionMetadata` sizing for every vision encoder; a deterministic dummy-sizing API makes KV-cache
  profiling hit the declared budget exactly; and a per-mm-item fan-out hook converts the user's mm-item knob into the correct attention-sequence buffer size for encoders that split a single mm item into many attention sequences (LLaVA-Next sub-image patches, Qwen-VL attention windows).

  ## Why

  KV-cache profiling already runs a multimodal dummy through the encoder, so the encoder workspace **is** preallocated today — just not at the size the user actually wants:

  1. The dummy sizer used an `input_seq_len`-based heuristic that iteratively halved on failure. Production requests larger than the heuristic triggered `Attention workspace size is not enough` plus a `cudaMalloc` resize on the first real request.
  2. Every encoder hardcoded `max_num_requests=8192` and `max_num_tokens=model_config.max_num_tokens` (LLM-side units, not encoder pre-merger tokens) in its `__init__`. Even when the user knew their multimodal workload, there was no way to declare it.
  3. The `8192` floor is simultaneously **too small** (Qwen2/2.5-VL windowed attention exceeds it once `max_pixels` is raised — produces broadcast errors like *"size of tensor a (3) must match the size of tensor b (5) at non-singleton dimension 0"*) and **too large** (≈1 MB per `AttentionMetadata` wasted for
  Pixtral/RADIO/single-image CLIP usage).

  Fix is threefold:

  - Let the user declare the encoder budget.
  - Propagate that budget through engine → mixin → metadata in one place.
  - Make the dummy sizing deterministic so preallocation matches production exactly.

  ## What this PR adds

  ### Public API knobs

  `tensorrt_llm/llmapi/llm_args.py` · `tests/unittest/api_stability/references/llm.yaml`

  | Field | Default | Meaning |
  | --- | --- | --- |
  | `TorchLlmArgs.encoder_max_batch_size: Optional[PositiveInt]` | `64` | Maximum number of multimodal items per encoder forward. Drives the encoder's `AttentionMetadata.max_num_requests` together with the per-mm-item fan-out. Explicit `None` falls back to `max_batch_size` (legacy behaviour, overallocates with the LLM-side default of 2048). |
  | `TorchLlmArgs.encoder_max_num_tokens: Optional[PositiveInt]` | `None` | Encoder pre-merger attention-token budget. Falls back to `max_num_tokens`. Sizes `AttentionMetadata.max_num_tokens` **and** drives the deterministic dummy. |

  ### Encoder mixin

  `tensorrt_llm/_torch/models/modeling_multimodal_encoder.py` (new)

  - `MultimodalEncoderMixin.setup_attn_metadata(max_num_requests, max_num_tokens)` — default contract for engine-driven `AttentionMetadata` construction. Replaces per-encoder hardcoded `8192` in `__init__`.
  - Migrated encoders: `CLIPVisionModel`, `PixtralVisionModel`, `Qwen2_5_VisionModel`, `Qwen3VisionModel`, `VisionTransformer` (RADIO), `SiglipVisionModel`. RADIO overrides to set `kv_layout="NHD"` for the FlashInfer backend; Qwen2/2.5-VL overrides to construct a full + window metadata pair.

  ### Engine wiring

  `tensorrt_llm/_torch/pyexecutor/model_engine.py` · `tensorrt_llm/_torch/pyexecutor/_util.py`

  - `PyTorchModelEngine._set_up_multimodal_encoder_attn_metadata` — walks `self.model.modules()`, calls `setup_attn_metadata` on every `MultimodalEncoderMixin` with sizes derived from the user-declared encoder budget. Mirrors `_set_up_attn_metadata` for the LLM backbone.
  - `PyTorchModelEngine.is_multimodal` — single canonical check (model inherits `MultimodalModelMixin` or input processor inherits `BaseMultimodalInputProcessor`); replaces the ad-hoc `MODEL_CLASS_VISION_ENCODER_MAPPING` lookup scattered across paths.

  ### Per-mm-item fan-out

  `tensorrt_llm/inputs/registry.py` · `tensorrt_llm/_torch/models/modeling_llava_next.py` · `tensorrt_llm/_torch/models/modeling_qwen2vl.py`

  - `BaseMultimodalInputProcessor.get_max_requests_per_mm_item() -> int` — worst-case attention sequences produced by one mm item, derived purely from this processor's own configuration (mirrors vLLM's `max_tokens_per_mm_item` pattern, but in attention-sequence units). Default `1`.
  - `LlavaNextInputProcessor` override: `image_size_to_num_patches` worst-case over `image_grid_pinpoints` (default config returns `5`).
  - `Qwen2VLInputProcessorBase` override: `ceil(processor.max_pixels / patch_size² / (window_size/patch_size)²)` (default Qwen2.5-VL returns `80`, computed from `size["longest_edge"]` on transformers ≥ 5 or legacy `max_pixels`).
  - Engine multiplies `encoder_max_batch_size × get_max_requests_per_mm_item()` when sizing the encoder's `max_num_requests`. The *"user knob is mm items, prealloc is attention sequences"* unit conversion lives in one place.

  ### Deterministic dummy sizing

  `tensorrt_llm/inputs/registry.py` · `tensorrt_llm/_torch/pyexecutor/_util.py` · Qwen-VL family

  - `BaseMultimodalInputProcessor.get_num_mm_tokens(width, height, num_frames=1)` plus `spatial_merge_unit` — single source of truth for the encoder-side token math, in pre-merger units (same as `encoder_max_num_tokens` and `AttentionMetadata.max_num_tokens`). Existing `get_num_tokens_per_image` /
  `get_num_tokens_per_video` delegate here, falling back to the HF processor when the model has not opted in.
  - `BaseMultimodalDummyInputsBuilder.get_size_with_most_features(max_tokens)` — inverse of the above; picks the `(width, height, num_frames)` that lands just under `max_tokens` without iterative halving.
  - Qwen-VL family (`Qwen2VLInputProcessorBase`, `Qwen3VLInputProcessorBase`) implements both, mirroring HF's `smart_resize` math.
  - `KvCacheCreator._create_dummy_mm_context_request` consumes `encoder_max_num_tokens` and the deterministic API directly; iterative halving is removed.
 
  ## Non-goals / follow-ups

  - **Multi-encoder models** (vision + audio simultaneously, e.g. Nemotron Nano VL2). The mixin and engine wiring already support multiple `MultimodalEncoderMixin` modules on the same model, but the deterministic dummy currently sizes a single modality. Multi-modality dummy sizing follows in a separate PR.
  - **Scheduler enforcement of `encoder_max_num_tokens`.** Today the knob is a *prealloc hint*, not an admission gate. The `encoder_max_batch_size=64` default is a safety cap because the scheduler does not yet check the encoder token budget at admission (vLLM's `max_num_encoder_input_tokens` model). Once the
  scheduler enforces it, the fallback can switch to `min(max_batch_size, encoder_max_num_tokens // tokens_per_mm_item)` — mirroring vLLM's `encoder_budget // max_tokens_per_mm_item`. TODO comments mark the relevant call sites.


tensorrt_llm/_torch/models/modeling_multimodal_utils.py
2ez4bz: Nit: should this be in `tensorrt_llm/_torch/models/modeling_multimodal_encoder.py` instead?

2ez4bz: Nit: should `metadata_cls` be a `ClassVar`?

2ez4bz: Maybe stupid question: "philosophically", what is the difference between this and `prepare_attn_metadata`? It seems most (if not all?) existing multimodal models define `prepare_attn_metadata` - should we unify that and `setup_attn_metadata`?

moraxu (approved):
LGTM

tensorrt_llm/llmapi/llm_args.py
achartier: I think we need to update `tests/unittest/api_stability/references/llm.yaml` for new API fields

achartier (approved):

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -2864,6 +2877,20 @@ class BaseLlmArgs(StrictBaseModel):
...
+            "Maximum batch size for the multimodal encoder's AttentionMetadata. "
+            "Falls back to `max_batch_size` when unset."),
+        status="prototype")
+
+    encoder_max_num_tokens: Optional[int] = Field(
+        default=None,
+        description=(
+            "Maximum number of tokens for the multimodal encoder's "
+            "AttentionMetadata. Falls back to `max_num_tokens` when unset."),
+        status="prototype")
```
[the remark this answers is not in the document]
2ez4bz: The suggested fix to use `PositiveInt` as the default makes sense. But, doesn't hurt to change the line in `get_encoder_runtime_sizes` to explicitly check `is not None` as well. Pydantic allows you to check which fields are set / unset, as well.

tests/unittest/_torch/modeling/test_mm_encoder_mixin.py
2ez4bz, line 29: Nit: if this is supposed to be `ClassVar`, then it would be better to declare it as such even in `_DummyEncoder` for readability.

tests/unittest/llmapi/test_llm_args.py
2ez4bz, line 206: Nit: can we use `@pytest.mark.parametrize` for these tests? with e.g. `kwargs, expected_runtime_sizes`. The test logic for each test function below seems identical.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -687,17 +688,18 @@ def __init__(self, model_config: ModelConfig[PretrainedConfig]):
...
+
+    def setup_attn_metadata(self, max_num_requests: int,
+                            max_num_tokens: int) -> None:
+        # Override: Qwen2/2.5-VL uses two metadata objects (full + window
+        # attention) instead of the mixin's single `attn_metadata`.
+        kwargs = dict(max_num_requests=max_num_requests,
+                      max_num_tokens=max_num_tokens,
+                      kv_cache_manager=None)
+        self.full_attn_metadata = self.metadata_cls(**kwargs)
+        self.window_attn_metadata = self.metadata_cls(**kwargs)
```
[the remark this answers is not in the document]
yechank-nvidia: Engine-internal invariant: `PyTorchModelEngine._set_up_mm_encoder_attn_metadata()` is the sole production call site and it always runs after model load.

Adding a runtime guard would violate the codebase guidance (AGENTS.md) of not adding "validation for scenarios that can't happen". Tests that bypass the engine are easy to fix locally — the existing AttributeError already points at the correct line.

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
yechank-nvidia: From Claude: 

Intentionally not a `ClassVar`. The value is bound per-instance from model_config.attn_backend in every encoder's `__init__` (CLIP, Pixtral, SigLIP, Radio, Qwen2VL, Qwen3VL — all do `self.metadata_cls = get_attention_backend(...).Metadata)`, so two instances of the same encoder class can resolve to different Metadata types. 

Marking it `ClassVar` would (a) be semantically narrower than the actual binding pattern and (b) trip mypy/pyright on every existing self.metadata_cls = ... assignment. 

The bare Type[AttentionMetadata] annotation expresses the mixin contract — "subclasses must bind this before setup_attn_metadata() is called" — without prescribing class- vs. instance-level binding.

yechank-nvidia: Different lifecycle: 

`setup_attn_metadata` is called once after model load by the engine to allocate the AttentionMetadata with capacity sizes (max_num_requests, max_num_tokens). 

`prepare_attn_metadata` is called per-forward to populate per-batch fields (seq_lens, request_ids, prompt_lens) on the already-allocated instance. 

Mirrors the LLM backbone's _set_up_attn_metadata (one-shot) vs per-iteration prepare.

tests/unittest/_torch/modeling/test_mm_encoder_mixin.py
yechank-nvidia: From the above comment, metadata_cls is not `ClassVar`, so the conditional doesn't apply.

tests/unittest/llmapi/test_llm_args.py
yechank-nvidia: Changed. Thx.

tensorrt_llm/llmapi/llm_args.py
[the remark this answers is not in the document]
yechank-nvidia: Addressed.

tensorrt_llm/llmapi/llm_args.py
yechank-nvidia: Added. Thx.

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
yechank-nvidia: Moved. Thx.

tensorrt_llm/llmapi/llm_args.py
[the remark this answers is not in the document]
yechank-nvidia: @coderrabbitai I'm talking to you. I'm saying your suggestion has been addressed.

nvpohanh (approved):

juney-nvidia (approved):
Approved from LLM API perspective

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -591,14 +591,33 @@ def prepare_encoder_only(self) -> None:
...
             get_global_attrs().attention_metadata = weakref.ref(self)
 
+        # Encoder-only metadata never carries a KV cache. ``__post_init__``
+        # only initializes the block-offset / block-id attributes when a
+        # ``kv_cache_manager`` is present, and the full ``prepare()`` is what
+        # otherwise sets them to ``None`` (its no-cache branch). This fast path
+        # skips ``prepare()``, so mirror that no-cache setup here -- the FMHA
+        # ``_run`` reads ``kv_cache_block_offsets`` and ``block_ids_per_seq``
+        # unconditionally and would otherwise raise ``AttributeError``.
+        self.kv_cache_params = KVCacheParams(use_cache=False)
```
yuxianq, line 613: We add some `prepare` code back to `prepare_encoder_only`, can we directly update `prepare` and use `prepare` instead of a new `prepare_encoder_only`? We may need to sync more code between them in the future. If there is perf issue to use `prepare`, do we have some data?

.claude/skills/trtllm-model-onboard-multimodal/SKILL.md
```diff
@@ -304,7 +304,18 @@ class {Name}Model(PreTrainedModel):
 
 ### Phase 3 — Input processor + dummy builder
 
-Subclass **both** `BaseMultimodalInputProcessor` (drives every real request) and `BaseMultimodalDummyInputsBuilder` (drives engine warmup / profiling — the base shrinks dummy image resolution until the synthetic prompt fits `input_seq_len`). Colocate in the modeling file. Reference: `Qwen3VLInputProcessorBase`.
+Subclass **both** `BaseMultimodalInputProcessor` (drives every real request) and `BaseMultimodalDummyInputsBuilder` (drives engine warmup / KV-cache profiling). Colocate in the modeling file. References: `Qwen3VLInputProcessorBase` (image+video), `Mistral3InputProcessor` (Pixtral).
+
+**Encoder KV-cache memory profiling (deterministic dummy sizing).** The KV-cache profiler sizes the encoder's memory contribution by running the encoder **once** on a worst-case dummy (held resident through the peak measurement), **decoupled** from the text-only LLM dummy. This replaced the old iterative resolution-halving dummy prompt. To opt in, the model exposes `encode_multimodal_inputs` (via `MultimodalModelMixin`) and the input processor implements the modality-agnostic dummy contract on `BaseMultimodalDummyInputsBuilder`:
```
2ez4bz: Nit: I'm not sure the "This replaced the old iterative resolution-halving dummy prompt" is adding much.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -591,14 +591,33 @@ def prepare_encoder_only(self) -> None:
         if extra_attrs is None:
             get_global_attrs().attention_metadata = weakref.ref(self)
 
+        # Encoder-only metadata never carries a KV cache. ``__post_init__``
```
2ez4bz: Nit: get rid of double backticks on changed lines?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -591,14 +591,33 @@ def prepare_encoder_only(self) -> None:
...
+        # only initializes the block-offset / block-id attributes when a
+        # ``kv_cache_manager`` is present, and the full ``prepare()`` is what
+        # otherwise sets them to ``None`` (its no-cache branch). This fast path
+        # skips ``prepare()``, so mirror that no-cache setup here -- the FMHA
+        # ``_run`` reads ``kv_cache_block_offsets`` and ``block_ids_per_seq``
```
2ez4bz: Nit: I don't think the technical details of why these were put here are necessary - it actually seems like it was more of an omission before that we _added_ public attributes such as `kv_cache_block_offsets` in `__post_init__` instead of declaring them as part of the dataclass.

tensorrt_llm/_torch/models/modeling_clip.py
```diff
@@ -16,20 +16,9 @@
 from ..modules.attention import Attention
 from ..modules.mlp import MLP
 from .hf_parameter_utils import get_parameter_device, get_parameter_dtype
+from .modeling_multimodal_encoder import MultimodalEncoderMixin
 from .modeling_utils import _load_weights_impl, register_auto_model
 
-try:
-    # Available in transformers<5
```
2ez4bz, line 22: Nice!

tensorrt_llm/_torch/models/modeling_mistral.py
```diff
@@ -463,6 +463,112 @@ def call_with_text_prompt(
...
+    # (``get_num_tokens_per_image``) must keep using the processor's LLM-side
+    # token count (Pixtral grid + ``[IMG_BREAK]``/``[IMG_END]`` framing), while
+    # encoder profiling needs the ViT attention-sequence length (pre-merge patch
+    # count). Different units, so the dummy path uses the private ``_vit_*``
+    # helpers below instead of ``get_num_mm_tokens``.
+    # ------------------------------------------------------------------
+    def _vision_geometry(self) -> Tuple[int, int, int, int]:
+        """``(patch_size, spatial_merge_size, num_channels, max_image_size)``,
+        read from the processor (mistral_common) or the HF ``vision_config``."""
+        vcfg = getattr(self.config, "vision_config", None)
```
2ez4bz: Can we check if any of these attributes are guaranteed to exist? I find that these AI tools' love of `getattr(x, "foo", None)` makes code less readable.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
```diff
@@ -109,7 +114,16 @@ def _prepare_qwen_vl_vision_attn_metadata(
     attn_metadata.cu_q_seqlens = cu_seqlens
     attn_metadata.cu_kv_seqlens = cu_seqlens
     attn_metadata.max_seq_len = max(seq_lens)
-    attn_metadata.prepare()
+    # The vision tower runs no-cache, context-only attention and supplies its
+    # own ``cu_seqlens`` above, so the heavy KV-oriented ``prepare()`` (kv_lens
+    # / prompt_lens / host_request_types setup) is unnecessary host work. Use
+    # the lean encoder-only path when the backend exposes it; fall back to the
+    # full ``prepare()`` for backends that do not (e.g. non-TRTLLM metadata).
+    prepare_encoder_only = getattr(attn_metadata, "prepare_encoder_only", None)
```
2ez4bz: Nit: couldn't we pass an explicit argument like `backend` instead? Seems more natural to me.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -703,20 +709,40 @@ def __init__(self, model_config: ModelConfig[PretrainedConfig]):
...
-        # Pre-allocated `arange` for the vision block's
-        # `rope_position_ids`; per-call code just slices `[:seq_len]`
-        # instead of allocating a fresh `(seq_len,) int32` + H->D copy.
-        # TODO: Make capacity dynamic with the encoder's `max_num_tokens`.
-        self.register_buffer(
-            "_rope_position_ids_buffer",
-            torch.arange(32768, dtype=torch.int32, device="cuda"),
-            persistent=False,
-        )
+        # Pre-allocated `arange` for the vision block's `rope_position_ids`;
```
2ez4bz: Nit: this is no longer initialized as `arange`?

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -837,10 +863,12 @@ def prepare_attn_metadata(
...
     ):
         if attn_metadata is None:
-            attn_metadata = self.metadata_cls(
-                max_num_requests=8192,  # TODO: Make this dynamic
-                max_num_tokens=8192,  # TODO: Make this dynamic
-                kv_cache_manager=None,
+            raise RuntimeError(
+                "Vision encoder AttentionMetadata is not initialized. "
+                "`setup_attn_metadata` (engine-driven via "
+                "`_set_up_multimodal_encoder_attn_metadata`, or called "
```
2ez4bz: Nit: I don't think we need to / should mention private helpers in these potentially user-facing error messages.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -1035,7 +1066,23 @@ def forward(self, multimodal_params: List[MultimodalParams]) -> List[torch.Tenso
         return embeds
 
 
-class Qwen3VLModelBase(PreTrainedModel):
+class Qwen3VLModelBase(PreTrainedModel, MultimodalModelMixin):
+    def encode_multimodal_inputs(
```
2ez4bz, line 1067: Just checking: was this enough to support things like deepstack features?

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -538,6 +460,75 @@ def _create_dummy_context_requests(
...
+
+        Returns an empty list when the model is not a multimodal-encoder model
+        or the processor has no dummy builder (then the encoder is not profiled
+        directly). Whether the *processor* has opted into deterministic dummy
+        sizing is detected below via ``NotImplementedError`` / empty demand — a
+        model with the encoder entry but no dummy builder just yields an empty
+        batch (no encoder profiling) until the builder is implemented.
+        """
+        # ``MultimodalModelMixin`` is the uniform ``encode_multimodal_inputs``
+        # entry; imported locally to avoid a pyexecutor -> models import cycle.
```
2ez4bz: Do we actually import things from `pyexecutor` in the models code? I just checked, it doesn't seem like there are any pre-existing direct imports from `pyexecutor` under `tensorrt_llm/_torch/models`, so if there are any, they would likely be indirect.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -538,6 +460,75 @@ def _create_dummy_context_requests(
...
+        from tensorrt_llm._torch.models.modeling_multimodal_mixin import \
+            MultimodalModelMixin
+        if not isinstance(self._model_engine.model, MultimodalModelMixin):
```
2ez4bz, line 483: Out of curiosity, is there a hard interface requirement on `MultimodalModelMixin` for the code below to work? At first glance, it seems to mostly be gated on whether the input processor class implements things like `get_mm_max_tokens_per_item` / `get_dummy_mm_data_for_tokens`, but I could be missing something.

Might be good to leave a comment if it's something else downstream that assumes the we are dealing with a `MultimodalModelMixin`?

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -699,6 +702,11 @@ def configure_kv_cache_capacity(self,
                 torch_peak_memory = torch.cuda.memory_stats(
                 )["allocated_bytes.all.peak"]
 
+                # Free the held encoder embeddings now that the peak (which they
+                # contributed to) has been recorded, so the steady-state
+                # measurement below doesn't count this transient dummy.
+                del encoder_profile_output
```
2ez4bz, line 713: Just checking: should `self._dummy_encoder_inputs` be reset as well?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1827,6 +1837,43 @@ def _set_up_attn_metadata(
 
         return self.attn_metadata
 
+    @property
+    def is_multimodal(self) -> bool:
+        """True iff this engine drives a multimodal model.
+
+        Primary signal: :class:`MultimodalModelMixin` (PR #13866) is the
```
2ez4bz: Nit: mentions of the actual PR is unnecessary.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1827,6 +1837,43 @@ def _set_up_attn_metadata(
...
+        canonical marker — multimodal LM classes inherit from it. Until
+        every model has migrated (Mistral done; Qwen-VL, Nemotron, Gemma,
+        Phi-4-MM, etc. pending), fall back to whether the input processor
+        subclasses :class:`BaseMultimodalInputProcessor`, which every
```
2ez4bz: Nit: I don't think our docs support `:class:` style references.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1827,6 +1837,43 @@ def _set_up_attn_metadata(
...
+        multimodal model necessarily provides at the data boundary.
+
+        TODO: Once all multimodal models inherit ``MultimodalModelMixin``
```
2ez4bz: Could you create a ticket and reference it here? 🙏

tensorrt_llm/inputs/registry.py
```diff
@@ -488,6 +485,52 @@ def get_num_multimodal_tokens(self):
...
+    # raises NotImplementedError and the hashing path falls back to the HF
+    # processor delegation it already used.
+    # ------------------------------------------------------------------
+    @property
+    def spatial_merge_unit(self) -> int:
+        """Encoder→LLM token ratio.
+
+        Default 1 (no spatial merging). Override on encoders that downsample
+        their output before the LLM (e.g. Qwen-VL family returns
+        ``spatial_merge_size ** 2``).
```
2ez4bz: Out of curiosity, do models use "unit" or "size" more often?

tensorrt_llm/inputs/registry.py
```diff
@@ -488,6 +485,52 @@ def get_num_multimodal_tokens(self):
...
+        """
+        return 1
+
+    def get_num_mm_tokens(
```
2ez4bz: I think this is very confusing, given that we have `get_num_tokens_per_image`, which ends up trying to call this, and `get_num_mm_tokens` accepts arguments that are very vision-specific.

Would it make more sense to move this to the innards of `get_num_tokens_per_image` / `get_num_tokens_per_video` for models that need it (e.g. Qwen) ?

tensorrt_llm/inputs/registry.py
```diff
@@ -561,19 +630,31 @@ def get_num_tokens_per_video(
...
+    ``get_num_tokens_per_image`` / ``..._video`` / ``..._audio``); a concrete
+    multimodal processor inherits from both mixins.
     """
-    Base class for generating dummy inputs. Specially for profiling
-    """
-
-    DEFAULT_IMAGE_MAX_DIM = 16384
-    DEFAULT_IMAGE_MIN_DIM = 128
 
     def __init__(self, **kwargs):
```
2ez4bz: Nit: looks like this method can now be entirely removed?

tensorrt_llm/inputs/registry.py
```diff
@@ -590,51 +671,50 @@ def config(self) -> PretrainedConfig:
...
+        """Build the worst-case dummy ``multimodal_data`` per modality budget.
+
+        The modality-agnostic entry the KV-cache encoder profiler calls, sizing
+        each modality to saturate its share of the token budget.
+
+        ``max_tokens_per_modality`` maps each modality from
+        :meth:`get_mm_max_tokens_per_item` to its share of the (shared)
+        ``encoder_max_num_tokens`` budget. The implementation owns every
+        modality-specific choice: a vision processor inverts the budget to a
+        media size and materializes ``pixel_values`` / ``grid_thw``; an audio
```
2ez4bz: Nit: I personally think the lines starting with this one, till the end of the paragraph, could be removed (they don't add that much more clarity to what this method is used for, at least for me). The paragraph after was clearer for me.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -3550,6 +3550,20 @@ class BaseLlmArgs(StrictBaseModel):
     max_num_tokens: Optional[int] = Field(
         default=8192, description="The maximum number of tokens.")
 
+    encoder_max_batch_size: Optional[PositiveInt] = Field(
```
2ez4bz: Do you think it would make sense to add these to the new config I'm defining [here](https://github.com/NVIDIA/TensorRT-LLM/pull/14829/changes#diff-bf140a489175e4c6e5931cd19cc57e06c1634483517f06c4e6fda96d5d26b056R519)? Also, might be worth making it explicit in the descriptions here whether the `max_num_tokens` / `max_batch_size` are per-modality, or whether they're shared proportionately across all modalities?

tests/unittest/_torch/modeling/test_modeling_mistral.py
```diff
@@ -629,3 +631,82 @@ def test_mistral_attention_swa_layer_types():
...
+    patches = (size["width"] // 14) * (size["height"] // 14)
+    assert patches <= budget
+    # Adding one more aligned step would exceed the budget (saturated).
+    nxt = size["width"] + unit
+    assert (nxt // 14) ** 2 > budget or nxt > 1540
+
+
+def test_dummy_get_size_rejects_non_positive_budget():
+    proc = _make_dummy_processor()
+    with pytest.raises(ValueError):
```
2ez4bz: Nit: can we make the test a bit more specific with a `match=r"..."` type argument?

tests/unittest/_torch/modeling/test_modeling_qwen2_5vl.py
```diff
@@ -424,3 +428,216 @@ def test_prepare_qwen_vl_mrope_config_pure_generation_reads_deltas_only():
...
+            "shortest_edge": min_pixels,
+            "longest_edge": max_pixels
+        }))
+    return instance
+
+
+_DUMMY_PROCESSORS = [Qwen2VLInputProcessorBase, Qwen3VLInputProcessorBase]
+
+
+@pytest.mark.parametrize("processor_cls", _DUMMY_PROCESSORS)
```
2ez4bz, line 479: Nit: can we also parameterize over `spatial_merge_size, expected_spatial_merge_unit`?

tests/unittest/inputs/test_multimodal_input_processor.py
```diff
@@ -0,0 +1,188 @@
...
+        @property
+        def spatial_merge_unit(self):
+            return spatial_merge_unit_value
+
+        if deterministic:
+
+            def get_num_mm_tokens(self, *, width, height, num_frames=1):
+                return width * height * num_frames
+
+    instance = _Sub.__new__(_Sub)
```
2ez4bz: Out of curiosity, why can't we directly instantiate `_Sub` here?

tests/unittest/inputs/test_multimodal_input_processor.py
```diff
@@ -0,0 +1,188 @@
...
+    frames = [Image.new("RGB", (8, 6)) for _ in range(3)]
+    # Each frame width*height=48, *3 frames=144 encoder tokens, //4 = 36
+    assert sub.get_num_tokens_per_video(video=frames) == 36
+    sub._processor._get_num_multimodal_tokens.assert_not_called()
+
+
+def test_get_num_tokens_per_video_falls_back_to_hf_processor():
+    sub = _make_subclass(deterministic=False, hf_post_merger_tokens=777)
+    frames = [Image.new("RGB", (8, 6)) for _ in range(3)]
+    assert sub.get_num_tokens_per_video(video=frames) == 777
```
2ez4bz: Nit: for values like these that get reused multiple times, can we assign them to a variable? The same thing happens in e.g. `test_get_num_tokens_per_image_falls_back_to_hf_processor`.

tests/unittest/llmapi/test_llm_args.py
```diff
@@ -189,6 +189,61 @@ def test_llm_args_with_model_kwargs(self, llm_args_cls):
         assert llm_args.model_kwargs['num_hidden_layers'] == 2
 
 
+@pytest.mark.parametrize("llm_args_cls", [TrtLlmArgs, TorchLlmArgs])
```
2ez4bz: Do we even want to support these 2 new args in `TrtLlmArgs` / should they be only for `TorchLlmArgs`.

tensorrt_llm/_torch/attention_backend/trtllm.py
yechank-nvidia: Hi @yuxianq, thanks for the comment.

I didn't add `prepare_encode_only`, but when I tested e2e with `prepare()` vs. `prepare_encode_only()`, I got faster speed on `prepare_encode_only()`.

Here is the microbenchmark.

```
  ┌───────┬──────┬───────────┬────────────────────────┬─────────┬───────┐
  │ batch │ seq  │ prepare() │ prepare_encoder_only() │  delta  │ ratio │
  ├───────┼──────┼───────────┼────────────────────────┼─────────┼───────┤
  │ 1     │ 1024 │ 47.2µs    │ 10.3µs                 │ +36.9µs │ 4.6x  │
  ├───────┼──────┼───────────┼────────────────────────┼─────────┼───────┤
  │ 1     │ 4096 │ 46.8µs    │ 10.4µs                 │ +36.4µs │ 4.5x  │
  ├───────┼──────┼───────────┼────────────────────────┼─────────┼───────┤
  │ 8     │ 1024 │ 48.2µs    │ 11.0µs                 │ +37.3µs │ 4.4x  │
  ├───────┼──────┼───────────┼────────────────────────┼─────────┼───────┤
  │ 32    │ 256  │ 47.7µs    │ 10.8µs                 │ +36.9µs │ 4.4x  │
  ├───────┼──────┼───────────┼────────────────────────┼─────────┼───────┤
  │ 64    │ 256  │ 47.6µs    │ 10.9µs                 │ +36.7µs │ 4.4x  │
  └───────┴──────┴───────────┴────────────────────────┴─────────┴───────┘
```

yechank-nvidia: I think I can make the little helper and share the code between `prepare()` and `prepare_encode_only()`, WDYT?

.claude/skills/trtllm-model-onboard-multimodal/SKILL.md
yechank-nvidia: Removed.

tensorrt_llm/_torch/attention_backend/trtllm.py
yechank-nvidia: Removed.

tensorrt_llm/_torch/attention_backend/trtllm.py
yechank-nvidia: Addressed.

tensorrt_llm/_torch/models/modeling_mistral.py
yechank-nvidia: Thx. Removed

tensorrt_llm/_torch/models/modeling_qwen3vl.py
yechank-nvidia: Removed.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
yechank-nvidia: So far, didn't encounter any issue yet.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
yechank-nvidia: No, we still need it. the buffer initialization is moved to `setup_attn_metadata`

tensorrt_llm/_torch/pyexecutor/_util.py
yechank-nvidia: Good catch — the dummy-data sizing is indeed gated on the input processor's `get_mm_max_tokens_per_item` & `get_dummy_mm_data_for_tokens`. 

The hard dependency on `MultimodalModelMixin` is `_encode_dummy_inputs`, which calls `model.encode_multimodal_inputs` (the mixin contract). 

The `isinstance` gate guards that call, and also intentionally scopes direct encoder profiling to mixin-migrated pilots (Qwen2-VL, Mistral); future models opt in by inheriting the mixin + implementing the processor dummy hooks. Added a comment to make this explicit.

tensorrt_llm/_torch/pyexecutor/_util.py
yechank-nvidia: I think this will be called just once, so resetting will be not necessary.

tensorrt_llm/_torch/pyexecutor/model_engine.py
yechank-nvidia: Removed.

tensorrt_llm/_torch/pyexecutor/_util.py
yechank-nvidia: You are correct on this. Thx. Removed.

tensorrt_llm/_torch/pyexecutor/model_engine.py
yechank-nvidia: Removed.

tensorrt_llm/inputs/registry.py
yechank-nvidia: Removed.

tensorrt_llm/inputs/registry.py
yechank-nvidia: At that time when I first added, I encountered some silent MRO bug, I prefer to keep it here.

tensorrt_llm/inputs/registry.py
yechank-nvidia: From Claude:

size is far more common (161 vs 18 hits), but it's a different quantity — spatial_merge_size is the linear factor (e.g. 2), while spatial_merge_unit is spatial_merge_size ** 2 (the encoder→LLM token ratio, e.g. 4).

tensorrt_llm/llmapi/llm_args.py
yechank-nvidia: > Do you think it would make sense to add these to the new config I'm defining here? 

Yes, I think unifying all multimodal-related thing into one big config, would be the final goal.

> Also, might be worth making it explicit in the descriptions here whether the max_num_tokens / max_batch_size are per-modality, or whether they're shared proportionately across all modalities?

I think for this PR, I meant shared proportionately across all modalities. Let me add that explanation. Later, we could have per modality knobs.

tensorrt_llm/inputs/registry.py
yechank-nvidia: Totally makes sense. Moved to model-specific logics.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
2ez4bz: Should we move the comment closer to that those lines?

tensorrt_llm/_torch/models/modeling_qwen3vl.py
2ez4bz: Nice!

tensorrt_llm/_torch/pyexecutor/_util.py
2ez4bz: Right, but won't we be holding on to more (host?) memory than necessary?

tensorrt_llm/inputs/registry.py
2ez4bz: Ah, gotcha.

tensorrt_llm/inputs/registry.py
2ez4bz: Hm... I have high confidence this is no longer necessary - does our test suite not catch that bug you're referring to?

tensorrt_llm/llmapi/llm_args.py
2ez4bz: Sounds good.

tensorrt_llm/inputs/registry.py
yechank-nvidia: Yes, I confirmed that currently this is not the issue anymore. Removed.

tensorrt_llm/_torch/models/modeling_qwen2vl.py
yechank-nvidia: Went with making prepare_encoder_only a base AttentionMetadata method defaulting to `prepare()` (TRTLLM overrides it with the lean path), so the call sites just call it directly — no getattr probe or backend branching needed.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: Thanks for the perf data, a little shared helper is enough, fix the potential perf issue of `prepare` is out of scope.

tensorrt_llm/_torch/pyexecutor/_util.py
yechank-nvidia: You are correct. Removed.

tensorrt_llm/_torch/pyexecutor/model_engine.py
yechank-nvidia: Done.

tests/unittest/inputs/test_multimodal_input_processor.py
yechank-nvidia: Done.

tests/unittest/_torch/modeling/test_modeling_mistral.py
yechank-nvidia: Done.

tests/unittest/_torch/modeling/test_modeling_qwen2_5vl.py
yechank-nvidia: Done.

tests/unittest/inputs/test_multimodal_input_processor.py
yechank-nvidia: Good question — we actually can instantiate it directly. Switched to call directly.

tests/unittest/llmapi/test_llm_args.py
yechank-nvidia: Moved to `TorchLlmArgs` only

tensorrt_llm/_torch/models/modeling_qwen3vl.py
yechank-nvidia: Done.

tensorrt_llm/_torch/attention_backend/trtllm.py
yechank-nvidia: Done

2ez4bz:
first try

Issue #218, opened 2023-10-31
building tensorRT-llm, TensorRt issue


arjunanand13:
Out of all tensorrt-9.1.0.post12.dev4 only one wheel is downloading in colab , that is "TensorRT-9.1.0.4/python/tensorrt-9.1.0.post12.dev4-cp310-none-linux_x86_64.whl". But after downloading ,I'm facing this error while importing it
"ImportError                               Traceback (most recent call last)
[<ipython-input-5-8dfd087091e3>](https://localhost:8080/#) in <cell line: 1>()
----> 1 import tensorrt

[/usr/local/lib/python3.10/dist-packages/tensorrt/__init__.py](https://localhost:8080/#) in <module>
     69 del _libs_wheel_imported
     70 
---> 71 from .tensorrt import *
     72 
     73 __version__ = "9.1.0.post12.dev4"

ImportError: libnvinfer.so.9: cannot open shared object file: No such file or directory"

byshiue:
Have you export the new TensorRT libs path into `LD_LIBRARY_PATH`?

arjunanand13:
Yeah like this
!export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/content/TensorRT-9.1.0.4/lib

arjunanand13:
Is their any sample documentation code

juney-nvidia:
@arjunanand13  Hi, it is suggested to build TensorRT-LLM by following the installation guide with using the docker based workflow. Have you tried with the docker based build flow before?

June

zmsunnyday:
Hi, I got the same issue when using vscode, but it worked well in shell without vscode. So weird.

byshiue:
It might the environment setting of using vscode.

nv-guomingz:
do u still have further issue or question now? If not, we'll close it soon.

PR #13531, opened 2026-04-27
[TRTLLM-11851][feat] Add MX-only P2P checkpoint loading support for TRTLLM

chienchunhung:
## Release Notes

* **New Features**
  * Added ModelX (MX) integration for faster checkpoint loading via peer-to-peer weight transfer, with automatic fallback to standard disk-based loading when MX is unavailable or not configured.
  * New configuration parameters: `mx_server_url` for MX server endpoint and `mx_preshard_strategy` to control weight sharding behavior.

### Summary

This PR is the **MX-only first slice** split out from PR #13045.

It adds `checkpoint_format="MX"` support to TRT-LLM's PyTorch backend using upstream `modelexpress.trtllm_live_transfer.MxLiveWeightLoader` and `publish_model_params`. GMS is intentionally excluded so reviewers can validate MX first.

Follow-up PRs will add:

1. GMS-only support (`LoadFormat.GMS`, `GMSBackend`, GMS args/tests)
2. MX+GMS validation/composition
3. Packaging extras once MX/GMS dependency onboarding is complete

#### MX Checkpoint Loader

Adds `MXCheckpointLoader` under `tensorrt_llm/_torch/models/checkpoints/mx/`.

Behavior:

- Subclasses `HfCheckpointLoader`, so HF disk fallback is inherited.
- Calls upstream `MxLiveWeightLoader(mx_server=url).load_weights(checkpoint_dir, mapping=..., model=...)`.
- Calls upstream `publish_model_params(model)` before `post_load_weights()` for source workers.
- Exposes `p2p_succeeded` so `ModelLoader` can skip normal weight mapping on full P2P success.
- Supports mixed P2P/disk fallback: if upstream returns `fallback_weights`, keep P2P-delivered tensors and merge only the returned fallback tensors through the standard disk pipeline.
- Avoids republishing from P2P receiver workers to prevent duplicate MX metadata / NIXL registrations.

#### MX Config

Adds MX-only prototype fields:

- `mx_server_url`
- `mx_server_query_timeout_s`
- `mx_preshard_strategy`

Behavior:

- `MODEL_EXPRESS_URL` is used as fallback for `mx_server_url` when `checkpoint_format="MX"`.
- `mx_server_query_timeout_s` lets deployments size source-discovery wait time.
- If unset, TRT-LLM probes MX first:
  - no registered source: use `MX_SOURCE_QUERY_TIMEOUT=30` for fast disk fallback
  - registered source exists: defer to upstream/default wait so targets can wait for long source disk-loads
- `mx_preshard_strategy="global"` fails fast until `LoadFormat.PRESHARDED` exists upstream.

#### ModelLoader Integration

Updates only the existing `LoadFormat.AUTO` path:

- Passes `model=model` to checkpoint loaders. Generic HF loaders ignore it; MX uses it for direct P2P writes.
- Always initializes `self.weight_mapper`, including the MX fast path, so `reload()` remains safe.
- Marks non-draft `Linear` modules as `_weights_presharded=True` after MX success.
- Skips normal weight mapping on full P2P success.
- Runs returned `fallback_weights` through the standard weight-loading path for partial MX fallback.
- Publishes source weights only when this worker did **not** receive via P2P.

#### Linear Marker

Adds `_weights_presharded = False` to `Linear`.

This PR does not route presharded tensors back through `load_weight_shard()` helpers; the marker is set after successful MX direct writes and is kept for the current ModelLoader path plus future `LoadFormat.PRESHARDED` work.

### What This PR Excludes

This PR intentionally does **not** include:

- `LoadFormat.GMS`
- `GMSBackend`
- `gms_socket_path`, `gms_mode`, `gms_tag`
- GMS RW/RO loading
- GMS tests
- `[gms]` / `[dynamo]` packaging extras

### Packaging Note

This PR does not add a `[mx]` extra yet.

For prototype testing:

```bash
pip install "modelexpress>=0.3.0,<0.4.0"
```

`modelexpress` is on PyPI but still needs NVIDIA OSS allowlist onboarding (tracked as MX-7). Once complete, restoring `pip install tensorrt_llm[mx]` is a small `setup.py` change.

### Running MX

```yaml
# config_mx.yaml
checkpoint_format: "MX"
mx_server_url: "http://mx-server:8001"
```

```bash
trtllm-serve <model> --config config_mx.yaml
```

Optional timeout override:

```yaml
checkpoint_format: "MX"
mx_server_url: "http://mx-server:8001"
mx_server_query_timeout_s: 1800
```

Python API:

```python
from tensorrt_llm import LLM

llm = LLM(
    model="<model>",
    checkpoint_format="MX",
    mx_server_url="http://mx-server:8001",
)
```

### Test Coverage

Added MX-only unit tests:
- `tests/unittest/llmapi/test_mx_args.py`
- `tests/unittest/_torch/models/checkpoints/mx/test_mx_checkpoint_loader.py`
- `tests/unittest/_torch/pyexecutor/test_model_loader_mx.py`

Coverage includes:
- MX config defaults and validation
- `MODEL_EXPRESS_URL` fallback
- `mx_server_query_timeout_s`
- `mx_preshard_strategy` validation
- MX loader registry and construction
- disk fallback paths
- full P2P success path
- partial fallback merge path
- source publish env restoration
- `MODEL_NAME` resolution
- model-loader fast path: mapper init, reload safety, draft exclusion, publish skip on P2P receive

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,392 @@
...
+
+# Defensive default for the upstream ``MX_SOURCE_QUERY_TIMEOUT`` env var.
+# The upstream ``MxLiveWeightLoader`` polls the MX server every 5 s for up
+# to ``MX_SOURCE_QUERY_TIMEOUT`` seconds (default 3600 = 1 hour) waiting
+# for a source. On a cold cluster (no donor up yet), this means the very
+# first replica blocks for an hour before falling back to disk. We cap
+# the default at 30 s so first-replica startup degrades gracefully; users
+# can still override via the env var or a future per-loader knob.
+# Tracked as MX-4 in §15 (non-blocking source-query API upstream).
+_MX_SOURCE_QUERY_TIMEOUT_DEFAULT_S = "30"
```
KavinKrishnan, line 56: The 30-second cap is correct for first-replica / source instances (don't poll forever for a non-existent source), but it's wrong for target replicas that legitimately need to wait for an in-flight source disk-load. We've measured Kimi K2.5 take ~22 minutes and DeepSeek-V3.2-NVFP4 take ~20 minutes to disk-load on the source. A target started during that window will hit 30s, fall through to disk, and the deployment loses the entire MX benefit.

Suggestion: do a fast probe via MxClient.list_sources() first to distinguish the two modes, or expose this as TorchLlmArgs.mx_server_query_timeout so users can size it per-deployment. The 30s setdefault should only apply when no sources exist after the probe.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,392 @@
...
+        except Exception:
+            # Deliberately broad: MX is an opportunistic fast path and HF
+            # disk loading remains the correctness path. Preserve the full
+            # traceback so unexpected upstream failures are diagnosable.
+            logger.warning(
+                f"MX P2P transfer failed; falling back to disk loading.\n{traceback.format_exc()}"
+            )
+            return self._fallback_to_disk(checkpoint_dir, mapping, **kwargs)
+
+        if fallback_weights:
```
KavinKrishnan, line 237: Falling through to a full disk load when even one tensor returned in fallback_weights discards a successful 90+ GB RDMA transfer. On large models with thousands of tensors a single mismatch (e.g., embedding table vocab differs between checkpoint revisions) costs the entire MX win.

The upstream MxLiveWeightLoader already handles size mismatches per-tensor (RDMAs the matched ones, returns only the unmatched). 

Suggestion: keep p2p_succeeded = True and pass the fallback_weights dict back to model.load_weights() so just those few tensors go through the standard pipeline. The Linear-level _weights_presharded marking can stay all-or-nothing for this MX-only slice (sufficient for the 100% match case we've seen with Kimi K2.5 and DSV3.2), with per-module marking arriving alongside LoadFormat.PRESHARDED.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,392 @@
...
+                checkpoint_dir,
+            )
+        else:
+            logger.info("MX P2P unavailable; loading from disk: %s", checkpoint_dir)
+        return super().load_weights(checkpoint_dir, mapping=mapping, **kwargs)
+
+    def publish_as_source(
+        self,
+        model,
+        mapping: Optional[Mapping] = None,  # noqa: ARG002
```
KavinKrishnan: The mapping parameter is silently ignored. If a future upstream publish_model_params signature accepts mapping for parallelism-aware identity, this argument should be plumbed through; if not, the argument should be removed. As-is, callers (model_loader.py:484) pass mapping=self.mapping thinking it has effect.

Suggestion: either add # TODO(MX-2): plumb mapping into publish_model_params when public API lands or drop the argument entirely.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,392 @@
...
+            return
+
+        try:
+            from modelexpress.trtllm_live_transfer import (  # type: ignore[import-not-found]
+                publish_model_params,
+            )
+        except ImportError:
+            logger.debug("modelexpress library not installed; skipping MX publish.")
+            return
+
```
KavinKrishnan, line 354: The os.environ set/restore for MODEL_EXPRESS_URL and MODEL_NAME is process-global mutation. Fine for the MPI-per-rank execution model where each rank is a separate Python process, but a Python LLM API caller that uses threading or multiprocessing inside a single interpreter would race here.

Suggestion: turn the existing # THREADSAFETY: comment (line 297) into an explicit assertion or warning when threading.active_count() > 1, so accidental misuse fails loud.

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -382,17 +396,53 @@ def init_meta_tensor(t: torch.Tensor):
...
+                            module._weights_presharded = True
+                    logger.info("MX P2P: marked main model Linear modules as "
+                                "presharded (strategy=per_module), skipping "
+                                "weight mapping pipeline.")
+
+                if not mx_p2p_succeeded:
+                    self.weight_mapper = checkpoint_loader.get_initialized_weight_mapper(
+                        model, config)
+                    self._call_load_weights(model.load_weights, weights,
+                                            self.weight_mapper)
```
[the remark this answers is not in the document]
KavinKrishnan: +1 to coderabbit comment

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -428,6 +478,14 @@ def init_meta_tensor(t: torch.Tensor):
                 raise NotImplementedError(
                     f"No load support for load format: {load_format}")
 
+            # MX source publish: expose raw weights to other ranks
+            # BEFORE post_load_weights() transforms them, so targets
+            # receive the pre-transform state.
+            if (load_format == LoadFormat.AUTO
```
KavinKrishnan: The condition is if load_format == LoadFormat.AUTO and hasattr(checkpoint_loader, 'publish_as_source'). There's no guard against publishing when mx_p2p_succeeded is True (i.e., we received via P2P, so we shouldn't republish — that would just re-register the same NIXL endpoints under a new ID).

Suggestion: skip the publish when getattr(checkpoint_loader, 'p2p_succeeded', False) is True, similar to the prior not_p2p_succeeded guard in our PR #218 patch script. Without this, second-generation replicas that received via P2P will also publish, leading to N copies of effectively the same data in Redis.

tensorrt_llm/_torch/pyexecutor/model_loader.py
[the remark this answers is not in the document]
chienchunhung: Good catch; updated the code. `self.weight_mapper = checkpoint_loader.get_initialized_weight_mapper(model, config)` is now hoisted before the `mx_p2p_succeeded` branch, so both the MX fast path and fallback path initialize it.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
chienchunhung: Good point; updated the timeout handling to distinguish those two cases.

The loader now does a best-effort fast probe before setting `MX_SOURCE_QUERY_TIMEOUT`:
- If a source is already registered, we do **not** stamp the 30s cap and let upstream `MxLiveWeightLoader` use the existing env/user/default timeout, so targets can wait for long donor disk-loads.
- If no source is registered, we temporarily apply the 30s cap for fast first-replica disk fallback.
- Added `TorchLlmArgs.mx_server_query_timeout_s` and plumbed it into `MXCheckpointLoader` so deployments can explicitly size the wait window.

The timeout override is now scoped to the actual `load_weights()` call via a temporary env context rather than being set in the constructor.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
chienchunhung: Agreed; updated the mixed-success path.

When upstream `MxLiveWeightLoader` returns non-empty `fallback_weights`, `MXCheckpointLoader` now keeps `p2p_succeeded = True` and returns that dict instead of falling back to a full HF disk load. `ModelLoader` then runs the returned fallback weights through the standard `_call_load_weights(model.load_weights, fallback_weights, self.weight_mapper)` path, so only those tensors are loaded through the disk pipeline while P2P-delivered tensors remain in place.

Also updated the warning to report the fallback tensor count and byte size, and added tests for this partial-fallback flow.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
chienchunhung: Agreed; removed the unused `mapping` parameter from `publish_as_source` and stopped passing `mapping=self.mapping` at the call site.

Today upstream `publish_model_params(model)` does not accept mapping and derives rank from `MPI.COMM_WORLD` internally, so keeping a no-op parameter was misleading. The MX-2 follow-up still tracks replacing the env-var handoff with a public upstream API once one exists.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
chienchunhung: Good point. Updated this from a comment-only warning into runtime protection.

`publish_as_source()` now wraps the `MODEL_EXPRESS_URL` / `MODEL_NAME` set-restore window in a module-level lock, and it emits a one-time warning when `threading.active_count() > 1`.

This does not make process-global env vars a perfect long-term API — a concurrent non-MX reader could still observe transient values — but it prevents writer-writer races in our code and makes threaded misuse visible. The real fix remains MX-2: replacing the env-var handoff with a public upstream publish/identity API that accepts explicit arguments.

tensorrt_llm/_torch/pyexecutor/model_loader.py
chienchunhung: Agreed; fixed.

The publish guard now skips `publish_as_source` when `checkpoint_loader.p2p_succeeded` is true. A worker that received weights via P2P already points at a real source, so republishing would create duplicate MX metadata and redundant NIXL registrations for the same bytes.

Added unit coverage in `test_model_loader_mx.py`: MX success path asserts `publish_as_source` is not called.

KavinKrishnan:
PR LGTM - lack approval privileges

tburt-nv (approved):
No problem with the setup.py comments

tensorrt_llm/_torch/models/checkpoints/hf/weight_loader.py
```diff
@@ -59,8 +60,9 @@ def _get_local_available_host_memory() -> int:
                                               op=_MPI.MIN)
         return available_host_memory
 
-    def load_weights(self, checkpoint_dir: str,
-                     mapping: Mapping) -> dict[str, Any]:
+    def load_weights(self, checkpoint_dir: str, mapping: Mapping,
+                     **kwargs) -> dict[str, Any]:
+        del kwargs  # Optional format-specific args (e.g. model=) are ignored.
```
2ez4bz: This is a claude-ism that is very confusing. `kwargs` are always passed as a local copy, so these `del` statements don't do anything. Please remove.

chienchunhung: This is just a common trick to silence the unused-argument lint. But I agree this can be confusing and it does not seem like our ruff-legacy lint complains about it. Removed the line as suggested.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -3782,6 +3759,42 @@ class TorchLlmArgs(BaseLlmArgs):
...
+        "is skipped. Options: "
+        "'per_module' — set ``_weights_presharded=True`` on each Linear "
+        "module after a successful MX P2P transfer (current prototype "
+        "behavior). "
+        "'global' — set ``LoadFormat.PRESHARDED`` for the whole model "
+        "(planned, not yet wired upstream — selecting it raises "
+        "at config construction). "
+        "Only used when checkpoint_format='MX'.",
+        status="prototype",
+    )
```
pcastonguay: Could we group those in a `ModelExpressConfig` class?

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -228,9 +228,14 @@ def get_guided_decoding_config(guided_decoding_backend: str,
...
-    checkpoint_loader = _construct_checkpoint_loader(llm_args.backend,
-                                                     llm_args.checkpoint_loader,
-                                                     llm_args.checkpoint_format)
+    checkpoint_loader = _construct_checkpoint_loader(
+        llm_args.backend,
+        llm_args.checkpoint_loader,
+        llm_args.checkpoint_format,
+        mx_server_url=llm_args.mx_server_url,
+        mx_model_name=llm_args.model,
+        mx_server_query_timeout_s=llm_args.mx_server_query_timeout_s,
```
pcastonguay: Once `ModelExpressConfig` class is created, just pass the config `mx_config = llm_args.mx_config`.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,456 @@
...
+# Defensive default for the upstream ``MX_SOURCE_QUERY_TIMEOUT`` env var.
+# The upstream ``MxLiveWeightLoader`` polls the MX server every 5 s for up
+# to ``MX_SOURCE_QUERY_TIMEOUT`` seconds (default 3600 = 1 hour) waiting
+# for a source. On a cold cluster (no donor up yet), this means the very
+# first replica blocks for an hour before falling back to disk. We cap
+# the default at 30 s so first-replica startup degrades gracefully; users
+# can still override via the env var or a future per-loader knob.
+# Tracked as MX-4 in §15 (non-blocking source-query API upstream).
+_MX_SOURCE_QUERY_TIMEOUT_DEFAULT_S = "30"
```
pcastonguay, line 56: Could we make this a user-configurable parameter, part of the (to-be-added) `ModelExpressConfig` class?

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -428,6 +491,18 @@ def init_meta_tensor(t: torch.Tensor):
...
```
pcastonguay: From Claude:

  After a successful MX P2P main-model load with speculative decoding enabled, the draft-weight load_weights call resets _p2p_succeeded to False, causing publish_as_source to fire on the source even though the  
  main model arrived via P2P. The flag tracks "did the last call use P2P", not "did the main model use P2P".
                                                                                                                                                                                                                   
  Repro flow:                                     
  1. Main load → _p2p_succeeded = True (line 252 of checkpoint_loader.py)                                                                                                                                          
  2. Draft load (line 462 of model_loader.py) calls MXCheckpointLoader.load_weights without model= → first line of override sets self._p2p_succeeded = False (line 179) → falls back to disk (line 182).           
  3. Publish gate (line 499) reads not p2p_succeeded → True → publishes the main-model bytes a second time, contradicting the comment on line 500-502 ("Republishing it would create duplicate MX metadata and NIXL
   registrations for the same bytes.")                                                                                                                                                                             
                                                                                                                                                                                                                   
  Current:         
```                                                                                                                                                                                                
      self._p2p_succeeded = False                                                                                                                                                                                  
      if self._mx_server_url is None or model is None:                                                                                                                                                             
          return self._fallback_to_disk(...)
```
  Fix:                                                                                                                                                                                                             
      Either (a) snapshot mx_p2p_succeeded once in ModelLoader.load after the main load and pass it explicitly into the publish gate (preferred — avoids stateful order dependence), or (b) make _p2p_succeeded    
  "sticky" in MXCheckpointLoader.load_weights by only resetting it when the load is actually a P2P attempt:                         
```                                                                               
          # only reset when we are about to attempt P2P, not on fallback paths                                                                                                                                     
          if self._mx_server_url is None or model is None:                                                                                                                                                         
              return self._fallback_to_disk(...)  
          self._p2p_succeeded = False              
```                                                                                                                                                                
      Note that (b) still leaves the flag reflecting "the last P2P-eligible call", which is also wrong — option (a) is the durable fix.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,456 @@
...
+
+        self._p2p_succeeded = True
+        logger.info(
+            "MX P2P weight transfer succeeded from %s",
+            self._mx_server_url,
+        )
+        return {}
+
+    def _resolve_query_timeout_override(
+        self, checkpoint_dir: str, MxClient, build_identity
```
pcastonguay: Add type hints for `MxClient` and `build_identity`. 
```
      def _resolve_query_timeout_override(                                                                                                                                                                         
          self, checkpoint_dir: str, MxClient: Any, build_identity: Callable[..., Any]
      ) -> Optional[str]:
```

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,456 @@
...
+                reason,
+                checkpoint_dir,
+            )
+        else:
+            logger.info("MX P2P unavailable; loading from disk: %s", checkpoint_dir)
+        return super().load_weights(checkpoint_dir, mapping=mapping, **kwargs)
+
+    def publish_as_source(
+        self,
+        model,
```
pcastonguay, line 322: Add type hint.
```suggestion
        model: nn.Module
```

tensorrt_llm/_torch/modules/linear.py
```diff
@@ -2599,6 +2578,11 @@ def __init__(
         self.use_cute_dsl_blockscaling_mm = use_cute_dsl_blockscaling_mm
         self.disable_deep_gemm = disable_deep_gemm
         self.fused_weight_shard_indices_mapping = fused_weight_shard_indices_mapping
+        # MX P2P can write already-sharded local weights directly into
+        # parameter buffers. The marker is set by ModelLoader after a
+        # successful MX transfer; this first MX-only slice does not route it
+        # back through the standard load_weight_shard helpers.
+        self._weights_presharded = False
```
pcastonguay: It doesn't look like this is used anywhere. If not needed for this PR, please remove and add in future PR that needs it.

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -382,17 +399,63 @@ def init_meta_tensor(t: torch.Tensor):
...
                 self.weight_mapper = checkpoint_loader.get_initialized_weight_mapper(
                     model, config)
-                self._call_load_weights(model.load_weights, weights,
-                                        self.weight_mapper)
+
+                if mx_p2p_succeeded:
+                    preshard_strategy = getattr(self.llm_args,
+                                                'mx_preshard_strategy',
+                                                'per_module')
+                    if preshard_strategy != 'per_module':
```
pcastonguay: Don't we already check that in `llm_args.py`? I would drop the runtime check until we need  support other `preshard_strategy` values.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -4042,6 +4042,38 @@ def validate_checkpoint_format(self):
...
         return self
 
+    @model_validator(mode="after")
+    def validate_mx_config(self) -> 'TorchLlmArgs':
+        # When MX is the active checkpoint format and the user did not
+        # explicitly set ``mx_server_url``, honor the ``MODEL_EXPRESS_URL``
+        # env var that the upstream ``modelexpress`` library reads. This
+        # lets orchestrators configure MX via the environment while keeping
+        # the resolved value visible on ``llm_args.mx_server_url``.
+        if (self.checkpoint_format == "MX" and self.mx_server_url is None):
```
pcastonguay: nit 
```suggestion
        if self.checkpoint_format == "MX" and self.mx_server_url is None:
```

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -382,17 +399,63 @@ def init_meta_tensor(t: torch.Tensor):
...
+                # A non-empty dict contains size-mismatched tensors that
+                # should be merged via the standard disk pipeline.
+                mx_p2p_succeeded = (hasattr(checkpoint_loader, 'p2p_succeeded')
+                                    and checkpoint_loader.p2p_succeeded)
                 self.weight_mapper = checkpoint_loader.get_initialized_weight_mapper(
                     model, config)
-                self._call_load_weights(model.load_weights, weights,
-                                        self.weight_mapper)
+
+                if mx_p2p_succeeded:
```
pcastonguay: We're adding MX specific code in the generic `ModelLoader` class , which makes it harder to read. 

Maybe we could add methods to `BaseCheckpointLoader`:
```
          def post_load_apply(self, model: nn.Module) -> None: ...    # MX overrides → mark presharded                                                                                                             
          def post_load_publish(self, model: nn.Module, checkpoint_dir: str) -> None: ...  # MX overrides → publish_as_source 
```
and call them here.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,456 @@
...
+        query_timeout_s: Optional[int] = None,
+    ):
+        super().__init__(
+            weight_loader=weight_loader,
+            weight_mapper=weight_mapper,
+            config_loader=config_loader,
+        )
+        # Align the backing attribute with the property override so any
+        # caller reading self._checkpoint_format directly also sees "MX".
+        self._checkpoint_format = "MX"
```
pcastonguay, line 117: Do we need both `_checkpoint_format` and:
```
   @property
    def checkpoint_format(self) -> str:
```
?

tensorrt_llm/_torch/models/checkpoints/hf/weight_loader.py
```diff
@@ -59,8 +60,8 @@ def _get_local_available_host_memory() -> int:
                                               op=_MPI.MIN)
         return available_host_memory
 
-    def load_weights(self, checkpoint_dir: str,
-                     mapping: Mapping) -> dict[str, Any]:
+    def load_weights(self, checkpoint_dir: str, mapping: Mapping,
+                     **kwargs) -> dict[str, Any]:
```
brb-nv, line 68: Are these supposed to be passed down but missed?

brb-nv: Ok, I think you implemented this for MX overload.

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -382,17 +399,63 @@ def init_meta_tensor(t: torch.Tensor):
...
+                    preshard_strategy = getattr(self.llm_args,
+                                                'mx_preshard_strategy',
+                                                'per_module')
+                    if preshard_strategy != 'per_module':
+                        raise NotImplementedError(
+                            f"mx_preshard_strategy='{preshard_strategy}' is not "
+                            "supported in this MX-only PR. Use "
+                            "mx_preshard_strategy='per_module' instead.")
+
+                    from tensorrt_llm._torch.modules.linear import Linear
```
brb-nv: Is this lazy import intentional?

brb-nv (approved):
Minor comments. Changes LGTM.

tensorrt_llm/_torch/pyexecutor/model_loader.py
chienchunhung: This is removed now. The lazy `Linear` import was only needed for the `_weights_presharded` marking path, which we dropped from this MX-only slice because the marker is not consumed by an exercised runtime path in this PR.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
chienchunhung: I kept both intentionally.

`MXCheckpointLoader` subclasses `HfCheckpointLoader`, whose constructor initializes the backing `_checkpoint_format` as `"HF"`. The property override makes `loader.checkpoint_format` return `"MX"`, but keeping `_checkpoint_format = "MX"` avoids an inconsistent object state for any legacy/internal call site that still reads the backing attribute directly.

I tightened the comment to explain this inheritance detail explicitly.

tensorrt_llm/_torch/pyexecutor/model_loader.py
chienchunhung: Good suggestions. Moved the format-specific post-load behavior behind `BaseCheckpointLoader` hooks.

tensorrt_llm/_torch/models/checkpoints/hf/weight_loader.py
chienchunhung: Right.

venkywonka (commented):
This PR seem to have no documentation update on this awesome new feature addition. If that is aimed at a follow-up PR then no worries, but if not, I'd recommend adding some docs:

Here are some places codex suggested:
-  `docs/source/features/checkpoint-loading.md`
- Add a dedicated `features/model-express-p2p-
  checkpoint-loading.md`, wire it into docs/source/index.rst
- Add small pointers/examples in `overview.md`, `trtllm-serve.rst`,
  and optionally `quickstart_advanced.py`

venkywonka:
also if you desire that this be tracked in telemetry as a feature usage, you might also want to update `_collect_features()` in `tensorrt_llm/usage/usage_lib.py`

chienchunhung:
> also if you desire that this be tracked in telemetry as a feature usage, you might also want to update `_collect_features()` in `tensorrt_llm/usage/usage_lib.py`

Good suggestion! Added the entries for both "checkpoint_format" and "load_format".

chienchunhung:
> This PR seem to have no documentation update on this awesome new feature addition. If that is aimed at a follow-up PR then no worries, but if not, I'd recommend adding some docs:
> 
> Here are some places codex suggested:
> 
> * `docs/source/features/checkpoint-loading.md`
> * Add a dedicated `features/model-express-p2p- checkpoint-loading.md`, wire it into docs/source/index.rst
> * Add small pointers/examples in `overview.md`, `trtllm-serve.rst`,
>   and optionally `quickstart_advanced.py`

Yes we plan to add the doc once the integration is more mature (for both MX and GMS).

pcastonguay (approved):

tensorrt_llm/_torch/models/checkpoints/base_weight_loader.py
```diff
@@ -91,15 +94,16 @@ def mark_consumed(self, prefix: str) -> int:
...
+    def load_weights(self, checkpoint_dir: str, mapping: Mapping,
+                     **kwargs) -> Union[Dict[str, Any], ConsumableWeightsDict]:
         """
         Loads weights from a checkpoint directory.
 
         Args:
             checkpoint_dir: A path to the checkpoint directory.
             mapping: A mapping object containing the distributed configuration.
+            **kwargs: Optional format-specific loader arguments. Generic HF
+                loaders ignore these; MX uses ``model`` for direct P2P writes.
```
mikeiovine: Probably do not need these details in the base class docstring

tensorrt_llm/_torch/modules/linear.py
mikeiovine: Where is this marker set? I don't see it anywhere

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,469 @@
...
+                logger.warning(
+                    f"MX P2P transfer failed; falling back to disk loading.\n{traceback.format_exc()}"
+                )
+                return self._fallback_to_disk(checkpoint_dir, mapping, **kwargs)
+
+        if fallback_weights:
+            fallback_bytes = sum(
+                tensor.numel() * tensor.element_size()
+                for tensor in fallback_weights.values()
+                if hasattr(tensor, "numel") and hasattr(tensor, "element_size")
```
mikeiovine: Do we need these `hasattr` calls? Shouldn't all tensors have these methods?

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,469 @@
...
+
+    def _has_any_source_instance(self, checkpoint_dir: str, MxClient, build_identity) -> bool:
+        """Best-effort fast probe for registered MX source instances."""
+        client = None
+        try:
+            identity = build_identity(model_name=self._resolve_publish_name(checkpoint_dir))
+            client = MxClient(server_url=self._mx_server_url)
+            list_resp = client.list_sources(identity=identity)
+            return bool(getattr(list_resp, "instances", []))
+        except Exception:
```
mikeiovine: Can we use a more specific exception type?

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
```diff
@@ -0,0 +1,469 @@
...
+
+    def _fallback_to_disk(
+        self, checkpoint_dir: str, mapping: Mapping, *, reason: Optional[str] = None, **kwargs
+    ) -> dict[str, Any]:
+        """Standard HF disk loading fallback."""
+        if reason is not None:
+            logger.info(
+                "MX P2P unavailable (%s); loading from disk: %s",
+                reason,
+                checkpoint_dir,
```
mikeiovine: nit: coding guidelines recommend using f-strings over `%s` (and I think f strings are used elsewhere in this file)

mikeiovine (approved):

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
chienchunhung: Updated.

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
chienchunhung: Updated to `except (AttributeError, RuntimeError, TimeoutError, grpc.RpcError):`

tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py
chienchunhung: Removed. `fallback_weights` is expected to contain tensors, so the defensive `hasattr` checks were unnecessary.

tensorrt_llm/_torch/modules/linear.py
chienchunhung: Removed `_weights_presharded` for this PR. This was originally added to support per module tracking under partial MX scenarios. Planned to add this back in follow-up PRs.

tensorrt_llm/_torch/models/checkpoints/base_weight_loader.py
chienchunhung: I removed the MX/HF-specific detail from the base class docstring and kept it generic.

venkywonka (approved):

juney-nvidia (approved):
Approved from API perspective.

chienchunhung:
/bot skip --comment "CI succeeded earlier; skip another CI run after rebasing with tiny conflict resolution."

chienchunhung:
/bot skip --comment "CI succeeded earlier; skip another CI run after rebasing with tiny conflict resolution."

pcastonguay:
/bot skip --comment "CI succeeded earlier; skip another CI run after rebasing with tiny conflict resolution."

chienchunhung:
/bot skip --comment "CI succeeded earlier; skip another CI run after rebasing with tiny conflict resolution."

PR #13542, opened 2026-04-28
[None][chore] Convert cubins in repository to compressed archives

tongyuantongyu:
## Release Notes

* **New Features**
  * Implemented tarball-based cubin binary archiving (`.cubin.tar.zst`) with direct embedded linking

* **Chores**
  * Updated kernel build pipelines to use new cubin archive distribution format
  * Migrated cubin embedding from C++ hex-arrays to direct binary embedding with namespace-scoped symbol linking

## Description

Convert cubin sources from C array to compressed raw binaries. Drastically reduced LFS stored file size of them (3.7G -> 155MB).

Also unified the `EXCLUDE_SM` macro meaning across the whole codebase.

## Test Coverage

No functional change. Exist tests verify nothing is broken.

## PR Checklist

## GitHub Bot Help


cpp/tensorrt_llm/kernels/decoderMaskedMultiheadAttention/decoderXQAImplPrecompiled.cpp
```diff
@@ -356,7 +356,7 @@ class XQAKernelList
     TKernelMeta const* mKernelMeta;
     unsigned int mKernelMetaCount;
     unsigned int mSM;
-    std::unordered_map<unsigned long long const*, CUlibrary> mCuLibs;
```
pengbowang-nv, line 359: Did you re-export all the cubins or did you just compressed them? I'm asking this because XQA pre-compiled kernels are built from an ancient version and by exporting using current branch I doubt if it is going to work. Can we keep XQA for now as it only contains ~100 kernels and will be removed later? Thanks!

cpp/CMakeLists.txt
```diff
@@ -26,6 +26,22 @@ include(sanitizers)
 
 project(tensorrt_llm LANGUAGES CXX)
 
+# Single source of truth for the inline ABI namespace. Read by both the C++ side
```
pengbowang-nv, line 29: This might be a collateral change brought by changing XQA, we can remove these it we just ignore XQA/DecoderMaskedAttention kernels for now.

pengbowang-nv (commented):
I think we can reduce a lot of changes if we ignore xqa for now

cpp/cmake/modules/tllm_cubin_archive.cmake
```diff
@@ -0,0 +1,242 @@
...
+    # downstream .o rebuild. POSIX `touch -r` is universal on Linux, which is
+    # the only platform cubinIncbin.h supports anyway.
+    add_custom_command(
+      OUTPUT "${_EXTRACTED}"
+      COMMAND ${CMAKE_COMMAND} -E tar xf "${_ARCHIVE_PATH}"
+      COMMAND touch -r "${_ARCHIVE_PATH}" "${_EXTRACTED}"
+      WORKING_DIRECTORY "${EXTRACT_DIR}"
+      DEPENDS "${_ARCHIVE_PATH}"
+      COMMENT "Extracting cubin: ${_CUBIN_FILE}"
+      VERBATIM)
```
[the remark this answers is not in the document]
tongyuantongyu: Windows support is dormant now for TensorRT LLM.

cpp/CMakeLists.txt
tongyuantongyu: Actually, this is required to determine the mangled symbol name for all cubins, not limited to XQA.

cpp/tensorrt_llm/kernels/decoderMaskedMultiheadAttention/decoderXQAImplPrecompiled.cpp
tongyuantongyu: In this PR, all cubins are extracted from the cubin.cpp files and compressed. I just verified the fmha_v2 and xqa scripts can generate new cubins but didn't ship the updated cubins.

cpp/tensorrt_llm/kernels/decoderMaskedMultiheadAttention/CMakeLists.txt
```diff
@@ -88,3 +87,11 @@ target_link_libraries(decoder_attention_src PUBLIC nvrtc_wrapper_src
...
 set_property(TARGET decoder_attention_src PROPERTY CUDA_RESOLVE_DEVICE_SYMBOLS
                                                    ON)
+
+# Embed cubins via INCBIN. Replaces the bin2c-style hex-array .cpp stubs that
+# previously lived under cubin/ and bloated the source tree by ~166 MB.
+include(tllm_cubin_archive)
+tllm_add_cubin_archive_sources(
+  decoder_attention_src "${CMAKE_CURRENT_SOURCE_DIR}/cubin"
+  ARCHS 80 86 89 90 120
+  IMPLICIT_FAMILY)
```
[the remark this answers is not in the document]
tongyuantongyu: We still need `xqa_kernel_cubin.cpp` to provide the content of:

```
extern XQAKernelMetaInfo const sXqaKernelMetaInfo[];
extern size_t const sXqaKernelMetaInfoSize;
```

juney-nvidia (approved):
Approved from OSS compliance perspective

cpp/kernels/fmha_v2/setup.py
```diff
@@ -3359,8 +3355,21 @@ def get_cubin_header(kernel_traits, specs_names):
                 kspec.enable_skip_softmax, mask_type):
             continue
         name = fname.replace('.', '_')
```
pengbowang-nv, line 3357: NIT: Although no cubin is affected in this PR, FMHA_v2 may generate names like `cubin/..._sm100.no_i2f_f2i.cu.cubin`, as this PR is now generating .cubin.cpp at compile time and use filename for symbol name, we may get a cpp symbol like `cubin_fmha_v2_fp16_2048_128_sm100.no_i2f_f2i_cu_cubin`

cpp/kernels/fmha_v2/setup.py
```diff
@@ -3583,12 +3592,17 @@ def get_lname_from_kname(kname: str) -> str:
         else:
             assert False, 'Something terrible happened'
 
+    def render_sm(sm):
```
pengbowang-nv, line 3595: By Codex: This maps regular SM100/SM120 FMHA metadata to the family guards as well. That breaks CUDA 12.7/12.8 builds: `build_wheel.py` defaults `CMAKE_CUDA_ARCHITECTURES` to `all`, and `setup_cuda_architectures()` includes 100/120 for CUDA >= 12.7, but only enables family targets for CUDA >= 12.9. So in that configuration `EXCLUDE_SM_100` is not defined, but `EXCLUDE_SM_100F` is defined, the generated SM100 metadata/declarations/launchers are compiled out even though SM100 is enabled.

My understanding for above comment is: we don't support family target (100f/120f) prior to CUDA 12.8, and we can have `CMAKE_CUDA_ARCHITECTURES_ORIG = 80;86;89;90;100;120` but `CMAKE_CUDA_ARCHITECTURES_FAMILIES = (nothing)`. Here we are gating more than we need (as originally here use sm100 not sm100f)

cpp/tensorrt_llm/kernels/decoderMaskedMultiheadAttention/decoderXQAImplPrecompiled.cpp
pengbowang-nv: I just want to confirm my understanding: this PR changed the behavior of cubin generation from `source -> cubin -> cubin.cpp (all before compile)` to `source -> cubin -> cubin.tar.zst (before compile) -> cubin.cpp (during compile)`. In the process of ` cubin.tar.zst -> cubin.cpp` this PR generated function symbol names from .cubin.tar.zst name. And this PR also changed how XQA kernels are stored (previously by array of unsigned long long and now array of unsigned char), and that is why the change here is needed?

cpp/kernels/xqa/gen_cubins.py
```diff
@@ -406,10 +402,21 @@ def generate_header_file_contents(
...
+        # `tensorrt_llm::TRTLLM_ABI_NAMESPACE::kernels` namespace, avoiding
+        # multi-definition collisions with other packages or other TRT-LLM
+        # ABI versions in the same final link.
+        # Match the definition site emitted by TLLM_INCBIN_NS in
+        # cpp/include/tensorrt_llm/common/cubinIncbin.h:
+        #   extern unsigned char const SYMBOL[];
+        #   extern unsigned int  const SYMBOL_len;
         cubin_data_array.append(
-            f"extern unsigned long long {cubin_variable_name}[];\n")
+            f'extern unsigned char const {cubin_variable_name}[];\n')
```
pengbowang-nv, line 417: If we are now using unsigned char to interpret cubin, and old cubin are generated using unsigned long long, is the layout consistent?

Note that this would not be a problem even if we are wrong here, as XQA cubins are not actually used now😀.

tongyuantongyu: XQA used `unsigned long long` merely to reduce the overhead of cubin:

With `unsigned char`, each byte is 6 characters `0x00, ` so a 6x bloat;
with `unsigned long long` each 8 bytes is 23 characters `0x0000'0000'0000'0000ULL, ` so a ~3x bloat.

I verified old and new methods produced the same byte sequence in the binary.

cpp/kernels/fmha_v2/setup.py
tongyuantongyu: Fixed.

Kernels targeting the whole family should always use the `F` suffixed macros to avoid conflict (these definitions are global). I add rules to define these macros even when the CUDA compiler does not support family conditionals (only 12.8 has public released version)

cpp/kernels/fmha_v2/setup.py
tongyuantongyu: Seems like `.no_i2f_f2i` is the only exception. I simply changed it to `_no_i2f_f2i`.

cpp/tensorrt_llm/kernels/decoderMaskedMultiheadAttention/decoderXQAImplPrecompiled.cpp
tongyuantongyu: Conceptually yes.

Practically, we are using the `.incbin` assembler directive and inline assembly to avoid the C array in `.cubin.cpp` altogether. The assembler directly convert the cubin binary to objects.

`unsigned long long` or `unsigned char` does not matter as long as the same bytes are reproduced: `cuLibraryLoadData(CUlibrary* library, void const* code, ...)` accepts `void*` pointer so both works.

cpp/kernels/fmha_v2/setup.py
pengbowang-nv: I see the fix is [this line](https://github.com/NVIDIA/TensorRT-LLM/pull/13542/changes/4deeb4c367206ff7dcc4fccb1cf1b05bfa9c8453..2adf2d52c56a659a498f82f666352e999c644460#diff-724efe7b1e20be7a1025ab4bfe1791231c22257311966910c36917b933c76ddfR459) , but before 12.8 the previous condition is always true, so exclude_sm_100f is still defined in this case?

tongyuantongyu: CUDA supports SM100/120 starting from 12.8, so it's correct behavior to always exclude them for earlier version (by defining the macro).

pengbowang-nv: I tested on cuda 12.8 and it will define `EXCLUDE_SM_100F` even if I set `-a "100-real"`, although due to various dependency problem I didn't finish a full build, I managed to print some varibles by adding `message(STATUS "FMHA defs: ${TRTLLM_GEN_FMHA_DEFS}")` in `cpp/tensorrt_llm/kernels/trtllmGenKernels/fmha/CMakeLists.txt`.

```
-- GPU architectures: 100
-- CUDA family target support: version=12.8.93; has_families=OFF; min_family=9999
-- Excluding SM 80
-- Excluding SM 86
-- Excluding SM 89
-- Excluding SM 90
-- Excluding SM 103
-- Excluding SM 120
-- CUDA architectures normalized: 100a-real
-- CUDA architecture families:

....

-- FMHA defs: EXCLUDE_SM_103;EXCLUDE_SM_100F
```

tongyuantongyu: Thanks for the careful verification. This flaw should be fixed now.

PerkzZheng (approved):

pengbowang-nv (approved):

PR #13544, opened 2026-04-28
[https://nvbugs/6029882][fix] Clamp tokens_info writes in computeSeqAndPaddingOffsets

bobboli:
## Summary

- Bound each `tokensInfo` write in `computeSeqAndPaddingOffsets` to the `num_tokens`-sized allocation, so the last block cannot scribble past the buffer when the packed `seqQLengths` sum exceeds `numTokens` under multi-token generation paths.
- The `std::max(numTokens, seqEnd)` extension is preserved for the legitimate `sum < numTokens` case (filling trailing pad slots); the per-write bound now keeps `sum > numTokens` from causing OOB.
- Unwaive `accuracy/test_llm_api_pytorch.py::TestDeepSeekR1::test_nvfp4_multi_gpus[throughput_mtp]`.

## Why

NVBug 6029882. Primary fault is a CUDA Warp MMU exception in `computeSeqAndPaddingOffsets<__nv_bfloat16, 256>`, with secondary `moeA2ADispatchKernel` SIGTRAPs on peer ranks waiting on dispatch metadata that never arrived. Core-dump triage attributed the fault to the last-block `std::max(numTokens, seqEnd)` branch flagged by the original `// FIXME(Eagle)` comment.

## Test plan

- [x] Built `RelWithDebInfo` with `-lineinfo` locally.
- [x] Ran the unwaived test on 8x B200 with `CUDA_ENABLE_COREDUMP_ON_EXCEPTION=1`. Pass: MMLU 86.96 / GSM8K 95.41 / CnnDailymail 27.89, no core dumps.
- [ ] CI reruns covering `GB200-8_GPUs-2_Nodes-PyTorch-2` (the original failing stage).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced memory safety in tensor computation kernels with improved bounds checking to prevent out-of-range memory access during token processing operations.
  * Corrected token indexing logic to ensure all computations remain within valid ranges throughout execution.

* **Tests**
  * Re-enabled previously skipped multi-GPU performance throughput test case for improved optimization verification and overall test coverage.


cpp/tensorrt_llm/kernels/gptKernels.cu
```diff
@@ -223,25 +223,32 @@ __global__ __launch_bounds__(THREADS_PER_BLOCK) void computeSeqAndPaddingOffsets
...
+        // (sized to numTokens on the caller side); the bound keeps the kernel safe even if
+        // the caller's seqQLengths sum exceeds numTokens under multi-token generation paths.
         for (int tokenIdx = threadIdx.x; tokenIdx < seqLength; tokenIdx += blockDim.x)
         {
-            params.tokensInfo[seqBegin + tokenIdx] = make_int2(batchIdx, tokenIdx);
+            int const idx = seqBegin + tokenIdx;
+            if (idx < params.numTokens)
+            {
+                params.tokensInfo[idx] = make_int2(batchIdx, tokenIdx);
+            }
```
JadoTu: Will it bring silent errors since there is no process when it falls into the else case.
In other word, the root cause of idx running out of numTokens is not fixed.

bobboli: Fair point — agreed the silent skip traded a hard fault for a worse failure mode (downstream attention reads stale `(batchIdx, tokenIdxInSeq)` and produces wrong outputs without crashing). Pushed `878fdc21eb` which converts the bound check from a silent skip to `printf(...) + __trap()` for both `tokensInfo` and `paddingOffsets`/`encoderPaddingOffsets`. The kernel still cannot scribble past the `[numTokens]`-sized allocation, but if the upstream `sum(seqQLengths) > numTokens` invariant ever breaks the run aborts with `(blockIdx, batchSize, capacity, seqBegin, seqEnd)` printed instead of producing silent garbage.

I traced both call paths on Blackwell looking for the actual root cause — for the trtllm-gen Python path `is_spec_decoding_enabled` is forced to `False` at `trtllm.py:1537` so `spec_decoding_generation_lengths` is never wired in and the kernel takes the safe `fixed_q_seqlen` branch. For the legacy XQA path the same bool gate at `attentionOp.cpp:491` keeps the buffer null and `multi_query_tokens=False`. The one suspicious line I found is `attentionOp.cpp:293` (`total_num_input_tokens = mCpSize > 1 ? num_requests : num_tokens`) which would set numTokens=num_requests while seqQLengths report runtime_draft_len+1 per slot, but I couldn't find an assignment that flips `mCpSize > 1` from the PyTorch path. Couldn't reproduce the original Warp MMU fault on single-node 8×B200 or 8×B300 across multiple iterations either, so root cause remains open.

So treating this PR as the memory-safety mitigation only, leaving NVBug 6029882 open for the upstream metadata investigation. The trap+printf at least guarantees that if it recurs in CI we get an actionable signature instead of an opaque MMU fault.

bobboli:
There are other reasons behind this bug.

PR #13559, opened 2026-04-28
[None][feat] Add Laguna model support (Poolside Laguna-XS.2)

DomBrown:
## Release Notes

* **New Features**
  * Added support for Laguna language models with MoE and specialized attention mechanisms.
  * Introduced NVFP4 quantization enabling 4-bit float precision.
  * Added FP8 block-scale MoE inference optimization for improved performance.
  * Enabled KV-cache FP8 quantization support.
  * Added reasoning parser support for Laguna models.

* **Improvements**
  * Enhanced quantization configuration handling with improved safety checks.
  * Extended module exclusion patterns to support regex expressions alongside glob patterns.

## Description

Adds TensorRT-LLM PyTorch-backend support for **Poolside's Laguna model family**, including the current generation **Laguna-XS.2** in bf16, FP8 block-scale, and NVFP4 variants.

Laguna-specific bits: per-head softplus output gate (`g_proj` pre-`o_proj`), per-layer attention-head counts, per-layer dual RoPE (yarn for full-attention, default linear for sliding-window), MiniMaxM2-style sigmoid routing with per-layer routed scaling, and both legacy (`mlp_only_layers` + `decoder_sparse_step`) and current (`mlp_layer_types`) MoE-vs-dense schemas.

The branch also lands several generally useful fixes that Laguna depends on:

- Compressed-tensors NVFP4 + FP8 `kv_cache_scheme` parsing in `ModelConfig.load_hf_quant_config` / `llm_utils.ModelLoader`.
- `re:`-prefixed regex patterns supported in `QuantConfig.is_module_excluded_from_quantization` (compressed-tensors writes regex-shaped exclusion lists).
- SM120 FP8 block-scale MoE Triton fallback generalized to cover Laguna's expert layout (CUTLASS path can't initialize on Blackwell server-class shapes).
- `laguna` model_type registered with the DeepSeek-R1-style reasoning parser.

## Validation: full GSM8K (1319 samples)

Hardware: NVIDIA RTX PRO 6000 Blackwell (SM120).

| Variant | flex-extract | avg | Init | Exec |
|---|---|---|---|---|
| Laguna-XS.2 (bf16) | 85.60 ± 0.97 | **86.28** | 95s | 110s |
| Laguna-XS.2-FP8 | 86.13 ± 0.95 | **86.66** | 57s | 78s |
| Laguna-XS.2-NVFP4 | 85.44 ± 0.97 | **85.60** | 56s | 49s |

FP8 and NVFP4 stay within ~1 point of bf16 (well inside `±0.97` stderr) and execute ~1.4× / ~2.2× faster respectively. All three variants load via the same overlay script and the same modeling code, with no per-checkpoint or per-quant-format patches.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/modules/fused_moe/fused_moe_cutlass.py
```diff
@@ -520,10 +519,7 @@ def _get_quant_method(self):
             if self.quant_config.layer_quant_mode.has_fp8_qdq():
                 return FP8QDQFusedMoEMethod()
             elif self.quant_config.layer_quant_mode.has_fp8_block_scales():
-                if get_sm_version() == 120:
-                    return DeepSeekFP8BlockScalesFusedMoEMethodDeepGemm()
-                else:
-                    return DeepSeekFP8BlockScalesFusedMoEMethod()
+                return DeepSeekFP8BlockScalesFusedMoEMethod()
```
nekorobov, line 533: Is it confirmed that new kernel is always faster than DeepGemm on sm120?

nekorobov (approved):

tensorrt_llm/_torch/modules/fused_moe/fused_moe_cutlass.py
DomBrown: I've just tested locally. DeepGemm is not supported on SM120 (triggers an internal assertion)

arysef (approved):
Doc changes look good

Funatiq (approved):
Runtime changes lgtm.

Superjomn (approved):
LGTM on the llmapi changes.

tests/integration/defs/accuracy/test_llm_api_pytorch.py
```diff
@@ -6235,6 +6235,60 @@ def test_fp8(self):
                           extra_evaluator_kwargs=self.EXTRA_EVALUATOR_KWARGS)
 
 
+class TestLagunaXS(LlmapiAccuracyTestHarness):
```
crazydemo, line 6341: please add this case to llm_function_rtx6k.txt. If this case is only for rtx6k series platform. Otherwise please add it to llm_fucntion_test.txt as well.  So that QA can include this case in our testing pipeline.

tests/integration/test_lists/test-db/l0_rtx_pro_6000.yml
```diff
@@ -26,6 +26,9 @@ l0_rtx_pro_6000:
   - test_e2e.py::test_ptp_quickstart_advanced[Qwen3-30B-A3B_fp8_hf-Qwen3/saved_models_Qwen3-30B-A3B_fp8_hf] # 3mins
   - test_e2e.py::test_ptp_quickstart_advanced[Qwen3-30B-A3B_nvfp4_hf-Qwen3/saved_models_Qwen3-30B-A3B_nvfp4_hf] # 2mins
   - test_e2e.py::test_ptp_quickstart_advanced[GPT-OSS-20B-gpt_oss/gpt-oss-20b]
+  - accuracy/test_llm_api_pytorch.py::TestLagunaXS::test_bf16
```
xinhe-nv, line 36: do we need to add these tests into tests/integration/test_lists/qa/llm_function_rtx6k.txt?

xinhe-nv:
@coderabbitai review the pr as a QA engineer, check the test coverage on both function and perf

DomBrown:
@coderabbitai

- There are GSM8K accuracy reference values provided.
- "No FP8-activation Triton kernel path" This is intentional.
- Lack of multi-GPU tests is intentional to reduce CI resource usage as this is a small model

xxi-nv (approved):
Thanks for the integration.
Ideally, the Triton kernels should be imported into TritonFusedMoE. However, this backend is somewhat out - of - date. And we have not decide whether it still has value to maintain.

Considering the current limitations of the MOE backend, this is not the ideal implementation.
However, I don't have a better choice either.
Approved. I'll start to investigate how to reorganize it to make it more scalable for integrating new kernels.

tensorrt_llm/_torch/configs/laguna.py
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from transformers.configuration_utils import PretrainedConfig
+
+
+class LagunaConfig(PretrainedConfig):
```
2ez4bz, line 21: Just to confirm: would this be necessary if we were using `transformers>=5.7.0`, which seems to be the first version where laguna support was added to `transformers`? If not, can we leave a comment to remove this once we upgrade?

tensorrt_llm/_torch/models/modeling_laguna.py
```diff
@@ -0,0 +1,680 @@
...
+                overridden_tp_size=shared_expert_tp_size,
+                is_shared_expert=True,
+                reduce_output=False,
+            )
+        else:
+            self.shared_expert = None
+
+        self.routed_scaling_factor = float(getattr(config, "moe_routed_scaling_factor", 1.0))
+
+    def forward(
```
2ez4bz, line 168: Nit: `@torch.inference_mode()`?

tensorrt_llm/_torch/models/modeling_laguna.py
```diff
@@ -0,0 +1,680 @@
...
+        router_logits = self.gate(hidden_states)
+
+        routed_output = self.experts(
+            hidden_states,
+            router_logits,
+            all_rank_num_tokens=all_rank_num_tokens,
+        )
+
+        if self.routed_scaling_factor != 1.0:
+            routed_output = routed_output * self.routed_scaling_factor
```
2ez4bz: Nit: would `routed_output *= self.routed_scaling_factor` work? (Skips a tensor allocation.)

tensorrt_llm/_torch/models/modeling_laguna.py
```diff
@@ -0,0 +1,680 @@
...
+
+        pos_embd_params = PositionalEmbeddingParams(
+            type=PositionEmbeddingType.rope_gpt_neox,
+            rope=rope_params,
+        )
+
+        gating = getattr(config, "gating", False)
+        self._use_gating = bool(gating)
+        self._gate_per_head = gating == "per-head" or gating is True
+        self._num_heads_local = None  # set after super().__init__
```
2ez4bz: Maybe stupid question: why is this necessary if it's being set overwritten on line 270? Does the `super().__init__(...)` call somehow require it to exist? If there's a reason, can we leave a comment for why this was necessary?

tensorrt_llm/_torch/models/modeling_laguna.py
```diff
@@ -0,0 +1,680 @@
...
+
+        if self._use_gating:
+            g_out = num_heads if self._gate_per_head else num_heads * self.head_dim
+            # In DEP mode (enable_attention_dp=True) the attention base class uses
+            # a local mapping with tp_size=1, so self.num_heads == num_heads (full).
+            # g_proj must match: no COLUMN sharding, each GPU holds the full weight.
+            g_tp_mode = (
+                None if model_config.mapping.enable_attention_dp else TensorParallelMode.COLUMN
+            )
+            self.g_proj = Linear(
```
2ez4bz, line 281: Does this attribute exist when `self._use_gating` does not evaluate to `True`? If not, for consistency in attributes, can we set it to `None`?

tensorrt_llm/_torch/models/modeling_laguna.py
```diff
@@ -0,0 +1,680 @@
...
+            gate = self.g_proj(gate_input)
+            gate = F.softplus(gate.float()).to(attn_output.dtype)
+            if self._gate_per_head:
+                shape = attn_output.shape
+                attn_output = (
+                    attn_output.view(*shape[:-1], self._num_heads_local, self.head_dim)
+                    * gate.unsqueeze(-1)
+                ).view(shape)
+            else:
+                attn_output = attn_output * gate
```
2ez4bz: Nit: `*=` ?

tensorrt_llm/llmapi/llm_utils.py
```diff
@@ -485,12 +485,53 @@ def _update_from_hf_quant_config(self) -> bool:
                         raise ValueError(
                             f"Unsupported weights_quant_strategy: {weights_quant_strategy}. "
                             "Supported strategies: 'channel', 'block'.")
+                elif (weights_quant_config["num_bits"] == 4
```
2ez4bz, line 510: I know this pattern precedes your PR but it seems weird that we have 2 places gating on basically the same checks. Do you happen to know if that's accidental / they could be merged in the future?

2ez4bz (approved):

tensorrt_llm/_torch/configs/laguna.py
DomBrown: I suspect you are right about this, yes. I agree a note would be useful.

tensorrt_llm/_torch/models/modeling_laguna.py
DomBrown: Nice catch. Fixed.

tensorrt_llm/llmapi/llm_utils.py
DomBrown: Agree this is less than ideal but outside the scope of the PR. I can explore whether this can be cleaned in a separate PR if needed.

tests/integration/test_lists/test-db/l0_rtx_pro_6000.yml
DomBrown: Will do.

DomBrown:
Added to QA lists and added some additional tests.

fredricz-20070104 (approved):
Approved.

StanleySun639 (approved):

xinhe-nv:
@coderabbitai path: "tests/**"
      instructions: |
        Act as a QA engineer reviewing test changes for TensorRT-LLM.

        QA test list hygiene (integration / release runs):
        - If the change adds or materially alters an integration test under
          tests/integration/defs/ (or otherwise affects what QA should run on a
          schedule), call out whether an entry is needed under
          tests/integration/test_lists/qa/. See tests/integration/test_lists/qa/README.md
          for which file to use (e.g. llm_function_core.txt for primary single-node
          multi-GPU functional cases; llm_function_multinode.txt for multi-node;
          llm_perf_*.yml for perf; llm_triton_integration.txt for Triton).
        - If the PR only touches unittest/ or narrow unit scope, say explicitly
          whether QA list updates are unnecessary or optional.

        Coverage expectations:
        - Assess whether new/changed tests cover happy path, important edge cases,
          and failure modes relevant to the feature or fix (skips, guards, env
          vars like LLM_MODELS_ROOT, GPU count, backend-specific paths).
        - Flag tests that assert overly brittle behavior (e.g. exact token match
          across speculative vs non-speculative paths) unless the product contract
          requires it.
        - Note missing negative tests, missing parametrization where multiple
          backends or dtypes apply, or tests that cannot fail when behavior regresses.

        Performance test coverage:
        - If the PR touches performance-sensitive paths (attention kernels, MoE
          routing/dispatch, KV cache management, scheduler, batching logic,
          CUDA graph capture, speculative decoding, or quantization kernels),
          check whether a perf test entry is present or updated in:
          (a) tests/integration/test_lists/test-db/l0_perf.yml or the
              appropriate per-GPU l0_*.yml (primary requirement — this is how
              tests enter pre-merge and scheduled CI); and
          (b) tests/integration/test_lists/qa/llm_perf_*.yml (single-node) or
              tests/integration/test_lists/qa/llm_perf_multinode.txt (multi-node)
              for QA scheduled runs.
          Flag if (a) is missing even when (b) is present, since test-db is
          the authoritative source for CI execution.
        - Flag if a test in tests/integration/defs/perf/test_perf_sanity.py is
          missing for a new model or serving configuration that warrants a
          latency/throughput baseline (examples/benchmark/ is deprecated).
        - Note if only functional correctness is tested for a change where a
          performance regression would not be caught (e.g. a kernel rewrite with
          no perf test, or a new scheduling policy with no throughput assertion).
        - Do not require perf tests for doc-only, infra, style, or pure refactor
          changes where no runtime behavior changes.

        Keep feedback actionable: suggest concrete list file names and whether
        coverage is sufficient, insufficient, or needs follow-up outside the PR.

xinhe-nv (approved):

joerowell:
## Build issues hit when targeting sm_120 (RTX PRO 6000 Blackwell workstation)

Built the PR against `nvcr.io/nvidia/tensorrt-llm/devel:latest` and ran Laguna-XS.2-FP8 (TP=1) and Laguna-M.1-NVFP4 (TP=4) end-to-end on 4× RTX PRO 6000 Blackwell (compute capability 12.0, `sm_120f`). Both generate coherent output, but the build requires three source changes. Posting in case they're useful upstream — the model card lists supported targets as H100/H200/B200, so sm_120 isn't in scope here, but these are small enough that fixing them keeps the door open.

### 1) `moeAlltoAllKernels.cu` — `cvt` PTX guards include sm_120 where ptxas rejects

Two FP8-pair conversion specializations gate on `__CUDA_ARCH__ >= 1000` (line 1013) and `>= 890` (line 1031). Both include sm_120 (`__CUDA_ARCH__ == 1200`), but ptxas for `sm_120f` rejects the specific type combos in those `asm volatile` blocks:

\`\`\`
ptxas .../moeAlltoAllKernels.ptx: Unexpected instruction types specified for 'cvt'
ptxas fatal: Ptx assembly aborted due to errors
\`\`\`

The cmake invocation for sm_120 already passes \`-DEXCLUDE_SM_100 -DEXCLUDE_SM_100F -DEXCLUDE_SM_103 -DEXCLUDE_SM_90\`, so the build system intends sm_120 to take a non-sm_100 path — the source just uses \`__CUDA_ARCH__\` comparisons instead of the \`EXCLUDE_SM_*\` macros at these two sites. A minimal fix is to exclude sm_120 from these blocks; the generic scalar \`vec_convert\` at line 1004 takes over (slower but correct):

\`\`\`diff
-#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000) && (__CUDA_ARCH__ != 1200) && (__CUDA_ARCH__ != 1201) && (__CUDA_ARCH__ != 1210)
\`\`\`

and similarly for the `>= 890` guard. Cleaner upstream fix would migrate to `#ifndef EXCLUDE_SM_120` matching the style used elsewhere.

### 2) `executorWorker` link fails on torch ABI when venv torch ≠ container torch

\`\`\`
/usr/bin/ld: ../libtensorrt_llm.so: undefined reference to `c10::TensorImpl::decref_pyobject() const'
\`\`\`

`decref_pyobject` exists in venv torch 2.10's `libc10.so` (which `build_wheel.py`'s setup_venv pulls from PyPI), but `executorWorker`'s link line picks up the container-preinstalled torch 2.8 `libc10.so` from `/usr/local/lib/python3.12/dist-packages/torch/lib/`. The rest of the build is fine — `libtensorrt_llm.so` itself is built consistently against venv torch — so this only affects the `executorWorker` standalone CLI. Easy enough to skip the target locally for our use case (Python `LLM` API), but the path-priority issue is a real torch-ABI hazard.

### 3) `setup_venv` interactive prompt blocks non-TTY builds

When `setup_venv` detects PyPI torch on top of an NGC container, it does \`input("Press Ctrl+C to stop, any key to continue...")\`. Crashes with `EOFError` under `docker run` without `-it`, nohup, CI, etc. A simple `if sys.stdin.isatty()` guard or a `--non-interactive` flag would suffice.

### Setup steps that aren't source patches but worth documenting

- NCCL: PR adds `createNCCLWindowTensor`, which needs NCCL ≥ 2.29 (symbolic memory APIs). The devel container ships NCCL 2.27.5 — needed `apt install libnccl2=2.30.4-1+cuda12.9 libnccl-dev=2.30.4-1+cuda12.9` before the build.
- `git lfs install && git lfs pull` is required — the build aborts with that exact suggestion when LFS cubins are missing. Worth mentioning in the docs/README.
- `tensorrt_llm/runtime/kv_cache_manager_v2/rawref/` is a Python C extension with its own `setup.py`; `build_wheel.py` doesn't install it. Had to `pip install` in that dir and symlink the resulting `_rawref.cpython-312-*.so` into the source tree.
- `laguna_minimal_overlay.sh` is referenced by the Laguna-XS.2 model card as "shipped with this PR" but isn't in the branch. We didn't actually need it (the venv pulls a compatible transformers version), so possibly just stale README text — but worth confirming.

End-to-end with the three patches above: Laguna-XS.2-FP8 (TP=1) and Laguna-M.1-NVFP4 (TP=4) both load and generate sensible output. Happy to share more detail or open a follow-up PR with the source fixes if useful.

PR #11158, opened 2026-01-31
[None][chore] Align LlmArgs with some Pydantic best practices

anish-shanbhag:
## Motivation

`LlmArgs` is the main user-facing interface for configuring TRTLLM, and it uses [Pydantic](https://docs.pydantic.dev/latest/) for validating custom user configuration, including YAML files passed with `--config` / `--extra_llm_api_options`. It's important that `LlmArgs` follows Pydantic best practices so that (a) it's easier for users to understand how to configure the LLM API, and (b) we reduce the surface area for subtle bugs related to configuration parsing / validation.

This PR is also a preparatory change for #8331 which will significantly simplify the parsing of `LlmArgs` from config YAML files. The changes here would enable a variety of custom parsing logic (e.g. [`update_llm_args_with_extra_dict`](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/llmapi/llm_args.py#L3402)) to be completely removed.

## Description

This PR aims to align `LlmArgs` with some Pydantic best practices that have been added to [`CODING_GUIDELINES.md`](https://github.com/anish-shanbhag/TensorRT-LLM/blob/ashanbhag/pydantic-best-practices/CODING_GUIDELINES.md#pydantic-guidelines).

The diff is unfortunately pretty large due to the interdependent nature of the changes here. ***To help with this, I recommend that reviewers read the changes in this order***:
1. Review the new Pydantic guidelines added to [`CODING_GUIDELINES.md`](https://github.com/anish-shanbhag/TensorRT-LLM/blob/ashanbhag/pydantic-best-practices/CODING_GUIDELINES.md#pydantic-guidelines). Some highlights include:
    - Ensuring that all fields have descriptions
    - Preferring built-in Pydantic features like `model_validator` / `model_dump` etc. instead of using custom methods like `validate()` or `to_dict()`
    - Using [discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) when a field needs to accept one of several possible config classes
2. Review the new tests added in `test_llm_args.py::TestPydanticBestPractices`, which enforce a subset of the guidelines above.
3. Review the rest of the changes, which all aim to just bring `LlmArgs` into compliance with the above guidelines. Some highlights include:
    - Adding descriptions to fields in various config classes
    - Migrating the speculative decoding configs / sparse attention configs to [discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions)
[...]

PR #8527, opened 2025-10-21
[None][doc] Add doc for torch.compile & piecewise cuda graph

yizhang-nv:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guide for using Torch Compile with Piecewise CUDA Graph in TensorRT LLM, covering setup, feature mechanics, performance optimization, advanced customization, and troubleshooting for common issues.

* **Chores**
  * Updated code quality configuration to exclude additional file types from scanning.

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

`/bot [-h] ['run', 'kill', 'skip', 'reuse-pipeline'] ...`

Provide a user friendly way for developers to interact with a Jenkins server.

Run `/bot [-h|--help]` to print this help message.

See details below for each supported subcommand.

<details>

`run  [--reuse-test (optional)pipeline-id --disable-fail-fast --skip-test --stage-list "A10-PyTorch-1, xxx" --gpu-type "A30, H100_PCIe" --test-backend "pytorch, cpp" --add-multi-gpu-test --only-multi-gpu-test --disable-multi-gpu-test --post-merge --extra-stage "H100_PCIe-TensorRT-Post-Merge-1, xxx" --detailed-log --debug(experimental)]`

Launch build/test pipelines. All previously running jobs will be killed.

`--reuse-test (optional)pipeline-id ` *(OPTIONAL)* : Allow the new pipeline to reuse build artifacts and skip successful test stages from a specified pipeline or the last pipeline if no pipeline-id is indicated. If the Git commit ID has changed, this option will be always ignored. The DEFAULT behavior of the bot is to reuse build artifacts and successful test results from the last pipeline.

[...]

PR #13574, opened 2026-04-28
[https://nvbugs/5615248][fix] Broader capture of piecewise cudagraph

brb-nv:
## Description

This is part of a fix for https://nvbugspro.nvidia.com/bug/5615248.

When `enable_piecewise_cuda_graph=True`, the piecewise CUDA graph capture set silently failed to cover `num_tokens` values that the runtime engine could legitimately request, causing those forward passes to fall back to eager execution.

The model-engine filter on the candidate list used `i <= max_num_tokens`, which is looser than the engine's actual reachable ceiling `max_batch_size * (max_seq_len - 1 - num_extra_decoding_steps)`. Entries above the real ceiling were advertised as captured but warmup silently failed to record them; the outer padding logic then padded prefill chunks to a target with no captured graph and dropped to eager.

**_Proposed fix:_**

* Add `_filter_piecewise_capture_num_tokens` helper and use it in `PyTorchModelEngine.__init__` to (a) cap the candidate list at the engine's reachable ceiling and (b) append the ceiling itself so ISLs in the gap between the next-largest candidate and the ceiling still get a graph. Drops are reported via a single `logger.warning` pointing at `max_seq_len` (this was silent earlier).
* Default `capture_num_tokens` (powers-of-2 + 256-stride) is preserved to bound startup time and CUDA graph memory; the new filter handles ceiling correctness.

**_Before/after:_**

`max_batch_size=1`, `max_seq_len=128`, `enable_piecewise_cuda_graph=true`. Reachable ceiling = `1 * (128 - 1) = 127`.

| ISL  | Before (main)                    | After (this PR)         |
| ---- | -------------------------------- | ----------------------- |
| 64   | piecewise (pad to 64)            | piecewise (pad to 64)   |
| 100  | **eager** (pad to 128, no graph) | piecewise (pad to 127)  |
| 107  | **eager** (pad to 128, no graph) | piecewise (pad to 127)  |
| 121  | **eager** (pad to 128, no graph) | piecewise (pad to 127)  |
| 127  | **eager** (pad to 128, no graph) | piecewise (pad to 127)  |

## Test Coverage

```
$ pytest tests/unittest/llmapi/test_llm_args.py::TestPiecewiseCudaGraphCaptureDefaults -s -v
```

## PR Checklist

## GitHub Bot Help

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA-graph capture candidate filtering to account for KV-cache constraints and extra decoding steps, preventing silent failures during warmup.
  * Added warning logs for capture candidates that exceed reachable limits.

* **Behavior Changes**
  * `capture_num_tokens` no longer receives automatic defaults; users must now explicitly specify values when enabling piecewise CUDA graphs.

### DIFF
--- a/tensorrt_llm/llmapi/llm_args.py
+++ b/tensorrt_llm/llmapi/llm_args.py
@@ -3520,6 +3520,13 @@ class TorchCompileConfig(StrictBaseModel):
         description=
         "The maximum number of CUDA streams to use for torch.compile.")
 
+    @model_validator(mode='after')
+    def set_default_capture_num_tokens(self) -> 'TorchCompileConfig':
+        if self.enable_piecewise_cuda_graph and self.capture_num_tokens is None:
+            self.capture_num_tokens = [2**i for i in range(8)
+                                       ] + [i for i in range(256, 3073, 256)]
+        return self
+
 
 class TorchLlmArgs(BaseLlmArgs):
     # PyTorch backend specific configurations
--- a/tensorrt_llm/llmapi/llm_args.py
+++ b/tensorrt_llm/llmapi/llm_args.py
@@ -3520,13 +3520,6 @@ class TorchCompileConfig(StrictBaseModel):
         description=
         "The maximum number of CUDA streams to use for torch.compile.")
 
-    @model_validator(mode='after')
-    def set_default_capture_num_tokens(self) -> 'TorchCompileConfig':
-        if self.enable_piecewise_cuda_graph and self.capture_num_tokens is None:
-            self.capture_num_tokens = [2**i for i in range(8)
-                                       ] + [i for i in range(256, 3073, 256)]
-        return self
-
 
 class TorchLlmArgs(BaseLlmArgs):
     # PyTorch backend specific configurations
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -98,6 +98,47 @@ class ModelEngine(ABC):
         return
 
 
+def _filter_piecewise_capture_num_tokens(
+    candidate_num_tokens: list[int],
+    max_num_tokens: int,
+    max_batch_size: int,
+    max_seq_len: int,
+    num_extra_decoding_steps: int = 0,
+) -> Tuple[list[int], list[int]]:
+    """Cap a piecewise CUDA graph capture-set at the largest forward-pass
+    `num_tokens` value the engine can ever construct.
+
+    A piecewise CUDA graph for `num_tokens = N` is only useful if a real
+    forward pass can ever flow N tokens through the model. Every in-flight
+    request must leave room for at least one decode token in its KV cache,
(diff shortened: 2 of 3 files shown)


tensorrt_llm/llmapi/llm_args.py
pcastonguay, line 3554: IIUC this is removed so that the default is `cuda_graph_batch_sizes` if not specified? This would be a breaking change no? If so, we should add `BREAKING: ` to PR title.

@yijingl-nvidia @Superjomn to review as well. Thanks.

brb-nv, line 3548: Hi @anish-shanbhag, do you have any concerns about this change? Asking because you added this here:
https://github.com/NVIDIA/TensorRT-LLM/pull/11158/changes#diff-bf140a489175e4c6e5931cd19cc57e06c1634483517f06c4e6fda96d5d26b056

If you think I should request review from someone else, please suggest their name. Thank you!

anish-shanbhag: My change above was mainly intended as a no-op refactor to move the logic from `TorchLlmArgs` into `TorchCompileConfig` directly (previously it was here: https://github.com/NVIDIA/TensorRT-LLM/pull/11158/changes#diff-bf140a489175e4c6e5931cd19cc57e06c1634483517f06c4e6fda96d5d26b056L3312). 

It looks like the logic in `TorchCompileConfig` was originally added by @yizhang-nv in https://github.com/NVIDIA/TensorRT-LLM/pull/8527, maybe Yi can review?

brb-nv: Thank you!

yijingl-nvidia: I've reviewed the code. The code does what the PR is said to do and I agree changing the code to the documented behavior is nice.

One worry I have is that now I've changed the default cuda_graph_batch_sizes to have 40+ values, by default this batch size list will be inherited by the piecewise CUDA graph, increasing server startup time and CUDA graph memory usage further.

On the bright side, KV cache compression methods like what is used in DeepSeek V4 and our team's work on improving server startup time should mitigate the effect of the startup time and memory increase.

brb-nv: I think a reasonable compromise is to keep the validator so that piecewise cudagraph count doesn't blow up but keep my fix to make sure the ceiling is always present.
https://github.com/NVIDIA/TensorRT-LLM/pull/13574/commits/d8859dfcf85a7633be0c4377de37bc5f727d3433

Updated the MR description accordingly.

yizhang-nv (approved):
LGTM.

tensorrt_llm/llmapi/llm_args.py
yijingl-nvidia: The updated PR LGTM!

yijingl-nvidia (approved):

pcastonguay (approved):

PR #13576, opened 2026-04-28
[None][fix] add kv_scales/inv_kv_scales to FP8BlockScalesLinearMethod and FP8RowwiseLinearMethod

jjia-droid:
## Problem

When a model uses FP8 block-scale weights (e.g. Qwen3-4B-Thinking-2507-FP8, which has weight_scale_inv 2-D block-scale tensors) with NVFP4 KV cache, the server crashes with an AttributeError:

  AttributeError: 'Linear' object has no attribute 'kv_scales'

This happens because attention.py accesses qkv_proj.kv_scales unconditionally whenever has_fp4_kv_cache() is True:

```
  if self.quant_config.layer_quant_mode.has_fp4_kv_cache():
      kv_scales_sf = self.qkv_proj.kv_scales      # <-- crashes here
      kv_scales_sf_inv = self.qkv_proj.inv_kv_scales

```
`FP8QDQLinearMethod` and `NVFP4LinearMethod `already create these parameters in create_weights(), but `FP8BlockScalesLinearMethod` and `FP8RowwiseLinearMethod` did not, so any FP8 block-scale or FP8 rowwise model + NVFP4 KV cache would hit this crash.

 ## Repro

  Serve a block-scale FP8 checkpoint (e.g. Qwen/Qwen3-4B-Thinking-2507-FP8, which uses weight_scale_inv 2-D block tensors) with kv_cache_dtype: nvfp4:
```

  trtllm-serve Qwen/Qwen3-4B-Thinking-2507-FP8 --kv_cache_dtype nvfp4
  # Before fix: AttributeError: 'Linear' object has no attribute 'kv_scales'
  # After fix:  server starts and runs normally

  ▎ Note for block-scale FP8 checkpoints: the checkpoint's hf_quant_config.json must declare "quant_algo": "FP8_BLOCK_SCALES" (not "FP8") so TRT-LLM selects
  ▎ FP8BlockScalesLinearMethod and loads weight_scale_inv correctly. A file copied from a per-tensor FP8 checkpoint will silently use the wrong linear method.
```

### How to use with a block-scale FP8 checkpoint

Checkpoints quantised in the HuggingFace fp8 block-scale format (weight_scale_inv 2-D tensors, weight_block_size=[128,128] in config.json) must have an hf_quant_config.json that declares the correct algo:

```
  {
    "quantization": {
      "quant_algo": "FP8_BLOCK_SCALES",
      "kv_cache_quant_algo": "FP8",
      "exclude_modules": ["lm_head"]
    }
  }
```

Without this file (or with a file copied from a per-tensor FP8 checkpoint that says quant_algo "FP8"), TRT-LLM selects FP8QDQLinearMethod, which loads the 2-D weight_scale_inv tensors incorrectly (silently uses weight_scale=1.0).

## Fix

Add kv_scales (shape [3], dtype float32) and inv_kv_scales to both FP8BlockScalesLinearMethod.create_weights() and
FP8RowwiseLinearMethod.create_weights(), initialised to all-ones.

The default unit global scales are intentional: these models typically do not carry k_scale / v_scale calibration tensors in their checkpoint (those only exist in per-tensor FP8 checkpoints like nvidia/Qwen3-8B-FP8). With unit global scales the NVFP4 KV cache relies entirely on its per-16-element block-local scales, which provide accurate quantization for any input range without pre-calibration.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced KV-cache quantization support for linear modules with improved parameter initialization during model creation.

## PR Checklist

- [ ] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help


tensorrt_llm/_torch/modules/linear.py
```diff
@@ -942,6 +942,12 @@ def create_weights(self, module: Linear, in_features: int,
         module.inv_input_scale = Parameter(torch.tensor(1.,
                                                         dtype=torch.float32),
                                            requires_grad=False)
+        # K, V scales for NVFP4 KV cache (default to 1.0 when not calibrated)
+        module.kv_scales = Parameter(torch.ones(3, dtype=torch.float32),
```
QiJune, line 946: tensorrt_llm/_torch/modules/attention.py:756-759 requires qkv_proj.kv_scales whenever FP4 KV cache is enabled, but the PR fixes that by adding attention/KV-cache state into two more LinearMethod.create_weights() implementations at linear.py:945-950 and linear.py:1093-1098.

That fixes the immediate crash, but the invariant is still scattered across quantized linear methods. A future linear method, or a KV-only FP4-cache path with a different QKV projection method, can hit the same bug again. KV cache scales are required by Attention, not by FP8 rowwise/block-scale GEMM.

Recommended middle-ground fix: keep the parameters attached to qkv_proj for loader compatibility, but make Attention centrally enforce their existence.


```python
  diff --git a/tensorrt_llm/_torch/modules/attention.py b/tensorrt_llm/_torch/modules/attention.py
  @@
       def create_weights(self):
           # self.attn has no weights but has states that are related to quant_config,
           # which could be modified after __init__
           self.attn.update_quant_config(self.quant_config)

  +        self.qkv_proj.create_weights()
  +        self._ensure_qkv_proj_kv_cache_scales()
           self.o_proj.create_weights()
           self.has_quant_scale = (self.o_proj.has_fp8_qdq or self.o_proj.has_nvfp4
                                   or self.o_proj.has_fp8_block_scales
                                   or self.o_proj.has_fp8_rowwise
                                   or self.o_proj.has_w4a8_nvfp4_fp8)
  +
  +    def _ensure_qkv_proj_kv_cache_scales(self):
  +        if not (
  +            self.quant_config is not None
  +            and self.quant_config.layer_quant_mode.has_fp4_kv_cache()
  +        ):
  +            return
  +
  +        if not hasattr(self.qkv_proj, "kv_scales"):
  +            self.qkv_proj.register_parameter(
  +                "kv_scales",
  +                nn.Parameter(torch.ones(3, dtype=torch.float32),
  +                             requires_grad=False),
  +            )
  +        if not hasattr(self.qkv_proj, "inv_kv_scales"):
  +            self.qkv_proj.register_parameter(
  +                "inv_kv_scales",
  +                nn.Parameter(torch.ones(3, dtype=torch.float32),
  +                             requires_grad=False),
  +            )
  +
  +        if self.qkv_proj.kv_scales.shape != torch.Size([3]):
  +                f"Expected qkv_proj.kv_scales shape [3], got "
  +                f"{tuple(self.qkv_proj.kv_scales.shape)}.")
  +        if self.qkv_proj.inv_kv_scales.shape != torch.Size([3]):
  +                f"Expected qkv_proj.inv_kv_scales shape [3], got "
  +                f"{tuple(self.qkv_proj.inv_kv_scales.shape)}.")

``` 

Then the two new blocks in FP8RowwiseLinearMethod.create_weights() and FP8BlockScalesLinearMethod.create_weights() can be removed. Existing FP8QDQLinearMethod/NVFP4LinearMethod can keep their current fields; the helper is idempotent.

PR #13577, opened 2026-04-28
[None][fix] Reject incompatible KV connector configurations at construction time

jthomson04:
## Summary

The KV connector contract has a number of implicit assumptions that the rest of `PyExecutor` can violate without any runtime check today, leading to silent miscompute. This PR adds an explicit `NotImplementedError` block in `_maybe_init_kv_connector_manager` for every configuration the connector cannot currently service correctly.

### Configurations now rejected at construction time

| Config | Reason |
|---|---|
| `KvCacheConfig.host_cache_size > 0` | Connector worker is registered only with the primary (GPU) pool via `kv_cache_manager.get_unique_primary_pool()` and has no view of the secondary (host) pool. Block ids handed to `update_state_after_alloc` / `request_finished` may already point to blocks evicted to host. The connector's load/save also runs on the forward stream while `KvCacheTransferManager` uses its own onboard/offload streams, so there's no synchronization between the two. |
| `pp_size > 1` **or** `cp_size > 1` | Pool registration is per-rank with no cross-rank coordination. The previous PP-only guard is generalized to PP **or** CP. |
| `max_beam_width > 1` | `RequestData.block_ids` in the connector contract has no notion of beams, so non-leading beams would save and restore the wrong KV state. |
| `enable_attention_dp` | Dummy requests inserted for cross-DP balancing flow through the connector scheduler/worker hooks indistinguishably from real requests. |
| VSWA (`KVCacheManager.is_vswa`) | The connector is registered with a single primary pool, but VSWA models allocate one pool per window size. Today this surfaces as a low-level C++ assertion in `KVCacheManager::getUniquePrimaryPool` (`getWindowSizesMetadata().size() == 1`); this PR surfaces it as a friendly `NotImplementedError` at the same place we reject the others. |
| `BaseMambaCacheManager` instances (Mamba / hybrid linear attention) | Mamba layers carry SSM state rather than per-token KV blocks, so the per-`DecoderLayer` load/save hooks have nothing meaningful to transfer. |

### Already covered elsewhere — no new guard added

- **Cross-attention / encoder-decoder.** Hard-blocked at the resource-manager level via `assert is_cross_attention is False` in `KVCacheManager._calculate_blocks_per_window` (resource_manager.py). The connector cannot reach a cross-attention setup, so a duplicate guard would be dead code.
- **`MambaHybridCacheManager` for Nemotron-Hybrid.** Already raises `NotImplementedError` at cache-manager construction (`_util.py`). The new `BaseMambaCacheManager` `isinstance` check generalizes this to any current/future Mamba manager and gives a uniform diagnostic with the other connector exclusions.

### What is *not* changed

- `enable_block_reuse` — the connector contract is built around it via `get_num_new_matched_tokens` adding on top of on-device reuse; existing tests cover it.
- `enable_chunked_prefill`, overlap scheduler, KV cache transceiver — already covered by dedicated connector tests; no behavioral change.
- LoRA / PEFT, guided decoding, CUDA graphs, MTP-Eagle in-place spec — orthogonal to KV layout.

The longer-term fix (extending the connector contract so the cache manager guarantees onboarding before connector save / no offload during connector load, plumbing per-tier pool tensors and beam/DP-aware metadata through `register_kv_caches` / `build_connector_meta`, etc.) is left for follow-up. This PR just stops the silent-misuse paths.

## Test plan

- [x] `tests/integration/defs/llmapi/test_llm_api_connector.py::test_connector_rejects_unsupported_config[host_offloading]`
- [x] `tests/integration/defs/llmapi/test_llm_api_connector.py::test_connector_rejects_unsupported_config[beam_search]`
- [x] `tests/integration/defs/llmapi/test_llm_api_connector.py::test_connector_rejects_unsupported_config[attention_dp]`
- [x] All three parametrized cases registered in `tests/integration/test_lists/test-db/l0_a10.yml`.
- VSWA, Mamba and CP are detected via state that requires either a specific model checkpoint (VSWA, Mamba) or distributed setup (CP), so they are guarded by code review rather than a parametrized test in this PR.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -61,7 +61,7 @@ from .hang_detector import HangDetector
 from .kv_cache_transceiver import KvCacheTransceiver
 from .llm_request import (ExecutorRequest, LlmRequest, LlmRequestState,
                           LlmResponse, get_draft_token_length)
-from .mamba_cache_manager import MambaHybridCacheManager
+from .mamba_cache_manager import BaseMambaCacheManager, MambaHybridCacheManager
 from .model_engine import ModelEngine
 from .perf_metrics_manager import PerfMetricsManager
 from .request_utils import (RequestBroadcaster, attach_py_objects_to_requests,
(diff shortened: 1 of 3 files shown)


pcastonguay:
@jthomson04 could you resolve the merge conflict? Thanks.

pcastonguay (approved):

PR #13590, opened 2026-04-29
[None][feat] Bind NUMA-aware CPU affinity in visual_gen workers

luyiyun1021:
## Release Notes

* **New Features**
  * Added CPU affinity configuration for optimized performance on multi-socket systems.

* **Refactor**
  * Consolidated CPU affinity management into shared utility functions for improved maintainability.

## Description

Each `visual_gen` diffusion worker process currently inherits whatever CPU affinity the parent / launcher sets, which on multi-GPU video generation is typically the full machine. This causes the host-side Python work (compile, CUDA launch, NCCL collective setup) to bounce across NUMA nodes and slows down end-to-end generation.

The LLM worker path already solves this in `BaseWorker._configure_affinity` by querying NVML for the GPU's NUMA-local CPU set and pinning the process via `psutil`. This PR extracts that logic into a free helper so both workers share one source of truth, then calls it from the diffusion worker entry point.

**Changes:**

1. `tensorrt_llm/llmapi/utils.py` — new free function `configure_cpu_affinity(device_id)` that contains the existing probing logic (env-var contract unchanged).
2. `tensorrt_llm/executor/base_worker.py` — `_configure_affinity` now delegates to the shared helper; `psutil` import dropped (no longer used here).
3. `tensorrt_llm/_torch/visual_gen/executor.py` — `run_diffusion_worker` calls `configure_cpu_affinity(device_id)` right after `torch.cuda.set_device`. Wrapped in try/except so a missing NVML / non-NUMA host falls through with a warning instead of crashing the worker.

**Drive-by:** 8 pre-existing single-quote (`'''`) docstrings in `utils.py` were converted to triple-double-quote (`"""`) to satisfy ruff `D300`. The `ruff-legacy` baseline for this file was stale — it expected 0 D300 violations but the file actually had 8. Without this drive-by fix, any edit to `utils.py` would fail the gate. Pure quote-style change, no semantic effect; the fix actually reduces total ruff-legacy violations in the file.

**Env-var contract** (identical to the existing LLM-worker path):

| `TLLM_NUMA_AWARE_WORKER_AFFINITY` | Behavior |
|---|---|
| unset | auto-configure if affinity is unconstrained; clear constraints otherwise |
| `1` | unconditionally auto-configure |
| `0` or any other value | disabled |

## Test Coverage

- Local 2 / 4 / 8-GPU LTX2 video generation runs: each worker logs `CPU affinity set to <NUMA-local CPUs> for optimal NUMA-aware scheduling.` with the expected device-local CPU list (~112 CPUs per GPU on B200 dual-socket).
- `TLLM_NUMA_AWARE_WORKER_AFFINITY=0` opt-out path verified — workers leave affinity untouched.
- LLM worker path unchanged in behavior (`_configure_affinity` is now a one-line delegate; semantics preserved).

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/visual_gen/executor.py
+++ b/tensorrt_llm/_torch/visual_gen/executor.py
@@ -16,6 +16,7 @@ import zmq
 from tensorrt_llm._torch.visual_gen.output import PipelineOutput
 from tensorrt_llm._torch.visual_gen.pipeline_loader import PipelineLoader
 from tensorrt_llm.executor.ipc import ZeroMqQueue
+from tensorrt_llm.llmapi.utils import configure_cpu_affinity
 from tensorrt_llm.logger import logger
 from tensorrt_llm.visual_gen.args import VisualGenArgs
 
@@ -353,6 +354,14 @@ def run_diffusion_worker(
         device_id = _local_rank % torch.cuda.device_count() if torch.cuda.is_available() else 0
         if torch.cuda.is_available():
             torch.cuda.set_device(device_id)
+            try:
+                configure_cpu_affinity(device_id)
+            except Exception as e:
+                logger.warning(
+                    f"[rank {rank}] NUMA-aware CPU affinity setup failed: {e}. "
+                    f"The worker will run without NUMA pinning, which may impact "
+                    f"performance."
+                )
 
         dist.init_process_group(
             backend="cuda:nccl,cpu:gloo" if torch.cuda.is_available() else "gloo",
--- a/tensorrt_llm/executor/base_worker.py
+++ b/tensorrt_llm/executor/base_worker.py
@@ -16,13 +16,11 @@ import datetime
 import enum
 import gc
 import json
-import os
 import weakref
 from pathlib import Path
 from queue import Queue
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
-import psutil
 import torch
 
 from tensorrt_llm.logger import logger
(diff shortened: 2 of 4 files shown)


tensorrt_llm/executor/base_worker.py
```diff
@@ -122,54 +120,10 @@ def __init__(
     def _configure_affinity(self, device_id):
         '''Probe and configure the CPU affinity of the worker based on NUMA topology.
```
zhenhuaw-me: Do we need to keep this helper or all call sites should use the new one?

luyiyun1021: I think the lattter one is better. All call sites should use the new func

tensorrt_llm/_torch/visual_gen/executor.py
```diff
@@ -353,6 +354,10 @@ def run_diffusion_worker(
         device_id = _local_rank % torch.cuda.device_count() if torch.cuda.is_available() else 0
         if torch.cuda.is_available():
             torch.cuda.set_device(device_id)
+            try:
+                configure_cpu_affinity(device_id)
+            except Exception as e:
+                logger.warning(f"[rank {rank}] NUMA-aware CPU affinity setup failed: {e}")
```
zhenhuaw-me: Should we provide some instruction for the user? E.g., safe to ignore / there might be some perf impact.

Or leave the warning to be in the util?

tensorrt_llm/llmapi/utils.py
```diff
@@ -622,6 +622,56 @@ def get_numa_aware_cpu_affinity(device_id):
     return cpu_affinity
 
 
+def configure_cpu_affinity(device_id: int) -> None:
```
zhenhuaw-me, line 625: Should we move this to some file that is out of the API part? Anyway, it's out of the scope of this PR.

zhenhuaw-me (approved):

tensorrt_llm/llmapi/utils.py
luyiyun1021: The affinity setup codes are previously in this file. May leave as followup.

tensorrt_llm/_torch/visual_gen/executor.py
luyiyun1021: Added a note that there might be some perf impact

PR #13606, opened 2026-04-29
[https://nvbugs/6072125][fix] harden allreduce rmsnorm fusion multigpu test

tcherckez-nvidia:
…u test

## Summary by CodeRabbit

* **Tests**
  * Improved distributed training test reliability and resource management by deferring port selection to MPI workers and adding synchronization barriers to prevent conflicts during teardown.

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

To see a list of available CI bot commands, please comment `/bot help`.

PR #13610, opened 2026-04-29
[https://nvbugs/6072808][fix] Retry on EADDRINUSE in AutoDeploy allreduce-fusion test

MrGeva:
The multi-GPU test_allreduce_fusion test reserves a port via get_free_port() in the parent and broadcasts it to MpiPoolSession workers, which then call dist.init_process_group("nccl"). There is a TOCTOU race: the port can be grabbed between the parent socket close and the NCCL bind, producing DistNetworkError(EADDRINUSE) and a CI failure (e.g. on DGX_H100-4_GPUs-AutoDeploy-1).

The spawn_multiprocess_job path already recovers from this via _PORT_CONFLICT_EXIT_CODE, but the MpiPoolSession path does not. Wrap submit_sync in a 5-attempt retry that fetches a fresh port and recreates the MPI pool when DistNetworkError(EADDRINUSE) is observed; other errors propagate immediately, mirroring the existing spawn_multiprocess_job behavior.

## Summary by CodeRabbit

* **Tests**
  * Enhanced test reliability by implementing retry logic to handle port allocation conflicts during distributed training initialization.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tests/unittest/auto_deploy/multigpu/transformations/library/test_allreduce_residual_rmsnorm_fusion.py
+++ b/tests/unittest/auto_deploy/multigpu/transformations/library/test_allreduce_residual_rmsnorm_fusion.py
@@ -3,6 +3,7 @@
 import pytest
 import torch
 from _dist_test_utils import get_device_counts
+from torch.distributed import DistNetworkError
 from torch.export import export
 
 from tensorrt_llm._torch.auto_deploy.custom_ops.distributed.trtllm_dist import (
@@ -152,16 +153,31 @@ def _test_allreduce_fusion(port: int, ModuleCls, strategy: str):
 def test_allreduce_fusion(device_count, ModuleCls, strategy):
     if device_count <= 1:
         pytest.skip("Require multi GPUs to run test_allreduce_fusion.")
-    port = get_free_port()
 
     n_workers = device_count
-    mpi_pool = MpiPoolSession(n_workers=n_workers)
-    try:
-        mpi_pool.submit_sync(
-            _test_allreduce_fusion,
-            port=port,
-            ModuleCls=ModuleCls,
-            strategy=strategy,
-        )
-    finally:
-        mpi_pool.shutdown()
+    # Retry on EADDRINUSE: there is a TOCTOU race between get_free_port() in
+    # the parent and dist.init_process_group("nccl") in the workers. The
+    # spawn_multiprocess_job path handles this internally; the MpiPoolSession
+    # path used here does not, so retry with a fresh port and pool.
+    max_retries = 5
+    last_exc: Exception | None = None
+    for attempt in range(max_retries):
+        port = get_free_port()
+        mpi_pool = MpiPoolSession(n_workers=n_workers)
+        try:
+            mpi_pool.submit_sync(
+                _test_allreduce_fusion,
+                port=port,
+                ModuleCls=ModuleCls,
+                strategy=strategy,
+            )
+            return
+        except DistNetworkError as e:
+            last_exc = e
+            if "EADDRINUSE" not in str(e) and "address already in use" not in str(e).lower():
+                raise
+        finally:
+            mpi_pool.shutdown()
+    raise RuntimeError(
+        f"Failed to initialize distributed group after {max_retries} attempts due to repeated port conflicts"
+    ) from last_exc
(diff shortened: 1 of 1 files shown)


tcherckez-nvidia:
already opened https://github.com/NVIDIA/TensorRT-LLM/pull/13606 which has different approach

tests/unittest/auto_deploy/multigpu/transformations/library/test_allreduce_residual_rmsnorm_fusion.py
```diff
@@ -152,16 +153,31 @@ def _test_allreduce_fusion(port: int, ModuleCls, strategy: str):
...
-    try:
-        mpi_pool.submit_sync(
-            _test_allreduce_fusion,
-            port=port,
-            ModuleCls=ModuleCls,
-            strategy=strategy,
-        )
-    finally:
-        mpi_pool.shutdown()
+    # Retry on EADDRINUSE: there is a TOCTOU race between get_free_port() in
```
suyoggupta: Time-of-check-to-time-of-use (TOCTOU) <-- TIL. let's add this to the comment?

tests/unittest/auto_deploy/multigpu/transformations/library/test_allreduce_residual_rmsnorm_fusion.py
```diff
@@ -152,16 +153,31 @@ def _test_allreduce_fusion(port: int, ModuleCls, strategy: str):
...
+    last_exc: Exception | None = None
+    for attempt in range(max_retries):
+        port = get_free_port()
+        mpi_pool = MpiPoolSession(n_workers=n_workers)
+        try:
+            mpi_pool.submit_sync(
+                _test_allreduce_fusion,
+                port=port,
+                ModuleCls=ModuleCls,
+                strategy=strategy,
```
suyoggupta, line 175: won't we need this logic across all the other multi-gpu tests? why do we add it only to allreduce_residual_rmsnorm_fusion ?

tests/unittest/auto_deploy/multigpu/transformations/library/test_allreduce_residual_rmsnorm_fusion.py
MrGeva: done

tests/unittest/auto_deploy/multigpu/transformations/library/test_allreduce_residual_rmsnorm_fusion.py
MrGeva: MpiPoolSession is required because the test exercises trtllm's MPI-mode allreduce ops, which only activate when is_ompi() is true. the other tests don't and can't run on the trtllm allreduce backend. I will comment that in the code

MrGeva:
> already opened #13606 which has different approach

@tcherckez-nvidia your PR: 
1. Move get_free_port() into the workers — rank 0 picks the port and mpi_broadcasts it to peers, instead of the parent picking it
  and passing it through submit_sync. Shrinks the TOCTOU window from "parent picks → MPI submit/spin-up → workers init NCCL"
  (milliseconds–seconds) to "worker rank 0 picks → broadcast → init NCCL" (microseconds).

  2. Add torch.distributed.barrier() before cleanup() — keeps all ranks aligned at teardown so a fast rank can't return to the
  parent and start the next parametrization while a slow rank still holds the rendezvous socket.

The PRs are complementary, not competing. your two ideas are both legitimate hardening:
  - The barrier addresses something my fix doesn't — leaked rendezvous between adjacent parametrizations.
  - Picking the port inside workers is genuinely cleaner (no large race window), but on shared CI hosts a microsecond window can and does hit; "smaller probability" ≠ "won't happen."

So let's merge both

tcherckez-nvidia:
Can we fix it other then trying 5 times? 
maybe there should be some sort of tests resource manager that assigns ports per process to avoid conflicts

tcherckez-nvidia (approved):

MrGeva:
@tcherckez-nvidia summatizing our conversation for now: its too complex to add such port manager and won't eliminate collisions because not all tests will be using it. I think this method gives a very high success rate. it solved the problem with the other dis tests that use pt dist. this test uses mpi dist so it also needs such mechanism

MrGeva:
/bot run --disable-fail-fast --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

MrGeva:
/bot run --disable-fail-fast --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

MrGeva:
/bot skip --comment "the failures are not related to this change, all AD tests passed"

PR #13614, opened 2026-04-29
[TRTLLM-11408][feat] Add VisualGen TP Support

belgarten-nv:
## Release Notes

* **New Features**
  * Added tensor-parallel size configuration option for visual generation tasks.
  * Introduced RMSNorm normalization module with multiple execution backends (NVFP4, CUDA-Tile, FlashInfer).

* **Bug Fixes**
  * Improved tensor-parallel handling in attention and transformer components; tensors are now correctly distributed across parallel groups.
  * Updated default all-reduce strategy to NCCL for improved distributed communication.

* **Refactor**
  * Enhanced device mesh initialization and distributed process group management.
  * Optimized attention QKV projection computation and layer reduction behavior for better performance under tensor-parallel configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13614)

## Description
- Add TP support to VisualGen for Attention and MLP layers.
- Setup necessary mappings and initializes tensorrt_llm process groups.
- Parallelizes QKV Projection and Attention.
- Adds custom RMSNorm to support parallel crosshead QKV normalization.
- Allows `torch.compile` around `dist.AllReduce` by factoring out device mesh slicing to constructor.
- Adds `--tp_size` flag to VisualGen example scripts.
 
## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/distributed/ops.py
```diff
@@ -897,7 +900,7 @@ def forward(
                 scale=all_reduce_params.scale,
                 bias=all_reduce_params.bias,
                 workspace=self.workspace,
-                group=self.mapping.tp_group,
+                group=self.tp_group,
```
NVShreyas: do we need these changes since self.tp_group seems to always be self.mapping.tp_group

tensorrt_llm/_torch/visual_gen/modules/attention.py
```diff
@@ -191,6 +214,8 @@ def _init_qkv_proj(self) -> None:
                     "k": (self.q_dim, self.kv_dim),
                     "v": (self.q_dim + self.kv_dim, self.kv_dim),
                 },
+                # TODO: Fused QKV TP support
+                reduce_output=False,
```
NVShreyas: is there anything stopping it or just untested?

tensorrt_llm/_torch/visual_gen/modules/rms_norm.py
```diff
@@ -0,0 +1,329 @@
...
+        allreduce_strategy: AllReduceStrategy = AllReduceStrategy.NCCL,
+    ):
+        super().__init__()
+
+        if use_gemma and not has_weights:
+            raise ValueError("has_weights must be True if use_gemma is True")
+        if quantize_type is not None:
+            if quantize_type != "nvfp4":
+                raise NotImplementedError(
+                    f"Quantize type {quantize_type} not implemented in RMSNorm")
```
NVShreyas: could we strip out some of these features that were used in LLMs but aren't needed for VisualGen?

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -143,7 +143,7 @@ def seq_parallel_size(self) -> int:
 
     @property
     def n_workers(self) -> int:
-        return self.dit_cfg_size * self.seq_parallel_size
+        return self.dit_cfg_size * self.dit_ulysses_size * self.dit_tp_size
```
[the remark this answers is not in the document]
NVShreyas: @belgarten-nv could you use self.seq_parallel_size or add a reminder to do it before merge

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -138,11 +138,13 @@ def build_mesh(self):
         if cls.device_mesh is not None:
             return
 
-        shape = tuple(self._dim_sizes[d] for d in self._dim_names)
+        dims = ("pp",) + (self._dim_names) + ("cp",)
+        shape = (1,) + tuple(self._dim_sizes[d] for d in self._dim_names) + (1,)
```
NVShreyas: Why is this needed?

tensorrt_llm/_torch/visual_gen/pipeline_loader.py
```diff
@@ -123,6 +123,12 @@ def _setup_visual_gen_mapping(self, config: DiffusionModelConfig) -> None:
         config.visual_gen_mapping = vgm
         config.mapping = vgm.to_llm_mapping()
 
+        if dist.is_initialized() and ws > 1:
+            from tensorrt_llm._utils import torch_pybind11_abi
+            from tensorrt_llm.bindings.internal.process_group import init_pg
+            init_pg(dist.group.WORLD, dist.group.WORLD,
+                    torch_pybind11_abi())
```
NVShreyas: why is this needed?

tensorrt_llm/_torch/visual_gen/modules/rms_norm.py
belgarten-nv: My idea was to leave it incase someone wants these things in the future. What's your guidance here?

NVShreyas: lets remove it and add a note saying it was copied from LLM/modules so someone can copy it over if theres something missing in the future

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -11,7 +11,7 @@
 from tensorrt_llm._torch.modules.layer_norm import LayerNorm
```
NVShreyas, line 11: we'll probably need to update Flux and LTX models too

tests/unittest/_torch/visual_gen/multi_gpu/test_tp_attention.py
```diff
@@ -0,0 +1,455 @@
...
+
+def _logic_tp_ulysses_combined(rank, world_size):
+    """TP + Ulysses combined matches F.sdpa reference on the full sequence.
+
+    4 GPUs: tp_size=2, ulysses_size=2.
+    Ulysses shards the sequence; TP shards the heads/weights.
+    """
+    assert world_size == 4
+    tp_size = 2
+    ulysses_size = 2
```
NVShreyas: can you also add tests for uneven sizes?

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -139,8 +130,36 @@ def __init__(
         }
 
         if dist.is_initialized() and world_size > 1:
+            self.setup_communicators()
             self.build_mesh()
-            self._build_vae_group()
+
+    def setup_communicators(self):
+
+        hostname = gethostname()
```
NVShreyas: if running inside docker on multiple nodes, would this return a unique hostname? Maybe we should use torchrun, slurm specific variables here. 
CC suggested something like this but feel free to use some other method if this doesnt make sense - 
```
def _get_node_rank(self) -> int:
  """Resolve node rank from whichever launcher is active.
  
  torchrun sets NODE_RANK.
  SLURM srun sets SLURM_NODEID.
  Plain mp.Process (single-node): no env var → 0.
  """
  for var in ("NODE_RANK", "SLURM_NODEID"):
      val = os.environ.get(var)
      if val is not None:
          return int(val)
  return 0  # single-node mp.Process: all ranks are co-located
```

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -139,8 +130,36 @@ def __init__(
...
+
+        host_to_ranks = {}
+        for rank, host in all_hosts:
+            host_to_ranks.setdefault(host, []).append(rank)
+
+        self.local_comm = None
+        for ranks in host_to_ranks.values():
+            # All global ranks from the default process group to participate in the call,
+            # even if some ranks are not part of the new process group being created
+            pg = dist.new_group(ranks=ranks, backend='cuda:nccl,cpu:gloo')
```
NVShreyas: nit: `ranks=sorted(ranks)` here to guarantee ordering. Not a hard requirement

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -142,6 +143,7 @@ def __init__(
             force_dynamic_quantization=model_config.force_dynamic_quantization
             if model_config
             else False,
+            reduce_output=False,
```
NVShreyas, line 147: whys reduce_output False for both ff_in and ff_out?

NVShreyas: nvm, this just means TP isnt enabled for WanImageEmbedding, right?

belgarten-nv: Exactly; the default parameter for reduce_output is True for Linear, and since we are passing the mapping here via the model_config, this would do the all reduce here. Since we aren't using TP for these layers, we need to disable.

NVShreyas:
@karljang / @yibinl-nvidia could you guys please review from FLUX / LTX perspective?

cc @chang-l for review

tensorrt_llm/_torch/visual_gen/modules/rms_norm.py
```diff
@@ -0,0 +1,108 @@
...
+                                            device=device),
+                                 persistent=False)
+
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        input_dtype = hidden_states.dtype
+        hidden_states = hidden_states.to(torch.float32)
+
+        variance = hidden_states.pow(2).mean(-1, keepdim=True)
+        if self.allreduce:
+            variance = self.allreduce(variance) / self.mapping.tp_size
```
karljang: Do we need to compute sum/count instead of mean/tp_size, in case the hidden_size is not divisible by tp_size?

tensorrt_llm/_torch/visual_gen/models/flux/joint_proj.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def load_weights(self, weight_dict: Dict[str, torch.Tensor], loader: DynamicLinearWeightLoader):
+        """Load from checkpoint proj_out.weight / proj_out.bias.
+
+        At TP=1, passes through to the inner Linear.
+        At TP>1, splits the weight along the input dim (columns) into
+        attn and mlp portions, then loads each as a vanilla ROW Linear.
+        """
+        if self.tp_size <= 1:
+            loader.load_linear_weights(self.proj, "proj_out", [weight_dict])
```
karljang, line 124: Do we need to call self.proj.create_weight() first?

tensorrt_llm/_torch/visual_gen/models/flux/joint_proj.py
```diff
@@ -0,0 +1,283 @@
...
+
+        The checkpoint stores a single fused weight with layout:
+        [Q: q_dim | K: kv_dim | V: kv_dim | gate: mlp_hid | up: mlp_hid]
+
+        At TP=1: loads directly into self.proj.
+        At TP>1: splits into QKV and MLP portions, then loads into
+        column-parallel sub-Linears which handle per-rank sharding.
+        """
+        if self.tp_size <= 1:
+            loader.load_linear_weights(self.proj, "to_qkv_mlp_proj", [weight_dict])
```
karljang, line 260: Do we need to call create_weight() first?

tensorrt_llm/_torch/visual_gen/models/flux/joint_proj.py
```diff
@@ -0,0 +1,283 @@
...
+            ]
+            mlp_dicts = [
+                {"weight": W_gate, "bias": B_gate},
+                {"weight": W_up, "bias": B_up},
+            ]
+        else:
+            qkv_dicts = [{"weight": W_q}, {"weight": W_k}, {"weight": W_v}]
+            mlp_dicts = [{"weight": W_gate}, {"weight": W_up}]
+
+        loader.load_linear_weights(self.qkv_proj, "to_qkv_mlp_proj", qkv_dicts)
```
karljang, line 302: Do we need to call create_weight() first?

tensorrt_llm/_torch/visual_gen/models/flux/transformer_flux.py
```diff
@@ -922,6 +935,12 @@ def load_weights(self, weights: dict) -> None:
             if callable(getattr(module, "create_weights", None)):
                 module.create_weights()
 
+            # Wrapper modules have no direct _parameters; handle before the guard.
+            if isinstance(module, FluxJointAttnMLPProj):
+                module_weights = loader.filter_weights(name, weights)
+                module.load_weights(module_weights, loader)
```
karljang, line 900: Do we need managed_prefixes, similar to flux2?

karljang (commented):
Thank you for the PR. 
My AI suggests that we should test the TP=1 case. Could you please take a look?

tensorrt_llm/_torch/visual_gen/modules/rms_norm.py
belgarten-nv: Yes, that's better. Will fix.

tensorrt_llm/_torch/visual_gen/models/flux/transformer_flux.py
belgarten-nv: Yes, would be less fragile. Will add.

belgarten-nv:
@karljang 
> Thank you for the PR. My AI suggests that we should test the TP=1 case. Could you please take a look?

Specifically the Flux models or more generally? By visual inspection, the output of the Flux models is nearly identical between the original output before my modifications and any level of TP (tested 1,2,4,8). Is there a specific code path it is concerned about or just generally considering TP1 as an edge case that we should be careful with?

tensorrt_llm/_torch/visual_gen/modules/attention.py
```diff
@@ -139,6 +161,19 @@ def __init__(
...
+            backend_num_heads = backend_num_heads // tp_size
+
+            assert backend_num_kv_heads % tp_size == 0
+            backend_num_kv_heads = backend_num_kv_heads // tp_size
+
+            self.num_attention_heads //= tp_size
+            self.num_key_value_heads //= tp_size
+
+            self.q_dim = self.num_attention_heads * self.head_dim
+            self.kv_dim = self.num_key_value_heads * self.head_dim
```
NVShreyas: this should be moved before the `_init_qkv_proj` call. the `fused_weight_shard_indices_mapping` needs the correct dims

belgarten-nv: I think whether it is before or after is actually an issue in the Linear layer. See `load_weights_fused_qkv_helper`: `load_weight_shard()` takes in the indices mapping and computes the TP shard, but `copy_weight_shard()` just directly copies the weight over without any regard for TP. (These models use `allow_partial_loading=False`, which is how I've missed this.)

Thus if we moved this before `_init_qkv_proj`, the `load_weight_shard`, we would get the smaller `q_dim` and `kv_dim`, but this would again be divided by tp_size in `load_weight_shard()`, which is wrong. However, if we leave it here, we the `copy_weight_shard()` will try to copy all of QKV, even though we've only loaded one TP shard. This would crash.

The right answer here is probably to make `copy_weight_shard()` TP aware (ie, have it compute the same tp shard that the `load_weight_shard()` does), but I'm not sure  yet how wide the consequences of this would be.

belgarten-nv: No, I'm wrong. `copy_weight_shard()` is made TP aware via `fused_weight_shard_indices_mapping`, I can't directly move this code since I still need the actual weight sizes to be correct, but I do need to update the `fused_weight_shard_indices_mapping` to divide by the TP size for the offsets/sizes.

karljang:
> @karljangKJ
> 
> > Thank you for the PR. My AI suggests that we should test the TP=1 case. Could you please take a look?
> 
> Specifically the Flux models or more generally? By visual inspection, the output of the Flux models is nearly identical between the original output before my modifications and any level of TP (tested 1,2,4,8). Is there a specific code path it is concerned about or just generally considering TP1 as an edge case that we should be careful with?

Actually, my 3 comments about create_weight() are on the TP1 path. Those were identified as a block path due to a crash. Could you please double check?

tensorrt_llm/_torch/visual_gen/modules/attention.py
NVShreyas: > update the fused_weight_shard_indices_mapping to divide by the TP size for the offsets/sizes.

Yep, I added a temp fix for it on my cosmos3 branch. it was failing for nvfp4 weights

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -281,16 +284,20 @@ def __init__(
         )
 
         # Self-attention with fused QKV.
-        # fuse_qk_norm_rope=True: use fused cross-head QK Norm + RoPE CUDA kernel
+        # fuse_qk_norm_rope: use fused cross-head QK Norm + RoPE CUDA kernel
         # to eliminate extra global memory round-trip between separate norm and RoPE.
+        # Disabled at TP>1 because the fused kernel computes RMSNorm variance over
+        # the local head shard only (no all-reduce), producing incorrect cross-head
+        # normalization. The unfused path uses TP-aware RMSNorm with all-reduce.
+        tp_size = model_config.mapping.tp_size if model_config.mapping else 1
```
NVShreyas: do you know how much perf drops cause of this change? If its a decent amount, we can add it to our future optimizations list

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -1079,6 +1097,7 @@ def _init_preprocessors(self, cross_pe_max_pos):
             )
         elif self.model_type.is_video_enabled():
             self.video_args_preprocessor = TransformerArgsPreprocessor(
+                config=self.model_config,
```
NVShreyas: this should be `model_config`?

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -1093,6 +1112,7 @@ def _init_preprocessors(self, cross_pe_max_pos):
             )
         elif self.model_type.is_audio_enabled():
             self.audio_args_preprocessor = TransformerArgsPreprocessor(
+                config=self.model_config,
```
NVShreyas: same here

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
NVShreyas, line 402: these should be updated to local as well

NVShreyas: here too

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -142,9 +143,55 @@ def __init__(
         }
 
         if dist.is_initialized() and world_size > 1:
+            self.setup_communicators()
```
NVShreyas: can we call this method only when TP is enabled?

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -107,7 +107,7 @@ class DiffusionModelConfig(BaseModel):
     mapping: Mapping = PydanticField(default_factory=Mapping)
     skip_create_weights_in_init: bool = False
     force_dynamic_quantization: bool = False
-    allreduce_strategy: AllReduceStrategy = PydanticField(default=AllReduceStrategy.AUTO)
+    allreduce_strategy: AllReduceStrategy = PydanticField(default=AllReduceStrategy.NCCL)
```
NVShreyas, line 110: anything preventing us to use AUTO?

tensorrt_llm/_torch/visual_gen/models/flux/joint_proj.py
```diff
@@ -0,0 +1,288 @@
...
+                skip_create_weights_in_init=skip_create_weights_in_init,
+                force_dynamic_quantization=force_dynamic_quantization,
+                weights_loading_config=WeightsLoadingConfig(
+                    weight_mode=WeightMode.FUSED_QKV_LINEAR,
+                ),
+                fused_weight_shard_indices_mapping={
+                    "q": (0, q_dim),
+                    "k": (q_dim, kv_dim),
+                    "v": (q_dim + kv_dim, kv_dim),
+                },
```
NVShreyas, line 210: just to double check, these are pre-sharded buffers?

tensorrt_llm/_torch/visual_gen/models/flux/joint_proj.py
```diff
@@ -0,0 +1,288 @@
...
+                quant_config=quant_config,
+                skip_create_weights_in_init=skip_create_weights_in_init,
+                force_dynamic_quantization=force_dynamic_quantization,
+                weights_loading_config=WeightsLoadingConfig(
+                    weight_mode=WeightMode.FUSED_GATE_UP_LINEAR,
+                ),
+                fused_weight_shard_indices_mapping={
+                    "gate": (0, self.mlp_hidden_dim),
+                    "up": (self.mlp_hidden_dim, self.mlp_hidden_dim),
+                },
```
NVShreyas, line 230: same here

tensorrt_llm/_torch/visual_gen/models/flux/attention.py
```diff
@@ -303,29 +334,33 @@ def __init__(
             layer_idx=layer_idx,
         )
 
-        # Combined output: [q_dim + mlp_hidden_dim] -> [hidden_size]
-        self.to_out = Linear(
-            self.q_dim + self.mlp_hidden_dim,
-            hidden_size,
+        # Output projection needs FULL dims (ROW parallel divides internally)
+        self.to_out = FluxJointAttnMLPProj(
+            attn_dim=self.kv_dim,
```
NVShreyas: should this be q_dim?

tests/unittest/_torch/visual_gen/multi_gpu/test_tp_attention.py
```diff
@@ -0,0 +1,507 @@
+"""Multi-GPU tests for Tensor Parallel (TP) Attention.
```
NVShreyas, line 16: copyright missing

tensorrt_llm/_torch/visual_gen/models/flux/attention.py
belgarten-nv: Yes, I messed up my output shaped (although q_dim == kv_dim for flux, but will fix).

tensorrt_llm/_torch/visual_gen/models/flux/joint_proj.py
belgarten-nv: No, this is wrong, we need to divide by tp_size on these

tensorrt_llm/_torch/visual_gen/models/flux/joint_proj.py
belgarten-nv: ditto above

tensorrt_llm/_torch/visual_gen/config.py
belgarten-nv: The current code path requires Ray for AllReduce workspace configuration. We could have a version without Ray, but it would involve passing the VisualGenMapping (or a similar object) and having a custom Distributor overload for our setup instead. Its unclear right now how large the impact of that would be, but it seems to add many new code paths through the distributed module.

NVShreyas (approved):

tensorrt_llm/_torch/visual_gen/modules/rms_norm.py
```diff
@@ -0,0 +1,105 @@
...
+
+import torch
+from torch import nn
+
+from tensorrt_llm._torch.distributed import AllReduce
+from tensorrt_llm.functional import AllReduceStrategy
+from tensorrt_llm.mapping import Mapping
+
+
+class RMSNorm(nn.Module):
```
chang-l: Is this RMSNorm module intended only for TP usage, or is the intent to use it across all VisualGen models in the future?

chang-l (commented):
Thanks for your contribution! Could you also update the docs to reflect this feature? Specifically, `examples/visual_gen/README.md` and `docs/source/models/visual-generation.md`

tensorrt_llm/_torch/visual_gen/modules/rms_norm.py
belgarten-nv: This is intended only for TP support, it removes many of the custom kernels featured in the general TRTLLM RMSNorm since they are not TP compliant. If you are looking for strong perf and do not need TP, that is the better module to use.

chang-l: Thanks. Can we rename w/ TP-specific suffix/prefix or somehow better reflect its intended usage, so developers or users don’t mistakenly use this module in other places?

chang-l: Thanks for the update – it seems `RMSNormTPAware` module only makes sense for cross-head norms like Wan/LTX2. In future, it would be better to unify/extend this as VisualGen's RMSNorm module for all models
cc. @NVShreyas @karljang for viz

chang-l (approved):

docs/source/models/visual-generation.md
```diff
@@ -37,13 +37,13 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
...
-| **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| **Wan 2.2** | Yes | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| **LTX-2** | Yes | Yes | No | Yes | Yes | No | No | Yes | Yes | Yes | Yes |
+| Model | FP8 blockwise | NVFP4 | TeaCache | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve | Attention2D | Ring Attention | Tensor Parallelism |
+|---|---|---|---|---|---|---|---|---|---|--|--|--|
+| **FLUX.1** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes |
+| **FLUX.2** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes |
+| **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| **Wan 2.2** | Yes | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| **LTX-2** | Yes | Yes | No | Yes | Yes | No | No | Yes | Yes | Yes | Yes | No |
```
yibinl-nvidia, line 47: Not a blocking question, what's the reason ltx-2 doesn't support TP yet? Any changes we need to make?

belgarten-nv: LTX-2 TP support was attempted but I found it to be overly complex and have poor performance. The implementation involved changes all over the LTX transformer and the preprocessor args, and interrupts most of its optimizations. We can no longer use the fused attention path for any TP work since the fused RMSNorm kernels are not TP aware. Additionally TP introduces sync points in every RMSNorm, every Attention layer and every MLP. This causes very large overhead to the point where I measured almost a 3x *slowdown* when going from default to TP=8.

In theory you certainly could do it, but I found to be a much larger project than the other models.

belgarten-nv:
/bot skip --comment "2 failing tests from AutoDeploy. Both spurious failures that are unrelated to code changes. Will submit separate MR to waive them."

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
belgarten-nv: Efficiency for Wan2.2 tends from ~88% on 2 GPU to around 55% on 8 GPUs. Will investigate performance optimizations in a future MR.

belgarten-nv:
/bot skip --comment "2 failing tests from AutoDeploy. Both spurious failures that are unrelated to code changes. Will submit separate MR to waive them."

PR #13635, opened 2026-04-30
[TRTLLM-11319][feat] VisualGen public output API + bench timing decomposition

zhenhuaw-me:
## Description

Reworks the public output story of `tensorrt_llm.visual_gen` so callers get a coherent, request-aware return type with measurable engine-side timing — and reworks the offline bench so its `latency` and engine `generation` metrics line up with what the serving path does.

### Public output API

- `VisualGenOutput` replaces the raw `MediaOutput` as the user-facing return type. Carries image/video/audio tensors plus `request_id`, `error`, `frame_rate` / `audio_sample_rate`, and engine-side `VisualGenMetrics`.
- `VisualGenResult` is a single Future-like awaitable handle that resolves to `VisualGenOutput` (single prompt) or `List[VisualGenOutput]` (batch). Batched `generate(List[str])` fans out into per-item outputs with per-item error semantics.
- `VisualGenOutput.save(path)` is the single user-facing way to persist a generated tensor to disk (image PNG/JPG/WEBP, video MP4/AVI).
- Public exceptions: dropped `VisualGenError` / `VisualGenParamsError` in favor of standard `RuntimeError` / `ValueError` / `NotImplementedError`. Aligns with vLLM/SGLang convention where diffusion engines raise standard exception types and callers catch by built-in type, not by package-specific wrappers.

### Internal types and contracts

- Internal `MediaOutput` renamed to `PipelineOutput` and gains three CUDA-event timing phases (`pre_denoise` / `denoise` / `post_denoise`) plus an executor-measured `generation` field on the wire response.
- Tightened batched-shape contract: image is always `(B, H, W, C)`, video `(B, T, H, W, C)`, audio `(B, channels, T_audio)`. Asserts in `split_visual_gen_output` make a violating pipeline fail loudly rather than silently corrupt per-item outputs by indexing along the wrong dim. LTX-2's `decode_audio` no longer unconditionally `.squeeze(0)`s, restoring the contract for `B=1`.

### Module boundaries

- Tensor-encoding code extracted from `tensorrt_llm/serve/media_storage.py` into a new `tensorrt_llm/media/encoding.py` module of free functions; `MediaStorage` keeps its file-storage role under `serve/`.
- The eight video-generation route handlers in `openai_server.py` are extracted into a sibling mixin (`tensorrt_llm/serve/openai_video_routes.py`) so `openai_server.py` stays under the per-file line budget. Behavior is strictly preserved.

### Engine reliability

- Fixed a timeout-leak in `DiffusionRemoteClient`: previously, a request that timed out in `aresult()` could have its late-arriving response pinned in `completed_responses` for the process lifetime, leaking a full `PipelineOutput` (including video tensors) per timeout. New `abandon_request_id` flow uses an abandoned-id set so `_store_response` drops late responses and `abandon_request_id` pops cached ones; both orderings handled under the same async lock.

### Server endpoints

- `/v1/images/edits` short-circuits to HTTP 501 `NotImplementedError`. Reason: no in-tree pipeline implements image editing today (Flux/Flux2 are gen-only; Wan/LTX-2 produce video). The handler used to dispatch through the generator and 500 on a downstream `None` check; 501 is the honest answer until an edit-capable pipeline lands. Route stays registered so re-enabling is a one-line change.

### Benchmark (`tensorrt_llm/bench/benchmark/visual_gen.py`)

- Per-run persisted media dir: `_run_benchmark` wraps each run in one `tempfile.TemporaryDirectory` and saves each result into it. Mirrors the server's persisted-write pattern (no per-request `mkstemp` / `unlink` overhead in the timing window).
- ffmpeg-optional fallback via `resolve_video_format("auto")`, matching how the server treats `output_format="auto"`. The bench no longer hard-fails on hosts without ffmpeg.
- Timing units: dropped `_ms` suffix and store wall-clock seconds end-to-end. `CudaPhaseTimer.fill()` divides `Event.elapsed_time` by 1000 once at the boundary so every downstream type carries seconds.
- Renamed the engine-side metric from `pipeline` (a class-name leak) to `generation` (describes what the timer measures, not how). Threaded through `VisualGenMetrics`, `DiffusionResponse`, server log lines, bench per-request and aggregate types.
- `VisualGenBenchmarkMetrics` gains `mean / median / std / min / max / percentiles_generation` alongside the existing latency aggregates. The `latency >= generation` gap is the encode + persist + IPC overhead the bench measures around the engine, and the report exposes that headroom.

### API stability

YAML coverage for the VisualGen public surface is explicitly deferred until the API exits prototype status (all classes are `@set_api_status("prototype")`).

## Test Coverage

- `tests/unittest/visual_gen/test_output.py` — `VisualGenOutput` dataclass, `VisualGenMetrics`, `to_visual_gen_output` / `split_visual_gen_output` success and error paths, batch fan-out, `VisualGenOutput.save()` routing, `VisualGenResult` sync `result()`, async `aresult()`, `__await__`, **timeout-leak fix** (5 new tests covering `abandon_request_id`'s two race orderings via a minimal `DiffusionRemoteClient` stub).
- `tests/unittest/_torch/visual_gen/test_visual_gen_params.py` — request-validation surface migrated to plain `ValueError`.
- `tests/unittest/_torch/visual_gen/test_trtllm_serve_endpoints.py` — FastAPI mock tests; `TestImageEdit` reduced to a 501 assertion plus the existing pydantic-validation 400 test.
- `tests/unittest/media/test_encoding.py` — PNG round-trip, `video_to_bytes` AVI bytes, `resolve_video_format` dispatch, ffmpeg fallback paths.
- `tests/integration/defs/examples/test_visual_gen.py` — Wan T2V and Flux end-to-end public-output contract checks (single + batch + async), including error-path semantics. Field-name assertions updated for the timing rename.
- `tests/integration/defs/visual_gen/test_visual_gen_benchmark.py` — bench JSON-key assertions extended with `mean_generation`.

## Summary by CodeRabbit

* **New Features**
  * Added performance metrics (pipeline latency, denoising timing) to generation outputs via `VisualGenMetrics`
  * Introduced `VisualGenOutput.save()` method for easy output persistence
  * Added OpenAI-compatible video generation endpoints

* **Improvements**
  * Refactored output handling for cleaner result objects with error tracking
  * Consolidated media encoding functionality into dedicated utilities

* **Documentation**
  * Updated all visual generation examples to use new output API

## PR Checklist

## GitHub Bot Help


examples/visual_gen/models/wan_t2v.py
```diff
@@ -65,8 +64,8 @@ def main():
         params=params,
     )
 
-    # --- Model-specific: video output ---
-    MediaStorage.save_video(output.video, args.output_path, frame_rate=params.frame_rate)
+    # --- Model-specific: video output (frame_rate carried on the output) ---
```
zhenhuaw-me:
```suggestion
```

examples/visual_gen/visual_gen_ltx2.py
```diff
@@ -414,13 +413,9 @@ def main():
...
 
-        # Save Output
-        MediaStorage.save_video(
-            output.video,
-            args.output_path,
-            audio=output.audio,
-            frame_rate=args.frame_rate,
-        )
+        # Save Output (frame_rate carried on the output; pass kwarg here only
+        # if the caller wants to override the rate the pipeline emitted)
```
zhenhuaw-me:
```suggestion
```

tensorrt_llm/media/encoding.py
```diff
@@ -0,0 +1,591 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zhenhuaw-me, line 1: Note for reviewers: this is moved from `media_storage.py`, so should be safe to skip.

tensorrt_llm/media/encoding.py
```diff
@@ -0,0 +1,591 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""Tensor → image/video encoding free functions.
+
+Used directly by :class:`tensorrt_llm.visual_gen.VisualGenOutput.save` and by
+the OpenAI-compatible server. Not re-exported from ``tensorrt_llm`` to keep the
+public top-level surface focused on engine APIs.
+"""
```
zhenhuaw-me:
```suggestion
"""Tensor → image/video encoding free functions."""
```

tensorrt_llm/serve/media_storage.py
```diff
@@ -1,755 +1,20 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Server-side storage handler for generated media assets.
+
+The tensor-encoding routines (``save_image``, ``save_video``,
+``image_to_bytes``, ``video_to_bytes``, ``resolve_video_format``) live in
+:mod:`tensorrt_llm.media.encoding`. This module retains :class:`MediaStorage`
+for the server-side role of persisting media to a managed location alongside
+its metadata (the ``trtllm-serve`` ``/v1/videos`` storage path).
 """
```
zhenhuaw-me:
```suggestion
"""Server-side storage handler for generated media assets."""
```

tensorrt_llm/visual_gen/output.py
```diff
@@ -0,0 +1,207 @@
...
+    image: Optional[torch.Tensor] = None
+    video: Optional[torch.Tensor] = None
+    audio: Optional[torch.Tensor] = None
+    frame_rate: Optional[float] = None
+    audio_sample_rate: Optional[int] = None
+    error: Optional[str] = None
+    metrics: Optional[VisualGenMetrics] = None
+
+    @classmethod
+    def _from_response(cls, resp: "DiffusionResponse") -> "VisualGenOutput":
```
zhenhuaw-me: Does it make sense to let the Pipeline class (or any other similar class) define a util to convert the Request and PipelineOutput to VisualGenOutput?

tensorrt_llm/visual_gen/visual_gen.py
```diff
@@ -688,56 +736,71 @@ def generate(
...
-            VisualGenResult: Call result() to get output dict.
+        Raises:
+            ValueError: ``inputs`` is empty or contains non-strings.
+            NotImplementedError: ``params`` is a list.
         """
+        if isinstance(params, list):
+            raise NotImplementedError(
+                "Per-item params (List[VisualGenParams]) are not supported in this "
+                "release; pass a single VisualGenParams shared across the batch. "
+                "See §5.1.2 of the VisualGen API design doc for the rationale."
```
zhenhuaw-me:
```suggestion
                "release; pass a single VisualGenParams shared across the batch."
```

tests/integration/defs/examples/test_visual_gen.py
```diff
@@ -391,20 +390,20 @@ def _generate_ltx2_video(llm_venv, output_subdir, linear_type="default"):
...
-            params=params,
-        )
-        MediaStorage.save_video(
-            output.video,
-            output_path,
-            audio=output.audio,
-            frame_rate=LTX2_T2V_FRAME_RATE,
-        )
+        output = visual_gen.generate(inputs=LTX2_T2V_PROMPT, params=params)
+        # AC-1/9/10/11: public output contract on a successful LTX-2 run.
```
zhenhuaw-me:
```suggestion
```

tests/integration/defs/examples/test_visual_gen.py
```diff
@@ -480,20 +478,19 @@ def _generate_ltx2_two_stage_video(llm_venv, output_subdir, linear_type="default
...
-        )
-        MediaStorage.save_video(
-            output.video,
-            output_path,
-            audio=output.audio,
-            frame_rate=LTX2_T2V_FRAME_RATE,
-        )
+        output = visual_gen.generate(inputs=LTX2_T2V_PROMPT, params=params)
+        # AC-1/9/10/11: same contract holds on the two-stage LTX-2 path; the
+        # second stage is rolled into ``post_denoise_ms`` per the plan.
```
zhenhuaw-me:
```suggestion
```

tests/integration/defs/examples/test_visual_gen.py
```diff
@@ -817,3 +814,209 @@ def test_wan_t2v_example(_visual_gen_deps, llm_root, llm_venv):
...
+# =============================================================================
+# Public API contract tests
+#
+# These exercise the AC-1..AC-11 contracts on real model runs:
+#   - VisualGenOutput shape, error semantics, metrics population
+#   - frame_rate populated by Wan T2V (== 16.0 per plan)
+#   - executor-measured pipeline_ms > 0 (AC-10)
+#   - generate(List[str]) batch fan-out (AC-4)
+#   - NotImplementedError guard for params: List[VisualGenParams] (AC-4)
+#   - generate_async / await path (AC-3)
```
zhenhuaw-me:
```suggestion
```

tensorrt_llm/visual_gen/output.py
```diff
@@ -0,0 +1,207 @@
...
+        if self.audio is not None:
+            raise VisualGenError("Saving audio-only outputs is not supported in this release.")
+
+        raise VisualGenError(
+            f"Cannot save output: request {self.request_id} carries no media "
+            "(image/video/audio are all None)."
+        )
+
+
+def _split_visual_gen_output(resp: "DiffusionResponse", batch_size: int) -> List[VisualGenOutput]:
```
zhenhuaw-me: Does it make sense to move this to "internal impl" files?

tensorrt_llm/serve/openai_server.py
```diff
@@ -1874,14 +1878,19 @@ async def openai_image_edit(self, request: ImageEditRequest,
...
+                            image_to_bytes(image)).decode('utf-8'),
                         revised_prompt=request.prompt,
                     ) for image in output_images
                 ],
                 size=f"{params.width}x{params.height}",
             )
 
+            e2e_ms = (time.perf_counter() - image_edit_start) * 1000.0
+            logger.info(
+                f"Image {image_id} edited and encoded: "
```
chang-l: We don't support any image-edit model yet, right?

tensorrt_llm/serve/openai_server.py
```diff
@@ -1891,396 +1900,6 @@ async def openai_image_edit(self, request: ImageEditRequest,
                 err_type="InternalServerError",
                 status_code=HTTPStatus.INTERNAL_SERVER_ERROR)
 
-    async def openai_video_generation_sync(self,
```
chang-l, line 1901: @JunyiXu-nv Can you please take a look at the server endpoint changes?

tensorrt_llm/bench/benchmark/visual_gen.py
```diff
@@ -352,6 +352,33 @@ def _run_benchmark(
     return outputs
 
 
+def _save_to_tempfile(result) -> None:
+    """Persist ``result`` to a throwaway tempfile so e2e timing covers encoding.
```
chang-l: Suggest dropping `_save_to_tempfile()` here and not folding encoding into `e2e_ms`. Three reasons:

1. The server streams via `FileResponse` from the persisted output — it never tempfile-then-deletes. So this inflates `e2e_ms` with work the serving system does not do, plus a real FS round-trip that is non-trivial for fast pipelines.
2. Encoding is CPU-side and would overlap with the next request's GPU work in a real server, but the bench runs the save inline in the awaiting coroutine and does not overlap either — so today the bench neither matches serving nor models the overlap.
3. The `.mp4` suffix routes through ffmpeg unconditionally, so the bench now fails hard on hosts without ffmpeg, where previously it did no encoding at all. (`resolve_video_format("auto")` would have fallen back to `.avi` / pure-Python.)

If we want encoding latency in the report, maybe surface it as a separate `encode_ms` measured around in-memory `image_to_bytes` / `video_to_bytes` (or just not including video encoding time in our bench). In this way, it could also make our bench not depend on ffmpeg. What do you think?

tensorrt_llm/visual_gen/visual_gen.py
```diff
@@ -537,62 +537,115 @@ async def _wait_ready_async(self):
...
-        # Use run_coroutine_threadsafe to execute in the background thread's event loop
         future = asyncio.run_coroutine_threadsafe(
             self.executor.await_responses(self.request_id, timeout=timeout),
             self.executor._event_loop,
         )
-
-        # Await the future in the current event loop
         response = await asyncio.wrap_future(future)
 
         if response is None:
```
chang-l, line 610: Bug (captured by cc): timeout here orphans the entry in `executor.completed_responses`. `await_responses` returns `None` without popping the id, so when the worker eventually delivers the response, `_store_response` puts it in the dict and nothing consumes it — `aresult` has already cached the timeout sentinel with `_finished=True`. Each timed-out request leaks a full `PipelineOutput` tensor for the process lifetime.

Fix: pop `request_id` from `executor.completed_responses` on this branch (or extend `await_responses` with an abandon-on-timeout mode that drops late arrivals).

tensorrt_llm/_torch/visual_gen/output.py
```diff
@@ -1,29 +1,207 @@
...
+        pre_denoise_ms=out.pre_denoise_ms,
+        denoise_ms=out.denoise_ms,
+        post_denoise_ms=out.post_denoise_ms,
+    )
+    results: List["VisualGenOutput"] = []
+    for i in range(batch_size):
+        results.append(
+            VisualGenOutput(
+                request_id=resp.request_id,
+                image=out.image[i] if out.image is not None else None,
```
chang-l, line 230: Bug(captured by cc): Slicing `[i]` assumes a leading batch dim of size `batch_size`, but `PipelineOutput`'s docstring permits both `(H, W, C)` and `(B, H, W, C)`. A pipeline returning the unbatched form would silently index along H and produce corrupted per-item outputs. Latent today (every in-tree pipeline returns batched), but worth an `assert out.image.shape[0] == batch_size` (same for `video`/`audio`) so it fails loudly when the contract is violated.

tensorrt_llm/bench/benchmark/visual_gen.py
```diff
@@ -352,6 +352,33 @@ def _run_benchmark(
     return outputs
 
 
+def _save_to_tempfile(result) -> None:
```
chang-l: Two practical concerns layered on the metric question above:

1. The `.mp4` suffix routes through ffmpeg unconditionally, so on a host without ffmpeg installed the bench now fails hard where it previously did no encoding at all. Consider `resolve_video_format("auto")` (falls back to `.avi` / pure-Python) so the bench remains runnable, or document the new dependency.
2. Even with ffmpeg available, the tempfile write-then-unlink is real disk I/O — for fast pipelines the FS round-trip is non-trivial relative to the engine time we are trying to measure, and it inflates `e2e_ms` with something the production server would never do (the server streams via `FileResponse` from the persisted output path, but never tempfile-then-delete).

If we *do* want encoding latency in the metric, in-memory `image_to_bytes` / `video_to_bytes` is a closer proxy without the FS round-trip; if we *do not*, then this whole helper goes away.

tensorrt_llm/serve/openai_server.py
```diff
@@ -1874,14 +1878,19 @@ async def openai_image_edit(self, request: ImageEditRequest,
...
-                                'utf-8'),
+                            image_to_bytes(image)).decode('utf-8'),
                         revised_prompt=request.prompt,
                     ) for image in output_images
                 ],
                 size=f"{params.width}x{params.height}",
             )
 
+            e2e_ms = (time.perf_counter() - image_edit_start) * 1000.0
+            logger.info(
```
chang-l: Following up on the image-edit-model question above: no in-tree pipeline implements image-edit `infer()` (Flux/Flux2 are gen-only, Wan/LTX-2 are video), and `parse_visual_gen_params` does not pass `ImageEditRequest.image` to any of them. So this handler 500s on every call. Suggest dropping both the route registration (`/v1/images/edits`) and this method until an edit-capable pipeline lands — a 404 is better than a guaranteed 500.

tensorrt_llm/visual_gen/output.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+
+        if self.image is not None:
+            saved = save_image(self.image, path, format=format, quality=quality)
+            return Path(saved)
+
+        if self.video is not None:
+            fr = frame_rate if frame_rate is not None else self.frame_rate
+            if fr is None:
+                raise VisualGenError(
```
chang-l: Missing `frame_rate` is a caller usage error, not an engine failure — `ValueError` matches the precedent `VisualGenParamsError(ValueError)` set elsewhere in the package and reads cleaner for callers writing `try/except ValueError`. Same for the no-media branch (line 133) and the audio-only branch (line 131).

tensorrt_llm/visual_gen/output.py
```diff
@@ -0,0 +1,136 @@
...
+        Returns:
+            :class:`pathlib.Path` of the saved file.
+
+        Raises:
+            VisualGenError: When the output carries an error, when no media
+                tensor is present, or when video output lacks a frame rate.
+        """
+        # Lazy imports keep the encoding stack and the engine error type out
+        # of import time for users who only construct outputs.
+        from tensorrt_llm.media.encoding import save_image, save_video
```
chang-l, line 109: Nit: the rationale is off — by the time anyone has `VisualGenOutput`, `tensorrt_llm.visual_gen.__init__` has already eagerly imported `visual_gen.py` (where `VisualGenError` lives). The actual reason this import has to be lazy is the cycle: `visual_gen.py` imports `VisualGenOutput` at module top, so `output.py` cannot import `visual_gen` at module top. Worth fixing the comment so a future reader does not move it back to module scope and break the cycle.

tensorrt_llm/serve/openai_server.py
zhenhuaw-me: No we don't. Fixed typo.

tensorrt_llm/serve/openai_server.py
zhenhuaw-me: These are moved to a separate source file `tensorrt_llm/serve/openai_video_routes.py`

tensorrt_llm/visual_gen/visual_gen.py
zhenhuaw-me: Fixed in 35c880f42a. Added `_abandoned_request_ids: Set[int]` plus an async `abandon_request_id(rid)` on `DiffusionRemoteClient` that pops `completed_responses[rid]` and inserts into the abandoned set under the lock. `_store_response` checks the set on insert and discards both the response and the set entry. The `aresult` timeout branch schedules `abandon_request_id` on the executor's event loop before flipping `_finished`, so both orderings (response arrived between break and abandon, or after) are covered under the same async lock. Five unit tests in `tests/unittest/visual_gen/test_output.py` exercise the contract end-to-end.

tensorrt_llm/_torch/visual_gen/output.py
zhenhuaw-me: The docstring was wrong... Fixed in 35c880f42a.

tensorrt_llm/serve/openai_server.py
zhenhuaw-me: Agree on the diagnosis. make the endpoint to return 501 - keep the endpoint thus the change can be focus on the pipeline implementation when enabling the "image-edit" functionality.

tensorrt_llm/visual_gen/output.py
zhenhuaw-me: Agree, and went broader than just this site after surveying vLLM and SGLang. Both libraries strongly prefer standard exceptions: vLLM has only one thin `VLLMValidationError(ValueError)` for OpenAI-style structured errors; SGLang's `multimodal_gen` uses plain `ValueError` / `RuntimeError` / `NotImplementedError` everywhere. Fixed in 35c880f42a by dropping both `VisualGenError` and `VisualGenParamsError` and removing the exports from `tensorrt_llm/__init__.py` and `tensorrt_llm/visual_gen/__init__.py`. Replacement map in `output.py.save()`:
- errored output → `RuntimeError`
- missing `frame_rate` → `ValueError`
- audio-only → `NotImplementedError`
- no media → `ValueError`

`aresult` single-prompt error path → `RuntimeError`. `_validate_request` → plain `ValueError`. Both classes were `@set_api_status("prototype")` so removal is allowed; tests updated.

tensorrt_llm/visual_gen/output.py
zhenhuaw-me: Fixed in 35c880f42a. Dropped all `VisualGen*Error` to prefer "standard python exceptions".

tensorrt_llm/bench/benchmark/visual_gen.py
zhenhuaw-me: Update the benchmarking to have per-run tempdir. The tempdir/encoding in trtllm-bench part is to be aligned with trtllm-serve where the storage/encoding are included.

We now have 2 metrics reported in benchmarking:
* `latency`: the e2e HTTP latency including storage/encoding.
* `generation`: the Python API latency.

I didn't add a `encode` yet since:
1. we can chose either `latency` or `generation` as the metric depending on the analysis target.
2. We are unlikely to optimize `encode` since it's in FFmpeg.
3. If we introduce some inference/encoding overlapping in trtllm-serve, it would be tricky to define `encode` since `encode + generation < latency` is possible.

Does this make sense?

tensorrt_llm/serve/media_storage.py
```diff
@@ -1,755 +1,13 @@
 #!/usr/bin/env python
```
JunyiXu-nv: Could we completely remove this file?

tensorrt_llm/serve/openai_video_routes.py
```diff
@@ -0,0 +1,493 @@
...
+            sync_video_start = time.perf_counter()
+            output = self.generator.generate(inputs=request.prompt, params=params)
+            if output.video is None:
+                return self.create_error_response(
+                    message="Video generation failed",
+                    err_type="InternalServerError",
+                    status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+                )
+
+            actual_output_path = save_video(
```
JunyiXu-nv: Could we move to the new `save` API usage?

JunyiXu-nv: Also for images.

tests/unittest/media/test_encoding.py
```diff
@@ -0,0 +1,161 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
JunyiXu-nv, line 1: Is this test included in any test list?

JunyiXu-nv (commented):
Left some questsions.

tensorrt_llm/serve/media_storage.py
zhenhuaw-me: Yeah, makes sense — only the empty stub was left. Removed.

tensorrt_llm/serve/openai_video_routes.py
zhenhuaw-me: Done — both video routes and the image URL path now go through `output.save`.

tests/unittest/media/test_encoding.py
zhenhuaw-me: It wasn't. Added both `test_encoding.py` and `test_output.py` to the list, and replaced a stale entry that was still pointing at the deleted `test_media_storage.py` while I was there.

tensorrt_llm/bench/benchmark/visual_gen.py
chang-l: Yes, this makes sense to me. I’m not trying to push inference/encoding overlap into trtllm-serve right now. My main point is that encoding is mostly CPU/FFmpeg work, and in a real serving setup it can often be hidden by an overlapping runner or async launch. So keeping both latency and generation sounds reasonable, but we should be very careful with exposed `latency` metric: for GPU-centric metrics like average TTFF/FPS/GPU, including save/encoding time can be misleading since that is usually not the part we care about.

chang-l (approved):

JunyiXu-nv (approved):

tensorrt_llm/_torch/visual_gen/output.py
```diff
@@ -1,29 +1,238 @@
...
+        # syncing the last event also covers the earlier ones (same stream).
+        self._end.synchronize()
+        ms_to_s = 1.0 / 1000.0
+        output.pre_denoise = float(self._pre_start.elapsed_time(self._denoise_start)) * ms_to_s
+        output.denoise = float(self._denoise_start.elapsed_time(self._post_start)) * ms_to_s
+        output.post_denoise = float(self._post_start.elapsed_time(self._end)) * ms_to_s
+        return output
+
+
+def to_visual_gen_output(resp: "DiffusionResponse") -> "VisualGenOutput":
```
QiJune, line 154: `to_visual_gen_output()` and `split_visual_gen_output()` feel misplaced in tensorrt_llm/_torch/visual_gen/output.py.

Both helpers convert an internal IPC/pipeline representation (DiffusionResponse / PipelineOutput) into the public VisualGenOutput API type. That is client/public API adaptation logic, not internal _torch pipeline output logic. Keeping them here makes the _torch layer know about the public visual_gen API, which reverses the intended dependency direction.

I suggest moving both helpers to tensorrt_llm/visual_gen/output.py, next to VisualGenOutput, and keeping _torch/visual_gen/output.py focused on PipelineOutput and CudaPhaseTimer.

tensorrt_llm/visual_gen/visual_gen.py
```diff
@@ -537,62 +548,123 @@ async def _wait_ready_async(self):
...
-            raise VisualGenError(f"Generation failed: {response.error_msg}")
+            return self._resolved_value()
 
-        self._result = response.output
+        self._resolved = self._build_resolved(response)
         self._finished = True
-        return self._result
+        return self._resolved_value()
+
+    def result(self, timeout: Optional[float] = None):
```
QiJune, line 638: result(timeout=...) applies the same timeout twice:

```python
future = asyncio.run_coroutine_threadsafe(self.aresult(timeout=timeout), loop)
return future.result(timeout=timeout)
```

The inner timeout is the generation-response timeout and runs cleanup via abandon_request_id(). The outer timeout is a cross-thread wait timeout. At the timeout boundary, the outer future.result() can raise before the inner coroutine finishes cleanup, so late large outputs may still leak into completed_responses.

I’d use the timeout only inside aresult():

```python
future = asyncio.run_coroutine_threadsafe(self.aresult(timeout=timeout), loop)
return future.result()
```

tensorrt_llm/_torch/visual_gen/output.py
zhenhuaw-me: This placement is deliberate, mirroring the C++ DTO convention: the public `VisualGenOutput` is a passive dataclass that knows nothing about `_torch`, and the internal layer is responsible for emitting it. That keeps the public surface stable — it can be imported without pulling in any pipeline implementation, and changes to `DiffusionResponse` / `PipelineOutput` never force edits to `tensorrt_llm/visual_gen/output.py`.

Note also that the only public→internal coupling at module load time is already absent: the `from tensorrt_llm.visual_gen.output import VisualGenMetrics, VisualGenOutput` lines inside `to_visual_gen_output` / `split_visual_gen_output` are function-local, so loading `_torch/visual_gen/output.py` does not import the public module. Inverting the direction would push `DiffusionResponse` / `PipelineOutput` knowledge into the public file, which is the trade I'd rather not make.

Happy to discuss further if you'd still prefer the flip — just wanted to lay out the reasoning before changing it.

tensorrt_llm/bench/benchmark/visual_gen.py
zhenhuaw-me: Agreed — the two metrics are intentionally separated for exactly that reason. `generation` is the GPU-centric figure (Python-API time around the engine call), while `latency` covers the full HTTP-equivalent path including encoding/persistence. Anyone deriving TTFF/FPS/SOL on top of `latency` would indeed mislead themselves; the right input there is `generation` (or its sub-phases). Worth noting in a future bench-results doc once we have one.

tensorrt_llm/_torch/visual_gen/output.py
zhenhuaw-me: This placement is deliberate, mirroring the C++ DTO convention: the public `VisualGenOutput` is a passive dataclass that knows nothing about `_torch`, and the internal layer is responsible for emitting it. That keeps the public surface stable — it can be imported without pulling in any pipeline implementation, and changes to `DiffusionResponse` / `PipelineOutput` never force edits to `tensorrt_llm/visual_gen/output.py`.

Note also that the only public→internal coupling at module load time is already absent: the `from tensorrt_llm.visual_gen.output import VisualGenMetrics, VisualGenOutput` lines inside `to_visual_gen_output` / `split_visual_gen_output` are function-local, so loading `_torch/visual_gen/output.py` does not import the public module. Inverting the direction would push `DiffusionResponse` / `PipelineOutput` knowledge into the public file, which is the trade I'd rather not make.

Happy to discuss further if you'd still prefer the flip — just wanted to lay out the reasoning before changing it.

tensorrt_llm/visual_gen/visual_gen.py
zhenhuaw-me: Good catch — fixed. Dropped the outer timeout so only `aresult` enforces it; that keeps the cleanup path (`abandon_request_id` + resolved error output) atomic and prevents a late-arriving response from leaking into `completed_responses`.

QiJune (approved):
LGTM

kaiyux (approved):

brb-nv (approved):
LGTM.

PR #13649, opened 2026-04-30
[None][feat] Emit per-rank Attention-DP iteration stats

indrajit96:
## Release Notes

* **Bug Fixes**
  * Improved accuracy of distributed inference metrics by excluding placeholder requests from request counters.

* **Improvements**
  * Enhanced per-iteration statistics delivery and synchronization across distributed ranks, ensuring statistics are aggregated only when all ranks are properly aligned.
  * Refined handling of unmatched iterations to prevent mixed-iteration statistics from being reported.

## Description

Attention-DP previously exposed rank-local iteration stats from rank 0 only. That undercounted scheduled and paused work when multiple ADP ranks were active, and dummy Attention-DP padding requests could appear in planner-facing metrics.

This PR makes rank 0 publish one `IterationStats` record per Attention-DP rank for each aligned Attention-DP iteration. Each rank piggybacks a small rank-local stats payload on the existing ADP rank-state allgather. Rank 0 waits until all ranks report the same iteration, then emits per-rank records tagged with `attentionDpRank=i`. Scheduled and paused fields stay rank-local so downstream consumers can see load distribution across ADP ranks. Engine-global queued fields stay on the rank 0 record only because the executor request queue is owned by rank 0. Dummy requests, including Attention-DP and CUDA graph dummies, are filtered out of request counts and KV-token metrics.

 ### Attention-DP stats flow

  ```mermaid
  flowchart TD
      A["Forward iteration N completes"] --> B["_process_iter_stats<br/>existing"]
      B --> C["_update_iter_stats<br/>existing, updated"]
      C --> D{"Attention-DP enabled"}

      D -- "No" --> E["_append_iter_stats<br/>existing, updated"]

      D -- "Yes" --> F["_queue_adp_iter_stats<br/>new"]
      F --> G["_make_adp_iter_stats_payload<br/>new"]
      G --> H["_pending_adp_iter_stats_payloads<br/>new state"]

      F --> I{"rank 0"}
      I -- "Yes" --> J["_pending_adp_iter_stats / req_stats / kv_iter_stats<br/>new state"]
      I -- "No" --> K["Payload only"]

      H --> L["Next scheduler pass"]
      J --> L
      K --> L

      L --> M["_next_adp_iter_stats_payload<br/>new"]
      M --> N["gather_all_rank_states<br/>existing, updated"]
      N --> O["all_rank_states"]

      O --> P["_finalize_pending_adp_iter_stats<br/>new"]
      P --> Q["Use rank 0 iter_id as target"]
      Q --> R["_drop_pending_adp_iter_stats_before<br/>new"]
      R --> S["_ensure_adp_zero_iter_stats_payload<br/>new"]
      S --> T{"All ranks match target iter"}

      T -- "No" --> U["Wait for next allgather"]
      U --> L

      T -- "Yes" --> V{"rank 0"}
      V -- "Yes" --> X["For each matching rank_state<br/>sorted by Attention-DP rank"]
      X --> Y["_make_adp_rank_iter_stats<br/>new"]
      Y --> Z["_append_iter_stats<br/>existing, updated<br/>tagged with attention_dp_rank"]
      Z --> AA["_clear_pending_adp_iter_stats_through<br/>new"]
      classDef newFunc fill:#dcfce7,stroke:#15803d,color:#14532d;
      classDef updatedFunc fill:#dbeafe,stroke:#1d4ed8,color:#1e3a8a;
      classDef newState fill:#fef3c7,stroke:#b45309,color:#78350f;
      classDef decision fill:#f3f4f6,stroke:#6b7280,color:#111827;

      class F,G,M,P,R,S,W,Y,AA newFunc;
      class B,C,E,N,Z updatedFunc;
      class H,J newState;
      class D,I,T,V decision;
  ```
  
  Legend:
  - Green: newly added helper functions
  - Blue: existing functions updated by this PR
  - Yellow: newly added pending-state structures

## Test Coverage

  - `tests/unittest/_torch/executor/test_adp_router.py`
    - Verifies `RankState` serialization includes the iter-stats payload.
    - Verifies ADP rank-state allgather piggybacks pending iter-stats payloads.

  - `tests/unittest/pyexecutor/test_iter_stats_populate.py`
    - Verifies Attention-DP and CUDA graph dummy requests are excluded from count and KV-token metrics.
    - Verifies rank-local ADP stats are emitted as one row per Attention-DP rank.
    - Verifies per-rank fanout waits for complete matching iteration payloads.
    - Verifies multiple pending payloads are processed oldest-first.
    - Verifies non-rank0 ranks align to rank 0’s target iteration with zero payloads when needed.
    - Verifies queued/request/KV stats remain rank-0-owned while scheduled fields are rank-local.

  - `tests/unittest/executor/test_stats_serializer.py`
    - Verifies ADP stats tuples serialize the supplied `attentionDpRank`.

### Manual validation on H100x2:
  - Ran targeted ADP unit tests:
    - `tests/unittest/_torch/executor/test_adp_router.py`
    - `tests/unittest/pyexecutor/test_iter_stats_populate.py`
  - Built TRT-LLM wheel from patched commit `164d91c61669a540a56b2411eb388d046244f15e`.
  - Ran Attention-DP smoke test with `tensor_parallel_size=2`, `enable_attention_dp=True`.
  - Verified Dynamo forward-pass metrics event plane observed planner-visible scheduled/queued metrics under Attention-DP.

  ## Validation Evidence

  Manual H100x2 validation used `tp_size=2`, `enable_attention_dp=True`, 8 prompts, and `TinyLlama/TinyLlama-1.1B-Chat-v1.0`.

  ### Before vs After TRT-LLM IterationStats

 Before this PR, rank 0 emitted only rank-local scheduled work. The previous aggregate design fixed total counts but collapsed the per-rank distribution. 
 After This PR emits one row per Attention-DP rank so consumers can see both total work and load balance across ranks.
 
  | Metric | Before | After |
  |---|---:|---:|
  | `maxContextRequests` | 3 | 7 |
  | `maxGenRequests` | 4 | 8 |
  | `sumContextRequests` | 4 | 8 |
  | `sumGenRequests` | 28 | 55 |
  | `sumScheduledRequests` | 32 | 63 |
  | `attentionDpRank` | `null` | `0` |

Below is before/after output derived from `stats = llm.get_stats(timeout=10)` for the same H100x2 Attention-DP workload.

### Before :

  ```json
  {"summary": {"maxContextRequests": 3, "maxGenRequests": 4, "records": 9, "sumContextRequests": 4, "sumGenRequests": 28, "sumScheduledRequests": 32}}
  {"attentionDpRank": null, "iter": 2, "numContextRequests": 0, "numGenRequests": 4, "numScheduledRequests": 4}
  {"attentionDpRank": null, "iter": 3, "numContextRequests": 0, "numGenRequests": 4, "numScheduledRequests": 4}
```

### After :

```json
  {"attentionDpRank": 0, "iterLatencyMS": 12.5, "inflightBatchingStats": {"numContextRequests": 1, "numCtxTokens": 100, "numCtxKvTokens": 10, "numGenRequests": 2, "numGenKvTokens": 20, "numPausedRequests": 1, "numPausedKvTokens": 5, "numQueuedContextRequests": 7, "numQueuedCtxTokens": 700, "numQueuedGenRequests": 8, "numQueuedGenKvTokens": 800, "numScheduledRequests": 3}}
  {"attentionDpRank": 1, "iterLatencyMS": 12.5, "inflightBatchingStats": {"numContextRequests": 3, "numCtxTokens": 300, "numCtxKvTokens": 30, "numGenRequests": 4, "numGenKvTokens": 40, "numPausedRequests": 2, "numPausedKvTokens": 25, "numQueuedContextRequests": 0, "numQueuedCtxTokens": 0, "numQueuedGenRequests": 0, "numQueuedGenKvTokens": 0, "numScheduledRequests": 7}}
```

  Interpretation:

  - Before: steady decode reported only rank 0’s local work, so a 2-rank Attention-DP engine could underreport scheduled decode work.
  - After: TRT-LLM emits one stats row per Attention-DP rank, tagged with `attentionDpRank`.
  - After: scheduled fields remain rank-local, so downstream consumers can see the per-rank load split instead of only an aggregate total.
  - After: total scheduled work is still recoverable by summing the per-rank rows.
  - After: queued fields remain nonzero only on rank 0, so queue pressure is not double-counted across ADP ranks.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1339,14 +1377,180 @@ def _update_iter_stats(self, stats, iter_latency_ms, num_completed_requests,
...
+        self._pending_adp_iter_stats_payloads[iter_id] = payload
+        self._synthetic_adp_iter_stats_iters.discard(iter_id)
+
+        if self.dist.rank == 0:
+            self._pending_adp_iter_stats[iter_id] = stats
+            self._pending_adp_req_stats[iter_id] = req_stats
+            # Capture KV iteration stats with this IterationStats object.
+            # ADP aggregation happens on a later allgather, when
+            # _latest_kv_iter_stats may already refer to a newer iteration.
+            self._pending_adp_kv_iter_stats[iter_id] = self._latest_kv_iter_stats
```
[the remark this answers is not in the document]
indrajit96: That's ok and expected outof scope for this PR. Added comments to clearify the behaviour

tests/unittest/pyexecutor/test_iter_stats_populate.py
```diff
@@ -457,6 +503,210 @@ def test_num_ctx_kv_tokens_ignores_iter_states_side_channel():
...
+    stats.iter = iter_id
+    stats.inflight_batching_stats = ifb
+    return stats
+
+
+def _build_adp_stats_harness(pending_stats, *, rank=0, tp_rank=0):
+    from tensorrt_llm._torch.pyexecutor.py_executor import PyExecutor
+
+    harness = _AdpStatsHarness()
+    harness.dist = types.SimpleNamespace(rank=rank, tp_rank=tp_rank)
```
achartier: Missing `harness.enable_attention_dp = True`

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
```diff
@@ -48,15 +88,48 @@ class RankState:
...
+            self.num_active_tokens,
+            *self.iter_stats.serialize(),
+        ]
 
     @classmethod
     def deserialize(cls, data: list[int]) -> RankState:
         """Deserialize from a flat list received via allgather."""
-        return cls(*data)
+        values = list(data)
+        rank_state_fields = fields(cls)[:3]
```
brb-nv: Will be nice to give the value 3 a meaningful variable name so that the usage below becomes easier to follow. Also, this could break easily if we add another field in the future?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1177,11 +1195,33 @@ def _update_iter_stats(self, stats, iter_latency_ms, num_completed_requests,
...
-        stats.inflight_batching_stats.num_paused_requests = len(
-            scheduled_batch.paused_requests)
+        # Attention-DP may add dummy requests to keep ranks aligned during
+        # distributed scheduling. Those placeholders are not user work, so
+        # count the request lists with a filter instead of using the cached
+        # batch counters, which include every scheduled item.
+        if getattr(self, "enable_attention_dp", False):
+            # The sum(...) scans are small and keep planner-facing metrics
+            # from treating Attention-DP dummy requests as real load.
+            num_context_requests = sum(
```
brb-nv, line 1240: Do we need to account for cudagraph dummies also? Or is that accounted for elsewhere?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1339,14 +1379,184 @@ def _update_iter_stats(self, stats, iter_latency_ms, num_completed_requests,
...
+    def _next_adp_iter_stats_payload(self) -> Optional[RankIterStatsPayload]:
+        """All ranks: return the oldest pending stats payload to piggyback."""
+        if not self._pending_adp_iter_stats_payloads:
+            return None
+        iter_id = min(self._pending_adp_iter_stats_payloads)
+        return self._pending_adp_iter_stats_payloads[iter_id]
+
+    def _ensure_adp_zero_iter_stats_payload(self, iter_id: int) -> None:
+        """All ranks: add a zero payload when this rank had no work for iter."""
+        if iter_id in self._pending_adp_iter_stats_payloads:
```
brb-nv: When would this happen? Is this when concurrency < dpSize?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1339,14 +1379,184 @@ def _update_iter_stats(self, stats, iter_latency_ms, num_completed_requests,
...
+            self._pending_adp_iter_stats.pop(pending_iter, None)
+            self._pending_adp_req_stats.pop(pending_iter, None)
+            self._pending_adp_kv_iter_stats.pop(pending_iter, None)
+
+    def _clear_pending_adp_iter_stats_through(self, iter_id: int) -> None:
+        """All ranks: clear local ADP stats once aggregation has caught up."""
+        for pending_iter in list(self._pending_adp_iter_stats_payloads):
+            if pending_iter > iter_id:
+                continue
+            self._pending_adp_iter_stats_payloads.pop(pending_iter, None)
```
brb-nv: Nit: Any chance we can remove code duplication between `_drop_pending_adp_iter_stats_before` and `_clear_pending_adp_iter_stats_through`?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2867,8 +3081,13 @@ def _fetch_new_requests(
             # (e.g. KV-cache-aware) that need new_requests to gather additional
             # info, the allgather position may need to be revisited.
 
+            # Piggyback this rank's oldest pending IterationStats payload on
+            # the ADP rank-state allgather, then use the gathered states below
+            # to aggregate/clear stats once every rank is aligned.
             all_rank_states = self.adp_router.gather_all_rank_states(
```
brb-nv, line 3120: Dumb question: Is this done unconditionally today? I was expecting to see it conditioned on something like "if collect_iter_stats".

brb-nv (commented):
Mostly informational questions.

tests/unittest/pyexecutor/test_iter_stats_populate.py
indrajit96: Nice catch fixed!

tensorrt_llm/_torch/pyexecutor/py_executor.py
indrajit96: " Is this done unconditionally today?"
Yes the ADP rank-state allgather itself is unconditional because routing needs the per-rank active request counts every iteration. I updated the stats-specific part to be gated on `enable_iter_perf_stats`: we only piggyback an iter-stats payload and call `_finalize_pending_adp_iter_stats(...)` when iteration stats are enabled.

tensorrt_llm/_torch/pyexecutor/py_executor.py
indrajit96: Done. I kept the two methods separate because they have different range semantics (`< iter_id` vs `<= iter_id`), 
But refactored the repeated pending-state cleanup into `_discard_pending_adp_iter_stats(...)`.

tensorrt_llm/_torch/pyexecutor/py_executor.py
indrajit96: Yes, that is one case. This happens when rank 0 has an iteration stats payload to publish, but another ADP rank has no real scheduled work for that same iteration.
concurrency < dpSize is a reasonable example

tensorrt_llm/_torch/pyexecutor/py_executor.py
indrajit96: Nice catch! 

Updted to use a generic dummy filter in `_update_iter_stats`. It now excludes ADP dummies, CUDA graph dummies, and other dummy requests

  I also updated the unit coverage to include `is_cuda_graph_dummy=True` cases, so the count fields and KV-token fields both verify that CUDA graph dummies are not reported as real load.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1339,14 +1386,244 @@ def _update_iter_stats(self, stats, iter_latency_ms, num_completed_requests,
...
+            return
+
+        if self.dist.rank == 0:
+            rank0_stats = self._pending_adp_iter_stats.get(iter_stats_iter)
+            if rank0_stats is None:
+                logger.warning(
+                    "Skipping attention-DP IterationStats fanout on "
+                    f"rank 0: pending IterationStats object is missing for "
+                    f"iter {iter_stats_iter}")
+                return
```
achartier: Only rank 0 will early return here. It should call `_clear_pending_adp_iter_stats_through` to be in-sync with other ranks.

indrajit96: Great find!
Fixed it!

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -551,6 +551,23 @@ def on_detected():
...
+        # not real measured local stats.
+        self._synthetic_adp_iter_stats_iters: Set[int] = set()
+        # Rank 0 only: full IterationStats objects waiting for ADP fanout.
+        self._pending_adp_iter_stats: Dict[int, IterationStats] = {}
+        # Rank 0 only: per-request stats preserved for compatibility.
+        # RequestStats remain rank-0-owned under Attention-DP.
+        self._pending_adp_req_stats: Dict[int,
+                                          Optional[List[RequestStats]]] = {}
+        # Rank 0 only: KV iteration stats captured with pending IterationStats.
+        self._pending_adp_kv_iter_stats: Dict[int, object] = {}
```
pcastonguay: Could all of those be part of a separate class for better encapsulation? Maybe some of the functions added in PyExecutor could also be part of that class (diff file too maybe?) to make the code easier to read.

indrajit96: Great idea. I refactored this into a separate ADPIterStatsBuffer helper in `tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py`
That class now owns the pending ADP iteration-stats

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1185,11 +1187,41 @@ def _update_iter_stats(self, stats, iter_latency_ms, num_completed_requests,
...
                 self._latest_kv_iter_stats = None
 
-        stats.inflight_batching_stats.num_context_requests = scheduled_batch.num_context_requests
-        stats.inflight_batching_stats.num_gen_requests = scheduled_batch.num_generation_requests
-        stats.inflight_batching_stats.num_scheduled_requests = stats.inflight_batching_stats.num_context_requests + stats.inflight_batching_stats.num_gen_requests
-        stats.inflight_batching_stats.num_paused_requests = len(
-            scheduled_batch.paused_requests)
+        def is_stats_dummy_request(req) -> bool:
+            return bool(
+                getattr(req, "is_dummy", False)
```
achartier: `is_dummy` is already a property combining the other 3 fields. Checking them again is overkill imo.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1435,7 +1464,10 @@ def _append_iter_stats(self,
         with self.stats_lock:
             if len(self.stats) > self.max_stats_len:
                 self.stats.pop(0)
-            self.stats.append((stats, req_stats, self._latest_kv_iter_stats))
+            entry = (stats, req_stats, kv_iter_stats)
+            if attention_dp_rank is not None:
+                entry = (stats, req_stats, kv_iter_stats, attention_dp_rank)
```
achartier: Nit: I think it'd be better to always include `attention_dp_rank` and check if it's None inside `_stats_serializer` rather than rely on the size of the tuple.

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
```diff
@@ -48,15 +88,49 @@ class RankState:
     rank: int
     num_active_requests: int = 0
     num_active_tokens: int = 0
+    iter_stats: RankIterStatsPayload = field(default_factory=RankIterStatsPayload)
+
+    def copy_iter_stats_from(self, iter_stats_payload: RankIterStatsPayload | None) -> None:
+        if iter_stats_payload is None:
+            return
+        self.iter_stats = RankIterStatsPayload.deserialize(iter_stats_payload.serialize())
```
achartier: Could we use `dataclasses.replace` and avoid a full serialization / deserialization?

tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py
```diff
@@ -0,0 +1,261 @@
...
+        if is_rank0:
+            self._rank0_iter_stats[iter_id] = stats
+            self._rank0_req_stats[iter_id] = req_stats
+            self._rank0_kv_iter_stats[iter_id] = kv_iter_stats
+
+    def next_payload(self) -> Optional[RankIterStatsPayload]:
+        """Return the oldest pending stats payload to piggyback."""
+        if not self._payloads:
+            return None
+        iter_id = min(self._payloads)
```
achartier: How big does `self._payloads` get? If it is large `min` could be pretty slow. We could use a sorted struct for `_payloads` or keep track the the minimum of insert/remove instead.

tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py
```diff
@@ -0,0 +1,261 @@
...
+        can see scheduling distribution and diagnose load imbalance. Scheduled
+        fields are rank-local. Queued fields remain rank-0/global because the
+        executor request queue lives on rank 0.
+        """
+        rank = rank_state.rank
+        payload = rank_state.iter_stats
+        source_ifb = rank0_stats.inflight_batching_stats
+
+        stats = IterationStats()
+        for attr in (
```
achartier: Should we also add a test comparing this hardcoded list to the `IterationStats` fields?

tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py
```diff
@@ -0,0 +1,261 @@
...
+from .scheduler.adp_router import RankIterStatsPayload, RankState
+
+
+@dataclass
+class ADPIterStatsRecord:
+    """Append-ready stats row produced by Attention-DP fanout."""
+
+    stats: IterationStats
+    req_stats: Optional[List[RequestStats]]
+    kv_iter_stats: object
```
achartier: Should that be `Dict[int, object] | None` for better type checking?

tensorrt_llm/_torch/pyexecutor/py_executor.py
indrajit96: Done. I switched this to use req.is_dummy directly instead of rechecking the underlying fields.

tensorrt_llm/_torch/pyexecutor/py_executor.py
indrajit96: Done. The stats entry now always includes attention_dp_rank, and the serializer handles None explicitly instead of relying on tuple length.

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
indrajit96: Done. Replaced the serialize/deserialize copy with dataclasses.replace.

tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py
indrajit96: Makes sense. 
The buffer now tracks the oldest pending ADP stats iter incrementally instead of calling min() on every pop.

tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py
indrajit96: Added a test guard to compare the hardcoded ADP IterationStats field list against the actual IterationStats fields so future schema drift is caught.

tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py
indrajit96: Makes sese
Updated the type to Dict[int, object] | None for better type checking.

achartier (approved):

pcastonguay (approved):

Issue #11823, opened 2026-03-02
[Feature]:  MLA trtllm backend


MrGeva:
### 🚀 The feature, motivation and pitch

Add MLA trtllm backend and test using GLM 4.7 Flash

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and checked the [documentation](https://nvidia.github.io/TensorRT-LLM/) and [examples](https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples) for answers to frequently asked questions.

suyoggupta:
FYI, the multi-stream overlap happens as part of a graph transform https://sourcegraph.com/r/github.com/NVIDIA/TensorRT-LLM/-/blob/tensorrt_llm/_torch/auto_deploy/config/default.yaml?L249-251

PR #13677, opened 2026-04-30
[#11823][feat] AutoDeploy's mla chunked prefill loop support

MrGeva:
A common scenario in serving is prefill with a cache hit — a new request shares a prefix with one already in flight. Instead of re-computing the prefix's KV state, we read it back from the cache and combine it with the genuinely new tokens.

  The bug we kept patching

  The attention kernel that does this work has a scratch buffer for compressing K and V into a fast format (FP8). That buffer is sized via a multiplier: N × max_tokens_per_call. The kernel then writes the actual K/V tokens into the buffer.

  The bug: AD's MLA path was reading the full [past + new] KV in one shot and handing it all to the kernel in a single call. Under high concurrency, "full" became 84,252 tokens — but N = 1 only sized the buffer for 15,360. The kernel happily wrote 5.5× past the end of
  its buffer → memory corruption → mysterious crashes downstream.

  Two prior fixes:
  - Bumped N from 1 → 16 (so 16 × 15,360 = 245,760 covers the observed worst case).
  - Added a separate workspace tensor for captured CUDA graphs (orthogonal CUDA-graph stability problem).

  The N=16 fix worked but was fragile: it was a hand-tuned constant chosen because it happened to be big enough for one workload. A bigger workload would re-introduce the bug.

  The principled fix

  Instead of making the buffer bigger to fit all the KV at once, call the kernel multiple times with smaller pieces of KV. Each call now writes ≤ N × max_tokens tokens, which the buffer is sized for. By construction, no overflow is possible.

  Think of it like this: instead of trying to mail a 200-page document in one envelope (and buying bigger envelopes whenever someone writes a longer document), cut the document into chapters that always fit in a standard envelope, mail each chapter, and reassemble at the
   destination.

  How attention can be split across calls

  Attention is essentially a weighted average. The naive split would be wrong — if I average half the values, then average the other half, then average those two averages, I lose the proper weighting.

  The trick is that attention's softmax has two pieces of state: the running maximum logit (m) and the running sum of exponentials (s). If I compute attention on chunk A and chunk B separately, and I know (m_A, s_A) and (m_B, s_B), I can correctly combine the two partial
   outputs into the full attention output. This is the same math behind FlashAttention's tiling.

  The kernel was already capable of writing out these (m, s) stats per call. We just weren't using that feature. There's a separate "merge" kernel that takes (running output, partial output, running stats, partial stats) and combines them properly.

  What the new code does

  1. Plan upfront (host-side, runs before the GPU work): walk through which sequences have how much past KV, greedily pack their KV into iterations bounded by N × chunk_size tokens each. For each iteration, record:
    - which sequences contribute (and how many tokens each)
    - per-sequence read offset into the cache
    - whether this sequence is making its "first appearance" (copy partial to running) or "continuing" (merge with running)
  2. Execute the chunk loop: for each iteration:
    - Read this chunk's slice of past KV from the cache.
    - Run the attention kernel on it (with softmax-stats output enabled).
    - Call the merge kernel to combine into the running result.
  3. Final causal pass: the new tokens themselves — which haven't been mixed in yet — get one more attention pass with a causal mask, then merged in.

  By the time we return, output holds the same answer the original one-shot call would have produced, but no individual kernel call ever wrote past its buffer.

  Why N can shrink from 16 to 4

  The packer in step 1 enforces the bound. If you give it more KV than N × chunk_size, it just creates more iterations. So N is no longer a "hope this is enough" upper bound — it's a knob trading per-call workspace against iteration count. 4 is the value the standard PyTorch backend uses, and the bench shows same perf as before and slightly better accuracy.

  What didn't change

  Three completely orthogonal things stayed put:
  - The first kernel call (RoPE + cache write for the new tokens) is unchanged.
  - The fresh-prefill path (no past KV) is unchanged — it has nothing to chunk.
  - The decode path is unchanged — decode only processes 1 token per request.

  Just the cache-reused prefill path got the loop treatment, mirroring how the standard PyTorch backend already does it.

## Summary by CodeRabbit

* **New Features**
  * Added piecewise compilation configuration support for DeepSeek-R1 model.
  * Added optional output buffer support for MLA attention operations.

* **Improvements**
  * Enhanced MLA attention with improved cache handling and chunked-prefill support.
  * Improved buffer lifecycle management for CUDA-graph captured operations.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/auto_deploy/custom_ops/mla/trtllm_mla.py
```diff
@@ -262,6 +310,41 @@ def reset(self, device: torch.device, max_batch: int, max_blocks_per_seq: int) -
...
 
+        # Chunked-prefill metadata (cache-reused / chunked-prefill path).
+        # Upper bound on iteration count: with N=4 and chunk_size = max_seq_len,
+        # the worst case is ceil(total_past_kv / (N * max_seq_len)) ≤
+        # ceil(max_batch * max_seq_len / (4 * max_seq_len)) = ceil(max_batch / 4).
+        # Round up to ``max_batch`` for safety; cost is ~MB.
+        self._max_chunked_loop_num = max_batch
+        L = self._max_chunked_loop_num
+        self.chunked_seq_len = torch.zeros(L, max_batch, dtype=torch.int32, device=device)
+        self.host_chunked_seq_len = torch.zeros(
```
galagam: Use `_chunked_seq_len_host` to align with the rest of the host-side tensors naming

tensorrt_llm/_torch/auto_deploy/custom_ops/mla/trtllm_mla.py
```diff
@@ -262,6 +310,41 @@ def reset(self, device: torch.device, max_batch: int, max_blocks_per_seq: int) -
         )
         self.fmha_scheduler_counter_decode = torch.zeros(1, dtype=torch.uint32, device=device)
 
+        # Chunked-prefill metadata (cache-reused / chunked-prefill path).
+        # Upper bound on iteration count: with N=4 and chunk_size = max_seq_len,
+        # the worst case is ceil(total_past_kv / (N * max_seq_len)) ≤
+        # ceil(max_batch * max_seq_len / (4 * max_seq_len)) = ceil(max_batch / 4).
+        # Round up to ``max_batch`` for safety; cost is ~MB.
+        self._max_chunked_loop_num = max_batch
```
galagam: Any reason not to use `ceil(max_batch / _TRTLLM_MLA_CHUNK_BATCH_SIZE)` ? 
I'm assuming batch_size can grow beyond the current use case.

Also, please modify the comment to refer to _TRTLLM_MLA_CHUNK_BATCH_SIZE, if the "4" constant will ever change, this will be very hard to understand.

galagam (approved):
LGTM

MrGeva:
/bot run --disable-fail-fast --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

MrGeva:
/bot run --disable-fail-fast --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

MrGeva:
/bot skip --comment "all tests passed except for a non related test"

MrGeva:
/bot run --disable-fail-fast --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

MrGeva:
/bot run --disable-fail-fast --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

MrGeva:
/bot run --stage-list "DGX_B200-8_GPUs-AutoDeploy-Post-Merge-1"

MrGeva:
/bot run --stage-list "DGX_B200-8_GPUs-AutoDeploy-Post-Merge-1"

MrGeva:
/bot skip --comment "failing tests are not related to AD, this PR is local to AD MLA and relevant tests passed"

PR #13687, opened 2026-05-01
[TRTLLM-12432][perf] ltx2: drop redundant pe all-gather in AV cross-attention

luyiyun1021:
@coderabbitai summary

## Description

Under Ulysses sequence parallelism, LTX-2's AV cross-attention K-side does an explicit `dist.all_gather` on the (cos, sin) RoPE tensors every block × every step (`BasicAVTransformerBlock._sp_gather_pe`). The pe is **deterministic** — precomputed once per `generate()` and held un-sharded inside `MultiModalTransformerArgsPreprocessor.prepare`, then sliced for Ulysses scatter. The all-gather rebuilds data the rank already had a copy of pre-shard.

This PR removes the cos/sin all-gathers by carrying both layouts on `TransformerArgs`.

### Design

| Field | Layout | Used by |
|---|---|---|
| `cross_positional_embeddings_local` | sharded along seq | Q-side rope on sharded query |
| `cross_positional_embeddings_full`  | un-sharded         | K-side rope on all-gathered K |

Both fields are populated once at construction (`MultiModalTransformerArgsPreprocessor.prepare`) from the same `static_cross_pe`. `LTXModel._shard_transformer_args` slices `_local = _shard_pe(args.cross_positional_embeddings_full)`; `_full` passes through. AV cross-attn reads `pe=*._local` and `k_pe=*._full` directly — no `_sp_gather_pe` call.

Why precomputed fields rather than a Python conditional or env-var gate? cuda_graph captures the kernel sequence, so a Python branch has no effect on the captured graph (we verified empirically that an env-var-gated approach left the AllGather kernel count unchanged). A field whose value is fixed at trace time is captured directly. Pre-computing in `_shard_transformer_args` instead of slicing inline keeps the sharding policy in one place and avoids creating fresh view objects per block × step.

### Changes

- `transformer_args.py`: replace `cross_positional_embeddings` with `_local` + `_full`; populate both at construction.
- `transformer_ltx2.py::LTXModel._shard_transformer_args`: slice `_local` from `_full`; pass `_full` through.
- `transformer_ltx2.py::BasicAVTransformerBlock.forward`: AV cross-attn reads `_local` (Q) / `_full` (K).
- Delete the now-unused `_sp_gather_pe` method.

Net diff: `+15 / −24` across two files.

### Lossless

cos/sin are deterministic and computed once. The previous shard→gather→concat round-trip is memcpy-only (no fp ops). Reusing the un-sharded source produces a bit-identical tensor for K-side rope. Q-side path is unchanged (just renamed `_local`).

### Validation

Setup: 2-GPU pure Ulysses (`dit_cfg=1, dit_ulysses=2`), single-stage 768×1280×121 fr, gs=3.0, seed=42, `cuda_graph + torch_compile` both ON. A/B by toggling git commits on the same worktree.

**Bit-identical** — raw video tensor sha256 (pre-encode), 40-step end-to-end:

```
baseline (HEAD~1):   371cee010e51cfc4ebb0af74a783a2f95b069ad49c70f703b07d406edc5e1c98
this patch (HEAD):   371cee010e51cfc4ebb0af74a783a2f95b069ad49c70f703b07d406edc5e1c98
```

**NCCL kernels** (10-step profiled iter, nsys with `--cuda-graph-trace=node`, sum over both GPUs):

| | baseline | this patch | Δ |
|---|---:|---:|---:|
| AllGather count | 7720 | **3880** | **−3840 (−49.7 %)** |
| AllGather total | 605.45 ms | 434.61 ms | **−170.84 ms (−28.2 %)** |
| SendRecv count | 7680 | 7680 | 0 (unchanged ✓) |
| SendRecv total | 1275.09 ms | 1277.20 ms | +2 ms (noise) |
| **Total NCCL** | **1880.54 ms** | **1711.83 ms** | **−168.71 ms (−9.0 %)** |

AllGather count is exactly halved — every cos/sin all-gather scheduled by `_sp_gather_pe` is removed from the captured graph. SendRecv (project-before-gather K/V collective) is untouched, as designed.

**E2E** (40-step, single timed run, post-warmup):

| | baseline | this patch |
|---|---|---|
| E2E | 26.104 s | 25.894 s (−210 ms, −0.8 %) |

Per-GPU NCCL saving (~85 ms / 10-step iter) translates to ~1 % E2E at 2 GPU; observed delta direction is consistent. Larger Ulysses sizes are expected to see similar ~9 % NCCL kernel-time reduction.

## Test Coverage

No new tests — change is a data-routing rewrite of an existing collective, covered by existing LTX-2 generate-path and `tests/unittest/_torch/visual_gen/...` pipeline tests. No-op for single-GPU / non-Ulysses configurations: `_shard_transformer_args` is not called and both fields hold the same un-sharded tensor.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/transformer_args.py
+++ b/tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/transformer_args.py
@@ -261,8 +261,17 @@ class MultiModalTransformerArgsPreprocessor:
         static_mask: torch.Tensor | None,
         static_pe: tuple[torch.Tensor, torch.Tensor],
         static_cross_pe: tuple[torch.Tensor, torch.Tensor],
+        static_cross_pe_local: tuple[torch.Tensor, torch.Tensor] | None = None,
     ) -> TransformerArgs:
-        """Build TransformerArgs for one denoise step with pre-computed static outputs."""
+        """Build TransformerArgs for one denoise step with pre-computed static outputs.
+
+        ``static_cross_pe_local`` is the sharded-contiguous variant of
+        ``static_cross_pe`` (matches the local Q shard for Q-side rope).  When
+        ``None`` (single-rank or caller didn't pre-shard) we fall back to
+        ``static_cross_pe`` so behavior is unchanged.  Pre-sharding it once in
+        ``prepare_text_cache`` instead of every step's ``_shard_transformer_args``
+        avoids the per-step slice + downstream non-contiguous reshape copy.
+        """
         transformer_args = self.simple_preprocessor.prepare(
             modality,
             static_context=static_context,
@@ -275,9 +284,11 @@ class MultiModalTransformerArgsPreprocessor:
             batch_size=transformer_args.x.shape[0],
             hidden_dtype=modality.latent.dtype,
         )
+        if static_cross_pe_local is None:
+            static_cross_pe_local = static_cross_pe
         return replace(
             transformer_args,
-            cross_positional_embeddings_local=static_cross_pe,
+            cross_positional_embeddings_local=static_cross_pe_local,
             cross_positional_embeddings_full=static_cross_pe,
             cross_scale_shift_timestep=cross_scale_shift_timestep,
             cross_gate_timestep=cross_gate_timestep,
--- a/tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/transformer_args.py
+++ b/tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/transformer_args.py
@@ -28,7 +28,12 @@ class TransformerArgs:
     timesteps: torch.Tensor
     embedded_timestep: torch.Tensor
     positional_embeddings: tuple[torch.Tensor, torch.Tensor]
-    cross_positional_embeddings: tuple[torch.Tensor, torch.Tensor] | None
+    # Cross-modal (AV) RoPE pe carried in two layouts:
+    #   *_local : sharded along seq (matches sharded Q for Q-side rope)
+    #   *_full  : un-sharded (matches all-gathered K for K-side rope)
+    # Equal at construction time; *_local is sliced inside _shard_transformer_args.
+    cross_positional_embeddings_local: tuple[torch.Tensor, torch.Tensor] | None
+    cross_positional_embeddings_full: tuple[torch.Tensor, torch.Tensor] | None
     cross_scale_shift_timestep: torch.Tensor | None
     cross_gate_timestep: torch.Tensor | None
     enabled: bool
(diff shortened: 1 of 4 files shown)


tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -1285,34 +1261,73 @@ def prepare_text_cache(
...
-        v_ctx = v_mask = v_pe = v_cross_pe = v_kv = None
-        a_ctx = a_mask = a_pe = a_cross_pe = a_kv = None
+        v_ctx = v_mask = v_pe = v_pe_local = v_cross_pe = v_cross_pe_local = v_kv = None
+        a_ctx = a_mask = a_pe = a_pe_local = a_cross_pe = a_cross_pe_local = a_kv = None
 
         if video_context is not None:
             v_ctx, v_mask, v_pe, v_cross_pe = self.video_args_preprocessor.prepare_text_cache(
                 video_context, video_context_mask, video_positions, dtype
             )
+            v_pe_local = self._shard_pe(v_pe, is_sharded=self.use_ulysses)
```
yibinl-nvidia, line 1271: `self.use_ulysses` is unitialized here. LTXModel `__init__` only keeps `use_ulysses` as a local variable and only stores `self.use_seq_parallel`, `self.seq_parallel_size`, and `self.seq_parallel_rank`. This makes `prepare_text_cache()` raise `AttributeError` before the denoise loop.

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -1356,6 +1373,8 @@ def forward(
                 text_cache.audio_mask,
                 text_cache.audio_pe,
```
yibinl-nvidia, line 1374: When `model_type` is `VideoOnly`, the preprocessor is `TransformerArgsPreprocessor`, whose `prepare()` signature accepts `static_pe_local` but not `static_cross_pe_local`. Passing this kwarg unconditionally makes the supported non-AV forward path fail with `TypeError: unexpected keyword argument static_cross_pe_local`; you may add the kwarg to the simple preprocessor or only pass it for the multimodal preprocessor.

tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/transformer_args.py
```diff
@@ -28,7 +28,12 @@ class TransformerArgs:
     timesteps: torch.Tensor
     embedded_timestep: torch.Tensor
     positional_embeddings: tuple[torch.Tensor, torch.Tensor]
-    cross_positional_embeddings: tuple[torch.Tensor, torch.Tensor] | None
```
yibinl-nvidia, line 31: Could you add tests to catch the issue above?

yibinl-nvidia (changes requested):
Left a few comments.

PR #13713, opened 2026-05-03
[https://nvbugs/6104831][fix] Disaggregated KV transfer: lifecycle, cancellation, and quiescence hardening

chienchunhung:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of disaggregated KV cache transfers with enhanced cancellation and timeout handling across context and generation phases.
  * Enhanced resource cleanup and lifecycle safety for asynchronous transfer operations to prevent memory leaks during errors or cancellations.
  * Better error handling and state management during KV cache transfer failures and request termination.

## Summary

For Disaggregated Serving under long-prompt bursts with client-side cancellations, the existing cleanup path could leave the deployment alive but permanently non-responsive (wedge): workers up, `/health` returns 200, but every post-burst request timed out.

This PR fixes the permanent wedge for the PyTorch backend. Root cause is a stack of cleanup/lifetime invariant gaps across C++ KV transfer, Python termination, NIXL cancel, and block reuse. Cancelled requests now stop scheduling immediately, but KV resources release only after C++ transfer is terminal — and deferred termination is retried exactly once when it becomes safe.

**Operator default**: mid-flight NIXL cancel + fail-closed-on-unquiesced is **opt-in via `TRTLLM_DISAGG_ENABLE_INFLIGHT_CANCEL=1`**. Default unset → no behavioral change vs. pre-PR baseline for that surface. The orthogonal lifetime, idempotency, RAII, and rc13-block-reuse fixes ship **always-on**.
Two late additions during the upstream/main merge cycle close two further failure modes that became deterministic in CI after this PR's `shared_ptr<LlmRequest>` lifetime extension turned previously-loud per-rank divergences into silent hangs:
- Commits `bdfdf8be02` (gen side) and `53a0692aa4` (ctx side) close the sig #9 cross-collective ABBA on the helix `pp1dp2cp2` / `pp2tp1cp2` parametrizations by removing three rank-asymmetric Python gates from disagg gen-side scheduling and two from ctx-side scheduling so every rank reaches the C++ `gatherRequestIds` collective together. See sig #9 and the rank-asymmetric-gates row in the change table.
- Commit `9a1a0329fb` (Option C) caps the per-iteration `wait_for` slice in `checkContextTransferStatus` / `checkGenTransferStatus` at 50 ms so the polling thread always cycles back to drive MPI/UCX forward progress. Without the cap, when neither `kv_transfer_sender_future_timeout_ms` nor `kv_transfer_timeout_ms` is configured, `effectiveSliceMs` fell through to `INT64_MAX` — and `wait_for(INT64_MAX)` on a not-yet-ready future blocked the polling thread indefinitely on transports without their own progress thread (notably `mpi_kvcache` over UCX-over-shm). See the cpp-gtest-wedge row in the failure-signatures and change tables.

Co-worked with @yifjiang.

## Failure signatures

| Sig | Symptom | Root cause |
| --- | ------- | ---------- |
| #1 | Sender-side `Broken promise` | cancel-after-ready erased the promise without fulfilling it |
| #4 | Generation event loop blocked forever | `checkGenTransferStatus(atLeastNum=1)` called `future.get()` on an unready future |
| #5 | Receiver-side `Broken promise` | queued cancel erased the receiver promise without fulfilling it |
| #6 | Receiver buffer-pool wedge | `requestSync()` early-return leaked a recv-buffer slot |
| #7 | First-request SIGSEGV / mutex wedge | lifetime + eval-order bugs after moving from raw pointer to `shared_ptr<LlmRequest>` |
| rc13 | No recovery even after #1–#7 fixed | rc13's default-on block reuse: Python deferred termination during in-flight context transfer, then skipped cleanup because it assumed early termination had already succeeded |
| #9 | Helix CI hang: gen ranks asymmetrically enter the C++ `gatherRequestIds` collective vs. proceed to ADP `tp_allgather` / PP step-boundary collective | rank-local Python gates over per-rank state (`fitting_disagg_gen_init_requests`, `_disagg_gen_kv_recv_started_ids`, `any(req.is_disagg_generation_transmission_in_progress)`) before the only call chain into the C++ cross-rank collective. Pre-PR raw-pointer lifetime crashed loudly on the same divergence; this PR's `shared_ptr<LlmRequest>` closes that UAF and makes the divergence silent, turning it into an ABBA deadlock |
| cpp-gtest-wedge | `test_asymmetric_executor[llama-4proc-mpi_kvcache-103]` deterministically hits gtest's 300 s internal timeout on every PR build | `effectiveSliceMs` in `checkContextTransferStatus` / `checkGenTransferStatus` fell through to `INT64_MAX` when neither `kv_transfer_sender_future_timeout_ms` nor `kv_transfer_timeout_ms` was configured (the cpp gtest's default `CacheTransceiverConfig` sets neither). `wait_for(INT64_MAX)` on a not-yet-ready future blocked the polling thread; transports without a dedicated progress thread (notably `mpi_kvcache` over UCX-over-shm) then never advanced the in-flight transfer → future never became ready → death spiral. Test passes in ~35 s with the 50 ms cap. |

## Correctness invariant

```
A request can be cancelled at the API level immediately,
but its KV resources cannot release/reuse until:
  1. C++ KV transfer is terminal, or transfer/buffer state is fail-closed;
  2. transfer-buffer ownership has been released or poisoned;
  3. block-reuse / prefix-cache references are accounted for;
  4. Python resource cleanup runs exactly once.
```
**Python request state alone is insufficient:** with block reuse, a cancelled request may still have KV blocks referenced by the reuse tree or pinned by an in-flight context transfer. Freeing too early → memory corruption. Forgetting to free → permanent wedge.

## What this PR changes

| # | Change | Default | Closes |
| - | ------ | ------- | ------ |
| 1 | C++ transceiver uses `std::shared_ptr<LlmRequest>` across async send/receive | always-on | sig #7 UAF (Python termination racing C++ workers) |
| 2 | Cancellation paths fulfil promises before dropping queued/ready work; recv-side per-request idempotency guards `std::promise` against double-fulfilment | always-on (idempotency); gated (cancel propagation) | sigs #1, #5 (`Broken promise`) |
| 3 | `checkGenTransferStatus(atLeastNum=1)` skips unready entries in non-blocking polling | always-on | sig #4 (gen loop blocked on unready future) |
| 4 | NIXL `TransferStatus::release()` → `nixlAgent::releaseXferReq()` on cancel | **gated** | wedged workers under terminal peer failure (NVBug 6104831 customer report) |
| 5 | Materialize `reqId` before `std::move(resp)` in `sendAndRemoveResponse` (C++ eval-order UAF after the `shared_ptr` migration in #1) | always-on | sig #7 variant |
| 6 | Python `py_request_id`-keyed idempotency on disagg gen init + KV recv setup | always-on | duplicate `addSequence()` / `request_and_receive_async()` calls |
| 7 | Fail-closed on unquiesced transfer: `BufferIndexHolder` RAII (always-on); `poison()` on cancel + tri-state ready-signal + Python `_fail_closed_for_unquiesced_disagg_transfer` | RAII always-on; poison + fail-closed **gated** | silent buffer-pool corruption under cancel races |
| 8 | Deferred cleanup after timeout cancel: do not free Python resources until C++ transfer status reports terminal | always-on (the deferred-cleanup pattern); gated (the `cancel_request()` call) | premature free while NIXL/C++ may still touch buffers |
| 9 | rc13 block-reuse-safe deferred termination: `RequestTransferMetadata{termination_requested, resources_freed}` + `EndTransferResult{completed, needs_termination}`; `_end_transfer_and_maybe_terminate()` retries termination if an earlier attempt was deferred | always-on | rc13 regression (dual-path lost cleanup obligation) |
| 10 | Remove rank-asymmetric Python gates in disagg gen-side scheduling (`_prepare_disagg_gen_init`, `_recv_disagg_gen_cache`, `_check_disagg_gen_transfer_status` — commit `bdfdf8be02`) AND ctx-side scheduling (`_executor_loop_pp` + `_executor_loop`'s `if num_fitting_reqs == 0 and not fitting_disagg_gen_init_requests` gates — commit `53a0692aa4`) so every rank reaches the corresponding C++ `gatherRequestIds` cross-rank collective together. C++ side handles empty `mSenderFutures` / `mRequesterFutures` cheaply (one empty allgather, no per-request loop work). | always-on | sig #9 (helix CI hang: cross-collective ABBA against `_can_queue::tp_allgather` under attention-DP / PP step-boundary collective under `gen_pp > 1`) |
| 11 | Cap `effectiveSliceMs` in `checkContextTransferStatus` / `checkGenTransferStatus` at `kMaxPollSliceMs = 50` ms (commit `9a1a0329fb` — Option C) so the polling thread always cycles back to drive MPI/UCX forward progress. The deadline-check observability above is **unchanged** — if `kvTransferTimeoutMs` is set, the deadline still fires within `kMaxPollSliceMs` of true expiry; cancellation semantics (gated by `TRTLLM_DISAGG_ENABLE_INFLIGHT_CANCEL`) unchanged. | always-on | cpp-gtest-wedge (transport-progress starvation on `mpi_kvcache` under default `CacheTransceiverConfig`) |
### 12. Opt-in flag: `TRTLLM_DISAGG_ENABLE_INFLIGHT_CANCEL` *(new)*
Gates the `{cancel propagation, poison signal, fail-closed}` triple — sections 4, the cancel-propagation half of 2, and 7 — behind a single env var. Default unset → no behavioral change vs. pre-PR baseline for that triple.
| Mode | `cancel_request` | C++ cancel chain | `has_poisoned_transfer_buffer` | Layer-5 fail-closed |
| ---- | ---------------- | ---------------- | ------------------------------ | ------------------- |
| **Default (unset / `=0`)** | no-op, returns `False` | dormant | returns `False` | never fires |
| `=1` | full path | active | reports `true` on poison | fires on poison |

## Tests

New / updated unit coverage:
* sender cancel-after-ready promise fulfillment
* receiver queued-cancel promise fulfillment
* non-blocking `checkGenTransferStatus(atLeastNum=1)`
* deferred termination surviving until async transfer completion
* partial-reuse path still skipping termination when none requested
* successful early free preventing duplicate termination
* direct `EndTransferResult.needs_termination` coverage
Integration regression coverage:
* `TestDeepSeekV3Lite::test_auto_dtype_with_helix[fifo_v2-cudagraph:with_padding-pp1dp2cp2]` and `[…-pp2tp1cp2]`. Both wedged at the 300 s hang-detector on CI build `#39529` (commit `5234311`); both PASS under the CI-default configuration (default `kv_transfer_timeout_ms=60000`, no `TRTLLM_DISAGG_ENABLE_INFLIGHT_CANCEL`) post-fix, with GSM8K accuracy `63.87` vs the `61.54` hypothesis-test threshold — i.e. the model actually serves the workload, not a no-op pass.
* `test_asymmetric_executor[llama-4proc-mpi_kvcache-103]` (cpp gtest): deterministically wedged at gtest's 300 s internal timeout on every pre-Option-C PR build. Passes in ~35 s post-Option-C (item #11), without any `CacheTransceiverConfig` timeout being configured.

## Notes / limitations

* The direct UCX path has a separate throughput/backpressure boundary under high concurrency. This PR primarily fixes the NIXL customer path and the shared disagg cleanup invariants.
* MLA send-path hardening included for parity; stress validation uses Qwen3-0.6B which primarily exercises the non-MLA formatter.
* The local CuTe DSL import guard used during development is not part of this PR.
* `TestQwen3NextInstruct::test_auto_dtype[use_py_transceiver=False]` (gen `pp=2`, NIXL → C++ transceiver, listed in `l0_dgx_b200.yml` with `TIMEOUT (60)`) matches the sig #9 trigger condition exactly (multi-rank gen + downstream unconditional collective + C++ transceiver). The extended timeout most plausibly masks the same hang. Worth re-running with the `TIMEOUT (60)` annotation removed once this PR lands; tracked as a follow-up in the investigation report's `08-next-steps-and-pr-map.md` item 12.
* **Cross-rank consensus enforcement (follow-up work, NOT in this PR)** — a follow-up implementation of doc 12's Path B (explicit horizontal-consistency layer) is prepared and locally validated. It ports the V2 transceiver's `_consensus_outcome` pattern into V1's C++ `checkContextTransferStatus` / `checkGenTransferStatus` as a four-pass pipeline (readiness consensus → local classify+cache → 2 outcome allgathers → state transitions), gated behind a new env var `TRTLLM_DISAGG_USE_CONSENSUS_OUTCOME` (default OFF, byte-identical to this PR's default behaviour). Local validation across 11 runs on 4 test families, 4 topologies, 3 transports, 4 models — all PASSED, 3 of 4 helix-class tests caught real cross-rank divergence events the consensus mechanism correctly deferred, cache mechanism survived multi-iteration deferral. See investigation doc `14-cross-rank-consistency-enforcement.md` for the full design + empirical results + updated landing plan. Decision on whether to fold this into this PR or land as a follow-up is gated on this PR's CI signal.

## PR Checklist

## GitHub Bot Help


yifjiang (commented):
Review notes from a read-through of the PR. Three concerns, all phrased as questions / follow-up suggestions rather than blockers.

cpp/tensorrt_llm/executor/cache_transmission/agent_utils/connection.cpp
```diff
@@ -143,8 +143,48 @@ void AgentConnection::send(DataContext const& ctx, void const* data, size_t size
...
+        {
+            // Cancel observed mid-send. Release the underlying backend
+            // transfer handle before unwinding so NIXL does not retain
+            // stale in-flight state below TRT-LLM's request cleanup.
+            bool const released = status->release();
+            TLLM_LOG_WARNING(
+                "AgentConnection::send cancelled while transfer was in progress (ctx tag=%d, remote=%s, "
+                "releaseAccepted=%d)",
+                ctx.getTag(), mRemoteAgentName.c_str(), released);
+            TLLM_CHECK_WITH_INFO(
```
yifjiang, line 167: `MooncakeTransferStatus` doesn't override the new `TransferStatus::release()`, so it inherits the base `return false;` and trips this CHECK on every cancel-mid-send (also cascading into pool poisoning via the catch in `cacheFormatter::format()`). Could you file a follow-up PR to add a `MooncakeTransferStatus::release()` override (a documented no-op returning `true` would be enough if Mooncake has no real mid-flight cancel API)?

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
```diff
@@ -468,6 +464,10 @@ void CacheFormatter::format(tensorrt_llm::batch_manager::TransferSession& sessio
             NVTX3_SCOPED_RANGE(sendBufferFun);
 
             TLLM_CHECK(pickUpConnections.size() == 1);
+            TLLM_CHECK_WITH_INFO(false,
```
yifjiang: The `TLLM_CHECK_WITH_INFO(false, ...)` here disables zcopy on this SEND path, but the same cancel-vs-freed-blocks race applies to four other zcopy paths the PR leaves untouched: `cacheFormatter.cpp:723` (recv), `mlaCacheFormatter.cpp:463` (recv), `rnnCacheFormatter.cpp:83/248`, and the layer-wise SEND at `cacheFormatter.cpp:411-415` (ships raw block pointers without the env var gate). Worth either guarding those too, or pivoting the design — making zcopy transfers non-cancellable (skip the cancel-flag check in `AgentConnection::send` / `recvReadySignal` for zcopy paths) would cover all paths uniformly and is memory-safe given the shared_ptr migration this PR already adds. Fine as a follow-up if you'd rather keep this scoped.

yifjiang (commented):
Three concerns, all phrased as questions / follow-up suggestions rather than blockers.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -3510,14 +3653,18 @@ def _check_disagg_ctx_cache_transfer_status(self, atLeastNum: int = 0):
...
-                # Otherwise, try at next iteration
-                if is_cancelled:
-                    request.py_kv_transfer_start_time = None
-                    request.state = LlmRequestState.DISAGG_CONTEXT_COMPLETE
-
-                    self._end_transfer_and_maybe_terminate(request)
+                # Ask C++ to cancel once, but do not free Python-side
+                # resources until C++ reports the transfer in finished/error
+                # status. cancel_request() is not a transport-quiescence proof.
+                if request_id not in self._disagg_timed_out_ctx_cancelled_ids:
```
yifjiang, line 4122: When poison fires on the C++ side (the buffer pool throws with *"process must restart"*), the call to `self._check_cache_transfer_errors("context requests")` a few lines below ultimately reaches `_handle_errors(..., charge_budget=False)` (line 3626) — which means the error budget is never consumed, `_fatal_error` is never set, and no shutdown is enqueued. So the pod stays alive serving 100% disagg errors with no signal to trtllm-serve callers, Dynamo's router, or kubelet readiness. Worth surfacing poison as fatal in Python (and/or via the readiness endpoint) so the "must restart" intent actually causes a restart — fine as a follow-up PR.

cpp/tensorrt_llm/batch_manager/baseTransBuffer.cpp
```diff
@@ -225,16 +311,46 @@ void BaseTransBufferManager::allocateBuffer()
...
     std::unique_lock lk(resource.mBuffersMutex);
-    resource.mBuffersCV.wait(
-        lk, [&resource, bufferCount]() { return static_cast<size_t>(resource.mConcurrence) < bufferCount; });
+    auto const predicate = [&resource, bufferCount]()
+    {
+        return resource.mPoisoned.load(std::memory_order_relaxed)
+            || static_cast<size_t>(resource.mConcurrence) < bufferCount;
+    };
+    if (!predicate())
+    {
```
pcastonguay: nit: this is not needed I think given the inner while loop.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
pcastonguay: I think this check should happen in the constructor.

pcastonguay (commented):
Overall lgtm. Two nits.

cpp/include/tensorrt_llm/batch_manager/cacheTransceiver.h
```diff
@@ -204,13 +205,17 @@ class BaseCacheTransceiver
 {
 public:
     virtual ~BaseCacheTransceiver() = default;
-    virtual void respondAndSendAsync(LlmRequest* llmRequest) = 0;
+    // These methods take std::shared_ptr<LlmRequest> so the transceiver and
+    // its async workers can hold a strong reference for the duration of the
+    // transfer. See the comment on CacheTransceiver::mSenderFutures for the
+    // lifetime invariant (kept in one place to avoid drift).
```
Tabrizian: Remove the comment.

cpp/include/tensorrt_llm/batch_manager/cacheTransceiver.h
```diff
@@ -276,8 +281,27 @@ class CacheTransceiver : public BaseCacheTransceiver
...
+    // - mRequesterFutures (gen side): on timeout, keep the entry tracked
+    //   via mTimedOutRequesterIds until the worker future resolves. A
+    //   timeout/cancel is not a quiescence proof on the recv side, so the
+    //   advertised receive buffers may still be written to until the worker
+    //   unwinds. See checkGenTransferStatus.
+    // - mSenderFutures (ctx side): erased immediately on completion,
+    //   exception, or timeout. Sender zombies empirically unwind on peer
+    //   teardown (decode-pod restart), and CacheSender::cancelRequest is
+    //   only required to clear bookkeeping for telemetry / re-enqueue
+    //   paths. See checkContextTransferStatus.
```
Tabrizian: Remove these comments.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
```diff
@@ -573,12 +586,26 @@ void CacheFormatter::format(tensorrt_llm::batch_manager::TransferSession& sessio
...
+        try
+        {
+            sendAllBuffers(session, deviceId, outputSplitCaches, bufferCoverTargetNum, preAllocSendBuffer,
+                bufferManager, targetInfo, pickUpConnections);
+        }
+        catch (...)
+        {
+            if (agentConnection != nullptr)
+            {
+                sendHolder.poison();
```
Tabrizian, line 634: Instead of poisoning the buffer can we clean up the buffer properly so that it can be reused by future runs? What's the exception here?

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
```diff
@@ -60,6 +60,28 @@
...
+///
+/// @note Logging policy for cacheTransceiver / dataTransceiver:
+///       new logs are added only at entry/exit boundaries
+///       (checkGenTransferStatus done / context-completion summaries)
+///       or for cancel-flow transitions ("Flipped ... cancel flag").
+///       Per-iteration polling-loop logs are deliberately omitted to
+///       keep TLLM_LOG_DEBUG output proportional to actual
+///       state-changing events. See PR #13713 review for the
+///       rationale; if you add a new TLLM_LOG_DEBUG here, audit
+///       whether it fires per iteration vs per state transition.
```
Tabrizian: Can we shorten the comment? Looks like a one line only change.

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
```diff
@@ -485,10 +525,29 @@ RequestStatuses CacheTransceiver::checkContextTransferStatus(
 {
     bool blockAll = !atLeastRequestNum.has_value();
     std::optional<int> senderFutureTimeoutMs = std::nullopt;
-    // If blockAll is true, we want to block and not use a timeout
-    if (!blockAll && mCacheTransceiverConfig.has_value())
+    std::optional<int> kvTransferTimeoutMs = std::nullopt;
+    if (mCacheTransceiverConfig.has_value())
     {
-        senderFutureTimeoutMs = mCacheTransceiverConfig->getKvTransferSenderFutureTimeoutMs();
+        // Behaviour change vs upstream/main: the overall transfer deadline
```
Tabrizian: What would "Behaviour change vs upstream/main:" mean when this change is merged to `main`?

chienchunhung: Those are stale comments; rephrased both call-site comments to describe the behavior directly without referencing pre-merge state.

cpp/tensorrt_llm/batch_manager/cacheTransceiver.cpp
chienchunhung: Done.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
chienchunhung: The exception this catch covers is mostly for `session.send` → `AgentConnection::send` throwing on cancel-mid-transfer, cancel-pre-notify, or transport failure (with rarer CUDA copy/sync errors). By the time we reach here on an AgentConnection, NIXL has already been handed the send buffer, and `status->release()` only returns the request handle to the backend pool — it does not prove the backend "stopped reading the underlying memory", so returning the slot via normal release() could let a future request overwrite the buffer while the backend is still draining it, which is the exact cancel-vs-freed-blocks race this PR closes. 

Waiting for true quiescence is not an option either, because the worker that would observe it is the one that just threw. Direct-UCX (`agentConnection == nullptr`) does not have this exposure since `session.send` is synchronous, so we keep the normal `release()` path there. 

To make the fail-closed signal actionable we also added `has_poisoned_transfer_buffer()` in the latest change, which lets `_check_cache_transfer_errors` route the next iteration through fatal error handling so the process restarts with a fresh pool.

cpp/include/tensorrt_llm/batch_manager/cacheTransceiver.h
chienchunhung: Done.

cpp/include/tensorrt_llm/batch_manager/cacheTransceiver.h
chienchunhung: Done.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
chienchunhung: Done.

cpp/tensorrt_llm/batch_manager/baseTransBuffer.cpp
chienchunhung: Done — removed the redundant outer if (!predicate())

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Fixed by surfacing poisoned transfer-buffer state from C++ to Python via has_poisoned_transfer_buffer(). _check_cache_transfer_errors() now treats that case as unrecoverable, so it goes through the fatal-error path and enqueues shutdown instead of serving repeated disagg errors.

cpp/tensorrt_llm/batch_manager/cacheFormatter.cpp
chienchunhung: I made some tweaks and moved the zero-copy disallowance to formatter construction, so `CacheFormatter`, `MLACacheFormatter`, and `RnnCacheFormatter` all reject `TRTLLM_TRY_ZCOPY_FOR_KVCACHE_TRANSFER` before any send/recv path can run. 

This covers KV send/recv, MLA send/recv, RNN send/recv, and the layer-wise KV send path uniformly, and removes the local branch-level hard-fail.

cpp/tensorrt_llm/executor/cache_transmission/agent_utils/connection.cpp
chienchunhung: Added MooncakeTransferStatus::release() as a documented no-op returning true, since the Mooncake path does not expose a per-request mid-flight release/cancel primitive at this layer.

pcastonguay (approved):

pcastonguay:
@chienchunhung the DCO check failed, probably one of your commit wasn't signed off. Can you check?

Tabrizian (approved):

pcastonguay:
Looks like the failing tests are not flaky tests, but seems to consistently fail with your changes. Can you prioritize investigating and fixing those? Thanks.

Shixiaowei02 (approved):

chienchunhung:
/bot run --disable-fail-fast --stage-list "DGX_H100-4_GPUs-CPP-1"

chienchunhung:
PR#14768 will replace this PR (PR#13713) as a safer subset of the changes with smaller blast radius: it only includes the changes required for resolving UAF memory corruption, segmentation fault, ..etc.

Issue #12359, opened 2026-03-19
[AutoDeploy]:  Super V3 with MTP performance


galagam:
### 🚀 The feature, motivation and pitch

Discuss with @govind-ramnarayan 

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and checked the [documentation](https://nvidia.github.io/TensorRT-LLM/) and [examples](https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples) for answers to frequently asked questions.

galagam:
Working on top of #12745 which hasn't merged yet due to CI issues.

I'm able to get some quick wins using `flashinfer.mamba.selective_state_update` to replace the current triton ssm kernel (the underlying kernel is `selective_state_update_kernel_simple_mtp` ). Will run some more tests and put out for review.

I also have a draft for `SpecDecAwareScheduler` - When MTP extend-step requests are in-flight, it can defer new prefill (context) requests up to k consecutive cycles to keep batches extend-only, so we can run full cuda graphs (some small k to avoid starvation and big increase in TTFT). 

Next: 
1. Revert unrelated (probably) config changes:
```
<     allreduce_strategy: SYMM_MEM
---
>     allreduce_strategy: NCCL
```
and
```
<   gather_logits_before_lm_head:
<     # TODO: fix https://github.com/NVIDIA/TensorRT-LLM/issues/9878 to enable by default
<     enabled: true
---
>     enabled: false
```

2. Fix two known host-device syncs:
https://github.com/NVIDIA/TensorRT-LLM/pull/12745/changes#diff-d6bde3a9e362bf1e5a7899e25fe993f9c9d2bc1b008a1e7750579d77227ddcd5R1414
https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/triton_backend_causal_conv.py#L109 

3. Enable piecewise CG
4. Consider tuning max_draft_len according to batch size

galagam:
Filed https://github.com/NVIDIA/TensorRT-LLM/pull/13711
for FlashInfer SSM kernel + extended tests

galagam:
Filed #13725 for replay SSM - WIP - perf is not where expected (better than previous AD version, but expected higher perf gain)

galagam:
AD:
[...]

PR #13725, opened 2026-05-04
[#12359][feat] AutoDeploy: Support SSM replay kernel for MTP with FlashInfer

galagam:
## Release Notes

* **New Features**
  * Added MTP speculative decoding configurations for SuperV3 model (single-GPU and multi-GPU setups).
  * Introduced SSM replay state-update capability for improved speculative decoding efficiency.

* **Improvements**
  * Updated SuperV3 + MTP configuration with optimized throughput-oriented limits.
  * Enhanced Mamba SSM kernel compatibility and cache handling.

* **Tests**
  * Added smoke test for SuperV3 MTP with SSM replay functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13725?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Description
Integrate replay_selective_state_update (tl.dot fast-forward) into the
AD flashinfer_ssm backend as an optional path controlled by:

      insert_cached_ssm_attention:
        backend: flashinfer_ssm
        ssm_replay: true

The replay kernel replaces the FlashInfer horizontal SSM scan in the MTP extend path with the same two-kernel algorithm the pytorch backend uses: 
- _replay_precompute_kernel: computes CB_scaled/decay_vec from cached B,dt,A
- _replay_state_update_kernel: fast-forwards SSM state via tl.dot on cached values, then computes output using new x/C and precomputed tensors

## Test Coverage

Unit smoke (tests/unittest/auto_deploy/singlegpu/smoke/test_ad_speculative_decoding.py):
- test_super_mtp_smoke: existing triton_ssm path (unchanged, all pass)
- test_super_mtp_ssm_replay_smoke: flashinfer_ssm + ssm_replay=True

Accuracy (B200, ws4_180gb, trtllm backend):
  nvfp4 + fp8: PASSED  (GSM8K ~92%, acceptance_rate ~45%)

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
```diff
@@ -154,47 +290,68 @@ def _flashinfer_cached_ssm(
...
-FLASHINFER_SUPPORTED_HEAD_DIMS = [64, 128]
-
-
 @AttentionRegistry.register("flashinfer_ssm")
 class FlashinferBackendSSM(BaseBackendSSM):
+    # When ssm_replay=True, use the replay kernel (tl.dot fast-forward) instead of FlashInfer.
+    # Disabled automatically when: block reuse enabled, tree attention, or SM < 80.
+    ssm_replay: bool = False
+
+    # Cache keys always passed as kwargs (follow constants at positions 16-17).
```
govind-ramnarayan: I'm not sure how much of a problem it is to have unused inputs - we've already had this problem at a smaller scale with speculative decoding without replay (when MTP is off, the intermediate ssm cache defaults to None) and thought it was preferable to adding a special kwargs path.

I think the design might be more natural with something like the following, similar to what we did for the "rewind" speculative resource handling. 

Each SSM FlashInfer kernel alway outputs resources in `get_cache_initializers()`:
- SSMResourceHandler
- RewindSSMResourceHandler
- various Replay SSM resources

Then depending on the speculative configuration, the cache manager either binds them or not. So if speculation is on but replay is off, bind the RewindSSMResourceHandler like here: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/shim/interface.py#L577-L586

And if replay is on, instead bind the Replay handlers. 

Edit: And a key with this was the args did *not* default to None, it was just left unmanaged and unused. I manually set it to None in a bunch of unit tests.

If there is a perf / memory usage issue with having these unused dummy arguments, let's confirm it and understand it so we can weigh it and see whether it merits changing the design of ResourceHandlers / cached ops.

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
```diff
@@ -89,11 +135,103 @@ def _flashinfer_cached_ssm(
...
+            dt_bias_hp,
+        ) = extend_inputs
+
+        preallocated_ssm_out_e = preallocated_ssm_out[
+            num_prefill_tokens : num_prefill_tokens + num_extend_tokens
+        ].view(num_extend, tokens_per_extend, num_heads, head_dim)
+
+        slot_idx_extend_i32 = slot_idx_extend.to(torch.int32)
+
+        use_replay = replay_old_x is not None
```
govind-ramnarayan: if we go the route described below where replay_old_x may be passed in non-replay states, then we can add `use_replay` to `BatchInfo` instead, and then branch based on that.

I think this is better anyway rather than relying on the presence or absence of caches - we already branch on various elements of BatchInfo (`num_extend`, `num_prefill`, `num_decode`, etc.) for making decisions about which path to go down.

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
```diff
@@ -154,47 +290,68 @@ def _flashinfer_cached_ssm(
...
     )
 
 
-# Flashinfer's selective_state_update kernel only supports these head dimensions
-FLASHINFER_SUPPORTED_HEAD_DIMS = [64, 128]
-
-
 @AttentionRegistry.register("flashinfer_ssm")
 class FlashinferBackendSSM(BaseBackendSSM):
+    # When ssm_replay=True, use the replay kernel (tl.dot fast-forward) instead of FlashInfer.
```
govind-ramnarayan: You wouldn't need this either - though it does seem like a good place to put this info as it feels like an internal kernel detail. Maybe it's okay - the other natural place would be spec_config if the flag is already there, but this feels a bit more natural.

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
galagam: It's a good question, I started off with the approach you mentioned - it works. There's no memory issue because we use empty tensors/None as placeholders.
However, we're just adding dead weight to the graph. With Python, using kwargs seems more natural than the C/C++ approach of positional args. It's not a prerequisite for enabling replay, we can do this with dummy positional args as well, but I think it's a good infrastructure change for cached operator.

@lucaslie @MrGeva - WDYT?

hnover-nv:
was there anything in particular you wanted me to review?  Or helpful to look at nsys traces if we aren't seeing much perf improvement?

galagam:
> was there anything in particular you wanted me to review? Or helpful to look at nsys traces if we aren't seeing much perf improvement?

@hnover-nv mostly for general awareness and to prevent dup. work. I'll ping you again when I move it out of draft. Still working on it.

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
```diff
@@ -89,11 +135,103 @@ def _flashinfer_cached_ssm(
...
+            num_prefill_tokens : num_prefill_tokens + num_extend_tokens
+        ].view(num_extend, tokens_per_extend, num_heads, head_dim)
+
+        slot_idx_extend_i32 = slot_idx_extend.to(torch.int32)
+
+        use_replay = replay_old_x is not None
+        if use_replay:
+            # Replay path: fast-forward SSM state via tl.dot on cached values.
+            # State is updated in-place; no disable_state_update needed.
+            replay_selective_state_update(
```
govind-ramnarayan, line 207: Note: Better understand this kernel - can we provide a single replay tensor that we slice appropriately? Ask Harris

hnover-nv: probably?  what tensors do you want to smash together?

govind-ramnarayan: Ideally, all the replay ones: 

`   replay_old_x: Optional[torch.Tensor] = None,  # [max_batch, T, nheads, head_dim]
    replay_old_b: Optional[torch.Tensor] = None,  # [max_batch, 2, T, ngroups, dstate]
    replay_old_dt: Optional[torch.Tensor] = None,  # [max_batch, 2, nheads, T] fp32
    replay_old_da_cumsum: Optional[torch.Tensor] = None,  # [max_batch, 2, nheads, T] fp32
    replay_cache_buf_idx: Optional[torch.Tensor] = None,  # [max_batch] int32
    replay_prev_num_accepted: Optional[torch.Tensor] = None,  # [max_batch] int32`

The reason for this is more for code cleanliness than anything. There are 6 separate buffers fed to this custom op when they semantically either all come together or none of them are there. Ideally we would be feeding something like a `ReplayBuffer` struct, but I believe due to some restrictions on torch custom ops the arguments here can't be structs (I believe it's true but maybe something to confirm). So, if we can package them together in a tensor that would be nice.

This has another benefit. Right now our [ResourceHandlers](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py#L1526) have two nice properties:
1) Each ResourceHandler corresponds to a single tensor.
2) Each ResourceHandler covers a fairly broad family of resources

If we have 6 separate ResourceHandlers, 2) is no longer there. It's not a travesty but would be nice to preserve if we can.

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
govind-ramnarayan: (Including the results of some convo with Lucas about this)

I think this will be nicer if we can move to one replay tensor that gets fed to the kernel or gets sliced appropriately by our custom op. Then it introduces just one other unused arg - which is the same as for the intermediate state buffers currently. I've mentioned this below to @hnover-nv .

I don't see anything fundamentally bad with kwargs either, but as it's only done for speculative SSM arguments so far, maybe instead of supporting kwargs in general, we can add code that sets these args to None before calling the custom op, if they belong to unmanaged speculative handlers. This is if we want to address the concern of many unused inputs to the graph op. That way the code change is restricted to the handling of speculative resources, which is exactly what motivated it. Is it possible?

If doing this is messy or there's another reason we can't do it I think the kwargs fallback is okay.

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
galagam: Thanks @govind-ramnarayan , that's an interesting suggestion, and if PyTorch would have supported dicts/structs already it could be ideal. Unfortunately, it's not supported... Stacking together the buffers together seems very messy, and will also incur perf loss due to the slicing / type conversions (we have both fp32 and int32 buffers).

> 2) Each ResourceHandler covers a fairly broad family of resources. If we have 6 separate ResourceHandlers, 2) is no longer there.

Can you explain what do you mean by "2) is no longer there." ? 

Another possible option, is to have a completely separate custom op for mamba with replay, but is seems like a lot of code dup.

govind-ramnarayan: Yeah I thought about the custom op - I think the code dup will cause headaches down the road and is not worth it... wdyt?

By "2) is no longer there" I meant that 
> Each ResourceHandler covers a fairly broad family of resources
is no longer true. 

ResourceHandlers are just very specific now - they effectively have no interpretation beyond "a tensor required for our particular custom op implementation" (as opposed to `KVPagedResourceHandler` which is both a tensor and also abstractly represents KV Paged resources). But this was always bound to diverge if a custom op takes multiple cached tensors, so I think it's fine.

I discussed with Lucas that if the number of `ResourceHandlers` is getting too high then we should probably move them to their own file, or at least out of `attention_interface.py`. I think it makes sense to me, `attention_interface.py` generally needs some rethinking / reorganization as it carries so much load for us right now. I think it's best to leave this for a follow-up cleanup task but I'll leave it up to you.

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
govind-ramnarayan: Okay re: this.

I think this is the main design issue left. From what I understand the issue with the previous design was mainly:

1. Many unused arguments to the custom op, which are empty tensors. There used to be just one (the intermediate ssm states) but now there are like 6 or 7.
2. The logic you use to choose the replay branch within the extend branch relies on the replay tensors being None when replay is off. Otherwise choosing replay vs non-replay would require some sort of global argument or new batch_info knob - I'm not a fan of the latter because it feels like a kernel-specific thing.

I think these are decent points. I am leaning towards setting all the unused spec-dec resources to "None" explicitly before calling the operation if it is clean, as this ties closer to the issue we are solving (that speculation requires the op to carry around different caches). But honestly I'm also fine with the kwargs since it seems like a natural extension of what we have and is fairly harmless.

galagam: @govind-ramnarayan @hnover-nv  I refactored to keep the existing approach, all arguments are positional, None if unused.
We can revisit the kwargs idea separatly from MTP work,

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
galagam: Good point, I moved SSM replay mode flag into BatchInfo

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
galagam: This is still needed at compile time, but BatchInfo.is_use_replay() will be used in runtime. Updated comment to clarify this.

tests/unittest/auto_deploy/singlegpu/smoke/test_ad_speculative_decoding.py
```diff
@@ -92,6 +92,70 @@ def test_super_mtp_smoke():
...
+        "max_tokens": 20,
+        "top_k": None,
+        "temperature": 0.0,
+        "seed": 42,
+    }
+
+    results = main(cfg)
+
+    prompts_and_outputs = results["prompts_and_outputs"]
+    assert len(prompts_and_outputs) == 1
```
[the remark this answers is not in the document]
galagam: added new test test_flashinfer_extend_replay_calls_replay_kernel to cover this

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -400,19 +406,69 @@ def _identify_managed_state_resources(
...
+            conv_ref,
+            conv_managed,
+            conv_spec,
+            replay_old_x,
+            replay_old_B,
+            replay_old_dt,
+            replay_old_dA_cumsum,
+            replay_cache_buf_idx,
+            replay_prev_num_accepted,
+        )
```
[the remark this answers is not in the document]
galagam: fixed

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
```diff
@@ -289,15 +362,43 @@ def get_cache_initializers(
...
+            ret["replay_cache_buf_idx"] = ReplayCacheBufIdxHandler()
+            ret["replay_prev_num_accepted"] = ReplayPrevNumAcceptedHandler()
+        else:
+            ret["intermediate_ssm_state_cache"] = SpecSSMResourceHandler.from_base(ssm_h)
+            ret["replay_old_x"] = None
+            ret["replay_old_b"] = None
+            ret["replay_old_dt"] = None
+            ret["replay_old_da_cumsum"] = None
+            ret["replay_cache_buf_idx"] = None
+            ret["replay_prev_num_accepted"] = None
```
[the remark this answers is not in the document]
galagam: fixed as suggested

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
```diff
@@ -63,7 +74,15 @@ def _flashinfer_cached_ssm(
...
-    ],  # [spec_state_size, max_draft_len+1, num_heads, head_dim, d_state]
+    ],  # [spec_state_size, max_draft_len+1, num_heads, head_dim, d_state]; None in replay mode
+    replay_old_x: Optional[torch.Tensor],  # [max_batch, T, nheads, head_dim]; None in non-replay
+    replay_old_b: Optional[torch.Tensor],  # [max_batch, 2, T, ngroups, dstate]; None in non-replay
+    replay_old_dt: Optional[torch.Tensor],  # [max_batch, 2, nheads, T] fp32; None in non-replay
+    replay_old_da_cumsum: Optional[
+        torch.Tensor
+    ],  # [max_batch, 2, nheads, T] fp32; None in non-replay
+    replay_cache_buf_idx: Optional[torch.Tensor],  # [max_batch] int32; None in non-replay
+    replay_prev_num_accepted: Optional[torch.Tensor],  # [max_batch] int32; None in non-replay
```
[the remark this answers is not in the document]
galagam: fixed

tensorrt_llm/_torch/auto_deploy/shim/interface.py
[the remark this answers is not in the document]
galagam: See https://github.com/NVIDIA/TensorRT-LLM/pull/13725/changes/3713e492b7113bfe8bfa9e95c38378a5eb8dd306

tests/unittest/auto_deploy/singlegpu/smoke/test_ad_speculative_decoding.py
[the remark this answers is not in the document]
galagam: See https://github.com/NVIDIA/TensorRT-LLM/pull/13725/changes/a627870fbf769c156bceb289c2c53e2df707b00c

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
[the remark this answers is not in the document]
galagam: See https://github.com/NVIDIA/TensorRT-LLM/pull/13725/changes/f2c0472c5e0a7e6dc7129587e4db5fcb40005356

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

govind-ramnarayan (approved):
Thanks! Looks good, we can revisit refactoring later

xinhe-nv (approved):
@coderabbitai review the pr as a QA engineer, check the test coverage.

galagam:
Blocked on #14640

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

galagam:

galagam:

galagam:

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"

examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml
```diff
@@ -2,17 +2,17 @@
 #
 runtime: trtllm
 compile_backend: torch-cudagraph
-max_batch_size: 128
-max_seq_len: 65536 # tunable
+max_batch_size: 64
+max_seq_len: 8192
```
Wanli-Jiang: curious why so small `max_seq_len`?

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
```diff
@@ -89,6 +117,15 @@ def _flashinfer_cached_ssm(
     num_prefill_tokens, num_extend_tokens, num_decode_tokens = batch_info.get_num_tokens()
     num_total_tokens = num_prefill_tokens + num_extend_tokens + num_decode_tokens
 
+    # Pre-hoist index type conversions so their kernels run before the extend
+    # causal_conv1d_update kernel rather than landing in the gap between it
+    # and the PDL-dependent replay precompute kernel.
+    slot_idx_i32 = slot_idx.to(torch.int32)
```
Wanli-Jiang, line 125: what is `slot_idx` here? would it be possible to be larger than max_int32?

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
```diff
@@ -303,7 +303,7 @@ def route_requests(
 
         def get_relax_value(req_item):
             scheduling_params = getattr(req_item.request, "py_scheduling_params", None)
-            if scheduling_params is None:
+            if scheduling_params is None or scheduling_params.attention_dp_rank is None:
```
Wanli-Jiang, line 306: ideally attention_dp_rank should be not None if enabled attention_dp, can we double check?

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
```diff
@@ -814,6 +849,31 @@ def get_intermediate_conv_states(self,
...
+        assert not self._use_cpp, "get_replay_old_dA_cumsum is not supported in CppMambaCacheManager"
+        return self._impl.get_replay_old_dA_cumsum(layer_idx)
+
+    def get_replay_cache_buf_idx(self) -> Optional[torch.Tensor]:
+        assert not self._use_cpp, "get_replay_cache_buf_idx is not supported in CppMambaCacheManager"
+        return self._impl.get_replay_cache_buf_idx()
+
+    def get_replay_prev_num_accepted_tokens(self) -> Optional[torch.Tensor]:
+        assert not self._use_cpp, "get_replay_prev_num_accepted_tokens is not supported in CppMambaCacheManager"
+        return self._impl.get_replay_prev_num_accepted_tokens()
```
Wanli-Jiang, line 1009: any further idea to support replay with cppmambacachemanager?

Wanli-Jiang (approved):

tensorrt_llm/_torch/modules/mamba/ssd_combined.py
```diff
@@ -47,9 +47,17 @@
...
+    # tile_shape_mnk_intra2[1:] = (headdim, chunk_size) instead of the correct
+    # tile_shape_mnk_intra1[1:] = (chunk_size, dstate), making gmem K-mode =
+    # chunk_size/16 mismatch smem K-mode = dstate/16. Only compiles when
+    # dstate == chunk_size (masked the bug for NemotronH). See
+    # https://github.com/flashinfer-ai/flashinfer/issues/3397 for upstream fix.
     return (chunk_size in _FLASHINFER_SSD_VALID_M_MODES
             and dstate in _FLASHINFER_SSD_VALID_M_MODES
-            and headdim in _FLASHINFER_SSD_VALID_HEAD_DIMS)
+            and headdim in _FLASHINFER_SSD_VALID_HEAD_DIMS
+            and dstate == chunk_size)
```
Wanli-Jiang, line 84: would it always be dstate==chunk_size for nemotron-H?

galagam: Yes, for NemotronH chunk_size=128, dstate=128 it's not a problem, which is probably why this went unnoticed. 
I filed https://github.com/flashinfer-ai/flashinfer/issues/3397 against FlashInfer to fix this issue, and added the check here to make sure we fallback to triton if the dims don't match.

tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
galagam: Not planned from my side. Do you know what's the projected perf gain?

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
galagam: I double-checked. attention_dp_rank is actually an optional per-request field that defaults to None in tensorrt_llm/scheduling_params.py:L20, so a None rank is expected and valid even when attention_dp is enabled. It means the request hasn't pinned itself to a specific rank and can be routed freely.
There's a similar handling in https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/pyexecutor/request_utils.py#L117-L119

tensorrt_llm/_torch/auto_deploy/custom_ops/mamba/flashinfer_backend_mamba.py
galagam: Per-request cache slot index in [0, max_batch_size], so no overflow in int32.
This is pre-existing cast, just moved up so it won't interfere with PDL later on.

examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml
galagam: This config targets low-concurrency / latency-bound serving - where MTP shines. It's deliberately sized smaller than the throughput-oriented super_v3.yaml (which uses batch 384 / seq 65536).

QiJune (approved):
LGTM

galagam:
/bot run --stage-list "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-Post-Merge-1,DGX_B200-8_GPUs-AutoDeploy-Post-Merge-1"

examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml
```diff
@@ -2,17 +2,17 @@
...
+max_num_tokens: 2048
 enable_chunked_prefill: false # TODO: Support chunked prefill w/ spec dec
 attn_backend: trtllm
 model_factory: AutoModelForCausalLM
 skip_loading_weights: false
 cuda_graph_config:
-  batch_sizes: [1, 2, 4, 8, 16, 24, 32, 64, 128]
+  max_batch_size: 64
```
MrGeva, line 13: why so low?

examples/auto_deploy/model_registry/configs/super_v3_mtp_ws4.yaml
```diff
@@ -0,0 +1,67 @@
+# Config for SuperV3 with MTP speculative decoding, WS=4 (4-GPU EP+BMM sharding).
+# Derived from super_v3_mtp.yaml with world_size: 4 to enable sharding.
+#
+runtime: trtllm
+compile_backend: torch-cudagraph
+world_size: 4
```
MrGeva: why have a yaml per specific ws? maybe call it _scale instead?

MrGeva: or _multi_gpu.yaml

galagam: Thanks for catching this, I didn't mean to commit this file...

examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml
galagam: See https://github.com/NVIDIA/TensorRT-LLM/pull/13725#discussion_r3332501036

What effective values are you seeing for MTP?

examples/auto_deploy/model_registry/configs/super_v3_mtp_ws4.yaml
MrGeva: or maybe just enable sharding it in the examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml since its a no-op in ws 1

PR #13732, opened 2026-05-04
[TRTLLMINF-54][feat] Add typed exception hierarchy + unified classifier

dpitman-nvda:
## Summary by CodeRabbit

* **Infrastructure**
  * Enhanced error classification system in the continuous integration pipeline to better distinguish between infrastructure failures, user-related issues, and pipeline interruptions, improving failure diagnosis and remediation.

## Description

Commit 1 of 5 (classifier-base, additive only).

Adds the new typed-exception machinery alongside the existing classifyInfraFailure() and the four SLURM_/K8S_ pattern lists. No call-site changes -- existing consumers continue to use classifyInfraFailure(); the new classify() is wired up by the next commit. This staging gives the new classes a soak window in which any CPS save/restore surprises can surface even with no consumers exercising them.

Types
- TrtllmCiException base
- InfraFailure with severity (TRANSIENT|PERSISTENT) and scope (SLURM|K8S|BOTH)
- UserFailure (do-not-retry default)
- PipelineInterruption (user abort / FlowInterrupted / SIGTERM)

Severity and scope are static final String constants on InfraFailure rather than Groovy enums to dodge known CPS save/restore enum pickling issues.

PATTERN_CATALOG
- 20 entries, scope-tagged. Behavior-equivalent to the existing classifyInfraFailure(ex, K8S_INFRA_FAILURE_PATTERNS, K8S_INFRA_SINGLE_RETRY_PATTERNS) combinations:
  - SLURM legacy entries (matched both consumers in legacy because the K8s consumer fed the SLURM list as its first pass) -> scope=BOTH (13 entries).
  - K8S legacy entries (matched only when K8s consumer passed extras) -> scope=K8S (7 entries).
  - "CANCELLED" was effectively dead in legacy (listed in single-retry but not in any failure list) -- omitted here. A separate optional commit can activate it with production log evidence.
  - Severity preserved one-for-one: PERSISTENT for the legacy single-retry entries (DUE TO TIME LIMIT, Permission denied, OOMKilled, Connection failed); TRANSIENT for the rest.

flattenThrowable(ex)
- Single cycle-safe walker over .cause AND .getSuppressed(). Replaces the two separate walks in classifyInfraFailure (which were inconsistent: the interrupt walk wasn't cycle-safe and didn't see suppressed exceptions; the pattern walk was cycle-safe and did). Used by classify() for both interrupt detection and pattern matching.

classify(Throwable ex, String scope)
- Returns one of: PipelineInterruption, InfraFailure, UserFailure.
- Scope isolation guard: typed InfraFailure passes through only if ex.scope == scope || ex.scope == BOTH. A SLURM-scoped InfraFailure leaking into a K8s consumer (e.g. inner SLURM retry exhausted, the typed exception bubbles into the outer K8s pod retry that wraps the SLURM dispatcher pod) is wrapped as UserFailure so the outer doesn't re-run a budget that the inner already burned. The original InfraFailure is preserved as cause for forensics.
- Uses Locale.ROOT for case-folding (avoids locale-specific lower-case surprises like the Turkish-I problem).

Behavior of pre-existing call sites is unchanged: nothing yet calls the new classify() or throws a typed exception.

## Test Coverage

N/A, this is a CI change

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/jenkins/L0_Test.groovy
+++ b/jenkins/L0_Test.groovy
@@ -165,6 +165,201 @@ K8S_INFRA_SINGLE_RETRY_PATTERNS = [
 // independently as production telemetry comes in.
 K8S_INFRA_RETRY_MAX = 2
 
+// ============================================================================
+// Typed-exception hierarchy + unified failure classifier.
+//
+// This block is additive — the existing classifyInfraFailure() and the four
+// SLURM_/K8S_ pattern lists above remain in place. Consumers continue to use
+// classifyInfraFailure(); the new classify(ex, scope) is wired up in a follow-
+// up commit (classifier-consumers).
+//
+// Behavior-equivalence note for reviewers: PATTERN_CATALOG below is a
+// scope-tagged restatement of the legacy lists. Every entry that the legacy
+// classifier would have matched (SLURM consumer: SLURM_INFRA_FAILURE_PATTERNS
+// only; K8s consumer: SLURM list + K8S_INFRA_FAILURE_PATTERNS) is reproduced
+// with the same severity. Patterns that were effectively dead in the legacy
+// (CANCELLED — listed in single-retry but not in the failure list) are
+// omitted; they can be activated as a separate, reviewable change.
+//
+// String constants (not Groovy enums) intentionally — Groovy enums declared at
+// pipeline-script top level have historically been tricky across CPS save/
+// restore boundaries when a build resumes from disk.
+// ============================================================================
+
+class TrtllmCiException extends RuntimeException {
+    TrtllmCiException(String msg)              { super(msg) }
+    TrtllmCiException(String msg, Throwable c) { super(msg, c) }
+}
+
+class InfraFailure extends TrtllmCiException {
+    static final String TRANSIENT  = "TRANSIENT"
+    static final String PERSISTENT = "PERSISTENT"   // single-retry-only
+    static final String SLURM      = "SLURM"
+    static final String K8S        = "K8S"
+    static final String BOTH       = "BOTH"
+
+    String severity         // TRANSIENT | PERSISTENT
+    String scope            // SLURM | K8S; never BOTH on a thrown instance
+    String detectedPattern  // "<typed:...>" for direct throws, matched substring for catalog fallback
+
+    InfraFailure(String msg, Throwable c, String sev, String sc, String pat) {
+        super(msg, c)
+        severity = sev
+        scope = sc
+        detectedPattern = pat
+    }
+}
+
+class UserFailure extends TrtllmCiException {
+    UserFailure(String msg, Throwable c) { super(msg, c) }
+}
+
+class PipelineInterruption extends TrtllmCiException {
+    PipelineInterruption(String msg, Throwable c) { super(msg, c) }
+}
+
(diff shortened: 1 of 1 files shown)


jenkins/L0_Test.groovy
```diff
@@ -165,6 +165,188 @@ K8S_INFRA_SINGLE_RETRY_PATTERNS = [
 // independently as production telemetry comes in.
 K8S_INFRA_RETRY_MAX = 2
 
+// ============================================================================
+// Typed-exception hierarchy + unified failure classifier.
+//
+// This block is additive — the existing classifyInfraFailure() and the four
+// SLURM_/K8S_ pattern lists above remain in place. Consumers continue to use
+// classifyInfraFailure(); the new classify(ex, scope) is wired up in a follow-
+// up commit (classifier-consumers).
```
tburt-nv, line 174: Will the legacy lists also be removed in a follow-up PR?

jenkins/L0_Test.groovy
```diff
@@ -165,6 +165,188 @@ K8S_INFRA_SINGLE_RETRY_PATTERNS = [
...
+        return new PipelineInterruption(ex.message ?: "interrupted", ex)
+    }
+    for (t in chain) {
+        if (t.getClass().name.contains("FlowInterruptedException")) {
+            return new PipelineInterruption(ex.message ?: "FlowInterrupted", ex)
+        }
+    }
+    if (ex instanceof hudson.AbortException && ex.message?.contains("script returned exit code 143")) {
+        return new PipelineInterruption(ex.message, ex)
+    }
```
tburt-nv: I think this scenario could occur if the machine shuts down in the middle of testing, which should be classified as a retryable exception at the node-level rather than an intentional pipeline cancellation.

I guess that might look like:
```
return new InfraFailure(ex.message, ex, InfraFailure.TRANSIENT, InfraFailure.BOTH, "script returned exit code 143")
```

Although I wonder if we could achieve the same effect by adding a new row to `PATTERN_CATALOG`.

tburt-nv (approved):

tburt-nv:
I think we could use "/bot run --stage-list" here to limit the testing pipeline to a single stage. Since it's already running, we can just check if any of the stages pass I think.

jenkins/L0_Test.groovy
dpitman-nvda: Yes, see this commit staged on my fork: https://github.com/dpitman-nvda/TensorRT-LLM/compare/typed-exception-3...dpitman-nvda:TensorRT-LLM:typed-exception-4

dpitman-nvda:
> I think we could use "/bot run --stage-list" here to limit the testing pipeline to a single stage. Since it's already running, we can just check if any of the stages pass I think.

Ack, I think it's most of the way through the x86 single GPU test suite, H100s are finishing up now.

jenkins/L0_Test.groovy
dpitman-nvda: Good call, let me push an update.

dpitman-nvda:
/bot skip --comment "The single stage passed, no idea why the pipeline is showing as failed"

PR #13736, opened 2026-05-04
[None][chore] Import torchvision when used, not at module-level.

tfogal:
## Summary by CodeRabbit

* **Refactor**
  * Reorganized imports across multiple model and utility modules for optimized loading behavior.

## Description

torchvision can be a bit of a pain when developing with torch from git, because it pins particular pytorch versions and so installing it will erase any local torch install.

In a large repo like TRT-LLM, a common fix I use is to just remove torchvision from requirements.txt as a local change, and just test with models that do not need the package. That was difficult in TensorRT-LLM due to how it was imported, so I'm curious if you would consider doing something like this change to make that path easier.

## Test Coverage

New tests were added as per 2ez4bz's review.

## GitHub Bot Help


2ez4bz (commented):
Two thoughts:

1. we could consider using `lazy_loader` python package. It seems already to be a transitive dependency of our environments, so maybe we could make it explicit in `requirements.txt`.
2. Even without it - should be add a test or a lint stage that checks `torchvision` is never imported at the module-level? Otherwise, you might need to keep making such changes in the future.

tfogal:
I like your thoughts, thanks!

> 1. we could consider using `lazy_loader` python package. It seems already to be a transitive dependency of our environments, so maybe we could make it explicit in `requirements.txt`.

Oh, interesting, I wasn't familiar with `lazy_loader`.
I am happy to do this but I'm new here and not sure on conventions; can I just add arbitrary packages or is there a wider review process for that?

> 2. Even without it - should be add a test or a lint stage that checks `torchvision` is never imported at the module-level? Otherwise, you might need to keep making such changes in the future.

Great point. I think I see a clever similar test in tests/unittest/standalone/test_standalone_import.py. I'll add this.

tfogal:
> I think I see a clever similar test in tests/unittest/standalone/test_standalone_import.py. I'll add this.

Pushed this. Still working on 1.

tests/unittest/test_lazy_torchvision_imports.py
2ez4bz: Isn't it too conservative to just check these particular modules? If I add a new `modeling_foo.py` that directly imports `torchvision` then this won't catch it.

2ez4bz (approved):

tests/unittest/test_lazy_torchvision_imports.py
tfogal: Yeah, true; we'd have to add the new file[s] manually in this approach.

Would you like us to glob the files there and try each? That'd be robust against that problem at the cost of a bit more I/O. Or is there another solution I'm not seeing?

2ez4bz: Yeah I think globbing against `*.py` under the `tensorrt_llm/_torch/models/` and `tensorrt_llm/inputs` directories would be a pretty decent first line of defense.

amukkara (approved):
Approving phi4mm changes.

tfogal:
> > I think I see a clever similar test in tests/unittest/standalone/test_standalone_import.py. I'll add this.
> 
> Pushed this. Still working on 1.

Apologies forgot to post here; I had updated it to use lazy.load earlier. Latest push just fixes some hook issues.

... but now I see there's a conflict, I'll address that in a sec, but should be fine to review again already.

tfogal:
/bot run

Previous run failed due to infra issues; the 3 systems that ran to completion succeeded.

yechank-nvidia (approved):

tfogal:
> Primary (Infrastructure): Investigate the rogue process (PID 2125557) that was consuming 20.18 GiB of the A10 GPU during the test run. This is likely a leaked process from a previous job on the same node. The CI infrastructure team should ensure proper GPU process cleanup between test jobs on the A10 node (a4u8g-0012.ipp1a1.colossus).
>
> For the PR author (tfogal): The PR itself is not the root cause of the failures. The lazy_loader dependency was successfully installed. The failing tests (test_torch_multi_arange, test_openai_chat_multimodal_example) failed due to A10 GPU memory exhaustion from an external process — not from the import refactoring changes. Re-triggering CI should pass if the GPU node is clean.

Going to merge main in hopes that this was either already fixed, or a transient infra failure.

PR #13745, opened 2026-05-05
[None][feat] Support Gemma4 multi-head_dim pools and host-side slicing to provide local view to Triton kernels for SWA

eopXD:
## Summary

- For the V1 C++ KV cache manager: Previously multi-pool has the implicit assumption that they all operate on same head_dim. To run Gemma4 correctly, this needs to be relaxed. (e.g. Gemma-4 MoE: SWA layers with `head_dim=256` + global-attention layers with `head_dim=512`) 
- Further more for the V1 C++ KV cache manager, it originally operates on the assumption of a window mapping to a single pool. The window -> pool mapping should not be the correct set-up of memory pool since we should provide flexibility to all three attributes - (window_size, dtype, head_dim). This MR abstracts out a "PoolConfiguration" concept for the v1 manager in order to further provide flexibility in pool management.
- For the auto deploy, attention grouping and kernel call plumbing originally operates on a single pool assumption. This does not match the multi-pool set-up we hope to achieve for modern LLM models. This MR plumbs to support multi-pool attenion grouping of the kv cache manager.
- For triton kernel, sliding-window attention will result in front-evicting pages. The kernels will require to have a local view of the pool as the KV cache block writes forwards, leaving the front blocks out. Such local view has to be supported within this PR because previously gemma3 model running MMLU (including long sequence requests) have been added to the AutoDeploy support.

## Test coverage

- [x] Existing VSWA model test cases should pass (gemma3-e2n) under the changes done in this merge request
- [x] Unit tests for triton kernel local view
    - For prefill, the full sequence is added.
    - For decode, front blocks are evicted.
- [x] test_cached_sequence_interface.py updated, multi-head_dim pool configuration is now managed.
-  Currently gemma4 model run is still waived out for other existing problems


eopXD:
/bot run --extra-stage "DGX_H100-1_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1"

cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h
```diff
@@ -896,6 +927,29 @@ class WindowBlockManager
...
+    //!          head_dim (the constructor's @c sizePerHead), so we read it from the
+    //!          first pool.  Returns 0 if no pools are present.
+    [[nodiscard]] SizeType32 getSizePerHead() const noexcept
+    {
+        if (mPools.empty())
+        {
+            return 0;
+        }
+        return mPools.front().sizePerHead;
+    }
```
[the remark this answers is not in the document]
eopXD: **Problem.** `WindowBlockManager` builds recurrent-state pools with `sizePerHead = -1` (kvCacheManager.cpp:803). The new `getSizePerHead()` accessor returns that sentinel directly and surfaces it through `getSizePerHeadPerWindow()` and the nanobind `size_per_head_per_window` property, so any caller iterating all windows could observe `-1` as a real head dimension and mis-shape buffers.

**Fix.** Guarded `getSizePerHead()` to also return `0` when `mPools.front().sizePerHead <= 0`, mirroring the existing empty-pool branch. Updated the comment to record that recurrent pools use `-1` as a sentinel and intentionally have no meaningful head dimension.

Applied in b6db6dea20.

examples/auto_deploy/model_registry/configs/gemma4_moe.yaml
```diff
@@ -10,6 +10,7 @@ attn_backend: triton_paged
 compile_backend: torch-cudagraph
 cuda_graph_config:
   batch_sizes: [1, 2, 4, 8, 16, 32, 64, 128, 256, 512]
+  max_batch_size: 512
```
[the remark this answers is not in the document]
eopXD: **Bot's claim.** The nested `cuda_graph_config.max_batch_size: 512` duplicates the top-level `max_batch_size: 512` and should be removed for consistency with `qwen3.5_moe_35b.yaml` etc.

**Why we're not changing it.** `CudaGraphConfig.max_batch_size` is a documented field on the Pydantic schema (`tensorrt_llm/llmapi/llm_args.py:117`) with semantics distinct from the top-level engine `max_batch_size` — it's the cuda-graph-capture cap. The post-init validator at `llm_args.py:142` enforces compatibility (`cuda_graph_config.max_batch_size == max(batch_sizes)` when both are explicit). The other configs that omit the nested value rely on auto-derivation from `batch_sizes`; our config sets it explicitly so the cap is unambiguous to anyone reading the file. This is style-only — no functional bug to fix here.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1187,6 +1317,44 @@ def nest_sequences(
...
+                self._stage_arg(f"cache_loc{suffix}", group_cache_loc)
+                if cu_num_pages_per_group is not None:
+                    self._stage_arg(f"cu_num_pages{suffix}", cu_num_pages_per_group[group_idx])
+                    self._stage_arg(f"last_page_len{suffix}", lpl_host)
+                if extra_page_per_seq_per_group is not None:
+                    self._stage_arg(
+                        f"extra_page_per_seq{suffix}", extra_page_per_seq_per_group[group_idx]
+                    )
+                if kv_page_offset_per_group is not None:
+                    self._stage_arg(f"kv_page_offset{suffix}", kv_page_offset_per_group[group_idx])
```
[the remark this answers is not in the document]
eopXD: **Problem.** Once `cache_loc_per_group is not None`, the group-0 fallback (the `if cache_loc_per_group is None` branch) never runs, and any group omitted from the per-group dicts silently reuses the previous batch's staged `cache_loc_g*` / `cu_num_pages_g*` / `extra_page_per_seq_g*` / `kv_page_offset_g*`. That's a silent correctness footgun against any future caller that forgets a group.

**Caveat.** The current sole caller (`ad_executor.py:864-867`) initializes the per-group dicts with all groups `1..N-1` and populates each in the per-pool loop (lines 906-924), so partial maps cannot occur today.

**Fix.** Added an `assert` over each provided `*_per_group` map that every group in `range(1, num_window_groups)` is present, so a future regression fails loud at staging time rather than at decode time. This matches the spirit of the bot's suggestion without forcing an opinionated fill-from-group-0 policy on the per-group API.

Applied in b6db6dea20.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/triton_paged_attention.py
```diff
@@ -1137,6 +1160,18 @@ def triton_paged_mha_with_cache(
...
+        # the context kernel computes page iteration bounds from global
+        # seq_len, overflowing the windowed cache_loc.
+        # seq_len_with_cache = cache_len + q_len, where cache_len is the
+        # number of prior-cached tokens.  Only cache_len needs capping.
+        q_lens = cu_seqlen[1 : num_prefill + 1] - cu_seqlen[:num_prefill]
+        page_counts = cu_num_pages[1 : num_prefill + 1] - cu_num_pages[:num_prefill]
+        max_cached = page_counts * kv_cache.shape[3]  # pages × page_size
+        cache_len_raw = seq_len_with_cache - q_lens
+        cache_len_capped = torch.minimum(cache_len_raw, max_cached)
+        seq_len_with_cache = cache_len_capped + q_lens
```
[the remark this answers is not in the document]
eopXD: **Problem.** `max_cached = page_counts * page_size` over-counts last-page slack when `last_page_len < page_size`. After the cap fires, Phase 1 of `_paged_context_kernel` (the `range(first_window_page, num_full_pages)` loop) computes `num_full_pages = first_q_kv_pos // PAGE_SIZE` from the (now-overcounted) `cache_len`, folds the final cached page into the no-mask iteration, and reads the slack slots whose KV positions still satisfy the causal mask (`kv_pos < cache_len`) but whose contents are uninitialized. Phase 2's per-token validity mask doesn't catch this because Phase 1 has already eaten the last cached page.

**Fix.** Replaced the cap with the suggested form:

```python
max_total_tokens = (page_counts - 1) * page_size + last_page_len[:num_prefill]
max_cached = torch.clamp(max_total_tokens - q_lens, min=0)
```

`last_page_len` here is computed from `(ip + sl - 1) % page_size + 1` upstream in `attention_interface.py:1317`, so it matches the last page of the full sequence (which, post-eviction, is also the last page of `cache_loc`). With this cap, `cache_len = (page_counts-1)*page_size + last_page_len`, so `num_full_pages = page_counts - 1` and the final cached page falls into Phase 2 where the validity mask gates the slack slots correctly.

Applied in b6db6dea20.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -853,16 +877,68 @@ def _prepare_inputs(
...
+                        cache_loc_per_group[pool_idx].extend(active_indices)
+                        prev = cu_num_pages_per_group[pool_idx][i]
+                        cu_num_pages_per_group[pool_idx].append(prev + num_active)
+                        kv_page_offset_per_group[pool_idx].append(page_offset_g)
+                        if len(all_indices) > front_removed + num_active:
+                            extra_page_per_seq_per_group[pool_idx].append(
+                                all_indices[front_removed + num_active]
+                            )
+                        else:
+                            extra_page_per_seq_per_group[pool_idx].append(-1)
```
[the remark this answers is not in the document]
eopXD: **Problem.** With `num_active = len(all_indices[front_removed:])`, we have `front_removed + num_active == len(all_indices)` identically. The branches at lines 910 and 919 (`if len(all_indices) > front_removed + num_active`) are always false, so `extra_page_per_seq*` is always `-1` in the VSWA path. The single-pool path (line 939) avoided this by computing `num_active_blocks_i = get_num_kv_blocks(end_compute_i)` independently, leaving the trailing reserved page available to slice off as the spillover.

**Fix.** Compute the active block count from the actual post-eviction token count, then slice `all_indices` into active vs. extra:

```python
active_token_count = min(end_compute_i, group_window)
pages_for_active = (active_token_count + tokens_per_block - 1) // tokens_per_block
num_active = max(0, len(all_indices) - front_removed)
num_active = min(num_active, pages_for_active)
active_indices = all_indices[front_removed : front_removed + num_active]
```

When `len(all_indices) > front_removed + num_active`, the trailing entry now correctly populates `extra_page_per_seq*`, giving VSWA writes a real spillover target.

Applied in b6db6dea20.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -118,8 +118,17 @@ def __init__(
...
+        # A single KVCacheManager hosts all KV pools (full-attention + per-window
+        # SWA) via the C++ per-window head_dim/dtype overrides; cross-pool
+        # admission, event flush, disagg transfer, and the shared radix tree all
+        # live in C++.
         self._kv_cache_manager: Optional[Union[KVCacheManager, MambaHybridCacheManager]] = None
+        # Per-pool sliding window sizes, published by the kvcache transform and
+        # consumed by the executor.  Pool index == position in this list, in the
+        # same order as the C++ manager's internal pool ordering (see
+        # ``size_per_head_per_window`` keys).
+        self._kv_group_windows: List[int] = []
```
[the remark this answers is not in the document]
eopXD: **Bot's claim.** The file lacks the NVIDIA/Apache copyright header.

**False positive.** The header is present at lines 1-14 of `tensorrt_llm/_torch/auto_deploy/shim/interface.py` and has been there since the very first commit on this branch (`feedb6afe7`). Verified via `git show feedb6afe7:tensorrt_llm/_torch/auto_deploy/shim/interface.py | head -5`. No change needed.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -634,13 +776,70 @@ def _allocate_unmanaged_resources(self) -> None:
...
+        # If the caller didn't pass a budget and there's no SWA pool, defer to
+        # free_gpu_memory_fraction inside the manager — same as the legacy path.
+        if total_max_tokens is None and not self._has_swa_window(size_per_head_per_window):
+            return None
+
+        if total_max_tokens is None:
+            # SWA present but no total budget — conservative single-sequence estimate.
+            # The C++ side will still bound this against available memory.
+            return self.info.max_seq_len
+
```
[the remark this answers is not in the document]
eopXD: **Problem.** When the caller passes `total_max_tokens=None` and an SWA pool is present, `_compute_total_token_budget` returns `self.info.max_seq_len`. `_prepare_kv_cache_config` then runs `kv_cache_config.max_tokens = min(kv_cache_config.max_tokens or max_tokens, max_tokens)`, silently shrinking any user-pinned `KvCacheConfig.max_tokens` (e.g. 10 000) down to one sequence's worth of tokens (`max_seq_len`).

This fires on the `else` branch of `initialize_resources` at `interface.py:1043`, where the comment claims we'll "use the original settings in kv_cache_config" — but for SWA models that promise was being broken.

**Fix.** Early-return `None` from `_compute_total_token_budget` when `self._kv_cache_config_original.max_tokens is not None`, so `_prepare_kv_cache_config` keeps the user's value untouched. The conservative `max_seq_len` fallback still kicks in for SWA models where the user did not pin a budget.

Applied in b6db6dea20.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -683,35 +903,50 @@ def _create_kv_cache_manager(self, max_tokens: Optional[int] = None) -> Dict:
...
         self._kv_cache_config_tuned = kv_cache_config
 
-        # 5. Assign KV views (compute block_offset_multiplier from first view's strides)
-        block_offset_multiplier = self._assign_kv_cache_views(kv_managed)
+        # 5. Assign KV views and update cache information.
+        block_offset_multiplier = 0
+        if kv_managed:
+            block_offset_multiplier = self._assign_kv_cache_views(
+                kv_managed, self._kv_cache_manager
+            )
```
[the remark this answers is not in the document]
eopXD: **Problem.** `KVCacheManager.__init__` clamps every window via `min(w, max_seq_len)` (`resource_manager.py:444-446`) before constructing the C++ pools. `_kv_group_windows`, however, is set at transform time from `handler.sliding_window` (`kvcache.py:404-407`) and never refreshed. Downstream callers in `ad_executor.py:888-893` then feed those stale values into `get_cache_indices(window_size=...)` / `get_num_front_blocks_removed(...)`, which would hit `mWindowBlockManagers.at(window).out_of_range` for any window > `max_seq_len`.

Gemma4-26B doesn't trip this today (`sliding_window=4096 < max_seq_len=8192`), but the staleness is a real footgun for any future model that declares `sliding_window > max_seq_len`.

**Fix.** After `KVCacheManager` construction, refresh `_kv_group_windows` from the manager's authoritative `max_attention_window_vec`, deduped via `dict.fromkeys` to preserve first-occurrence order (matching the C++ pool ordering). A log line fires when the values actually change, so a regression would be visible immediately.

Applied in b6db6dea20.

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
```diff
@@ -216,41 +269,98 @@ def _apply(
...
+
+        # --- Pass 2: insert cached attention nodes with correct group metadata ---
+        for idx, (attn_node, qkv, cache_in_nodes, constants, group_idx) in enumerate(layer_infos):
+            layer_meta_nodes_std = meta_nodes_std
+            if is_multi_group and group_idx > 0:
+                std_arg_names = self.attn_descriptor.get_standard_metadata_args()
+                layer_meta_nodes_std = list(meta_nodes_std)
+                for arg_pos, arg_name in enumerate(std_arg_names):
+                    if arg_name in vswa_group_nodes.get(group_idx, {}):
+                        layer_meta_nodes_std[arg_pos] = vswa_group_nodes[group_idx][arg_name]
```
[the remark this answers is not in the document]
eopXD: **Problem.** Pass 2 (`kvcache.py:435-453`) only swaps `layer_meta_nodes_std`. For backends like `TrtllmAttention`, the group-sensitive inputs flow through `meta_nodes_extra` via `prepare_trtllm_metadata(batch_info_host, cu_num_pages, cache_loc)`, so every non-zero group still references the block-offset node built from group-0 `cache_loc` / `cu_num_pages` — broken VSWA routing for those layers.

**Why we're not addressing this PR.** This PR's CI exercises the `triton_paged` backend (see `examples/auto_deploy/model_registry/configs/gemma4_moe.yaml`), whose `prepare_triton_paged_metadata` only consumes group-agnostic inputs (`cu_seqlen`, `seq_len_with_cache`), so the test surface here is unaffected. The `TrtllmAttention` + VSWA path is genuinely broken but **not regressed by this PR** — it never worked.

**Follow-up plan.** Tracked as a follow-up: build per-group `meta_nodes_extra` alongside `layer_meta_nodes_std` (mirror the `vswa_group_nodes[group_idx]` mapping into `meta_nodes_extra` for any arg that is in `vswa_swappable_bases`), and add a multi-window `TrtllmAttention` regression test. Keeping it out of this PR to avoid scope creep.

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
```diff
@@ -216,41 +269,98 @@ def _apply(
...
+                    base = arg_name.removesuffix(host_suffix)
+                    if base in vswa_swappable_bases:
+                        is_host = arg_name.endswith(host_suffix)
+                        group_arg = f"{base}_g{gi}{host_suffix if is_host else ''}"
+                        vswa_group_nodes[gi][arg_name] = self._add_or_retrieve_input(
+                            gm, cm, group_arg
+                        )
+
+        # --- Pass 2: insert cached attention nodes with correct group metadata ---
+        for idx, (attn_node, qkv, cache_in_nodes, constants, group_idx) in enumerate(layer_infos):
```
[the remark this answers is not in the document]
eopXD: **Problem.** Both Pass-1 and Pass-2 loops in this transform use `enumerate()` but never read the index, so Ruff B007 keeps flagging them.

**Fix.** Dropped `enumerate()` on both loops (`for attn_node in source_attn_nodes:` and `for (attn_node, qkv, ..., prepared_mask) in layer_infos:`). No behavior change; B007 cleared.

Applied in b6db6dea20.

tests/unittest/auto_deploy/singlegpu/transformations/library/test_kv_cache.py
```diff
@@ -600,3 +600,575 @@ def test_insert_cached_mla_attention_preserves_non_flashinfer_backend():
...
+        tokens_per_block=max_seq_len,
+        max_tokens=batch_size * max_seq_len,
+        free_gpu_memory_fraction=0.0,
+    )
+    cm = CachedSequenceInterface(
+        max_seq_len=max_seq_len,
+        max_batch_size=batch_size,
+        device="cuda",
+        kv_cache_config=kv_cache_config,
+    )
```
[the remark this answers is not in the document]
eopXD: **Problem.** `CachedSequenceInterface.__init__` requires `max_num_tokens` (made positional-required by upstream commit `253f17eab5`, well before this branch's base). Seven of the new VSWA test instantiations introduced in this PR — at lines 983, 1017, 1049, 1082, 1128, 1205, 1328 — omit it, so those tests would `TypeError` at runtime.

**Fix.** Added `max_num_tokens=default_max_num_tokens(max_seq_len, batch_size)` to all seven instantiations, matching the existing pattern earlier in the file. After the change, all 13 `CachedSequenceInterface(` calls in the test file pass `max_num_tokens`.

Applied in b6db6dea20.

eopXD:
/bot run --extra-stage "DGX_H100-1_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1"

eopXD:
/bot run --extra-stage "DGX_H100-1_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1"

eopXD:
/bot run --extra-stage "DGX_H100-1_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-1_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

nvpohanh:
@SimengLiu-nv @VALLIS-NERIA could you help to review this? Thanks!

StanleySun639 (approved):

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

SimengLiu-nv (commented):
## Concern 1: Restore compatibility for direct `triton_paged_context` callers

The PR makes `kv_page_offset` a required positional argument before `sm_scale` in `triton_paged_context`:

https://github.com/NVIDIA/TensorRT-LLM/blob/09fd33d9c2d16bf082529405724d483f566cab86/tensorrt_llm/_torch/auto_deploy/custom_ops/attention/triton_paged_attention.py#L1060-L1069

Existing callers in the Triton paged attention tests still use the old call shape, passing `seq_len_with_cache` followed directly by `sm_scale`:

https://github.com/NVIDIA/TensorRT-LLM/blob/09fd33d9c2d16bf082529405724d483f566cab86/tests/unittest/auto_deploy/singlegpu/custom_ops/attention/test_triton_paged_attention.py#L311-L320

https://github.com/NVIDIA/TensorRT-LLM/blob/09fd33d9c2d16bf082529405724d483f566cab86/tests/unittest/auto_deploy/singlegpu/custom_ops/attention/test_triton_paged_attention.py#L1063-L1073

Those calls now bind `sm_scale` as `kv_page_offset` and leave the required `sm_scale` argument missing, so the existing tests fail before the kernel launches. Please either make `kv_page_offset` optional with a zero-vector default for the old direct-call path, or update all direct callers to pass an explicit offset tensor.

## Concern 2: Keep query and KV positions in the same coordinate space after SWA eviction

The new VSWA path slices each window's page table after evicted front blocks and passes `front_removed` as `kv_page_offset`:

https://github.com/NVIDIA/TensorRT-LLM/blob/09fd33d9c2d16bf082529405724d483f566cab86/tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py#L907-L920

Then `triton_paged_mha_with_cache` caps `seq_len_with_cache` to the local window length:

https://github.com/NVIDIA/TensorRT-LLM/blob/09fd33d9c2d16bf082529405724d483f566cab86/tensorrt_llm/_torch/auto_deploy/custom_ops/attention/triton_paged_attention.py#L1399-L1407

Inside `_paged_context_kernel`, KV positions are converted back to global coordinates using `page_offset`, but query positions and validity math still use the capped local `cache_len` / `total_kv_len`:

https://github.com/NVIDIA/TensorRT-LLM/blob/09fd33d9c2d16bf082529405724d483f566cab86/tensorrt_llm/_torch/auto_deploy/custom_ops/attention/triton_paged_attention.py#L798-L806

https://github.com/NVIDIA/TensorRT-LLM/blob/09fd33d9c2d16bf082529405724d483f566cab86/tensorrt_llm/_torch/auto_deploy/custom_ops/attention/triton_paged_attention.py#L821-L826

Once `front_removed > 0`, this compares local query positions against global KV positions and can make valid boundary pages look invalid, producing incorrect causal/SWA masking during long chunked prefill. Gemma4 is configured to use `triton_paged`, so this is on the target path:

https://github.com/NVIDIA/TensorRT-LLM/blob/09fd33d9c2d16bf082529405724d483f566cab86/examples/auto_deploy/model_registry/configs/gemma4_moe.yaml#L9

Please keep separate values for local page iteration bounds and global token coordinates, or add `kv_page_offset * page_size` back to the query-side position math. This should also get a regression test where `kv_page_offset > 0`.

chuangz0 (approved):

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
suyoggupta, line 1215: I think we need MMMU here for the multimodal version of gemma4
@bmarimuthu-nv : please review.

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
```diff
@@ -316,20 +384,75 @@ def _apply(
...
+                constants.append(sink_token_length)
+
+            layer_infos.append(
+                (attn_node, qkv, cache_in_nodes, constants, group_idx, prepared_mask)
+            )
+
+        # --- Multi-group setup: register metadata groups and create graph placeholders ---
+        num_groups = len(handler_groups)
+        is_multi_group = num_groups >= 2
+        vswa_group_nodes: dict[int, dict[str, "Node"]] = {}
```
suyoggupta: @eopXD : are these changes still necessary given that we don't need multiple pools?

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
```diff
@@ -14,7 +14,30 @@
...
+    (``register_window_groups``), allocate per-group metadata placeholders for
+    groups ``1..N-1``, and insert cached attention ops with their layer's
+    group's metadata nodes.
+
+After the transform, ``group_idx == pool_idx`` holds everywhere: the executor
+reads per-pool windows from ``cache_seq_interface.kv_group_windows`` and
+issues per-window queries (``get_cache_indices(request, window_size=W)``)
+against the unified ``KVCacheManager``; the C++ side routes each call to the
+correct pool via its ``mLayerToWindowSize`` map.
+"""
```
suyoggupta, line 40: stale now?

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
```diff
@@ -1181,43 +1122,49 @@ def test_nvfp4(self, ep_size, attention_dp):
...
+        sampling_params = SamplingParams(end_id=None,
+                                         pad_id=None,
+                                         n=1,
+                                         use_beam_search=False)
         with AutoDeployLLM(model=self.MODEL_PATH,
                            tokenizer=self.MODEL_PATH,
                            world_size=registry_world_size,
                            yaml_extra=yaml_paths) as llm:
-            task = MMMU(self.MODEL_NAME)  # noqa: F821
+            task = MMLU(self.MODEL_NAME)
```
bmarimuthu-nv: Can we both the changes? Keep MMMU, MMLU and GSM8k

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -1001,6 +1014,8 @@ def get_constants(cls, source_attn_node: Node) -> List[Constant]:
             num_heads_hint,
             num_kv_heads_hint,
             head_dim_hint,
+            # Note: sink_token_length is a runtime config (KvCacheConfig), not a model
+            # parameter.  It is appended by the kvcache transform at insertion time.
```
nvchenghaoz: duplicate comments... Can we delete these two lines? Or delete the 954 - 955.

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
eopXD: I don't understand the comment here. May you elaborate further?

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
eopXD: Multiple pooling still exists, but within the KV cache manager.
The grouping is still necessary.

May you elaborate further on your concerns?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
eopXD: Removed. Thank you for the catch.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
eopXD: Kept. Thank you for the catch.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
eopXD: Recovered.

eopXD:
@SimengLiu-nv Thank you for carefully auditing the kernel changes. It is indeed incorrect.
To make the scope of change tight, I will limit the change of this MR to only V1 kv cache manager relaxation for multi-head_dim support and its related AutoDeploy plumbing.

For your concerns of the kernels, I will have a follow-up PR to resolve that.

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

cpp/tensorrt_llm/batch_manager/cacheTransBuffer.cpp
```diff
@@ -253,6 +253,28 @@ CacheTransBufferManager::CacheTransBufferManager(
 {
     // TODO: FP4 dataSize
     TLLM_CHECK(mCacheManager);
+    // TODO(disagg-multi-dtype): Per-pool dtype dispatch in formatter / transfer buffer
```
SimengLiu-nv, line 256: Can you guard the unsupported cases? Looks like a silent failure now.

SimengLiu-nv (approved):
The KV cache part looks good to me with one comment.
Approve base on the promise that concerns here https://github.com/NVIDIA/TensorRT-LLM/pull/13745#pullrequestreview-4278771198  will be resolved in follow-up PR by https://github.com/NVIDIA/TensorRT-LLM/pull/13745#issuecomment-4447492340.

cpp/tensorrt_llm/batch_manager/cacheTransBuffer.cpp
eopXD: The new code segment right under the TODO is the guard.
Marking this thread as resolved. Thank you for the review.

yechank-nvidia (approved):
LGTM as torch-models-devs. This PR only touches AD section, and it was covered by AD reviewers.

nv-guomingz (approved):

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

danielafrimi (approved):

tensorrt_llm/_torch/auto_deploy/transform/library/kvcache.py
eopXD: Resolving it for now. Please raise if there is any further question.

cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp
```diff
@@ -9798,3 +9799,513 @@ TEST_F(KVCacheManagerTest, KvCacheConnector_DecodeBlockBoundary_ParityWithBaseli
             << ", withConnector=" << withConnector[i] << "); see issue #13320";
     }
 }
+
+// =============================================================================
+// Per-window head_dim / dtype overrides on BlockManager and KVCacheManager,
+// plus per-window front-eviction bookkeeping on GenerationRequest.
+// Production surface added in commits 1c273f0594 and 823d6ea0e1; the previous
```
eopXD: Remove comment that points to specific commit(s). Please state the code behavior ground-up and avoid cross-referencing.

cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp
```diff
@@ -9798,3 +9799,513 @@ TEST_F(KVCacheManagerTest, KvCacheConnector_DecodeBlockBoundary_ParityWithBaseli
...
+// tests (KVCacheManagerVariableWindowAttentionWithReuseTest, FP4BlockScaleManagementTest)
```
eopXD: Please check on `KVCacheManagerVariableWindowAttentionWithReuseTest` and `FP4BlockScaleManagementTest`, do these tests really need to exist now?

cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp
```diff
@@ -9798,3 +9799,513 @@ TEST_F(KVCacheManagerTest, KvCacheConnector_DecodeBlockBoundary_ParityWithBaseli
...
+// only exercise uniform-shape pools.
+// =============================================================================
+
+namespace
+{
+//! \brief Find the absolute pool index whose window size matches @p windowSize.
+SizeType32 findPoolIdxForWindow(BlockManager const& blockManager, SizeType32 windowSize)
```
eopXD: Why do we need this extra function? Can't we just query BlockManager for the window size and non-existing window sizes are already handled with error?

cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp
```diff
@@ -9798,3 +9799,513 @@ TEST_F(KVCacheManagerTest, KvCacheConnector_DecodeBlockBoundary_ParityWithBaseli
...
+    auto const largePoolIdx = findPoolIdxForWindow(blockManager, largeWindow);
+    EXPECT_EQ(blockManager.getPool(smallPoolIdx).sizePerHead, smallSizePerHead);
+    EXPECT_EQ(blockManager.getPool(largePoolIdx).sizePerHead, largeSizePerHead);
+    EXPECT_NE(blockManager.getPool(smallPoolIdx).sizePerHead, blockManager.getPool(largePoolIdx).sizePerHead);
+    // blockSize = numKvHeads * sizePerHead * tokensPerBlock — should also differ accordingly.
+    EXPECT_EQ(blockManager.getPool(smallPoolIdx).blockSize, numKvHeads * smallSizePerHead * tokensPerBlock);
+    EXPECT_EQ(blockManager.getPool(largePoolIdx).blockSize, numKvHeads * largeSizePerHead * tokensPerBlock);
+}
+
+TEST_F(KVCacheManagerTest, BlockManagerTestPerWindowDataType)
```
eopXD: Please fuse this testing to the unit test above. We don't need separate unit tests. We can extend another window size that contains a different data type.

cpp/tests/unit_tests/batch_manager/kvCacheManagerTest.cpp
```diff
@@ -9798,3 +9799,513 @@ TEST_F(KVCacheManagerTest, KvCacheConnector_DecodeBlockBoundary_ParityWithBaseli
...
+    // so the small window's block count stays the same and the large window's halves.
+    EXPECT_EQ(overSmallPrim, baselineSmallPrim);
+    EXPECT_EQ(overSmallSec, baselineSmallSec);
+    EXPECT_LT(overLargePrim, baselineLargePrim);
+    EXPECT_LT(overLargeSec, baselineLargeSec);
+    // Large window pays double bytes/token under override → roughly half the blocks.
+    EXPECT_NEAR(static_cast<double>(overLargePrim) / baselineLargePrim, 0.5, 0.01);
+}
+
+TEST_F(KVCacheManagerTest, KVCacheManagerSWAEvictionCountPerWindow)
```
eopXD, line 9913: Do we have existing tests that tests for front-eviction for the SWA scheme? If so, please combine this test with the existing ones.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/flashinfer_attention.py
```diff
@@ -577,10 +577,18 @@ def get_prepare_extra_metadata_info(
     def get_cache_initializers(
         cls, source_attn_node: Node, cache_config: KvCacheConfig
     ) -> ResourceHandlerDict:
+        """Build the per-layer KV handler used by the kvcache transform.
+
+        ``sliding_window`` is propagated into the handler so that handler
```
eopXD: Too long, just say:
```
``sliding_window`` is propagated into the handler so layers
with different windows land in separate pools.
```

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/triton_paged_attention.py
```diff
@@ -1495,9 +1495,17 @@ def get_prepare_extra_metadata_info(
     def get_cache_initializers(
         cls, source_attn_node: Node, cache_config: KvCacheConfig
     ) -> ResourceHandlerDict:
+        """Build the per-layer KV handler used by the kvcache transform.
+
+        ``sliding_window`` is read from the source attention node and passed into
```
eopXD: Ditto, please try to  comment in a straight-forward fashion.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py
```diff
@@ -853,7 +845,12 @@ def get_standard_metadata_args(cls) -> List[str]:
     def get_cache_initializers(
         cls, source_attn_node: Node, cache_config: KvCacheConfig
     ) -> ResourceHandlerDict:
-        """Return only KV cache handler (no workspace handler, managed like flashinfer)."""
+        """Return only the KV cache handler (no workspace handler; managed like flashinfer).
+
+        ``sliding_window`` is propagated into the handler so that handler equality
```
eopXD: Ditto, straight-forward wording.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1602,23 +1792,34 @@ def __init__(
...
             kv_layout: Memory layout - "HND" or "NHD". Default is "HND".
+            sliding_window: Sliding window size for this layer. 0 means full attention.
         """
         self.num_kv_heads = num_kv_heads
         self.head_dim = head_dim
         self.dtype = dtype
         self.kv_factor = kv_factor
         assert kv_factor in [1, 2], f"Invalid kv_factor: {kv_factor}"
         self.kv_layout = kv_layout
+        self.sliding_window = (
```
eopXD, line 1894: Why do we need the `if-else` here if a default value is already set?

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -658,17 +837,46 @@ def _prepare_inputs(
...
+        # on SequenceInfo).  Per-window queries on the manager route to the
+        # correct C++ pool via mLayerToWindowSize.
+        kv_group_windows = self.cache_seq_interface.kv_group_windows
+        is_vswa = len(kv_group_windows) >= 2
+
+        # Cache hot lookups so the per-request loop avoids repeated C++
+        # dispatch / hasattr calls.
         _tokens_per_block = kv_cache_manager.tokens_per_block
         _use_mamba = hasattr(kv_cache_manager, "mamba_cache_index")
+
```
eopXD: Unnecessary extra line here.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -658,17 +837,46 @@ def _prepare_inputs(
...
         state_slot_idx: List[int] = []
 
-        batch_cache_indices = kv_cache_manager.get_batch_cache_indices(
-            [r.py_request_id for r in ordered_requests]
-        )
+        # Per-pool variants for VSWA (pools 1..N-1); pool 0 reuses the base
+        # lists above.  These are the data-transport shape that the kernels
+        # consume; their source is now a single per-window get_cache_indices()
+        # call against the unified manager.
+        cache_loc_per_group: Optional[Dict[int, List[int]]] = None
```
eopXD: I think we should generalize this logic.

We should not have a diverging logic of "vswa" using `cache_loc_per_group`, `cu_num_pages_per_group`, `extra_page_per_seq_per_group` and full attention using `cache_loc`, `cu_num_pages`, and `extra_page_per_seq`.

Full attention should be a case covered by VSWA logic.
Please re-approach the code and make the logic sound for both VSWA and single window full attention.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -133,8 +133,17 @@ def __init__(
 
         self._resource_lookup: ResourceHandlerDict = {}
         self._caches: Dict[str, torch.Tensor] = {}
-        # KVCacheManager (or MambaHybridCacheManager) for managed resources
+        # KVCacheManager (or MambaHybridCacheManager) for managed resources.
```
eopXD: I don't see a need in editing the comment here.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -291,37 +300,109 @@ def _get_mamba_state_params(
...
-        the reference (via __eq__) are collected for managed allocation.
+    ) -> Tuple[
+        ResourceHandlerDict,
+        Dict[int, int],
+        Dict[int, "DataType"],
+        Dict[str, int],
+    ]:
+        """Identify all managed KV resources and their per-window shape maps.
+
+        A single C++ ``KVCacheManager`` now hosts every KV pool (full-attention
```
eopXD: Please state the comment directly upon its behavior and not its evolutioning history.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -291,37 +300,109 @@ def _get_mamba_state_params(
...
+        the order in which layers register their handlers.
 
         Returns:
-            Tuple of (reference_handler, managed_resources_dict).
-            reference_handler is None if no KV paged resources exist.
+            Tuple of:
+                - kv_managed: ResourceHandlerDict — every KVPagedResourceHandler
+                  in registration order; passed to the C++ ctor as the layer
+                  list.
+                - size_per_head_per_window: Dict[int, int] — maps each effective
```
eopXD: Why do we need 2 separate variables (`size_per_head_per_window`, `dtype_per_window`) as return?
Having a window size mapping to a tuple of (head_dim, dtype) should do.

Also, I don't think we need `window_per_layer`. The construction of kv_cache_config.max_attention_window can be derived from `kv_managed`.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -469,31 +572,70 @@ def _prepare_kv_cache_config(
 
     def _build_kv_cache_kwargs(
         self,
-        kv_ref: Optional[KVPagedResourceHandler],
         kv_managed: ResourceHandlerDict,
         kv_cache_config: KvCacheConfig,
+        size_per_head_per_window: Optional[Dict[int, int]] = None,
```
eopXD: Why can't we just use `kv_ref` as the source of construction? Do we really need to create two extra parameters passed?

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -602,18 +746,27 @@ def _create_and_assign_state_views(
...
             kv_managed: Dict of KV resources managed by the cache manager.
+            manager: The KVCacheManager that owns these resources.
 
         Returns:
             block_offset_multiplier derived from the first KV cache view's strides.
         """
         block_offset_multiplier = 0
         for idx, (name, h) in enumerate(kv_managed.items()):
-            view = self._kv_cache_manager.get_buffers(idx, kv_layout=h.kv_layout)
+            view = manager.get_buffers(idx, kv_layout=h.kv_layout)
```
eopXD: I don't see the meaning of this change. Why?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -331,6 +331,13 @@ def __init__(
         is_estimating_kv_cache: bool = False,
         execution_stream: Optional[torch.cuda.Stream] = None,
         linear_attention_metadata: Optional[LinearAttentionMetadata] = None,
+        # Optional per-window head_dim / dtype overrides forwarded to the C++ ctor.
```
eopXD: Can we encapsulate this with something like PoolConfiguration?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -1343,6 +1420,131 @@ def get_iteration_stats(self):
...
+        When ``self.dtype_per_window`` is non-empty, dtype is resolved per
+        layer (or per the supplied ``window_size`` for the homogeneous fast
+        path).  Otherwise ``dtype_default`` (or ``self.dtype``) is used.
+
+        Handles NVFP4 scaling-factor overhead, computed per dtype.
+        """
+        if dtype_default is None:
+            dtype_default = self.dtype
+
+        # Homogeneous fast path: a single dtype applies to every layer in
```
eopXD: Why do we need two separate paths? The homogeneous case should be covered by a correctly generalized heterogeneous calculation. Please re-approach this.

cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h
```diff
@@ -511,6 +510,12 @@ class GenerationRequest
         for (auto const [windowSize, metadata] : windowSizeToMetadata)
         {
             mCacheBlockIds[windowSize] = std::vector<std::vector<KVCacheBlock::IdType>>(beamWidth);
+            // Front-eviction count is tracked per-window: SWA pools may evict blocks while
+            // full-attention pools never do, and each pool's index math needs its own count.
+            // Pre-existing behavior (a single shared counter) only worked when one
```
eopXD: Don't need to document the pre-existing behaviors. Please adjust.

cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h
```diff
@@ -546,9 +551,21 @@ class GenerationRequest
         return mNumTokens;
     }
 
-    [[nodiscard]] SizeType32 getNumFrontBlocksRemoved() const
+    //! \brief Front-evicted blocks for the specified window only.
+    //! \details Only SWA windows ever evict; full-attention windows always return 0.
+    //!          Callers iterating per-pool MUST pass the relevant windowSize so that
+    //!          SWA evictions are not attributed to non-SWA pools.
+    //!          The no-argument / aggregate-across-windows overload was deliberately
+    //!          removed: implicit aggregation hid window-qualification bugs at call sites.
```
eopXD: Remove this historical documenting style comment.

cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h
```diff
@@ -1988,13 +2087,20 @@ class BaseKVCacheManager
     /// @param extraCostMemory Additional memory cost to account for CacheTransBufferManager::preAllocBufferSize
     /// @param kvFactor Factor for KV cache size calculation (typically 2 for key+value)
     /// @param maxBatchSize Maximum batch size
+    /// @param linearAttentionMetadata Optional metadata for linear / recurrent state pools.
+    /// @param sizePerHeadPerWindow Optional per-window override of @p sizePerHead.  Lets a single
```
eopXD: Adding this in parallel to the original `dtype` `sizePerHead` parameter causes confusion.

Let us encapsulate a `PoolConfiguration` concept to allow us to gather information like:
- What window sizes we are dealing with in this KV cache manager.
- What head_dim, data type does the pool in each window size is.

This encapsulation allows us to further support mixed-head_dim under the same window size (that is, multi-pool under a window_size) easier.

With a unified encapsulation, we can avoid exceptions we need to guard, for instance, to deal with both parameter co-existing and forcing one to over-power another.

Make sure to audit kvCacheManagerTest.cpp when you are approaching to resolve this comment. All existing callers of this class should reflect upon this change.

cpp/tensorrt_llm/batch_manager/kvCacheManager.cpp
```diff
@@ -4122,12 +4166,23 @@ BlocksPerWindow BaseKVCacheManager::calculateMaxNumBlocks(executor::KvCacheConfi
...
+        if (LinearAttentionMetadata::hasLinearCache(windowSize))
+        {
+            cacheSizeBytesPerTokenPerWindow[windowSize] = 1;
+            continue;
+        }
+        // Per-window head_dim and dtype: if the caller supplied an override map,
+        // use it; otherwise fall back to the scalars. This budgets pools whose
+        // (head_dim, dtype) differ across windows (Gemma4-style mixed attention)
+        // through the same single-manager path the rest of the C++ machinery uses.
+        auto const sizePerHeadIt = sizePerHeadPerWindow.find(windowSize);
```
eopXD: Why does this logic needs to exist here? It should be correctly done under `calculateCacheSizePerTokenForSingleWindowSize`.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -717,11 +921,11 @@ def _prepare_inputs(
             cu_seqlen=cu_seqlen,
             input_pos=input_pos,
             batch_info=batch_info,
-            cache_loc=cache_loc,
-            cu_num_pages=cu_num_pages,
-            extra_page_per_seq=extra_page_per_seq,
             slot_idx=state_slot_idx,
             prompt_lens=prompt_lens,
+            cache_loc_per_pool=cache_loc_per_pool,
```
eopXD: Just extend the variable in the original position.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -291,37 +296,86 @@ def _get_mamba_state_params(
...
+        C++ ``mLayerToWindowSize`` routing.
 
         Returns:
-            Tuple of (reference_handler, managed_resources_dict).
-            reference_handler is None if no KV paged resources exist.
+            Tuple of:
+                - kv_managed: ResourceHandlerDict — every KVPagedResourceHandler
+                  in registration order; passed to the C++ ctor as the layer
+                  list.
+                - shape_per_window: Dict[int, Tuple[int, DataType]] — maps each
```
eopXD: I think the correct semantic here should be "Pool information". This map has an implicit assumption that a window size mapping to a (head_dim, datatype) pair. We should allow a window_size to have multiple pools in the future. The design here should be flexible enough to accommodate that. Please adjust.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -469,31 +540,72 @@ def _prepare_kv_cache_config(
...
+                dtype_per_window={w: dt for w, (_, dt) in shape_per_window.items()},
+            )
             kv_cache_kwargs.update(
                 {
                     "kv_cache_type": kv_cache_type,
                     "num_layers": len(kv_managed),
                     "num_kv_heads": [h.num_kv_heads for h in kv_managed.values()],
-                    "head_dim": kv_ref.head_dim,
-                    "dtype": torch_dtype_to_binding(kv_ref.dtype),
+                    "head_dim": default_head_dim,
```
eopXD, line 610: This goes back to my comment above under kvCacheManager.h.

We should avoid having an explicit, un-encapsulated information that blocks our flexibility in future pool capability expansion.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -73,6 +74,17 @@
     int]]  # window_size -> (blocks_in_primary_pool, blocks_in_secondary_pool)
 
 
+@dataclass
+class PoolConfiguration:
+    """Per-window overrides for mixed-shape KV pools.
```
eopXD: A single pool configuration should describe
- The attention window size
- The head dimension of the attention
- The datatype for the attention

A collection of configuration collects the above and feeds it to the KV cache manager for resource allocation.

cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h
```diff
@@ -1977,9 +2081,9 @@ class BaseKVCacheManager
     ///          of memory requirements. The weighting considers both the window size and the number of
     ///          layers using each window size, as well as the sum of cache sizes per token for each window.
     /// @param config KV cache configuration parameters
-    /// @param dtype Data type used for KV cache values
+    /// @param dtype Default KV-cache data type (used for windows without a matching pool entry)
```
eopXD, line 2139: Do we still need the `dtype` parameter or `sizePerHead` parameter?
I think this will become a legacy interface.

The newly added `poolConfigurations` parameter should cover the case of only needing a single pool. Let us remove this usage so such inflexible design does not last in the future.

You will need to carefully audit all existing usage to make sure that the caller updates its usage and removes the parameter as well.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
eopXD: Backlogging this as a follow-up as this does not directly contribute towards the goal of this MR.

cpp/include/tensorrt_llm/batch_manager/kvCacheManager.h
eopXD: Backlog as this is not aligned with MR's goal. Will follow-up.

eopXD:
A footnote for the current change.

`mNumFrontBlocksRemovedPerWindow` and `getDataTypePerWindow` has the same contradiction to the "pool" concept a resource manager should essentially have. This is a larger scope refactor and should be done after this pull request. For the sake of this pull request, now we will have it as-is.

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

SimengLiu-nv (approved):
LGTM

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot run --disable-fail-fast --extra-stage "DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, H100_PCIe-AutoDeploy-1"

eopXD:
/bot skip --comment "The only failing pipeline `H100_PCIe-PyTorch-Perf-1` is not related to the scope of change of this  merge request. The pipeline has passed before too in the CI attempts of this merge request. To avoid from running the long pipeline again, we have AutoDeploy-related pipeline passing now, which proves that the MR should be good now. Given the above, I have enough confidence to use skip here and merge the MR."

nvchenghaoz (approved):

PR #13758, opened 2026-05-05
[https://nvbugs/5805494][fix] Limit maximum warmup token count to prevent crash in autotuner

dbari:
## Summary by CodeRabbit

* **Bug Fixes**
  * Implemented a maximum token count cap (8,192 tokens) during model warmup initialization, limiting tensor dimensions and optimizing resource consumption.
  * Warmup request-shape selection and autotuner warmup processes now respect this token limit to prevent excessive memory usage.

* **Tests**
  * Added regression test validating that model warmup enforces the token count cap across different configurations.

## Description

Caps the per-rank token count used by every warmup forward pass at a new `PyTorchModelEngine.max_warmup_tokens = 8192`, matching the `MAX_PROFILE_BUCKET` already used by every tunable op in `trtllm_gen_custom_ops.py`. The cap is applied both in `_get_max_shape_warmup_requests` (used by the pre- and post-CUDA-graph general warmups) and in `_run_autotuner_warmup`.

Discovered with DeepSeek-R1, `tp=8`, `ep=8`, `enable_attention_dp=True`, `max_num_tokens=16384`. The autotuner warmup ran a forward pass at the full 16384 per-rank context shape. After the MoE all-gather the TRT-LLM-Gen FP8 block-scale activation kernel saw `16384 * dp_size = 131072` tokens, and `permutedIdx * innerDim` overflowed int32 inside `DevKernel.cu:412`, producing an illegal memory access during init:

`RuntimeError: [TensorRT-LLM][ERROR] CUDA runtime error in cudaLaunchKernelEx(...): an illegal memory access was encountered (.../trtllmGenKernels/blockScaleMoe/DevKernel.cu:412)`

The autotuner only validates tactics on profile buckets up to 8192 (see `MAX_PROFILE_BUCKET` and `round_rule` in `tensorrt_llm/_torch/custom_ops/trtllm_gen_custom_ops.py`), so warmup shapes above the bucket size yield no tuning benefit while pushing the kernel into territory it was never validated on. Capping the warmup shape at the same value keeps tuning, warmup and the kernel's validated range consistent.

This was previously caught only at runtime by the scheduler distributing context tokens across ranks unevenly, so production inference does not trigger it; the autotuner warmup, where every rank runs at `max_num_tokens` simultaneously, is the only path that reliably hits the overflow on supported configurations.

## Test Coverage

Adds `test_warmup_caps_token_count` in `tests/unittest/_torch/executor/test_pytorch_model_engine.py` (picked up automatically via `unittest/_torch/executor` in the existing `l0_h100.yml` / `l0_b200.yml` / `l0_b300.yml` lists).

The test:
- Builds a `DummyModelEngine` with `max_num_tokens=16384` and a KV cache sized to `max_seq_len * max_batch_size = 32768` tokens, with sanity assertions that the cap (rather than KV cache or batch sizing) is the binding bound.
- Wraps `_create_warmup_request` with a recording wrapper that delegates to the original, then runs the full `model_engine.warmup()` end-to-end, exercising both general-warmup callsites and `_run_autotuner_warmup` with the real `autotune()` context on a single GPU.
- Asserts that no warmup call requested more tokens than `max_warmup_tokens`. Fails with the recorded list and the resolved `model_engine.py` path when the cap is removed, passes with the cap.

Manual verification on the 8-GPU DeepSeek-R1 deployment that originally crashed: `trtllm-bench --max_num_tokens 16384` now completes the autotuner warmup cleanly and proceeds to throughput measurement.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -504,6 +504,13 @@ class PyTorchModelEngine(ModelEngine):
 
         self.kv_cache_dtype_byte_size = self.get_kv_cache_dtype_byte_size()
 
+        # Upper bound on the per-rank token count used for warmup forward passes
+        # (both the max-shape general warmup and the autotuner warmup). Each
+        # tunable op currently generates tuning buckets only up to 8192 (see
+        # MAX_PROFILE_BUCKET in trtllm_gen_custom_ops.py and related files), so
+        # warmup shapes larger than this do not improve tuning quality.
+        self.max_warmup_tokens = 8192
+
     def register_forward_pass_callable(self, callable: Callable):
         self.forward_pass_callable = callable
 
@@ -684,7 +691,8 @@ class PyTorchModelEngine(ModelEngine):
         kv_cache_manager = resource_manager.get_resource_manager(
             self.kv_cache_manager_key)
         token_num_upper_bound = min(self.max_num_tokens,
-                                    self.batch_size * (self.max_seq_len - 1))
+                                    self.batch_size * (self.max_seq_len - 1),
+                                    self.max_warmup_tokens)
         curr_max_num_tokens = kv_cache_manager.get_num_available_tokens(
             token_num_upper_bound=token_num_upper_bound,
             max_num_draft_tokens=self.original_max_draft_len)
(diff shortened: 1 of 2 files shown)


DomBrown (approved):
LGTM

PR #13768, opened 2026-05-05
[Bugfix] Forward secondary_offload_min_priority to KVCacheManager in PyTorch executor

Saddss:
## Summary

The PyTorch backend's `KVCacheManager` (in `tensorrt_llm/_torch/pyexecutor/resource_manager.py`) builds its kwargs from `KvCacheConfig` but omits `secondary_offload_min_priority`. As a result, the user-supplied value (set via the YAML passed to `trtllm-serve --config`, the OpenAI server config, the Python `LLM` API, etc.) is silently dropped and the C++ default (30) is always used for the PyTorch backend.

The C++ `KVCacheManager` constructor and its nanobind binding already accept this parameter:

```cpp
nb::arg("cache_type") = tbk::CacheType::kSELF,
nb::arg("secondary_offload_min_priority") = std::nullopt,
nb::arg("event_manager") = nullptr,
```

and the executor-level Python `KvCacheConfig.__init__` already forwards it into the executor proto. Only the PyTorch-executor-side kwargs entry is missing.

## Fix

Add the field next to the other `kv_cache_config`-derived kwargs, in the same position as the C++ binding's parameter ordering (between `cache_type` and `enable_partial_reuse`).

## Test plan

- [x] Set `secondary_offload_min_priority: 0` in a YAML config; before this PR the canOffload check (`pri >= getPriorityIdx(secondary_offload_min_priority)`) keeps using the C++ default 30 — priority=0 blocks bypass D2H. After the PR they are eligible for D2H to secondary memory as the user requested.
- [x] Existing CI continues to pass (no behavior change for users that don't set the field).

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Enabled configuration of secondary offload minimum priority settings for KV cache management, allowing finer control over cache offloading behavior.


Saddss:
Hi @NVIDIA/trt-llm-kv-cache-manager-devs, @chang-l, @eopXD, and @yizhang-nv, could you please take a quick look at this PR when you have a chance?

This is a very small PyTorch-backend fix in `resource_manager.py`: `secondary_offload_min_priority` is already plumbed through `KvCacheConfig` and accepted by `KVCacheManagerCpp`, but the PyTorch executor never forwards it, so user-configured values are ignored and the backend always falls back to the default `30`.

The PR only adds the missing kwargs entry. Would appreciate a review. Thanks.

nvpohanh (approved):

nvpohanh:
@Saddss should we add a unit test to cover this?

Saddss:
> @Saddss should we add a unit test to cover this?

thanks，done

Saddss:
Could you please re-open a CI for me?

> /bot run

Saddss:
> /bot run --disable-fail-fast

@karljang Thanks for the follow-up. It seems we need multiple CI attempts to get it to pass?

PR #13773, opened 2026-05-05
[None][feat] FlashInfer NVFP4 MoE backend (SM120/SM121) for Nemotron …

farazkh80:
## Description

Adds a new MoE backend, `FlashInferNvfp4Sm12xFusedMoE`, for Nemotron-Super-120B-NVFP4 on SM120 (RTX 5090 / RTX PRO 6000 / GB202) and SM121 (DGX Spark / GB10). It is **auto-selected on the `CUTLASS` path** when `quant_algo == NVFP4`, SM is 120/121, and `flashinfer` is importable — otherwise falls back to plain `CutlassFusedMoE`. There is no new user-facing `moe_backend` value; the heuristic in `get_moe_cls()` mirrors the existing `MEGAMOE_DEEPGEMM` `can_implement`-gated pattern.

Hybrid composition:
- **Prefill (`x.shape[0] >= 64`)** routes through the inherited `CutlassFusedMoE` NVFP4 GroupGEMM. b12x's 12-CTA-per-token MMA pattern is suboptimal at large `m`.
- **Decode (`x.shape[0] <  64`)** dispatches to FlashInfer's `B12xMoEWrapper.run`. Beats CUTLASS by ~17.6 % TPOT.

## Hardware constraints (rejected by `can_implement` or `__init__`)

NVFP4 only; SM120/121 only; bf16/fp16 activation only; `ep_size == 1` only; no MoE alltoall; activations limited to `Relu2` and `Swiglu`; no `swiglu_gptoss_style`. `Fp4QuantizedTensor` input rejected on the decode path (b12x quantizes activations internally).

## Performance

Single RTX PRO 6000 Blackwell (SM120, 97 GB), Nemotron-Super-120B-NVFP4, TRT-LLM 1.3.0rc14, FlashInfer 0.6.8, cutlass-dsl 4.5.0; ISL=2048, OSL=1024, 5 reqs, conc=1, KV reuse off, `cuda_graph_config: {batch_sizes: [1]}`, `max_num_tokens=2048`.

| Variant | Total tput (tok/s) | TTFT P50 (ms) | TPOT P50 (ms) |
| --- | ---: | ---: | ---: |
| `moe_backend: CUTLASS` (pre-PR) | 70.58 | 154.67 | 13.97 |
| FlashInfer-only (pure b12x)     | 85.32 | 229.26 | 11.50 |
| **`moe_backend: CUTLASS` (this PR, auto-promoted)** | **85.92** | **154.53** | **11.49** |

Matches CUTLASS on TTFT (prefill on CUTLASS), matches pure b12x on TPOT (decode on b12x), beats both on total throughput. Tokens/Watt +20.4 % vs CUTLASS.

## GSM8K accuracy parity (1319 samples, 8-shot CoT, greedy)

| Backend | Accuracy | Δ vs CUTLASS |
| --- | ---: | ---: |
| `CUTLASS` (baseline) | 92.418 % | — |
| `CUTLASS` (auto-promoted, this PR) | 92.267 % | −0.151 pp |

Statistically indistinguishable (≈ 2 questions out of 1319; well within 95 % binomial CI of ±1.5 pp).

## PR Checklist

  - [x] PR follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md).
  - [x] Test cases provided.
  - [x] No new dependencies added.
## GitHub Bot Help

## Summary by CodeRabbit

* **New Features**
  * Added FlashInfer as a selectable MoE backend option in model configuration for optimized Mixture of Experts inference on SM120/SM121 GPUs with NVFP4 quantization
  
* **Documentation**
  * Updated developer guide with FlashInfer MoE backend capabilities matrix and supported constraints documentation

### DIFF
--- a/tensorrt_llm/_torch/modules/fused_moe/MOE_DEVELOPER_GUIDE.md
+++ b/tensorrt_llm/_torch/modules/fused_moe/MOE_DEVELOPER_GUIDE.md
@@ -147,7 +147,7 @@ Still on old path (standalone, with embedded communication):
 | `fused_moe_deepgemm.py` | `DeepGemmFusedMoE` | SM100/SM103 | FP8 Block Scales on Blackwell | `EXTERNAL_COMM` |
 | `fused_moe_densegemm.py` | `DenseGEMMFusedMoE` | SM100/SM103 | NVFP4 min-latency; CuTe DSL dense GEMM packs all experts into one matrix (vs Cutlass per-expert scatter), efficient for small token counts | `EXTERNAL_COMM` |
 | `fused_moe_cute_dsl.py` | `CuteDslFusedMoE` | SM100/SM103 | High throughput NVFP4, generally faster than Cutlass | `EXTERNAL_COMM` |
-| `fused_moe_cute_dsl_b12x.py` | `CuteDslB12xFusedMoE` | SM120/SM121 | NVFP4; auto-selected on the `CUTLASS` path | `EXTERNAL_COMM` |
+| `fused_moe_cute_dsl_b12x.py` | `CuteDslB12xFusedMoE` | SM120/SM121 | NVFP4 hybrid CUTLASS-prefill / FlashInfer NVFP4 MoE decode — best perf on RTX PRO 6000 (SM120) and DGX Spark (SM121); select via the `CUTEDSL` backend path (auto-promoted when flashinfer is importable) | `EXTERNAL_COMM` |
 | `mega_moe/mega_moe_deepgemm.py` | `MegaMoEDeepGemm` | SM100/SM103 | W4A8_MXFP4_MXFP8 via DeepGEMM `fp8_fp4_mega_moe` fused dispatch+GEMM+act+GEMM+combine kernel; requires `hidden_size % 512 == 0` | `FUSED_COMM` |
 | `fused_moe_triton.py` | `TritonFusedMoE` | SM90 only | GPT-OSS on Hopper (requires `swiglu_gptoss_style=True`) | (legacy path) |
 | `fused_moe_wide_ep.py` | `WideEPMoE` | All GPUs | Deprecating — use ConfigurableMoE instead | (legacy path) |
@@ -193,7 +193,7 @@ Each backend's `can_implement(quant_algo, dtype_activation, swiglu_gptoss_style,
 | Unquantized (BF16/FP16) | Y (SM80+) | N | N | N | N | N | Y (SM90, BF16) | Y | Y |
 | FP8 QDQ | Y (SM89+) | N | N | N | N | N | Y (SM90) | Y | Y |
 | FP8 Block Scales | Y (SM90, SM120) | Y (SM100/103) | Y (SM100/103) | N | Y (SM100/103) | N | N | Y | Y |
-| NVFP4 | Y (SM100/103/120/121) | Y (SM100/103) | N | Y (SM100/103) | Y (SM100/103) | N | N | Y | Y |
+| NVFP4 | Y (SM100/103/120/121) | Y (SM100/103) | N | Y (SM100/103) | Y (SM100/103/120/121) | N | N | Y | Y |
 | W4A8 NVFP4 FP8 | N | Y (SM100/103) | N | N | N | N | N | N | N |
 | W4A16 MXFP4 | Y (SM90) | Y (SM100/103) | N | N | N | N | Y (SM90) | N | N |
 | W4A8 MXFP4 FP8 | Y (SM100/103) | Y (SM100/103) | N | N | N | N | Y (SM90) | N | N |
--- a/tensorrt_llm/_torch/modules/fused_moe/MOE_DEVELOPER_GUIDE.md
+++ b/tensorrt_llm/_torch/modules/fused_moe/MOE_DEVELOPER_GUIDE.md
@@ -147,6 +147,7 @@ Still on old path (standalone, with embedded communication):
 | `fused_moe_deepgemm.py` | `DeepGemmFusedMoE` | SM100/SM103 | FP8 Block Scales on Blackwell | `EXTERNAL_COMM` |
 | `fused_moe_densegemm.py` | `DenseGEMMFusedMoE` | SM100/SM103 | NVFP4 min-latency; CuTe DSL dense GEMM packs all experts into one matrix (vs Cutlass per-expert scatter), efficient for small token counts | `EXTERNAL_COMM` |
 | `fused_moe_cute_dsl.py` | `CuteDslFusedMoE` | SM100/SM103 | High throughput NVFP4, generally faster than Cutlass | `EXTERNAL_COMM` |
+| `fused_moe_cute_dsl_b12x.py` | `CuteDslB12xFusedMoE` | SM120/SM121 | NVFP4; auto-selected on the `CUTLASS` path | `EXTERNAL_COMM` |
 | `mega_moe/mega_moe_deepgemm.py` | `MegaMoEDeepGemm` | SM100/SM103 | W4A8_MXFP4_MXFP8 via DeepGEMM `fp8_fp4_mega_moe` fused dispatch+GEMM+act+GEMM+combine kernel; requires `hidden_size % 512 == 0` | `FUSED_COMM` |
 | `fused_moe_triton.py` | `TritonFusedMoE` | SM90 only | GPT-OSS on Hopper (requires `swiglu_gptoss_style=True`) | (legacy path) |
 | `fused_moe_wide_ep.py` | `WideEPMoE` | All GPUs | Deprecating — use ConfigurableMoE instead | (legacy path) |
--- a/tensorrt_llm/_torch/modules/fused_moe/create_moe.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/create_moe.py
@@ -298,9 +298,9 @@ def create_moe_backend(
             without_comm=without_comm,
             activation_type=activation_type,
         )
-    elif issubclass(moe_cls, CutlassFusedMoE):
-        # CutlassFusedMoE and any of its subclasses (e.g. FlashInferNvfp4Sm12xFusedMoE)
-        # share the same constructor signature.
+    elif moe_cls in (CutlassFusedMoE, FlashInferNvfp4Sm12xFusedMoE):
+        # CuteDslFusedMoE and DeepGemmFusedMoE also subclass CutlassFusedMoE but
+        # have narrower constructors, so they take their own branches below.
         return moe_cls(
             routing_method=routing_method,
             num_experts=num_experts,
--- a/tensorrt_llm/_torch/modules/fused_moe/create_moe.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/create_moe.py
@@ -34,6 +34,27 @@ def get_moe_cls(
     if override_quant_config is not None:
         quant_config = override_quant_config
     if moe_backend.upper() == "CUTLASS":
+        # Auto-promote to FlashInferNvfp4Sm12xFusedMoE (hybrid CUTLASS-prefill
+        # / b12x-decode) on SM120 / SM121 + NVFP4 when flashinfer is available.
+        # Falls back to plain CutlassFusedMoE otherwise.
+        if quant_config is not None and quant_config.quant_mode.has_nvfp4():
+            from tensorrt_llm._utils import get_sm_version
+            sm_version = get_sm_version()
+            if sm_version in FlashInferNvfp4Sm12xFusedMoE._SUPPORTED_SM_VERSIONS:
+                try:
+                    import flashinfer  # noqa: F401
+                    logger.info(
+                        "Auto-selecting FlashInferNvfp4Sm12xFusedMoE for hybrid "
+                        "CUTLASS-prefill / b12x-decode (SM%d + NVFP4).",
+                        sm_version,
+                    )
+                    return FlashInferNvfp4Sm12xFusedMoE
+                except ImportError:
+                    logger.warning(
+                        "FlashInferNvfp4Sm12xFusedMoE eligible (SM%d + NVFP4) "
+                        "but flashinfer is not importable; using CutlassFusedMoE.",
+                        sm_version,
+                    )
         return CutlassFusedMoE
     elif moe_backend.upper() == "VANILLA":
         return VanillaMoE
@@ -139,23 +160,6 @@ def get_moe_cls(
                 "Falling back to CutlassFusedMoE.")
             return CutlassFusedMoE
         return MegaMoEDeepGemm
-    elif moe_backend.upper() == "FLASHINFER_NVFP4SM12X":
-        # FlashInferNvfp4Sm12xFusedMoE is the hybrid CUTLASS-prefill /
-        # b12x-decode NVFP4 MoE backend for SM120/SM121. Hard-error rather
-        # than silently falling back to CUTLASS so a misconfigured request
-        # is loud at startup, not a silent perf regression.
-        if quant_config is None or not quant_config.quant_mode.has_nvfp4():
-            raise ValueError(
-                "FlashInferNvfp4Sm12xFusedMoE requires NVFP4 quantization "
-                f"(got quant_config={quant_config}).")
-        from tensorrt_llm._utils import get_sm_version
-        sm_version = get_sm_version()
-        if sm_version not in FlashInferNvfp4Sm12xFusedMoE._SUPPORTED_SM_VERSIONS:
-            sm_list = "/".join(f"SM{v}" for v in sorted(
-                FlashInferNvfp4Sm12xFusedMoE._SUPPORTED_SM_VERSIONS))
-            raise ValueError(f"FlashInferNvfp4Sm12xFusedMoE requires {sm_list} "
-                             f"(got SM{sm_version}).")
-        return FlashInferNvfp4Sm12xFusedMoE
     else:
         raise ValueError(f"Unsupported moe backend: {moe_backend}")
 
--- a/tensorrt_llm/_torch/modules/fused_moe/create_moe.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/create_moe.py
@@ -34,27 +34,6 @@ def get_moe_cls(
     if override_quant_config is not None:
         quant_config = override_quant_config
     if moe_backend.upper() == "CUTLASS":
-        # Auto-promote to CuteDslB12xFusedMoE (hybrid CUTLASS-prefill
-        # / b12x-decode) on SM120 / SM121 + NVFP4 when flashinfer is available.
-        # Falls back to plain CutlassFusedMoE otherwise.
-        if quant_config is not None and quant_config.quant_mode.has_nvfp4():
-            from tensorrt_llm._utils import get_sm_version
-            sm_version = get_sm_version()
-            if sm_version in CuteDslB12xFusedMoE._SUPPORTED_SM_VERSIONS:
-                try:
-                    import flashinfer  # noqa: F401
-                    logger.info(
-                        "Auto-selecting CuteDslB12xFusedMoE for hybrid "
-                        "CUTLASS-prefill / b12x-decode (SM%d + NVFP4).",
-                        sm_version,
-                    )
-                    return CuteDslB12xFusedMoE
-                except ImportError:
-                    logger.warning(
-                        "CuteDslB12xFusedMoE eligible (SM%d + NVFP4) "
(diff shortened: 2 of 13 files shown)


tbraun96:
Disclosure: I work on [Atlas](https://github.com/Avarok-Cybersecurity/atlas). We've been running NVFP4 MoE on `sm_120` / `sm_121` for a while and hit the FlashInfer side of this gap pretty hard. Sharing the patch in case it's useful prior art for the SM120/SM121 backend wiring here.

Two things that bit us on FlashInfer + NVFP4 on consumer Blackwell that may or may not already be on your radar:

**1. E2M1 conversion PTX is Hopper/SM10-only.** FlashInfer's CUTLASS headers gate `CUDA_PTX_FP4FP6_CVT_ENABLED` for SM ranges that exclude `sm_120`/`sm_121`. The kernel still emits `.tile::scatter4` PTX on those archs and faceplants at JIT time. We patch around this with a software conversion using `__float_as_uint` and bit-twiddling. Patch script in the public repo:

[`docker/gb10/fix_flashinfer_e2m1_sm121.py`](https://github.com/Avarok-Cybersecurity/atlas/blob/main/docker/gb10/fix_flashinfer_e2m1_sm121.py)

About 30 lines. Disables the hardware E2M1 path for SM121 specifically and falls back to the software conversion. Roughly 32x speedup over the broken-PTX path on a 35B baseline (1.1 to 35 tok/s for us) so worth catching on the SM120 side too.

**2. NVFP4 MoE backend dispatch bug.** Upstream FlashInfer's `select_nvfp4_moe_backend()` returned `None` for `k_cls` on `sm_121`, which made vLLM fall through to a broken path silently. We patch that one separately:

[`docker/gb10/fix_flashinfer_nvfp4_moe_backend.py`](https://github.com/Avarok-Cybersecurity/atlas/blob/main/docker/gb10/fix_flashinfer_nvfp4_moe_backend.py)

If FlashInfer's gating already handles this on SM120 cleanly in your branch, ignore. If not, this is the call site that was lying to callers.

For end-to-end NVFP4 numbers on `sm_121` with these patches in (Qwen3.6-35B-A3B-NVFP4, MTP K=2): 214.6 tok/s decode at c=1, sparkrun-benchmark-validated bundle attached on this PR:

[`Avarok-Cybersecurity/atlas-recipes#2`](https://github.com/Avarok-Cybersecurity/atlas-recipes/pull/2)

Glad to see Nemotron NVFP4 landing for SM120/121. Happy to dig into either of the patches above if you want to confirm whether your backend wiring handles them differently.

tbraun96:
Disclosure: I work on [Atlas](https://github.com/Avarok-Cybersecurity/atlas). We've been running NVFP4 MoE on `sm_120` / `sm_121` for a while and hit the FlashInfer side of this gap pretty hard. Sharing the patch in case it's useful prior art for the SM120/SM121 backend wiring here.

Two things that bit us on FlashInfer + NVFP4 on consumer Blackwell that may or may not already be on your radar:

**1. E2M1 conversion PTX is Hopper/SM10-only.** FlashInfer's CUTLASS headers gate `CUDA_PTX_FP4FP6_CVT_ENABLED` for SM ranges that exclude `sm_120`/`sm_121`. The kernel still emits `.tile::scatter4` PTX on those archs and faceplants at JIT time. We patch around this with a software conversion using `__float_as_uint` and bit-twiddling. Patch script in the public repo:

[`docker/gb10/fix_flashinfer_e2m1_sm121.py`](https://github.com/Avarok-Cybersecurity/atlas/blob/main/docker/gb10/fix_flashinfer_e2m1_sm121.py)

About 30 lines. Disables the hardware E2M1 path for SM121 specifically and falls back to the software conversion. Roughly 32x speedup over the broken-PTX path on a 35B baseline (1.1 to 35 tok/s for us) so worth catching on the SM120 side too.

**2. NVFP4 MoE backend dispatch bug.** Upstream FlashInfer's `select_nvfp4_moe_backend()` returned `None` for `k_cls` on `sm_121`, which made vLLM fall through to a broken path silently. We patch that one separately:

[`docker/gb10/fix_flashinfer_nvfp4_moe_backend.py`](https://github.com/Avarok-Cybersecurity/atlas/blob/main/docker/gb10/fix_flashinfer_nvfp4_moe_backend.py)

If FlashInfer's gating already handles this on SM120 cleanly in your branch, ignore. If not, this is the call site that was lying to callers.

For end-to-end NVFP4 numbers on `sm_121` with these patches in (Qwen3.6-35B-A3B-NVFP4, MTP K=2): 214.6 tok/s decode at c=1, sparkrun-benchmark-validated bundle attached on this PR:

[`Avarok-Cybersecurity/atlas-recipes#2`](https://github.com/Avarok-Cybersecurity/atlas-recipes/pull/2)

Glad to see Nemotron NVFP4 landing for SM120/121. Happy to dig into either of the patches above if you want to confirm whether your backend wiring handles them differently.

lcoleman0422:
@farazkh80 when I try to run this it crashes my spark. Granted I had a larger config then you. I assume I am missing something. 

@tbraun96 would you mind sharing a branch?

xxi-nv:
@QiJune @kaiyux Please help to review the skills added in this PR. And may I confirm whether the developers could add skills into this repo directly?

tensorrt_llm/_torch/modules/fused_moe/MOE_DEVELOPER_GUIDE.md
```diff
@@ -187,19 +188,47 @@ Communication strategies are auto-selected at runtime by `CommunicationFactory`
...
+| INT4 WoQ (W4AFP8) | N | N | N | N | N | N | N | N | Y | N |
+
+### FlashInferNvfp4Sm12xFusedMoE — composition, dispatch policy, and constraints
+
+`FlashInferNvfp4Sm12xFusedMoE` is selected via `--moe_backend FLASHINFER_NVFP4SM12X` (or `moe_config.backend: FLASHINFER_NVFP4SM12X` in YAML). It is gated on **NVFP4 quantization + SM120 (RTX 5090 / RTX PRO 6000 / GB202) or SM121 (DGX Spark / GB10)**, and depends at runtime on `flashinfer-python ≥ 0.6.8` (provides `flashinfer.B12xMoEWrapper` and `flashinfer.cute_dsl.utils.convert_sf_to_mma_layout`) plus the matched `nvidia-cutlass-dsl == 4.4.1` trio (`-libs-base`, `-libs-cu13`).
+
+Internally it dispatches per call by activation-row count `m`:
+
+- **`m >= _PREFILL_VIA_CUTLASS_THRESHOLD` (prefill)** routes through the inherited `CutlassFusedMoE.quantize_input` + `run_moe` (CUTLASS NVFP4 GroupGEMM). The b12x kernel's 12-CTA-per-token MMA pattern is suboptimal at large `m`; CUTLASS GroupGEMM wins per-layer.
+- **`m <  _PREFILL_VIA_CUTLASS_THRESHOLD` (decode)** dispatches to FlashInfer's `B12xMoEWrapper.run` — a kernel purpose-built for `m=1` / small routed-row counts. Beats CUTLASS by ~17.6 % TPOT on Nemotron-Super-120B-NVFP4.
```
xxi-nv: Please remove the unnecessary code details in this Doc.
This Doc serves as a relatively high-level guide. And expose too many details will make it hard to maintain the Doc and some kind of duplication with the code.

tests/unittest/_torch/modules/moe/test_cute_dsl_b12x_moe_backend.py
```diff
@@ -0,0 +1,170 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
xxi-nv, line 1: Please add the necessary test in test_moe_module.py and test_moe_backend.py to make sure the added backend could reuse the uniformed test framework.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_flashinfer_nvfp4_sm12x.py
```diff
@@ -0,0 +1,359 @@
...
+
+# ActivationType -> b12x activation string. b12x currently exposes "relu2"
+# (Nemotron-style x * relu(x)) and "silu" (SwiGLU-style x * silu(gate)).
+_ACTIVATION_MAP = {
+    ActivationType.Relu2: "relu2",
+    ActivationType.Swiglu: "silu",
+}
+
+
+class FlashInferNvfp4Sm12xFusedMoE(CutlassFusedMoE):
```
xxi-nv: Could you help to explain why we have to introduce a new backend? Do we have a another way to do this, something like heuristic selection? 
What's the relation between FlashInfer and CutlassFusedMoE?
May I know the motivation here to inherit from CutlassFusedMoE?

It is highly recommended to do a refactor with the skills https://gitlab-master.nvidia.com/ftp/trtllm-agent-toolkit/-/merge_requests/41#143127cc66c0ca5221561b39ccb27e7766f76cf3.

xxi-nv (changes requested):

tensorrt_llm/_torch/modules/fused_moe/fused_moe_flashinfer_nvfp4_sm12x.py
xxi-nv: A following question: do we have any chance to add more quantization in the FalshInfer path, if yes, then I would like to remove the quantization type and platform info from the `FlashInferNvfp4Sm12xFusedMoE`.

farazkh80: Thanks for the comment @xxi-nv, and for the moe-develop skills. I used them and made some design changes.

On FI precision support: as of flashinfer-python 0.6.8/0.6.10, B12xMoEWrapper is NVFP4-only and SM120/SM121-only. Because of that, B12x is no longer user-facing: moe_backend: `CUTLASS` auto-promotes to `FlashInferNvfp4Sm12xFusedMoE` on SM120/121 + NVFP4 + importable FlashInfer, otherwise it falls back to CutlassFusedMoE, following the existing `can_implement-gated` promotion pattern (similar to `MEGAMOE_DEEPGEMM-style` can_implement)

I also removed `FLASHINFER_NVFP4SM12X` from `MoeConfig.backend`, so there is no user-facing `FLASHINFER_*` config value anymore. The class name is internal only, so I’d prefer to defer renaming until we have a second FlashInfer-backed MoE path. `FlashInferNvfp4Sm12xFusedMoE` still subclasses CutlassFusedMoE because it reuses the NVFP4 weight-loading lifecycle and only overrides quantize_input, run_moe, and post_load_weights; we also keep `CutlassFusedMoE` for prefill since it performs better.

tests/unittest/_torch/modules/moe/test_cute_dsl_b12x_moe_backend.py
farazkh80: added, please let me know if they look good!

tensorrt_llm/_torch/modules/fused_moe/MOE_DEVELOPER_GUIDE.md
farazkh80: removed almost all of it since I changed the design a bit. Let me know if you prefer to remove the oneline file map as well.

requirements.txt
```diff
@@ -72,6 +72,7 @@ tiktoken
 blobfile
 openai-harmony==0.0.4
 nvidia-cutlass-dsl==4.5.0; python_version >= "3.10"
+nvidia-cutlass-dsl-libs-cu13==4.5.0; python_version >= "3.10"
```
farazkh80: CUDA 13 runtime libs for cutlass-dsl; not pulled automatically by the main wheel, but required on SM120/SM121 to avoid a ptxas "Unexpected Instruction types specified for '_mma'" ICE during b12x kernel JIT.

tensorrt_llm/_torch/modules/fused_moe/create_moe.py
```diff
@@ -33,6 +34,27 @@ def get_moe_cls(
...
+        # Auto-promote to FlashInferNvfp4Sm12xFusedMoE (hybrid CUTLASS-prefill
+        # / b12x-decode) on SM120 / SM121 + NVFP4 when flashinfer is available.
+        # Falls back to plain CutlassFusedMoE otherwise.
+        if quant_config is not None and quant_config.quant_mode.has_nvfp4():
+            from tensorrt_llm._utils import get_sm_version
+            sm_version = get_sm_version()
+            if sm_version in FlashInferNvfp4Sm12xFusedMoE._SUPPORTED_SM_VERSIONS:
+                try:
+                    import flashinfer  # noqa: F401
+                    logger.info(
```
xxi-nv: May I know whether the Prefill and Decode has different platforms?
If Prefill and Decode run both on SM120, then the heuristic here will not works, right?

xxi-nv: It looks that you are using `_route_to_cutlass `as the runtime switch.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_flashinfer_nvfp4_sm12x.py
```diff
@@ -0,0 +1,359 @@
...
+
+    def _route_to_cutlass(self, x) -> bool:
+        """Return ``True`` iff this call should fall back to the inherited
+        CUTLASS path (prefill chunk). ``Fp4QuantizedTensor`` inputs always
+        stay on the b12x path (which rejects them) so the existing error
+        message is preserved."""
+        return isinstance(x, torch.Tensor) and x.shape[0] >= self._PREFILL_VIA_CUTLASS_THRESHOLD
+
+    def post_load_weights(self):
+        """Build the b12x weight dict and instantiate ``B12xMoEWrapper``.
```
xxi-nv: It seems that the skills work partially.
It is suggested to apply the `post_load_weights` to a dedicated `quantization_method`.
The principle here is to leave all the implementation details into the `quantization_method`.
E.G.
https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/modules/fused_moe/quantization.py#L877

xxi-nv (approved):
Thanks for the refactoring. It looks much better. However, there are still some tasks that could make it more consistent with our design.

Approved and please address the comment.

tensorrt_llm/_torch/modules/fused_moe/MOE_DEVELOPER_GUIDE.md
xxi-nv: File map is needed so that the agent could lean where to find what it needed.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_flashinfer_nvfp4_sm12x.py
xxi-nv: Thanks for the refactoring.

May I confirm that the kernel implementation of FLASHINFER_NVFP4SM12X is also Cutlass C++?

farazkh80: it is actually in cuteDSL. but for prefill we use cutlass so I thought we could put it here.

Cutlass is the suggested MoE backend in sm120/121. so by adding this special handling, we essentially let the user get better output tp/s without changing their config.

xxi-nv: Wow, if we want to make it consistent with the current design of backend. I will recommend to integrate the newly added FLASHINFER_NVFP4SM12X to CuteDSLFusedMoE, because they both implement the kernel by cuteDSL.
And you can update the heuristic in 'AUTO' mode in MOE, but the AUTO mode in MOE is not so ready yet.
Please check https://github.com/xxi-nv/TensorRT-LLM/blob/bfcf0f31a52335622341c820270d0695bac1feaf/tensorrt_llm/_torch/model_config.py#L270-L293.

But I have to say, it is kind of complicated and not a good scalable way to add so many kernels in a easy way.

I need to take some time to rethink the definition of backends.  @litaotju Any suggestions?

tensorrt_llm/_torch/modules/fused_moe/create_moe.py
farazkh80: `get_moe_cls` is a static class-selection gate (SM + quant + flashinfer importable), and `_route_to_cutlass(x)` is the per-call runtime switch that routes m ≥ 64 to `CutlassFusedMoE.run_moe` and m < 64 to `B12xMoEWrapper.run`. 

Both fire on the same SM120/121 device. `get_moe_cls` itself doesn't differentiate prefill vs decode, that's `_route_to_cutlass`'s job.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_flashinfer_nvfp4_sm12x.py
farazkh80: extracted into `NVFP4CuteDslB12xFusedMoEMethod(NVFP4CutlassFusedMoEMethod)` in `quantization.py` next to `NVFP4CuteDslFusedMoEMethod`, so all b12x weight prep lives in the quant method. 

Backend's `post_load_weights` is now inherited from `CutlassFusedMoE` as a thin pass-through to `self.quant_method.post_load_weights(self)`.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_flashinfer_nvfp4_sm12x.py
farazkh80: Good suggestion, I restructured it as a child of CuteDslFusedMoE backend. FYI, we are going to add another w4a16 MoE from b12x through flashinfer in a new PR, so the CuteDSLFusedMoE backend should have more than 1 use case.

The w4a16 MoE is for nemo nano 3.5 model and the nvfp4 MoE is for nemo super 120b nvfp4.

yuanjingx87 (approved):

tensorrt_llm/_torch/modules/fused_moe/fused_moe_cute_dsl_b12x.py
```diff
@@ -0,0 +1,285 @@
...
+        token_final_scales: torch.Tensor,
+        x_sf: Optional[torch.Tensor] = None,
+        is_sf_swizzled: bool = True,
+        output_dtype: Optional[torch.dtype] = None,
+        tuner_num_tokens: Optional[int] = None,
+        tuner_top_k: Optional[int] = None,
+        moe_output: Optional[torch.Tensor] = None,
+        enable_alltoall: Optional[bool] = None,
+    ) -> torch.Tensor:
+        if self._route_to_cutlass(x):
```
xxi-nv, line 225: I am a little worried about this heuristic.
It means the MOE backend could be switched dynamically according to the shape of X.
E.G.
This means even in the GEN phase, the CHANGE of shape of X may select CutlassFusedMoE, and the next forward may choose CuteDslB12xFusedMoE.
But currently, the MOE doesn't support dynamic weights loading.

All the weights are loaded at the first time model loading and it will only load the weights according to your current design of CuteDslB12xFusedMoE. If the MOE backend is dynamically switched to CutlassFusedMoE, the weights layout may not satisfy the requirement.

I highly recommend letting the user choose the backend statically. 
If we can not do the heuristic selection of backends in https://github.com/xxi-nv/TensorRT-LLM/blob/bfcf0f31a52335622341c820270d0695bac1feaf/tensorrt_llm/_torch/model_config.py#L270-L293., then don't do it.


I know you want to save some time for the user and make it easy, but we are not ready yet. I am trying to introduce the AUTOTUING capability across MOE backends, which will choose different MOE backends according to the silicon time.

farazkh80: If I understand correctly, the concern is that `_route_to_cutlass` swaps backends mid-forward and weights wouldn't match. But both layouts are pre-staged once at load time on the same module.

CuteDslB12xFusedMoE.post_load_weights calls super().post_load_weights first (quantization.py:3008), which loads the full CUTLASS NVFP4 layout exactly as CutlassFusedMoE expects, and only then adds the b12x views/wrapper as new attributes, so both contracts are always satisfied on the same module, and  _route_to_cutlass only picks a compute path, never swaps a layout. 

Collapsing it would erase the only reason this backend exists (≈17.6 % decode TPOT win on Nemotron-Super-120B-NVFP4; CUTLASS GroupGEMM still wins at large m). This is something the cross-backend autotuner can absorb once it lands.

PR #13783, opened 2026-05-05
[None][feat] Wan 2.2 serve: extend video generation API

wu6u3tw:
## Description

Adds four serve-side capabilities to the Wan 2.2 video generation pipeline.
All are independent, backward-compatible additive extensions to existing
public-facing schemas (no internal API changes).

1. **VideoGenerationRequest knobs**: `guidance_scale_2`, `boundary_ratio`,
   `num_frames`. `guidance_scale_2`/`boundary_ratio` are routed through
   `extra_params` to match `WanPipeline.infer`'s `extra.get(...)` reads.
   Also defaults `guidance_scale_2 = 3.0` on Wan 2.2 A14B when unset
   (was `= guidance_scale`, which is wrong for two-stage CFG).

2. **`fixed_latent_path`**: `VisualGenArgs` option to load a pre-computed
   latent tensor and substitute it for sampled latents at request time —
   used for MLPerf-deterministic generation. Lazy-loaded on first
   inference (pipeline construction runs inside `MetaInitMode`, where
   `torch.load` on a meta tensor would raise). Mirrors the existing
   `extra_attrs` pattern used by `spatial_upsampler_path` /
   `distilled_lora_path`. Warmup passes `_use_fixed_latent=False`.

3. **`response_format = "video_path"`**: server returns JSON with the
   on-disk path instead of streaming bytes — for filesystem-shared
   deployments (e.g. Lustre). Avoids holding HTTP open through encode
   + transfer of multi-MB payloads. Default remains `"video_bytes"`.

4. **`output_format = "raw"`**: uncompressed uint8 RGB byte dump
   (T·H·W·3 layout, `.rgb.bin`). No ffmpeg dependency, no encode
   latency. Worst-case payload ~224 MB at 720x1280x81. Default remains
   `"auto"`.

## Test Coverage

- Existing `tests/unittest/_torch/visual_gen/test_wan22_t2v_pipeline.py`
  exercises the WAN pipeline path; new fields are additive and do not
  change existing test behavior.
- Manual verification: `trtllm-serve <Wan2.2-T2V-A14B-Diffusers>` then
  POST `VideoGenerationRequest` with each new field combination.

## PR Checklist

- Test cases provided for new code paths (additive backward-compatible fields; no new code paths require dedicated tests beyond manual server verification).
- Any new dependencies have been scanned for license and vulnerabilities (none added).
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes (no ownership change).
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR (n/a).

## GitHub Bot Help


zhenhuaw-me:
@wu6u3tw Can you clarify the motivation of this change? We are refactoring the Python APIs, once we have a "beta" python API, we plan to get back to the serving endpoint changes.

wu6u3tw:
@zhenhuaw-me 
I am adding support for adding param: gaidence_scale_2 and fixed_latened pt into trtllm-serve which is a supported param in non-serve run

wu6u3tw:
@zhenhuaw-me Let's sync offline for the api change plan.

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
```diff
@@ -97,6 +97,16 @@ def __init__(self, model_config):
                 "Use cache_backend='none' or 'cache_dit' (not 'teacache')."
             )
 
+        # MLPerf-deterministic fixed latent (lazy-loaded on first real inference).
+        # __init__ runs under MetaInitMode where torch.load triggers aten.set_ on
+        # a meta tensor and raises MetaInitException, so we only stash the path.
+        self.fixed_latent_path: Optional[str] = (
```
zhenhuaw-me: If this is a debug purpose or benchmarking purpose knob? If it's a debug knob, we prefer to use a env such as `TLLM_MLPERF_FIXED_LATENT_PATH`.

Tin-Yin: Different latent for given inputs/seeds across frameworks.

If it's expected? Or it's should be something like acceptable variants? @o-stoner  for viz.

Source of the latent: PT.

tensorrt_llm/media/encoding.py
```diff
@@ -559,6 +561,22 @@ def save_video(
 
     if format in ("mp4", "avi"):
         return _save_encoded_video(video, audio, output_path, frame_rate, audio_sample_rate)
+    if format == "raw":
```
zhenhuaw-me: `pt`

tensorrt_llm/serve/openai_protocol.py
```diff
@@ -1420,21 +1420,55 @@ class VideoGenerationRequest(OpenAIBaseModel):
...
+        ge=0.0,
+        le=1.0,
+        description=(
+            "Boundary ratio in [0, 1] that switches CFG from "
+            "guidance_scale to guidance_scale_2 once the timestep crosses "
+            "boundary_ratio * num_train_timesteps. Required when "
+            "guidance_scale_2 should take effect."))
     guidance_rescale: Optional[float] = Field(
         default=None, description="Classifier-free guidance rescale.")
+    num_frames: Optional[int] = Field(
```
zhenhuaw-me, line 1469: OpenAI API seems to use `seconds * fps`. cc @JunyiXu-nv

JunyiXu-nv: `fps` is actually also not a standard OpenAI API. I think it's ok to have an extra field directly indicate the number of frames, since it'll not break the normal OAI sdk usage.

wu6u3tw:
Fixed based on the  reviews.

tensorrt_llm/serve/openai_protocol.py
wu6u3tw: So, if we really need 81 frames strictly, what is your suggestions?

zhenhuaw-me: I think we can specify through `num_frames` and we can comment that it will overwrite seconds.

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
```diff
@@ -460,13 +472,26 @@ def forward(
...
+                    self.fixed_latent_path,
+                    map_location=self.device,
+                    weights_only=True,
+                )
+                logger.info(
+                    f"Loaded fixed latent from {self.fixed_latent_path}, "
+                    f"shape={self.fixed_latent.shape}"
+                )
+            latents = self.fixed_latent.to(device=self.device, dtype=self.dtype)
         else:
```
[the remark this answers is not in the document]
wu6u3tw: Addressed in 933ae99 — pipeline now validates `self.fixed_latent.shape` against the computed `(batch, in_channels, latent_frames, latent_h, latent_w)` after lazy-loading and raises a clear `ValueError` (showing both shapes plus the path) when they disagree.

tensorrt_llm/serve/openai_server.py
```diff
@@ -2001,8 +2010,14 @@ async def _parse_video_generation_request(
...
             if "guidance_scale" in form and form["guidance_scale"]:
                 data["guidance_scale"] = float(form["guidance_scale"])
+            if "guidance_scale_2" in form and form["guidance_scale_2"]:
+                data["guidance_scale_2"] = float(form["guidance_scale_2"])
+            if "boundary_ratio" in form and form["boundary_ratio"]:
+                data["boundary_ratio"] = float(form["boundary_ratio"])
             if "guidance_rescale" in form and form["guidance_rescale"]:
                 data["guidance_rescale"] = float(form["guidance_rescale"])
+            if "num_frames" in form and form["num_frames"]:
+                data["num_frames"] = int(form["num_frames"])
```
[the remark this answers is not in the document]
wu6u3tw: Addressed in 76d3adc — multipart parsing now uses `form.get(...) not in (None, "")` for `guidance_scale_2`, `boundary_ratio`, and `num_frames`, so valid `0` / `0.0` values are no longer dropped. (Code moved from `openai_server.py` to `openai_video_routes.py`.)

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
wu6u3tw: Made it a hybrid in a42d32d — env var `TLLM_MLPERF_FIXED_LATENT_PATH` now takes precedence over `VisualGenArgs.fixed_latent_path`. So you can use the env var for the debug/benchmark knob workflow.

On the motivation: this is for MLPerf-deterministic generation. Different frameworks (e.g. PT vs ours) sample different latents from the same seed/inputs, which makes cross-framework numerical comparisons noisy. Pinning the latent to a reference tensor (sourced from PT) eliminates that variability so we can compare the rest of the pipeline. Keeping the `VisualGenArgs.fixed_latent_path` config field also preserves symmetry with `spatial_upsampler_path` / `distilled_lora_path` for deployment-time wiring.

tensorrt_llm/media/encoding.py
wu6u3tw: Addressed in bc5dd2f — the raw RGB byte dump path is replaced with `torch.save` to `.pt`, and `output_format = "pt"` is now the wire-level value (`"raw"` is gone).

tensorrt_llm/serve/openai_protocol.py
wu6u3tw: Addressed in a796cf1 — `num_frames` docstring is now explicit that it takes precedence over both `seconds` and `fps` (both are ignored when set), matching your suggestion.

wu6u3tw:
@zhenhuaw-me friendly ping — when you have a moment, could you take another look?

Since your last round of feedback, I've pushed:

- **a42d32d** — made `fixed_latent_path` a hybrid: env var `TLLM_MLPERF_FIXED_LATENT_PATH` now takes precedence over the `VisualGenArgs.fixed_latent_path` config field, so the debug/benchmark workflow you suggested is supported. (Left the discussion open for your call on whether the hybrid is acceptable, or you'd prefer env-var-only.)
- **933ae99** — added shape validation for the lazy-loaded fixed latent (CodeRabbit's suggestion, confirmed by them).
- **76d3adc** — zero-value-safe multipart parsing for `guidance_scale_2` / `boundary_ratio` / `num_frames` (CodeRabbit's suggestion, confirmed by them).
- **bc5dd2f** — replaced the raw RGB byte dump with `torch.save (.pt)`, matching your `pt` comment.
- **a796cf1** — made `num_frames` docstring explicit that it overrides both `seconds` and `fps`, per your suggestion.

All CI checks are green. Thanks!

chang-l (approved):
Approve to unblock — please also wait for @zhenhuaw-me to sign off as well

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -526,6 +526,24 @@ class VisualGenArgs(StrictBaseModel):
     # Skip warmup inference after loading (useful for testing or fast startup)
     skip_warmup: bool = False
 
+    # Path to a pre-computed fixed latent tensor (.pt file).
+    # When set, the pipeline replaces sampled latents with this tensor at
+    # request time (lazy-loaded on first inference). Used for MLPerf
+    # deterministic generation across runs.
+    fixed_latent_path: Optional[str] = PydanticField(
```
zhenhuaw-me: I see that we have an env var now. Can we drop this config from API?

wu6u3tw: Done in f4b5f51 — dropped `VisualGenArgs.fixed_latent_path` (and its `extra_attrs` plumbing in `from_pretrained`). The WAN pipeline now reads the path exclusively from `TLLM_MLPERF_FIXED_LATENT_PATH`, treating it as a pure debug/benchmark knob rather than part of the serving config surface.

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
wu6u3tw: Followed up in f4b5f51 — switched to env-var-only as you preferred. `VisualGenArgs.fixed_latent_path` is gone and the pipeline now reads `TLLM_MLPERF_FIXED_LATENT_PATH` directly. Resolving.

zhenhuaw-me (approved):

wu6u3tw:
@zhenhuaw-me I am wondering in you API change, is the extra param wired into your new API?

wu6u3tw:
Covered in other PRs

PR #13799, opened 2026-05-06
[https://nvbugs/5615248][fix] Beam history copies only on terminal steps

brb-nv:
## Description

This addresses https://nvbugspro.nvidia.com/bug/5615248. Two host-side optimizations for the beam-search decode path:

* `_prepare_beam_history` - only copy beam history on terminal steps.
    * The per-step D2H copies are skipped entirely on non-terminal steps via a small predictor entirely using information available on host.
    * On potentially-terminal steps they run on a private side stream so they never block the main stream. The side-stream completion is recorded into `SamplerEvent` and awaited host-side in `update_requests`.
    * A predictor-miss falls back to a synchronous `.cpu()` so correctness is preserved for any unforeseen finish pattern.

* `seq_slots` / `seq_lens` H2D - cast once per step at the top of `_process_requests` and reused. Eliminates 2 redundant H2D launches/step on the typical concurrency=1 / single-strategy beam-search path. Multi-strategy keeps the existing per-group fall-back.

Workload: TinyLlama-1.1B-Chat-v1.0 on L40S, ISL=100 / OSL=20, beam_width=10, `max_batch_size=1`, piecewise CUDA graphs, 5 runs × 16 requests pooled per side (n = 80). Sign convention: negative Δ = improvement.

| metric | base median (ms) | exp median (ms) | Δmedian (ms) |
|---|---:|---:|---:|
| TTFT | 10.895 | 10.803 | −0.092 |
| E2E | 83.918 | 83.470 | −0.448 |
| ITL | 0.367 | 0.365 | −0.0016 |

## Reproduction

A companion validation branch carries the bench harness used to produce the table above (kept off this MR to keep the diff minimal):

  - Branch: `user/brb/nvbug-5615248-beam-host-overhead`
  - Repro doc: `nvbugs_5615248/trtllm_bench/REPRO_V6_V7.md`

## Test Coverage

```
$ pytest tests/unittest/_torch/sampler/test_beam_search_speculative_d2h.py -s -v
$ pytest tests/unittest/_torch/sampler/test_beam_search.py -s -v
$ pytest tests/unittest/_torch/sampler/test_beam_search_util.py -s -v
```

## PR Checklist

## GitHub Bot Help

## Summary by CodeRabbit

* **Refactor**
  * Improved beam search sampling efficiency with optimized asynchronous data transfer handling to reduce synchronization overhead during batch processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13799)


tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar: As it is now, the sync D2H would appear to break overlap scheduling whenever `need_history.item()` is true. I suspect that this may reduce perf for other workloads. Unless we can exclude this, it would probably be good to restrict this optimization to workloads for which it has been demonstrated to help.

Alternatively, have you considered overlapping the D2H with compute by using CUDA streams? Such an approach might be beneficial for all workloads.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1139,18 +1153,83 @@ def _copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
...
+                self._batch_stream_ctx = torch.cuda.stream(self._d2h_side_stream)
+                self._batch_stream_ctx.__enter__()
+            dest.copy_(src, non_blocking=True)
+        else:
+            with torch.cuda.stream(self._d2h_side_stream):
+                dest.copy_(src, non_blocking=True)
+        return dest
+
+    @contextlib.contextmanager
+    def _side_stream_d2h_batch(self) -> Iterator[None]:
```
ixlmar: The batch copy feature carries significant complexity. Are repeated `with torch.cuda.stream:` blocks indeed so expensive that this results in a net benefit? How much latency difference does this make?

If we really need this, it would be nice to get rid of the de-facto-global (there is only one sampler) flags `_batch_stream_ctx` and `_in_side_stream_batch`. Instead:

* `_side_stream_d2h_batch` could yield a `SideStreamCopier` (or similar), so one could write
  ```python
  with self._side_stream_d2h_batch() as copier:
    dst1 = copier.copy_to_host(src1)
    dst2 = copier.copy_to_host(src2)
    ...
  ```

  If nested batches should be fused as well (`prev_in_batch` logic), one could use the more Pythonic `contextvars` (which also avoids pitfalls with threading etc.), but this is likely going to be slower. I would suggest to forego this complexity and explicitly pass around the `copier` to manually address cases where such fusion is needed.
  Perhaps one could also do `if torch.cuda.current_stream() is _d2h_side_stream: ...`, which might be fast enough and simplify the code.

* Alternatively, one could leverage threading in the way that the `AsyncWorkerMixin` is doing it (it maintains a separate thread with it's own fixed CUDA stream). Have you considered modifying/extending `AsyncWorkerMixin` to implement this optimization?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1139,18 +1153,83 @@ def _copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
...
+        dest = torch.empty_like(src, device="cpu", pin_memory=prefer_pinned())
+        if self._in_side_stream_batch:
+            # Lazily enter the stream context on the first copy so empty
+            # batches (predictor-driven all-skip steps) cost nothing.
+            if self._batch_stream_ctx is None:
+                self._batch_stream_ctx = torch.cuda.stream(self._d2h_side_stream)
+                self._batch_stream_ctx.__enter__()
+            dest.copy_(src, non_blocking=True)
+        else:
+            with torch.cuda.stream(self._d2h_side_stream):
```
ixlmar: Can `torch.cuda.StreamContext` be reused? `with torch.cuda.stream(self._d2h_side_stream): ... -> with self._d2h_side_stream_ctx: ...` (might eliminate the need for the batched-copy feature).

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1139,18 +1153,83 @@ def _copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
...
+        """
+        prev_in_batch = self._in_side_stream_batch
+        prev_batch_stream_ctx = self._batch_stream_ctx
+        self._in_side_stream_batch = True
+        self._batch_stream_ctx = None
+        try:
+            yield
+        finally:
+            if self._batch_stream_ctx is not None:
+                self._batch_stream_ctx.__exit__(None, None, None)
```
ixlmar: To me, directly calling `__enter__` and `__exit__` appears to undermine the idea of context managers, is quite difficult to read, and more error prone:

```python
with self._side_stream_d2h_batch():
   ...
   with self._side_stream_d2h_batch():
      # trigger first copy here, calls __enter__
   ...
   # at this point, the inner context manager will have called __exit__, although subsequent copies could still have reused the stream context
```

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1066,6 +1072,14 @@ def _async_worker_init(self, enable_async_worker: bool) -> None:
...
         self._async_worker: futures.ThreadPoolExecutor | None = None
         self._async_worker_futures: list[futures.Future[Any]] = []
+        # Side-stream D2H state (lazily initialized; see
+        # _async_d2h_via_side_stream and _side_stream_d2h_batch). The pending
+        # flag is cleared by _record_sampler_event after recording the
+        # side-stream event into SamplerEvent.
+        self._d2h_side_stream: torch.cuda.Stream | None = None
+        self._d2h_side_stream_pending: bool = False
+        self._in_side_stream_batch: bool = False
+        self._batch_stream_ctx: Any | None = None
```
ixlmar: `Any` -> `torch.cuda.StreamContext`

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1139,18 +1153,83 @@ def _copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
...
     def _record_sampler_event(self) -> SamplerEvent:
         cuda_event = torch.cuda.Event()
         cuda_event.record()
 
+        # Record a separate event on the side stream rather than calling
+        # current_stream.wait_stream(side); the latter would serialize the
+        # next iteration's main-stream work behind these copies. The event
+        # is synchronized host-side by SamplerEvent.synchronize().
+        side_stream_event: torch.cuda.Event | None = None
+        if self._d2h_side_stream_pending:
```
ixlmar: I think renaming `_d2h_side_stream_pending` to `_d2h_side_stream_sync_pending` would make the intention clearer.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1066,6 +1072,14 @@ def _async_worker_init(self, enable_async_worker: bool) -> None:
         self._enable_async_worker = enable_async_worker
         self._async_worker: futures.ThreadPoolExecutor | None = None
         self._async_worker_futures: list[futures.Future[Any]] = []
+        # Side-stream D2H state (lazily initialized; see
+        # _async_d2h_via_side_stream and _side_stream_d2h_batch). The pending
+        # flag is cleared by _record_sampler_event after recording the
+        # side-stream event into SamplerEvent.
+        self._d2h_side_stream: torch.cuda.Stream | None = None
```
ixlmar: Since this is done only once, I would suggest to always initialize `self._d2h_side_stream: torch.cuda.Stream` (or even `torch.cuda.StreamContext`, cf. other comments) here and get rid of the various `None` checks/asserts.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -2207,6 +2286,10 @@ def __init__(self, args: Args):
 
         self._async_worker_init(args.enable_async_worker)
 
+        # 1-step-lagged host mirror of beam_search_store.first_finish_reasons,
+        # consumed by _predict_beam_history_terminal.
+        self._prev_first_finish_reasons_host: dict[int, torch.Tensor] = {}
```
ixlmar: `first_finish_reasons` is a `torch.Tensor`, what does the dict key correspond to here? If there is a need for inhomogeneous dimensions, could we use `list[torch.Tensor]` instead?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3042,42 +3132,85 @@ def _prepare_beam_history(
...
-            torch.gather(input=current_path, dim=0, index=cache_indirection, out=new_path)
-            return new_path
-
+        sampled_log_probs_dev: torch.Tensor | None = None
+        sampled_logprobs_indices_dev: torch.Tensor | None = None
+        cum_logprobs_dev: torch.Tensor | None = None
         if request.py_return_log_probs:
             log_probs_store = self.store.log_probs_store
-            sampled_log_probs = log_probs_store.sampled_log_probs[
+            sampled_log_probs_dev = log_probs_store.sampled_log_probs[
```
ixlmar: Evaluate `self._predict_beam_history_terminal` before `if request.py_return_log_probs:` and handle copies in this same block. Like this, `sampled_log_probs_dev`, `sampled_logprobs_indices_dev`, and `cum_logprobs_dev` can be limited to this block and we can avoid the `| None` with all the entailed checks/asserts.

ixlmar: This is a general observation: We should try to reorder variable declarations and control flow to get rid of `variable: SomeType | None` (in particular for variables whose "noneness" is correlated and/or gated by some flag), `if variable is not None`, and `assert variable is not None` as far as possible (the base branch had very few of those).

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3042,42 +3132,85 @@ def _prepare_beam_history(
...
+        current_path_host: torch.Tensor | None = None
+        sampled_log_probs_host: torch.Tensor | None = None
+        sampled_logprobs_indices_host: torch.Tensor | None = None
+        cum_logprobs_host: torch.Tensor | None = None
+        if self._predict_beam_history_terminal(
+            request,
+            num_generated_tokens=num_generated_tokens,
+            num_tokens=num_tokens,
+        ):
+            cache_indirection_host = self._async_d2h_via_side_stream(cache_indirection_dev)
```
ixlmar: This seems to bypass `self._copy_to_host`, which was introduced to support confidential compute (CC @dhansen-nvidia). Can we integrate the enhancements from this PR into the `AsyncWorkerMixin` somehow to reduce the number of possible code paths?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3042,42 +3132,85 @@ def _prepare_beam_history(
...
+                sampled_logprobs_indices_host = self._async_d2h_via_side_stream(
+                    sampled_logprobs_indices_dev
+                )
+                cum_logprobs_host = self._async_d2h_via_side_stream(cum_logprobs_dev)
+
+        def _builder() -> BeamHistory | None:
+            if not need_history.item():
+                return None
+
+            if cache_indirection_host is not None and current_path_host is not None:
```
ixlmar: Separate `if`s would seem less brittle.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3042,42 +3132,85 @@ def _prepare_beam_history(
...
+        if self._predict_beam_history_terminal(
+            request,
+            num_generated_tokens=num_generated_tokens,
+            num_tokens=num_tokens,
+        ):
+            cache_indirection_host = self._async_d2h_via_side_stream(cache_indirection_dev)
+            current_path_host = self._async_d2h_via_side_stream(current_path_dev)
+            # The three log-prob *_dev tensors are bound iff py_return_log_probs;
+            # the conjunction also narrows them for mypy.
+            if (
```
ixlmar: Separate `if`s would seem less brittle.

tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar: If optionality (`| None`) cannot be avoided, then for correlated fields we could use `SomeCompoundType | None` to make correlations explicit.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3272,11 +3443,15 @@ def _prepare_beam_histories(
...
         the corrected tokens and logprobs for each beam of a request.
+
+        The loop is wrapped in _side_stream_d2h_batch so the side stream is
+        made current at most once per call instead of once per copy.
         """
-        return [
-            self._prepare_beam_history(req, finish_reasons=finish_reasons[req.py_seq_slot])
-            for req in requests
-        ]
+        with self._side_stream_d2h_batch():
```
ixlmar: Just realizing that `with self._side_stream_d2h_batch():` changes the default stream. Thus, any work submitted in this context (incl. non-D2H tasks) would appear to end up in the wrong stream. Suggest to extend `AsyncWorkerMixin` instead, or refactor using the explicit `copier` pattern sketched above instead of global and/or thread-local state.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3621,6 +3803,8 @@ def _sample_batched_by_strategy(
         req_offsets: torch.Tensor,
         seq_slots: torch.Tensor,
         seq_lens: Optional[torch.Tensor] = None,
+        seq_slots_cuda: Optional[torch.Tensor] = None,
```
ixlmar: Can we do without `Optional` (`_process_requests` always provides that)?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3042,42 +3132,85 @@ def _prepare_beam_history(
...
+            if not need_history.item():
+                return None
+
+            if cache_indirection_host is not None and current_path_host is not None:
+                # Predictor-hit: side-stream D2H already awaited via SamplerEvent.
+                cache_indirection = cache_indirection_host
+                current_path = current_path_host
+            else:
+                # Predictor-miss fallback: synchronous D2H on the main stream.
+                cache_indirection = cache_indirection_dev.cpu()
```
ixlmar: This code path should also be covered in test cases. Since this syncs and the tests forbid any additional syncs, the tests likely need further adaptation.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3262,6 +3397,42 @@ def _check_stop_words_length(request: LlmRequest) -> bool:
...
+        Known miss: all beams hit end_id on the same step from a clean
+        state. Rare in practice; covered by _builder's .cpu() fallback.
+        """
+        if num_generated_tokens >= request.py_max_new_tokens or num_tokens >= self.max_seq_len:
+            return True
+        if request.py_stop_words_list is not None:
+            return True
+        assert request.py_seq_slot is not None
+        prev = self._prev_first_finish_reasons_host.get(request.py_seq_slot)
+        if prev is not None and bool((prev > 0).any().item()):
```
ixlmar: Nit: `prev.any().item()`

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3262,6 +3397,42 @@ def _check_stop_words_length(request: LlmRequest) -> bool:
             return longest_stop_word_len > 1
         return False
 
+    def _predict_beam_history_terminal(
```
ixlmar: Naming suggestion: `_predict_beam_search_is_likely_finishing`

tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar: On another look, I think a simple solution could be the `copier` approach with something like

```python
def copy_to_host(self, src):
  dst = torch.zeros_like(src, ...)
  self._copy_list.append_task((dst, dst))
  return dst

def __exit__(self, ...):
  with self.side_stream:
    # process self._copy_list
```

ixlmar: Also, one could have `copier.record_event() -> torch.cuda.Event`.

tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar: In any case, the new code paths require unit tests to cover beam search with confidential compute enabled.

ixlmar (commented):
The changes make sense, but we need to figure out how to keep the overall sampler code maintainable and how to not regress any existing features.

Since the side-stream D2H copies are supposed to overlap with compute, do we even need the speculative D2H-optout with fallback to `.cpu()` and resulting potential syncs that need to be covered in unit tests and may alter performance characteristics unexpectedly (for big enough models the risk of breaking overlap scheduling by `.cpu()` may outweigh any perf benefit)?

Generally, this is a form of speculative execution with the usual performance implications. Although I agree that the `.cpu()` branch would appear to be hit only rarely, this might be unexpected to users running benchmarks etc. and thus perhaps it would be better to gate this optimization by some configuration option (having "spec" in its name)?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3042,42 +3132,85 @@ def _prepare_beam_history(
...
+            else:
+                # Predictor-miss fallback: synchronous D2H on the main stream.
+                cache_indirection = cache_indirection_dev.cpu()
+                current_path = current_path_dev.cpu()
 
-            # enqueue async D2H copies
-            sampled_log_probs = self._copy_to_host(sampled_log_probs)
-            sampled_logprobs_indices = self._copy_to_host(sampled_logprobs_indices)
-            cum_logprobs = self._copy_to_host(cum_logprobs)
+            new_path = torch.zeros_like(current_path)
```
ixlmar: Maybe keep `_post_process_path` instead of inlining (more readable and more expressive)

ixlmar: Similarly, the previous body of the (longer variant of) `_maybe_postprocess_logprobs` is perhaps better kept as a separate function.

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1a3672d8d8ccbd0509772cf22273402b2edc0b4b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1a3672d8d8ccbd0509772cf22273402b2edc0b4b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1a3672d8d8ccbd0509772cf22273402b2edc0b4b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1a3672d8d8ccbd0509772cf22273402b2edc0b4b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1a3672d8d8ccbd0509772cf22273402b2edc0b4b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1a3672d8d8ccbd0509772cf22273402b2edc0b4b

mikeiovine (approved):
Stamp on behalf of spec decode reviewers. Should be reviewed by sampler owners

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: I added new tests exercising the speculative codepath specifically.

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: I have a separate copier now. Can you please take a look once again?

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Hope newer state of the MR looks better.

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: I have added a few dataclasses with logical grouping. Please let me know if that makes sense.

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: The speculative feature is now opt-in and turned off when CC is enabled. Can you please have a look once again?

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Added some dataclasses to group relevant variables together.

Most of the `assert variable is not None` in this MR have been added to make mypy happy.

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Switched to list of tensors now.

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Done.

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Hope the current refactor handled this better.

brb-nv:
Hi @ixlmar, I refactored to keep the current codepath as is and made the incoming feature opt-in. It's also turned off when CC is being used. Guarding it with an env variable for now. Can you please take a look once again?

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Can you please review the refactored MR?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1050,11 +1057,108 @@ class SampleStateTorch(SampleState[SampleStateTensorsHostTorch, SampleStateTenso
...
+        assert not self._flushed, "copy_to_host called after the copier was flushed"
+        dst = torch.empty_like(src, device="cpu", pin_memory=prefer_pinned())
+        self._tasks.append((dst, src))
+        return dst
+
+    def _flush(self) -> None:
+        if not self._tasks:
+            return
+        self._side_stream.wait_stream(torch.cuda.current_stream())
+        with torch.cuda.stream(self._side_stream):
```
ixlmar: Can this be changed to the following?

```python
self._side_stream_ctx =  torch.cuda.stream(self._side_stream)
...
with self._side_stream_ctx:
   ...
```

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1139,7 +1245,17 @@ def _copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
             dest.copy_(src, non_blocking=True)
         return dest
 
-    def _record_sampler_event(self) -> SamplerEvent:
+    def _make_side_stream_copier(self) -> _SideStreamCopier:
```
ixlmar: Implementing the context manager protocol directly on `_SideStreamCopier` does not force the caller to use `with`. Suggestion:

```
@contextmanager
def _make_side_stream_copier() -> Generator[_SideStreamCopier, None, None]:
  try:
    copier = ...
    yield copier
    copier.record_event()  # can inline .flush into .record_event and drop copier._flushed
  finally:
    cleanup()
```

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -2207,6 +2327,27 @@ def __init__(self, args: Args):
 
         self._async_worker_init(args.enable_async_worker)
 
+        # Opt-in switch for the speculative beam-history D2H path; default off
```
ixlmar: We already have `TorchSampler.Args`, so you can mimic how `LlmArgs.disable_flashinfer_sampling` is relayed.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -2207,6 +2327,27 @@ def __init__(self, args: Args):
...
+        # uses the safe per-step _copy_to_host. Stopgap until exposed in LlmArgs.
+        # The speculative path bypasses AsyncWorkerMixin's _copy_to_host, so it
+        # is incompatible with the async worker and is force-disabled there.
+        self._use_speculative_beam_history_d2h: bool = (
+            os.environ.get("TRTLLM_ENABLE_BEAM_SEARCH_SPECULATIVE_D2H", "0") == "1"
+        )
+        if self._use_speculative_beam_history_d2h and self.async_worker_enabled():
+            logger.warning(
```
ixlmar, line 2337: After moving the env var into `LlmArgs.`, this could be validated early in `LlmArgs.validate...`.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3006,6 +3151,15 @@ def _prepare_beam_history(
         If logprobs are requested, the function also corrects the stored logprobs for each beam.
         The function returns a BeamHistory object that contains the corrected tokens and logprobs for each beam.
 
+        Dispatches to one of two D2H strategies based on
+        self._use_speculative_beam_history_d2h:
+          - Default: _prepare_beam_history_default, _copy_to_host per step.
+          - Speculative (TRTLLM_ENABLE_BEAM_SEARCH_SPECULATIVE_D2H=1):
+            _prepare_beam_history_speculative skips definitely-non-terminal
```
ixlmar: "definitely" -> "likely"

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3042,95 +3197,227 @@ def _prepare_beam_history(
...
+                log_probs=log_probs_host,
+            )
 
         def _builder() -> BeamHistory | None:
             if not need_history.item():
                 return None
 
-            new_path = _post_process_path()
-            new_logprobs, new_logprobs_indices, cum_logprobs = _maybe_postprocess_logprobs()
+            if host_snapshot is not None:
```
ixlmar, line 3252: This looks correct, but could we reduce the amount of duplication between `_prepare_beam_history_speculative` and `_prepare_beam_history_default`? This would facilitate testing and eliminate the need to keep two near-copies in sync. Suggestion:

* Remove `_prepare_beam_history_default` and make it a special case of `_prepare_beam_history_speculative` (call it `_prepare_beam_history`).

* Change `if self._predict_beam_search_is_likely_finishing...` -> `if no_speculation or self._predict_beam_search_is_likely_finishing...`.

* Change the argument from `side_stream_copier: _SideStreamCopier | None = None` to `d2h_copier: Callable[[torch.Tensor], torch.Tensor]` (mandatory). Pass in the relevant bound method via this argument, either `side_stream_copier.copy_to_host` or `self.copy_to_host`.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3258,25 +3558,76 @@ def _check_stop_words_length(request: LlmRequest) -> bool:
             _, cumsum = request.py_stop_words_list
             if -1 in cumsum:
                 cumsum = cumsum[: cumsum.index(-1)]
-            longest_stop_word_len = np.max(np.diff(cumsum, prepend=0), initial=0).item()
+            longest_stop_word_len = int(np.max(np.diff(cumsum, prepend=0), initial=0).item())
```
ixlmar: `np.diff(np.array(cumsum, dtype=np.int32))...`?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3258,25 +3558,76 @@ def _check_stop_words_length(request: LlmRequest) -> bool:
...
         return False
 
+    def _predict_beam_search_is_likely_finishing(
+        self,
+        request: LlmRequest,
+        *,
+        num_generated_tokens: int,
+        num_tokens: int,
+    ) -> bool:
+        """Predict whether this step might finalize the beam history.
```
ixlmar: nit: "might trigger beam history finalization"

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3258,25 +3558,76 @@ def _check_stop_words_length(request: LlmRequest) -> bool:
...
+
+        True triggers an eager side-stream D2H copy; False skips it. False
```
ixlmar: This paragraph is a bit out of context now, perhaps it can be removed.

tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar, line 3474: Out of scope, but host overhead could be reduced by caching the result of this method (e.g. as `request._py_...`.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3258,25 +3558,76 @@ def _check_stop_words_length(request: LlmRequest) -> bool:
...
+        if not self._use_speculative_beam_history_d2h:
+            builders = [
+                self._prepare_beam_history(req, finish_reasons=finish_reasons[req.py_seq_slot])
+                for req in requests
+            ]
+            return builders, None
+
+        # Speculative path: queue D2H copies onto one shared side-stream
+        # copier; flushed on exit.
+        with self._make_side_stream_copier() as side_stream_copier:
```
ixlmar: Nit:

```python
ctx = self._make_side_stream_copier() if self._use_speculative_beam_history_d2h else nullcontext()
with ctx as side_stream_copier:
            builders = [
                self._prepare_beam_history(
                    req,
                    finish_reasons=finish_reasons[req.py_seq_slot],
                    side_stream_copier=side_stream_copier,
                )
                for req in requests
            ]
        return builders, side_stream_copier.record_event() if side_stream_copier is not None else None
```

ixlmar (commented):
Still need to look at the new test.

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/1216a563b7b6125c739d3c2e319177e6f68ff98b

I'm still forced to keep the ugly int() because mypy complains:
```
error: Returning Any from function declared to return "bool"  [no-any-return]
```

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/commit/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/commit/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/commit/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/commit/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/commit/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Makes sense. Updated. https://github.com/NVIDIA/TensorRT-LLM/commit/1216a563b7b6125c739d3c2e319177e6f68ff98b

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Makes sense. Leaving this as is for now.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1055,11 +1056,96 @@ class SampleStateTorch(SampleState[SampleStateTensorsHostTorch, SampleStateTenso
...
+class _BeamHistoryDeviceSlices:
+    """Device-side beam-history slices. `log_probs` is bound iff log-probs are requested."""
+
+    cache_indirection: torch.Tensor
+    current_path: torch.Tensor
+    log_probs: _BeamHistoryLogProbsSlices | None
+
+
+@dataclass(kw_only=True, frozen=True)
+class _BeamHistoryHostSnapshot:
```
ixlmar: Currently, there is no difference between `_BeamHistoryHostSnapshot` and `_BeamHistoryDeviceSlices`, so we could use the same type for both.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1055,11 +1056,96 @@ class SampleStateTorch(SampleState[SampleStateTensorsHostTorch, SampleStateTenso
...
+        self._tasks: list[tuple[torch.Tensor, torch.Tensor]] = []
+        self.event: torch.cuda.Event | None = None
+
+    def copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
+        """Queue a non-blocking D2H copy of src and return its pinned-CPU dst."""
+        dst = torch.empty_like(src, device="cpu", pin_memory=prefer_pinned())
+        self._tasks.append((dst, src))
+        return dst
+
+    def record_event(self) -> torch.cuda.Event | None:
```
ixlmar: Maybe call this `commit` or similar, to express that this actually executes the work.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1055,11 +1056,96 @@ class SampleStateTorch(SampleState[SampleStateTensorsHostTorch, SampleStateTenso
...
+        self,
+        side_stream: torch.cuda.Stream,
+        side_stream_ctx: torch.cuda.StreamContext,
+    ) -> None:
+        self._side_stream = side_stream
+        self._side_stream_ctx = side_stream_ctx
+        self._tasks: list[tuple[torch.Tensor, torch.Tensor]] = []
+        self.event: torch.cuda.Event | None = None
+
+    def copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
```
ixlmar: Maybe call it `stage_copy_to_host` or similar.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1144,7 +1236,27 @@ def _copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
...
+
+        On exit, queued copies are flushed and an event is recorded;
+        retrieve it from `copier.event` (`None` if nothing was queued).
+        The `@contextmanager` form forces callers to use `with`.
+        """
+        copier = _SideStreamCopier(self._d2h_side_stream, self._d2h_side_stream_ctx)
+        try:
+            yield copier
+        finally:
+            copier.record_event()
```
ixlmar: This will enqueue copies even when the outer context raises an exception. Let's maybe remove the `try`-`finally` altogether.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -1144,7 +1236,27 @@ def _copy_to_host(self, src: torch.Tensor) -> torch.Tensor:
...
+
+    def _record_sampler_event(
+        self, side_stream_event: torch.cuda.Event | None = None
+    ) -> SamplerEvent:
+        """Record a SamplerEvent on the main stream. side_stream_event,
```
ixlmar: Add an empty line after the short-form description (PEP-257).

tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar: I still see both variants spelled out. The idea was to only have one `_prepare_beam_history_(..., d2h_copier: Callable[[torch.Tensor], torch.Tensor])` (a slight adaptation of the current `_prepare_beam_history_speculative` as outlined above). The calling code then only needs to pass `d2h_copier=self._copy_to_host if not use_side_stream else side_stream_copier.copy_to_host`.

tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar: That is because mypy cannot know the dtype of the array (it is not part of the type annotation). We typically use `cast(int, ...)` in those cases (many occurrences in that module already).

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -3262,25 +3559,72 @@ def _check_stop_words_length(request: LlmRequest) -> bool:
...
         return False
 
+    def _predict_beam_search_is_likely_finishing(
+        self,
+        request: LlmRequest,
+        *,
+        num_generated_tokens: int,
+        num_tokens: int,
+    ) -> bool:
+        """Predict whether this step might trigger beam history finalization.
```
ixlmar: Nit: "might" -> "is likely to" (hadn't noticed this on the previous pass...)

tensorrt_llm/_torch/pyexecutor/sampler.py
ixlmar: Perhaps add a "TODO" comment.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -3841,6 +3816,17 @@ class TorchLlmArgs(BaseLlmArgs):
         "irrespective of confidential compute state.",
         status="prototype")
 
+    enable_speculative_beam_history_d2h: bool = Field(
```
ixlmar, line 3819: The current implementation ties speculation and side-stream copy together, although both are orthogonal features. Is there any scenario in which we would not want to use the side stream? Otherwise, we could have a separate knob which simply alters the behavior of `TorchSampler._copy_to_host`. Thus, the latter would either use the async worker if needed / confidential compute is enabled, or otherwise side-stream D2H, unless users disable that explicitly via `LlmArgs`. We would still need to add a context manager, so `TorchSampler._copy_to_host` would morph into `TorchSampler._batch_d2h_copier: ContextManager[Callable[[torch.Tensor], torch.Tensor]`, but otherwise the changeset might become simpler. (CC @Funatiq)

ixlmar (commented):
Had a superficial look at the tests as well (looking good)

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Sorry, I missed your point. Updating.

tensorrt_llm/llmapi/llm_args.py
brb-nv: This sounds like a large change we can take up in a follow-up MR. Created a JIRA task for myself to follow-up:
https://jirasw.nvidia.com/browse/TRTLLM-12730

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Calling this commit now. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Renamed to `stage_copy_to_host`. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Updated. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Looks better. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Added TODO: https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

brb-nv: JIRA to follow-up: https://jirasw.nvidia.com/browse/TRTLLM-12729

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Right, should've used `cast`. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

tensorrt_llm/_torch/pyexecutor/sampler.py
brb-nv: Unified here. https://github.com/NVIDIA/TensorRT-LLM/pull/13799/commits/ebdd632cf81fa99900c165d27d78ebc877503cd6

juney-nvidia (approved):

PR #13809, opened 2026-05-06
[TRTLLMINF-54][feat] SlurmConfig boundary throws typed InfraFailure

dpitman-nvda:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error classification for SLURM job failures with more reliable and safer retry behavior.

## Description

Commit 3 of 5 (typed-slurm-producer).

Migrates the one OUR-code producer of an infra-failure substring into a typed throw. The catch boundary at L0_Test.groovy that wraps SlurmConfig.checkJobStatus()'s "is no longer active" rethrow now throws InfraFailure(TRANSIENT, SLURM, "<typed:slurm-job-inactive>") instead of a plain Exception with an enriched message string. The other 21 patterns in PATTERN_CATALOG come from third-party sources (Jenkins remoting plugin, durable-task plugin, kubelet, JDK NIO, SLURM CLI, SSH) that we cannot typed-throw at; the catalog remains as fallback for those.

Why this matters for nested SLURM-inside-K8s retry correctness:

The SLURM dispatcher pod runs inside a K8s pod, so a SLURM stage is wrapped by both retries: an inner runLLMTestlistOnSlurm retry loop AND an outer runKubernetesPodWithInfraRetry. Before this commit, the SLURM retry's "is no longer active" exhaustion would bubble out as a plain Exception, and the outer K8s consumer would call classify() which would substring-match "is no longer active" against the catalog (scope=BOTH in the new catalog) and treat it as a fresh K8s infra failure -- effectively doubling the retry budget across the same root cause.

After this commit, the inner SLURM retry's exhaustion bubbles out as a typed InfraFailure(scope=SLURM). The outer K8s classify() hits the scope-isolation guard (added in the classifier-base commit) and wraps it as UserFailure -- the outer rethrows without retry. End-to-end trace documented in the inline comment at the catch site.

A focused test for this nested behavior would force-throw a typed InfraFailure(SLURM) from inside an inner SLURM retry, run from a K8s parallelJobs entry that wraps it via runKubernetesPodWithInfraRetry, and assert that the outer retries 0 times (the SLURM-scoped throw is rejected by the outer's scope guard). Smoke test pattern in line with the prior synthetic-replay tests gated behind a build parameter.

## Test Coverage

N/A, this is a CI change

## PR Checklist

## GitHub Bot Help


zeroepoch (approved):

dpitman-nvda:
/bot skip --comment "Got through all but one stage successfully, calling it good"

PR #13811, opened 2026-05-06
[None][feat] Indexer topk opt

pcastonguay:
@coderabbitai summary

<!--
Please write the PR title by following this template:

**[JIRA ticket/NVBugs ID/GitHub issue/None][type] Summary**

Valid ticket formats:
  - JIRA ticket: [TRTLLM-1234] or [FOOBAR-123] for other FOOBAR project
  - NVBugs ID: [https://nvbugs/1234567]
  - GitHub issue: [#1234]
  - No ticket: [None]

Valid types (lowercase): [fix], [feat], [doc], [infra], [chore], etc.

Examples:
  - [TRTLLM-1234][feat] Add new feature
  - [https://nvbugs/1234567][fix] Fix some bugs
  - [#1234][doc] Update documentation
  - [None][chore] Minor clean-up

Alternative (faster) way using CodeRabbit AI:

**[JIRA ticket/NVBugs ID/GitHub issue/None] @coderabbitai title**

NOTE: "@coderabbitai title" will be replaced by the title generated by CodeRabbit AI, that includes the "[type]" and title.
For more info, see /.coderabbit.yaml.

-->

## Description

## GSM8K accuracy (1319 samples, 5-shot, greedy)

| Metric | baseline | opt | Δ |
|---|---:|---:|---:|
| `exact_match,strict-match` | 0.9545 ± 0.0057 | 0.9553 ± 0.0057 | +0.0008 |
| `exact_match,flexible-extract` | 0.9538 ± 0.0058 | 0.9545 ± 0.0057 | +0.0008 |
| eval time (s) | 207.3 | 209.2 | +1.9 |

Within stderr — **accuracy unchanged**.

## Serving performance (concurrency 256, ISL 8192 / OSL 1024, gen_only)

Dataset: `random-v2/deepseek-v4-flash-8192-1024-20000-ratio-08_for_serve.json`. 1× ctx server + 1× gen server, max-concurrency 256, 1 round, non-streaming.

| Metric | baseline  | opt | Δ |
|---|---:|---:|---:|
| request throughput (req/s) | 2.169 | 2.280 | **+5.13 %** |
| output throughput (tok/s) | 2024.4 | 2128.2 | **+5.13 %** |
| total token throughput (tok/s) | 18112 | 19041 | **+5.13 %** |
| mean TTFT (ms) | 84338.8 | 78120.9 | **−7.37 %** |
| median TTFT (ms) | 84337.5 | 78119.1 | −7.37 % |
| p99 TTFT (ms) | 84379.4 | 78150.7 | −7.38 % |
| mean TPOT (ms) | 33.29 | 33.72 | +1.29 % |
| median TPOT (ms) | 33.34 | 33.77 | +1.29 % |
| p99 TPOT (ms) | 33.40 | 33.80 | +1.21 % |
| mean ITL (ms) | 3150.9 | 3191.7 | +1.29 % |
| median ITL (ms) | 3457.7 | 3496.3 | +1.12 % |
| mean E2EL (ms) | 115368 | 109551 | **−5.04 %** |
| duration (s) | 118.0 | 112.3 | −4.88 % |

→ **+5.1 % throughput, −7.4 % TTFT, −5.0 % E2EL.** TPOT regresses ~1.3 %.



Before / After comparison — indexer top-K decode optimization                                                                                                                                                              
                                                                                   
  Target kernel: topKPerRowDecode 

 | Metric | Before | After | Delta |                                                                                                                                                                                        
  |---|---|---|---|                                                                                                                                                                                                          
  | Mean per-call duration | 123.12 us | 10.48 us | **11.7× faster** |
  | Min / Max per-call | 122.78 / 124.61 us | 2.24 / 19.17 us | much wider range (split path) |                                                                                                                              
  | Total kernel time | 36.94 ms | 6.29 ms | **−83.0%** |                                                                                                                                                                    
  | Invocations | 300 | 600 | 2× (split + merge passes) |                                                                                                                                                                    
  | Share of total GPU time | 9.86% | 1.92% | −7.94 pp |     

<!--
Please explain the issue and the solution in short.
-->

## Test Coverage

<!--
Please list clearly what are the relevant test(s) that can safeguard the changes in the PR. This helps us to ensure we have sufficient test coverage for the PR.
-->

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/kernels/IndexerTopK.h
+++ b/cpp/tensorrt_llm/kernels/IndexerTopK.h
@@ -27,6 +27,13 @@ TRTLLM_NAMESPACE_BEGIN
 
 namespace kernels
 {
+// Number of blocks-per-row used by the multi-block split + merge dispatch path of
+// invokeIndexerTopKDecode. Returns 1 when the single-block path is preferred.
+// Callers that allocate aux buffers must use this same helper to size them, and
+// must pass the same splitWorkThreshold they will pass to invokeIndexerTopKDecode
+// (a value <= 0 selects the internal default).
+int computeIndexerTopKDecodeBlocksPerRow(int numRows, int numColumns, int splitWorkThreshold = 0);
+
 void invokeIndexerTopKDecode(float const* logits, int const* seqLens, int* indices, float* outLogitsAux,
     int* outIndicesAux, int const splitWorkThreshold, int const numRows, int const numColumns, int const stride0,
     int const stride1, int const next_n, int const topK = 2048, int const* preIdx = nullptr, int const preIdxStride = 0,
--- a/cpp/tensorrt_llm/kernels/indexerTopK.cu
+++ b/cpp/tensorrt_llm/kernels/indexerTopK.cu
@@ -21,6 +21,7 @@
 #include "tensorrt_llm/common/envUtils.h"
 #include "tensorrt_llm/kernels/heuristicTopKDecode.h"
 #include "tensorrt_llm/kernels/noAuxTcKernels.h"
+#include <algorithm>
 #include <cfloat>
 #include <cooperative_groups.h>
 #include <cooperative_groups/reduce.h>
@@ -162,6 +163,20 @@ __device__ bool processHistogramStep(int const* indices, float const* logits, in
     int& thresholdBinIdx, SmemOutputType& smemOutput, int* smemThresholdBinIdx, int* smemFinalDstIdx,
     int* smemFinalBinSize, int* smemFoundTopKValues, SmemFinalType& smemFinal, int stride1, int rowStart, int topK)
 {
+    // Step 0 is the fp16 fast path; if it could not resolve top-K (threshold bin
+    // exceeded kNumFinalItems) we restart the fp32 radix from scratch in steps 1-3.
+    // Discard any candidates step 0 wrote into smemOutput so step 1 doesn't
+    // double-count valid entries that fall under both fp16 and fp32 thresholds
+    // (this is what produced 2x duplicated indices when many -FLT_MAX padding
+    // entries dominated the threshold bin in the multi-block merge path).
+    if constexpr (step == 1)
+    {
+        if (threadIdx.x == 0)
+        {
+            smemFoundTopKValues[0] = 0;
+            smemFinalDstIdx[0] = 0;
+        }
+    }
     // Clear the histogram.
 #pragma unroll
     for (int idx = threadIdx.x; idx < kNumBins; idx += kNumThreadsPerBlock)
@@ -498,6 +513,12 @@ static __device__ void topKPerRowJob(int const* indices, float const* logits, in
             for (int ii = 0; ii < kNumFinalItemsPerThread; ++ii)
             {
                 finalLogits[ii] = -FLT_MAX;
+                // Indices must be paired with the -FLT_MAX sentinel logit so unused
+                // slots survive SortDescendingBlockedToStriped and the post-sort copy
+                // as -1 rather than garbage. Without this, when the threshold bin is
+                // dominated by -FLT_MAX padding (multi-block merge path) the trailing
+                // top-K slots receive arbitrary register contents.
+                finalIndices[ii] = -1;
             }
 
             // Read the elements from SMEM.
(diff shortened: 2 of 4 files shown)


lfr-0531 (approved):

longcheng-nv (changes requested):
cc @pcastonguay

Before zooming into the diff, I'd like to align on the broader trajectory of this kernel.

## Strategic context

The TRT-LLM top-K kernel is migrating from single-CTA → multi-CTA. The strategic implementation of that step **already landed on `main`** as #12354 — *single-pass multi-CTA cluster top-K in CuTe DSL*, which uses cluster barriers + DSMEM to do the cross-CTA histogram merge in one pass and **reuses the exact same radix micro-kernel** that `topKPerRowDecode` uses internally.

This PR introduces a parallel-but-different multi-CTA strategy on the CUDA C++ side: a "split + global aux + merge" pattern with two kernel launches (the pre-cluster idiom). Because both implementations share the same inner micro-kernel, the gap between them is purely in the launch / scheduling / merge layer, which makes a head-to-head benchmark cleanly attributable.

## Asks before merging

1. **Head-to-head against `CuteDSLTopKDecodeSinglePassMultiCTAClusterRunner` (#12354) on B200.**
   For the (BS × seq_len × cr) grid in (2). Since the micro-kernel is identical, this isolates *"double-launch + global-aux merge"* vs *"cluster + DSMEM in-kernel"*. If #12354 wins on Blackwell, please reposition this PR's hardware scope (e.g., sm_90 / pre-cluster only) and state it explicitly in the description.

2. **Drive the dispatch constants from a systematic sweep, not from a single workload.**
   `kDecodeMinColsPerSubBlock=2048`, `kDecodeTargetTotalBlocks=132`, and `kMaxBlocksPerRowDecode=10` are currently picked ahead of time. Multi-CTA scheduling is tightly coupled with occupancy and L2 footprint; constants tuned from one (concurrency=256, ISL=8192, OSL=1024) trajectory will not generalize. Suggested grid:

   - `BS ∈ {1, 2, 4, 8, 16, 32, 64, 128, 132, 256}` — covers the `smTarget=1` transition at 132
   - `seq_len ∈ {4K, 8K, 16K, 32K, 64K, 128K, 256K}` × `cr ∈ {1, 4}` — crosses both 12288 and 200000 boundaries
   - H100 *and* B200

   Report `µs/call` (before/after) as a heatmap + per-BS curves; pick the constants from the resulting envelope. Also: `kDecodeTargetTotalBlocks=132` is hard-coded for H100; B200 is 148 SM. Consider runtime `cudaDeviceGetAttribute(cudaDevAttrMultiProcessorCount, ...)`.

3. **Isolate baseline / bug-fix / perf in a 3-way table.**
   The current "after" column bundles the bug fix in commit `bf7f808111` together with the perf change. Please report:

   | Variant | GSM8K | kernel µs/call | serving throughput |
   |---|---|---|---|
   | baseline | | | |
   | baseline + commit `bf7f808111` (bug fix only) | | | |
   | baseline + all 4 commits | | | |

   Without this, the +5.13% throughput claim is partially attributable to the correctness fix.

4. **End-to-end coverage at lower concurrency.**
   Commit messages 1–2 motivate the change with "BS=32", but the only serving run is at `concurrency=256`, where decoder steady-state BS is much larger than 32 and the new heuristic does not engage on the hot decode path. Please add a run at `concurrency ∈ {32, 64}` so the regime the PR targets is actually exercised end-to-end.

5. **Split commit `bf7f808111` into a standalone `[None][fix]` PR to `main`.**
   The two fixes (`processHistogramStep<step==1>` reset, `finalIndices = -1` sentinel pairing) are pre-existing bugs in the multi-block merge path. They affect `main` too — large BS / `numColumns ≥ 200000` already routes to the same multi-block radix and inherits both. Landing the fix on `main` should not depend on the DSv4 → main merge timeline.

## Follow-ups for the GVR top-K side (not gating this PR)

- GVR top-K is still a **single-CTA** implementation today. Once it evolves to a multi-CTA variant, the dispatch boundary between GVR and the multi-CTA radix path introduced here will need to be re-derived: rerun the same (BS × seq_len × cr) sweep with both kernels in their multi-CTA form, and pick the GVR ↔ radix crossover from the head-to-head envelope. Until that point, current single-CTA-GVR vs multi-CTA-radix numbers should be treated as an interim baseline only — the dispatch position cannot be considered final.
- `IndexerTopKOp.cpp` now sizes aux buffers dynamically as `{N, blocksPerRow, K}` (range 1–10) instead of fixed `{N, 10, K}`. Please confirm CUDA Graph capture/replay on the GVR warm-up path is unaffected (no warmup-vs-replay shape diff).

longcheng-nv (approved):

PR #13815, opened 2026-05-06
[None][feat] Exact multimodal KV blockhashing

venkywonka:
## Description

Today's multimodal KV-cache block hashing assumes each multimodal item occupies one contiguous `[pos, pos+length)` span in the prompt. Production VLM preprocessors already break that assumption.

image

| Model | Real run shape |
|---|---|
| **Nemotron Nano v2 VL**   | 1 video -> 8 runs x 258 patches, separated by 13-token `Frame N sampled at X.XX seconds:` gaps |
| **Nemotron Nano v2 VL** (EVS-pruned) | 1 video -> 8 runs with lengths `[1026, 2, 2, 2, 2, 2, 2, 2]` |
| **Qwen3-VL** (Video-MME) | 1 video -> 2 runs `[10, 890)` and `[898, 1778)` separated by "`<vision_end><0.2 seconds><vision_start>`" |
| **Nano-Omni v3** | video + audio item, frame separator + appended audio block, reuses v2 VL's preprocessor |

### What goes wrong under the contiguous claim

The producer emits `multimodal_lengths[i] = num_mm_tokens` (a *count* of mm tokens, not a span width). Both consumers then re-derive a span as `[pos, pos + length)`, which is not true for any item whose patches sit in non-contiguous runs:

- **C++ block-key extra keys** (`blockKey.cpp::generateBlockHashExtraKeys`) - `MmKey.startOffset` is computed prompt-relative (`block_start - pos`), which only equals the item-local offset for contiguous items. Tail blocks past `pos + length` receive **no `MmKey` at all**, even when they hold real mm patches (e.g. positions `[2088, 2183)` of an 8-frame Nemotron video).
- **Python V2 token augmentation** (`KVCacheManagerV2._augment_tokens_for_block_reuse`) - overwrites positions inside the claimed span with digest-derived synthetic IDs. Real text tokens that fall in the gaps (`Frame 2 sampled at 2.51 seconds:`) are **erased** before hashing. Tail mm positions outside the span keep their placeholder IDs and carry **no digest contribution at all**.

### Practical impact - correctness debt, not a hot incident

Surface-level cross-prompt collisions are rare in production *today*:

- The prefix hash chain means any digest mismatch breaks the chain at block 0, so two prompts with different videos diverge before the buggy bookkeeping ever matters.
- Gap text is deterministic for a given (video, sampling config), so same-digest requests have identical gap content - the masking happens to coincide with the truth.
- Tail-block placeholder IDs are reserved tokens that don't appear in normal text, so accidental tail-block collisions across different prompts are structurally unlikely.

What the PR actually closes is structural fragility that might surface as preprocessors evolve:

- The V2 augmentation relies on gap-text determinism never slipping. Tokenizer drift, locale formatting changes, or mixed-modality preprocessors (Nano-Omni v3 with audio appended) would silently serve wrong KV.
- Correct tail-block reuse past the truncated span is **accidental** - the hash isn't causally tied to the trailing mm content. EVS-pruned and Nano-Omni-style layouts (where the "tail" *is* most of the runs) need this causation to be enforced.

### What this PR adds

Three optional flat arrays alongside the existing fields, so they cross the Python <-> C++ <-> serialization boundary cleanly:

```
multimodal_item_run_cu_offsets : List[int]   # prefix sum into runs, size: n_item_runs + 1
multimodal_run_positions       : List[int]   # prompt start of each run, flat across items, size: n_item_runs
multimodal_run_lengths         : List[int]   # length of each run, flat across items, size: n_item_runs
```

End-to-end plumbing:

- **Producer** (NEW): `tensorrt_llm/inputs/multimodal.py::_find_mm_token_runs_from_mask` extracts the runs from the per-token mm mask, called from `multimodal_hashing_process` in `inputs/registry.py`.
- **API surface**: `MultimodalInput` Python dataclass gains the three fields plus `_validate_multimodal_runs` (all-or-nothing, monotonic `cu`, `sum(run_lengths(item)) == multimodal_lengths[item]`, runs ordered + non-overlapping, values fit in int32, `pos+len <= INT32_MAX`). C++ `executor::MultimodalInput`, `serialization.cpp`, and the nanobind binding follow; `__setstate__` accepts both legacy 4-tuple and new 7-tuple state for backward-compatible deserialization.
- **Request object**: `GenericLlmRequest` gains 3 optional `shared_ptr<vector<SizeType32>>` fields; `executor_request_to_llm_request` and the nanobind ctor forward them.
- **Two consumers** refactored into `resolve + run-based + legacy-fallback` helpers:
  - C++ `generateBlockHashExtraKeys` -> `resolveRunMetadata` / `generateRunBasedExtraKeys` / `generateLegacyContiguousExtraKeys`.
  - Python `_augment_tokens_for_block_reuse` -> `_resolve_multimodal_run_metadata` / `_augment_tokens_with_mm_run_metadata` / `_augment_tokens_with_contiguous_mm_metadata`. `gen_multimodal_cache_key_tokens` (renamed from `gen_multi_modal_tokens`) gains a `token_offset` arg so multi-run items don't re-materialize the full per-item synthetic sequence.

**Backward compatible**: when run buffers are absent on a request, the legacy contiguous path runs exactly as before. Run buffers are only consulted when the producer emits them.

## Test Coverage

- **C++ unit tests** (`cpp/tests/unit_tests/batch_manager/blockKeyTest.cpp`)
  - `GenerateExtraKeysUsesExactMultimodalRuns` - non-contiguous runs produce per-run `MmKey`s with correct item-local offsets; gap-only blocks produce no key; second run's offset reflects accumulated prior run lengths.
  - `GenerateExtraKeysDisablesHashingForIncompleteRunMetadata` - partial run metadata logs a warning and falls back to no extra keys (request continues).
- **C++ serialization round-trip** (`cpp/tests/unit_tests/executor/serializeUtilsTest.cpp::MultimodalInputWithUuids`) extended to cover the 3 new fields.
- **Python unit tests** (`tests/unittest/_torch/executor/test_kv_cache_v2_multimodal_runs.py`) - exact-runs path, out-of-slice run skipping, incomplete-metadata `ValueError`, contiguous-fallback preserved, `gen_multimodal_cache_key_tokens(token_offset)` semantics.
- **Validator** (`tests/unittest/llmapi/test_llm_kv_cache_events.py::test_multimodal_input_exact_run_buffers`) - `from_components`, paired-fields-required, sum-mismatch, int32 overflow.
- **Bindings** (`tests/unittest/bindings/test_executor_bindings.py`) - `MultimodalInput` ctor with run buffers, state-tuple round-trip with both 4-tuple (legacy) and 7-tuple (new) state.

## Performance and accuracy validation

Rule for all perf tables: positive delta means feature is slower.

### Main perf numbers

Nemotron Omni synthetic run. 4x GB200. `tp_size=4`. `/v1/completions`. No profiler. Each cell has 3 runs x 9 measured requests = 27 measured requests per arm.

| Chunked prefill | Block reuse | Baseline mean | Feature mean | Delta | 95% range |
| --- | --- | ---: | ---: | ---: | ---: |
| off | off | 60.16 ms | 60.03 ms | -0.13 ms / -0.21% | -0.39% to -0.04% |
| off | on | 60.39 ms | 61.49 ms | +1.10 ms / +1.83% | -0.42% to +4.08% |
| on | off | 61.17 ms | 59.52 ms | -1.64 ms / -2.65% | -9.69% to +4.39% |
| on | on | 60.23 ms | 60.28 ms | +0.05 ms / +0.08% | -4.88% to +5.05% |

Takeaway: no clear perf regression. Only positive deltas are `+1.83%` and `+0.08%`, and both ranges cross zero.

### Extra perf checks

| Check | Baseline | Feature | Result |
| --- | ---: | ---: | --- |
| Qwen2-VL image, C++ KV, block reuse on, warm median | 0.203896 s | 0.202823 s | feature/baseline `0.994738` |
| Qwen2-VL Video-MME, C++ KV, block reuse off, warm median | 1.134069 s | 1.143502 s | feature/baseline `1.008318` |
| Qwen2-VL Video-MME, C++ KV, block reuse on, warm median | 0.943674 s | 0.951920 s | feature/baseline `1.008738` |
| Qwen2-VL Video-MME, Python KVCacheManagerV2, block reuse off, warm median | 1.106186 s | 1.118415 s | feature/baseline `1.011055` |
| Qwen2-VL Video-MME, Python KVCacheManagerV2, block reuse on, warm median | 0.918805 s | 0.930309 s | feature/baseline `1.012520` |
| Qwen2.5-VL-7B Video-MME, block reuse off, direct generate time | 0.5644 s | 0.5706 s | feature/baseline `1.0110` |
| Qwen2.5-VL-7B Video-MME, block reuse on, direct generate time | 0.5789 s | 0.5557 s | feature/baseline `0.9599` |

Worst warm repeated ratio observed: `1.012520` (+1.25%).

### Accuracy numbers

| Check | Baseline | Feature | Result |
| --- | --- | --- | --- |
| Qwen2-VL MMMU image sample | accuracy `40.0`, MM keys `654`, stored blocks `741` | accuracy `40.0`, MM keys `654`, stored blocks `741` | pass: output text and input IDs matched |
| Qwen2-VL Video-MME, C++ KV, block reuse off | answer `A`, token IDs `[32, 151645]` | answer `A`, token IDs `[32, 151645]` | pass: gold was `A` |
| Qwen2-VL Video-MME, C++ KV, block reuse on | answer `A`, token IDs `[32, 151645]` | answer `A`, token IDs `[32, 151645]` | pass: gold was `A` |
| Qwen2-VL Video-MME, Python KVCacheManagerV2, block reuse off | answer `C` | answer `C` | parity only: gold was `A` |
| Qwen2-VL Video-MME, Python KVCacheManagerV2, block reuse on | answer `C` | answer `C` | parity only: gold was `A` |
| Qwen2.5-VL-7B Video-MME, block reuse off | answer `A`, token IDs `[32, 151645]` | answer `A`, token IDs `[32, 151645]` | parity only: gold was `D` |
| Qwen2.5-VL-7B Video-MME, block reuse on | answer `A`, token IDs `[32, 151645]` | answer `A`, token IDs `[32, 151645]` | parity only: gold was `D` |

Claim boundary: perf numbers above show no clear regression. Accuracy numbers above show one passing local Video-MME sample plus branch-vs-baseline parity in the other completed video lanes. No scorer-backed Nemotron Omni Video-MME accuracy number yet.

## PR Checklist

- [x] PR follows TRT-LLM CODING GUIDELINES.
- [x] Test cases provided for new code paths.
- [x] No new third-party dependencies.
- [x] CODEOWNERS updated if ownership changes - N/A.
- [x] Documentation updated as needed - pending review (no public-API docs reference the contiguous-span assumption).
- [x] Update tava architecture diagram - N/A (no significant design change to the KV-cache layer).
- [x] The reviewers assigned automatically/manually are appropriate for the PR.

### DIFF
--- a/cpp/include/tensorrt_llm/batch_manager/llmRequest.h
+++ b/cpp/include/tensorrt_llm/batch_manager/llmRequest.h
@@ -149,7 +149,7 @@ public:
         std::optional<executor::ContextPhaseParams> const& contextPhaseParams = std::nullopt,
         std::optional<CacheSaltIDType> cacheSaltID = std::nullopt, std::optional<TimePoint> arrivalTime = std::nullopt,
         std::optional<std::vector<std::tuple<std::string, int>>> agent_hierarchy = std::nullopt,
-        std::optional<std::shared_ptr<std::vector<SizeType32>>> multimodalItemRunCuSeqlen = std::nullopt,
+        std::optional<std::shared_ptr<std::vector<SizeType32>>> multimodalItemRunCuOffsets = std::nullopt,
         std::optional<std::shared_ptr<std::vector<SizeType32>>> multimodalRunPositions = std::nullopt,
         std::optional<std::shared_ptr<std::vector<SizeType32>>> multimodalRunLengths = std::nullopt)
         : mRequestId(requestId)
@@ -175,7 +175,7 @@ public:
         , mMultimodalPositions(std::move(multimodalPositions))
         , mMultimodalLengths(std::move(multimodalLengths))
         , mMultimodalUuids(std::move(multimodalUuids))
-        , mMultimodalItemRunCuSeqlen(std::move(multimodalItemRunCuSeqlen))
+        , mMultimodalItemRunCuOffsets(std::move(multimodalItemRunCuOffsets))
         , mMultimodalRunPositions(std::move(multimodalRunPositions))
         , mMultimodalRunLengths(std::move(multimodalRunLengths))
         , mMultimodalEmbedding(std::move(multimodalEmbedding))
@@ -416,10 +416,10 @@ public:
             {
                 mMultimodalUuids = std::make_shared<std::vector<std::optional<std::string>>>(multimodalUuids.value());
             }
-            if (auto const& multimodalItemRunCuSeqlen = multimodalInput->getMultimodalItemRunCuSeqlen())
+            if (auto const& multimodalItemRunCuOffsets = multimodalInput->getMultimodalItemRunCuOffsets())
             {
-                mMultimodalItemRunCuSeqlen
-                    = std::make_shared<std::vector<SizeType32>>(multimodalItemRunCuSeqlen.value());
+                mMultimodalItemRunCuOffsets
+                    = std::make_shared<std::vector<SizeType32>>(multimodalItemRunCuOffsets.value());
             }
             if (auto const& multimodalRunPositions = multimodalInput->getMultimodalRunPositions())
             {
@@ -960,9 +960,9 @@ public:
         return mMultimodalUuids;
     }
 
-    [[nodiscard]] std::optional<std::shared_ptr<std::vector<SizeType32>>> getMultimodalItemRunCuSeqlen() const
+    [[nodiscard]] std::optional<std::shared_ptr<std::vector<SizeType32>>> getMultimodalItemRunCuOffsets() const
     {
-        return mMultimodalItemRunCuSeqlen;
+        return mMultimodalItemRunCuOffsets;
     }
 
     [[nodiscard]] std::optional<std::shared_ptr<std::vector<SizeType32>>> getMultimodalRunPositions() const
@@ -2075,7 +2075,7 @@ protected:
     std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalPositions{std::nullopt};
     std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalLengths{std::nullopt};
     std::optional<std::shared_ptr<std::vector<std::optional<std::string>>>> mMultimodalUuids{std::nullopt};
-    std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalItemRunCuSeqlen{std::nullopt};
+    std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalItemRunCuOffsets{std::nullopt};
     std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalRunPositions{std::nullopt};
     std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalRunLengths{std::nullopt};
     std::optional<TensorPtr> mMultimodalEmbedding{std::nullopt};
@@ -2396,7 +2396,7 @@ public:
         std::optional<executor::ContextPhaseParams> const& contextPhaseParams = std::nullopt,
         std::optional<CacheSaltIDType> cacheSaltID = std::nullopt, std::optional<TimePoint> arrivalTime = std::nullopt,
         std::optional<std::vector<std::tuple<std::string, int>>> agent_hierarchy = std::nullopt,
-        std::optional<std::vector<SizeType32>> multimodalItemRunCuSeqlen = std::nullopt,
+        std::optional<std::vector<SizeType32>> multimodalItemRunCuOffsets = std::nullopt,
         std::optional<std::vector<SizeType32>> multimodalRunPositions = std::nullopt,
         std::optional<std::vector<SizeType32>> multimodalRunLengths = std::nullopt)
         : Base(requestId, maxNewTokens, std::make_shared<std::vector<TokenIdType>>(std::move(inputTokens)),
@@ -2433,8 +2433,8 @@ public:
             numReturnSequences, std::move(eagleConfig), skipCrossAttnBlocks, returnPerfMetrics,
             std::move(guidedDecodingParams), languageAdapterUid, allottedTimeMs, contextPhaseParams, cacheSaltID,
             arrivalTime, std::move(agent_hierarchy),
-            multimodalItemRunCuSeqlen.has_value()
-                ? std::make_shared<std::vector<SizeType32>>(std::move(multimodalItemRunCuSeqlen.value()))
+            multimodalItemRunCuOffsets.has_value()
+                ? std::make_shared<std::vector<SizeType32>>(std::move(multimodalItemRunCuOffsets.value()))
                 : std::optional<std::shared_ptr<std::vector<SizeType32>>>(std::nullopt),
             multimodalRunPositions.has_value()
                 ? std::make_shared<std::vector<SizeType32>>(std::move(multimodalRunPositions.value()))
--- a/cpp/include/tensorrt_llm/batch_manager/llmRequest.h
+++ b/cpp/include/tensorrt_llm/batch_manager/llmRequest.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2025, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2022-2026, NVIDIA CORPORATION.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -148,7 +148,10 @@ public:
         std::optional<MillisecondsType> allottedTimeMs = std::nullopt,
         std::optional<executor::ContextPhaseParams> const& contextPhaseParams = std::nullopt,
         std::optional<CacheSaltIDType> cacheSaltID = std::nullopt, std::optional<TimePoint> arrivalTime = std::nullopt,
-        std::optional<std::vector<std::tuple<std::string, int>>> agent_hierarchy = std::nullopt)
+        std::optional<std::vector<std::tuple<std::string, int>>> agent_hierarchy = std::nullopt,
+        std::optional<std::shared_ptr<std::vector<SizeType32>>> multimodalItemRunCuSeqlen = std::nullopt,
+        std::optional<std::shared_ptr<std::vector<SizeType32>>> multimodalRunPositions = std::nullopt,
+        std::optional<std::shared_ptr<std::vector<SizeType32>>> multimodalRunLengths = std::nullopt)
         : mRequestId(requestId)
         , mPromptLen(inputTokens->size())
         , mMaxNewTokens(maxNewTokens)
@@ -172,6 +175,9 @@ public:
         , mMultimodalPositions(std::move(multimodalPositions))
         , mMultimodalLengths(std::move(multimodalLengths))
         , mMultimodalUuids(std::move(multimodalUuids))
+        , mMultimodalItemRunCuSeqlen(std::move(multimodalItemRunCuSeqlen))
+        , mMultimodalRunPositions(std::move(multimodalRunPositions))
+        , mMultimodalRunLengths(std::move(multimodalRunLengths))
         , mMultimodalEmbedding(std::move(multimodalEmbedding))
         , mMropeRotaryCosSin(std::move(mropeRotaryCosSin))
         , mMropePositionDeltas(mropePositionDeltas)
@@ -397,6 +403,32 @@ public:
         {
             mMropeRotaryCosSin = executor::detail::toITensor(mRopeConfig.value().getMRopeRotaryCosSin());
             mMropePositionDeltas = mRopeConfig.value().getMRopePositionDeltas();
+        }
+
+        auto multimodalInput = req.getMultimodalInput();
+        if (multimodalInput)
+        {
+            mMultimodalHashes
+                = std::make_shared<std::vector<std::vector<SizeType32>>>(multimodalInput->getMultimodalHashes());
+            mMultimodalPositions = std::make_shared<std::vector<SizeType32>>(multimodalInput->getMultimodalPositions());
+            mMultimodalLengths = std::make_shared<std::vector<SizeType32>>(multimodalInput->getMultimodalLengths());
+            if (auto const& multimodalUuids = multimodalInput->getMultimodalUuids())
+            {
+                mMultimodalUuids = std::make_shared<std::vector<std::optional<std::string>>>(multimodalUuids.value());
+            }
+            if (auto const& multimodalItemRunCuSeqlen = multimodalInput->getMultimodalItemRunCuSeqlen())
+            {
+                mMultimodalItemRunCuSeqlen
+                    = std::make_shared<std::vector<SizeType32>>(multimodalItemRunCuSeqlen.value());
+            }
+            if (auto const& multimodalRunPositions = multimodalInput->getMultimodalRunPositions())
+            {
+                mMultimodalRunPositions = std::make_shared<std::vector<SizeType32>>(multimodalRunPositions.value());
+            }
+            if (auto const& multimodalRunLengths = multimodalInput->getMultimodalRunLengths())
+            {
+                mMultimodalRunLengths = std::make_shared<std::vector<SizeType32>>(multimodalRunLengths.value());
+            }
         }
 
         auto loraConfig = req.getLoraConfig();
@@ -926,6 +958,21 @@ public:
     [[nodiscard]] std::optional<std::shared_ptr<std::vector<std::optional<std::string>>>> getMultimodalUuids() const
     {
         return mMultimodalUuids;
+    }
+
+    [[nodiscard]] std::optional<std::shared_ptr<std::vector<SizeType32>>> getMultimodalItemRunCuSeqlen() const
+    {
+        return mMultimodalItemRunCuSeqlen;
+    }
+
+    [[nodiscard]] std::optional<std::shared_ptr<std::vector<SizeType32>>> getMultimodalRunPositions() const
+    {
+        return mMultimodalRunPositions;
+    }
+
+    [[nodiscard]] std::optional<std::shared_ptr<std::vector<SizeType32>>> getMultimodalRunLengths() const
+    {
+        return mMultimodalRunLengths;
     }
 
     [[nodiscard]] std::optional<TensorPtr> getMultimodalEmbedding() const
@@ -2028,6 +2075,9 @@ protected:
     std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalPositions{std::nullopt};
     std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalLengths{std::nullopt};
     std::optional<std::shared_ptr<std::vector<std::optional<std::string>>>> mMultimodalUuids{std::nullopt};
+    std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalItemRunCuSeqlen{std::nullopt};
+    std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalRunPositions{std::nullopt};
+    std::optional<std::shared_ptr<std::vector<SizeType32>>> mMultimodalRunLengths{std::nullopt};
     std::optional<TensorPtr> mMultimodalEmbedding{std::nullopt};
     std::optional<TensorPtr> mMropeRotaryCosSin{std::nullopt};
     std::optional<SizeType32> mMropePositionDeltas{std::nullopt};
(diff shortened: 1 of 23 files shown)


cpp/include/tensorrt_llm/batch_manager/llmRequest.h
```diff
@@ -146,7 +146,10 @@ class GenericLlmRequest
         std::optional<SizeType32> languageAdapterUid = std::nullopt,
         std::optional<MillisecondsType> allottedTimeMs = std::nullopt,
         std::optional<executor::ContextPhaseParams> const& contextPhaseParams = std::nullopt,
-        std::optional<CacheSaltIDType> cacheSaltID = std::nullopt, std::optional<TimePoint> arrivalTime = std::nullopt)
+        std::optional<CacheSaltIDType> cacheSaltID = std::nullopt, std::optional<TimePoint> arrivalTime = std::nullopt,
+        std::optional<std::shared_ptr<std::vector<SizeType32>>> multimodalItemRunCuOffsets = std::nullopt,
```
2ez4bz, line 152: For my own knowledge / out of curiosity, what does `Cu` refer to here?

cpp/tensorrt_llm/batch_manager/blockKey.cpp
```diff
@@ -17,82 +17,242 @@
...
 {
-    auto const multimodalHashes = llmRequest.getMultimodalHashes();
-    auto const multimodalPositions = llmRequest.getMultimodalPositions();
-    auto const multimodalLengths = llmRequest.getMultimodalLengths();
-    auto const multimodalUuids = llmRequest.getMultimodalUuids();
+    if (multimodalUuids && *multimodalUuids && itemIdx < (*multimodalUuids)->size())
+    {
+        return (*(*multimodalUuids))[itemIdx];
+    }
+    return std::nullopt;
```
2ez4bz, line 88: What does `std::nullopt` signify? Maybe leave a comment?

cpp/tensorrt_llm/batch_manager/blockKey.cpp
```diff
@@ -17,82 +17,242 @@
...
+        return {RunMetadataStatus::kInvalid, {}};
     }
 
-    std::vector<MmKey> extraKeys;
-    extraKeys.reserve((*multimodalPositions)->size());
-    std::array<uint8_t, 32> mmHashArray;
+    auto const& itemRunCuOffsets = *(*multimodalItemRunCuOffsets);
+    auto const& runPositions = *(*multimodalRunPositions);
+    auto const& runLengths = *(*multimodalRunLengths);
+    if (itemRunCuOffsets.size() != itemCount + 1 || itemRunCuOffsets.empty() || itemRunCuOffsets.front() != 0
```
2ez4bz: Maybe stupid question - why is the expected count `itemCount + 1`?

Also, is `itemRunCuOffsets.empty()` redundant? I believe `itemCount` is `size_t` which should be unsigned, so if it is indeed empty, `itemRunCuOffsets.size() != itemCount + 1` should automatically evaluate to `true`? Of course, it does have the benefit of being more explicit.

cpp/tensorrt_llm/batch_manager/blockKey.cpp
```diff
@@ -17,82 +17,242 @@
...
+        || itemRunCuOffsets.back() < 0 || static_cast<size_t>(itemRunCuOffsets.back()) != runPositions.size()
+        || runPositions.size() != runLengths.size())
+    {
+        TLLM_LOG_WARNING("Invalid multimodal run metadata shape; disabling multimodal cache hash extra keys");
```
2ez4bz: I'm not sure how convenient it is to do in C++, but there are several conditions being checked for validity here. From a UX / debugging perspective, having the values / conditions logged would help.

Similar comment for other places where we log such warnings.

cpp/tensorrt_llm/batch_manager/blockKey.cpp
```diff
@@ -17,82 +17,242 @@
...
     return extraKeys;
 }
+} // namespace
+
+namespace tensorrt_llm::batch_manager::kv_cache_manager
+{
+std::vector<MmKey> generateBlockHashExtraKeys(
+    tensorrt_llm::batch_manager::LlmRequest const& llmRequest, SizeType32 startTokenIdx, SizeType32 endTokenIdx)
+{
+    auto const multimodalHashes = llmRequest.getMultimodalHashes();
```
2ez4bz, line 288: I know this pattern precedes your changes, but I wanted to check: for container data types, would it make sense to have all of these be `const` references? Or are the return types already references?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -90,6 +90,137 @@ class Role:
...
+    # Convert 8 x int32 hash to 32-byte digest (big-endian, matching v1 C++
+    # getNthByte which extracts MSB first).
+    assert len(
+        hash_ints) == 8, f"Expected 8 int32 hash values, got {len(hash_ints)}"
+    return b''.join(v.to_bytes(4, 'big', signed=True) for v in hash_ints)
+
+
+def _as_int64_cpu_tensor(values: Sequence[int] | torch.Tensor) -> torch.Tensor:
+    # int64 is expected dtype for tensor indexing
+    return torch.as_tensor(values, dtype=torch.int64).cpu()
```
2ez4bz: Just checking, since somebody had made a similar comment before: do we ever expect `values` to be a GPU tensor? If so, wouldn't that introduce a D2H sync once per request lifetime, or once per forward pass of the model?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -90,6 +90,137 @@ class Role:
...
+
+
+def _resolve_multimodal_run_metadata(
+        req: LlmRequest) -> Optional[_MmRunMetadata]:
+    # NOTE: the following optional fields help in exact mm token position mixing in blockhash key generation
+    item_run_cu_offsets = getattr(req, "multimodal_item_run_cu_offsets", None)
```
2ez4bz: Just checking: are the `getattr` calls necessary? Shouldn't `LlmRequest` be guaranteed to have these new fields now that you've added them? If so, then let's remove them, as it makes the code more readable.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -90,6 +90,137 @@ class Role:
...
+    run_ends: torch.Tensor
+    run_item_indices: torch.Tensor
+    run_item_offsets: torch.Tensor
+
+
+def _hash_to_digest(hash_ints: Sequence[int]) -> bytes:
+    # Convert 8 x int32 hash to 32-byte digest (big-endian, matching v1 C++
+    # getNthByte which extracts MSB first).
+    assert len(
+        hash_ints) == 8, f"Expected 8 int32 hash values, got {len(hash_ints)}"
```
2ez4bz: Is this the same assert as the one in line 66 of `cpp/tensorrt_llm/batch_manager/blockKey.cpp` ? Do we need both?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -90,6 +90,137 @@ class Role:
...
+                                         metadata: _MmRunMetadata, start: int,
+                                         end: int) -> list[TokenIdExt]:
+    overlap_mask = (metadata.run_ends > start) & (metadata.run_positions < end)
+    overlap_run_indices = torch.nonzero(overlap_mask).flatten()
+    if overlap_run_indices.numel() == 0:
+        return result
+
+    overlap_run_item_indices = metadata.run_item_indices[overlap_run_indices]
+    overlap_item_indices = torch.unique_consecutive(overlap_run_item_indices)
+    for item_idx in overlap_item_indices.tolist():
```
2ez4bz: Just checking: do we need the `tolist()` here?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -90,6 +90,137 @@ class Role:
...
+        hash_ints) == 8, f"Expected 8 int32 hash values, got {len(hash_ints)}"
+    return b''.join(v.to_bytes(4, 'big', signed=True) for v in hash_ints)
+
+
+def _as_int64_cpu_tensor(values: Sequence[int] | torch.Tensor) -> torch.Tensor:
+    # int64 is expected dtype for tensor indexing
+    return torch.as_tensor(values, dtype=torch.int64).cpu()
+
+
+def _resolve_multimodal_run_metadata(
```
2ez4bz, line 144: Nit: some code comments for _why_ certain steps are being done would help readers a lot in understanding what is going on 🙏

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -90,6 +90,137 @@ class Role:
...
+    # we assume multimodal chunks are contiguous
+    # so mm_token_end_position = mm_token_start_position + mm_token_length
+    # this assumption is not valid for video data, which may contain interleaved text tokens
+    positions = _as_int64_cpu_tensor(multimodal_positions)
+    lengths = _as_int64_cpu_tensor(multimodal_lengths)
+
+    item_ends = positions + lengths
+    overlap_mask = (item_ends > start) & (positions < end)
+    overlap_item_indices = torch.nonzero(overlap_mask).flatten()
+    for item_idx in overlap_item_indices.tolist():
```
2ez4bz: Similar question here re `tolist()`.

tensorrt_llm/inputs/multimodal.py
```diff
@@ -86,6 +87,20 @@ class MultimodalInput:
     returned in KV cache events.
     """
 
+    multimodal_item_run_cu_offsets: Optional[List[int]] = None
+    """Optional offsets into the flat multimodal run arrays.
+
+    Length is ``len(multimodal_hashes) + 1``. Runs for item ``i`` live in
```
2ez4bz: Nit: `sed -i 's/``/`/g'`.

tests/unittest/llmapi/test_llm_kv_cache_events.py
```diff
@@ -409,6 +409,78 @@ def test_multimodal_input_from_components_with_uuids():
     assert mm_input_no_uuids.multimodal_uuids is None
 
 
+def test_multimodal_input_exact_run_buffers():
+    """Test optional exact run buffers on Python MultimodalInput."""
+    import torch
```
2ez4bz: Nit: move imports to module-level.

tests/unittest/llmapi/test_llm_kv_cache_events.py
```diff
@@ -409,6 +409,78 @@ def test_multimodal_input_from_components_with_uuids():
...
+            multimodal_positions=[1],
+            multimodal_lengths=[2],
+            multimodal_item_run_cu_offsets=[0, 1],
+            multimodal_run_positions=[1],
+            multimodal_run_lengths=[3],
+        )
+
+
+def test_multimodal_input_rejects_exact_run_int32_overflow():
+    from tensorrt_llm.inputs.multimodal import MultimodalInput
```
2ez4bz: Ditto.

cpp/include/tensorrt_llm/batch_manager/llmRequest.h
venkywonka: "Cumulative"

cpp/tensorrt_llm/batch_manager/blockKey.cpp
venkywonka: means we don't layer UUIDs in the key hashing (so just content-hashing)

cpp/tensorrt_llm/batch_manager/blockKey.cpp
venkywonka: i think they are kinda references in the sense that they return shared ptrs to vectors - so even though the getter functions don't don't return const references explicitly as `const &`, the values they return are not vectors themselves but `std::optional<std::shared_ptr<...>>` ptrs - so this isnt' causing a full vector copy or anything

tensorrt_llm/_torch/pyexecutor/resource_manager.py
venkywonka: yea we do need both as theys are mutually exclusive kv cache managers

tensorrt_llm/_torch/pyexecutor/resource_manager.py
venkywonka: its expected to reside in cpu - better to raise loudly than defensively .cpu() it - will add that

tensorrt_llm/_torch/pyexecutor/resource_manager.py
venkywonka: `overlap_item_indices` is a 1d cpu torch tensor, so if i don't `.tolist()` it, `item_idx` will not be a scalar, `int(item_idx)` will. 
maybe doing that is better than materializing a list out of `overlap_item_indices`... decent catch

cpp/tensorrt_llm/batch_manager/blockKey.cpp
venkywonka: have added some helpful comments

cpp/tensorrt_llm/batch_manager/blockKey.cpp
venkywonka: have added helpful comments.

cpp/tensorrt_llm/batch_manager/blockKey.cpp
venkywonka: left a comment.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
venkywonka: done

tensorrt_llm/inputs/multimodal.py
venkywonka: done

SimengLiu-nv (commented):
When any exact-run field is present but the run metadata is incomplete or malformed, this returns an empty extra-key vector. That does not disable KV reuse; it builds ordinary text/placeholder-only BlockKeys, so two requests with the same prompt placeholder IDs but different multimodal content can collide and reuse the wrong KV. Since the legacy multimodal hashes/positions/lengths are still available, this path should either fall back to the legacy extra-key generation or make the request uncacheable instead of returning reusable empty keys.


Assume two requests differ only by image content.

### Setup

- `tokens_per_block = 4`
- Prompt tokens for both requests: `[101, IMG, IMG, 102, 103]`
- `IMG` is the same placeholder token ID in both requests.
- Request A image hash: `hashA`
- Request B image hash: `hashB`
- Legacy metadata is present:
  - `multimodal_hashes = [hashA/hashB]`
  - `multimodal_positions = [1]`
  - `multimodal_lengths = [2]`
- Exact-run metadata is malformed:
  - `multimodal_item_run_cu_offsets = [0, 1]`
  - `multimodal_run_positions = null`
  - `multimodal_run_lengths = null`

### What Happens Today

1. `generateBlockHashExtraKeys()` sees at least one exact-run field, so it calls `resolveRunMetadata()`.
2. `resolveRunMetadata()` returns `kInvalid` because the three run fields are not all present.
3. The current code returns `{}` from `blockKey.cpp`.
4. `buildBlockKeys()` still creates reusable `BlockKey`s with `extraKeys = {}`.
5. For block `[101, IMG, IMG, 102]`, Request A and Request B now have identical block keys:
   - same token IDs: `[101, IMG, IMG, 102]`
   - same `loraTaskId`
   - same `cacheSaltID`
   - same `extraKeys = {}`
6. The KV cache can reuse Request A’s cached block for Request B, even though `hashA != hashB`.

### Why This Is Wrong

`extraKeys` is the only thing that makes identical multimodal placeholder tokens distinguish different media content. Returning an empty vector does not mean “disable reuse”; it means “hash this like ordinary text.” For multimodal placeholders, that is unsafe.

### Safe Behavior

Preferred behavior:

- If exact-run metadata is invalid but legacy `multimodal_positions`, `multimodal_lengths`, and `multimodal_hashes` are valid, fall back to `generateLegacyContiguousExtraKeys(...)`.

Stricter behavior:

- Mark the request or block as uncacheable instead of producing normal text-only block keys.

Unsafe behavior:

- Returning `{}` while still committing/reusing the block.

tensorrt_llm/inputs/multimodal.py
```diff
@@ -16,6 +16,7 @@
 
 # Default hasher
 default_hasher = blake3
+_INT32_MAX = 2_147_483_647
```
SimengLiu-nv: Would be better to replace the exact value with function calls or `2**31 - 1`.

SimengLiu-nv (approved):
LGTM.

cpp/tensorrt_llm/batch_manager/blockKey.cpp
2ez4bz: Cool, thanks for explaining.

moraxu (approved):
Thanks for the comments in `resource_manager.py`

tests/unittest/_torch/executor/test_kv_cache_v2_multimodal_runs.py
```diff
@@ -0,0 +1,120 @@
...
+        vocab_size + 4,
+    ]
+
+
+def test_augment_tokens_for_block_reuse_uses_exact_multimodal_runs():
+    vocab_size = 1000
+    digest = b"".join(v.to_bytes(4, "big", signed=True) for v in _HASH_INTS)
+    mm_tokens = gen_multimodal_cache_key_tokens(vocab_size, digest, 4)
+
+    manager = SimpleNamespace(vocab_size=vocab_size)
```
2ez4bz: Could the original classes have been used themselves?

tests/unittest/_torch/executor/test_kv_cache_v2_multimodal_runs.py
```diff
@@ -0,0 +1,120 @@
...
+import pytest
+import torch
+
+import tensorrt_llm._torch.pyexecutor.resource_manager as resource_manager
+from tensorrt_llm._torch.pyexecutor.resource_manager import KVCacheManagerV2
+from tensorrt_llm.runtime.kv_cache_manager_v2._block_radix_tree import (
+    gen_multimodal_cache_key_tokens,
+)
+
+_HASH_INTS = [1, 2, 3, 4, 5, 6, 7, 8]
```
2ez4bz: Nit: make a tuple?

2ez4bz (approved):

QiJune (approved):
LGTM

PR #13821, opened 2026-05-07
[TRTLLM-12342][feat] Ring Attention, Unified Context Parallel for VisualGen

NVShreyas:
WAN 2.2 14B perf on N GB200s - 

| GPUs | Config          | 40-step (s) | Speedup vs 1-GPU | Efficiency |
|-----:|:----------------|------------:|-----------------:|-----------:|
| 1    | cfg1            | 348.40      | 1.00x            | 100%       |
| 2    | cfg2            | 174.00      | 2.00x            | 100%       |
| 4    | cfg2·ul2        | 95.89       | 3.63x            | 91%        | |
| 8    | cfg2·ul4        | 48.06       | 7.25x            | 91%        |
| 16   | cfg2·ul4·r2     | 25.60       | 13.61x           | 85%        |
| 24   | cfg2·ul4·r3     | 18.46       | 18.87x           | 79%        ||
| 32   | cfg2·ul8·r2     | 14.12       | 24.67x           | 77%        |
| 64   | cfg2·ul8·r4     | 8.48        | 41.08x           | 64%        |

## Summary by CodeRabbit

* **New Features**
  * Ring-based sequence parallelism for distributed inference and generation
  * New `--ring_size` CLI argument to configure ring parallelism

* **Refactor**
  * Unified sequence sharding abstraction for simplified parallel model configuration and execution

* **Tests**
  * Added comprehensive multi-GPU tests for ring parallelism and combined ring-Ulysses configurations

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/docs/source/models/visual-generation.md
+++ b/docs/source/models/visual-generation.md
@@ -38,7 +38,7 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models 
 ### Feature Matrix
 
 | Model | FP8 blockwise | NVFP4 | TeaCache | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve | Attention2D | Ring Attention |
-|---|---|---|---|---|---|---|---|---|---|
+|---|---|---|---|---|---|---|---|---|---|--|--|
 | **FLUX.1** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes |
 | **FLUX.2** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes |
 | **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
@@ -118,7 +118,7 @@ TeaCache caches transformer outputs when timestep embeddings change slowly betwe
 - **CFG Parallelism** (`--cfg_size 2`): Splits positive/negative guidance prompts across GPUs.
 - **Ulysses Parallelism** (`--ulysses_size N`): Splits the sequence dimension across GPUs for longer sequences.
 - **Parallel VAE** (`--parallel_vae_size N`): Shards the final VAE decode along a spatial axis across GPUs, useful to reduce VAE latency and improve GPU utilization (Constraint: `parallel_vae_size ≤ world_size`). Currently only supported for WAN models.
-- **Attention Parallel**: There are 2 methods supported to run attention parallel - 
+- **Attention Parallel**: There are 2 methods supported to run attention parallel. Both of these methods require the attention backend to support LSE (only FA4 currently) - 
     - **Attention2D Parallelism** (`--attn2d_row_size N`, `--attn2d_col_size M`): Shards the sequence axis across a 2D `N x M` device mesh, all-gathering Q along rows and K/V along columns so each rank computes a sub-block of the attention matrix (total CP degree = `N * M`; not currently combinable with Ulysses).
     - **Ring Attention Parallelism** (`--ring_size N`): Shards the sequence axis across a 1D ring of `N` ranks and streams K/V blocks around the ring so each rank computes its attention output without materializing the full K/V (mutually exclusive with Attention2D).
 ## Developer Guide
--- a/docs/source/models/visual-generation.md
+++ b/docs/source/models/visual-generation.md
@@ -37,13 +37,13 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models 
 
 ### Feature Matrix
 
-| Model | FP8 blockwise | NVFP4 | TeaCache | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve |
+| Model | FP8 blockwise | NVFP4 | TeaCache | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve | Attention2D | Ring Attention |
 |---|---|---|---|---|---|---|---|---|---|
-| **FLUX.1** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes |
-| **FLUX.2** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes |
-| **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| **Wan 2.2** | Yes | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes |
-| **LTX-2** | Yes | Yes | No | Yes | Yes | No | No | Yes | Yes |
+| **FLUX.1** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes |
+| **FLUX.2** | Yes | Yes | Yes | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes |
+| **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| **Wan 2.2** | Yes | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| **LTX-2** | Yes | Yes | No | Yes | Yes | No | No | Yes | Yes | Yes | Yes |
 
 [^1]: FLUX models use embedded guidance and do not have a separate negative prompt path, so CFG parallelism is not applicable.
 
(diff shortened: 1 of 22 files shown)


tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -186,21 +185,14 @@ class ParallelConfig(StrictBaseModel):
     parallel_vae_split_dim: Literal["width", "height"] = "width"
 
     # DiT Parallelism
-    dit_dp_size: int = PydanticField(1, ge=1)
+    dit_dp_size: int = PydanticField(1, ge=1)  # Not yet supported
     dit_tp_size: int = PydanticField(1, ge=1)  # Not yet supported
     dit_ulysses_size: int = PydanticField(1, ge=1)  # Supported
-    dit_ring_size: int = PydanticField(1, ge=1)  # Not yet supported
+    dit_ring_size: int = PydanticField(1, ge=1)  # Supported
```
chang-l, line 191: Since this PR now supports composing ring and Ulysses, `ParallelConfig.seq_parallel_size` below still returns only `dit_ring_size` when ring is enabled. The public `n_workers`/`total_parallel_size` path will launch 2 workers for `ring_size=2, ulysses_size=2`, then `VisualGenMapping` expects 4 and rejects the config. Please make the public parallel size account for `dit_ring_size * dit_ulysses_size` and add coverage through `VisualGenArgs`/`ParallelConfig`.

tensorrt_llm/_torch/visual_gen/utils.py
```diff
@@ -33,3 +41,214 @@ def postprocess_video_tensor(video: torch.Tensor) -> torch.Tensor:
...
+        Validates head divisibility only when Ulysses is part of the seq
+        group — ring and attn2d shard the sequence axis and have no
+        head-count constraint.
+        """
+        if vgm is None:
+            return cls(size=1, rank=0, group=None)
+
+        size = vgm.seq_size
+        rank = vgm.seq_rank
+        group = vgm.seq_group
```
chang-l: `seq_group` is defined as a method on `VisualGenMapping`, but this reads it as a property. Active sharding stores a bound method in `_group`, so `dist.all_gather(..., group=self._group)` later receives the method object instead of a `ProcessGroup`. Either call `vgm.seq_group()` here or make `seq_group` a `@property`.

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -1211,11 +1173,11 @@ def configure_audio_ulysses(self, audio_seq_len: int) -> None:
         known.  The decision is cached — ``forward()`` uses it without
         re-checking.
         """
-        if not self.use_seq_parallel:
+        if not self._sharder.is_active:
             self._audio_is_sharded = False
             return
 
-        self._audio_is_sharded = audio_seq_len % self.seq_parallel_size == 0
+        self._audio_is_sharded = audio_seq_len % self._sharder.size == 0
```
chang-l: `_sharder.size` is now ring/CP x Ulysses, not just Ulysses. When audio length is not divisible, this leaves audio unsharded while `audio_attn1` may still be a Ring/Attention2D wrapper; only the Ulysses wrapper has the plain-attention toggle. Please require divisibility for non-Ulysses CP or add a real plain-attention fallback for those modes.

tensorrt_llm/_torch/visual_gen/utils.py
```diff
@@ -33,3 +41,214 @@ def postprocess_video_tensor(video: torch.Tensor) -> torch.Tensor:
...
+        RoPE layouts that occur across visual_gen models.  Returns the
+        pair unchanged when the seq axis cannot be inferred unambiguously
+        (e.g. ``seq_len == head_dim``); callers can fall back to explicit
+        ``shard(..., dim=...)`` calls in that case.
+        """
+        if rope is None or not self.is_active:
+            return rope
+
+        cos, sin = rope
+        candidates = [d for d, n in enumerate(cos.shape) if n == seq_len]
```
chang-l: Inferring the RoPE sequence axis this way can leave it unsharded when another dimension equals `seq_len`. LTX split RoPE is `[B, H, S, D]`, so `H == S` causes full RoPE to be paired with sharded `x`, which can shape-mismatch or apply wrong positions. Please let callers pass the sequence dim or prefer the known RoPE layouts explicitly.

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
```diff
@@ -439,3 +446,163 @@ def support_fused_qkv(cls) -> bool:
...
+        self._p2p_reqs.clear()
+
+    def _ensure_buffers(self, q: torch.Tensor, k: torch.Tensor) -> None:
+        B, S, H, D = q.shape
+        H_kv = k.shape[2]
+        key = (B, S, H, H_kv, D, q.device, q.dtype, k.dtype)
+        if key == self._buf_key:
+            return
+        self._kv_bufs = k.new_empty(2, 2, B, S, H_kv, D)
+        self._out_buf = q.new_empty(B, S, H, D)
```
chang-l: `_out_buf = q.new_empty(...)` inherits Q's bf16 dtype but`_out_buf` should accumulate in fp32 and cast back on return. The merge uses fp32 LSEs, there is a rounding error bf16<->fp32 in every ring hop and can drift accuracy esp. when ring size grows.

docs/source/models/visual-generation.md
```diff
@@ -37,13 +37,13 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
 
 ### Feature Matrix
 
-| Model | FP8 blockwise | NVFP4 | TeaCache | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve |
+| Model | FP8 blockwise | NVFP4 | TeaCache | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve | Attention2D | Ring Attention |
```
chang-l, line 40: After adding two columns, the separator row below still has only 10 cells. Please update it to 12 cells or this table will render incorrectly in MyST/Markdown.

docs/source/models/visual-generation.md
```diff
@@ -113,12 +113,14 @@ TeaCache caches transformer outputs when timestep embeddings change slowly betwe
...
 ### Multi-GPU Parallelism
 
-Three parallelism modes can be combined:
+5 parallelism modes can be combined:
 
 - **CFG Parallelism** (`--cfg_size 2`): Splits positive/negative guidance prompts across GPUs.
 - **Ulysses Parallelism** (`--ulysses_size N`): Splits the sequence dimension across GPUs for longer sequences.
 - **Parallel VAE** (`--parallel_vae_size N`): Shards the final VAE decode along a spatial axis across GPUs, useful to reduce VAE latency and improve GPU utilization (Constraint: `parallel_vae_size ≤ world_size`). Currently only supported for WAN models.
-
+- **Attention Parallel**: There are 2 methods supported to run attention parallel -
```
chang-l: `git diff --check` fails here because of trailing whitespace. It also reports trailing whitespace in `tensorrt_llm/_torch/visual_gen/mapping.py` and `tests/unittest/_torch/visual_gen/multi_gpu/test_visual_gen_mapping.py`, so please clean those before merge (reported by CC).

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -161,6 +164,22 @@ def build_mesh(self):
...
+        if self.cp_size * self.ulysses_size > 1:
+            cp_idx = self._dim_names.index("cp")
+            uly_idx = self._dim_names.index("ulysses")
+            if abs(cp_idx - uly_idx) != 1:
+                raise RuntimeError(
+                    "seq_group requires cp and ulysses adjacent; "
+                    f"fix _DEVICE_MESH_DIM_ORDER (got {self._dim_names!r})"
+                )
+            # Linearisation: seq_rank = cp_rank * ulysses_size + ulysses_rank (cp outer).
+            VisualGenMapping.seq_mesh = cls.device_mesh["cp", "ulysses"]._flatten(
```
chang-l, line 190: [Captured by CC] Class-level singleton concern: this caches `seq_mesh` on the `VisualGenMapping` class, and `cls.device_mesh` is also a singleton on `DeviceMeshTopologyImpl`. Combined with the early-return at the top of `build_mesh()` (`if cls.device_mesh is not None: return`), the very first `VisualGenMapping` instantiated in the process wins.

A later instance with different `_dim_sizes` (e.g. the refiner, which has its own `refiner_dit_ring_size` / `refiner_dit_ulysses_size` in `ParallelConfig`) silently inherits the first instance's mesh + `seq_mesh` flatten + dim names. No consistency check is performed, so a misconfigured refiner will run on the wrong mesh without any error.

At minimum, please add an assertion that the cached `cls.device_mesh.mesh_dim_names` and shape match this instance's `_dim_names`/`_dim_sizes` before returning. Long-term, key the cache on `(dim_names, dim_sizes)` instead of storing it as a process-wide singleton on the parent class.

chang-l (commented):
[sorry if i missed it] Did you add test coverage for Ring+Ulysses (Ring+Ulysses+CFG) combined?

tensorrt_llm/_torch/visual_gen/config.py
NVShreyas: Oh I had enabled this when testing in my other branch and forgot to add it here. 
Updated

tensorrt_llm/_torch/visual_gen/utils.py
NVShreyas: done

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
NVShreyas: added divisibility check for non ulysses CP. Ideally we should add toggles for CP as well similar to ulysses

tensorrt_llm/_torch/visual_gen/utils.py
NVShreyas: done

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
NVShreyas: good catch, done

docs/source/models/visual-generation.md
NVShreyas: updated

tensorrt_llm/_torch/visual_gen/mapping.py
NVShreyas: This should not affect anything currently since there is only 1 instance of `VisualGenMapping`. But I added an assertion and a note about instantiated multiple in the future

NVShreyas:
> Ring+Ulysses (Ring+Ulysses+CFG) combined?

Yes Ring+ulysses tests exist. I have added new tests for CFG+Ulysses+Ring

chang-l (approved):
Thanks for the effort!

NVShreyas:
/bot skip --comment "perf regressions are localized to LLMs and unrelated to visual_gen and current changes"

PR #13864, opened 2026-05-07
[TRTLLM-12467][feat] EPD improvements

venkywonka:
## Summary

This PR restructures the multimodal disaggregated (Encoder/Prefill/Decode) serving pipeline to properly track encoder-output embedding lengths separately from prompt-side MM token counts.

Core problem: Previously multimodal_lengths (prompt-side counts that may include special/framing tokens) was used to split encoder outputs, causing shape mismatches (TRTLLM-12175).

Solution: Introduce multimodal_embedding_lengths as a first-class field throughout the pipeline:
```python
# Before: tuple return, no embedding lengths
def get_prompt_token_ids(inputs, handles)
  -> (ids, lengths, positions)
# After: typed dataclass with embedding lengths
def build_disagg_prefill_multimodal_inputs(
    inputs, handles
) -> DisaggPrefillMultimodalInputs
  # .multimodal_lengths = prompt spans
  # .multimodal_embedding_lengths
  #   = encoder output rows
```

Key changes:

1. New `DisaggPrefillMultimodalInputs` dataclass replaces `(ids, lengths, positions)` tuple
2. `MultimodalResult` now carries sparse `mm_embedding_request_indices` for mixed text/MM batches
3. Validation extracted from `MultimodalInput` into reusable standalone functions
4. NemotronNano gains full EPD support with `NanoV2VLMultimodalEncoder`
5. Qwen3VL video+image placeholder expansion generalized
`_is_disagg → _is_mm_disagg` clarifies scope vs generic disagg serving
6. Dedicated EPD Accuracy e2e harness that runs video-mme benchmark for nano-omni-v3[fp8, nvfp4], qwen3vl-2b. 

## Out of Scope

The following are intentionally **not** addressed in this PR and are tracked as follow-up work:

1. **EVS (Efficient Video Sampling) video token pruning + EPD** — combining EVS-based video token pruning with the disaggregated Encoder/Prefill/Decode pipeline.
2. **Video + audio + EPD** — combining simultaneous video and audio modalities with the disaggregated Encoder/Prefill/Decode pipeline.
3. **Full native E/PD multi-gpu disaggregation** - E and PD currently only works if the workers are colocated and sharing same gpu. This is because, today only the cuda IPC handles are transported between E and PD workers - the actual gpu buffers aren't. Therefore the consumer uses the ipc handles to open gpu buffers owned by the producer. That read is non-garbage only if the consumer worker runs on the same device as the producer. If they are in different devices, the responsibility of moving the gpu tensors across devices is currently not done by TRTLLM natively.

## Validation

| gpu | model | benchmark | validation |
|---|---|---|---|
| 8xH100 | Nemotron Nano V3 Omni FP8 | VideoMME | Threshold 62.785; AGG 69.667; E/PD 68.667 |
| 8xH200 | Qwen3-VL-30B BF16 | VideoMME | AGG baseline 62; E/PD 62 |
| 2xB100 | Qwen3-VL-2B Instruct | VideoMME | Threshold 47.785; AGG 54.667; E/PD 54.667 |
| 2xB100 | Nemotron Nano V3 Omni NVFP4 | VideoMME | Threshold 61.285; AGG 68.334; E/P/D 69 |

## Release Notes

* **Refactor**
  * Enhanced multimodal disaggregated serving infrastructure for improved embedding handling and shared tensor lifetime management across distributed execution.
  * Standardized multimodal input processing APIs across model implementations.

* **Tests**
  * Added comprehensive integration tests for disaggregated multimodal serving scenarios including video processing.


tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -81,12 +81,32 @@
...
+        if key in _RESERVED_MM_DATA_KEYS:
+            continue
+        extra_args[key].append(value)
+
+
+def _metadata_to_list(value: Any) -> List[int]:
+    if value is None:
+        return []
+    if torch.is_tensor(value):
+        value = value.detach().cpu().tolist()
```
venkywonka: these methods might be unnecessary

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -827,10 +847,7 @@ def _prepare_inputs(
             # store extra arguments
             if request.py_multimodal_data is not None:
                 has_context_multimodal_data = True
-                for k, v in request.py_multimodal_data.items():
-                    if k in _RESERVED_MM_DATA_KEYS:
-                        continue
-                    extra_args[k].append(v)
+                _append_mm_extra_args(extra_args, request.py_multimodal_data)
```
venkywonka: why are these changes related to epd support?

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
```diff
@@ -256,6 +256,23 @@ def get_multimodal_embeddings(
     return [all_embeddings]
 
 
+def _get_cached_multimodal_embeddings(
+        multimodal_params: List[MultimodalParams]) -> List[torch.Tensor]:
+    cached_embeddings = []
+    for param in multimodal_params:
```
venkywonka, line 290: each param corresponds to a request? or a multimodal item?

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
```diff
@@ -256,6 +256,23 @@ def get_multimodal_embeddings(
...
+        embeds = param.multimodal_data.get("multimodal_embedding")
+        if embeds is None:
+            continue
+        if isinstance(embeds, list):
+            embeds = torch.cat(embeds, dim=0)
```
venkywonka, line 297: what entry points can lead to embeds being a list?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -810,6 +836,29 @@ def _video_tubelet_geometry(self, t: int, T: int, ih: int, iw: int) -> Tuple[int
         return num_tubelets, wh
 
 
+class NanoV2VLMultimodalEncoder(NanoV2VLVisionEncoder):
```
venkywonka, line 890: i don't think a new child class NanoV2VLMultimodalEncoder is necessary

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -2364,6 +2498,8 @@ def _resample_audios(
 )
 
 
+@support_multimodal_disaggregated
+@register_vision_encoder(NanoV2VLMultimodalEncoder)
```
venkywonka, line 2556: was nano validated on EPD?

tensorrt_llm/_torch/modules/rms_norm.py
```diff
@@ -87,6 +87,47 @@ def __init__(
                 return_hp_output = False
         self.return_hp_output = return_hp_output
 
+    def _torch_rms_norm(
```
venkywonka: what is this change? why is this relevant? is this due to an improper rebase conflict resolution?

tests/integration/defs/disaggregated/disagg_test_utils.py
```diff
@@ -106,7 +107,9 @@ def _run_worker(
         yaml.dump(worker_config, f)
         f.flush()
         cmd = [
-            "trtllm-serve",
+            sys.executable,
```
venkywonka: are these changes necessary?

tests/unittest/_torch/modeling/test_modeling_nemotron_nano_v2_vl.py
```diff
@@ -14,13 +15,16 @@
 from test_modeling_nemotron_h import extract_decode_logprobs
 
 from tensorrt_llm import LLM
+from tensorrt_llm._torch.models import modeling_nemotron_nano as nemotron_nano
 from tensorrt_llm._torch.models.modeling_multimodal_utils import get_multimodal_embeddings
 from tensorrt_llm._torch.models.modeling_nemotron_nano import (
     NanoV2VLInputProcessor,
+    NanoV2VLMultimodalEncoder,
```
venkywonka, line 22: what's the difference between NanoV2VLMultimodalEncoder and NanoV2VLVisionEncoder?

tests/unittest/_torch/modeling/test_modeling_nemotron_nano_v2_vl.py
```diff
@@ -36,6 +40,143 @@
...
+    video = SimpleNamespace(frames=[object()], metadata={}, audio=None)
+    expanded_ids, mm_lengths, mm_offsets, layout = processor.get_prompt_token_ids(
+        {
+            "prompt": "Question <video> answer",
+            "multi_modal_data": {"video": [video]},
+        },
+        [{"tensor_size": (4, 16)}],
+    )
+
+    assert expanded_ids == [101, 30, 20, 20, 31, 55, 30, 20, 20, 31, 102]
```
venkywonka: thsi test isn't really useful if magicmock is used right?

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -60,6 +61,96 @@
...
+            special_token_offsets.append(flat_mm_offset)
+
+        write_pos += mm_token_num
+        flat_mm_offset += prompt_mm_length
+        mm_handle_idx += 1
+
+    if write_pos != final_length:
+        raise RuntimeError(f"Write position mismatch: {write_pos} != {final_length}")
+
+    return DisaggPrefillMultimodalInputs(
```
2ez4bz, line 143: Just match what agg is doing; we were returning incomplete stuff before.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -267,8 +358,14 @@ def get_num_tokens_per_video(
     ) -> int:
         merge = self.config.vision_config.spatial_merge_size
         if video_grid_thw is not None:
-            t, h, w = (int(x) for x in video_grid_thw)
```
2ez4bz, line 269: Old code assumed single `video_grid_thw` - ??

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -267,8 +358,14 @@ def get_num_tokens_per_video(
...
-            return t * (h // merge) * (w // merge)
+            grid_thw = torch.as_tensor(video_grid_thw)
+            if grid_thw.ndim == 1:
+                grid_thw = grid_thw.reshape(1, 3)
+            if grid_thw.ndim != 2 or grid_thw.shape[-1] != 3:
+                raise ValueError("video_grid_thw must have shape [3] or [num_segments, 3]")
+            return sum(
+                int(t) * (int(h) // merge) * (int(w) // merge) for t, h, w in grid_thw.tolist()
```
2ez4bz: Can we stay in torch land?

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -267,8 +358,14 @@ def get_num_tokens_per_video(
     ) -> int:
         merge = self.config.vision_config.spatial_merge_size
         if video_grid_thw is not None:
-            t, h, w = (int(x) for x in video_grid_thw)
-            return t * (h // merge) * (w // merge)
+            grid_thw = torch.as_tensor(video_grid_thw)
```
2ez4bz: Are we not passing it in as a tensor already from the caller?

tensorrt_llm/llmapi/llm.py
```diff
@@ -596,8 +597,17 @@ def _preprocess(
                     "Multimodal disaggregated inference is not supported for this model"
                 )
             mm_handles = disaggregated_params.multimodal_embedding_handles
-            prompt_token_ids, mm_token_length, mm_token_positions = self.input_processor.get_prompt_token_ids(
-                inputs, mm_handles)
+            # TODO(TRTLLM-12419): Pass encoder-side MM layout through
```
2ez4bz: https://jirasw.nvidia.com/browse/TRTLLM-12869

tests/integration/test_lists/test-db/l0_dgx_b200.yml
```diff
@@ -35,6 +35,7 @@ l0_dgx_b200:
   - disaggregated/test_disaggregated.py::test_disaggregated_deepseek_v3_lite_fp8_ucx[DeepSeek-V3-Lite-fp8]
   - disaggregated/test_disaggregated.py::test_disaggregated_deepseek_v3_lite_fp8_nixl[DeepSeek-V3-Lite-fp8]
   - disaggregated/test_disaggregated.py::test_disaggregated_gpt_oss_120b_harmony[gpt_oss/gpt-oss-120b]
+  - accuracy/test_disaggregated_serving.py::test_disaggregated_videomme_vlm_epd[nemotron_nano_v3_omni_nvfp4]
```
xinhe-nv: please add the test into qa/llm_function_core.txt, too.

tests/integration/test_lists/test-db/l0_dgx_h100.yml
```diff
@@ -38,6 +38,7 @@ l0_dgx_h100:
   - accuracy/test_disaggregated_serving.py::TestGemma3_1BInstruct::test_auto_dtype[False]
   - accuracy/test_disaggregated_serving.py::TestGemma3_1BInstruct::test_auto_dtype[True]
   - accuracy/test_disaggregated_serving.py::TestGemma3_1BInstruct::test_kv_cache_v2_nixl_python
+  - accuracy/test_disaggregated_serving.py::test_disaggregated_videomme_vlm_epd[nemotron_nano_v3_omni_fp8]
```
xinhe-nv: please add the test into qa/llm_function_core.txt, too.

tests/integration/test_lists/test-db/l0_h100.yml
```diff
@@ -147,6 +147,7 @@ l0_h100:
   - disaggregated/test_disaggregated.py::test_disaggregated_deepseek_v3_lite_fp8_tp1_single_gpu_mtp[DeepSeek-V3-Lite-fp8]
   - disaggregated/test_disaggregated.py::test_disaggregated_deepseek_v3_lite_fp8_tp1_two_mtp[DeepSeek-V3-Lite-fp8]
   - disaggregated/test_disaggregated.py::test_disaggregated_deepseek_v3_lite_fp8_ucx_tp1_single_gpu[DeepSeek-V3-Lite-fp8]
+  - disaggregated/test_disaggregated_multimodal.py::test_qwen3vl_epd_video_split_item_runs[Qwen3-VL-8B-Instruct] TIMEOUT (30)
```
xinhe-nv: please add the test into qa/llm_function_core.txt, too.

xinhe-nv (approved):

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
2ez4bz: Seconded. If this is the input, should an error be raised instead, as it would point to incorrect usage? Not to mention, the `embeds` may not be concatenate-able, which would also lead to an error.

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
```diff
@@ -291,10 +327,15 @@ def find_input_mm_embeds(
     Note:
         - Supports both individual batching (len(mm_embeds) == len(multimodal_params))
           and pre-concatenated batching (len(mm_embeds) == 1)
+        - Call get_attached_multimodal_embeddings before this helper when
```
2ez4bz, line 330: Is this to make sure the embeddings are "normalized" to an expected format?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -878,6 +886,29 @@ def _video_tubelet_geometry(self, t: int, T: int, ih: int, iw: int) -> Tuple[int
...
 
 
+class NanoV2VLMultimodalEncoder(NanoV2VLVisionEncoder):
+    def __init__(self, model_config: ModelConfig[transformers.PretrainedConfig], *args, **kwargs):
+        super().__init__(model_config)
+
+    def forward(self, multimodal_params: List[MultimodalParams]) -> List[torch.Tensor]:
+        for param in multimodal_params:
+            modality_type = param.multimodal_data["modality_type"]
+            if modality_type == "audio":
```
2ez4bz, line 905: These checks are confusing since Nano v3 (which unfortunately is spaghetti'ed into the same classes) does support these modalities. Is this just for EPD/? If so, could you make that clear in the docstring for this class?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -2227,6 +2258,93 @@ def __call__(
             "multimodal_data": multimodal_data,
         }
 
+    def build_disagg_prefill_multimodal_inputs(
+        self, inputs: TextPrompt, mm_handles: List[Dict[str, Any]]
+    ) -> DisaggPrefillMultimodalInputs:
+        text_prompt = inputs.get("prompt")
+        if not text_prompt:
+            raise ValueError("Text prompt is required but not provided")
```
2ez4bz: Is this an actual requirement? Lines 2305-2307 would suggest `prompt_token_ids` are fine as well, we just need one or the other.

tensorrt_llm/_torch/pyexecutor/llm_request.py
```diff
@@ -314,6 +322,7 @@ def __init__(self,
             chunk_size=self._chunk_size) if return_generation_logits else None
         self._log_probs = LogProbStorage() if return_log_probs else None
         self._mm_embeddings: Optional[List[Dict[str, Any]]] = None
+        self._shared_tensor_lifetime_refs: List[torch.Tensor] = []
```
2ez4bz: Nit: leave a code comment for why this was added? Specifically, it would be good to precisely state whether this is for CPU or GPU tensors we're somehow passing across the process barriers (with typical examples), or both.

tensorrt_llm/_torch/pyexecutor/llm_request.py
```diff
@@ -926,6 +936,49 @@ def convert_wordlist(word_list) -> List[List[int]]:
...
+    if values is None:
+        return None
+    if not isinstance(values, list):
+        raise TypeError(f"{field_name} must be a list")
+    if not all(isinstance(value, int) for value in values):
+        raise TypeError(f"{field_name} must contain only integers")
+    return values
+
+
+def get_multimodal_embedding_lengths(request: Any) -> Optional[List[int]]:
```
2ez4bz: Just checking: should the type for `request` by hardened? If so, does it allow for getting rid of the `getattr` calls on lines 952 and 967?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -291,9 +291,39 @@ def is_generation_model(self) -> bool:
 @dataclass(kw_only=True)
 class MultimodalResult:
     mm_embeddings: List[torch.Tensor]
+    mm_embedding_lengths: List[List[int]]
```
2ez4bz, line 295: Nit: can you add code comments for why these 2 are added?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -291,9 +291,39 @@ def is_generation_model(self) -> bool:
...
+    mm_embedding_request_indices: List[int]
     # Can be used to include e.g. `mrope_position_ids`, etc.
     extra_data: Optional[Dict[str, Any]] = None
 
+    def __post_init__(self) -> None:
+        if len(self.mm_embedding_lengths) != len(self.mm_embeddings):
+            raise ValueError("mm_embedding_lengths batch size does not match mm_embeddings")
```
2ez4bz: Nit: include values in error message? Could use walrus operator in the line above + reuse variables here.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -291,9 +291,39 @@ def is_generation_model(self) -> bool:
...
+                raise ValueError(
+                    f"mm_embedding shape mismatch: {len(mm_embedding)} != {sum(mm_embedding_lengths)}"
+                )
+
+    @classmethod
+    def from_model_outputs(
+        cls, model_outputs: Dict[str, Any], num_context_requests: int
+    ) -> "MultimodalResult":
+        result = cls(
+            mm_embeddings=model_outputs.pop("mm_embeddings"),
```
2ez4bz: Why are we popping them? Doesn't this modify the caller's value, which is potentially dangerous? If there is a good reason to do, please document it. I would prefer to avoid this altogether, though.

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -356,26 +385,24 @@ def update_requests(
...
-            if len(mm_embedding) != sum(request.multimodal_lengths):
-                raise ValueError(
-                    f"mm_embedding shape mismatch: {len(mm_embedding)} != {sum(request.multimodal_lengths)}"
-                )
 
-            request.py_result.append_mm_embeddings(mm_embedding, request.multimodal_lengths)
+        request_indices = state.data.mm_embedding_request_indices
+        for result_index, (request_index, mm_embedding) in enumerate(
+            zip(request_indices, mm_embeddings)
+        ):
```
2ez4bz, line 422: Nit: `zip(..., strict=True)` ?

tensorrt_llm/executor/result.py
```diff
@@ -201,6 +215,7 @@ def __init__(self,
             CompletionOutput(i) for i in range(self.sampling_params.best_of)
         ]
         self._context_logits: Optional[torch.Tensor] = None
+        self._shared_tensor_lifetime_refs: List[Any] = []
```
2ez4bz: Could you similarly add comments for why this is necessary?

tensorrt_llm/executor/result.py
```diff
@@ -818,8 +846,23 @@ def __init__(
             postproc_params=generation_request.postproc_params,
         )
         self._generation_request = generation_request
+        multimodal_params = getattr(generation_request, "multimodal_params",
+                                    None)
+        if multimodal_params is not None:
+            # `generate_async` deletes `request.multimodal_params` after
+            # submit(), but multiprocess workers may restore request-side shared
```
2ez4bz: ! was this the source of the bug?

tensorrt_llm/disaggregated_params.py
```diff
@@ -62,6 +62,23 @@ class DisaggregatedParams:
...
     mrope_position_deltas_handle: Optional[Dict[str, Any]] = None
 
+    # Private lifetime anchors for SharedTensorContainer dicts. The dict only
+    # names shared storage; it does not keep that storage alive.
+    _shared_tensor_lifetime_refs: List[Any] = field(default_factory=list, repr=False, compare=False)
+
+    def __getstate__(self) -> Dict[str, Any]:
+        state = {
+            name: getattr(self, name)
+            for name in self.__dataclass_fields__
```
2ez4bz: Is there a difference between this vs `dataclasses.fields(self)`?

venkywonka: `__dataclass_fields__` is a dict of names→Field, `fields(self)` is a tuple of Field objects.. they're equivalent...
looks like `self.__dataclass_fields__` is a common pattern in custom pickle hooks... but agreed `fields(self)` is the public/idiomatic api so will prefer that

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
venkywonka: yea that's confusing my bad! updated in the docstring. the newly added adapter class `NanoV2VLMultimodalEncoder` merely gates on invalid cases (audio) and makes sure `forward()` returns a single tensor instead of a tuple

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
venkywonka: yep that's a bug thank you for the catch!

tensorrt_llm/_torch/models/modeling_qwen3vl.py
venkywonka: yes that's unnecessarily defensive, have tightened contract to always keep in torch land

tensorrt_llm/_torch/models/modeling_qwen3vl.py
venkywonka: yes sir

tensorrt_llm/_torch/models/modeling_utils.py
```diff
@@ -698,12 +698,14 @@ class SomeVLModel(...):
     """
 
     def wrapper(model_cls: Type[nn.Module]) -> Type[nn.Module]:
+        registered = False
```
venkywonka, line 758: One model class can have many arch aliases (eg, NemotronH_Nano_VL_V2, NemotronH_Nano_Omni_Reasoning_V3). previous changes breaked on first registration. this fixes that.

tensorrt_llm/_torch/pyexecutor/llm_request.py
venkywonka: yes thank you for that, have added it!

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
venkywonka: previously, `mm_embeds` was assumed to only originate from the in-process encoder in the model-specific forwards... so in EPD flow `mm_embeds` ended up empty. They are actually attached inside `multimodal_params.multimodal_data.multimodal_embedding` ... so extracting it from there should be done outside of `find_input_mm_embeds` 's contract. 

basically intention is to make sure `find_input_mm_embeds`'s contract always expects non-None `mm_embeds`

tensorrt_llm/_torch/pyexecutor/sampler.py
venkywonka: done

tensorrt_llm/_torch/pyexecutor/sampler.py
venkywonka: done

tensorrt_llm/_torch/pyexecutor/sampler.py
venkywonka: yep that's a bug thanks for the catch!

tensorrt_llm/_torch/pyexecutor/sampler.py
venkywonka: done

tests/integration/test_lists/test-db/l0_dgx_h100.yml
venkywonka: thanks! done.

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -264,31 +358,16 @@ def get_num_tokens_per_video(
         video_grid_thw: Optional[torch.Tensor] = None,
         **kwargs,
     ) -> int:
```
venkywonka, line 360: callout: have removed the full process fallback as its anyways a bug when that path is reached and hence fails with ValueError now.

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
2ez4bz: Gotcha. What happens if the video has audio and the media IO was setup with `--media_io_kwargs '{"video": {"extract_audio": true}}'` ?

tensorrt_llm/_torch/pyexecutor/llm_request.py
2ez4bz: @venkywonka GitHub makes it hard to view new diffs for rebased changes - did you no longer need `self._shared_tensor_lifetime_refs` ?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -2994,16 +3124,44 @@ def forward(
             ctx_params = multimodal_params[:num_context_requests]
             if self.video_pruning_rate > 0:
                 self._validate_evs_context_batch(ctx_params, num_context_requests)
-            if not _is_disagg():
+            raw_ctx_params = [param for param in ctx_params if has_raw_multimodal_payload(param)]
```
2ez4bz, line 3147: Nit: for readability, can we move input validation to a helper function or method?

tensorrt_llm/_torch/pyexecutor/sampler.py
```diff
@@ -291,9 +291,63 @@ def is_generation_model(self) -> bool:
...
+        }
+        result = cls(
+            mm_embeddings=model_outputs["mm_embeddings"],
+            mm_embedding_lengths=model_outputs["mm_embedding_lengths"],
+            mm_embedding_request_indices=model_outputs["mm_embedding_request_indices"],
+            extra_data={
+                key: value for key, value in model_outputs.items() if key not in result_keys
+            },
+        )
+        for request_index in result.mm_embedding_request_indices:
```
2ez4bz: Doesn't this make more sense to be in the `__post_init__`?

tests/integration/defs/accuracy/test_disaggregated_serving.py
```diff
@@ -519,6 +526,218 @@ def _show_kvcache_time(kv_cache_perf_dir, max_lines=100):
...
+    """Launch separate encoder and combined prefill/decode llmapi instances."""
+    env_updates = dict(extra_env or {})
+    # Tell PD to consume encoder handles, not raw multimodal payloads.
+    env_updates["TLLM_MULTIMODAL_DISAGGREGATED"] = "1"
+    with contextlib.ExitStack() as stack:
+        stack.enter_context(mock.patch.dict(os.environ, env_updates))
+        if "TMPDIR" in env_updates:
+            # tempfile caches TMPDIR. Keep the cache in sync with the env.
+            stack.enter_context(
+                mock.patch.object(tempfile, "tempdir", env_updates["TMPDIR"]))
```
2ez4bz: Out of curiosity, which line(s) of code in this context manager uses the `tempfile.tempdir` function?

tests/integration/defs/accuracy/test_disaggregated_serving.py
```diff
@@ -519,6 +526,218 @@ def _show_kvcache_time(kv_cache_perf_dir, max_lines=100):
...
+        sampling_params=sampling_params,
+        extra_evaluator_kwargs=extra_evaluator_kwargs,
+        max_workers=VideoMME.MAX_BATCH_SIZE,
+        expected_quant_algo=expected_quant_algo,
+    )
+
+
+@pytest.mark.timeout(DEFAULT_TEST_TIMEOUT)
+@skip_pre_hopper
+@pytest.mark.skip_less_device(2)
```
2ez4bz: For my own benefit, what in the code tells the encoder and the PD workers to use different GPUs?

tests/unittest/_torch/executor/test_chunked_logits.py
```diff
@@ -5,6 +5,21 @@
 logits handling through device-side fragments and batched host transfers.
 """
 
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
2ez4bz: Lol. Shouldn't this be at top of the file? Could also revert since it's unrelated to this change.

tests/unittest/_torch/pyexecutor/test_multimodal_embedding_lengths.py
```diff
@@ -0,0 +1,270 @@
...
+from tensorrt_llm._torch.pyexecutor.sampler import EarlyStopWithMMResult, MultimodalResult
+from tensorrt_llm._torch.shared_tensor import SharedTensorContainer
+from tensorrt_llm.bindings.executor import FinishReason
+
+
+@pytest.mark.parametrize(
+    "req,expected",
+    [
+        (
+            SimpleNamespace(multimodal_lengths=[3, 5], py_multimodal_data=None),
```
2ez4bz, line 23: Can the actual request object be used? Or is it too annoying to instantiate?

tests/unittest/_torch/multimodal/test_qwen3vl_disagg_prompt.py
```diff
@@ -0,0 +1,149 @@
...
+            video_token_id=4,
+            vision_start_token_id=1,
+            placeholder_id=99,
+        )
+
+
+def test_qwen3vl_video_token_count_sums_single_video_grid_rows():
+    """One logical video may be represented by multiple processor grid rows."""
+    processor = object.__new__(Qwen3VLInputProcessorBase)
+    object.__setattr__(
```
2ez4bz: Jusat checking: does `processor._config = ...` not work?

2ez4bz (approved):

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -878,6 +887,39 @@ def _video_tubelet_geometry(self, t: int, T: int, ih: int, iw: int) -> Tuple[int
         return num_tubelets, wh
 
 
+class NanoV2VLMultimodalEncoder(NanoV2VLVisionEncoder):
```
moraxu, line 890: Should we update https://github.com/moraxu/TensorRT-LLM/blob/main/docs/source/models/supported-models.md#multimodal-feature-support-matrix-pytorch-backend ?

tensorrt_llm/_torch/pyexecutor/sampler.py
venkywonka: `num_context_requests` was previously not a member of the class, so `__post_init__` could not validate using it... i have included it so we can seperate concerns

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
venkywonka: thanks for the catch! updated

tensorrt_llm/_torch/models/modeling_llava_next.py
yechank-nvidia: Non-blocking comment) @2ez4bz Calling this out. Don't we support DISAGG for Llava-Next?

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -4254,36 +4255,42 @@ def _forward_step_mm_encoder_only(
...
-                sum(request.multimodal_lengths)
-                for request in scheduled_requests.context_requests
-                if request.multimodal_lengths is not None
-            ]
+            return {
+                'mm_embeddings': [],
+                'mm_embedding_request_indices': [],
+                'mm_embedding_lengths': [],
+            }
+        active_multimodal_requests = [(request_idx, request)
```
yechank-nvidia: In mm_encoder_only, `request.py_multimodal_data` is not None is too broad for Qwen-VL because text-only prompts still carry `mrope_config`. A text-only request, or a mixed text-only/MM encoder batch, is treated as active multimodal and then fails because `multimodal_embedding_lengths` is absent. Please filter active encoder requests by real image/video/audio payload or explicit `multimodal_embedding_lengths`, not by metadata presence.

StanleySun639 (approved):

nvpohanh (approved):

amukkara (approved):

laikhtewari (approved):

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
```diff
@@ -35,10 +35,18 @@
 
 
 # Make this a runtime lookup rather than a module-wide constant for easier unit testing.
-def _is_disagg() -> bool:
+# MM E/P split flag. Not generic disaggregated serving.
+def _is_mm_disagg() -> bool:
     return os.getenv(_MULTIMODAL_ENV_NAME, "0") == "1"
```
chang-l, line 40: Are we actually using this env var in any of our EPD use cases?

chang-l (approved):

tensorrt_llm/llmapi/llm.py
bmarimuthu-nv, line 577: AutoDeploy also doesn't support E/P disagg today. Can we error out for `if self.args.backend == "_autodeploy"` as well?

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -58,18 +58,51 @@
...
-# Non-tensor multimodal metadata consumed by _store_prefill_multimodal_metadata.
-# These keys must NOT leak into the generic extra_args dict — entries there
-# are expected to be tensors, and these are lists or nested dicts.
+# Non-forward multimodal metadata consumed by _store_prefill_multimodal_metadata.
+# These keys must NOT leak into the generic extra_args dict: entries there are
+# expected to be tensors passed to the model forward.
 _RESERVED_MM_DATA_KEYS = frozenset(
     {
         "layout_metadata",
+        "multimodal_embedding_lengths",
```
bmarimuthu-nv, line 70: I think only this update is needed in this file. Other changes can be reverted, provided we error out in `llm.py` saying mm_disagg is not yet supported in AD.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -58,18 +58,51 @@
...
 )
 
 
+def _append_mm_extra_args(
+    extra_args: Dict[str, List[torch.Tensor]], multimodal_data: abc.Mapping[str, Any]
+) -> None:
+    for key, value in multimodal_data.items():
+        if key in _RESERVED_MM_DATA_KEYS:
+            continue
+        if key == "multimodal_embedding" and not torch.is_tensor(value):
```
bmarimuthu-nv: AD today is not doing multimodal E/P handoff. In AD multimodal, the prefill forward receives raw modality tensors and runs the encoder inside the AD wrapper. So `multimodal_embedding` is not something AD should expect or consume today.

This change is not needed once we llm.py to fail early.

If I were to keep it, we can do:

```
if key == "multimodal_embedding":
      raise NotImplementedError(
          "AutoDeploy does not support precomputed multimodal_embedding handoff. "
          "AutoDeploy expects raw multimodal inputs and runs the encoder in prefill."
      )
```

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
venkywonka: by the time we're inspecting `param.multimodal_data["modality_type"]`, ~the audio is already its own item tagged modality_type == "audio"~ we would have already extracted the audio payload in `param.multimodal_data["video"].get("audio")`- so when extract_audio is true, unfortunately this will raise too in the below lines. i shall add a TODO tracking an improvement for this. thanks for calling it out!

venkywonka: it is actually NOT gated by `if modality_type=="audio"` but by `audio_data = param.multimodal_data[modality_type].get("audio")` when `modality_type=="video"`.

tensorrt_llm/_torch/pyexecutor/model_engine.py
venkywonka: great catch! will update it!

tensorrt_llm/_torch/models/modeling_llava_next.py
venkywonka: yep this seems to be a pre-existing contradiction, where it advertises that it supports disagg but crashes at forward. i think to be on safer side and validate things, i shall revert disagg changes for this file, explicitly change this model to NOT support disagg until we have validated it. i shall also create a jira for this

tensorrt_llm/llmapi/llm.py
venkywonka: gotcha! yes thats valid, have done fail-fast in llm.py entry-point itself. thanks!

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
venkywonka: yep, done! thanks!

tensorrt_llm/_torch/models/modeling_llava_next.py
2ez4bz: I think this is a misleading raise - we do support E/P/D disagg for this model: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/_torch/multimodal/test_mm_encoder_standalone.py#L31

2ez4bz: I think the changes in this file can just be backed out for now.

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
venkywonka: so currently dynamo doesn't use it, but its used to gate redundantly instantiating the encoder in the PD workers when not needed (as well as its usage in the tests )

bmarimuthu-nv (approved):
LGTM AutoDeploy, thanks!

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
venkywonka: i believe would benefit from using this flag so gpu memory for the encoder is saved.

tensorrt_llm/_torch/models/modeling_llava_next.py
venkywonka: have updated in latest commit

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

tburt-nv:
The pipeline in https://github.com/NVIDIA/TensorRT-LLM/pull/13864#issuecomment-4614726781 failed due to a cluster issue and is tearing itself down; I'll trigger a new one now.

tburt-nv:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

tests/integration/test_lists/test-db/l0_h100.yml
```diff
@@ -41,7 +41,7 @@ l0_h100:
   - unittest/_torch/modules/moe/test_moe_backend.py::test_moe_backend -k "CUTLASS"
   # ------------- MoE: test_single_gpu (by backend) ---------------
   - unittest/_torch/modules/moe/test_moe_module.py::test_configurable_moe_single_gpu -k "CUTLASS"
-  - unittest/_torch/multimodal
+  - unittest/_torch/multimodal -k "not nemotron_nano_v2_vl_fp8"
```
2ez4bz, line 45: Nit: maybe leave a comment for why it's excluded?

tests/integration/test_lists/test-db/l0_h100.yml
```diff
@@ -296,6 +296,7 @@ l0_h100:
       orchestrator: mpi
   tests:
   # ------------- PyTorch tests ---------------
+  - unittest/_torch/multimodal/test_mm_encoder_standalone.py -k "nemotron_nano_v2_vl_fp8"
```
2ez4bz: Ditto.

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

venkywonka:
/bot run --extra-stage "DGX_H100-PyTorch-Post-Merge-1, DGX_H100-PyTorch-Post-Merge-2"

Tabrizian (approved):

PR #13866, opened 2026-05-07
[TRTLLM-7081][refactor] Add MultimodalModelMixin

2ez4bz:
## Summary by CodeRabbit

* **New Features**
  * Introduced standardized multimodal input processing framework for consistent embedding integration across models.
  * Added validation to detect and prevent conflicting multimodal input specifications.

* **Refactor**
  * Updated multimodal embedding pipeline in Mistral3VLM for improved consistency with standardized framework.

* **Tests**
  * Added unit tests for multimodal input preparation workflow.

## Description

* Why?

All of our existing multimodal models follow a similar pattern:

1. select context requests from the current batch
2. select those that have multimodal parameters
3. run `get_multimodal_embeddings`, which hides encoder
    caching and chunked-prefill reuse behavior with an implicit
    API contract.
4. run `find_input_mm_embeds`, which slices embeddings
    for the current forward chunk.
5. run `fuse_input_embeds`, which places multimodal embeddings
    into text embeddings.
6. Call the underlying language model's `forward` method with
    the fused embeddings.

This means these features are all opt-in, and the code pattern is
duplicated, with developers needing to be aware of several implicit
contracts.

* What?

This commit adds a new `MultimodalModelMixin` that specifies
an interface that hides most of this, allowing model developers to
just implement the forward pass for multimodal embeddings.

As an illustration, it refactors the mistral small model to use this
new mixin.

### Flowchart for new mixin

```mermaid
flowchart TD
    A[Mixin forward] --> B[Count context requests]
    B --> C[Mixin prepare_multimodal_inputs]
    C --> D[Select context multimodal params]
    D --> E{Any params?}
    E -- No --> Z[Return input_ids with no inputs_embeds]

    E -- Yes --> F[Build encoder kwargs via model hook]
    F --> G[Find uncached params]
    G --> H{Encoder work needed?}
    H -- No: all selected params already have cached embeddings --> J[Gather cached embeddings]
    H -- Yes --> I[Call model encode_multimodal_inputs]
    I --> K[Validate primary tensor contract]
    K --> L[Cache per-request embedding slices]
    L --> J

    J --> M[after_full_multimodal_embeddings hook]
    M --> N[Find active chunk embeddings]
    N --> O{Any active MM rows?}
    O -- No: current chunk has no MM rows to fuse --> P[Return text-only chunk inputs]
    O -- Yes --> Q[after_active_multimodal_embeddings hook]
    Q --> R[fuse_input_embeds]
    R --> S[Return PreparedLlmInputs]
    S --> U[Build LLM forward kwargs via model hook]
    U --> T[Call language model forward]

    classDef base fill:#d8f5d0,stroke:#2f8f46,color:#143d1f;
    classDef model fill:#d7e9ff,stroke:#2f6fb3,color:#12365d;
    classDef optional fill:#fff2b8,stroke:#b38b00,color:#4d3b00;
    classDef decision fill:#f5f5f5,stroke:#707070,color:#202020;

    class A,B,C,D,G,J,K,L,N,P,R,S,T,Z base;
    class I model;
    class F,M,Q,U optional;
    class E,H,O decision;
```

Legend:

- Green: framework-owned mixin behavior.
- Blue: required model implementation.
- Yellow: optional model hooks.

Clarifications:

- `Encoder work needed? -> No` means every selected context multimodal request already has cached embeddings, so the mixin gathers cached rows without calling the encoder.
- `Any active MM rows? -> No` means `find_input_mm_embeds` returned no rows for this forward chunk. This can happen when all MM rows are covered by KV-cache reuse or when the current chunk does not overlap the request's multimodal placeholder rows.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_mistral.py
+++ b/tensorrt_llm/_torch/models/modeling_mistral.py
@@ -1,6 +1,6 @@
 import copy
 import dataclasses
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Sequence, Tuple
 
 import torch
 import torchvision
@@ -22,9 +22,10 @@ from tensorrt_llm._torch.models.checkpoints.mistral.weight_mapper import \
     MistralWeightMapper
 from tensorrt_llm._torch.models.modeling_mistral_large3 import (
     Mistral3Gate, MistralLarge3ForCausalLM)
+from tensorrt_llm._torch.models.modeling_multimodal_mixin import (
+    MultimodalEncoderOutput, MultimodalModelMixin, PreparedLlmInputs)
 from tensorrt_llm._torch.models.modeling_multimodal_utils import (
-    _MULTIMODAL_ENV_NAME, _is_disagg, find_input_mm_embeds, fuse_input_embeds,
-    get_multimodal_embeddings)
+    _MULTIMODAL_ENV_NAME, _is_disagg)
 from tensorrt_llm._torch.models.modeling_utils import (DecoderModel,
                                                        DecoderModelForCausalLM,
                                                        _load_weights_impl,
@@ -548,7 +549,7 @@ class MistralCommonInputProcessor(Mistral3InputProcessor):
         placeholder_placement=MultimodalPlaceholderPlacement.BEFORE_TEXT,
         content_format=ContentFormat.STRING,
     ))
-class Mistral3VLM(PreTrainedModel):
+class Mistral3VLM(MultimodalModelMixin, PreTrainedModel):
     """Mistral3VLM implementation for TRTLLM.
 
     NOTE: for the time being, image tokens are only placed after the text (see
@@ -662,6 +663,18 @@ class Mistral3VLM(PreTrainedModel):
         )
 
     @property
+    def language_model(self) -> torch.nn.Module:
+        return self.llm
+
+    @property
+    def multimodal_token_ids(self) -> torch.Tensor:
+        return self._image_token_ids
+
+    @property
+    def text_embedding_layer(self) -> torch.nn.Module:
+        return self.llm.model.embed_tokens
+
+    @property
     def draft_config(self):
         return self.llm.draft_config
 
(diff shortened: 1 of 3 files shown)


tensorrt_llm/_torch/models/modeling_mistral.py
```diff
@@ -680,50 +693,93 @@ def vocab_size_padded(self) -> int:
     def infer_max_seq_len(self) -> int:
         return self.llm.infer_max_seq_len()
 
+    def encode_multimodal_inputs(
+        self,
+        multimodal_params: Sequence[MultimodalParams],
+        **encoder_kwargs: Any,
+    ) -> MultimodalEncoderOutput:
+        mm_embeds = self._vision_forward(list(multimodal_params))
+        if len(mm_embeds) != 1:
```
2ez4bz, line 702: One idea I had was to have the `MultimodalEncoderMixin` support kwargs at class definition time, e.g.:

```python
class Foo(MultimodalEncoderMixin, ..., supports_chunked_prefill=True):
    ...
```

in which case we can have explicit checks in the mixin itself for whether `mm_embeds` is a single tensor rather than in each model's `encode_multimodal_inputs`. Curious to get other people's thoughts on this.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+    - Special multimodal tokens occupy token positions but do not have rows in
+      this tensor.
+    """
+
+    embeddings: torch.Tensor
+    cacheable: bool = True
+
+
+@dataclass(frozen=True)
+class MultimodalForwardInputs:
```
yechank-nvidia: Seems it is for LLM, right? Can we re-name as `MultimodalLLMForwardInputs`? Cause when I first saw it, I thought it was the input to the Multimodal whole model(encoder + LLM) part.

tensorrt_llm/_torch/models/modeling_mistral.py
yechank-nvidia: It would also be good to re-unite all the functionalities.(e.g. support_disaggregated).

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+
+from .modeling_multimodal_utils import (
+    find_input_mm_embeds,
+    fuse_input_embeds,
+    get_multimodal_embeddings,
+)
+
+
+@dataclass(frozen=True)
+class MultimodalEncoderOutput:
```
yechank-nvidia, line 33: I think this would be too much, but do you think we can unify the load weights logic for encoder?

The main reason is that, we are not using the meta tensor for Encoder, so it gives us warning but some users think that this was error.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+            text_token_indices=forward_kwargs.get("text_token_indices"),
+            mm_token_indices=forward_kwargs.get("mm_token_indices"),
+        )
+        return MultimodalForwardInputs(
+            input_ids=fused_input_ids,
+            inputs_embeds=inputs_embeds,
+            extra_embeds=fused_extra_embeds,
+        )
+
+    def _get_or_encode_multimodal_embeddings(
```
yechank-nvidia, line 224: Can you add some function description that this function gets cached embedding or doing embedding forward on the docstring?

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+            multimodal_params=list(multimodal_params),
+            encoder_kwargs=encoder_kwargs,
+        )
+        primary = self._require_primary_embedding(embeddings)
+        # Validate post-gather so cached-only paths (KV reuse, all-cached chunked
+        # prefill) are also checked, not just paths that ran the encoder.
+        self._validate_primary_embedding_rows(primary, multimodal_params)
+        return MultimodalEncoderOutput(embeddings=primary, cacheable=True)
+
+    def _find_active_multimodal_embeddings(
```
yechank-nvidia, line 257: Can I ask why this wrapper needed?

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+        this method (`get_multimodal_embeddings`, `find_input_mm_embeds`,
+        `fuse_input_embeds`) operate on the returned list and therefore see
+        only `has_content()` params. Models overriding this hook must
+        preserve that invariant.
+        """
+        return [
+            param for param in list(multimodal_params)[:num_context_requests] if param.has_content()
+        ]
+
+    def after_full_multimodal_embeddings(
```
yechank-nvidia, line 117: Also, not sure what this function is doing.

tensorrt_llm/_torch/models/modeling_mistral.py
moraxu: I don't think I've come across class definition kwargs in this codebase yet but definitely a good idea

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+    fuse_input_embeds,
+    get_multimodal_embeddings,
+)
+
+
+@dataclass(frozen=True)
+class MultimodalEncoderOutput:
+    """Output produced by a model-owned multimodal encoder hook.
+
+    Stage 1 supports a strict primary tensor output:
```
moraxu: Stage 1 of this jira?

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+        self,
+        multimodal_params: Sequence[MultimodalParams],
+        **encoder_kwargs: Any,
+    ) -> MultimodalEncoderOutput:
+        """Run model-specific multimodal encoder work."""
+        raise NotImplementedError
+
+    @property
+    def multimodal_token_ids(self) -> Optional[Sequence[int] | torch.Tensor]:
+        """Return token ids whose text embeddings are replaced by MM embeds.
```
moraxu: Looking at the definition in `Mistral3VLM`, this refers to **placeholder** token ids? Might be worth mentioning in the comment/method name?

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+        *,
+        input_ids: torch.Tensor,
+        multimodal_params: Sequence[MultimodalParams],
+        encoder_output: MultimodalEncoderOutput,
+        **forward_kwargs: Any,
+    ) -> tuple[torch.Tensor, MultimodalEncoderOutput]:
+        """Optional hook after cached/full request embeddings are available."""
+        return input_ids, encoder_output
+
+    def after_active_multimodal_embeddings(
```
moraxu, line 133: Might be worth including an example of when this is needed in the comments

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
```diff
@@ -0,0 +1,342 @@
...
+        into `fuse_input_embeds`; only inputs the helper actually consumes
+        are surfaced here. Models needing to bypass token filtering should
+        pass pre-computed `text_token_indices`/`mm_token_indices`.
+        """
+        if mm_token_ids is not None and not isinstance(mm_token_ids, torch.Tensor):
+            mm_token_ids = torch.tensor(
+                list(mm_token_ids), dtype=input_ids.dtype, device=input_ids.device
+            )
+
+        if text_token_indices is not None and mm_token_indices is not None:
```
moraxu: I haven't seen this chunk of code in `Mistral3VLM`, does this PR fix an actual bug?

moraxu (commented):
I'd add two clarifications to `Flowchart for new mixin`:
1. `Encoder work needed` --> `No` - when is that the case?
2. `Any active MM rows?` --> `No` - when is that the case?

tensorrt_llm/_torch/models/modeling_mistral.py
2ez4bz: @yechank-nvidia @moraxu am I interpreting it correctly that you're generally onboard with the idea?

If so, I agree with keeping the broader capability-flag work out of this PR. We can follow up on a more general capability surface for things like chunked prefill and `support_disaggregated` once another model or two have adopted the mixin.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: Good idea; filed TRTLLM-12555 to look into this.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: Good point. I will rename this to `PreparedLlmInputs`: shorter than `MultimodalLLMForwardInputs` (which I found to be a mouthful 😅 ), but still clear that this is the prepared input bundle passed to the language model rather than the whole multimodal model input. WDYT?

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: Done. I will expand the docstring to make the "cache-or-encode" behavior explicit and mention the primary tensor validation.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: For now it is intentionally just an adapter around `find_input_mm_embeds,` but keeping this as a named internal stage lets `prepare_multimodal_inputs` own the full sequence while the internals can be replaced later.

The model-specific escape hatches remain the public hooks around it:

* `after_full_multimodal_embeddings` for pre-slicing adjustments such as Nemotron Nano EVS.
* `after_active_multimodal_embeddings` for post-slicing transforms such as Qwen3 deepstack splitting. I will make the docstring say that more clearly.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: This is the pre-active-slicing hook: cache lookup / encoder execution has finished, but we have not called `find_input_mm_embeds` yet.

I will expand the docstring to allude to this. The concrete case I had in mind is Nemotron Nano EVS-style behavior, where the model may need to rewrite `input_ids` based on encoder-side token counts before active-row slicing.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: This came from the staged refactor plan, but that wording should not leak into the public docstring. I will replace it with self-contained contract wording.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: Right; these are placeholder token IDs in `input_ids`. I will keep the hook name aligned with the existing `mm_token_ids` helper argument, but update the docstring to explicitly call them placeholder tokens.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: Good point. I will keep the public docstring generic and add a short code comment below it with Qwen3-VL deepstack as the concrete future-migration example.

tensorrt_llm/_torch/models/modeling_multimodal_mixin.py
2ez4bz: Good catch. This was codex being an eager beaver and fixing a bug that doesn't exist. Will back out.

2ez4bz:
> I'd add two clarifications to `Flowchart for new mixin`:
> 
> 1. `Encoder work needed` --> `No` - when is that the case?
> 2. `Any active MM rows?` --> `No` - when is that the case?

@moraxu Done (updated PR description). Also color coded the flowchart to better illustrate what the mixin provides, what the model has to implement, and what it can optionally implement.

moraxu (approved):

yechank-nvidia (approved):
LGTM.

PR #13872, opened 2026-05-08
[TRTLLM-12670][feat] add /start_profile and /stop_profile endpoints to trtllm…

JunyiXu-nv:
…-serve

Expose iteration-scoped runtime profiling over HTTP on the OpenAI-compatible trtllm-serve endpoint, mirroring SGLang's developer-profiling interface.

- POST /start_profile accepts output_dir, num_steps, start_step, activities (any of "CPU", "GPU", "CUDA_PROFILER"). With "CUDA_PROFILER" only, the server skips torch.profiler entirely so it composes cleanly with `nsys profile -c cudaProfilerApi`.
- POST /stop_profile terminates the active window and flushes the trace.
- PyExecutor.start_profile/stop_profile reuse the existing profile_start_iters / profile_stop_iters / cudaProfilerStart-Stop plumbing so env-var-based profiling behavior is preserved.
- _profiler() now builds torch.profiler lazily on the first start iteration so runtime configuration takes precedence over the TLLM_TORCH_PROFILE_TRACE env var and consecutive start/stop windows work.
- GenerationExecutor, BaseWorker, and GenerationExecutorRpcProxy proxy the new calls into the worker process; non-PyTorch backends fall through to a warning.
- LLM.start_profile / stop_profile added as @set_api_status("prototype").
- Smoke test covers default values, explicit overrides, and the stop path via mocked generator (no GPU required).

## Dev Engineer Review

- Added `POST /start_profile` and `POST /stop_profile` to `trtllm-serve`.
- Added runtime profiling across LLMs, executors, workers, proxies, and `PyExecutor`.
- Added iteration-scoped scheduling, lazy profiler creation, multi-rank traces, activity selection, and consecutive profiling windows.
- Added input validation for `CPU`, `GPU`, and `CUDA_PROFILER`.
- Added proxy-worker acknowledgements, timeout handling, and shutdown-state protection.
- Added unsupported-backend warnings and profiling documentation.
- Test-list entries use valid paths and cover endpoint smoke tests, executor profiling, and proxy synchronization.
- CI results remain mixed. PR_Github `#49393` succeeded, but multiple main pipelines failed.

## QA Engineer Review

Added tests cover:

- OpenAI server defaults, custom arguments, backend errors, and event-loop responsiveness.
- PyExecutor scheduling, cancellation, validation, and clean failure state.
- Proxy acknowledgements, error propagation, timeouts, serialization, thread ownership, and shutdown races.
- `test_trtllm_serve_profile_example`.

The endpoint smoke test, `test_profile_endpoints.py`, and `test_proxy_profile_sync.py` are listed in `tests/integration/test_lists/test-db/l0_a10.yml`.

Verdict: sufficient.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/profiling.py
+++ b/tensorrt_llm/_torch/pyexecutor/profiling.py
@@ -1,0 +1,809 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Runtime profiling controller for ``PyExecutor``.
+
+The ``PyExecutorProfileManager`` owns the entire profile-state machine
+that backs the ``trtllm-serve`` ``/start_profile`` / ``/stop_profile``
+HTTP endpoints (and the equivalent programmatic
+``LLM.start_profile`` / ``LLM.stop_profile`` calls):
+
+* ``start_profile`` / ``stop_profile`` — caller-thread entry points
+  that schedule a profile window via the executor's request queue.
+  These run on the FastAPI thread-pool worker (or a user thread for
+  the in-process / RPC paths).
+* ``apply_start_config`` / ``apply_stop_config`` — replay the
+  broadcasted config on every rank so all ranks fire the same start
+  and stop iterations.
+* ``profile_step`` — context manager driving the per-iteration
+  ``torch.profiler`` / ``cudaProfilerStart`` / ``Stop`` calls. Runs
+  on the executor thread because torch.profiler / Kineto require
+  the start and stop calls to happen on the same thread.
+* ``cleanup`` — best-effort tear-down on executor shutdown.
+
+The profile state itself (``profile_start_iters``, ``profile_stop_iters``,
+``_runtime_profile_*`` fields, ``_profile_state_lock``,
+``_profile_enabled``) lives on the ``PyExecutor`` instance so callers
+and unit tests can introspect it directly. The manager holds a weak
+reference back to the executor and operates on those fields. This
+keeps ``py_executor.py`` mostly orchestration / thin forwarding while
+isolating the lifecycle logic in this module for review.
+"""
+
+import os
+import time
+import uuid
+import weakref
+from contextlib import contextmanager
+from typing import Dict, List, Optional
+
+import torch
+
+from tensorrt_llm.logger import logger
+from tensorrt_llm.tools.layer_wise_benchmarks import get_calibrator
+from tensorrt_llm.tools.profiler.host_profile_tools.host_profiler import get_global_profiler
+
+# Environment-variable names used by both the env-var-driven legacy
+# path and the runtime HTTP-endpoint path. They are duplicated as
+# module-level constants here so callers that only need the manager
+# do not have to import from ``py_executor.py``.
+PROFILE_START_STOP_ENV_VAR_NAME = "TLLM_PROFILE_START_STOP"
+PROFILE_TRACE_ENV_VAR_NAME = "TLLM_TORCH_PROFILE_TRACE"
+PROFILE_LOG_RANKS_ENV_VAR_NAME = "TLLM_PROFILE_LOG_RANKS"
+
+
+def load_iteration_indexes(env_var: str):
+    """Parse an iteration-range environment variable.
+
+    Format: ``"start1-stop1,start2-stop2,..."`` for ranges or
+    ``"iter1,iter2,..."`` for single iterations.
+
+    Returns:
+        Two ``frozenset[int]`` -- (start_iters, stop_iters).
+    """
+    spans = os.environ.get(env_var, None)
+    starts, stops = [], []
+
+    if spans:
+        spans = spans.split(",")
+
+        for span in spans:
+            try:
+                if "-" in span:
+                    start, stop = span.strip().split("-")
+                    starts.append(int(start))
+                    stops.append(int(stop))
+                else:
+                    it = int(span.strip())
+                    starts.append(it)
+                    stops.append(it)
+            except ValueError as e:
+                raise ValueError(
+                    f"Cannot parse span in environment variable `{env_var}`: {e}"
+                ) from None
+
+    return frozenset(starts), frozenset(stops)
+
+
+class PyExecutorProfileManager:
+    """Owns the runtime-profiling state machine for ``PyExecutor``.
+
+    Instantiated from ``PyExecutor.__init__`` after the executor has
+    populated ``profile_start_iters`` / ``profile_stop_iters`` /
+    ``_runtime_profile_*`` / ``_profile_state_lock`` /
+    ``_profile_enabled``. The manager keeps a ``weakref`` back to the
+    executor so it can read ``iter_counter``, ``is_warmup``,
+    ``executor_request_queue``, ``dist``, and ``global_rank`` on demand
+    without creating a cycle that defeats garbage collection.
+
+    All methods follow the same threading invariants as the inline
+    code they replace; see the per-method docstrings for details.
+    """
+
+    def __init__(self, executor: "PyExecutor") -> None:  # noqa: F821
+        self._executor_ref = weakref.ref(executor)
+
+    # ---- helpers ---------------------------------------------------
+
+    @property
+    def _executor(self) -> "PyExecutor":  # noqa: F821
+        ex = self._executor_ref()
+        if ex is None:
+            raise RuntimeError(
+                "PyExecutorProfileManager: owning PyExecutor has been "
+                "garbage collected; profile control calls are no longer "
+                "valid."
+            )
+        return ex
+
+    @staticmethod
+    def _activities_from_names(
+        names: Optional[List[str]],
+    ) -> List["torch.profiler.ProfilerActivity"]:
+        if not names:
+            return [
+                torch.profiler.ProfilerActivity.CPU,
+                torch.profiler.ProfilerActivity.CUDA,
+                torch.profiler.ProfilerActivity.XPU,
+            ]
+        mapping = {
+            "CPU": torch.profiler.ProfilerActivity.CPU,
+            "GPU": torch.profiler.ProfilerActivity.CUDA,
+            "CUDA": torch.profiler.ProfilerActivity.CUDA,
+            "XPU": torch.profiler.ProfilerActivity.XPU,
+        }
+        resolved: List[torch.profiler.ProfilerActivity] = []
+        for name in names:
+            key = name.upper()
+            if key == "CUDA_PROFILER":
+                # CUDA_PROFILER controls cudaProfilerStart/Stop only;
+                # it is not a torch.profiler activity.
+                continue
+            if key in mapping and mapping[key] not in resolved:
+                resolved.append(mapping[key])
+        return resolved
+
+    # ---- HTTP-callable entry points --------------------------------
+
+    def start_profile(
+        self,
+        output_dir: Optional[str] = None,
+        num_steps: Optional[int] = None,
+        start_step: int = 0,
+        activities: Optional[List[str]] = None,
+    ) -> None:
+        """Schedule a runtime profile window on every rank.
+
+        Mirrors the env-var driven ``TLLM_PROFILE_START_STOP`` /
+        ``TLLM_TORCH_PROFILE_TRACE`` path but is callable at runtime
+        (e.g. from the ``trtllm-serve /start_profile`` HTTP handler).
+
+        Runs on the caller thread (FastAPI worker thread / user
+        thread). Marks ``_runtime_profile_pending_start_iter`` locally
+        so a concurrent second call is rejected immediately, then
+        broadcasts the config on rank 0 via the executor's request
+        queue so every rank sees the same start.
+        """
+        executor = self._executor
+        # Reject overlapping profile windows. Check locally on the
+        # caller thread so the HTTP handler gets a clean error instead
+        # of the executor thread crashing on ``assert not enabled``
+        # inside ``profile_step``.
+        with executor._profile_state_lock:
+            already_active = executor._profile_enabled
+        if already_active or executor._runtime_profile_pending_start_iter is not None:
+            # Use ``RequestError`` (a ``RuntimeError`` subclass) so the
+            # ``GenerationExecutor`` error monitor treats this as a
+            # per-call rejection rather than a fatal engine error.
+            from tensorrt_llm.executor.utils import RequestError
+
+            raise RequestError(
+                "Profiling is already in progress (or a pending start has "
+                "been scheduled); call stop_profile first before starting "
+                "a new window. (state: _profile_enabled="
+                f"{already_active}, _runtime_profile_pending_start_iter="
+                f"{executor._runtime_profile_pending_start_iter}, "
+                f"profile_start_iters="
+                f"{sorted(executor.profile_start_iters)[:5]}, "
+                f"profile_stop_iters="
+                f"{sorted(executor.profile_stop_iters)[:5]})"
+            )
+
+        # Reject ``num_steps <= 0``. The HTTP layer already rejects
+        # this via the ``StartProfileRequest`` Pydantic schema, but
+        # programmatic callers (``LLM.start_profile``, env-var-driven
+        # paths, unit tests) bypass that schema. With
+        # ``num_steps == 0``, ``stop_iter`` would equal ``start_iter``
+        # in ``apply_start_config`` below, ``profile_step`` would
+        # discard the stop marker as stale, and the profile window
+        # would run forever until an explicit ``stop_profile()``.
+        if num_steps is not None and num_steps <= 0:
+            raise ValueError(
+                f"start_profile: num_steps must be >= 1 if provided "
+                f"(got num_steps={num_steps!r}). Pass num_steps=None to "
+                "run until an explicit stop_profile() call."
+            )
+
+        if activities is None:
+            activities = ["CPU", "GPU"]
+
+        if output_dir is None:
+            # ``/tmp`` is the documented developer default for the
+            # profile trace directory (matches the server docs); the
+            # user-facing override is the
+            # ``TLLM_TORCH_PROFILER_DIR`` env var or the
+            # ``output_dir`` request body field.
+            output_dir = os.environ.get(
+                "TLLM_TORCH_PROFILER_DIR",
+                "/tmp",  # nosec B108
+            )
+
+        # Fail fast if the directory cannot be created. Returning
+        # success here would schedule a profile window that later
+        # blows up in ``export_chrome_trace()`` on the executor
+        # thread, after the HTTP caller has already been told the
+        # request was accepted. ``RequestError`` (a ``RuntimeError``
+        # subclass) routes through the same per-call rejection
+        # plumbing as the "already in progress" guard above, so the
+        # HTTP handler can map it to a 5xx with a meaningful body.
+        try:
+            os.makedirs(output_dir, exist_ok=True)
+        except OSError as e:
+            from tensorrt_llm.executor.utils import RequestError
+
+            raise RequestError(f"Failed to create profile output_dir {output_dir}: {e}") from e
+
+        # Include a unique profile id (monotonic timestamp + short
+        # uuid fragment) so successive start/stop cycles under the
+        # same output_dir don't silently overwrite each other's
+        # traces.
+        profile_id = f"{int(time.time())}-{uuid.uuid4().hex[:8]}"
+
+        # Build the per-request config that every rank consumes. We
+        # do NOT pre-compute start_iter here: each rank must derive
+        # it from its own ``iter_counter`` at the moment the
+        # broadcast is processed (inside
+        # ``_handle_special_queue_items``), so that rank 0 doesn't
+        # get a head start by pre-applying with its caller-thread
+        # ``iter_counter`` while rank 1's executor loop has already
+        # advanced past the start point. All ranks are in sync on
+        # iter_counter at the broadcast point, so they all compute
+        # the same start_iter/stop_iter.
+        profile_config = {
+            "output_dir": output_dir,
+            "profile_id": profile_id,
+            "activities": list(activities),
+            "start_step": int(max(0, start_step)),
+            "num_steps": num_steps,
+        }
+
+        # Mark pending on the caller thread too so a second
+        # concurrent ``start_profile()`` is rejected by the guard
+        # above. The actual ``profile_start_iters`` / trace_path
+        # state is only populated when the broadcast reaches
+        # ``_handle_special_queue_items``. Use a sentinel (0) to
+        # mean "pending but not yet scheduled".
+        executor._runtime_profile_pending_start_iter = 0
+        executor._runtime_profile_activities = list(activities)
+
+        # Broadcast to every rank via the request queue so non-zero
+        # ranks update their own ``profile_start_iters`` in
+        # lockstep. This is also what wakes the idle executor loop
+        # (the PROFILE_START_REQUEST_ID item gets broadcasted via
+        # ``RequestBroadcaster`` inside
+        # ``_fetch_and_enqueue_requests``, then dropped inside
+        # ``_handle_special_queue_items`` — no forward pass runs for
+        # this iteration, so the chrome trace captures only real
+        # user workload).
+        rank = getattr(getattr(executor, "dist", None), "rank", 0)
+        eq = getattr(executor, "executor_request_queue", None)
+        if rank == 0 and eq is not None:
+            try:
+                eq.enqueue_profile_start_request(profile_config)
+            except (RuntimeError, OSError, ValueError, AttributeError) as e:
+                # Roll back the pending markers we just set so a
+                # subsequent ``start_profile()`` is not rejected as
+                # "already pending", and so non-zero ranks (which
+                # never received the broadcast) don't drift out of
+                # sync with rank 0. Then surface the failure to the
+                # HTTP layer rather than silently logging and
+                # continuing.
+                executor._runtime_profile_pending_start_iter = None
+                executor._runtime_profile_activities = None
+                raise RuntimeError(
+                    f"start_profile: failed to enqueue profile-start broadcast item: {e}"
+                ) from e
+
+    def stop_profile(self) -> None:
+        """Schedule the in-progress runtime profile to stop.
+
+        The actual ``torch.profiler.stop()`` +
+        ``export_chrome_trace()`` call happens inside
+        ``profile_step`` on the executor thread, so torch.profiler /
+        Kineto's "stop must run on the same thread that called
+        start" invariant is respected.
+
+        Two extra cases are handled so no profile window is leaked:
+
+        * If ``start_profile()`` was called but the engine has not
+          yet reached the scheduled ``start_iter`` (e.g. the caller
+          stopped before sending any requests), cancel the pending
+          start so profiling does not begin later.
+        * Otherwise schedule a stop iteration. When called via an
+          HTTP handler the server layer is responsible for tickling
+          the executor loop so the stop iteration actually runs even
+          when the engine would otherwise be idle (see
+          ``OpenAIServer.stop_profile``).
+        """
+        executor = self._executor
+        with executor._profile_state_lock:
+            currently_enabled = executor._profile_enabled
+
+        if not currently_enabled and executor._runtime_profile_pending_start_iter is not None:
+            # Pending runtime start has not fired yet; cancel it
+            # cleanly on this rank.  Also broadcast the stop so
+            # non-zero ranks clear their own pending starts —
+            # otherwise a ``start_profile``-then-``stop_profile``
+            # cycle on an idle multi-rank deployment would leave
+            # stale pending starts on the subordinates.
+            self.apply_stop_config()
+            rank = getattr(getattr(executor, "dist", None), "rank", 0)
+            eq = getattr(executor, "executor_request_queue", None)
+            if rank == 0 and eq is not None:
+                try:
+                    eq.enqueue_profile_stop_request()
+                except (RuntimeError, OSError, ValueError, AttributeError) as e:
+                    # Local cancel already happened on this rank,
+                    # but subordinate ranks would still be carrying
+                    # the pending start. Surface the broadcast
+                    # failure so the caller can retry rather than
+                    # silently leaving the cluster out of sync.
+                    raise RuntimeError(
+                        f"stop_profile: failed to enqueue profile-stop broadcast item: {e}"
+                    ) from e
+            return
+
+        # Apply locally so diagnostic state (profile_stop_iters,
+        # pending_stop_iter) is visible on the caller thread
+        # immediately.
+        self.apply_stop_config()
+
+        # Broadcast to every rank so non-zero ranks add the same
+        # stop_iter to their own profile_stop_iters. Also wakes the
+        # idle executor loop on rank 0; the broadcast then
+        # propagates the wake to all ranks inside
+        # ``_fetch_and_enqueue_requests``.
+        rank = getattr(getattr(executor, "dist", None), "rank", 0)
+        eq = getattr(executor, "executor_request_queue", None)
+        if rank == 0 and eq is not None:
+            try:
+                eq.enqueue_profile_stop_request()
+            except (RuntimeError, OSError, ValueError, AttributeError) as e:
+                # Subordinate ranks won't see the stop. Bail out
+                # with a real error rather than silently waiting on
+                # a stop that will never fire.
+                raise RuntimeError(
+                    f"stop_profile: failed to enqueue profile-stop broadcast item: {e}"
+                ) from e
+
+        # Wait (with timeout) for the stop to actually fire so that
+        # by the time stop_profile() returns the chrome trace has
+        # been exported to disk. Without this wait, callers that hit
+        # the HTTP /stop_profile endpoint and then immediately try
+        # to read the trace file would see an empty directory.
+        deadline = time.monotonic() + 30.0
+        while time.monotonic() < deadline:
+            with executor._profile_state_lock:
+                if not executor._profile_enabled:
+                    return
+            time.sleep(0.005)
+        logger.warning(
+            "stop_profile: executor loop did not fire the scheduled stop "
+            "within 30s; the trace may not be fully flushed yet."
+        )
+
+    # ---- broadcast handlers (called on every rank) -----------------
+
+    def apply_start_config(self, config: Dict) -> None:
+        """Apply a broadcasted profile-start config to this rank.
+
+        Called from two places: directly on the caller thread from
+        ``start_profile()`` (so
+        ``_runtime_profile_pending_start_iter`` becomes visible for
+        re-entrancy checks) AND from ``_handle_special_queue_items``
+        when the broadcasted ``PROFILE_START_REQUEST_ID`` item
+        arrives. The operation is idempotent: the second call on
+        rank 0 sees the same state and is a no-op.
+        """
+        executor = self._executor
+        activities = config.get("activities") or ["CPU", "GPU"]
+        # ``/tmp`` matches the documented developer default; see
+        # ``start_profile`` for the rationale.
+        output_dir = config.get("output_dir") or "/tmp"  # nosec B108
+        profile_id = config.get("profile_id") or f"{int(time.time())}-{uuid.uuid4().hex[:8]}"
+        # Prefer the legacy absolute ``start_iter`` form (older unit
+        # tests and env-var driven paths set this directly). For the
+        # broadcast path the caller sends ``start_step`` instead so
+        # every rank computes its own ``start_iter`` from its
+        # current ``iter_counter`` — this avoids a race where rank
+        # 0's local apply sets a past start_iter that rank 1 never
+        # reaches in time to fire.
+        if "start_iter" in config:
+            start_iter = int(config["start_iter"])
+        else:
+            start_step = int(config.get("start_step", 0))
+            current_iter = int(getattr(executor, "iter_counter", 0))
+            start_iter = current_iter + 1 + start_step
+        num_steps = config.get("num_steps")
+        stop_iter = config.get("stop_iter")
+        if stop_iter is None and num_steps is not None:
+            stop_iter = start_iter + int(num_steps)
+
+        # Rank-specific trace path so every rank writes its own file
+        # when running with TP/PP > 1.
+        trace_filename = f"trtllm-trace-{profile_id}-rank-{executor.global_rank}.json"
+        trace_path = os.path.join(output_dir, trace_filename)
+
+        # First apply on this rank? If the start_iter is already in
+        # profile_start_iters it means either rank 0's local apply
+        # ran (before the broadcast echo-back) or a prior broadcast
+        # already applied — in both cases the pending marker
+        # belongs to the original apply, and re-setting it here
+        # would overwrite a subsequent clear by ``profile_step()``
+        # and leak stale state.
+        is_first_apply = start_iter not in executor.profile_start_iters
+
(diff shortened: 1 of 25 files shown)


xinhe-nv (approved):

Superjomn (approved):
LGTM on the llmapi changes.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1103,13 +1227,365 @@ def profile_step():
...
+                f"start_profile: num_steps must be >= 1 if provided "
+                f"(got num_steps={num_steps!r}). Pass num_steps=None to "
+                "run until an explicit stop_profile() call.")
+
+        if activities is None:
+            activities = ["CPU", "GPU"]
+
+        if output_dir is None:
+            # ``/tmp`` is the documented developer default for the profile
+            # trace directory (matches sglang and the server docs); the
```
Tabrizian: perhaps remove mentions of sglang here and everywhere else?

JunyiXu-nv: Thanks for reviewing! Removed all places mentioning sgl.

Tabrizian (approved):

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1150,13 +1274,365 @@ def profile_step():
...
+                    torch_profiler.export_chrome_trace(active_torch_trace_path)
+                    logger.info(f"Profiling stopped at iteration "
+                                f"{self.iter_counter}, "
+                                f"trace saved to {active_torch_trace_path}")
                 torch.cuda.cudart().cudaProfilerStop()
                 calibrator.stop()
+                with self._profile_state_lock:
+                    self._profile_enabled = False
+
+    def start_profile(self,
```
QiJune, line 1695: This adds quite a bit of profiling state-machine logic directly into py_executor.py. Could we move it into a separate profiling.py helper/controller instead?

A separate PyExecutorProfileManager that owns start_profile(), stop_profile(), apply_start_config(), apply_stop_config(), profile_step(), and cleanup would keep py_executor.py mostly as orchestration/thin forwarding, and make the profiling lifecycle much easier to review and maintain.

tensorrt_llm/serve/openai_server.py
```diff
@@ -1850,6 +1859,107 @@ async def openai_responses_delete_response(
...
+        The underlying ``GenerationExecutor.start_profile`` call may
+        block (it waits for the worker subprocess to ack on the
+        IPC-proxy path, up to ~60s). We run it on a worker thread via
+        ``asyncio.to_thread`` so the FastAPI event loop stays
+        responsive to other endpoints during that wait.
+        """
+        if request is None:
+            request = StartProfileRequest()
+        try:
+            await asyncio.to_thread(
```
QiJune, line 3037: One concern with the current /start_profile and /stop_profile implementation is the threading model around the IPC queues.

The HTTP handlers call the backend through asyncio.to_thread() so the FastAPI event loop is not blocked while waiting for the worker ack. However, in the IPC proxy path, GenerationExecutorProxy.start_profile()/stop_profile() then directly access ZMQ-backed IpcQueue objects from that thread-pool thread:

```python
  - request_queue.put(StartProfileRequest(...)) /request_queue.put(StopProfileRequest())
  - profile_ack_queue.get(...)
```

to writes. Even read-only access to the same socket from different threads is unsafe. Also, asyncio.to_thread() does not guarantee that repeated calls will run on the same worker thread, so concurrent or back-to-back profile calls may touch the same sockets from different threads.

Could we avoid doing ZMQ socket operations inside the to_thread() callback? I think there are two possible directions:

1. Make the endpoints scheduled/best-effort controls: enqueue the control request on the existing queue owner path and return once the request is accepted/scheduled. In that case the docs/response should not promise that /stop_profile returns only after the trace is on disk.

2. Keep the synchronous “trace is flushed before /stop_profile returns” contract, but introduce a safe ownership boundary: for example, a dedicated control/queue-owner thread that is the only thread touching the ZMQ queues, while the HTTP handler waits on a thread-safe Future/Queue for the ack.

The key requirement is that all operations on a given ZMQ-backed IpcQueue stay on one owning thread.

tensorrt_llm/_torch/pyexecutor/py_executor.py
JunyiXu-nv: Thanks for reviewing! I've extracted them to a standalone file. Could you help take another look?

tensorrt_llm/serve/openai_server.py
JunyiXu-nv: Thanks for the review! Pinned all profile-control IPC to a dedicated single-worker ThreadPoolExecutor, `start_profile/stop_profile` now `.submit()` onto it so every ZMQ op on `request_queue`/`profile_ack_queue` runs on one owning thread, and `_wait_profile_ack` still blocks the HTTP handler so the "trace flushed before `/stop_profile` returns" contract is preserved.

tensorrt_llm/executor/proxy.py
```diff
@@ -529,6 +567,136 @@ def abort_request(self, request_id: int) -> None:
...
+        ``AttributeError`` on ``None``. Raise ``RuntimeError`` instead — the
+        HTTP handlers already map it to a controlled response.
+        """
+        executor = self._profile_control_executor
+        if executor is None:
+            raise RuntimeError(
+                f"{kind}_profile: the executor proxy is shutting down or "
+                "already shut down; profile control is unavailable.")
+        executor.submit(self._profile_control_call, kind, request,
+                        timeout).result()
```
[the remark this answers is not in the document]
JunyiXu-nv: Good catch — confirmed and fixed in 35f3da7122.

`pre_shutdown()` sets `doing_shutdown = True` (proxy.py:863) and sends the worker quit sentinel (proxy.py:889) before `shutdown()` clears `_profile_control_executor`, so a profile call landing in that window submitted successfully and then blocked for the whole `_START_PROFILE_ACK_TIMEOUT_S` against workers that were already exiting.

`_submit_profile_control` now checks `doing_shutdown` first and raises `RuntimeError` immediately. Added `test_profile_control_during_pre_shutdown_raises_runtime_error[start|stop]`, which asserts both that the call is rejected and that it returns in under a second rather than after the ack timeout, and that neither ZMQ queue is touched.

One thing worth flagging that the suggested diff would have missed: `_bare_proxy()` in the test helper constructs the proxy via `__new__`, so `doing_shutdown` does not exist there. Adding the check alone would have broken every existing test in that file with `AttributeError`. The helper now sets `doing_shutdown = False` to match what `GenerationExecutor.__init__` does in production.

tensorrt_llm/_torch/pyexecutor/profiling.py
```diff
@@ -0,0 +1,809 @@
...
+                "Profiling is already in progress (or a pending start has "
+                "been scheduled); call stop_profile first before starting "
+                "a new window. (state: _profile_enabled="
+                f"{already_active}, _runtime_profile_pending_start_iter="
+                f"{executor._runtime_profile_pending_start_iter}, "
+                f"profile_start_iters="
+                f"{sorted(executor.profile_start_iters)[:5]}, "
+                f"profile_stop_iters="
+                f"{sorted(executor.profile_stop_iters)[:5]})"
+            )
```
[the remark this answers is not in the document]
JunyiXu-nv: Confirmed and fixed in f142580f86.

The check-then-set really was non-atomic, and it reproduces: a harness firing 8 concurrent `start_profile()` calls admits **2** callers and enqueues **2** profile-start broadcasts within the first couple of trials against the pre-fix code.

`start_profile()` now does the re-entrancy check and claims the pending sentinel inside one critical section. Validation and broadcast moved into `_prepare_and_broadcast_start()` so all three failure paths (`num_steps <= 0`, `output_dir` creation, enqueue failure) roll the claim back through a single `finally` under the same lock.

The other sites you listed are covered too:
- `apply_stop_config()` now makes the whole cancel-vs-schedule decision under the lock and returns which path it took. `stop_profile()` consumes that return value instead of repeating the check outside the lock, which also removed the duplicated broadcast block.
- `apply_start_config()` publishes the whole window (pending markers, `_runtime_profile_*` fields, and the iter sets) under the lock.
- `profile_step_fn` flips the pending markers in the same critical section as `_profile_enabled`, so a caller thread can no longer observe "not enabled but still pending" and reject a legitimate restart.

Regression tests in `tests/unittest/_torch/executor/test_profile_endpoints.py`: `test_concurrent_start_profile_admits_exactly_one_caller` (8 threads released off a barrier, 50 trials, asserts exactly one admitted caller and exactly one broadcast), `test_start_profile_rolls_back_claim_when_broadcast_fails`, and `test_apply_profile_stop_config_reports_cancel_vs_schedule`.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -4830,14 +4792,15 @@ def _executor_loop_overlap(self):
...
+                            if self.guided_decoder is not None:
+                                # add_batch must be called again to have updated new tokens.
+                                self.guided_decoder.add_batch(scheduled_batch)
+                                guided_decoder_failed_requests = self.guided_decoder.execute(
+                                    batch_outputs['logits'])
 
-                        sample_state = self._sample_async(
-                            scheduled_batch, batch_outputs)
+                            sample_state = self._sample_async(
+                                scheduled_batch, batch_outputs)
```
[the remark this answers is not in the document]
JunyiXu-nv: Good catch — fixed in f142580f86.

The two scopes genuinely are non-nested, with `_update_requests` for the previous batch running between them, so the overlap loop can't wrap the iteration in one scope. I took the disambiguation option: `_build_step_scope_label` gained an optional `phase` argument and the sampling scope now emits `step[...] sample`. Exactly one bare `step[...]` scope marks each iteration again, matching the non-overlap loop and the PP loop.

Also documented the split in the `_build_step_scope_label` docstring and in both user-facing profiling docs (`trtllm-serve.rst` and `run-benchmark-with-trtllm-serve.md`), noting that `sample`-suffixed labels must not be counted as iteration boundaries.

tests/unittest/llmapi/apps/_test_trtllm_serve_profile.py
```diff
@@ -0,0 +1,207 @@
...
+    response = asyncio.run(server.start_profile(StartProfileRequest()))
+
+    assert response.status_code == 200
+    assert _extract_json(response) == {"message": "Profiling started"}
+    server.generator.start_profile.assert_called_once_with(
+        output_dir=None,
+        num_steps=None,
+        start_step=0,
+        activities=["CPU", "GPU"],
+    )
```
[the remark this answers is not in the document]
JunyiXu-nv: Thanks — fixed in f142580f86.

- **Stale filename**: correct, `test_profile_endpoints_bugs.py` doesn't exist and neither does `test_http_handler_rejects_double_start_with_409`. The reference now points at `tests/unittest/_torch/executor/test_profile_endpoints.py` for the backend-side rejection.
- **409 path**: rather than only fixing the pointer, the HTTP-side 409 mapping is now covered here directly — `test_start_profile_double_start_returns_409`, parametrized over both message forms the handler matches (`already in progress` and `pending`).
- **`OSError` / `ValueError` / `TimeoutError` branches**: covered by `test_start_profile_non_runtime_errors_return_500` and `test_stop_profile_non_runtime_errors_return_500` (the stop handler doesn't catch `ValueError`, so it's parametrized over the two types it does catch). Added `test_start_profile_unexpected_error_propagates` / `test_stop_profile_unexpected_error_propagates` to pin the deliberate choice that unlisted exceptions reach FastAPI instead of being swallowed as a generic 500.
- **stop_profile success body**: the existing `test_stop_profile` already asserts the complete body via `== {"message": "Profiling stopped"}`; the new 500-path tests assert their full bodies the same way.
- **CI registration**: verified. `_test_trtllm_serve_profile.py` runs through the `test_e2e.py::test_trtllm_serve_profile_example` wrapper, and that wrapper, `unittest/_torch/executor/test_profile_endpoints.py`, and `unittest/executor/test_proxy_profile_sync.py` are all listed in `tests/integration/test_lists/test-db/l0_a10.yml`.

JunyiXu-nv:
`test_disaggregated_perf_metrics[TinyLlama-1.1B-Chat-v1.0]` has now failed twice (pipelines #55911, #55963), so recording why I do not think it is caused by this PR.

Both failures are the same assertion in `tests/test_common/perf_metrics_utils.py:162`:

```
assert disagg_arrival <= ctx_server_arrival + time_tolerance_seconds
```

Observed overshoot: **29 ms** (#55911) and **7 ms** (#55963), against a `time_tolerance_seconds` default of **5 ms**.

Why this cannot be caused by this PR:

1. The diff touches nothing in arrival-time recording — no changes to `disagg_server_arrival_time`, `server_arrival_time`, or any timestamp path.
2. The failure direction is wrong for a latency regression. `disagg_arrival` is stamped at the disagg router; `ctx_server_arrival` is stamped at the ctx server when it receives the *forwarded* request. For this assert to fail, the downstream stamp must be **earlier** than the upstream one, which is only possible through clock disagreement between the two processes. Any extra latency introduced by this PR would push `ctx_server_arrival` **later**, making the assertion *more* likely to pass.

The tolerance's own comment says it exists to absorb "a local network ping time ... by taking consideration of the error of NTP (1/2 ping time)". 5 ms is tight for cross-process wall-clock comparison on a shared A10 runner, which matches both observed overshoots.

The test is not currently waived on main. cc test owners — this looks like a tolerance that needs widening rather than a product bug.

JunyiXu-nv:
**Correcting my previous comment.** I claimed this failure "cannot be caused by this PR". That was wrong, and the data does not support it.

`test_disaggregated_perf_metrics[TinyLlama-1.1B-Chat-v1.0]` has now failed **3/3** runs here (#55911, #55963, #55987), with overshoots of 29 ms, 7 ms and 46 ms against the 5 ms tolerance. Meanwhile the same `A10-PyTorch-3` stage passed on two other PRs run in the same window:

| run | A10-PyTorch-3 |
| --- | --- |
| #17464 pipeline #55909 | SUCCESS |
| #17202 pipeline #55910 | SUCCESS |
| **#13872 pipeline #55911** | **FAILURE** |
| **#13872 pipeline #55963** | **FAILURE** |
| **#13872 pipeline #55987** | **FAILURE** |

My earlier argument was that a latency regression could only push `ctx_server_arrival` later and therefore make the assert *more* likely to pass. That reasoning only holds if the timestamps cannot move relative to each other for other reasons, which is not something I had established.

A more likely mechanism, which is specific to this PR but **not** to its product code: this PR adds three test files to `l0_a10.yml`, one of them at line 35 — *above* `test_disaggregated_perf_metrics` at line 65:

- `unittest/_torch/executor/test_profile_endpoints.py`
- `test_e2e.py::test_trtllm_serve_profile_example`
- `unittest/executor/test_proxy_profile_sync.py`

Inserting an entry above it shifts every subsequent entry, changing shard assignment when the list is split across `A10-PyTorch-1/2/3` and changing which tests it runs alongside. A 5 ms tolerance on cross-process `get_steady_clock_now_in_seconds()` timestamps is fragile enough for that to matter on a shared A10 runner.

What still argues against a product regression: the PR touches nothing on the arrival-timestamp path — `openai_disagg_server.py` and `responses_utils.py` are both unmodified; the serve-side diff is confined to `openai_server.py` and `openai_protocol.py`.

So: correlated with this PR, but most likely through test-list sharding rather than the profiling code. Worth confirming which shard the test lands in before/after, and whether the tolerance should be widened.

JunyiXu-nv:
**Heads-up: the current head (`97a37b5`) carries a temporary diagnostic commit that must not be merged.** It will be reverted as soon as the CI data is collected.

### Why

On this PR the disagg `/steady_clock_offset` handshake takes **30–150 ms** in CI, against **2–4 ms** for every other PR — including PR #17586 measured on the *same physical node* (`a4u8g-0059.ipp2a2`), on the same day, against the same base. That inflated one-way delay is fed straight into the NTP estimator (`offset = (d_f - d_b) / 2`); since both processes are co-located and share `CLOCK_MONOTONIC`, the true offset is exactly `0`, so the entire estimate is error. That is what breaks the `disagg_arrival <= ctx_server_arrival` causality assertion by the observed 21–33 ms.

### What has already been excluded

In the affected process (the `trtllm-serve` FastAPI + proxy process):

- the two new routes are dormant;
- the profile-control `ThreadPoolExecutor` spawns no thread until first submit;
- the executor-loop changes run in a **separate** process under the multi-process IPC executor CI uses (`Orchestrator is creating IPC executor`, no single-process worker), so they cannot contend for its GIL.

Ordering within the shard, CPU starvation, node/cluster, coverage instrumentation, and clock-epoch effects were all ruled out from CI artifacts.

### What this commit tests

The profile ack queue is the only remaining runtime-live delta — it brings its own `zmq.Context` (hence an extra I/O thread), a bound loopback TCP socket, and an HMAC key. CI logs confirm it is live: this PR generates **8** HMAC keys per run where other PRs generate **6**.

This commit drops it (`TLLM_ABLATE_PROFILE_ACK_QUEUE=0` restores it) and logs every clock probe. The warm-up probe is a fresh-session first request — exactly what the pre-fix single-sample code measured — so its delay is directly comparable to the historical 30–150 ms, and the per-probe series separates a one-off connection-setup cost from a persistent stall.

### Expected failures on this commit

The profile ack-path tests (`test_proxy_profile_sync.py`, plus the ack-dependent cases in `test_profile_endpoints.py` / `_test_trtllm_serve_profile.py`). `test_disaggregated_perf_metrics` runs earlier in the same A10 shard, so its measurement is unaffected. `--disable-fail-fast` is used so these deliberate failures do not cost the run.

### Rebase note

This head was rebased onto current main (51 commits). The profiler refactor conflicted with main's change that guards the per-iteration `torch.cuda.Event` creation/record behind `should_capture_timing`; that change has been ported into `profiling.py`.

JunyiXu-nv:
Re-pushed as `1d40908` — two unrelated issues on the previous head:

1. **`Pre-commit Check` failed** (`ruff-format`, `profiling.py`). Mine: the line I ported during the rebase was not ruff-format canonical. Fixed **in the feature commit**, not the temporary one, so the formatting stays correct after the diagnostic commit is reverted. Full `pre-commit run` over all 26 changed files now passes with no files modified.
2. **`blossom-ci` failed with `Failed to create new client` (exit 255)** — infrastructure, not this PR. Authorization, LDAP verification, and the vulnerability scan all passed; only "Start ci job" failed, unable to reach the CI server. Retrying.

The temporary ablation commit and its rationale are unchanged from the previous comment.

JunyiXu-nv:
## Ablation result: the profile ack queue is the trigger

Pipeline #56953 ran with the queue removed, on node `a4u8g-0054.ipp2a2` — the **same physical node** as build #54585, which ran this PR *with* the queue:

| | node | `profile_ack_queue` | handshake delay |
|---|---|---|---|
| build 54585 | `a4u8g-0054.ipp2a2` | **present** | **61 / 35 ms** |
| pipeline 56953 | `a4u8g-0054.ipp2a2` | **absent** | **3.6 / 2.7 ms** |

Same PR, same node, same version (1.3.0rc25). Ablation verified live: HMAC-key count dropped 8 → 6, `profile_ack` absent from the log.

```
ABLATION-PROBE :46619 warmup  delay=3.62ms offset=1.21ms
ABLATION-PROBE :46619 probe#1 delay=2.12ms offset=0.17ms
ABLATION-PROBE :45969 warmup  delay=2.67ms offset=0.51ms
ABLATION-PROBE :45969 probe#1 delay=1.56ms offset=0.32ms
```

The **warm-up** probe matters most: it is a fresh-session first request, i.e. exactly what the pre-fix single-sample code measured, so it is directly comparable to the historical 30–150 ms. It dropped too — this is not an artifact of the fix's warm-up probe hiding a connection-setup cost. `offset` fell from ~74 ms to 0.17/0.32 ms, the causality assertion did not fire, and the test passed on the first attempt (build 56184 needed three).

## Next: which process pays

Not yet known. The queue exists in both the `trtllm-serve` process (binds) and the executor worker process (connects). The worker side can only exist if the proxy publishes an address, so the separable modes are `off` (measured fast), `proxy-only` (this run), `full` (measured slow). proxy-only slow ⇒ cost is in `trtllm-serve`; proxy-only fast ⇒ cost is in the executor worker.

This run also adds an event-loop lag monitor to both processes, to separate loop starvation from time spent in the kernel/network leg.

Head is now `77c6493`. Profile ack-path tests are expected to fail; the ack wait is short-circuited outside `full` mode so they fail fast rather than burning their 60s/35s timeouts.

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
## Localization result: the cost is in the `trtllm-serve` process, and it is loop starvation

Pipeline #56981, mode `proxy-only` (proxy binds the queue, worker never connects — confirmed by `PROFILE-ACK-QUEUE-WORKER created=False`):

| | warm-up delay | serve-process peak loop lag | disagg peak loop lag |
|---|---|---|---|
| ctx | **60.2 ms** | **53.0 ms** | 6.0 ms |
| gen | **97.2 ms** | **93.0 ms** | 1.0 ms |
| probe #1 (both) | 2.0 / 1.9 ms | 1.0 ms | 1.0 ms |

Three things follow:

1. **proxy-only is slow**, so the cost is in the `trtllm-serve` process — the executor worker side is not involved.
2. **Serve-process loop lag tracks the delay almost exactly** (53 vs 60, 93 vs 97), so this is event-loop starvation, not time in the kernel/network leg. The disagg loop stays at 1–6 ms, ruling that side out.
3. **Only the first probe is slow.** The stall is one-time, not a per-request cost.

## Not yet: what actually blocks the loop

A standalone repro shows that creating the extra `zmq.Context` + PAIR socket + bind + HMAC key costs nothing by itself — 0.5–0.9 ms of loop lag at 3 vs 4 queues. So the queue is not paying this cost directly; it is perturbing something else in the process.

The same repro puts gen-2 GC pauses on a 2M-object heap at 4–272 ms, which is the right order of magnitude. But that is suggestive only: a few extra objects shift the collector's phase randomly, which would not produce the 10-of-10 slow runs actually observed. So the next run measures instead of assuming.

Head `a9eaf1b` adds two probes to the `trtllm-serve` process, because the blocker could be either kind. A GC pause holds the GIL inside C and never yields, so a sampling thread cannot see it — `gc.callbacks` can. A pure-Python blocker does yield every switch interval, so a sampler gets slots during it and can name the code — `gc.callbacks` would say nothing there. Running both covers both cases in one run.

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
/bot run --stage-list "A10-PyTorch-1,A10-PyTorch-2,A10-PyTorch-3"

JunyiXu-nv:
## Root cause confirmed, diagnostics removed, fix in place

Head is now `60d0ac9` — two commits, no diagnostic code.

### The ablation, repeated

The earlier "root cause confirmed" claim rested on a **single** `off` run, which was too thin: if the stall were machine-wide noise, one fast run could just have landed in a quiet moment, and being on the same node as a slow run does not fix that. So `off` was repeated. Pipeline #57064, again on `a4u8g-0054.ipp2a2`:

```
warmup delay = 4.45 / 1.50 ms      probe#1 = 2.40 / 1.07 ms
serve peak-loop-lag = 1.0 ms (all four handshakes)
lag-events = []   slow-io = []     disagg peak-loop-lag = 1.0 ms
```

Not a single lag event over 10 ms anywhere. Tally across 17 runs:

| mode | runs | handshake delay |
|---|---|---|
| `off` | 2 | 3.6/2.7, 4.45/1.50 ms |
| `proxy-only` | 5 | 60/97, 49/26, 186/38, 201/78, 110/65 ms |
| `full` | 10 | 30–150 ms |

Perfect separation. Random machine noise would not sort itself this way.

### Localization

- Cost is in the `trtllm-serve` process: `proxy-only` (proxy binds, worker never connects) reproduces the stall in full.
- It is loop starvation, not network: serve-process loop lag tracks the delay one-for-one (53↔60, 93↔97, 195↔201 ms), and the stall ends 1.0 ms before the handler runs.
- The loop thread is not running: 0.61–0.92 ms of CPU across stalls of 99–195 ms.
- Ruled out along the way: GC (`gc-pauses` always empty), Python code holding the GIL, the executor worker process, ordering within the shard, CPU starvation, node/cluster, coverage instrumentation, clock epoch, and blocking writes to the stdout pipe (`slow-io` always empty).

### Still open

The kernel-level mechanism. In one run nothing in the process advanced a Python frame at all — including the sampler — and the disagg process stalled at the same time. Why binding one extra loopback PAIR socket at startup produces that is not explained here. It is recorded rather than guessed at.

### The fix

The ack channel is now established on first use instead of at startup. The proxy binds its end when profiling is first requested and passes the address on the request itself; the worker connects on first ack. A deployment that never profiles never creates the socket — which is both the fix and the better design, since the socket was pure cost there anyway.

`WorkerCommIpcAddrs.profile_ack_queue_addr` is gone, as nothing hands the address over at startup any more.

JunyiXu-nv:
## Fix verified; one real regression found and fixed

### The lazy ack channel works

Pipeline #57074, node `a4u8g-0059.ipp2a2` — the same physical node where this PR previously measured 36/47 ms (build 55878) and 52/29 ms (54606):

| | with eager queue (same node) | with lazy queue |
|---|---|---|
| handshake delay | 36/47, 52/29 ms | **2.12 / 1.90 ms** |
| applied offset | ~74 ms of pure error | **0.17 / 0.048 ms** |
| HMAC keys generated | 8 | **6** |
| causality assertion | failed, 3 attempts | **passed, 1 attempt** |

The HMAC count is the load-bearing number: delays move with machine load, but the queue count is deterministic, so it confirms the lazy path is actually taken rather than the run getting lucky.

All three profile suites pass on the rewired ack channel — `test_proxy_profile_sync` 12 passed, `test_profile_endpoints` 12 passed, `test_trtllm_serve_profile_example` 17 passed. That last one drives `/start_profile` and `/stop_profile` through the HTTP layer against the real IPC proxy, so the synchronous "trace is on disk when this returns 200" guarantee survives the change.

### The regression

The pipeline still failed, on something unrelated to the handshake:

```
AttributeError: PyExecutor object has no attribute _profile_enabled
tests/unittest/_torch/executor/test_kv_pool_rebalance.py::TestPpLoopDrainWiring  (4 tests, x86 and SBSA)
```

This PR wraps forward+sample in `_step_scope()`, which reads `_profile_enabled`. That fixture builds its executor with `object.__new__(PyExecutor)` and stubs "exactly the attributes one iteration touches" — so the new one has to be declared there. Fixed by adding `exe._profile_enabled = False` to `_make_pp_loop_executor()`.

I took that over making `_step_scope()` defensive with `getattr(..., False)`: production code should not loosen its own invariants to accommodate a test double, and that call is on the per-iteration path.

Checked the rest of the blast radius rather than assuming: ten test files construct `PyExecutor` via `object.__new__`, and only this one reaches `_step_scope`. `test_py_executor.py` uses the pattern 35 times across every executor loop and passes 149/149. The executor stage was 1062 passed / 4 failed, and those 4 are exactly the ones fixed here.

Head is now `c29f879`.

JunyiXu-nv:
## Everything this PR touches is green; the one failure is unrelated flake

Pipeline #57090 came back `UNSTABLE`. Across all 4632 artifacts there is exactly **one** failing test:

```
A30-CPP-1 :: test_unit_tests[batch_manager-80]
subprocess.CalledProcessError: ctest ... returned non-zero exit status 8
```

That is a C++ ctest. This PR changes no C++ — `git diff --name-only origin/main...HEAD` matches zero `.cpp/.cu/.h/.hpp/.cc` files. And the test is broadly flaky right now: in the last 7 days it has failed **30 times across ~20 different PRs plus post-merge/main**, of which this PR accounts for 1.

| PR | failures (7d) |
|---|---|
| 18033 | 2 |
| post-merge / main | 1 |
| **13872 (this PR)** | **1** |
| 10520, 16132, 16331, 16911, 17377, 17434, 17485, 17662, 17668, 17693, 17695, 17781, 17849, 17870, 17884, 17899, 17921, … | 1 each |

Everything this PR does touch passed, on `a4u8g-0064.ipp2u2`:

| check | result |
|---|---|
| disagg handshake delay | **1.23 / 1.47 ms** (was 36/47 ms on this PR) |
| applied clock offset | **0.15 / 0.26 ms** (was ~74 ms of pure error) |
| HMAC keys generated | **6** (was 8 — confirms the lazy path is taken) |
| `test_disaggregated_perf_metrics` | **passed, first attempt, 0 assertions** |
| executor unit tests (x86 / SBSA) | **1066 / 1066 passed, 0 FAILED** |
| `test_proxy_profile_sync` | **12 passed** |
| `test_profile_endpoints` | **12 passed** |
| `test_trtllm_serve_profile_example` | **17 passed** |

The executor stage went 1062 passed / 4 failed → 1066 passed, which is exactly the four `test_kv_pool_rebalance` fixtures fixed in the last push. No test was retried anywhere in the pipeline, so nothing else failed and recovered either.

Re-running CI for the flaky C++ test.

JunyiXu-nv:
## Second run: different unrelated failure, and almost no coverage

Pipeline #57123 failed on one test, and it is not the same one as last time:

```
DGX_B200-8_GPUs-PyTorch-2 :: test_parallel_vae.py::TestParallelVAEDecode::test_decode_width_2gpu
torch.distributed.DistNetworkError: Failed to recv, got 0 bytes
TCP client failed to connect/validate to host localhost:32061 - timed out (try=19, timeout=600000ms)
```

CI's own analysis concluded "PR likely to blame?: No", but I checked rather than taking that on faith, and the honest picture is more nuanced than for the previous flake:

- `test_decode_width_2gpu` ran **233 times in the last 7 days: 229 passed, 3 skipped, 1 failed** — and that 1 is this run. So it is rare (0.4%), not broadly flaky the way `batch_manager` is (30 failures across ~20 PRs in the same window).
- What makes it still unrelated is the shape, not the rate: it is a `torch.distributed` TCPStore rendezvous timeout (600 s, 19 retries) in a VisualGen multi-GPU VAE test. This PR touches no VisualGen code, no distributed init, and no C++ — there is no causal path from a lazily-created IPC queue in `trtllm-serve` to a TCPStore handshake in a separate B200 job.

Worth flagging: **this run produced almost no coverage.** 268 artifacts against 4632 in #57090, and zero `test_disaggregated_perf_metrics` artifacts — the B200 failure tripped fail-fast and aborted the single-GPU stages before they ran. So #57123 neither confirms nor contradicts anything about this PR.

The run that actually exercised this PR is still #57090: handshake 1.23/1.47 ms, HMAC keys 6, disagg test passed first attempt, executor unit tests 1066/1066 on both architectures, and all three profile suites green (12/17/12). Its only failure was the `batch_manager` C++ ctest, which this PR cannot touch (zero C++ files changed) and which failed 30 times across ~20 other PRs plus main in the same week.

Re-running.

JunyiXu-nv:
## Three runs, three different unrelated failures

| run | pipeline | artifacts | failure | that test's 7-day record |
|---|---|---|---|---|
| 1 | #57090 | **4632 (complete)** | `batch_manager` C++ ctest | 30 failures across ~20 PRs + main |
| 2 | #57123 | 268 (aborted early) | VAE B200 TCPStore timeout | 1 failure / 233 runs |
| 3 | #57150 | 188 (aborted early) | UCX/MPI endpoint timeout, `test_gen_first[adp-mtp2]` | 14 failures / 453 runs, across 7 sources — **5 of them on post-merge main** |

Run 3 is the one worth spelling out, because CI rated it "Confidence: Medium — the PR does touch disaggregated serving". The failure is `ucx send failed: Endpoint timeout` → `PMIX ERROR: UNREACHABLE` → `MPI_ABORT` on a Slurm H100 node: the KV-cache transport layer between MPI ranks. This PR adds two HTTP endpoints, makes one IPC queue lazy, and changes a clock handshake — none of that reaches UCX or MPI. And the same test failed 5 times on main in the same week, where this PR does not exist.

Runs 2 and 3 also carry no evidence either way about this PR: fail-fast aborted them before the single-GPU stages ran, so neither produced a single `test_disaggregated_perf_metrics` artifact (268 and 188 artifacts, against 4632 in run 1).

**Run #57090 remains the only complete exercise of this PR, and everything it touches passed:**

| check | result |
|---|---|
| disagg handshake delay | 1.23 / 1.47 ms (was 36/47 ms with the eager queue, same node class) |
| applied clock offset | 0.15 / 0.26 ms (was ~74 ms of pure error) |
| HMAC keys generated | 6 (was 8 — confirms the lazy path) |
| `test_disaggregated_perf_metrics` | passed, first attempt, 0 assertions |
| executor unit tests (x86 / SBSA) | 1066 / 1066 passed, 0 FAILED |
| `test_proxy_profile_sync` | 12 passed |
| `test_profile_endpoints` | 12 passed |
| `test_trtllm_serve_profile_example` | 17 passed |

Re-running once more.

JunyiXu-nv:
## Pausing re-runs: the RDMA/UCX fabric is unhealthy cluster-wide

Four runs, four different failures, none of them this PR — and the last two are the same class:

| run | pipeline | artifacts | failure |
|---|---|---|---|
| 1 | #57090 | **4632 (complete)** | `batch_manager` C++ ctest |
| 2 | #57123 | 268 | VAE B200 TCPStore rendezvous timeout |
| 3 | #57150 | 188 | `ucx send failed: Endpoint timeout` → `MPI_ABORT` |
| 4 | #57168 | 117 | `ibv_create_ah ... Connection timed out` → `NIXL_ERR_NOT_FOUND` |

Runs 3 and 4 are both UCX/RDMA transport failures on DGX H100 multi-GPU disagg stages, so I checked whether that is specific to this PR. It is not.

**`test_disaggregated_serving` has failed 40 times in the last 24 hours across 11 different PRs plus post-merge main**, concentrated on exactly those stages:

```
DGX_H100-4_GPUs-PyTorch-Others-2     10
DGX_H100-4_GPUs-PyTorch-DeepSeek-1    8
DGX_H100-4_GPUs-PyTorch-Others-1      8
DGX_H100-2_GPUs-PyTorch-Others-1      6
DGX_H100-2_GPUs-PyTorch-Others-2      5
```

This PR accounts for 5 of those 40. Re-running will not converge until the fabric is healthy, so I have stopped re-triggering rather than keep burning CI.

Runs 2-4 also carry no evidence about this PR either way: fail-fast aborted them before the single-GPU stages ran — 117-268 artifacts against 4632, and zero `test_disaggregated_perf_metrics` artifacts in any of them.

**Run #57090 remains the only complete exercise of this PR, and everything it touches passed:**

| check | result |
|---|---|
| disagg handshake delay | 1.23 / 1.47 ms (was 36/47 ms with the eager queue, same node class) |
| applied clock offset | 0.15 / 0.26 ms (was ~74 ms of pure error) |
| HMAC keys generated | 6 (was 8 — confirms the lazy path is taken) |
| `test_disaggregated_perf_metrics` | passed, first attempt, 0 assertions |
| executor unit tests (x86 / SBSA) | 1066 / 1066 passed, 0 FAILED |
| `test_proxy_profile_sync` | 12 passed |
| `test_profile_endpoints` | 12 passed |
| `test_trtllm_serve_profile_example` | 17 passed |

Suggest waiting for the RDMA issue to clear before re-running.

JunyiXu-nv:
Rebased onto latest main (64 commits) and resolved four conflicts. Head is now `366d795e37`, `MERGEABLE`, full pre-commit clean.

The conflicts were semantic, not textual — two of them would have silently reverted main's work if resolved by taking this branch's side:

| conflict | what main did | resolution |
|---|---|---|
| `l0_a10.yml` | moved the MoE test tree `_torch/modules/moe/` → `_torch/moe/` | took main's path, merged in our entry |
| `openai_server.py` imports | added `ChatCompletionMessageParam`, `ChatCompletionToolsParam`, `StreamOptions` | union of 35 symbols, each verified to exist in `openai_protocol.py` |
| `openai_server.py` routes | refactored the RL routes into `_register_rl_control_routes()` | kept our profile routes + main's helper call, after checking the helper registers exactly the three routes our inline copy did |
| `l0_a10.yml` (clockfix) | replaced `- unittest/tools` with an explicit per-file list to keep `test_layer_wise_benchmarks.py` off A10 (nvbugs/6337228) | kept main's list and its comment, merged in our one entry |

Re-triggering CI.

JunyiXu-nv:
Rebased again — this time to pick up a fix, not just to clear a conflict.

Seven of the last eight runs died on things this PR does not touch, and the recurring one was `test_combined_fmha.py::test_flashinfer_context_fallback_scope` on DGX_B200. Rather than keep re-running, I checked where that test actually stands on main:

```
f152eb29af  2026-09-01  [https://nvbugs/6697142][fix] Wire the stub the way
                        `TrtllmAttention.forward` does and parametrize over both… (#18504)
```

It has a tracked bug (nvbugs/6697142) and a landed fix, and it is not in `waives.txt` — so it was genuinely fixed, not waived around. The branch was based on `73ee6b0b19` (08-31 10:25), a day older than that fix, which is why every run kept hitting it. The 24 failures visible in the last 24 h all belong to a single post-merge main build, i.e. also from before the fix.

Head is now `0e462d0f79`, based directly on `f152eb29af`, zero commits behind main, `MERGEABLE`, full pre-commit clean. The rebase produced no conflicts — `rerere` reused the four resolutions from the previous rebase (MoE test-tree move, the import union, the RL-route refactor, and the per-file `unittest/tools` list).

Worth recording so it does not get repeated: I had earlier concluded from OpenSearch that this test was "0 failures in the last 12 h" and treated that as a green light. That was wrong — the index runs more than six hours behind, so the recent windows read as "0 runs", not "0 failures". Commit history is the reliable signal here; failure statistics are not.

Re-triggering.

JunyiXu-nv:
Rebased onto main tot (220 commits). Head is `30bd6ba570`, `MERGEABLE`, full pre-commit clean, zero commits behind.

Two conflicts:

**`py_executor.py`** — main modified the inline `profile_step` while this PR moves that whole block into `profiling.py`. Resolved by keeping the delegation structure here and porting main's two changes into `profiling.py`:

- the new `adp_dummy_ctx_tokens` / `adp_dummy_gen_tokens` fields in the iteration log line;
- the per-iteration reset of those two counters.

The reset is the one that mattered. It is behavioural, not cosmetic: dropping it would let the counters accumulate forever and silently report wrong numbers in every iteration log, with nothing failing to flag it. I diffed main's actual changes to `_profiler` over this window and ported them item by item rather than picking whichever side looked right.

**`l0_a10.yml`** — main added `test_msgspec_route.py`, this PR adds `test_steady_clock_sync.py`; kept both.

Also note main moved `test_kv_pool_rebalance.py` into `tests/unittest/_torch/executor/kv_cache/`. Git followed the rename, and the `_profile_enabled` fixture line this PR added is still there — verified after the rebase, not assumed.

Re-triggering CI.

JunyiXu-nv:
Rebased onto main tot (53 commits) to pick up `7db313ce7a` (#18894), which fixes the fused finish-reason guard that broke the last two runs. Head is `497f7e861f`, `MERGEABLE`, zero behind main, no conflicts, full pre-commit clean.

For the record on those two runs, since the usual "it fails on other PRs too" argument did not apply here — `test_modeling_out_of_tree` had failed exactly twice in 9653 runs over a week, and both were ours. Three independent lines showed it was not this PR:

- **Lineage.** The failing assertion `new_tokens.shape == (...)` lives in `_can_fuse_finish_reasons`, a function added whole by `3933f62170` (#18714) at 16:50 that day. This branch had rebased onto a base cut at 17:23 — 33 minutes later. The GB10 stage's 2518 passing runs were all on bases predating it.
- **Drift.** Run one failed on GB10; run two failed on RTX5090, RTX5080, DGX_H100 and H100_PCIe, and the 13 failures there all reduced to the same `Sampling failed`. A defect in this PR would not migrate across platforms like that; a guard on the shared sampler decode path would.
- **Surface.** This PR changes zero files under `sampler/`, and the traceback runs from `_sample_async` down into `finish_reasons.py` without leaving the sampler.

Re-triggering.

liji-nv (approved):

PR #13873, opened 2026-05-08
[TRTLLM-12503][feat] Parallel VAE independent scaling and fix arg passing

NVShreyas:
## Summary by CodeRabbit

* **New Features**
  * Parallel VAE configuration now uses explicit numeric sizing (`--parallel_vae_size`) instead of boolean enable/disable flag for finer control over VAE parallelization across GPUs.
  * **Default is now 1** - Would prevent errors when scaling Ulysses or CFG as previously assumed VAE size = world size led to errors with some shapes when scaling beyond 4-8 gpus.
  * Can treat this as another dimension but recommended is to maximize the VAE size

* **Configuration Updates**
  * Example configurations and test profiles updated to use the new `parallel_vae_size` parameter format with appropriate GPU-count values.

* **Tests**
  * Added validation tests for parallel VAE size constraints and behavior across different GPU configurations.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/examples/visual_gen/configs/wan2.2-t2v-fp8-8gpu.yaml
+++ b/examples/visual_gen/configs/wan2.2-t2v-fp8-8gpu.yaml
@@ -23,6 +23,6 @@ attention:
 parallel:
   dit_cfg_size: 2
   dit_ulysses_size: 4
-  enable_parallel_vae: true
+  parallel_vae_size: 8
 cuda_graph:
   enable_cuda_graph: false
--- a/examples/visual_gen/visual_gen_wan_t2v.py
+++ b/examples/visual_gen/visual_gen_wan_t2v.py
@@ -228,7 +228,12 @@ def parse_args():
         "Can be set independently of --attn2d_row_size; asymmetric meshes (e.g. 1x4 or 4x1) are valid. "
         "Cannot be combined with --ulysses_size (not yet implemented).",
     )
-    parser.add_argument("--disable_parallel_vae", action="store_true", help="Disable parallel VAE")
+    parser.add_argument(
+        "--parallel_vae_size",
+        type=int,
+        default=1,
+        help="Number of ranks used for parallel VAE. 1 disables parallel VAE.",
+    )
 
     # CUDA graph
     parser.add_argument(
@@ -333,7 +338,7 @@ def main():
             "dit_ulysses_size": args.ulysses_size,
             "dit_attn2d_row_size": args.attn2d_row_size,
             "dit_attn2d_col_size": args.attn2d_col_size,
-            "enable_parallel_vae": not args.disable_parallel_vae,
+            "parallel_vae_size": args.parallel_vae_size,
         },
         torch_compile={
             "enable_torch_compile": not args.disable_torch_compile,
--- a/tensorrt_llm/_torch/visual_gen/config.py
+++ b/tensorrt_llm/_torch/visual_gen/config.py
@@ -95,7 +95,11 @@ class ParallelConfig(StrictBaseModel):
            2x2 mesh: Q gathered across row group, K/V gathered across col group
     """
 
-    enable_parallel_vae: bool = True
+    parallel_vae_size: int = PydanticField(
+        1,
+        ge=1,
+        description="Number of ranks used for VAE parallelism. 1 disables parallel VAE.",
+    )
     parallel_vae_split_dim: Literal["width", "height"] = "width"
 
     # DiT Parallelism
@@ -589,10 +593,6 @@ class DiffusionModelConfig(BaseModel):
 
     # Unified parallelism mapping (populated by setup_visual_gen_mapping)
     visual_gen_mapping: Optional[Any] = None  # VisualGenMapping (lazy import)
-
-    # VAE parallelism (promoted from ParallelConfig for pipeline_loader)
-    enable_parallel_vae: bool = True
-    parallel_vae_split_dim: Literal["width", "height"] = "width"
 
     dynamic_weight_quant: bool = False
 
--- a/tensorrt_llm/_torch/visual_gen/mapping.py
+++ b/tensorrt_llm/_torch/visual_gen/mapping.py
@@ -123,7 +123,7 @@ class VisualGenMapping(DeviceMeshTopologyImpl):
         self.parallel_vae_size = parallel_vae_size
         self._vae_ranks = list(range(self.parallel_vae_size))
         self._vae_group: Optional[ProcessGroup] = None
-        self._vae_adj_groups: list[ProcessGroup] = []
+        self._vae_adj_groups: list[Optional[ProcessGroup]] = []
         self._attn2d_row_group: Optional[ProcessGroup] = None
         self._attn2d_col_group: Optional[ProcessGroup] = None
         self._order = order
@@ -212,18 +212,19 @@ class VisualGenMapping(DeviceMeshTopologyImpl):
         """Create the process group used by parallel VAE."""
         if self.parallel_vae_size <= 1:
             return
-        if self.parallel_vae_size == self.world_size:
-            self._vae_group = dist.group.WORLD
-        else:
-            pg = dist.new_group(self._vae_ranks, use_local_synchronization=True)
-            if self._rank in self._vae_ranks:
-                self._vae_group = pg
 
+        pg = dist.new_group(self._vae_ranks, use_local_synchronization=False)
+        if self._rank in self._vae_ranks:
+            self._vae_group = pg
+
+        adj_groups: list[Optional[ProcessGroup]] = [None] * (self.parallel_vae_size - 1)
         for i in range(self.parallel_vae_size - 1):
             ranks = [self._vae_ranks[i], self._vae_ranks[i + 1]]
-            pg = dist.new_group(ranks, use_local_synchronization=True)
+            pg = dist.new_group(ranks, use_local_synchronization=False)
             if self._rank in ranks:
-                self._vae_adj_groups.append(pg)
+                adj_groups[i] = pg
+        if self._rank in self._vae_ranks:
+            self._vae_adj_groups = adj_groups
 
     # ------------------------------------------------------------------
     # Rank decomposition
@@ -308,7 +309,7 @@ class VisualGenMapping(DeviceMeshTopologyImpl):
         return self._vae_group
 
     @property
-    def vae_adj_groups(self) -> list[ProcessGroup]:
+    def vae_adj_groups(self) -> list[Optional[ProcessGroup]]:
         return self._vae_adj_groups
 
     # ------------------------------------------------------------------
--- a/tensorrt_llm/_torch/visual_gen/mapping.py
+++ b/tensorrt_llm/_torch/visual_gen/mapping.py
@@ -123,6 +123,7 @@ class VisualGenMapping(DeviceMeshTopologyImpl):
         self.parallel_vae_size = parallel_vae_size
         self._vae_ranks = list(range(self.parallel_vae_size))
         self._vae_group: Optional[ProcessGroup] = None
+        self._vae_adj_groups: list[ProcessGroup] = []
         self._attn2d_row_group: Optional[ProcessGroup] = None
         self._attn2d_col_group: Optional[ProcessGroup] = None
         self._order = order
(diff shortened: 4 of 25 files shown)


tests/integration/defs/perf/visual_gen_perf_utils.py
```diff
@@ -175,9 +174,6 @@ def build_visual_gen_db_entry(
         "b_enable_two_stage": bool(
             server_config.get("spatial_upsampler_path") or server_config.get("distilled_lora_path")
         ),
-        "b_enable_parallel_vae": bool(
-            _get_nested_value(server_config, "parallel.enable_parallel_vae", False)
-        ),
```
NVShreyas, line 180: @zhenhuaw-me please review this and let me know if we wanna keep backward compat or if I need to change anything else.

tensorrt_llm/_torch/visual_gen/pipeline.py
```diff
@@ -468,19 +468,29 @@ def _setup_cache_acceleration(
...
 
-        # Uses all ranks today; replace with a subset to dedicate specific ranks to VAE.
-        pg = dist.new_group(list(range(dist.get_world_size())))
+        vgm = self.model_config.visual_gen_mapping
+        if vgm is None:
+            return
+
+        vae_ranks = vgm.vae_ranks
+        if dist.get_rank() not in vae_ranks:
+            return
```
yibinl-nvidia: This early return could violate the PyTorch `new_group` [contract](https://docs.pytorch.org/docs/2.11/distributed.html#torch.distributed.new_group) 

> This function requires that all processes in the main group (i.e. all processes that are part of the distributed job) enter this function, even if they are not going to be members of the group. Additionally, groups should be created in the same order in all processes.

[Here](https://github.com/NVIDIA/TensorRT-LLM/blob/b9c95338e10986c9c7a9e6307633396fe81ecf70/tensorrt_llm/_torch/visual_gen/modules/vae/parallel_vae_interface.py#L128) (outside this code change) only VAE ranks call the later new_groups, i.e. the `new_group` function inside `_build_adj_groups` is only reachable by VAE ranks.

tests/integration/defs/perf/visual_gen_perf_utils.py
zhenhuaw-me: It should be fine at this stage where we are refactoring the APIs. cc @taianz-nv

zhenhuaw-me (approved):
LGTM in general.

tests/integration/defs/perf/visual_gen_perf_utils.py
zhenhuaw-me: @NVShreyas we may need to change this to something like `b_parallel_vae_size` since these are part of the matching keys.

NVShreyas: sure I will do that, thanks!

tensorrt_llm/_torch/visual_gen/pipeline.py
NVShreyas: thanks, good point! I have moved `vae_adj_group` creation to mapping so every rank calls it but is only used by VAE ranks.

tests/integration/defs/perf/visual_gen_perf_utils.py
NVShreyas: Added `l_parallel_vae_size`

chang-l (commented):
Retracted — re-posting as a pending review for staging. Please disregard.

tensorrt_llm/_torch/visual_gen/pipeline.py
```diff
@@ -622,7 +633,8 @@ def decode_latents(
         extra_latents: Optional[Dict[str, Tuple[torch.Tensor, Callable]]] = None,
     ):
         """Execute VAE decoding. Only rank 0 performs decoding.
-        If parallel VAE is enabled, all processes perform decoding.
+        If parallel VAE is enabled, only ranks in the VAE process group
+        perform decoding; non-participating ranks return None placeholders.
```
chang-l: The new docstring says *"only ranks in the VAE process group perform decoding; non-participating ranks return None placeholders"* — but the body wasn't updated. It routes by `self._parallel_vae_enabled` then `self.rank == 0`, which only matches the docstring because rank 0 happens to always be in `vae_ranks`.

CC surfaced this bug surface: if anyone later makes `vae_ranks` not start at 0 (e.g., `vae_ranks = [4,5,6,7]` for some scheduling reason), then rank 0 would have `_parallel_vae_enabled=False`, fall into the `self.rank == 0` branch, and try to do a single-GPU `decode_fn(latents)` — meanwhile ranks 4–7 are inside parallel decode collectives waiting for each other → hang or wrong output.

Suggest replacing `if self.rank == 0:` with `if self.rank in vgm.vae_ranks` (or equivalent) so the docstring's claim is enforced by the code.

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -113,6 +120,10 @@ def __init__(
         self.ulysses_size = ulysses_size
         self.attn2d_row_size = attn2d_row_size
         self.attn2d_col_size = attn2d_col_size
+        self.parallel_vae_size = parallel_vae_size
+        self._vae_ranks = list(range(self.parallel_vae_size))
```
chang-l, line 124: `vae_ranks` is computed without reference to the DiT mesh. Today, with the default `dit_dim_order`, the first `parallel_vae_size` ranks happen to land inside a single CFG group, which is what makes the asymmetric path work. But:

- If `dit_dim_order` is changed, the first N ranks may straddle CFG groups.
- With the Ulysses + CP combination flagged in the earlier PR's TODO, "first N ranks" is no longer obviously the right N.

Could we either:
- Derive `vae_ranks` from the mesh (e.g., "the first `parallel_vae_size` ranks along the outermost dim"), or
- At minimum, validate at construction time that `vae_ranks` falls inside a single CFG group, and `raise NotImplementedError` otherwise with a pointer to this comment?

Happy with either; I just don't want this to silently regress when someone tunes `dit_dim_order`.

examples/visual_gen/configs/wan2.2-t2v-fp8-8gpu.yaml
```diff
@@ -23,6 +23,6 @@ attention:
 parallel:
   dit_cfg_size: 2
   dit_ulysses_size: 4
-  enable_parallel_vae: true
+  parallel_vae_size: 8
```
chang-l, line 26: All three modified configs use`parallel_vae_size = world_size` – does it make sense to have this?
- `wan2.2-t2v-fp8-8gpu.yaml`: `parallel_vae_size: 4` (< world_size)

Or, do you think we can keep scaling till 8GPUs for Wan vae?

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -95,7 +95,11 @@ class ParallelConfig(StrictBaseModel):
            2x2 mesh: Q gathered across row group, K/V gathered across col group
     """
 
-    enable_parallel_vae: bool = True
+    parallel_vae_size: int = PydanticField(
+        1,
```
chang-l, line 99: Minor: `ge=1` plus `parallel_vae_size <= world_size` in `VisualGenMapping` cover the basics, but a couple of mismatches will still fail late, deep inside `ParallelVAEFactory.from_vae`:

- `parallel_vae_size` doesn't divide the VAE spatial dim along `parallel_vae_split_dim`.
- `parallel_vae_size` doesn't align with the DiT mesh (see comment on `mapping.py`).

Not blocking, but a config-time validator here would give friendlier errors than a `ValueError` inside the factory at load time.

tensorrt_llm/_torch/visual_gen/mapping.py
```diff
@@ -196,6 +208,24 @@ def _build_attn2d_groups(self) -> None:
             f"VisualGenMapping._build_attn2d_groups: row_size={row_size}, col_size={col_size}"
         )
 
+    def _build_vae_group(self) -> None:
+        """Create the process group used by parallel VAE."""
+        if self.parallel_vae_size <= 1:
+            return
+
+        pg = dist.new_group(self._vae_ranks, use_local_synchronization=False)
```
chang-l, line 253: One-line comment would help future readers: `use_local_synchronization=False` because PyTorch requires every rank to participate in `new_group`, otherwise non-VAE ranks deadlock on the next world-collective — without a comment, someone will wonder why both spellings exist.

examples/visual_gen/visual_gen_wan_t2v.py
```diff
@@ -228,7 +228,12 @@ def parse_args():
         "Can be set independently of --attn2d_row_size; asymmetric meshes (e.g. 1x4 or 4x1) are valid. "
         "Cannot be combined with --ulysses_size (not yet implemented).",
     )
-    parser.add_argument("--disable_parallel_vae", action="store_true", help="Disable parallel VAE")
+    parser.add_argument(
+        "--parallel_vae_size",
```
chang-l, line 232: Do we need to update the VisualGen docs to clarify that parallel VAE can be asymmetric, and it can use an arbitrary number of GPUs smaller than world_size?

chang-l (commented):
Do we have any e2e tests or pipeline tests with this feature, i.e.,`world_size > parallel_vae_size`?

tensorrt_llm/_torch/visual_gen/pipeline.py
NVShreyas: good point. I have made `self._parallel_vae_enabled` a global flag for all ranks so that all ranks know which ranks should participate in the decode. If parallel VAE is disabled for some reason then only rank 0 will decode.

tensorrt_llm/_torch/visual_gen/mapping.py
NVShreyas: Good catch, I added a validation check to ensure vae_ranks fall inside the same cfg group.

examples/visual_gen/configs/wan2.2-t2v-fp8-8gpu.yaml
NVShreyas: we can scale to 8 GPUs for WAN. I have an e2e test with world_size=4 and parallel_vae_size=2 for WAN which will test parallel_vae_size < world_size.

tensorrt_llm/_torch/visual_gen/config.py
NVShreyas: I've added a validator for the second one in the VisualGenMapping constructor so it will raise before model load if anything is misaligned.
For the first one, it depends on the requested H, W shapes which makes it request dependent and hard to check at construction time unless we assume some default shapes which I dont think is ideal

tensorrt_llm/_torch/visual_gen/mapping.py
NVShreyas: done

examples/visual_gen/visual_gen_wan_t2v.py
NVShreyas: I added 1 line under `Multi-GPU Parallelism` in `docs/models/visual-generation.md` describing what it does. Let me know if I need to add it somewhere else too.

tests/unittest/_torch/visual_gen/multi_gpu/test_wan_pipeline_parallel.py
```diff
@@ -0,0 +1,340 @@
...
+        seed=SEED,
+    )
+    _free(hf_pipe)
+
+    assert trtllm_video.numel() == hf_video.numel(), (
+        f"Element count mismatch — TRTLLM {tuple(trtllm_video.shape)} "
+        f"({trtllm_video.numel()}) vs HF {tuple(hf_video.shape)} ({hf_video.numel()})"
+    )
+
+    cos_sim = _cosine_similarity(trtllm_video, hf_video)
```
chang-l, line 303: @yibinl-nvidia  Instead of using cosine similarity in pixel space, this may be a good place to apply your LPIPS script/function in the E2E test.

examples/visual_gen/configs/wan2.2-t2v-fp8-8gpu.yaml
chang-l: >I have an e2e test with world_size=4 and parallel_vae_size=2 for WAN which will test parallel_vae_size < world_size.

Great – thanks 

>we can scale to 8 GPUs for WAN.

Does it still achieve speedup with `parallel_vae_size=8`, or does scaling stop beyond 4 GPUs?

chang-l (approved):

tests/unittest/_torch/visual_gen/multi_gpu/test_wan_pipeline_parallel.py
chang-l: btw, please switch to LPIPS once PR #13567 is merged.

examples/visual_gen/configs/wan2.2-t2v-fp8-8gpu.yaml
NVShreyas: Yes, it helps the most when cfg=2, ulysses=4. In very early tests, it brought down latency by 7.65% on 8 gpus. I'm curious to see the pattern when scaling to 16 on gb200

LarryXFly (approved):

ruodil (approved):

PR #13886, opened 2026-05-08
[None][fix] Make indexer top-K launch policy device-aware

mingyangHao:
@coderabbitai summary

### Magic number audit

| Constant | Before | After |
|---|---|---|
| `kDecodeTargetTotalBlocks` | `132` (H100 SM count) | **deleted** — replaced by `getSchemeXBounds().smCount` |
| `kNumBins` | hardcoded `2048` × 2 function-local copies | hoisted to namespace scope; single source of truth |
| `kSortingAlgorithmThreshold` | `12288` (file-scope + 2 function-local copies) | `6 * kNumBins` (namespace scope only; duplicates removed) |
| `kDecodeMinColsPerSubBlock` | `2048` | `kNumBins` |
| `kDefaultSplitWorkThreshold` | `200 * 1000` (file-scope + 2 function-local copies) | unchanged value; duplicates removed |
| `kMaxBlocksPerRowDecode` | `10` | unchanged (empirical cap on aux-buffer / merge SMEM) |

### Perf data (B200, fp32 logits, cols=196608, next_n=1, no preIdx)

Production batch sizes (BS=1/32/128 — covering the original PR's perf table):

| topK | BS=1 | BS=32 | BS=128 |
|---:|---|---|---|
| **512 baseline (us)** | 51.1 | 65.2 | 95.0 |
| **512 new (us)** | 51.1 | 67.3 | 95.2 |
| **512 speedup** | 1.00x | 0.97x | 1.00x |
| **1024 baseline (us)** | 56.0 | 72.6 | 105.7 |
| **1024 new (us)** | 55.9 | 72.9 | 109.9 |
| **1024 speedup** | 1.00x | 1.00x | 0.96x |
| **2048 baseline (us)** | 63.1 | 79.8 | 122.8 |
| **2048 new (us)** | 63.1 | 79.3 | 121.9 |
| **2048 speedup** | 1.00x | 1.01x | 1.01x |

Cliff band (BS=100-300):

| topK | BS=100 | BS=132 | BS=148 | BS=200 | BS=300 |
|---:|---|---|---|---|---|
| **512 baseline (us)** | 93.5 | 288.4 | 384.0 | 409.8 | 460.9 |
| **512 new (us)** | 89.3 | 107.6 | 112.2 | 144.0 | 228.4 |
| **512 speedup** | **1.05x** | **2.68x** | **3.42x** | **2.85x** | **2.02x** |
| **1024 baseline (us)** | 93.6 | 289.0 | 386.3 | 413.3 | 465.3 |
| **1024 new (us)** | 90.2 | 109.0 | 113.3 | 145.1 | 230.7 |
| **1024 speedup** | **1.04x** | **2.65x** | **3.41x** | **2.85x** | **2.02x** |
| **2048 baseline (us)** | 107.7 | 296.6 | 392.1 | 419.4 | 473.3 |
| **2048 new (us)** | 105.6 | 124.7 | 130.3 | 172.9 | 274.1 |
| **2048 speedup** | **1.02x** | **2.38x** | **3.01x** | **2.43x** | **1.73x** |
## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/kernels/indexerTopK.cu
+++ b/cpp/tensorrt_llm/kernels/indexerTopK.cu
@@ -39,6 +39,13 @@ TRTLLM_NAMESPACE_BEGIN
 
 namespace kernels
 {
+// Radix histogram bin count, used by topKPerRowJob's per-step distribution
+// pass. 2048 = 2^11 maps directly to the fp16 fast path's 11-bit bin index
+// (sign + exponent + top mantissa bits, mask 0x7FF). All file-scope dispatch
+// thresholds below derive from this so that a future change to the histogram
+// width does not require re-tuning the heuristics.
+constexpr int kNumBins = 2048;
+
 namespace
 {
 
@@ -625,9 +632,6 @@ static __global__ __launch_bounds__(kNumThreadsPerBlock) void topKPerRowPrefill(
 #if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
     cudaGridDependencySynchronize();
 #endif
-    // The number of bins in the histogram.
-    static constexpr int kNumBins = 2048;
-
     // The row computed by this block.
     int rowIdx = blockIdx.x + offsetIndex;
 
@@ -650,15 +654,11 @@ template <int kNumThreadsPerBlock, bool useRadixSort, bool multipleBlocksPerRow 
     typename InputT = float>
 static __global__ __launch_bounds__(kNumThreadsPerBlock) void topKPerRowDecode(InputT const* logits, int const* seqLens,
     int* outIndices, int stride0, int stride1, int const topK, int next_n, int compressRatio,
-    float* outLogits = nullptr,
-    int const numBlocksToMerge = 0, int const* indices = nullptr)
+    float* outLogits = nullptr, int const numBlocksToMerge = 0, int const* indices = nullptr)
 {
 #if (defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 900))
     cudaGridDependencySynchronize();
 #endif
-    // The number of bins in the histogram.
-    static constexpr int kNumBins = 2048;
-
     // The row computed by this block.
     int rowIdx = blockIdx.x;
 
@@ -699,22 +699,24 @@ static __global__ __launch_bounds__(kNumThreadsPerBlock) void topKPerRowDecode(I
 namespace
 {
 
-// Below this many columns, when blocksPerRow == 1, the kernel uses the
-// insertion-sort variant (smaller SMEM footprint than cub::BlockRadixSort).
-constexpr int kSortingAlgorithmThreshold = 12288;
-// Default for splitWorkThreshold: at/above this many columns, always use the
-// multi-block split path (per-block work would otherwise become too large).
-// Callers may override via the splitWorkThreshold parameter.
+// Insertion vs radix crossover for the single-block path. Radix always pays a
+// histogram-clear + per-step scan over kNumBins bins (~4 refinement passes);
+// insertion only maintains a topK SMEM array. The crossover is where insertion's
+// O(numColumns * lg topK) catches up to radix's O(numColumns + kNumBins) per
+// pass — measured empirically around 6 histograms of work.
+constexpr int kSortingAlgorithmThreshold = 6 * kNumBins;
+// Force the multi-block split-and-merge path above this column count: per-block
+// work would otherwise dwarf the merge-pass cost. Callers may override via the
+// splitWorkThreshold parameter. Tuned empirically; below this width the merge
+// launch overhead isn't worth saving the per-block radix cost.
 constexpr int kDefaultSplitWorkThreshold = 200 * 1000;
-// Cap on blocks-per-row in the multi-block path. Bounds aux-buffer size and the
-// merge-step input size.
+// Cap blocks-per-row in the multi-block path. Bounds the aux-buffer size
+// (numRows * blocksPerRow * topK * sizeof(int32)) and the merge-step input width
+// so the second-pass merge kernel stays SMEM-resident.
 constexpr int kMaxBlocksPerRowDecode = 10;
-// Approximate target for total launched blocks (~ SM count of H100/B200), used to
-// decide how aggressively to split rows when there are too few of them to fill SMs.
-constexpr int kDecodeTargetTotalBlocks = 132;
-// Each sub-block should process at least this many columns; otherwise the radix-pass
-// fixed overhead (histogram clear + scan + syncs) starts to dominate.
-constexpr int kDecodeMinColsPerSubBlock = 2048;
+// Each sub-block must amortize the radix histogram overhead, i.e. cover at least
+// one full histogram pass worth of columns.
+constexpr int kDecodeMinColsPerSubBlock = kNumBins;
 
 // Scheme X bound calculator — shared between fp32 and bf16/fp16 dispatchers.
 // Caches hardware attrs (SM count, L2 capacity) and the small-N threshold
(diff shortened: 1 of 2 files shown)


cpp/tensorrt_llm/kernels/indexerTopK.cu
```diff
@@ -690,52 +689,166 @@ static __global__ __launch_bounds__(kNumThreadsPerBlock) void topKPerRowDecode(f
...
+
+    int blocksPerRow = smCount / numRows;
+    if (blocksPerRow <= 1 && numRows <= previousPowerOfTwo(smCount))
+    {
+        // Power-of-two batches just below the SM count can still benefit from a
+        // 2-way split on very wide rows. Keep non-power-of-two tail batches on
+        // the single-block path to avoid adding a second wave indiscriminately.
+        blocksPerRow = 2;
+    }
+    return std::max(1, std::min(maxBlocksPerRow, blocksPerRow));
```
longcheng-nv: **bump-to-2 contradicts the wave-fit principle that motivates the rest of this PR.**

The 1.18–1.27× wins at `BS=32` are explained by single-wave fit (`32 × 4 = 128 ≤ 148 SM` vs `32 × 5 = 160 → 1.08 wave`). But this `blocksPerRow = 2` bump fires for `numRows ∈ (smCount/2, prevPow2(smCount)]`. On B200 (`smCount=148, prevPow2=128`):

| numRows | blocksPerRow | total blocks | waves on B200 |
|---:|---:|---:|---:|
|  74 | 2 (`smCount/74=2`) | 148 | 1.00 |
| 100 | 2 (bump)           | **200** | **1.35** |
| 128 | 2 (bump)           | **256** | **1.73** |

So `BS=128, cols ≥ 131K` lands at **1.73 waves** — the exact failure mode the rest of the PR avoids. The comment "Power-of-two batches just below the SM count" is also too narrow — the band includes plenty of non-pow2 batches like 100.

Two follow-ups:
1. Do you have perf data here? The table only has `BS ∈ {1, 32}`; the bump only fires for `BS > smCount/2`.
2. If multi-wave is empirically winning thanks to halved per-block work + amortized merge, the comment should justify that; otherwise consider tightening with `numRows * 2 <= smCount` to keep the wave-fit invariant clean.

tests/unittest/_torch/thop/parallel/test_indexer_topk.py
```diff
@@ -279,22 +328,31 @@ def test_indexer_topk_decode(batch_size, next_n, index_topk, num_tokens, compres
...
+# Regression coverage for serving-style decode inputs: the logits tensor keeps a
+# static maximum width, while per-row effective lengths are ragged and may be much
+# shorter. This catches invalid indices that pure max(row_end)-sized unit inputs
+# can hide.
+@pytest.mark.parametrize("batch_size", [16, 32, 64])
 @pytest.mark.parametrize("next_n", [1, 2])
 @pytest.mark.parametrize("index_topk", [2048])
-@pytest.mark.parametrize("num_tokens", [4096, 16384, 32768])
+@pytest.mark.parametrize("num_columns", [4096, 8192, 16384])
 @pytest.mark.parametrize("compress_ratio", [1, 4])
```
longcheng-nv: **Test-coverage regression — the deleted `test_indexer_topk_decode_sm_saturation` parametrization covers the exact band the new policy decides in.**

The old test parametrized `batch_size ∈ {16, 32, 128} × num_tokens ∈ {4096, 16384, 32768}`. This new ragged-width test drops `batch_size=128` and `num_tokens=32768` — both of which exercise the smTarget/occupancyBlocks transition that this PR rewires. The bump-to-2 logic (see inline thread on `indexerTopK.cu:783–791`) also lands inside that band.

The new ragged-width check is *orthogonally* valuable (numerical correctness under static-width + ragged effective rows). Suggest **keeping both** — rename the old test to e.g. `test_indexer_topk_decode_launch_policy_transitions` and parametrize `BS ∈ {1, 32, 64, 100, 128, 256}` so future regressions in the wave-fit logic are caught. CI cost is small (both are unit tests).

cpp/tensorrt_llm/thop/IndexerTopKOp.cpp
```diff
@@ -102,22 +108,44 @@ void indexer_topk_decode(th::Tensor const& logits, th::Tensor const& seq_lens, t
...
-    th::Tensor aux_indices = th::empty({0}, th::TensorOptions().dtype(th::kInt32).device(logits.device()));
-    th::Tensor aux_logits = th::empty({0}, th::TensorOptions().dtype(th::kFloat32).device(logits.device()));
-    if (blocks_per_row > 1)
+    int const splitWorkThreshold = kDefaultSplitWorkThreshold;
+    auto stream = at::cuda::getCurrentCUDAStream(logits.get_device());
+    bool const canUseHeuristic = compress_ratio == 1 && preIdxPtr != nullptr && logits_stride_1 == 1
+        && index_topk == tk::kHeuristicTopK && preIdxCount == tk::kHeuristicSize && preIdxStride >= preIdxCount
+        && num_columns < splitWorkThreshold && heuristicScratchPtr != nullptr;
+    int const minSplitColumns
+        = tk::computeIndexerTopKDecodeMinSplitColumns(static_cast<int32_t>(index_topk), splitWorkThreshold);
```
longcheng-nv: **This `canUseHeuristic` predicate is newly duplicated by this PR — and the two copies have already drifted within the PR itself.**

Verified against base SHA `ffe2442587`: pre-PR, `IndexerTopKOp.cpp` did **not** check the heuristic predicate (it only checked `blocks_per_row > 1` for aux allocation). The gate lived solely in `invokeIndexerTopKDecode` at base `indexerTopK.cu:742-752`. This PR introduces a second copy here so the Op can decide to skip aux allocation when heuristic applies.

The two new copies are not even strictly equivalent in source:
- Here (Op): `... && num_columns < splitWorkThreshold && heuristicScratchPtr != nullptr`
- cu side (`indexerTopK.cu:918-921`): `... && belowSplitWorkThreshold && heuristicScratch != nullptr`, where `belowSplitWorkThreshold = effectiveSplitWorkThreshold <= 0 || numColumns < effectiveSplitWorkThreshold`

The `<= 0` branch never fires at the Op call site (constant `200_000`), so the divergence is benign *today* — but linguistic drift inside a single PR is exactly the maintenance hazard you'd predict for a duplicated 8-condition predicate. The GVR path will keep gaining dispatch conditions (multi-K kernel, dtype variants), so easier to fix now.

**Suggest** extracting an inline helper in `IndexerTopK.h`:

```cpp
inline bool canUseHeuristicTopKDecode(
    int topK, int compressRatio, int stride1,
    int preIdxCount, int preIdxStride, void const* preIdx,
    void const* heuristicScratch, int numColumns, int splitWorkThreshold);
```

Op uses it to decide "skip aux allocation"; cu entry uses it to short-circuit to `launchHeuristicTopKDecode`. One source of truth.

cpp/tensorrt_llm/kernels/indexerTopK.cu
```diff
@@ -811,20 +973,16 @@ void invokeIndexerTopKPrefill(float const* logits, int const* rowStarts, int con
...
     {
-        int numRadixBlocks = numRows - kSortingAlgorithmThreshold;
-        topKPerRowPrefill<kNumThreadsPerBlock, true>
-            <<<numRadixBlocks, kNumThreadsPerBlock, topK * sizeof(int32_t), stream>>>(
-                logits, rowStarts, rowEnds, indices, stride0, stride1, topK, kSortingAlgorithmThreshold);
+        topKPerRowPrefill<kNumThreadsPerBlock, false><<<numRows, kNumThreadsPerBlock, topK * sizeof(int32_t), stream>>>(
+            logits, rowStarts, rowEnds, indices, stride0, stride1, topK, 0);
     }
 
     sync_check_cuda_error(stream);
```
longcheng-nv, line 1138: **This is a behavior change (algorithm-selection fix), not pure refactor — and it's not flagged in the PR description.**

The old code branched on row count using `kSortingAlgorithmThreshold = 12288`, which is a *column* boundary used here as a *row* threshold — clearly a bug. The new `numColumns > topK` boundary is semantically correct.

But it also changes the algorithm picked at the boundary `numRows ≪ 12288, numColumns > topK`:
- Old: insertion sort (all rows ≤ 12288 → all went through the false branch)
- New: radix sort (because `numColumns > topK`)

This could be a perf cliff for prefill shapes that previously hit the wrong-but-fast branch (small `numRows`, wide cols). Two requests:

1. Make the PR description / commit message explicitly call this out as a behavior change with bug-fix justification.
2. Add 1-2 perf rows for prefill at e.g. `numRows=64, numColumns ∈ {2048, 4096}, topK=2048` to confirm no large regression.

longcheng-nv (commented):
Reviewed end-to-end with focus on substantive correctness / perf / coverage concerns. Posted 4 inline comments + this top-level point on perf table coverage. Detailed analysis is in the inline threads.

## Perf table coverage gaps (top-level / PR description)

Table is `K ∈ {512, 1024, 2048} × BS ∈ {1, 32} × cols ∈ {131K, 196K}` = 9 cells. The policy change affects considerably more state space:

- **`BS ∈ {64, 96, 128, 148}`** — exactly the wave-fit edge and the bump-to-2 band (see inline thread on `indexerTopK.cu`). Nothing currently shows whether the win generalizes here or whether it regresses.
- **`BS=1, N≈8K`** — production GVR heuristic path on DeepSeek decode; this PR refactors the single-block dispatcher and the aux-skip logic that this path traverses. One row would prove no regression.
- **Prefill** — see inline thread on `invokeIndexerTopKPrefill`. Algorithm-selection change deserves at least 1-2 prefill perf rows.

Adding ~3-4 rows would close the loop.

Overall: solid framing on device-aware launch + wave-fit. The headline `BS=32 × cols=131K = 1.27×` wins are clean — the inline comments are about widening confidence the win generalizes (1-3) and maintainability hardening (4).

cpp/tensorrt_llm/kernels/indexerTopK.cu
mingyangHao: Addressed. I restored the bounded 2-way split for the largest power-of-two band below SM count, but updated the comment to explicitly call out that this can create  a second wave and is kept only because wide-row measurements show the reduced per-block work amortizes the merge pass. 
 Added the requested BS>SM/2 perf rows:
     BS100/128 improve ~1.8-2.2x versus no-bump; BS256 is unchanged because the bounded rule does not trigger.

longcheng-nv:
Looking at the force-pushed `fb2be5e2` (vs the prior `2ec03c9` that I'd previously cleared on substance) — the new commit makes the dispatch materially different, and I have four asks before I can re-ACK.

**1. The `numRows ∈ [smCount/2, smCount)` force-bp=2 branch reintroduces the multi-wave bump I'd originally flagged.**

The new `computeIndexerTopKDecodeBlocksPerRow` splits into `if (numRows < smCount/2) sweep` else `bp = min(2, maxByCols)`. On B200 (smCount=148) the `else` branch fires for `BS ∈ [74, 256]` and produces:
- BS=100 → 200 blocks → 1.35 waves
- BS=128 → 256 blocks → 1.73 waves
- BS=148 → 296 blocks → 2.0 waves

The new in-code comment justifies the bump with `BS=131=125us, BS=132=312us, BS=148=390us — the cliff disappears entirely with bp=2`, which establishes that **bp=1 hits a wave-scheduling cliff**, but does *not* establish that **bp=2 in this band is faster than bp=1 post-cliff**. The post-cliff bp=1 number is 390us at BS=148 — what is bp=2 at BS=100/128/148? Please add the bp=2 absolute timings to the PR description / comment so the bump is justified end-to-end, not just on cliff-avoidance.

**2. Test comment contradicts the code.**

The new comment above `test_indexer_topk_decode_launch_policy_transitions` says:
> Covers the single-block / multi-block path transition at `batch_size == kDecodeTargetTotalBlocks (= 132)`. batch_size < 132 keeps the single-block-per-row path; at and above 132 the launch policy routes to the multi-block split-and-merge path with bp = 2 …

But the actual threshold in the code is `numRows < smCount / 2`, which is 74 on B200 and 66 on H100 — not 132. Also `kDecodeTargetTotalBlocks` is deleted in this PR, so the comment references a non-existent constant. Please update the comment to match the real threshold (`smCount / 2`).

**3. Undocumented behavior change: SM-count query failure now hard-aborts.**

```diff
- int const smCount = bounds.smCount > 0 ? bounds.smCount : 132;
+ TLLM_CHECK_WITH_INFO(bounds.smCount > 0, "indexerTopK: failed to query device SM count");
+ int const smCount = bounds.smCount;
```

Previous revs had a soft 132 fallback; this rev throws. Two questions: (a) is there any test/CI environment (CPU-only sandbox, mocked CUDA) where `cudaDeviceGetAttribute(MultiProcessorCount)` can fail or return ≤0 and that previously kept working via the fallback? (b) If yes, please restore the fallback. If no, please call this out explicitly in the commit message — it is currently silent.

**4. CI has not run on `fb2be5e2`.**

The last `PR_Github` / `L0_MergeRequest_PR` pipeline ran on `2ec03c9` (#48396, completed 2026-05-14T19:48Z SUCCESS). The current head is `fb2be5e2` (force-pushed 2026-05-15T02:42Z) with a substantively different dispatch function — please request a fresh `/bot run` so we have CI signal on the actual code under review.

The PR-type bump (`[chore] → [perf]`) is appropriate given the dispatch change, but it also means this isn't a pure refactor anymore and the perf table in the PR description should be expanded accordingly. cc @mingyangHao.

Issue #13697, opened 2026-05-01
Add a AD specific config and trtllm common config via in model registry


bmarimuthu-nv:
Set all user exposed and/or trtllm common knobs in `<model>.yaml`
AD specific configs for the model in `<model>_ad_default.yaml` (hidden from the user and trtllm-serve command)

If the backend is autodeploy, it loads `<model>_ad_default.yaml`  and then `<model>.yaml`
If trtllm pytorch (let's say the model works in both workflows), it just loads `<model>.yaml`

For the user, the `trtllm-serve` command and args/config yaml stay the same between AD and pytorch flow except for adding --backend _autodeploy  for AD.

model registry should support this easily

guan404ming:
Hi @bmarimuthu-nv could I help with this issue? Thanks!

bmarimuthu-nv:
Hi @guan404ming , thanks for your interest. That would be great, feel free to share your thought/approach or just directly submit a PR. Happy to answers questions if any.

guan404ming:
Hi @bmarimuthu-nv. Thanks for kind word, really appreciate that.

For this issue I plan to split this into two PRs:
  1. First PR: add the shared model config loader and wire the PyTorch flow to load common `<model>.yaml` defaults. This PR will also add unit tests for missing files, merge order, nested config merging, and optional AD default loading behavior.
  2. Second PR: add the AutoDeploy-specific integration so AD loads `<model>_ad_default.yaml` first, then `<model>.yaml`, and add the actual model config YAML files.

This keeps the common loader/merge behavior separate from the AD-specific config changes. I just implemented the first one in #13898, please help take a look when you are available and feel free to let me know if there is any issue, thanks!

PR #13898, opened 2026-05-08
[#13697][feat] Add model YAML defaults loader

guan404ming:
## Release Notes

* **New Features**
  * Enhanced model configuration loading to merge both architecture-level YAML defaults and model-provided defaults for more comprehensive configuration.

* **Tests**
  * Added comprehensive test coverage for model configuration loading and merging functionality.

## Description

related to #13697

I add a shared loader for per-model default YAML configs.

- support loading `<model>.yaml` defaults for PyTorch flow and adds the merge behavior needed for a follow-up AutoDeploy PR to load `<model>_ad_default.yaml` before `<model>.yaml`. 
- YAML lookup currently uses the HF `architectures[0]` name as the model key.

## Test Coverage

Added unit tests for:
- missing config files
- `<model>.yaml` loading
- optional `<model>_ad_default.yaml` loading
- merge precedence
- nested dict merge behavior
- invalid top-level YAML types

## PR Checklist

## GitHub Bot Help


karljang:
@guan404ming ,
Apologies for the delayed response. I just noticed this PR.
Would you be able to fix the conflicts, so that we can proceed?
Thank you!

guan404ming:
Yes sure, not worry about that. Just update and please take another look, thanks!

bmarimuthu-nv:
Hi @guan404ming , 
Thanks for picking this up, this is a great start! The loader + deep-merge utilities and the test coverage are solid foundations. I have some design-direction feedback to share before this goes further, since we're actively evolving how model configs work across the project.

1. Config location — AD defaults should be internal
The current PR puts everything under tensorrt_llm/model_configs/. Ideally, the AD-specific configs (`<model>_ad_default.yaml`) need to be hidden from the user and from `trtllm-serve`. They should live inside the package at something like `tensorrt_llm/_torch/auto_deploy/config/` (an internal directory), while user-facing configs (common knobs like max_batch_size, kv_cache_config, etc.) stay in `examples/auto_deploy/model_registry/configs/`.

This two-layer split is a key part of the design:

Internal AD defaults: transforms, compile_backend, attn_backend — AD-specific knobs only
User-facing configs: common knobs that work the same for both PyTorch and AutoDeploy backends

2. Lookup key — use HF model name, not architecture class
The PR currently looks up configs by architectures[0] (e.g., LlamaForCausalLM). The problem is that many models share the same architecture class but need different configs — for example, `meta-llama/Llama-3.1-8B-Instruct` and `meta-llama/Llama-3.3-70B-Instruct` are both LlamaForCausalLM but may need different world_size, different transforms, etc.

There's an existing model registry at [`[examples/auto_deploy/model_registry/models.yaml](https://github.com/NVIDIA/TensorRT-LLM/blob/main/examples/auto_deploy/model_registry/models.yaml)`](https://github.com/NVIDIA/TensorRT-LLM/blob/main/examples/auto_deploy/model_registry/models.yaml) with 200+ models. Each entry maps a HuggingFace model name to a list of composable YAML config files via `yaml_extra`:

```yaml
- name: meta-llama/Llama-3.3-70B-Instruct
  yaml_extra: ['dashboard_default.yaml', 'world_size_4.yaml', 'llama3_3_70b.yaml']
```

The new loader should extend this existing registry pattern rather than creating a separate loading path. The issue (#13697) asks us to split configs into two layers **within this registry** — AD-internal defaults vs user-facing common knobs — not to create a parallel lookup mechanism.

3. Integration — needs to wire into `serve.py` and `build_and_run_ad.py`
The current PR only integrates into `model_loader.py.` For this to work end-to-end, the registry lookup needs to be consumed by:

* trtllm-serve (tensorrt_llm/commands/serve.py): auto-inject AD defaults as yaml_extra when --backend _autodeploy is used
* build_and_run_ad.py: the --use-registry flow should pull from the same package-level registry
* Integration tests: same registry, not reimplemented lookup logic

The yaml_extra composition mechanism already exists in the codebase — the registry should produce a list of YAML file paths that get merged through that existing pipeline, rather than introducing a separate deep-merge path in model_loader.py.

4. Parallel field translation
One thing that's missing: when a user provides tensor_parallel_size in their config (which is the standard TRT-LLM field), AutoDeploy currently rejects it because it expects world_size. The registry layer should handle this translation transparently so users can use the same config YAML for both backends.

5. Security — sanitize arch (from CodeRabbit review)
The CodeRabbit suggestion about sanitizing the architecture string before constructing file paths is valid and worth keeping if any arch-based lookup remains.

Summary: The core loading/merge utilities you've built are useful, but the PR needs to evolve into a centralized model registry with the two-layer config split described in the issue. This is a larger scope than what's here — happy to discuss the further if you'd like to continue working on it. Let me know if you have questions!

guan404ming:
Thanks, that makes sense. Agreed this should extend the existing registry with the two-layer split rather than add a separate loader. I think the merge order I have is fine, it's the lookup key (arch) and the file location that are off.

So I'm clear on the rework: each model gets a `<model>_ad_default.yaml` for the AD-internal knobs that ships inside
the pkg and stays hidden from `trtllm-serve`, plus a `<model>.yaml` for the common knobs both backends share. AD loads ad_default then model, PyTorch loads just model, user `--config` still wins over both. Lookup goes through models.yaml by HF name instead of architectures[0]. I'd reuse the merge util from this PR and drop the arch loader and the `tensorrt_llm/model_configs/` dir.

Few things I want to confirm that:
1. Does models.yaml itself need to move into the package too, so serve.py and model_loader.py can resolve it at runtime? Only the AD-internal yamls would be hidden.
2. For the two layers, do you want two yaml_extra-style lists on each entry so we keep the composable mechanism, or literally two files per model?

On scope, I think this is cleaner as two PRs:

- First one moves the resolver (get_registry_yaml_extra / _merge_yaml_files) out of build_and_run_ad.py into a package module, adds the two-layer dir plus wheel packaging for a handful of models to start, and points --use-registry at the shared resolver. No behavior change.
- Second one wires it into serve.py (inject ad_default when --backend _autodeploy) and model_loader.py (HF-name lookup), adds the field translation, migrates the rest of the configs.

I will keep the first PR a straight refactor and puts the actual behavior changes in the second. Please feel free to let me know if that split works or if you'd rather keep it as one, thanks!

bmarimuthu-nv:
Hi @guan404ming yes,
> Does models.yaml itself need to move into the package too, so serve.py and model_loader.py can resolve it at runtime? Only the AD-internal yamls would be hidden.

Yes, this makes sense. We would need this.

> For the two layers, do you want two yaml_extra-style lists on each entry so we keep the composable mechanism, or literally two files per model?

We can just put the AD default file before the model config in the same yaml_extra list:
```
  yaml- name: meta-llama/Llama-3.3-70B-Instruct
  yaml_extra: ['dashboard_default.yaml', 'llama3_3_70b_ad.yaml', 'llama3_3_70b.yaml']
```

The `_ad.yaml` files live inside the package (e.g., under `tensorrt_llm/_torch/auto_deploy/config/model_registry_internal/configs/`) so users don't see them, while user-facing configs stay in `examples/auto_deploy/model_registry/configs/`. The existing last-file-wins merge handles precedence. The registry loader just needs to resolve files from two directories.


I think it's okay to make it a single PR where we do all of these in one go. I'll help with the merging.

guan404ming:
Hi @bmarimuthu-nv, really appreciate your reply. I just finished the implementation
- models.yaml + the AD-internal `_ad.yaml` configs now ship in-package under `tensorrt_llm/_torch/auto_deploy/config/model_registry_internal/`
- lookup is by HF name through a shared resolver wired into `serve.py`, `build_and_run_ad.py`
- `tensor_parallel_size` is translated to `world_size` and the old arch path is gone
  
Currently I split the 15 active per-model configs (each round-trip verified) and left out the shared fragments (dashboard_default, world_size_N, ...) and commented-out entries. Would you like me to split those as well in this PR, or would you prefer to keep the scope to the active per-model configs for now? Happy to go either way and also feel free to let me know if there is anything could be better, thanks!

bmarimuthu-nv:
Thanks @guan404ming, could you rebase on the latest main and push the branch, I can kick off tests

guan404ming:
Yeah sure, just rebased and resolved the conflicts, thanks!

guan404ming:
Hi @bmarimuthu-nv I think your review is toward this pr. Here is some update:
  
- Reverted the model_loader.py changes, PyTorch path is untouched.
- Moved model_config_loader.py into tensorrt_llm/_torch/auto_deploy/ 
- Dropped load_model_common_defaults since its only consumer was the reverted path.

Everything now lives in the AD subsystem, with serve.py's `_autodeploy` branch as the only outside touchpoint. Thanks!

tensorrt_llm/commands/serve.py
```diff
@@ -370,9 +370,12 @@ def launch_server(
             llm = PyTorchLLM(**llm_args)
         elif backend == '_autodeploy':
             from tensorrt_llm._torch.auto_deploy import LLM as AutoDeployLLM
+            from tensorrt_llm._torch.auto_deploy.model_config_loader import \
+                inject_autodeploy_registry_defaults
 
             # AutoDeploy does not support build_config
             llm_args.pop("build_config", None)
+            inject_autodeploy_registry_defaults(llm_args)
```
bmarimuthu-nv: @guan404ming This logic (parallel field translation + prepending internal yaml_extra) is purely AD-internal — serve.py shouldn't need to know about the registry at all.
Can we push inject_autodeploy_registry_defaults down into the AutoDeploy LLM class itself (either in AutoDeployLLM.__init__ or the AD llm_args preprocessing layer). That way the AD subsystem handles its own defaults internally, and serve.py stays untouched.

guan404ming: That makes sense to me. Just moved the injection into AutoDeployLLM.__init__ (applied to kwargs before super().__init__, so it still runs ahead of the yaml_extra settings-source resolution). serve.py is now untouched relative to main.

bmarimuthu-nv:
Thanks @guan404ming , I see some regression tests failing. Can you add these fixes:

1. Fix test_mrope_delta_cache.py: Update test_qwen_registry_configs_explicitly_enable_mrope_delta_cache to read from the new _ad.yaml files in tensorrt_llm/_torch/auto_deploy/config/model_registry_internal/configs/ (e.g., qwen3.5_moe_35b_ad.yaml) instead of the now-stripped user-facing YAML files, OR add a test helper that merges _ad.yaml + the user-facing YAML before asserting on the merged config.

2. Fix resolve_registry_yaml_extra path resolution: The new function must also search for user-facing shared configs (like dashboard_default.yaml, world_size_*.yaml) in examples/auto_deploy/model_registry/configs/ relative to the source tree (as the old get_registry_yaml_extra did with Path(__file__).resolve().parent / "model_registry" / "configs"). The installed dist-packages path does not contain the examples/ directory, so the current fallback path …/dist-packages/examples/auto_deploy/model_registry/configs is incorrect.

3. Verify gemma3n_e2b_it test with merged config loading: After fixing (2), re-run test_gemma3n_e2b_it to confirm the executor error was caused by missing attn_backend: flashinfer from the unloaded _ad.yaml. If it still fails, investigate executor logs for a separate root cause.

4. Add integration test for the new two-layer loading: Add a test that verifies resolve_registry_yaml_extra can correctly locate and merge both the _ad.yaml (internal) and user-facing YAML files for a known model like meta-llama/Llama-3.1-8B-Instruct.

guan404ming:
@bmarimuthu-nv All four addressed:
  
  1. mrope test now reads the internal _ad.yaml files via INTERNAL_CONFIGS_DIR.
  2. Added user_dirs param to resolve_registry_yaml_extra; integration tests and build_and_run_ad.py pass their source-tree configs dir, so installed packages no longer hit the bogus dist-packages/examples/ path.
  3. With (2), gemma3n resolves gemma3n_e2b_it_ad.yaml (attn_backend: flashinfer) again; will confirm e2e in the next CI run.
  4. Added a two-layer loading test (test_resolve_yaml_extra_real_registry_two_layers) plus a user_dirs override test.

bmarimuthu-nv:
@guan404ming could rebase again, there seems to be some conflicts

xinhe-nv (approved):

guan404ming:
@bmarimuthu-nv 
Just rebased and resolved the conflicts, thanks!

bmarimuthu-nv:
@guan404ming can you please rebase again!

guan404ming:
Yeah, sure~

guan404ming:
Hi @bmarimuthu-nv, could you help take another look at this, really appreciate your time!

PR #13919, opened 2026-05-08
[TRTLLM-12339][feat] Support T5 and BART in the PyTorch backend

cascade812:
## Description
This PR adds the core encoder-decoder request flow, including encoder execution, decoder context/generation handling, cross-attention metadata, and cross-KV cache management.

## Key Changes

- Add PyTorch backend support for T5 model execution.
- thop.attention support for encoder-decoder cross-attention.
- Add dual KV cache pool support:
  - self-attention KV cache for decoder tokens
  - cross-attention KV cache for encoder outputs
- Add precision conversion support for encoder-decoder model weights.
- Support `beam_width > 1` for encoder-decoder generation.
- Plumb encoder-side inputs through LLM API request handling and executor flow.
- Add unit and integration coverage for encoder-decoder request handling, scheduling, KV cache behavior, and T5 execution.

Below features are not supported in this PR, will come in separate PR. 
- cuda graph
- TP

## Test Coverage
- Added T5 PyTorch integration tests.
- Added encoder-decoder request API tests.
- Added dual-pool KV cache and scheduler unit tests.

## PR Checklist

## GitHub Bot Help

## Summary by CodeRabbit

* **New Features**
  * Added full encoder–decoder model support (BART, mBART, T5) and cross‑attention paths.
  * Exposed encoder_tokens and encoder output handling in the public APIs for encoder→decoder workflows.
  * Added relative position bias support in attention.

* **Improvements**
  * Dual‑pool KV‑cache management and coordinated scheduling for self/cross attention to improve encoder‑decoder admission and budgeting.
  * Scheduler/executor now handle encoder phases, readiness gating, and encoder streaming.

* **Tests**
  * New unit and integration tests covering encoder‑decoder, dual‑pool KV cache, and scheduling.


pengbowang-nv:
Hi, I'd first like to know if it is possible that we may split this single PR into multiple smaller PRs, so as to reduce the difficulty for reviewer and change scope? Thanks!

cascade812:
> Hi, I'd first like to know if it is possible that we may split this single PR into multiple smaller PRs, so as to reduce the difficulty for reviewer and change scope? Thanks!

I agree that smaller PRs would be easier for review. But it would be quite difficult to split this PR at this stage, because the changes are not cleanly separable into self-contained pieces.

cpp/tensorrt_llm/batch_manager/capacityScheduler.cpp
```diff
@@ -319,7 +332,46 @@ std::tuple<RequestVector, RequestVector> GuaranteedNoEvictScheduler::impl(
...
+                if (isEncoderInit)
+                {
+                    // Encoder admission does not reserve self- or cross-pool
+                    // blocks.  Without a cross manager the dual-pool contract
+                    // cannot be satisfied later by decoder context, so surface
+                    // this and skip rather than silently admitting a request
+                    // that cannot complete.
+                    if (!reservedCrossBlocks)
+                    {
+                        TLLM_LOG_WARNING(
```
mikeiovine: Should this be a hard error instead of a warning?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -52,6 +54,53 @@
 # Default KV layout for flashinfer
 # HND = [max_num_pages, kv_factor, num_kv_heads, page_size, head_dim]
 DEFAULT_KV_LAYOUT = "HND"
+_RELATIVE_ATTENTION_BIAS_KWARGS = (
+    "relative_attention_bias",
+    "relative_attention_max_distance",
+)
+
+
+@lru_cache(maxsize=None)
```
mikeiovine: No need for LRU cache, this version of python should just support `@functools.cache`

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -52,6 +54,53 @@
...
+    if not IS_FLASHINFER_AVAILABLE:
+        return False
+
+    try:
+        context_params = inspect.signature(
+            flashinfer.prefill.trtllm_batch_context_with_kv_cache
+        ).parameters
+        decode_params = inspect.signature(
+            flashinfer.decode.trtllm_batch_decode_with_kv_cache
+        ).parameters
```
mikeiovine: I am not a big fan of this - do we need to update the min flashinfer version?

tensorrt_llm/_torch/models/modeling_t5.py
```diff
@@ -0,0 +1,1089 @@
...
+        return act_fn(gate) * up
+
+    return gated_act_fn
+
+
+def _clamp_fp16_infs(hidden_states: torch.Tensor) -> torch.Tensor:
+    if hidden_states.dtype != torch.float16:
+        return hidden_states
+
+    return torch.clamp(hidden_states, min=-64000.0, max=64000.0)
```
mikeiovine: Is this required? Curious if we put it here for debugging or if the model genuinely needs it for numerical stability

tensorrt_llm/_torch/modules/encoder_decoder_layer.py
```diff
@@ -0,0 +1,55 @@
...
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        attn_metadata: AttentionMetadata,
+        position_ids: Optional[torch.IntTensor] = None,
+        **kwargs,
+    ) -> torch.Tensor: ...
+
+
+class EncoderDecoderLayer(nn.Module, ABC):
```
mikeiovine: Having a well defined interface is a good idea but I think the idiom for most existing modeling code is just to inherit from `nn.Module` and create an ad hoc `forward` signatures with whatever that particular model happens to need. We can probably get rid of these - not much advantage in having layers for distinct modules share a class hierarchy.

mikeiovine (commented):
Only partial review so far - still need to check the files under pyexecutor/

tensorrt_llm/_torch/models/modeling_t5.py
cascade812: Yes, this is for T5 fp16 numerical stability rather than debugging. HF T5 applies the same guard after residual additions. Just updated it to mirror HF’s exact clamp logic instead of using a hard-coded 64000.

tensorrt_llm/_torch/modules/encoder_decoder_layer.py
cascade812: removed

cpp/tensorrt_llm/batch_manager/capacityScheduler.cpp
cascade812: I changed this from a warning/skip path to TLLM_CHECK_WITH_INFO() so it becomes a hard error, and updated the Python scheduler paths/tests for the same behavior.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
cascade812: I removed the runtime probing and made relative attention bias explicitly unsupported for trtllm-gen because relative attention bias is not supported by any flashinfer-python version.

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
mikeiovine, line 1558: Consider checking `is_decoder_context_request_ready` here for consistency with PyMicroBatchScheduler. Had to double check the correctness of this; I think it's fine because of the filter, so this suggestion is more for code clarity

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -395,6 +400,8 @@ def __init__(
         # kv cache events
         self.kv_cache_manager = self.resource_manager.resource_managers.get(
             ResourceManagerType.KV_CACHE_MANAGER)
+        self.cross_kv_cache_manager = self.resource_manager.resource_managers.get(
+            ResourceManagerType.CROSS_KV_CACHE_MANAGER)
```
mikeiovine: This appears to be unused

tensorrt_llm/_torch/pyexecutor/llm_request.py
mikeiovine, line 349: Coderabbit was complaining about needing to move encoder output here. But I think we can revisit that when (if?) we add pipeline parallelism

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -595,6 +602,26 @@ def on_detected():
...
+                raise NotImplementedError(
+                    "pp_size > 1 is not supported for encoder-decoder models "
+                    "in the PyTorch flow; encoder send/recv hooks are out of "
+                    "scope. Set pp_size=1 to run T5/BART/mBART.")
+            if not self.disable_overlap_scheduler:
+                raise NotImplementedError(
+                    "Overlap scheduler is not yet wired for encoder-decoder "
+                    "models. Set disable_overlap_scheduler=True for "
+                    "encoder-decoder runs.")
+
```
mikeiovine, line 702: Might want to add one for CUDA graphs

tensorrt_llm/_torch/attention_backend/vanilla.py
```diff
@@ -364,16 +407,75 @@ def no_kv_cache_forward(
...
+            v_s = repeat_kv(v_s, num_kv_groups)
+
+            qk_scale = None
+            if self.q_scaling is not None:
+                qk_scale = 1 / (math.sqrt(head_dim) * self.q_scaling)
+
+            # SDPA's is_causal flag implies square attention. For
+            # cross-attention (different Q/K lengths) we never apply a causal
+            # mask: the decoder Q attends to all encoder K/V tokens.
+            sdpa_is_causal = is_causal and (end_q - start_q) == (end_k -
```
brb-nv: Dumb question: Won't there be a case where `(end_q - start_q) == (end_k - start_k)` even for enc-dec models? Should we rather gate on `metadata.is_cross` here?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1343,12 +1345,26 @@ def _run(
                                  or metadata.max_seq_len)
         max_context_length = min(metadata.max_seq_len - 1,
                                  metadata.max_num_tokens)
+        relative_attention_bias = forward_args.relative_attention_bias
+        relative_attention_max_distance = forward_args.relative_attention_max_distance
+        has_relative_attention_bias = relative_attention_bias is not None
+        position_embedding_type = (int(PositionEmbeddingType.relative)
```
brb-nv: Why do we need to override here? Why can't this flow down from upstream with the right value to begin with?

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -398,6 +403,78 @@ def update_helix_param(
...
+                defaults to 0 (context phase, nothing cached yet).
+
+        Returns:
+            A new ``AttentionMetadata`` of the same subclass as ``self``,
+            with ``seq_lens_kv`` set to ``encoder_seq_lens`` so that
+            ``is_cross`` becomes ``True``.
+        """
+        cross_md = copy.copy(self)
+        cross_md._saved_tensors = {}
+        if self.is_cuda_graph:
```
brb-nv, line 474: How often is this function called? Only a few times during warmup? Or more often than that?

brb-nv (approved):
Mostly information questions.

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -680,16 +715,16 @@ void attention(torch::Tensor q, std::optional<torch::Tensor> k, std::optional<to
 
     bool const use_sage_attn
         = sage_attn_num_elts_per_blk_q > 0 || sage_attn_num_elts_per_blk_k > 0 || sage_attn_num_elts_per_blk_v > 0;
-    TLLM_CHECK_WITH_INFO(is_mla_enable || is_fused_qkv || use_sage_attn,
-        "Context attention only allows these non-MLA cases: fused QKV; separate QKV with SageAttention");
-    TLLM_CHECK_WITH_INFO(update_kv_cache, "KV cache update cannot be disabled now");
+    TLLM_CHECK_WITH_INFO(is_mla_enable || is_fused_qkv || use_sage_attn || cross_attention,
+        "Only fused QKV is supported for non-MLA non-cross attention now");
+    TLLM_CHECK_WITH_INFO(update_kv_cache || cross_attention, "KV cache update cannot be disabled now");
```
pengbowang-nv: Does the info and condition mismatch?

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1343,12 +1345,26 @@ def _run(
...
+                                   if has_relative_attention_bias else
+                                   self.position_embedding_type)
 
         helix_active = metadata.helix_position_offsets is not None
+        encoder_seq_lens_arg = (metadata.kv_lens_cuda_runtime
+                                if metadata.is_cross else None)
+        # Cross-attention treats decoder beams as already-expanded rows and
+        # reads request-scoped encoder K/V, so kernel beam indirection stays off.
+        kernel_beam_width = 1 if metadata.is_cross else metadata.beam_width
+        prefer_trtllm_gen = _TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION or metadata.is_cross
```
pengbowang-nv: I think maybe it's better to assert against _TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION is on for cross case, as it is an environment variable controlled by user.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1461,12 +1479,38 @@ def _run(
...
+            if metadata.is_cross and k is not None and v is not None:
+                k_flat = k.contiguous().view(k.shape[0], -1)
+                v_flat = v.contiguous().view(v.shape[0], -1)
+                cross_kv_input = torch.cat([k_flat, v_flat], dim=1).contiguous()
+            k_arg = None if metadata.is_cross else k
+            v_arg = None if metadata.is_cross else v
+            q_arg = q
+            is_fused_qkv_arg = is_fused_qkv
+            legacy_attention_kwargs = {}
+            if metadata.is_cross:
```
pengbowang-nv: A dump question: are we faking that input is a fused-QKV tensor here? Why is that needed?

cpp/tensorrt_llm/batch_manager/capacityScheduler.cpp
```diff
@@ -305,9 +312,25 @@ std::tuple<RequestVector, RequestVector> GuaranteedNoEvictScheduler::impl(
...
+                    crossSummary = crossKvCacheManager->analyzePrefixReuse(uniqueTokens, *req);
+                }
+                if (isEncoderInit)
+                {
+                    // Encoder admission does not reserve self- or cross-pool
+                    // blocks. Without a cross manager the dual-pool contract
+                    // cannot be satisfied later by decoder context, so fail
+                    // fast instead of admitting a request that cannot complete.
+                    TLLM_CHECK_WITH_INFO(reservedCrossBlocks.has_value(),
+                        "Encoder-init request %lu requires a cross_kv_cache_manager.", req->mRequestId);
```
Funatiq: Do we need to check this per request? This seems more like a requirement that we should verify globally.

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
```diff
@@ -61,6 +61,33 @@ def _call_with_optional_summary(
...
+        if is_decoder_context_request_ready(req):
+            filtered_requests.append(req)
+            continue
+
+        logger.debug(
+            "Skipping context request %s until encoder output is ready.",
+            getattr(req, "py_request_id", req.request_id),
+        )
+
+    return filtered_requests
```
Funatiq, line 89: It seems like we are describing a state between ENCODER_INIT and CONTEXT_INIT here. Wouldn't it make more sense to introduce an explicit request state (e.g. ENCODER_IN_PROGRESS) for this? Then the scheduler could just ignore requests in this state.

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
```diff
@@ -623,8 +677,16 @@ def get_lora_task_id(req: LlmRequest):
 
         if chunks_present:
             # Partition: non-last-chunk first, last-chunk at end
-            not_last_chunk = [r for r in context_requests if not r.is_last_context_chunk]
-            last_chunk = [r for r in context_requests if r.is_last_context_chunk]
+            not_last_chunk = [
+                r
+                for r in context_requests
+                if r.is_encoder_init_state or not r.is_last_context_chunk
+            ]
```
Funatiq: Is this correct? Do we want to mix encoder requests and chunked context requests here?
Shouldn't the encoder requests be separated from the context requests completely?

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
```diff
@@ -901,6 +963,17 @@ class GuaranteedNoEvictPolicy(SchedulerPolicyBase):
...
+    Encoder-decoder support: when ``cross_kv_cache_manager`` is configured
+    on the parent scheduler and ``no_schedule_until_state=ENCODER_INIT``,
+    encoder-init requests are considered in the same *scheduler pass* as
+    context/generation requests. This does not mean encoder and decoder
+    context execute in the same model iteration: encoder admission only
+    admits encoder compute and leaves both KV pools untouched until the
+    request transitions to ``CONTEXT_INIT`` on a later decoder-context
+    iteration. The in-pass classification preserves the legacy invariant
+    that encoder and decoder context never collide on the same self pool
+    budget.
```
Funatiq, line 1006: It's not entirely clear to me why we check the `reserved_cross_blocks` for ENCODER_INIT requests while it states here that we do not need to touch the KV pools.

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -251,32 +257,44 @@ def _get_model_kv_cache_manager_cls(self, model_engine: PyTorchModelEngine):
                 )
         return cls
 
-    def _per_manager_cache_cost(self, manager_cls, model_config,
+    def _per_manager_cache_cost(self,
+                                manager_cls,
+                                model_config,
+                                kv_cache_config: Optional[KvCacheConfig] = None,
                                 **extra_kwargs) -> CacheCost:
```
Funatiq, line 304: nit: Add a comma after the last parameter for consistency.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2416,6 +2443,23 @@ def _executor_loop(self):
 
                 finished_requests = []
 
+                # Split off encoder-init requests before any decoder-side
+                # preparation so the self-pool ``prepare_resources`` and
+                # the decoder forward step never see them. Decoder context is
+                # dispatched in a later iteration, so encoder admission does
+                # not need cross-pool blocks for same-iteration decoder work.
+                encoder_requests = self._split_encoder_decoder_context_requests(
+                    scheduled_batch)
```
Funatiq: The cleaner solution would be to add a separate list for encoder requests to `ScheduledRequests` and fill it in the schedulers. I know it would require more refactoring, but I believe it's the right thing to do.

cascade812: Agreed. I've separated encoder requests from decoder context requests at the scheduler output boundary. context_requests now means decoder context only, and encoder-init requests are carried through encoder_requests.

tensorrt_llm/_torch/pyexecutor/py_executor.py
cascade812: Added

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
cascade812: This is effectively a transient readiness condition between encoder work being queued and decoder context being allowed to consume its output. I considered an explicit request state, but LlmRequestState is a shared C++ enum with range-based scheduler semantics, so adding `ENCODER_IN_PROGRESS` would be a larger cross-language state-machine change and would still need scheduler special-casing. Based on it, I think the better way is to keep the state as `CONTEXT_INIT` and make the readiness gate explicit with `drop_decoder_context_requests_waiting_for_encoder_output`.

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
cascade812: As I separated encoder requests and context requests, this change has been reverted.

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
cascade812: Good catch. I changed this so ENCODER_INIT only verifies that a cross_kv_cache_manager exists. It no longer checks or reserves cross-KV blocks at encoder admission time.

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
cascade812: Added. Btw, I also renamed `is_decoder_context_request_ready` into `is_decoder_context_request_waiting_for_encoder_output` to avoid confusion. Based on its implementation, 'is_decoder_context_request_waiting_for_encoder_output' is a better name without causing confusion.

tensorrt_llm/_torch/pyexecutor/_util.py
cascade812: I checked the surrounding signatures in this file and the local style is mixed: several nearby aligned multi-line signatures don’t use a trailing comma after the last parameter. I’ll keep this one consistent with the nearby aligned style for now, unless you’d prefer that I reformat the whole signature to the vertical style with trailing commas.

tensorrt_llm/_torch/pyexecutor/llm_request.py
cascade812: For now, I updated PyResult.get_diff() to normalize diff.encoder_output via detach().cpu() before returning the diff. I’ll revisit this path when adding PP support.

tensorrt_llm/_torch/attention_backend/trtllm.py
cascade812: This is intentional for the T5 cached path. T5 does not set `pos_embd_params` on the generic attention module because its relative position handling is a learned attention-bias table, not a RoPE-style backend positional embedding. That learned table is passed as `relative_attention_bias` at forward time.

The TRTLLM kernel needs `PositionEmbeddingType.relative` exactly when `relative_attention_bias` is provided, so this local override keeps those two inputs coupled. We could add another forward arg for `position_embedding_type`, but today it would just duplicate `relative_attention_bias is not None`, so I’d prefer to keep this localized unless another caller needs those values to differ.

tensorrt_llm/_torch/attention_backend/vanilla.py
cascade812: Yes, that could happen. I updated this to `(is_causal and not is_cross and (end_q - start_q) == (end_k - start_k))`

tensorrt_llm/_torch/attention_backend/interface.py
cascade812: it is called once per decoder batch forward without cuda graph, not per layer; with cuda graph, cross metadata is created once, and then reused on later forwards.

tensorrt_llm/_torch/pyexecutor/scheduler/scheduler.py
```diff
@@ -114,12 +173,20 @@ def context_requests(self) -> RequestList:
...
     def append_context_request(self, request: LlmRequest) -> None:
         if request.is_last_context_chunk:
             self.context_requests_last_chunk.append(request)
         else:
             self.context_requests_chunking.append(request)
 
+    def reset_encoder_requests(self, encoder_requests: RequestList | None = None) -> None:
+        self.encoder_requests = (
+            encoder_requests if encoder_requests is not None else self.encoder_requests
+        )
```
Funatiq: It doesn't look like this function is used anywhere. We can remove it. `encoder_requests` can be reset without the function. `reset_context_requests` only exists because of the two separate lists for context requests.

cascade812: Removed.

cpp/include/tensorrt_llm/batch_manager/capacityScheduler.h
```diff
@@ -115,14 +126,14 @@ class GuaranteedNoEvictScheduler : public BaseCapacityScheduler
 
     [[nodiscard]] std::tuple<RequestVector, RequestVector> operator()(
         kv_cache_manager::BaseKVCacheManager const& kvCacheManager,
-        OptionalRef<kv_cache_manager::BaseKVCacheManager const> crossKvCacheManager,
```
eopXD, line 118: Why are we dropping the `const` here?

cpp/tensorrt_llm/batch_manager/capacityScheduler.cpp
```diff
@@ -364,30 +407,48 @@ std::tuple<RequestVector, RequestVector> GuaranteedNoEvictScheduler::impl(
...
+    if (crossKvCacheManager)
+    {
+        scheduledCrossBlocksManager.emplace(*crossKvCacheManager, mTwoStepsLookAhead);
+    }
     SizeType32 numScheduledPeftPages{0};
     std::unordered_set<uint64_t> seenTaskIds;
 
     // Keep track of blocks contributed by requests in context phase
     auto [newlyContributedContextBlocks, newlyContributedCrossContextBlocks]
-        = prefillWithChunkedContextsAlreadyExecuting(activeRequests, kvCacheManager);
```
eopXD, line 388: Why are we passing `crossKvCacheManager` here? Does not look relevant to the goal towards this file, which is to introduce the capacity scheduler to the ENCODER_INIT state of the request.

cpp/tests/unit_tests/batch_manager/capacitySchedulerTest.cpp
```diff
@@ -2268,3 +2268,189 @@ TEST_F(CapacitySchedulerTest, MaxUtilizationNoReuseWhenDisabled)
...
+    auto req = std::make_shared<LlmRequest>(reqId, executorReq);
+    // executor::Request ctor sets kENCODER_INIT when encoderInputTokenIds is present.
+    EXPECT_EQ(req->getState(), LlmRequestState::kENCODER_INIT);
+    return req;
+}
+} // namespace
+
+// GuaranteedNoEvict: a single encoder-init request is admitted without
+// consuming self- or cross-pool blocks.
+TEST_F(CapacitySchedulerTest, EncoderInitGuaranteedNoEvictAdmits)
```
eopXD: `EncoderInitGuaranteedNoEvictAdmits` and `EncoderInitMaxUtilizationAdmits` are trivial.
The latter two tests supersets them. Possibly prune them?

cpp/tensorrt_llm/common/attentionOp.cpp
```diff
@@ -1755,8 +1755,9 @@ int AttentionOp::enqueueContext(EnqueueContextParams<T> const& params, cudaStrea
         preprocessingParams.qkv_bias = params.qkv_bias;
         preprocessingParams.tokens_info = decoder_params.tokensInfo;
         preprocessingParams.seq_lens = params.context_lengths;
-        // Indicate if chunked-context is used (i.e. q_seqlen > kv_seqlen).
-        preprocessingParams.cache_seq_lens = params.sequence_lengths;
+        // For cross-attention this is the decoder-side length used by the preprocessing
+        // kernel to decide whether to store encoder K/V into the cross-KV cache.
+        preprocessingParams.cache_seq_lens = isCrossAttention() ? params.context_lengths : params.sequence_lengths;
```
eopXD, line 1742: I am suspicious on why do you need the change here.
Does this change breaks the contract from the comment you have removed, which the variable "indicate if chunked-context is used (i.e. q_seqlen > kv_seqlen)."?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -60,7 +61,7 @@
 from .hang_detector import HangDetector
 from .kv_cache_transceiver import KvCacheTransceiver
 from .llm_request import (ExecutorRequest, LlmRequest, LlmRequestState,
-                          LlmResponse)
+                          LlmResponse, get_draft_token_length)
```
eopXD: Unnecessary addition here of `get_draft_token_length`?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -682,7 +683,37 @@ def prepare_resources(self, scheduled_batch: ScheduledRequests):
...
+                if batch_request_infos:
+                    self.impl.add_sequence_batch(batch_request_infos,
+                                                 batch_llm_requests)
+
+                # Cross KV is written once from encoder K/V projection and
+                # then remains fixed for decoder generation.
+                self.impl.refresh_blocks()
+                return
+
+            # Collect first-chunk requests eligible for batch add_sequence.
```
eopXD: Why changing `add_sequence_batch` to `add_sequence`?

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -898,6 +929,16 @@ def update_resources(self,
                          scheduled_batch: ScheduledRequests,
                          attn_metadata: "AttentionMetadata" = None,
                          kv_cache_dtype_byte_size: float = None):
+        if self.kv_cache_type == CacheTypeCpp.CROSS:
+            for request in scheduled_batch.context_requests:
+                self.impl.store_context_blocks(request)
+            return
+
+        if not self.is_draft:
```
eopXD: There is no draft token for the encoder and cross K/V. May you re-evaluate the addition.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -898,6 +929,16 @@ def update_resources(self,
                          scheduled_batch: ScheduledRequests,
                          attn_metadata: "AttentionMetadata" = None,
                          kv_cache_dtype_byte_size: float = None):
+        if self.kv_cache_type == CacheTypeCpp.CROSS:
```
eopXD: Same comment as the `if` statement section above for `prepare_resources`. Please re-evaluate how to approach this.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -682,7 +683,37 @@ def prepare_resources(self, scheduled_batch: ScheduledRequests):
             # wait for all pending work to finish before launching offload/onboarding/partial copy
             self.impl.sync_transfer_manager_with_buffer_manager()
 
-            # Collect first-chunk requests eligible for add_sequence_batch.
+            if self.kv_cache_type == CacheTypeCpp.CROSS:
```
eopXD: I don't think the addition here is necesary. The same calls already exist in the function.
Plus, we don't expect generation requests in the scheduled batch in the cross KV manager.

Adding an assertion to guard your expectation should be better than adding a complete copy of the same code that bloats the codebase.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -1784,6 +1825,33 @@ def pin_blocks(self, request_id: int):
     def copy_batch_block_offsets(self, dst_tensor: torch.Tensor,
                                  request_ids: List[int], beam_width: int,
                                  num_context: int, num_seqs: int):
+        if self.kv_cache_type == CacheTypeCpp.CROSS and beam_width > 1:
```
eopXD, line 2072: Amend with more context:
```
  # This branch is invoked via cross_md.kv_cache_manager aliasing — there's no direct
  # cross_kv_cache_manager.copy_batch_block_offsets(...) call in the codebase.
```

tests/integration/defs/llmapi/test_llm_api_pytorch_t5.py
```diff
@@ -0,0 +1,526 @@
...
+        num_return_sequences,
+        exact_match,
+        id=f"{feature_id}-{model_name}",
+        marks=marks,
+    )
+
+
+_TEST_CASES = [
+    # Primary coverage: v1 cache manager and beam search.
+    _test_case(
```
eopXD, line 146: Do you mind styling the tests cases like. Should be quick in the agentic era.
```
_test_case(
 /* param1= */ XXX,
 /* param2 = */ XXX,
 ...
)
```
That will increase readability much more.

tests/unittest/llmapi/test_llm_args.py
```diff
@@ -340,6 +340,20 @@ def test_KvCacheConfig_declaration():
     assert pybind_config.attention_dp_events_gather_period_ms == 10
 
 
+def test_KvCacheConfig_rejects_cross_kv_cache_fraction_below_zero():
```
eopXD: Remove the following two trivial tests.

tests/unittest/_torch/modeling/test_modeling_enc_dec.py
```diff
@@ -0,0 +1,1351 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
eopXD: Does this test just directly overlaps with how `tests/unittest/llmapi/test_encoder_decoder_request_api.py` is testing e2e regarding the encoder-decoder model?

tests/unittest/_torch/executor/test_request_utils.py
```diff
@@ -57,6 +61,54 @@ def create_mock_request_with_py_schedule_params(attention_dp_rank=None, attentio
     return mock_request
 
 
+def test_executor_request_to_llm_request_preserves_encoder_tokens():
```
eopXD: e2e should test cover this simple conversion.

tests/unittest/_torch/executor/test_dual_pool_kv_cache.py
```diff
@@ -0,0 +1,906 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
eopXD, line 1: Please audit the file and remove decorative attribute / mock-shape assertions

Funatiq (approved):

tensorrt_llm/llmapi/llm_args.py
eopXD, line 2797: default to none, docstring says 50%.

cpp/tensorrt_llm/thop/attentionOp.cpp
```diff
@@ -487,6 +511,15 @@ class Runner : public RunnerBase
...
                 enqueue_params.v_stride_in_bytes = v->strides()[0] * v->element_size();
             }
+            if (cross_attention && cross_kv.has_value() && encoder_input_lengths.has_value())
+            {
+                auto const& cross_kv_tensor = cross_kv.value();
+                auto const& enc_lens = encoder_input_lengths.value();
+                enqueue_params.cross_kv = static_cast<T const*>(cross_kv_tensor.data_ptr());
+                enqueue_params.num_encoder_tokens = static_cast<int32_t>(cross_kv_tensor.size(0));
+                enqueue_params.cross_kv_length
+                    = enc_lens.slice(0, seq_offset, seq_offset + num_seqs).max().item<int32_t>();
```
eopXD: The `.item` call introduces a GPU->CPU sync, can we avoid this?

pengbowang-nv (commented):
For thop op and CPP, attention backend part.

cpp/include/tensorrt_llm/batch_manager/capacityScheduler.h
cascade812: Non-const crossKvCacheManager is needed for MaxUtilizationScheduler, where it calls mutating APIs on the cross manager: startScheduling() and schedulingRemoveSequence().

Even though it's not needed for GuaranteedNoEvictScheduler and StaticBatchScheduler, here I changed it to const for API uniformity.

mikeiovine (approved):

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
pengbowang-nv: The trtllm-gen cross-attention path passes raw `q` (shape `[N, q_hidden]`) into `torch.ops.trtllm.qkv_preprocessing`, which assumed QKVPacked input. The thop fallback at `trtllm.py:1498-1502` pads Q for exactly this reason:                                                                                                         
```
q_arg = q.new_zeros((q.shape[0], q_hidden_size + 2 * kv_hidden_size))                                                                                                                    
q_arg[:, :q_hidden_size] = q
```
The trtllm-gen branch here doesn't.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -810,10 +824,13 @@ def forward_impl(
         attention_mask_data: Optional[torch.Tensor],
         mrope_config: Optional[dict],
         attention_sinks: Optional[torch.Tensor] = None,
+        relative_attention_bias: Optional[torch.Tensor] = None,
+        relative_attention_max_distance: int = 0,
         has_lora: bool = False,
     ):
         mrope_rotary_cos_sin = None
         mrope_position_deltas = None
+        rel_attn_max_distance = relative_attention_max_distance
```
pengbowang-nv: Unnecessary mixed usage of `rel_attn_max_distance` and `relative_attention_max_distance`

tensorrt_llm/_torch/modules/cross_attention.py
```diff
@@ -0,0 +1,291 @@
...
+        Used as a fallback when ``cross_attn_metadata`` is not provided.
+        Only single-request batches are unambiguous; multi-request batches
+        require the caller to supply ``cross_attn_metadata`` with explicit
+        ``seq_lens_kv``.
+        """
+        num_encoder_tokens = encoder_hidden_states.shape[0]
+        num_requests = int(attn_metadata.seq_lens.numel())
+        if num_requests == 1:
+            return torch.tensor([num_encoder_tokens], dtype=torch.int32)
+        if num_encoder_tokens % num_requests != 0:
```
pengbowang-nv: When will this function be used? I think we should always require encoder_seq_lens instead of inferring them in an unreliable way.

cpp/tensorrt_llm/batch_manager/capacityScheduler.cpp
cascade812: `prefillWithChunkedContextsAlreadyExecuting` is invoked for CONTEXT_INIT requests which requires cross kv cache, not for ENCODER_INIT request.

tensorrt_llm/_torch/pyexecutor/py_executor.py
cascade812: used at line 4480:
`request.draft_tokens = request.py_draft_tokens if get_draft_token_length(
                request) > 0 else []`

tensorrt_llm/_torch/pyexecutor/resource_manager.py
cascade812: Address the bloat concern by pulling the cross-KV path out of prepare_resources into a dedicated _prepare_cross_kv_resources helper. I think it's better than a unified refactor for two reasons:
1. cross KV has fundamentally different semantics from the decoder self-attention cache, eg, no draft token, never grow at decoding step.
2. reduce the risk to break existing logic for decoder self-attention cache . 

On the assertion suggestion:  the cross manager is dispatched the same unfiltered scheduled_batch as every other manager (ResourceManager.prepare_resources loops all of them), so generation_requests is non-empty on every decode step — an assert not generation_requests would fail immediately.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
cascade812: reverted

tensorrt_llm/_torch/pyexecutor/resource_manager.py
cascade812: Applied the same treatment as prepare_resources — the cross-KV update now is in a dedicated _update_cross_kv_resources helper, same reason as previous.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
cascade812: This is not a issue as it returns directly on line 935 for cross kv cache.

cpp/tensorrt_llm/common/attentionOp.cpp
cascade812: No, it doesn't break logic for self-attention. The special case is cross-attention. In the cross-attention preprocessing kernel, `cache_seq_lens` is used by this gate:

bool const store_encoder_kv_cache = (decoder_seq_len == decoder_cache_seq_len);

For cross-attention context, we need that gate to be true so encoder K/V are written into the cross-KV cache. However, cross-attention callers do not consistently use `sequence_lengths` as decoder length: in the PyTorch cross-attention path it can represent the encoder/cross-KV length, while other paths may keep it decoder-side. The actual encoder length is already passed separately via `encoder_seq_lens`.

So this keeps the original self-attention/chunked-context behavior, while making cross-attention use decoder context length for this decoder-side cache-write gate. I agree the original comment should stay; I added an extra cross-attention-specific note instead of replacing it.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
cascade812: Done.

cpp/tests/unit_tests/batch_manager/capacitySchedulerTest.cpp
cascade812: Addressed

cpp/tensorrt_llm/thop/attentionOp.cpp
cascade812: Good catch, changed to read from host_past_key_value_lengths which is on CPU.

tensorrt_llm/_torch/attention_backend/trtllm.py
cascade812: I removed `or metadata.is_cross` here. Now backend selection follows one rule for both self- and cross-attention: use TRTLLM-Gen only when the user opts in globally (TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION=1) and is_supported() confirms the hardware/config supports it. Otherwise both fall back to the legacy.

tensorrt_llm/_torch/attention_backend/trtllm.py
cascade812: It's needed to reuse the existing fused-QKV preprocessing kernel.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
cascade812: fixed

venkywonka:
@coderabbitai full review

tests/integration/defs/llmapi/test_llm_api_pytorch_t5.py
```diff
@@ -0,0 +1,678 @@
...
+_FLAN_T5_XXL_MIN_GPU_MEMORY_MB = 80_000
+_FREE_GPU_MEMORY_FRACTION = 0.2
+_CROSS_KV_CACHE_FRACTION = 0.5
+_EXPECTED_TRANSLATION_FRAGMENT = "Haus"
+_EXPECTED_OUTPUT_TOKEN_IDS_BY_MODEL = {
+    "t5-small": [644, 4598, 229, 19250],
+    "t5-base": [644, 4598, 229, 19250],
+    "t5-large": [644, 4598, 229, 19250],
+    "flan-t5-small": [644, 4598, 229, 9685],
+    "byt5-small": [258, 35, 119, 114],
```
venkywonka, line 50: curious, why bart isn't part of the models tested here? also, since its not just t5-specific, maybe renaming this file to test_llm_api_encoder_decoder.py or along those lines to not keep it model-specific?

tensorrt_llm/llmapi/llm.py
```diff
@@ -325,6 +326,76 @@ def disaggregated_params(self) -> dict:
...
+                 pos: int,
+                 *,
+                 token_ids_are_scalar: bool = False) -> Any:
+        if maybe_batched is None:
+            return None
+        if token_ids_are_scalar and cls._is_token_id_list(maybe_batched):
+            return maybe_batched
+        if isinstance(maybe_batched, list):
+            return maybe_batched[pos]
+        return maybe_batched
```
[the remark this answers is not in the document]
venkywonka: not sure if maybe_batched can be tuple-typed. if that's the case this might be a valid bug

venkywonka (approved):
lgtm from llmapi! thanks!

cpp/include/tensorrt_llm/batch_manager/capacityScheduler.h
eopXD: So is this a stylistic change? I don't think that is a reason to drop a safeguard.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
eopXD: I think this is just finding another way to bloat the code.

For 1. , just skip
For 2, a separate module will duplicate usage call and may diverge into misaligned convention that hides bugs.

I think we can find a way to modulize the existing logic and reuse it for the cross-KV.

pengbowang-nv (approved):

cpp/include/tensorrt_llm/batch_manager/capacityScheduler.h
cascade812: Changed it back for GuaranteedNoEvictScheduler and StaticBatchScheduler.

tensorrt_llm/_torch/pyexecutor/resource_manager.py
cascade812: Changed to unified function to remove duplicate

tests/integration/defs/llmapi/test_llm_api_pytorch_t5.py
cascade812: Added a separate test_llm_api_encoder_bart.py

eopXD:
Looks like the CI process did not recognize your added tests.

```
[2026-06-08T02:55:33.226Z] Invalid cases:
[2026-06-08T02:55:33.226Z] {"invalid": ["llmapi/test_llm_api_pytorch_bart.py::test_bart_pytorch_generate_encoder_decoder_trtllm_gen_attention[trtllm-gen-bf16-kv-v1-cuda-graph-off-greedy-bart-large-cnn]", "llmapi/test_llm_api_pytorch_t5.py::test_t5_pytorch_generate_encoder_decoder_end_to_end[bf16-kv-v1-cuda-graph-on-beam2-t5-small]", "llmapi/test_llm_api_pytorch_bart.py::test_bart_pytorch_generate_encoder_decoder_end_to_end[bf16-kv-v1-cuda-graph-on-beam2-bart-large-cnn]", "llmapi/test_llm_api_pytorch_bart.py::test_bart_pytorch_generate_encoder_decoder_end_to_end[bf16-kv-v1-cuda-graph-on-greedy-bart-large-cnn]", "llmapi/test_llm_api_pytorch_t5.py::test_t5_pytorch_generate_encoder_decoder_end_to_end[bf16-kv-v2-cuda-graph-on-greedy-t5-small]", "llmapi/test_llm_api_pytorch_bart.py::test_bart_pytorch_generate_encoder_decoder_end_to_end[bf16-kv-v2-cuda-graph-on-greedy-bart-large-cnn]", "llmapi/test_llm_api_pytorch_t5.py::test_t5_pytorch_generate_encoder_decoder_end_to_end[bf16-kv-v1-cuda-graph-on-greedy-t5-small]"]}
[2026-06-08T02:55:33.226Z] 
[2026-06-08T02:55:33.226Z] Invalid test names found, please correct them first!!!
[2026-06-08T02:55:33.226Z] 
script returned exit code 1
```

Plaese fix this before attempting to re-trigger.

tburt-nv:
pipeline 42424 actually passed:
- `[2026-06-10T07:39:02.468Z] DGX_B200-PyTorch-3/test_unittests.py::test_unittests_v2[unittest/kv_cache_manager_v2_tests/] <- test_unittests.py PASSED`
- `[2026-06-10T07:29:31.808Z] DGX_B200-PyTorch-1/test_unittests.py::test_unittests_v2[unittest/_torch/thop/parallel] <- test_unittests.py PASSED`

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1633,7 +1675,7 @@ def _run(
                 head_size=self.head_dim,
                 quant_mode=self.quant_mode,
                 q_scaling=self.q_scaling,
-                position_embedding_type=self.position_embedding_type,
+                position_embedding_type=forward_args.position_embedding_type,
```
yuxianq: Please move this to the forward_args section

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1669,6 +1711,12 @@ def _run(
                 # stay as literal ``None`` until DeepSeek V4 sparse-MLA lands.
                 sparse_mla_topk_lens=None,
                 compressed_kv_cache_pool_ptr=None,
+                cross_attention=metadata.is_cross,
+                cross_kv=forward_args.cross_kv,
+                encoder_input_lengths=forward_args.encoder_input_lengths,
+                relative_attention_bias=forward_args.relative_attention_bias,
+                relative_attention_max_distance=forward_args.
+                relative_attention_max_distance,
```
yuxianq: Please move these new arguments to the correct sections.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1516,7 +1558,7 @@ def _run(
             assert metadata.num_contexts == metadata.num_seqs
 
         use_trtllm_gen = False
-        if _TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION:
+        if _TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION and not metadata.is_cross:
```
yuxianq: Don't add `not metadata.is_cross` here, move it to `trtllm_gen_backend.is_supported`

tests/integration/test_lists/test-db/l0_h100.yml
```diff
@@ -166,6 +166,37 @@ l0_h100:
   - disaggregated/test_disaggregated_single_gpu.py::test_disaggregated_logits[True-TinyLlama-1.1B-Chat-v1.0]
   - disaggregated/test_disaggregated_single_gpu.py::test_disaggregated_logprobs[False-TinyLlama-1.1B-Chat-v1.0]
   - disaggregated/test_disaggregated_single_gpu.py::test_disaggregated_logprobs[True-TinyLlama-1.1B-Chat-v1.0]
+  - llmapi/test_llm_api_pytorch_bart.py::test_bart_pytorch_generate_encoder_decoder_end_to_end[bf16-kv-v1-cuda-graph-off-greedy-bart-large-cnn]
```
yuxianq, line 169: Which arch support cross attention with your PR? We only test sm90, how about sm100?

cascade812: Cross attention is now supported on both SM90 and SM100 by falling back to THOP attention.
I'll have a separate PR that adds SM100 cross-attention support to trtllm-gen.

yuxianq: THOP uses trtllm-gen for sm100 and XQA/fmha_v2 for sm90, if trtllm-gen cross-attention kernels are aded in a follow-up PR, this PR will only contain sm90 support, right?

tensorrt_llm/_torch/attention_backend/trtllm.py
cascade812: @yuxianq can you clarify this? I think new arguments are already in the forward_args section?

tensorrt_llm/_torch/attention_backend/trtllm.py
cascade812: same question as below

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -1211,29 +1228,36 @@ def update_resources(self,
...
-            # and eventually overflows mCacheBlockIndices.
-            runtime_draft_len = (request.py_rewind_len +
-                                 request.py_num_accepted_draft_tokens)
-            extra_rewind = self._kv_reserve_draft_tokens - runtime_draft_len
-            if extra_rewind > 0:
-                self.rewind_kv_cache(request, extra_rewind)
+        is_cross = self.kv_cache_type == CacheTypeCpp.CROSS
+        if not is_cross:
+            if not self.is_draft:
+                _update_kv_cache_draft_token_location(self, scheduled_batch,
```
eopXD, line 949: Nit: This looks like a drive-by fix to achieve symmetry of the v2 manager's path of `update_resources`. I think we can remove this in case this dillutes the goal of this MR. A follow-up for this addition should also be good. My nit comment is to remove it here.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: You put `encoder_input_lengths/relative_attention_bias /relative_attention_max_distance` in the "Literals intentionally None" section, please move it to the "Per-call (AttentionForwardArgs)" section.

tests/integration/test_lists/test-db/l0_h100.yml
cascade812: Currently it supports attn_backend="TRTLLM" inside which it falls back to thop for cross attention on both sm90 and sm100.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1418,6 +1428,38 @@ def _run(
...
+            else:
+                assert q.shape[1] == qkv_hidden_size
+            k = None
+            v = None
+            forward_args.is_fused_qkv = True
+        else:
+            forward_args.cross_kv = None
+            forward_args.encoder_input_lengths = None
+
+        forward_args.position_embedding_type = (
```
yuxianq: Why we need to determine the position_embedding_type during forward? It is a static info, we should update the position_embedding_type in attention backend's init and remove position_embedding_type from forward_args.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: forward_args.position_embedding_type should move to the "Per-call (AttentionForwardArgs)" section, but I prefer to not add position_embedding_type to forward_args, see my previous comment.

tests/integration/test_lists/test-db/l0_h100.yml
yuxianq: There are `trtllm_gen_backend` and thop in trtllm attention backend, `trtllm_gen_backend` means flashinfer's trtllm-gen kernels (we will refactor it in https://github.com/NVIDIA/TensorRT-LLM/pull/15204 to make it clearer), but we also integrate trtllm-gen kernels to thop. If there is no trtllm-gen kernels for cross attention, thop also cannot support it. Do you mean we already have cross attention support in trtllm-gen, but you have not enabled it for `trtllm_gen_backend`?

eopXD (approved):
The changes under resource_manager.py come from a perspective of "if not is_cross", from a perspective of supporting the cross attention path you are trying to achieve. This is understandable.

However it is a slight code smell since you are branching out the main code path.You may  check https://github.com/cascade812/TensorRT-LLM/pull/1 for reference.

All other changes look good to me now. Thank you for your patience.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1418,6 +1428,38 @@ def _run(
...
+        if metadata.is_cross:
+            forward_args.update_kv_cache = k is not None and v is not None
+            if k is not None and v is not None:
+                k_flat = k.contiguous().view(k.shape[0], -1)
+                v_flat = v.contiguous().view(v.shape[0], -1)
+                forward_args.cross_kv = torch.cat([k_flat, v_flat],
+                                                  dim=1).contiguous()
+            else:
+                forward_args.cross_kv = None
+            forward_args.encoder_input_lengths = metadata.kv_lens_cuda_runtime
```
yuxianq: We have already pass `metadata.kv_lens_cuda_runtime` to thop, don't need to pass `encoder_input_lengths` again, we can remove `encoder_input_lengths` from forward_args

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1727,9 +1776,11 @@ def forward(
...
         forward_args.is_fused_qkv = not metadata.is_cross and k is None
         forward_args.update_kv_cache = not metadata.is_cross or k is not None
-        assert (forward_args.is_fused_qkv and k is None
-                and v is None) or (not forward_args.is_fused_qkv
-                                   and k is not None and v is not None)
+        assert (forward_args.is_fused_qkv and k is None and v is None) or (
+            not forward_args.is_fused_qkv and k is not None
+            and v is not None) or (metadata.is_cross
+                                   and not forward_args.update_kv_cache
+                                   and k is None and v is None)
```
yuxianq: Instead of adding assertion, we should update forward_args.update_kv_cache for cross attention here instead of L1432

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1418,6 +1428,38 @@ def _run(
...
     ) -> None:
+        if metadata.is_cross:
+            forward_args.update_kv_cache = k is not None and v is not None
+            if k is not None and v is not None:
+                k_flat = k.contiguous().view(k.shape[0], -1)
+                v_flat = v.contiguous().view(v.shape[0], -1)
+                forward_args.cross_kv = torch.cat([k_flat, v_flat],
+                                                  dim=1).contiguous()
+            else:
+                forward_args.cross_kv = None
```
yuxianq: Remove this line since cross_kv is None by default.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1418,6 +1428,38 @@ def _run(
...
+                fused_q[:, :q_hidden_size].copy_(q)
+                q = fused_q
+            else:
+                assert q.shape[1] == qkv_hidden_size
+            k = None
+            v = None
+            forward_args.is_fused_qkv = True
+        else:
+            forward_args.cross_kv = None
+            forward_args.encoder_input_lengths = None
```
yuxianq: Remove this else block since they are None by default.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1669,6 +1711,12 @@ def _run(
                 # stay as literal ``None`` until DeepSeek V4 sparse-MLA lands.
                 sparse_mla_topk_lens=None,
                 compressed_kv_cache_pool_ptr=None,
+                cross_attention=metadata.is_cross,
```
yuxianq: Use the same name for kwargs, e.g., `is_cross=metadata.is_cross`

tensorrt_llm/_torch/attention_backend/trtllm.py
cascade812: Fixed it by setting pos_embd_params to PositionEmbeddingType.relative when init t5.

tests/integration/test_lists/test-db/l0_h100.yml
cascade812: yes, that's what I meant. The underlying trtllm-gen path used through thop already has the sm100 kernels needed for cross attention, but the the explicit trtllm_gen_backend wrapper doesn't enable cross attention yet because it doesn't carry the encoder KV inputs and currently rejects meata.is_cross. So this PR supports cross attention on sm90 an sm100 through attn_backend="TRTLLM" via THOP. The coming PR is to enable the explicit trtllm_gen_backend path for cross attention which may need adjustment as you have refactored flashinfer's trtllm-gen kernels in your PR.

yuxianq: Good, can you add some tests to b200 test list to guard them? Thanks~

cascade812: Thanks, added~

PR #13926, opened 2026-05-09
[TRTLLM-12440][feat] Add GMS-only weight sharing support

chienchunhung:
## Release Notes

* **New Features**
  * Added GPU Memory Service (GMS) support for model loading and memory management with configurable socket path, mode, and weight tag options.
  * Introduced `LoadFormat.GMS` configuration option for GPU memory backend integration.

* **Improvements**
  * Added `is_weights_preloaded()` method to check weight preload status during model initialization.
  * Enhanced resource cleanup for model engines with idempotent, deterministic teardown.
  * Implemented configuration validation to prevent incompatible GMS and Mixture-of-Experts settings.

* **Tests**
  * Added comprehensive GMS backend and model loader test coverage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13926?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Summary

Adds `LoadFormat.GMS` so multiple TRT-LLM instances on the same node can zero-copy share model weights via the [GPU Memory Service (GMS)](https://github.com/ai-dynamo/dynamo/tree/main/lib/gpu_memory_service) pool. The first instance loads weights as the writer (RW); subsequent peers materialize them read-only (RO) without disk I/O or per-instance copies.

## Scope

In scope:
- `LoadFormat.GMS` enum value and nested `GmsConfig` (`socket_path`, `mode`, `tag`) on `TorchLlmArgs`.
- New `GMSBackend` adapter under `tensorrt_llm/_torch/memory/`, lazily imported only when GMS is selected.
- `ModelLoader` GMS branch with explicit RW / RO / unexpected-state handling, plus a guard that refuses to commit an unpopulated model to the pool.
- Cleanup hook released via `PyTorchModelEngine.__del__`.

Adjacent (sub-PR, ~15 lines): remove the redundant `MXCheckpointLoader.p2p_succeeded` property; `is_weights_preloaded()` (the abstract hook from `BaseCheckpointLoader`) is now the single accessor. Tests updated accordingly.

Out of scope:
- Networking and MX integration (the MX checkpoint format already merged in #13531 composes orthogonally with `LoadFormat.GMS`).
- Packaging the `gpu_memory_service` dependency. Same OSS-allowlist concern as MX; users install it manually for now.

## Test Coverage

New unit test files (mock-based, CPU CI):
- `tests/unittest/llmapi/test_gms_args.py` — `GmsConfig` validation and `LoadFormat.GMS` Pydantic surface.
- `tests/unittest/_torch/memory/test_gms_backend.py` — `GMSBackend` lifecycle and helpers.
- `tests/unittest/_torch/pyexecutor/test_model_loader_gms.py` — `ModelLoader` GMS RW / RO / failure / edge-case branches.

Updated:
- `tests/unittest/_torch/models/checkpoints/mx/test_mx_checkpoint_loader.py` — six call sites switched to `is_weights_preloaded()` after removing the redundant `p2p_succeeded` property.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/memory/gpu_memory_backend.py
+++ b/tensorrt_llm/_torch/memory/gpu_memory_backend.py
@@ -70,7 +70,7 @@ class GPUMemoryBackend(Protocol):
     def mem_pool_scope(
         self,
         device: Optional[torch.device] = None,
-    ) -> "Iterator[None]":
+    ) -> Iterator[None]:
         """Return a context manager scoping CUDA allocations to the backend pool."""
         ...
 
@@ -349,31 +349,35 @@ class GMSBackend:
         device_index = self._device_index
         tracked_storage_ptrs: set[int] = set()
 
-        with torch.no_grad():
-            for _name, tensor, tensor_type in _iter_module_tensors(model):
-                if tensor_type != "parameter" or tensor is None or not tensor.is_cuda:
-                    continue
+        # No torch.no_grad() guard: ``tensor.data = X`` bypasses autograd by
+        # definition (it's a data-attribute assignment, not an autograd-tracked
+        # operation), and ``replacement.copy_(tensor)`` is an in-place op on a
+        # freshly created tensor with no autograd history. Model loading also
+        # runs outside any grad-enabled context.
+        for _name, tensor, tensor_type in _iter_module_tensors(model):
+            if tensor_type != "parameter" or tensor is None or not tensor.is_cuda:
+                continue
 
-                # GMS tracks whole storage allocations, so use the storage
-                # base pointer instead of a tensor view's offset pointer.
-                storage_base_ptr = tensor.untyped_storage().data_ptr()
-                if _ptr_in_gms(gms_client, int(storage_base_ptr)):
-                    continue
-                if storage_base_ptr in tracked_storage_ptrs:
-                    continue
-                tracked_storage_ptrs.add(storage_base_ptr)
+            # GMS tracks whole storage allocations, so use the storage
+            # base pointer instead of a tensor view's offset pointer.
+            storage_base_ptr = tensor.untyped_storage().data_ptr()
+            if _ptr_in_gms(gms_client, int(storage_base_ptr)):
+                continue
+            if storage_base_ptr in tracked_storage_ptrs:
+                continue
+            tracked_storage_ptrs.add(storage_base_ptr)
 
-                nbytes = _storage_nbytes(tensor)
-                base_va = gms_client.create_mapping(size=nbytes, tag=self._tag)
-                replacement = _tensor_from_pointer(
-                    int(base_va),
-                    list(tensor.shape),
-                    list(tensor.stride()),
-                    tensor.dtype,
-                    device_index,
-                )
-                replacement.copy_(tensor)
-                tensor.data = replacement
+            nbytes = _storage_nbytes(tensor)
+            base_va = gms_client.create_mapping(size=nbytes, tag=self._tag)
+            replacement = _tensor_from_pointer(
+                int(base_va),
+                list(tensor.shape),
+                list(tensor.stride()),
+                tensor.dtype,
+                device_index,
+            )
+            replacement.copy_(tensor)
+            tensor.data = replacement
 
     def finalize_write(self, model: nn.Module) -> int:
         """Register tensors, commit them, and transition this client to RO.
--- a/tensorrt_llm/_torch/memory/gpu_memory_backend.py
+++ b/tensorrt_llm/_torch/memory/gpu_memory_backend.py
@@ -195,6 +195,12 @@ class GMSBackend:
                 tag=self._tag,
             )
         except Exception as e:
+            # TODO(GMS-API): narrow once upstream documents the expected
+            # client-creation failure modes (socket missing, connection
+            # refused, daemon crashed). Today the upstream API doesn't
+            # enumerate them, so the broad catch protects the
+            # "GMS optional / fall through to disk" guarantee at the cost
+            # of also swallowing programming bugs (e.g. signature drift).
             logger.warning(
                 "Failed to connect to GMS at %s (mode=%s, tag=%s): %s",
                 socket_path,
@@ -203,6 +209,7 @@ class GMSBackend:
                 e,
             )
             self._client = None
+            self._is_rw = None
             return False
 
         self._is_rw = self._client.granted_lock_type == GrantedLockType.RW
@@ -347,7 +354,15 @@ class GMSBackend:
 
         gms_client = self._client
         device_index = self._device_index
-        tracked_storage_ptrs: set[int] = set()
+        # Map original storage base ptr -> new GMS base_va. Serves two roles:
+        #   1. Dedups create_mapping() / copy_() calls when multiple tensors
+        #      share a storage (e.g. tied embeddings, sliced fused params).
+        #   2. Lets us rebind every view of a shared storage to the SAME
+        #      GMS mapping, so all tied/aliased parameters end up backed
+        #      by the GMS pool. Without this, only the first encountered
+        #      view was rebound; subsequent views still pointed at the
+        #      original non-GMS storage and were missed by finalize_write.
+        storage_to_gms_va: dict[int, int] = {}
 
         # No torch.no_grad() guard: ``tensor.data = X`` bypasses autograd by
         # definition (it's a data-attribute assignment, not an autograd-tracked
@@ -363,12 +378,14 @@ class GMSBackend:
             storage_base_ptr = tensor.untyped_storage().data_ptr()
             if _ptr_in_gms(gms_client, int(storage_base_ptr)):
                 continue
-            if storage_base_ptr in tracked_storage_ptrs:
-                continue
-            tracked_storage_ptrs.add(storage_base_ptr)
 
-            nbytes = _storage_nbytes(tensor)
-            base_va = gms_client.create_mapping(size=nbytes, tag=self._tag)
+            base_va = storage_to_gms_va.get(storage_base_ptr)
+            first_for_this_storage = base_va is None
+            if first_for_this_storage:
+                nbytes = _storage_nbytes(tensor)
+                base_va = gms_client.create_mapping(size=nbytes, tag=self._tag)
+                storage_to_gms_va[storage_base_ptr] = base_va
+
             replacement = _tensor_from_pointer(
                 int(base_va),
                 list(tensor.shape),
@@ -376,7 +393,11 @@ class GMSBackend:
                 tensor.dtype,
                 device_index,
             )
-            replacement.copy_(tensor)
+            # Copy original bytes exactly once per storage; subsequent
+            # views of the same storage just rebind onto the same GMS
+            # mapping without recopying.
+            if first_for_this_storage:
+                replacement.copy_(tensor)
             tensor.data = replacement
 
     def finalize_write(self, model: nn.Module) -> int:
@@ -511,6 +532,7 @@ class GMSBackend:
             logger.warning("GMS cleanup error: %s", e)
         finally:
             self._client = None
+            self._is_rw = None
 
 
 def _ptr_in_gms(gms_client, ptr: int) -> bool:
--- a/tensorrt_llm/_torch/memory/gpu_memory_backend.py
+++ b/tensorrt_llm/_torch/memory/gpu_memory_backend.py
@@ -1,0 +1,433 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""GPU Memory Backend protocol and GMS implementation.
+
+Defines a thin protocol that adapts the upstream GPU Memory Service
+(``gpu_memory_service`` from ``ai-dynamo/dynamo``) into TRT-LLM's
+``ModelLoader`` pipeline. The protocol intentionally exposes only the
+operations TRT-LLM needs at specific points in the loading lifecycle —
+all heavy lifting (CUDA VMM, FD passing, zero-copy tensor construction)
+is delegated to the GMS library's stable public Python primitives.
+
+Design notes:
+- We deliberately avoid the upstream ``setup_gms()`` monkey-patch entry
+  point. TRT-LLM owns the integration policy in ``ModelLoader``; this
+  backend is a thin adapter over the GMS library's primitive API.
+- The protocol has a single point of contact per concern, so when the
+  upstream GMS Python API drifts, only the methods on ``GMSBackend``
+  need to change — the call sites in ``model_loader.py`` are stable.
+
+Operating modes:
+- **RW (Read-Write)**: First worker loads weights via the normal
+  checkpoint loader pipeline, with allocations captured by a GMS-managed
+  CUDA memory pool. After loading, weights are committed for read-only
+  access by other workers and the client transitions to RO mode in place.
+- **RO (Read-Only)**: Subsequent workers zero-copy import already-committed
+  weights from the GMS pool. ``post_load_weights()`` must run BEFORE
+  materialization so that module aliases are set up correctly.
+"""
+
+from contextlib import contextmanager
+from typing import Iterator, Optional, Protocol, runtime_checkable
+
+import torch
+from torch import nn
+
+from tensorrt_llm.logger import logger
+from tensorrt_llm.mapping import Mapping
+
+
+@runtime_checkable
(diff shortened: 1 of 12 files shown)


tensorrt_llm/_torch/memory/gpu_memory_backend.py
galletas1712: do we want a proper API on the GMS side for this, so TRTLLM and consume his more easily?

galletas1712: is there any way to make the other buffers be allocated in a mempool in the first place? it might memory accounting easier, especially on shadow replenishment where there is a risk of OOM.

but if not possible this is fine for now.

tensorrt_llm/_torch/pyexecutor/model_loader.py
galletas1712, line 635: torch.cuda.empty_cache() may crash if there are still active mappings in the GMS mempool. we would typically do this only when everything has been unmapped.

tests/unittest/_torch/models/checkpoints/mx/test_mx_checkpoint_loader.py
galletas1712, line 192: I haven't really read through the MX path, but is the goal to allow MX to load weights into the GMS mempool?

tests/unittest/_torch/memory/test_gms_backend.py
galletas1712, line 9: probably not in this file, but a good e2e test for GMS is a simple engine restart with weights still resident in GMS. This would allows to see if KV cache memory accounting is done properly.

this would be a really useful smoke test before we actually put in shadow engine failover.

tensorrt_llm/_torch/pyexecutor/model_loader.py
hhzhang16, line 520: The PR skips meta-tensor materialization and `model.to("cuda")` for `LoadFormat.GMS` (above in this function), then only enters `mem_pool_scope` here. In Dynamo’s TRT-LLM shim, the normal loader runs inside `gms_use_mem_pool`, so model allocation, loading, and post-load work all land in the GMS pool before commit. As written, RW can try to load into meta params or allocate outside GMS, then fail or publish an incomplete layout.

hhzhang16, line 638: Should this wait until after the `post_load_*` functions that happen later?

tensorrt_llm/_torch/memory/gpu_memory_backend.py
chienchunhung: That'd be ideal! 

We're currently reaching into `gpu_memory_service.client.torch.{module,tensor}._iter_module_tensors` / `_tensor_from_pointer` and the `_mappings` attribute on the client, which are all underscore-prefixed and not part of upstream's stable surface. 

The ideal interface could be a public `gms_client.move_untracked_params(model)` (or equivalent) that hides the iteration + mapping creation + data swap dance, plus a public `mapping_contains(ptr)` predicate to replace `_ptr_in_gms`. 

Let me know what you think of this proposal; for this PR we will continue as is for now.

chienchunhung: Updated. `Meta init`, `model.to("cuda")`, `load`, and `post_load_*` now all allocate inside the pool from the start. move_untracked_params is now defensive (kept for C++-allocator-bypass). The shadow-replenishment OOM you flagged is exactly the case helped most: no more 2× peak from `move_untracked_params` copying strays in.

tensorrt_llm/_torch/pyexecutor/model_loader.py
chienchunhung: `GMSBackend.connect()` installs upstream's `patch_empty_cache()` from `gpu_memory_service.integrations.common.patches`, which monkey-patches `torch.cuda.empty_cache` to skip GMS-backed VMM regions. So by the time we reach this `empty_cache()`, GMS mappings are immune.

That said, the monkey-patch approach is not ideal either, but I am not sure if we have better option to bypass it.

tests/unittest/_torch/models/checkpoints/mx/test_mx_checkpoint_loader.py
chienchunhung: Ye, that's the design intent of the two-axis split PR #13531 set up:
- checkpoint_format = how weights arrive (HF disk, MX P2P over NIXL, future ones).
- load_format = where they live (AUTO = standard CUDA, GMS = node-shared pool).

The two compose orthogonally on the TRT-LLM side because the GMS RW branch in ModelLoader.load calls checkpoint_loader.load_weights(...) without caring which loader is registered, so any CUDA allocation the loader does inside mem_pool_scope lands in the GMS pool and move_untracked_params then sweeps anything that escaped.

**Real gap** on the MX side that blocks shipping (MX, GMS) today: the upstream MxLiveWeightLoader API has no parameter for a destination allocator / memory pool — MxLiveWeightLoader(mx_server=url) and load_weights(ckpt, mapping=, model=) are the entire surface we use. Until upstream MX exposes a `memory_pool=` (or equivalent) on `MxLiveWeightLoader.load_weights`, the (MX, GMS) mode can't be claimed correct. So this PR ships (HF, GMS), (MX, AUTO), and (HF, AUTO); (MX, GMS) is left unsupported.

tests/unittest/_torch/memory/test_gms_backend.py
chienchunhung: Strongly agree. This would require wiring up GMS daemon in our CI environment and installation. I will get to this (i.e., wiring GMS up and creating e2e test) in my follow-up PR.

tensorrt_llm/_torch/pyexecutor/model_loader.py
chienchunhung: Good call; updated.

chienchunhung: Yes — `finalize_write` should publish the `post-post_load_*` layout, otherwise RO peers materialize a stale model. Folded into the `mem_pool_scope` widening on your other comment: `post_load_apply`, `post_load_publish`, and `post_load_weights` now all run inside the pool, then the scope exits, then `finalize_write` commits. 

New test `test_gms_rw_post_load_runs_inside_pool_before_finalize` asserts the exact ordering as an invariant so a future refactor can't silently re-narrow it. This also unifies the RW exit shape with the existing RO branch which already calls `post_load_weights` before materialize_module.

tests/unittest/_torch/models/checkpoints/mx/test_mx_checkpoint_loader.py
galletas1712: afaik MX also has a similar "model materialization" API as GMS? just something to look out for since it might not be compatible with GMS.

tensorrt_llm/_torch/memory/gpu_memory_backend.py
galletas1712: ideally we don't need to move any untracked parameters (I think you addressed this below), so maybe this `move_untracked_params(model)` API isn't needed? do you think a `mapping_contains(ptr)` is still useful though?

galletas1712 (approved):

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -461,6 +472,193 @@ def init_meta_tensor(t: torch.Tensor):
...
+                        # (including draft_model when speculative decoding is
+                        # present).
+                        for module in model.modules():
+                            if hasattr(module,
+                                       'post_load_weights') and not getattr(
+                                           module, '_weights_removed', False):
+                                module.post_load_weights()
+
+                        gms_backend.materialize_module(model)
+                        gms_post_load_handled = True
```
hhzhang16, line 680: Can we run the normal module `post_load_weights()` pass after `materialize_module()` in the RO path?Right now this branch calls `post_load_weights()` before GMS materializes the weights, while params are still meta, then sets `gms_post_load_handled=True`, so the shared post-load block below is skipped. Dynamo's current shim does a small pre-materialization hook to create references, then materializes GMS tensors, then runs module `post_load_weights()` against real CUDA tensors.

Some hooks may inspect, derive, or rebind real weight tensors, so skipping the post-materialize pass can make RO engines diverge from RW engines.

chienchunhung: The RO branch in `ModelLoader.load` has two viable orderings: run `post_load_weights()` before `materialize_module_from_gms()` (current) or after (what you suggested). The current order is intentional although both have certain failure modes.

**Materialize-first** crashes hard for any model whose `post_load_weights()` wires layer aliases — e.g. `LlamaForCausalLM` setting `layer.next_attn = self.model.layers[idx+1].self_attn`. PyTorch registers `next_attn` as a child module, so the RW writer commits both the canonical path (`model.layers.1.self_attn.o_proj.weight`) and the alias path (`model.layers.0.next_attn.o_proj.weight`) to GMS — same physical tensor, two keys. On the RO reader, `materialize_module_from_gms()` walks the catalog and tries to resolve `model.layers.0.next_attn`, which is still `None` until the alias-wiring hook runs, so it raises `AttributeError: Cannot resolve 'o_proj'`. This is not a memory issue (the duplicate keys cost nothing — both resolve to the same `nn.Module` once aliases exist); it's a structural materialization failure that blocks Llama, DeepSeek, and any model with `next_attn`-style aliases. The bug was originally surfaced and fixed in [ai-dynamo/dynamo PR #7053](https://github.com/ai-dynamo/dynamo/pull/7053) (see the `discussion_r2105412837` thread for the repro).

**Post-load-first** (current) trades that crash for a soft divergence in Python-side state on the RO peer. Tensor data still shares zero-copy: even when a hook on the RO side runs against meta tensors and registers a new `nn.Parameter`, `materialize_module_from_gms()` subsequently rebinds it to the committed GMS storage. What diverges is non-tensor state computed inside the hook from weight data — cached scales, derived booleans, fingerprints — which compute incorrectly on meta and don't get re-run because `_weights_removed` blocks the second pass. This affects only modules whose hooks read weight data; pure alias-wiring hooks are unaffected. 

Hard-crash > soft-divergence in severity, so the current ordering is the right call for this PR. The principled fix (a separate pre-materialize alias hook so the structural and data-consuming portions can run in different phases) requires refactoring every model class that uses `post_load_weights()` aliases, which is filed as a `TODO(GMS-PHASE3)` follow-up.

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -461,6 +472,193 @@ def init_meta_tensor(t: torch.Tensor):
...
+
+                            # Catch-all for any tensors not on CUDA yet.
+                            model.to("cuda")
+
+                            weight_source = (model.llm_checkpoint_dir
+                                             if hasattr(model,
+                                                        'llm_checkpoint_dir')
+                                             else checkpoint_dir)
+                            weights = checkpoint_loader.load_weights(
+                                weight_source, mapping=self.mapping)
```
pcastonguay: Do we need to pass `model` here?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -107,8 +106,7 @@ def Field(default: Any = ...,
 
 
 class CudaGraphConfig(StrictBaseModel):
-    """
-    Configuration for CUDA graphs.
+    """Configuration for CUDA graphs.
```
pcastonguay: I would bundle these cleanup changes in a separate PR since they are not related to GMS changes.

tensorrt_llm/_torch/memory/gpu_memory_backend.py
```diff
@@ -0,0 +1,555 @@
...
+            raise RuntimeError("GMS client not connected. Call connect() first.")
+
+        from gpu_memory_service.client.torch.module import _iter_module_tensors
+        from gpu_memory_service.client.torch.tensor import _tensor_from_pointer
+
+        gms_client = self._client
+        device_index = self._device_index
+        tracked_storage_ptrs: set[int] = set()
+
+        with torch.no_grad():
```
pcastonguay: nit: I don't think this is needed here.

tensorrt_llm/_torch/memory/gpu_memory_backend.py
```diff
@@ -0,0 +1,555 @@
...
+        ...
+
+    def has_committed_weights(self) -> bool:
+        """Whether committed weights are ready to be imported (RO-ready)."""
+        ...
+
+    def mem_pool_scope(
+        self,
+        device: Optional[torch.device] = None,
+    ) -> "Iterator[None]":
```
pcastonguay:
```suggestion
    ) -> Iterator[None]:
```

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -461,6 +472,193 @@ def init_meta_tensor(t: torch.Tensor):
...
+                            checkpoint_loader.checkpoint_format)
+                    elif gms_backend.is_rw is False:
+                        # RO path: run post-load hooks first because GMS
+                        # materializes by walking the final module tree. The
+                        # hook order creates any aliases/derived parameter
+                        # attributes before the zero-copy GMS tensors replace
+                        # the meta tensors across the whole model tree
+                        # (including draft_model when speculative decoding is
+                        # present).
+                        for module in model.modules():
```
pcastonguay, line 668: Should `checkpoint_loader.post_load_apply` and `checkpoint_loader.post_load_publish` be called for `gms_backend.is_rw is False` too?

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -479,18 +677,26 @@ def init_meta_tensor(t: torch.Tensor):
...
+                    model,
+                    checkpoint_dir=checkpoint_dir,
+                    weights_preloaded=weights_preloaded)
+
+                for module in model.modules():
+                    if hasattr(module, 'post_load_weights') and not getattr(
+                            module, '_weights_removed', False):
+                        module.post_load_weights()
+
+            # TODO(GMS-MOE-LB): the MoE load balancer's
```
pcastonguay: Should we raise an error if `load_format == LoadFormat.GMS and moe_config.load_balancer`?

tests/unittest/_torch/models/checkpoints/mx/test_mx_checkpoint_loader.py
chienchunhung: Ack.

tensorrt_llm/_torch/memory/gpu_memory_backend.py
chienchunhung: Agreed that ideally we should not explicitly move untracked params.

tensorrt_llm/_torch/pyexecutor/model_loader.py
chienchunhung: For the GMS-only support this works fine without passing `model`, which is aligned with when `checkpoint_format='HF`. 

For MX+GMS support, we will need `model` to be passed here. 

I updated the code to add `model` so that we don't miss it in the future.

tensorrt_llm/llmapi/llm_args.py
chienchunhung: Done.

tensorrt_llm/_torch/memory/gpu_memory_backend.py
chienchunhung: Removed the `with torch.no_grad()` wrapper.

tensorrt_llm/_torch/memory/gpu_memory_backend.py
chienchunhung: Done.

tensorrt_llm/_torch/pyexecutor/model_loader.py
chienchunhung: Right. Today the `RO` path skips both hooks, which is harmless for `HfCheckpointLoader` because both are no-ops; but this could break the contract for any other loader (e.g., MX) where `post_load_apply` may have receiver-relevant work. Updated to call both on the `RO` branch with `weights_preloaded=True`.

tensorrt_llm/_torch/pyexecutor/model_loader.py
chienchunhung: Good idea; added `validate_gms_moe_compat` model validator on `TorchLlmArgs` that raises ValueError when `load_format == LoadFormat.GMS` and `moe_config.load_balancer is not None`.

pcastonguay (approved):

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -3697,6 +3684,80 @@ def validate_preshard_strategy(self) -> 'ModelExpressConfig':
...
+            ValueError: If ``mode`` is not one of ``{'auto', 'rw', 'ro'}``,
+                or if ``tag`` is empty / whitespace-only.
+
+        Returns:
+            ``self`` (Pydantic ``model_validator`` contract).
+        """
+        if self.mode not in ("auto", "rw", "ro"):
+            raise ValueError(
+                f"gms_config.mode must be 'auto', 'rw', or 'ro', got '{self.mode}'."
+            )
```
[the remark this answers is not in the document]
pcastonguay: @chienchunhung let's address

pcastonguay:
There are a few CodeRabbit comments that seems to raise valid issues, can you address after the request cancellation work.

tensorrt_llm/llmapi/llm_args.py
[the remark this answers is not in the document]
chienchunhung: Done.

brb-nv (approved):
Changes under `tensorrt_llm/_torch/models/checkpoints/mx/checkpoint_loader.py` look good to me.

venkywonka (approved):

PR #13938, opened 2026-05-09
[None][feat] Keep DSv4 o_a_proj as FP8, and port vLLM's fused_inv_rope_fp8_quant

lishicheng1996-nv:
## Issue Description

DeepSeek-V4 stores `o_a_proj` (the o_lora projection) natively as FP8 e4m3 with UE8M0 1×128 block scales — the same on-disk format as `o_b_proj` and the MoE expert weights. But while the rest of the model consumes those weights in their native FP8 form, the upstream MLA path upcasts `o_a_proj` to BF16 at load time (`weight_dequant`) and runs a BF16 BMM.

## What this PR does

Gated on the existing `use_cute_dsl_blockscaling_bmm` config flag in `extra_llm_api_options` (`TorchLlmArgs.use_cute_dsl_blockscaling_bmm`, default `False`). On SM100 with that flag enabled, on DSv4:

1. Skip the load-time `weight_dequant` of `o_a_proj`; dispatch the BMM through `cute_dsl_fp8_bmm_blackwell` with FP8 weights.
2. Replace the `mla_rope_inplace` + `fp8_batched_quantize_1x128_permute102` pair with a single Triton kernel `fused_inv_rope_fp8_quant` (ported from vLLM), so the BF16 latent is read once, RoPE-rotated in registers, FP8-quantized in registers, and FP8 + UE8M0 scales written once.

With the flag off (default), behavior is byte-identical to upstream.

### Per-step kernel sequence on the o_proj path

**Before (upstream BF16 BMM):**
```
[load-time]   weight_dequant      o_a_proj FP8 + UE8M0 -> BF16  (one shot)
[per-step 1]  mla_rope_inplace    in-place inverse RoPE on BF16 latent
[per-step 2]  bmm_out             BF16 BMM (latent x dequant'd o_a_proj)
[per-step 3]  o_b_proj GEMM       fp8_swap_ab_gemm
```

**After (FP8 BMM + fused front-end):**
```
[per-step 1]  fused_inv_rope_fp8_quant     in-place inverse RoPE +
                                            FP8 1x128 block-scaled quant +
                                            UE8M0-compatible scale layout
[per-step 2]  cute_dsl_fp8_bmm_blackwell   FP8 latent x FP8 weight
[per-step 3]  o_b_proj GEMM                fp8_swap_ab_gemm (unchanged)
```

### How to enable

```yaml
# extra_llm_api_options.yaml
use_cute_dsl_blockscaling_bmm: true
```

## Perf

Pro TEP8 c128 8k-1k, mtp=0, all other gates ON, MHC fused HC OFF:

| Kernel sequence | sys tok/s | Δ vs upstream |
|---|---|---|
| Upstream BF16 BMM (`mla_rope_inplace` + `bmm_out`) | 1730.52 | — |
| FP8 BMM only (`mla_rope_inplace` + `permute102_quant` + `cute_dsl_fp8_bmm`) | 1725.99 | -0.26% |
| **This PR (`fused_inv_rope_fp8_quant` + `cute_dsl_fp8_bmm`)** | **1756.88** | **+1.52%** |

The FP8-BMM-only intermediate is essentially flat-to-slightly-negative because the per-call `permute102_quant` adds ~9 MB HBM traffic that offsets the bandwidth saving from FP8 weight. The fusion eliminates that round-trip and lands the genuine bandwidth-savings on top.

## Accuracy

gsm8k Pro TEP8 c128 (`max_batch=16`, `max_num_tokens=8192`):

| Arm | flexible-extract | strict-match | avg |
|---|---|---|---|
| Upstream BF16 BMM | 96.21 | 96.29 | 96.25 |
| **This PR** | **96.36** | **96.36** | **96.36** |

Δ = +0.11 pp, within ±0.5 pp tolerance and ±1.34 pp eval stderr.

## Front-end fusion: vLLM port

The fused kernel is a port of vLLM's `fused_inv_rope_fp8_quant`:
- Source: https://github.com/vllm-project/vllm/blob/ecd0b60aad2f4e28dd00ababfc1402690d88cbed/vllm/v1/attention/ops/deepseek_v4_ops/fused_inv_rope_fp8_quant.py
- License: Apache-2.0 (compatible with TRT-LLM's Apache-2.0). Verbatim attribution + modifications block at the top of the new file.

Modifications relative to vLLM upstream:
1. Added `IS_NEOX: tl.constexpr` branch (DSv4 uses NEOX RoPE; vLLM upstream is interleaved-only).
2. Output FP8 buffer is fully contiguous `[n_groups, num_tokens, d]` so `cute_dsl_fp8_bmm_blackwell` consumes it via `make_ptr(.data_ptr())` without an adapter.
3. Dropped the UE8M0 INT32 packed-scale path (TRT-LLM's SM100 consumer takes raw FP32 a-side scales).
4. Rebound `cos_sin_cache` to TRT-LLM's `[max_pos, 2, rope_dim/2]` layout.
5. Replaced `direct_register_custom_op` with `torch.library.custom_op` to match the rest of `tensorrt_llm/_torch/custom_ops/`.

## Test Coverage

- [x] Unit test — `tests/unittest/_torch/custom_ops/test_fused_inv_rope_fp8_quant.py` compares the fused op vs the (`mla_rope_inplace` → `fp8_batched_quantize_1x128_permute102`) reference at `num_tokens in {3, 64, 257, 512}` on NEOX. Asserts <1% relative error in dequantized space.
- [x] Multi-rank end-to-end gsm8k accuracy A/B (numbers above).
- [x] Multi-rank perf A/B (numbers above).
- [x] ~~Cross-platform CI~~ — `cute_dsl_fp8_bmm_blackwell` is SM100-only; the dispatch gates on `is_sm_100f()`. Ops register cleanly on other SMs but fall back to the legacy path.

## PR Checklist

- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work)).
- Any new dependencies have been scanned for license and vulnerabilities (vLLM port: Apache-2.0, attribution preserved).
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes (no ownership changes).
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR (no architecture-level change).

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/custom_ops/__init__.py
+++ b/tensorrt_llm/_torch/custom_ops/__init__.py
@@ -1,3 +1,5 @@
+# Imported first to register the op before the chain via ..modules.attention re-enters this package.
+from . import triton_fused_inv_rope_fp8_quant  # noqa: F401  # isort: skip
 from ..cuda_tile_utils import IS_CUDA_TILE_AVAILABLE
 from ..cute_dsl_utils import IS_CUTLASS_DSL_AVAILABLE
 from ..flashinfer_utils import IS_FLASHINFER_AVAILABLE
--- a/tensorrt_llm/_torch/modules/attention.py
+++ b/tensorrt_llm/_torch/modules/attention.py
@@ -1618,14 +1618,29 @@ class MLA(nn.Module):
                 requires_grad=False,
             )
             if is_sm_100f():
-                self.o_a_proj_dequant = nn.Parameter(
-                    torch.empty(
-                        (self.n_local_groups, self.o_lora_rank,
-                         self.num_heads * self.qk_head_dim // self.num_groups),
-                        dtype=self.dtype,
-                    ),
-                    requires_grad=False,
-                )
+                # On DSv4 with the cute_dsl FP8 BMM enabled, keep o_a_proj in
+                # its native FP8 e4m3 form (no load-time dequant) so
+                # cute_dsl_fp8_bmm_blackwell can consume it directly.
+                if self.is_deepseek_v4 and self.use_cute_dsl_blockscaling_bmm:
+                    self.o_a_proj = nn.Parameter(
+                        torch.empty(
+                            (self.n_local_groups, self.o_lora_rank,
+                             self.num_heads * self.qk_head_dim //
+                             self.num_groups),
+                            dtype=torch.float8_e4m3fn,
+                        ),
+                        requires_grad=False,
+                    )
+                else:
+                    self.o_a_proj_dequant = nn.Parameter(
+                        torch.empty(
+                            (self.n_local_groups, self.o_lora_rank,
+                             self.num_heads * self.qk_head_dim //
+                             self.num_groups),
+                            dtype=self.dtype,
+                        ),
+                        requires_grad=False,
+                    )
         else:
             self.k_b_proj_trans_scale = None
             self.v_b_proj_scale = None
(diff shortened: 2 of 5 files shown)


tensorrt_llm/_torch/modules/attention.py
```diff
@@ -462,6 +462,12 @@ def __init__(
 
         self.use_cute_dsl_blockscaling_mm = config.use_cute_dsl_blockscaling_mm
         self.use_cute_dsl_blockscaling_bmm = config.use_cute_dsl_blockscaling_bmm
+        # TRTLLM_DSV4_O_PROJ_FP8=1 forces the FP8 BMM path on the DSv4
+        # o_a_proj (o_lora projection): the weight stays in its native FP8 e4m3
+        # form (no load-time dequant) and dispatches via
+        # cute_dsl_fp8_bmm_blackwell instead of BF16 bmm_out.
+        if os.environ.get("TRTLLM_DSV4_O_PROJ_FP8", "0") == "1":
```
lfr-0531: We already have `use_cute_dsl_blockscaling_bmm` in llm_args, and we can enable it by setting `use_cute_dsl_blockscaling_bmm: true` in the `extra_llm_api_options`. So I think we don't need `TRTLLM_DSV4_O_PROJ_FP8.`

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1618,14 +1630,32 @@ def create_weights(self):
...
+                        torch.empty(
+                            (self.n_local_groups, self.o_lora_rank,
+                             self.num_heads * self.qk_head_dim //
+                             self.num_groups),
+                            dtype=torch.float8_e4m3fn,
+                        ),
+                        requires_grad=False,
+                    )
+                else:
+                    self.o_a_proj_dequant = nn.Parameter(
```
lfr-0531, line 1635: Looks like we can drop this bf16 path now and keep only the FP8 one.

tensorrt_llm/_torch/custom_ops/__init__.py
```diff
@@ -1,3 +1,8 @@
+# Imported first so the `trtllm::fused_inv_rope_fp8_quant_vllm_port` op is
```
lfr-0531: We can simplify these comments, including the ones in `attention.py`.

tensorrt_llm/_torch/modules/attention.py
lishicheng1996-nv: Thanks for the review! Do we need keep the bf16 path for non-sm100 or use_cute_dsl_blockscaling_bmm=False (which is by default)?
Side question: should we by default enable use_cute_dsl_blockscaling_bmm for DSV4?

lfr-0531: Ah, you're right. Let's keep this BF16 path for future Hopper support.

lfr-0531 (approved):

PR #13477, opened 2026-04-26
[None][perf] Scheme X L2-aware dispatcher and PDL launchers for sparse-attention GVR Top-K

longcheng-nv:
## Summary

Follow-up to #12385 (Temporally-Correlated Heuristic-guided Indexer TopK).
This PR adds the per-(BS, N) **Scheme X dispatcher**, **PDL launchers**,
**CUDA Graph warmup**, and renames the inner micro-kernel symbols to
**GVR** (Guess-Verify-Refine).

- **Scheme X dispatcher (v1.1 + v1.2.3)**: per-(BS, N) routing between the
  heuristic kernel and the radix fallback, derived from `MultiProcessorCount`
  and `L2CacheSize` queried once at runtime. Closes the regression band
  around BS=128 / N=70K where the heuristic alone was slower than the
  radix path.
- **PDL on heuristic launchers**: switch `launchHeuristicTopKDecode` to
  `cudaLaunchKernelEx` with `cudaLaunchAttributeProgrammaticStreamSerialization`;
  call `cudaTriggerProgrammaticLaunchCompletion()` at the kernel epilogue.
  Symmetric with the radix path that already used PDL via
  `invokeIndexerTopKDecode`.
- **CUDA Graph safety**: `warmup_heuristic_topk_decode` helper invoked from
  the `Indexer` setup hook (`layer_idx == 0`) so the dispatcher's
  `cudaGetDevice` / `cudaDeviceGetAttribute` queries land outside any
  capture region; the cached `sm_count` and `L2CacheSize` are populated
  before any graph capture begins.
- **Rename**: inner micro-kernel symbols change from `heuristicTopK*` to
  `gvrTopK*` (`gvr` = Guess-Verify-Refine, the algorithm of the upcoming
  algorithm note). Public dispatcher / launcher names remain unchanged.

### Commit Breakdown (5 commits)

1. **fix**: Warm up heuristic TopK dispatcher for CUDA Graph safety
2. **feat**: L2-aware BS-threshold dispatcher (Scheme X v1.1) for GVR Top-K
3. **feat**: Lower GVR Top-K small-N threshold to 12288 (Scheme X v1.2.3)
4. **perf**: Enable PDL on heuristic TopK launcher and kernel epilogues
5. **refactor**: Rename inner GVR micro-kernel symbols

### Key Files

| File | Change |
|------|--------|
[...]

PR #13948, opened 2026-05-09
[None][feat] Multi-K (512/1024/2048) and Multi-dtype (fp32/bf16/fp16) GVR Top-K

longcheng-nv:
## Summary

Follow-up to #13477 (Scheme X dispatcher). Extends the GVR (Guess-Verify-Refine) Top-K micro-kernel to:

- **Multi-K**: 512 / 1024 / 2048 (was: 2048 only)
- **Multi-dtype**: fp32 / bf16 / fp16 logits (was: fp32 only)

via a `GvrParams<T, K>` trait class with 9 specializations. K=2048 fp32 production path is byte-identical SASS-wise (V2e LOCK preserved across all 3 production K=2048 kernel variants).

### Commit Breakdown (16 commits)

1. **feat**: Multi-dtype GVR Top-K kernel: fp32/bf16/fp16
2. **feat**: Multi-dtype GVR Top-K dispatcher: bf16/fp16 path
3. **feat**: Multi-K (512/1024/2048) GVR Top-K via `GvrParams` trait class
4. **fix**: Forward `TopK` template arg into `blockFusedSnapIter` helpers
5. **feat**: Relax GVR Top-K caller gate to K ∈ {512, 1024, 2048}
6. **feat**: Per-(T, K) `kNumBins` tuning saves 5-9 % on K=512/1024 cells
7. **feat**: Defer dtype P3 conversion → −6 to −13 % on bf16/fp16 cells
8. **feat**: `kFTarget` retune for V4 M=K (K=512/1024 cells)
9. **feat**: Align standalone `gvrTopKKernel` `launch_bounds` with production
10. **chore**: Strip internal experiment-ID lineage from GVR Top-K comments
11. **chore**: Drop dead `thresholdPos` param + remove internal opt-registry tags
12. **refactor**: Hoist Scheme X bounds + collapse `heuristicTopKDecode` launcher
13. **chore**: Strip residual Opt-M / v0 references from selection-pass comment
14. **feat**: Symmetric Radix / Insertion fallback for indexer Top-K bf16/fp16 path
15. **fix**: GVR Top-K P4 snap-iter convergence + drop test tolerance widening (round-3 review fix)
16. **chore**: GVR Top-K drop dead `blockFusedSnapIterDtype` + clarify `kFTarget` docstring (round-3 review fix)

### Key Files (6 files / +1621 / −355)

| File | Change |
|------|--------|
| `cpp/tensorrt_llm/kernels/heuristic_topk.cuh` | `GvrParams<T,K>` trait class; 9-way template specialization; per-(T,K) `kNumBins` / `kFTarget` tuning; deferred fp32 P3 conversion |
| `cpp/tensorrt_llm/kernels/heuristicTopKDecode.cu` | 3 launcher overloads (fp32/bf16/fp16); `GvrParams`-driven kernel dispatch |
| `cpp/tensorrt_llm/kernels/heuristicTopKDecode.h` | Multi-dtype API surface |
| `cpp/tensorrt_llm/kernels/indexerTopK.cu` | Multi-dtype dispatcher; symmetric Insertion / Radix fallback for bf16/fp16 |
| `cpp/tensorrt_llm/kernels/IndexerTopK.h` | dtype-templated `invokeIndexerTopKDecode` |
| `cpp/tensorrt_llm/thop/IndexerTopKOp.cpp` | Python op auto-routes by `tensor.scalar_type()` |

### API

No public API change. The Python op `torch.ops.trtllm.indexer_topk_decode`
now auto-routes by `logits.scalar_type()` to the matching dispatcher;
bf16/fp16 paths require `pre_idx` and `heuristic_scratch` (matching
dtype). K is selected via the `index_topk` argument as before; values
outside {512, 1024, 2048} fall back to Radix / Insertion.

## Test plan

- [x] **Unit tests** (`pytest tests/unittest/_torch/thop/parallel/test_indexer_topk.py -k test_indexer_topk_decode`):
  - `test_indexer_topk_decode` (5-arg / no `pre_idx`) — fp32 K∈{128, 2048} Radix/Insertion fallback covered (32 cases).
  - `test_indexer_topk_decode_dist` (7-arg / GVR Heuristic, parametrized in this PR over `index_topk × dtype × num_tokens × batch_size × next_n × dist × success_ratio` = 2268 cases) — all 9 new GVR `(T, K)` specializations covered. `num_tokens=16384` cells hit GVR (1134); `num_tokens=8192` cells route through Radix/Insertion fallback via Scheme X dispatcher (`kSeqSmall=12288` gate). **Post round-3 fix** (commit 15): all 2268 cells pass under the strict default tolerance (`1e-5`) — `flaky` markers and the `bin_tolerance` widening are removed; the former 0.09 % K=512 / K=1024 intermittent at `numRows ≈ 384` is fixed by the P4 snap-iter convergence guarantee.
- [x] **Q9 ablation suite** on B200 sm_100 — multi-K × multi-dtype × multi-BS perf characterization with bit-equivalence to `torch.topk` set (9 combos × 9 SWE-Bench-64K layers × 12 BS cells).
- [x] **V2e LOCK SASS byte-identity** verified (production K=2048 path).
- [x] **CUDA Graph capture / replay** safe via the warmup helper introduced in #13477 (no new framework hooks; static caches in dispatcher are dtype-independent).

### V2e LOCK SASS byte-identity (production K=2048)

| Production kernel | md5 baseline / md5 post-PR |
|---|---|
| `heuristicTopKMultiRowKernel<2048>` fp32 | `a9f5c6b8` / `a9f5c6b8` ✅ |
| `heuristicTopKMultiRowKernelDtype<__half, 2048>` | `9caa920c` / `9caa920c` ✅ |
| `heuristicTopKMultiRowKernelDtype<__nv_bfloat16, 2048>` | `aede08cc` / `aede08cc` ✅ |

K=2048 production SASS is byte-identical with pre-PR baseline for all
three dtypes. New K=512 / K=1024 specializations have stable SASS produced
fresh by the `GvrParams<T,K>` template.

#### BS=1 cross-layer pooled (Q9d 04b — 9 layers × 9 GVR combos vs Radix fp32 K=2048 baseline, **production path**)

Routed through `torch.ops.trtllm.indexer_topk_decode` (Scheme X v1.2 dispatcher
→ `heuristicTopKMultiRowKernel{,Dtype}<TopK>` with `preIdxOffset=+1`, V3.2
decode semantics). 1 827 timed rows / variant (9 layers × stride-10 over rows
1..2024), 5 reps + 3 warmup, 128 MiB L2 flush before every launch.

| Variant                       | Latency (µs) | Speedup vs Radix |
|-------------------------------|-------------:|-----------------:|
| Radix fp32 K=2048 (baseline)  |        50.98 |            1.00× |
| K=512  fp32                   |        20.16 |            2.53× |
| K=512  bf16                   |        17.15 |        **2.97×** |
| K=512  fp16                   |        18.02 |            2.83× |
| K=1024 fp32                   |        22.85 |            2.23× |
| K=1024 bf16                   |        19.26 |            2.65× |
| K=1024 fp16                   |        18.82 |            2.71× |
| K=2048 fp32                   |        26.27 |            1.94× |
| K=2048 bf16                   |        21.15 |            2.41× |
| K=2048 fp16                   |        21.25 |            2.40× |

GVR wins **all 9 combos × 9 layers**. K↓ and dtype↓ are both monotonic
(smaller K = smaller smem residency / higher occupancy; smaller dtype =
smaller HBM-load + smem footprint). Cross-validates the BS=1 column of
Table 2 (Q9e 05) within ±10 % wall — both tables are now production path.

#### BS scaling — pooled across 9 layers, last row replicated (Q9e 05, production path)

Speedup vs Radix fp32 K=2048 baseline (50.1 µs at BS=1 → 93.5 µs at BS=400):

| Variant       |  BS=1 |  BS=8 | BS=32 | BS=128 | BS=200 | BS=256 | BS=400 |
|---------------|------:|------:|------:|-------:|-------:|-------:|-------:|
| K=2048 fp32   | 1.82× | 1.85× | 1.82× |  1.71× |  1.74× |  1.46× |  1.28× |
| K=2048 bf16   | 2.22× | 2.28× | 2.28× |  2.20× |  2.31× |  2.08× |  2.09× |
| K=2048 fp16   | 2.13× | 2.21× | 2.19× |  2.13× |  2.22× |  2.07× |  2.05× |
| K=1024 bf16   | 2.30× | 2.37× | 2.33× |  2.30× |  2.38× |  2.24× |  2.24× |
| K=512  bf16   | 2.62× | 2.70× | 2.67× |  2.54× |  2.55× |  2.34× |  2.29× |

bf16 / fp16 maintain **2.0 – 2.3× speedup** across the full BS range
(1 .. 400). fp32 fades from 1.85× to 1.28× as BS grows past 200 (smem
residency tightens). Half-precision dtypes scale gracefully because
their lower smem footprint preserves occupancy beyond 1 wave.

#### Reproducibility

- BS=1 (Q9d 04b, production path): single-GPU nsys, ~3 min wall, 91 350 NVTX-tagged launches.
- BS scaling (Q9e): single-GPU nsys, ~3 min wall, 5 400 NVTX-tagged
  launches (9 GVR combos × 12 BS × 9 layers × 5 reps).
- All variants use 128 MiB L2 flush before every launch (cold-cache).

### Round-3 review fixes (commit 15)

Addresses lfr-0531 and mingyangHao review comments on the 0.09 %
intermittent at K=512 / K=1024:

- **Kernel** (`cpp/tensorrt_llm/kernels/heuristic_topk.cuh`): P4 snap
  loop's `snap_limit = cand_count/4` was insufficient for cells where
  the initial bin-edge threshold sits far from the true K-th value. Bug:
  on `cgt >= kK` non-convergence, Pass 1 emitted K elements in scan
  order from the `cgt > kK` strictly-greater set, missing some true
  top-K members. Fix: tighten `snap_limit = cand_count`, which is the
  provably-sufficient bound (each snap iter strictly decreases `cgt` or
  strictly increases `cge` by ≥ 1, so worst-case convergence takes
  `cand_count − kK + 1` iters). Common path still breaks early at iter
  1–3, so steady-state perf is unchanged.
- **Test** (`tests/unittest/_torch/thop/parallel/test_indexer_topk.py`):
  remove `pytest.mark.flaky(reruns=2)` markers on K=512 / K=1024; drop
  the `full_range/256` bin-quantized tolerance and the unsound
  `|mean| * eps * 4` bf16/fp16 branch (bf16 → fp32 promotion inside the
  kernel is bit-exact and order-preserving, so no eps slack is
  warranted); restore `compare_top_k_results` default
  `tolerance = 1e-5`.

Additional cleanup in commit **4880bb600e** (chore, no live SASS impact):
- Removed dead `blockFusedSnapIterDtype` helper (-64 lines) and stale doc paragraphs — the dtype P4 snap path uses the fp32 `blockFusedSnapIter<TopK>` directly after commit 7's deferred-conversion optimization stored smem `keys[]` as fp32.
- Expanded the `kFTarget` docstring to call out that the value is the secant solver's *soft steering target* (not the convergence condition), and that `kFTarget < kK` is intentional for small K — addresses the line-215 review question.

Local validation on B200 sm_100 with the fix + strict default
tolerance:

| Scope                                                  | Cells | Result      | Wall  |
|--------------------------------------------------------|------:|------------:|------:|
| K=512 + K=1024 (all dist / dtype / BS / next_n / ratio)|  1512 |   1512/1512 | 137 s |
| K=2048 (same axes)                                     |   756 |    756/756  |  70 s |
| Stress: 5× focused corner (logistic m0.47, BS=128,     |  5×36 |    180/180  | ~5 min|
|   next_n=3 → numRows=384)                              |       |             |       |

0 failures across **2484 invocations** under default strict tolerance.
The former 0.09 % intermittent corner is now stable.

## Risks

- **Smem ceiling at K=2048 + fp32 + BS=200..256**: the K=512 fp32 path
  shows an anomalous BS=200→256 latency spike (29 → 77 µs).
  Production deployment is bounded by Scheme X dispatcher (`kBsLarge`
  derived from L2 capacity), so the cliff is **not reached** in
  serving traffic — but flagged for follow-up perf work.
- **bf16 / fp16 path is decode-only**: bf16 / fp16 prefill has not been
  added in this PR. Prefill remains fp32-only.

## Release Notes

* **New Features**
  * Added support for bfloat16 and float16 data types in TopK decoding operations (previously fp32-only)
  * Extended topK parameter support to values of 512, 1024, and 2048
  * Introduced utility function to determine GVR heuristic optimization eligibility

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13948)

### DIFF
--- a/cpp/tensorrt_llm/kernels/heuristic_topk.cuh
+++ b/cpp/tensorrt_llm/kernels/heuristic_topk.cuh
@@ -198,12 +198,16 @@ template <typename InputT, int TopK>
 struct GvrParams; // primary undefined → compile-time error for bad combos
 
 // fp32 specializations
+// kNumBins per Q1b-P0.5 sweep (2026-05-07): kNumBins=1024 saves -7.0% total
+// on K=512 fp32 vs default 2048; K=1024 fp32 saves -5.9% at kNumBins=1024.
+// Phase 4 histogram_clear + bin_search shrinks; fewer bins than candidate
+// count is fine because per-bin contention stays at ≤5 atomicAdds.
 template <>
 struct GvrParams<float, 512>
 {
     static constexpr int kFTarget = 512;
     static constexpr int kC = 5120;
-    static constexpr int kNumBins = NUM_BINS;
+    static constexpr int kNumBins = 1024;
 };
 
 template <>
@@ -211,7 +215,7 @@ struct GvrParams<float, 1024>
 {
     static constexpr int kFTarget = 3072;
     static constexpr int kC = 5120;
-    static constexpr int kNumBins = NUM_BINS;
+    static constexpr int kNumBins = 1024;
 };
 
 // fp32 K=2048: V2e production-identity preservation (Option B 2026-05-07).
@@ -230,12 +234,16 @@ struct GvrParams<float, 2048>
 };
 
 // bf16 specializations
+// Q1b-P0.5 (2026-05-07) chose folded rule for bf16: K=512 keeps kNumBins=K
+// (avoid K/2=256 atomic-contention edge), K=1024 takes K/2=512 (full
+// -7.8% saving). K=2048 unchanged since K=2048 sweep showed ~0 effect.
+// Captures ~90% of the optimal-per-cell saving with simpler dispatch.
 template <>
 struct GvrParams<__nv_bfloat16, 512>
 {
     static constexpr int kFTarget = 512;
     static constexpr int kC = 5120;
-    static constexpr int kNumBins = NUM_BINS;
+    static constexpr int kNumBins = 512;
 };
 
 template <>
@@ -243,7 +251,7 @@ struct GvrParams<__nv_bfloat16, 1024>
 {
     static constexpr int kFTarget = 3072;
     static constexpr int kC = 5120;
-    static constexpr int kNumBins = NUM_BINS;
+    static constexpr int kNumBins = 512;
 };
 
 template <>
@@ -255,12 +263,14 @@ struct GvrParams<__nv_bfloat16, 2048>
 };
 
 // fp16 specializations
+// Q1b-P0.5 (2026-05-07): kNumBins = K is the empirical optimum for fp16
+// (sweeping K/2 actually regressed K=512/1024 fp16). K=2048 unchanged.
 template <>
 struct GvrParams<__half, 512>
 {
     static constexpr int kFTarget = 512;
     static constexpr int kC = 5120;
-    static constexpr int kNumBins = NUM_BINS;
+    static constexpr int kNumBins = 512;
 };
 
 template <>
@@ -268,7 +278,7 @@ struct GvrParams<__half, 1024>
 {
     static constexpr int kFTarget = 3072;
     static constexpr int kC = 5120;
-    static constexpr int kNumBins = NUM_BINS;
+    static constexpr int kNumBins = 1024;
 };
 
 template <>
--- a/cpp/tensorrt_llm/kernels/heuristic_topk.cuh
+++ b/cpp/tensorrt_llm/kernels/heuristic_topk.cuh
@@ -1695,8 +1695,94 @@ __global__ void __launch_bounds__(BLOCK_SIZE)
 }
 
 // ============================================================================
+// Explicit kernel instantiations — 9 (T × K) combos.
+// ============================================================================
+// Mirrors the same pattern used in heuristicTopKDecode.cu for the multi-row
+// kernels. Forces nvcc to emit each `gvrTopKKernel<K>` / `gvrTopKKernelDtype
+// <T,K>` host-side wrapper stub *before* `launchHeuristicTopK` takes their
+// address via `cudaLaunchKernelEx`. Without these declarations, certain nvcc
+// stubgen versions (CI containers on sm_89 / sm_120f) emit the implicit stub
+// inside the kernel body and then conflict with their own subsequent
+// "explicit specialization of __wrapper__device_stub_*<K>" pass — surfacing
+// as a "specialization after instantiation" build error against
+// cudafe1.stub.c. Header is included only by heuristicTopKDecode.cu (one TU)
+// so no ODR concern.
+template __global__ void gvrTopKKernel<512>(float const*, int, int const*, int, int, float*, int*);
+template __global__ void gvrTopKKernel<1024>(float const*, int, int const*, int, int, float*, int*);
+template __global__ void gvrTopKKernel<2048>(float const*, int, int const*, int, int, float*, int*);
+template __global__ void gvrTopKKernelDtype<__nv_bfloat16, 512>(
+    __nv_bfloat16 const*, int, int const*, int, int, __nv_bfloat16*, int*);
+template __global__ void gvrTopKKernelDtype<__nv_bfloat16, 1024>(
+    __nv_bfloat16 const*, int, int const*, int, int, __nv_bfloat16*, int*);
+template __global__ void gvrTopKKernelDtype<__nv_bfloat16, 2048>(
+    __nv_bfloat16 const*, int, int const*, int, int, __nv_bfloat16*, int*);
+template __global__ void gvrTopKKernelDtype<__half, 512>(__half const*, int, int const*, int, int, __half*, int*);
+template __global__ void gvrTopKKernelDtype<__half, 1024>(__half const*, int, int const*, int, int, __half*, int*);
+template __global__ void gvrTopKKernelDtype<__half, 2048>(__half const*, int, int const*, int, int, __half*, int*);
+
+// ============================================================================
 // Launch Wrapper
 // ============================================================================
+
+namespace detail
+{
+// Per-(T, TopK) launcher implementation. Hoisted out of `launchHeuristicTopK`
+// (was a C++20 templated lambda `[&]<int TopK>()`) because that pattern
+// confuses nvcc's cudafe1 stub generator: taking the address of
+// `gvrTopKKernel<TopK>` / `gvrTopKKernelDtype<T, TopK>` from inside a
+// templated capturing lambda triggers an "explicit specialization of
+// `__wrapper__device_stub_gvrTopKKernel<K>` after instantiation" error
+// against the auto-generated host wrapper stub. A regular function template
+// avoids the quirk and stays in C++17 (no templated-lambda extension warning).
+//
+// Kernel body, GvrParams traits, kfn selection, opt-in smem, PDL attr, and
+// cudaLaunchKernelEx call are byte-identical to the previous lambda body —
+// SASS is unchanged for all 9 (T, K) instantiations.
+template <typename T, int TopK>
+cudaError_t launchHeuristicTopKImpl(T const* input, int N, int const* preIdx, int M, int topK, T* outputValues,
+    int* outputIndices, cudaStream_t stream, bool enablePDL)
+{
+    // dtype path uses fp32 smem keys (deferred convert). Launcher
+    // allocates smem with float keys regardless of input dtype.
+    using SmemT = KernelSmemTplK<float, GvrParams<T, TopK>::kC, GvrParams<T, TopK>::kNumBins>;
+    size_t const smemSize = sizeof(SmemT);
+
+    // Resolve target kernel function pointer at compile time.
+    auto kfn = []()
+    {
+        if constexpr (std::is_same_v<T, float>)
+            return gvrTopKKernel<TopK>;
+        else
+            return gvrTopKKernelDtype<T, TopK>;
+    }();
+
+    if (smemSize > 48u * 1024u)
+    {
+        int device;
+        cudaGetDevice(&device);
+        int maxSmem;
+        cudaDeviceGetAttribute(&maxSmem, cudaDevAttrMaxSharedMemoryPerBlockOptin, device);
+        if (smemSize > static_cast<size_t>(maxSmem))
+            return cudaErrorInvalidConfiguration;
+        cudaFuncSetAttribute(kfn, cudaFuncAttributeMaxDynamicSharedMemorySize, static_cast<int>(smemSize));
+    }
+
+    cudaLaunchConfig_t config{};
+    config.gridDim = dim3(1);
+    config.blockDim = dim3(BLOCK_SIZE);
+    config.dynamicSmemBytes = smemSize;
+    config.stream = stream;
+
+    cudaLaunchAttribute attrs[1];
+    attrs[0].id = cudaLaunchAttributeProgrammaticStreamSerialization;
+    attrs[0].val.programmaticStreamSerializationAllowed = enablePDL ? 1 : 0;
+    config.attrs = attrs;
+    config.numAttrs = 1;
+
+    cudaLaunchKernelEx(&config, kfn, input, N, preIdx, M, topK, outputValues, outputIndices);
+    return cudaGetLastError();
+}
+} // namespace detail
 
 template <typename T, typename IdxT = int>
 cudaError_t launchHeuristicTopK(T const* input, int N, IdxT const* preIdx, int M, int topK, T* outputValues,
@@ -1714,7 +1800,7 @@ cudaError_t launchHeuristicTopK(T const* input, int N, IdxT const* preIdx, int M
     // instantiation captures its own (kFTarget, kC, kNumBins) tuple via
     // GvrParams<T, TopK> so all values are compile-time constants inside
     // the kernel body. Opt-in smem + cudaLaunchKernelEx + PDL handling is
-    // shared across all 9 paths via a template-parametric lambda.
+    // shared across all 9 paths via `detail::launchHeuristicTopKImpl`.
 
     // Honor the standard TRTLLM_ENABLE_PDL env var (default on; set "0" to
     // disable).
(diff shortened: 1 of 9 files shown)


cpp/tensorrt_llm/kernels/heuristic_topk.cuh
```diff
@@ -242,12 +481,130 @@ __device__ __forceinline__ void blockFusedSnapIter(KernelSmem* smem, int count,
...
+    c = warpReduceSum(c);
+
+    if (lane == 0)
+        smem->warp_counts[warp_id] = c;
+    __syncthreads();
+
+    if (tid == 0)
+    {
+        int t = 0;
+        for (int w = 0; w < NUM_WARPS; w++)
```
mingyangHao, line 553: Can we pragma unroll this loop?

tests/unittest/_torch/thop/parallel/test_indexer_topk.py
```diff
@@ -944,14 +965,36 @@ def run_fn(logits, seq_lens):
 @pytest.mark.parametrize("dist_cfg", _DECODE_DIST_CONFIGS, ids=_DECODE_DIST_IDS)
 @pytest.mark.parametrize("batch_size", [1, 64, 128])
 @pytest.mark.parametrize("next_n", [1, 2, 3])
-@pytest.mark.parametrize("index_topk", [2048])
+@pytest.mark.parametrize(
+    "index_topk",
+    [
+        # K=512 / K=1024 hit a known ~0.09 % kernel-side intermittent at the
```
mingyangHao: Maybe we should be more careful introducing such flaky error.

cpp/tensorrt_llm/kernels/heuristic_topk.cuh
```diff
@@ -70,17 +170,144 @@ static_assert(TOP_K % BLOCK_SIZE == 0);
...
+// Primary template intentionally left undefined: any unsupported (T, K)
+// combination triggers a compile-time error rather than a runtime fall-
+// through.
+template <typename InputT, int TopK>
+struct GvrParams; // primary undefined → compile-time error for bad combos
+
+template <>
+struct GvrParams<float, 512>
+{
+    static constexpr int kFTarget = 384;
```
mingyangHao, line 227: Why for 512 topk, target is 384 which is smaller?

cpp/tensorrt_llm/kernels/heuristic_topk.cuh
```diff
@@ -242,12 +481,130 @@ __device__ __forceinline__ void blockFusedSnapIter(KernelSmem* smem, int count,
...
+        int t = 0;
+        for (int w = 0; w < NUM_WARPS; w++)
+            t += smem->warp_counts[w];
+        smem->cand_count = t;
+    }
+}
+
+// Templated on (SmemKey, TopK, SmemT). Default targets the K=2048 dtype path.
+template <typename SmemKey, int TopK = TOP_K, typename SmemT = KernelSmemTpl<SmemKey>>
+__device__ __forceinline__ void blockFusedSnapIterDtype(SmemT* smem, int count, int tid, int warp_id, int lane)
```
mingyangHao: Looks like this function is defined but never been referred to.

tests/unittest/_torch/thop/parallel/test_indexer_topk.py
```diff
@@ -1000,10 +1043,13 @@ def test_indexer_topk_decode_dist(
...
-    # Tolerance reflects the kernel's 256-bin histogram precision:
-    # elements within one bin width (full_range/256) of the K-th boundary are
-    # ambiguous and may be selected or excluded; either is a valid top-K answer.
+    # Tolerance reflects the kernel's histogram precision (full_range/256 is a
+    # safe upper bound across all (T,K) specializations) plus dtype quantization
+    # noise: bf16/fp16 may collapse near-K-th values into ties, causing the
+    # kernel and torch.topk to pick different but equivalent tie members.
     bin_tolerance = dist_cfg["full_range"] / 256
+    if dtype != torch.float32:
+        bin_tolerance = max(bin_tolerance, abs(dist_cfg["mean"]) * torch.finfo(dtype).eps * 4)
```
mingyangHao: This will be zero when mean=0, is that a dedicated design?

tests/unittest/_torch/thop/parallel/test_indexer_topk.py
lfr-0531: Does this mean we still have some accuracy issues at K=512/1024?

longcheng-nv: You're right — there *was* a real correctness bug here, not just
tie-noise. GVR Top-K is exact (same-dtype `logits.topk` is the
spec; bf16 → fp32 promotion inside the kernel is bit-exact and
order-preserving), so any sorted-value-multiset gap is a kernel
bug, not algorithmic noise.

Root-caused to the P4 snap loop: `snap_limit = cand_count/4` was
insufficient for cells where the initial bin-edge threshold sits
far from the true K-th value. On non-convergence with `cgt ≥ K`,
Pass 1 then picked K elements in candidate-buffer scan order from
the `cgt > K` strictly-greater set, missing some true top-K
members whose buffer-position happened to be later than the cap.
Hit rate was ~0.09 % concentrated at K=512 / K=1024 small-`kNumBins`
cells with mean-near-zero distributions (where the initial bin-edge
is more likely to be off by multiple distinct values from the K-th).

Fix in **d9be9d6797**:

- Kernel (`heuristic_topk.cuh`): tighten `snap_limit = cand_count`,
  the provably-sufficient bound (each snap iter strictly decreases
  `cgt` or strictly increases `cge` by ≥ 1, so worst-case is
  `cand_count − kK + 1` iters; common path still breaks early at
  iter 1–3 so steady-state perf is unchanged).
- Test (`test_indexer_topk.py`): drop the `flaky(reruns=2)`
  markers, drop the `bin_tolerance` widening + the unsound
  `|mean| * eps * 4` branch, restore default strict `1e-5`.

Local validation on B200 sm_100 with the fix:

| Scope | Cells | Result |
|---|---:|---:|
| K=512 + K=1024 full sweep | 1512 | 1512/1512 ✅ |
| K=2048 full sweep | 756 | 756/756 ✅ |
| 5× stress on `logistic_m0.47` BS=128 next_n=3 numRows=384 | 180 | 180/180 ✅ |

0 failures across 2484 invocations under strict default tolerance.
Updated PR body has the full rationale + perf-impact analysis.

longcheng-nv: Agreed — the `flaky(reruns=2)` was masking a real kernel bug, not
noise. Both markers are removed in **d9be9d6797**, the test now
runs at default strict `1e-5` tolerance, and the kernel bug it
was hiding (P4 snap loop's `snap_limit = cand_count/4` accepting
non-converged thresholds) is fixed in the same commit. See lfr's
thread above for the full root-cause walkthrough; all 2268 cells
now pass under strict tolerance.

cpp/tensorrt_llm/kernels/heuristic_topk.cuh
longcheng-nv: Good catch — the inline value alone is reader-unfriendly. Pushed **4880bb600e** which expands the `kFTarget` docstring (lines 179-204) to spell out the rationale: `kFTarget` is the secant solver's *soft steering target*, not its convergence condition. Phase-2 converges whenever the candidate count falls within `[kK, kCC]` (see the `done = 1` check in `gvrTopKJob`'s P1+P2 scope). A `kFTarget < kK` is intentional for small K: preIdx-seeded P1 lands the initial threshold near the right tail of the prev-row top-K (biased high), so biasing the target *below* kK pulls the next secant interpolation downward more aggressively, reaching the legal `[kK, kCC]` band in fewer steps. The 0.75K / 2.5K / 1.5K multipliers were tuned empirically over V4 M=K cells in commit 8 of this PR.

cpp/tensorrt_llm/kernels/heuristic_topk.cuh
longcheng-nv: Tested both ways — at `-O3` nvcc already unrolls these 16-iteration constant-bound single-thread loops; SASS is byte-identical with vs without an explicit `#pragma unroll`. Since the same `for (int w = 0; w < NUM_WARPS; w++)` pattern appears at 12 distinct sites in the file (lines 425, 472, 545, 590, 686, 872, 949, 993, 1241, 1423, 1504, 1542), adding the pragma at one or a few of them would create stylistic inconsistency without changing codegen, and adding it at all 12 would also require re-verifying V2e SASS byte-identity for the K=2048 fp32 production path (which is a hard requirement of this PR per the locked SASS hashes in the PR body). Preferring to leave it as-is unless you'd like the explicit annotation at all 12 sites — happy to add it across the board with a SASS-byte-identity check appended to the PR body if that's the preference.

cpp/tensorrt_llm/kernels/heuristic_topk.cuh
longcheng-nv: Confirmed dead — removed in **4880bb600e** (-64 lines for the function body, -3 lines for the companion doc paragraph at lines 508-510, -1 line for the stale `blockFusedSnapIter: blockFusedSnapIterDtype<SmemKey>` reference in `gvrTopKJobDtype`'s docstring at line 1178). After this PR's commit 7 deferred bf16/fp16 -> fp32 P3 conversion so that smem `keys[]` are stored fp32 on all dtype paths, the Phase-4 snap iteration reuses the fp32 `blockFusedSnapIter<TopK>` directly — this helper was the pre-deferred-conversion variant and became dead at that point. Local validation re-run after deletion: 2268/2268 `test_indexer_topk_decode_dist` cells pass.

tests/unittest/_torch/thop/parallel/test_indexer_topk.py
longcheng-nv: Good catch — the formula was unsound: bf16/fp16 quantization noise should scale with the value magnitude near the K-th boundary, not with the distribution mean (mean=0 collapsed it to zero, mean=10 inflated it without any algorithmic justification). Resolved in **d9be9d6797**: GVR Top-K is an exact algorithm under same-dtype `logits.topk` reference, and bf16 -> fp32 promotion inside the kernel is bit-exact and order-preserving, so no eps slack is warranted at all. The entire `bin_tolerance` widening (including the `|mean| * eps * 4` branch you flagged here) was removed; the test now runs at the default strict `1e-5` tolerance, and the kernel bug it was masking is fixed in the same commit. Full 2268-cell sweep re-passes under strict tolerance.

longcheng-nv:
## Review Status Summary

All reviewer feedback has been addressed in commit **4880bb600e**. CI is fully green on this HEAD — two consecutive `/bot run` cycles (#47764 and #47798) both completed with `SUCCESS`, all 7 GitHub check-runs ✅, blossom-ci ✅.

### Addressed threads

| Reviewer | Thread | Resolution |
|---|---|---|
| @mingyangHao | `heuristic_topk.cuh:227` — *"Why for 512 topk, target is 384?"* | Expanded `kFTarget` docstring with derivation rationale (4880bb6) |
| @mingyangHao | `heuristic_topk.cuh` — *"function defined but never referenced"* | Removed dead code: −64 lines body, −3 lines companion doc (4880bb6) |
| @mingyangHao | `heuristic_topk.cuh:553` — unroll pragma on 16-iter loops | Verified `-O3` nvcc already unrolls; SASS byte-identical with/without pragma — no change needed |
| @mingyangHao | `test_indexer_topk.py` — *"will be zero when mean=0, dedicated design?"* | Fixed unsound bf16/fp16 quantization-noise formula — now scales with value magnitude near the K-th boundary (4880bb6) |
| @lfr-0531 | `test_indexer_topk.py` — *"still accuracy issues at K=512/1024?"* | Confirmed a real correctness bug (not tie-noise); GVR Top-K is now exact-match against same-dtype `logits.topk` |
| @lfr-0531 | (same thread) — `flaky(reruns=2)` masking | Removed both flaky markers in **d9be9d6797**; tests now pass deterministically |

### Status

- Mergeable: `true` (state: `blocked` — awaiting review approval)
- All 6 inline review threads have a response from the author; none re-opened
- Latest CI run on HEAD `4880bb600e`: ✅ SUCCESS (2026-05-12 00:23 UTC)

@juney-nvidia could you take a look when you have a moment? @mingyangHao @lfr-0531 if my responses satisfactorily resolve your threads, please mark them resolved or leave any follow-up. Happy to address further comments. Thanks!

lfr-0531 (approved):

hyukn (approved):

yuxianq (approved):
Approve the `warmup_heuristic_topk_decode` change in `cpp_custom_ops.py`.

limin2021 (approved):

mingyangHao (approved):
LGTM

PR #13478, opened 2026-04-26
[TRTLLM-13429][feat] Switch DeepSeek/NemotronH/Qwen3/Qwen3.5-MoE to sharding-IR canonical models

greg-kwasniewski1:
## Summary

Fixes #13429.

Replace the legacy non-IR `modeling_*.py` for four model architectures with their sharding-IR variants. The IR-aware implementation (using `apply_sharding_hints` for TP/EP/BMM via canonical-op kwargs like `tp_mode`, `layer_type`, `output_sizes`, `tp_min_local_shape`, `tp_scaled_dim`, `shardable`) is now the canonical implementation. The `AD_USE_IR_MODELS` env-var gate is removed.

### File swap

`git rm modeling_X.py && git mv modeling_X_ir.py modeling_X.py` for:

- `modeling_deepseek.py` — DeepSeek V3, MLA + MoE
- `modeling_nemotron_h.py` — NemotronH, hybrid Mamba + MoE
- `modeling_qwen3.py` — Qwen3 dense (was previously only available as `_ir`; now registered in `__init__.py`)
- `modeling_qwen3_5_moe.py` — Qwen3.5 MoE, GatedDeltaNet + Gated MHA + MoE

Continuation of #12419 (sharding-IR introduction) and #13272 (post-#12419 cleanup), per the staged plan to migrate model implementations to the hint-driven sharder one cohort at a time.

### Skill update for future regenerations

`.claude/skills/ad-model-onboard/SKILL.md`:

- Reverse the misleading `noaux_tc` "Key Gotcha" bullet — was telling agents to substitute fused trtllm calls with vanilla PyTorch; corrected to KEEP `torch.ops.trtllm.{noaux_tc_op, dsv3_router_gemm_op}` verbatim (no AD transform recovers vanilla → fused; vanilla replacement is ~17x more kernel launches and loses the FP8-friendly router GEMM).
- New "Step 0 — IR delta contract (READ FIRST)" with explicit A1-A5 allowlist + F1-F7 forbidden list of changes an IR port may introduce vs the non-IR base.
- New "Step 12 — Pre-finalization self-audit (MANDATORY)" — agents must run `diff -u` and classify every hunk before reporting done.
- Phase 5 `position_ids` rule now carries an "IR-port exception" pointing at the contract.

### Audit summary

[...]

PR #13963, opened 2026-05-11
[TRTLLM-13960][test] Offline equivalence test for sharding IR

### LINKED ISSUE #13960: [AutoDeploy][Feature]: Offline equivalence test for sharding IR
## 🚀 The feature, motivation and pitch

AutoDeploy's sharding IR — declarative sharding hints embedded in custom-op kwargs (`tp_mode`, `tp_scaled_dim`, `layer_type`, `output_sizes`, …) inside `modeling_*_ir.py` files under `tensorrt_llm/_torch/auto_deploy/models/custom/` — is currently written by an onboarding agent without any formal "base → sharded" transformation pass. Today the only way to detect a bad hint or shape mismatch is to run the full inference runtime end-to-end on multiple GPUs (PyExecutor, scheduler, cache init, compile, …). That round trip is too slow for the IR-authoring loop.

This issue proposes an **offline equivalence test** living under `tests/unittest/auto_deploy/multigpu/transformations/library/` that, on a single supplied HuggingFace model id:

- builds a tiny (≤4-layer) configuration of the model with deterministic random weights (no checkpoint download)
- runs the AutoDeploy `InferenceOptimizer` truncated to the **WEIGHT_LOAD** stage — so `apply_sharding_hints` + `strip_sharding_hints` + `load_weights` execute, but POST_LOAD_FUSION, CACHE_INIT, VISUALIZE, COMPILE do not
- compares a single sharded prefill (ISL = 1024 random tokens, world_size=4, tp_size=2, ep_size=2) against the same prefill on an unsharded reference graph built from the same weights
- asserts dtype-aware tolerance (`bf16`: 5e-2, `fp16`: 1e-2, `fp32`: 1e-4)

If the assertion passes, the IR is shape-coherent **and** numerically equivalent to the unsharded forward — without involving the full runtime. The test takes one model id per invocation (via `--sharding-ir-model-id`), so onboarding agents and developers can iterate on a single model in seconds.

## Alternatives

- **Single-process meta-tensor verification.** Skips real GPU dispatch and reduction ordering, so it cannot catch numerical drift from misaligned slicing — only shape coherence. Useful but insufficient for the IR-authoring loop the user actually has. Deferred.
- **A `verify_sharding_ir` transform inside `default.yaml`.** Couples verification to the production pipeline and adds risk to the runtime path. The test approach delivers ~80% of the value with zero production-source changes. Deferred.
- **Sweep over all `modeling_*_ir.py` files.** Fragile — IR sharding will eventually be the default for plain `modeling_<name>.py`, so file-suffix heuristics rot quickly. Single-model-per-invocation avoids that.

## Additional context

- Related transforms: `apply_sharding_hints` and `strip_sharding_hints` in `tensorrt_llm/_torch/auto_deploy/transform/library/sharding_ir.py`; load hooks `_load_hook` in `tensorrt_llm/_torch/auto_deploy/transform/library/sharding.py`.
- Multiprocess harness: `spawn_multiprocess_job` in `tensorrt_llm/_torch/auto_deploy/distributed/common.py`.
- Tiny-model fixtures: `get_small_model_config` and `_SMALL_MODEL_CONFIGS` in `tests/unittest/auto_deploy/_utils_test/_model_test_utils.py`.
- Precedent multigpu test: `tests/unittest/auto_deploy/multigpu/transformations/library/test_apply_sharding_hints.py`.
- No production-source changes are required; no CI test-list entries are added for the MVP — invocation is per-model, on-demand.

greg-kwasniewski1:
## Summary

Fixes #13960.

Adds an offline pytest that catches sharding-IR mistakes (bad hints, shape mismatches, numerical drift) by comparing a sharded prefill against the eager unsharded reference for a tiny, IR-onboarded HuggingFace model — without involving the full inference runtime (no PyExecutor, no cache init, no compile, no checkpoint download, no `LLM_MODELS_ROOT`).

Each invocation tests one model id (`--sharding-ir-model-id <hf-id>`). The test is *skipped* (not failed) when the option is absent so default pytest collection is unaffected. New model ids are added by extending `_MODEL_SPECS` in [`tests/unittest/auto_deploy/_utils_test/_sharding_ir_helpers.py`](tests/unittest/auto_deploy/_utils_test/_sharding_ir_helpers.py).

The test workflow per rank:
1. Instantiate a tiny (4-layer) IR model via `transformers.<Config>()` + the registered `modeling_*_ir.py` class. No factory, no filesystem.
2. Re-init weights with a CPU `torch.Generator` seeded identically across ranks → bit-identical unsharded weights everywhere.
3. Eager unsharded forward → reference logits.
4. Snapshot `state_dict()`.
5. `torch_export_to_gm` → `apply_sharding_hints` + `strip_sharding_hints` under `world_size=4, tp_size=2, ep_size=2` (exercises both TP and EP — including the last-EP-rank conditional branch in `MoEShardableNode.apply`).
6. `gm_sharded.load_state_dict(snapshot, strict=False)` fires the registered `_load_hook`s, slicing weights per rank.
7. Sharded forward + `torch.testing.assert_close` against the reference with dtype-aware tolerance.

## Files added

- `tests/unittest/auto_deploy/_utils_test/_sharding_ir_helpers.py` — model registry, programmatic config + model build, deterministic random init, dtype-aware tolerance.
- `tests/unittest/auto_deploy/multigpu/transformations/library/conftest.py` — `--sharding-ir-model-id` pytest option + `sharding_ir_model_id` fixture (skips when absent).
- `tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_ir_equivalence.py` — the test, plus a module-level `sys.path` scrub that removes any uncompiled `tensorrt_llm/` tree (no-op in single-tree CI; matters only for worktree-based dev setups).

## Test plan

- [x] Test passes locally on 4×H100 for `Qwen/Qwen3-8B`: `pytest auto_deploy/multigpu/transformations/library/test_sharding_ir_equivalence.py --sharding-ir-model-id Qwen/Qwen3-8B` (~28s).
- [x] Test is skipped (not failed) when `--sharding-ir-model-id` is absent.
- [x] CI: green on the targeted AutoDeploy stages (triggered after PR opens).

Refs: #13960

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test infrastructure for validating sharding equivalence across multi-GPU distributed configurations, ensuring numerical correctness between sharded and eager model execution.
  * Added pytest configuration and fixtures to support parameterized IR model testing with environment variable control.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13963)


greg-kwasniewski1:

greg-kwasniewski1:

tests/unittest/auto_deploy/utils/sharding_ir_helpers.py
```diff
@@ -0,0 +1,396 @@
...
+def enable_ir_models() -> None:
+    """Set ``AD_USE_IR_MODELS=1`` so the custom-model registry registers IR variants.
+
+    Must be called before any import of
+    ``tensorrt_llm._torch.auto_deploy.models.custom`` so the env-var guarded
+    block in that package's ``__init__.py`` registers the IR variants. Safe to
+    call repeatedly. Once IR sharding becomes the default representation this
+    env-var gate will go away and the call becomes a no-op.
+    """
+    os.environ["AD_USE_IR_MODELS"] = "1"
```
nvchenghaoz: Not a blocker for this PR, but I feel like we should not use the env var "AD_USE_IR_MODELS" in general. If there is an alternative way to solve this without the env var, could you file a ticket to delete this env var?

nvchenghaoz (approved):
Looks good, just one nit: 

Can we rename the `_utils_test to utils`? And rename the `_sharding_ir_helpers.py` to `sharding_ir_helpers.py`?

greg-kwasniewski1:

tests/unittest/auto_deploy/utils/sharding_ir_helpers.py
greg-kwasniewski1: I suggest doing this after [this PR](https://github.com/NVIDIA/TensorRT-LLM/pull/13478) is merged. From then onward, I suggest not keeping the "dual" models `modeling_*.py` and `modeling_*_ir.py`, but enforce that all modeling codes should use sharding IR. But until we have this duality, we need a way to switch between them.

greg-kwasniewski1:

greg-kwasniewski1:

greg-kwasniewski1:
### PR scope cleanup (force-pushed, 6 commits)

This PR was opened before #13478 merged. Now that #13478 has made sharding-IR the canonical modeling path (no more `modeling_*_ir.py` / `AD_USE_IR_MODELS` gate), I've stripped the PR back to its minimum and refined the test's call convention so it doesn't depend on per-modeling-file `forward()` slot ordering.

#### Final scope: 4 files, +1038 / -3

| File | Status |
|---|---|
| `tests/unittest/auto_deploy/_utils_test/_sharding_ir_helpers.py` | new (+476) |
| `tests/unittest/auto_deploy/multigpu/transformations/library/conftest.py` | new (+58) |
| `tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_ir_equivalence.py` | new (+455) |
| `.claude/skills/ad-sharding-ir-port/SKILL.md` | +49 / -3 |

No core modeling code, no test-list churn, no waives.txt edits, no unrelated reverts.

#### What was dropped vs. the pre-cleanup branch

- All references to the `AD_USE_IR_MODELS` env var (now a no-op post-#13478) and the `enable_ir_models()` helper that toggled it.
- Modifications to canonical `modeling_*.py` files (`modeling_nemotron_h.py`, `modeling_qwen3_5_moe.py`) that re-ordered `forward()` parameters. Replaced by switching the test itself to kwarg-based export (commit `b298d70a`), which mirrors the production export path (`transform/library/export_to_gm.py::ExportToGM` already uses `args=()` + kwargs).
- The unused `modeling_llama3_ir.py` edits (file is dead code -- not registered, not imported).
- An accidental revert of #14392 (K2.5 multimodal dep8 fix for NVBug 6182617) and the related test-list / waives churn -- that fix had to stay.
- Three `upstream/main` merge commits absorbed by a rebase-equivalent reset.

#### Test does not touch modeling files, even for DeepSeek-V3

`modeling_deepseek.py` has two latent assumptions on `DeepseekV3Config` -- `rope_scaling["factor"]` is always present, and `moe_layer_freq` is always set -- that hold for every real DeepSeek-V3 checkpoint (yarn rope_scaling with `factor`, explicit `moe_layer_freq`) but not for a default-constructed `DeepseekV3Config()`. Per same scope-discipline as the `forward()`-ordering case, the right answer is to provide a realistic tiny config in the test helper (`_TINY_KWARGS_UNIVERSAL`) rather than mutate the canonical modeling file. See commit `d6a768e0`.

#### Validation

| Model | tp-only | ep-only | tep | attn-dp |
|---|---|---|---|---|
| qwen3 | pass | pass | pass | pass |
| nemotron_h | pass | pass | pass | pass |
| qwen3_5_moe | pass | pass | pass | pass |
| deepseek | (sweep re-running) | (sweep re-running) | **pass** | (sweep re-running) |

deepseek/tep (the most general 4-rank TP+MoE-EP grid, default in conftest) confirmed locally with `rel_rmse < tol`. Full 4x4 re-sweep is running in parallel with CI; an updated table follows as a second comment.

#### Skill update

`.claude/skills/ad-sharding-ir-port/SKILL.md` now mandates running this equivalence test for all dist configs as Step 11b of the porting workflow, with pass criterion `rel_rmse < tol` (both parsed from the test's own log line -- no hardcoded tolerance in the skill, so `REL_RMSE_TOL` / `SHARDING_IR_REL_RMSE_TOL` changes don't rot the skill). It also drops hardcoded `world_size: 8` / `world_size=2 and 8` literals in favor of "auto-detect max-available GPUs and run once".

#### Default-CI cost

The test is gated on `--sharding-ir-modeling-file`; the conftest skips it when absent. A default `pytest tests/unittest/` invocation pays only the ~1s `skipped` cost, not the ~25-30s/cell of an actual run.

#### Final commit list

```
d6a768e09e [TRTLLM-13960][test][autodeploy] sharding-IR equivalence: tiny config covers DeepSeek-V3
2de4a1321d [TRTLLM-13960][chore][autodeploy] ad-sharding-ir-port skill: tighten max-available world_size checklist item
e0c4f6b2ae [TRTLLM-13960][chore][autodeploy] ad-sharding-ir-port skill: drop hardcoded world_size literals
b298d70a10 [TRTLLM-13960][test][autodeploy] sharding-IR equivalence: export with kwargs, not positional
540567f443 [TRTLLM-13960][chore][autodeploy] ad-sharding-ir-port skill: mandate sharding equivalence test
96bde08d03 [TRTLLM-13960][test][autodeploy] Offline equivalence test for sharding IR
```

greg-kwasniewski1:

greg-kwasniewski1:
### Follow-up: scrub removal + corrective fix for the DeepSeek-V3 patch

Two new commits land on top of the cleanup branch:

**1. [`38b5dc873e`](https://github.com/NVIDIA/TensorRT-LLM/pull/13963/commits/38b5dc873e) — drop dev-only `sys.path` scrub from the test file.**
`_scrub_unbuilt_tensorrt_llm_from_sys_path` was a local convenience for running this test out of a git worktree separate from the build root. CI and the typical contributor workflow use a single tree where the editable-install `tensorrt_llm` *is* the source tree, so the scrub is a no-op for everyone except the dev who needed multi-worktree juggling. Dropping it also lets `pytest` / `torch` be normal top-of-module imports instead of `# noqa: E402` after a runtime `sys.path` mutation — only the local `_sharding_ir_helpers` import keeps the `# noqa: E402` because it genuinely depends on the `_HELPERS_DIR` insert.

**2. [`fc3669ad52`](https://github.com/NVIDIA/TensorRT-LLM/pull/13963/commits/fc3669ad52) — switch the DeepSeek-V3 patch from a universal kwarg to a per-family quirk.**
The previous helper commit ([`d6a768e0`](https://github.com/NVIDIA/TensorRT-LLM/pull/13963/commits/d6a768e09e)) set `rope_scaling = None` in the universal tiny-kwargs dict to clear `DeepSeekV3Attention.__init__`'s unconditional `config.rope_scaling["factor"]` lookup. That turns out to be unsafe: in transformers 5.x, `PretrainedConfig.__setattr__` aliases `rope_scaling` ↔ `rope_parameters`, so the universal `setattr(config, "rope_scaling", None)` *also* blanks `config.rope_parameters`, which `Qwen3_5MoeTextRotaryEmbedding` reads at construction time. A 16-cell local sweep caught the regression: qwen3_5_moe failed 4/4 with `TypeError: 'NoneType' object is not subscriptable` at `rope_params["rope_theta"]`.

The corrective fix moves the DeepSeek-V3 patch out of the universal kwargs and into a new `_apply_per_family_quirks` hook gated on `kv_lora_rank` (a marker that exists natively on `DeepseekV3Config` and on no other config in the current matrix). It sets `config.rope_scaling = {"rope_type": "default", "factor": 1.0}`: `factor` clears the unconditional DeepSeek-V3 lookup, `rope_type = "default"` routes `_init_rope` through the vanilla rotary branch (correct stimulus for a sharding equivalence test), and nothing else in the matrix exposes `kv_lora_rank` natively so the patch never fires elsewhere.

To keep that feature-presence dispatch trustworthy, `build_ir_model` now runs `_apply_per_family_quirks` **before** the universal-kwargs `setattr` loop. The universal pass intentionally writes many family-specific keys (`kv_lora_rank`, `q_lora_rank`, `ssm_state_size`, …) onto every config to keep one tiny-kwargs dict universal, which would otherwise make every config look like every family by the time quirks run.

**Validation (16/16):**

| model         | tp-only | ep-only | tep | attn-dp |
|---------------|---------|---------|-----|---------|
| qwen3         | ✓       | ✓       | ✓   | ✓       |
| nemotron_h    | ✓       | ✓       | ✓   | ✓       |
| qwen3_5_moe   | ✓       | ✓       | ✓   | ✓       |
| deepseek      | ✓       | ✓       | ✓   | ✓       |

Modeling code untouched. PR scope still = 3 new test files + 1 skill update.

greg-kwasniewski1:

greg-kwasniewski1:

greg-kwasniewski1:

greg-kwasniewski1:
/bot skip --comment "All 4 failures are in unrelated PyTorch-backend tests (TestDeepSeekV3Lite::test_nvfp4_batch_waiting, test_overlap_scheduler[Ray] x2, TestLagunaXS::test_bf16). This PR's diff is test-infrastructure only under tests/unittest/auto_deploy/multigpu/; no production paths touched. All 5 AutoDeploy CI stages green (5627 passed, 0 failed)."

greg-kwasniewski1:
/bot skip --comment "All 4 failures are in unrelated PyTorch-backend tests (TestDeepSeekV3Lite test_nvfp4_batch_waiting, test_overlap_scheduler Ray, TestLagunaXS test_bf16). This PRs diff is test-infrastructure only under tests_unittest_auto_deploy_multigpu, no production paths touched. All 5 AutoDeploy CI stages green (5627 passed, 0 failed)."

PR #13944, opened 2026-05-09
[None][feat] VisualGen: Attention2D + Ulysses & Multi-GPU LPIPS Evals

juney-nvidia:
## Release Notes

**New Features**
* Attention2D now supports combination with Ulysses head parallelism for improved sequence sharding.
* Added ring-based column parallelism with overlapped execution mode.
* Introduced new `RingAttention` implementation for distributed attention computation.

**Improvements**
* Enhanced device mesh configuration for unified CFG/TP/CP/Ulysses parallelism.
* Improved parallel configuration logging to report all parallelism modes.
* Expanded test coverage for combined Attention2D and Ulysses configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13944?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

To see a list of available CI bot commands, please comment `/bot help`.

PR #13978, opened 2026-05-11
[TRTLLM-11457][feat] Async Ulysses pipeline (Enabled for LTX-2 + WAN)

luyiyun1021:
@coderabbitai summary

## Description

Adds an opt-in **async Ulysses pipeline** for sequence parallelism on diffusion video models (LTX-2 + WAN 2.2). The pipeline overlaps per-rank V/Q/K compute (`GEMM → RMSNorm → RoPE`) with cross-rank data movement (PyTorch `_SymmetricMemory` CUDA-IPC peer access + symm-mem `barrier` release fence) on a single dedicated per-device side stream. SMs stay free for the next V/Q/K while the GPU **Copy Engine** drains the prior V/Q/K's peer push.

**Opt-in gate**: `parallel_config.async_ulysses` (default **False**). Both LTX-2 (`LTX2Attention.__init__(async_ulysses=...)`) and WAN 2.2 self-attn read the same gate. Non-Ulysses, audio, cross-attn, and `ulysses_size == 1` paths are unchanged.

**Requires PyTorch ≥ 2.5** for `torch._C._distributed_c10d._SymmetricMemory.empty_strided_p2p` / `rendezvous` / `barrier`. No NCCL device-API dependency.

### Enabling

YAML config (`trtllm-serve --config foo.yaml`):

```yaml
parallel_config:
  ulysses_size: 4         # any P > 1
  async_ulysses: true     # default False — flip to enable the async pipeline
```

Python API (`VisualGenArgs`):

```python
from tensorrt_llm._torch.visual_gen.config import VisualGenArgs, VisualGenParallelConfig

args = VisualGenArgs(
    parallel_config=VisualGenParallelConfig(
        ulysses_size=4,
        async_ulysses=True,
    ),
    ...
)
```

A/B comparison against the baseline blocking-A2A path needs zero code change — flip the bool.

### Stream lanes per attention call

```
default      [V_GEMM+norm+rope+permute]   [Q_GEMM+norm+rope+permute]   [K_GEMM+norm+rope+permute]                            [SDPA]
                      │ev.record                  │ev.record                  │ev.record
                      ▼                           ▼                           ▼                                                ▲
side_stream      [V_push]                    [Q_push]                    [K_push]   [barrier × 3] ─► ev_done.record ──────────┘
                                                                                                     default waits ev_done
```

Issue order V → Q → K. V's CE push overlaps with Q's compute; Q's push overlaps with K's compute. Three barriers fire only at `_join_async`, after which the default stream waits the tail event before SDPA. Side stream is a per-device singleton (`UlyssesAttention._side_stream_by_device`) so all transformer layers on the same device share one comm lane — avoids per-layer stream proliferation under cuda_graph capture / inductor.

### C++ ops surface

```c++
// Phase 1 — default stream: lazy-alloc slot ring + fused permute+scatter kernel.
std::tuple<Tensor, SendHandle> ulysses_a2a_async_prepare(Tensor input, ProcessGroup pg);

// Phase 2a — side stream: CE peer push only (no barrier).
void ulysses_a2a_async_push(SendHandle send_h, ProcessGroup pg);

// Phase 2b — side stream: symm-mem barrier release fence (channel=0).
//   Called N times at join, once per deferred push.
void ulysses_a2a_async_barrier(ProcessGroup pg);
```

`AsyncUlyssesOp` caches a canonical `SymmetricMemory` handle on the first slot allocation (`mCanonicalHandle`); `emitBarrier()` uses it directly instead of scanning slots each call.

### `forward_async` API (closure-based)

```python
@torch.compiler.disable(recursive=False)
def _issue_async(self, perm_4d):
    recv, send_h = torch.ops.trtllm.ulysses_a2a_async_prepare(perm_4d, self._pg_boxed)
    ev = torch.cuda.Event(); ev.record()
    with torch.cuda.stream(self._async_side_stream):
        ev.wait()
        torch.ops.trtllm.ulysses_a2a_async_push(send_h, self._pg_boxed)
    self._pending_barriers += 1
    return recv

@torch.compiler.disable(recursive=False)
def _join_async(self):
    with torch.cuda.stream(self._async_side_stream):
        for _ in range(self._pending_barriers):
            torch.ops.trtllm.ulysses_a2a_async_barrier(self._pg_boxed)
        ev_done = torch.cuda.Event(); ev_done.record()
    self._pending_barriers = 0
    torch.cuda.current_stream().wait_event(ev_done)

def forward_async(self, compute_q, compute_k, compute_v, **attn_kwargs):
    v_4d = compute_v(); v_5d = self._issue_async(v_4d)
    q_4d = compute_q(); q_5d = self._issue_async(q_4d)
    k_4d = compute_k(); k_5d = self._issue_async(k_4d)
    self._join_async()
    # ulyssesPostUnscatterKernel ▶ SDPA ▶ reverse a2a — all in the outer compile region.
```

`@torch.compiler.disable(recursive=False)` on `_issue_async` / `_join_async` is the only stream-switch boundary. The caller's `compute_q/k/v` closures execute on the default stream between successive `_issue_async` graph breaks, so their bodies sit inside the outer block's inductor graph and get full `GEMM+norm+RoPE` fusion (verified `nvjet_sm100_ootst_..._Avec16UE4M3_Bvec16UE4M3_TNT` kernels emitted under NVFP4).

### `_SymmetricMemory` slot ring

`AsyncUlyssesOp` owns a 3-slot ring (V/Q/K per layer — intra-layer slot-reuse hazard requires ≥ 3, cross-layer reuse is safe because `_join_async` syncs the side stream before next layer's V starts). First call at `(numel, dtype)` triggers `empty_strided_p2p + rendezvous`; subsequent calls reuse the cached `(handle, peer_ptrs)`. Ops cached by `pg->getGroupName()` so multiple PGs (e.g. ulysses subgroup × cfg subgroup) coexist cleanly. Allocation is **not graph-capture-safe** (rendezvous / cudaMalloc inside) — `forward_async` requires one out-of-capture warmup pass before cuda_graph capture starts.

## Key Optimizations

The four mechanisms below are what make this PR's async path actually faster than the baseline blocking A2A. Listed bottom-up from hardware-level resource choice to host-side launch reduction.

### 1. Copy-Engine peer push (avoid SM contention)

V/Q/K peer-to-peer data movement runs entirely on the GPU **Copy Engines**, never on SM-resident NCCL kernels:
- Eager path: one `cudaMemcpyBatchAsync` over the `P-1` per-peer chunks (driver fans out across CEs).
- Capture path: per-peer `cudaMemcpyAsync` loop (because `cudaMemcpyBatchAsync` isn't graph-capture-safe until CUDA 13.5 — see Follow-ups #1).

The Copy Engine handles all `P-1` peer pushes, leaving SMs **fully free** for the next V/Q/K `GEMM → RMSNorm → RoPE`. Compared to NCCL all-to-all (Send/Recv kernels on SMs competing for the same compute resource as the GEMMs), the CE-based push has **disjoint hardware resources** from compute — no channel multiplexing or ring-topology hacks needed.

### 2. Fuse permute with self-chunk scatter (bypass CE local-D2D bandwidth cap)

`ulyssesPermuteScatterKernel` does two things in one launch:
- Permute `[B, S, H, D] → [P, B, S, H/P, D]`
- Write the **local rank's self chunk** directly to the recv slot's self position (`slot.basePtr + my_rank * chunk_bytes`); only the `P-1` peer chunks go to `slot.sendBuf`.

Without the fuse, the alternative is: SM permute → sendBuf; then CE runs `P` memcpys (`P-1` peer pushes + 1 local sendBuf → recv self-slot D2D copy). The CE's **local D2D channel is hardware-capped at ~100 GB/s** (independent from the NVLink/NVSwitch peer channel at ~400+ GB/s — they don't share bandwidth, but local D2D's own ceiling is low). Since the self chunk is the same size as one peer chunk, the local copy would **become the critical path** of the entire a2a — the `P-1` peer pushes finish 4× faster while we wait for one local D2D.

Fusing moves the self-chunk write into the SM kernel, using **HBM bandwidth (~5–10 TB/s)** — one to two orders of magnitude above CE local-D2D. The alternative of green-context isolation for the local copy (separate CE queue) requires complex multi-stream/queue coordination; fusing bypasses that whole subsystem.

### 3. Deferred barriers (push×N FIFO, barrier×N drain at join)

V/Q/K's a2a is split into two C++ ops, `ulysses_a2a_async_push` (CE push only) and `ulysses_a2a_async_barrier` (symm-mem barrier release fence). Side-stream sequence:

```
[push_V, push_Q, push_K, barrier, barrier, barrier]
```

Originally the op was push+barrier fused, producing `[push, barrier, push, barrier, push, barrier]` where each barrier kernel cut the CE FIFO and forced the next push to wait. Each barrier kernel drops from **~164 µs → ~93 µs (-43%)** under the deferred scheme, and per-step CE PTOP throughput improves by **~4 ms** (WAN 8-GPU step capture). Python state tracks `_pending_barriers` on `UlyssesAttention`; `_join_async` drains exactly N barriers on the side stream before the default stream waits the tail event.

### 4. Dedup `fp4_quantize` across QKV self-attn

Under async ulysses the video self-attn is forced to `SEPARATE_QKV` (so V/Q/K can stream-pipeline independently), but with NVFP4 static quant the three Linears all share `input_scale`:

```python
# Verified bit-equal across WAN 36+36 / LTX-2 42 self-attn layers (modelopt invariant)
to_q.input_scale == to_k.input_scale == to_v.input_scale
```

Pre-quantize hidden_states once and share the resulting `Fp4QuantizedTensor` across all three Linears via the existing fp4-shortcut path in `NVFP4LinearMethod._input_prepare` (which just unpacks `.fp4_tensor` / `.scaling_factor` for GEMM — no internal swizzle so the SF layout is the same as Linear's own quantize path). Eliminates **2 of 3 quantize launches per self-attn block** (verified **−368 launches in WAN A14B 8-GPU step capture**; identical kernel count between async-OFF FUSE_QKV and async-ON SEPARATE_QKV+dedup).

Gating:
- Structural (`_maybe_share_qkv_quantize`, set in `__init__`): `SEPARATE_QKV` + NVFP4 + not `force_dynamic_quantization`.
- Runtime: `getattr(to_q, "input_scale", None) is not None` — guards against checkpoints that exclude individual Linears from NVFP4 (e.g. LTX-2 `transformer_blocks.10.attn1`, bf16-only — falls back to per-Linear quantize).

### Isolated all-to-all (B200, single rank)

`bench_alltoall_ab.py` — single-tensor a2a latency (ms), B=2, S_total=6144, H=32, D=128, warmup=30, bench=100. `naive_fused` = `dist.all_to_all_single` on 5D fused QKV; `naive_split` = 3 separate 4D a2a; `symm_ce` = the new symm-mem CE pipeline (per V/Q/K call).

| P | mode  | naive_fused      | naive_split      | symm_ce               |
|---|-------|------------------|------------------|-----------------------|
| 2 | eager | 0.805 (1.00×)    | 0.872 (0.92×)    | **0.641 (1.26×)**     |
| 2 | graph | 0.862 (1.00×)    | 0.889 (0.97×)    | **0.685 (1.26×)**     |
| 4 | eager | 0.419 (1.00×)    | 0.483 (0.87×)    | **0.339 (1.24×)**     |
| 4 | graph | 0.431 (1.00×)    | 0.496 (0.87×)    | **0.370 (1.16×)**     |
| 8 | eager | 0.251 (1.00×)    | 0.294 (0.85×)    | **0.194 (1.29×)**     |
| 8 | graph | 0.262 (1.00×)    | 0.309 (0.85×)    | 0.287 (0.91×)         |

`symm_ce` is 1.16–1.29× faster than the fused baseline in every cell except `P=8 graph` (where cuda_graph capture amortizes Python launch overhead in the `naive_fused` 1-collective path enough that the merged baseline wins).

### End-to-end on B200 (production scale)

| model | ws | uly | cfg | OFF (s) | ON (s) | saved | E2E |
|---|---|---|---|---|---|---|---|
| LTX-2 (768×1280, 121 f, 40 steps, NVFP4, `torch.compile`+`cuda_graph`) | 2 | 2 | 1 | 21.453 | 20.767 | 0.686 | **−3.20%** |
|  | 4 | 2 | 2 | 13.154 | 12.884 | 0.270 | **−2.05%** |
|  | 8 | 4 | 2 | 9.486 | 9.338 | 0.148 | **−1.56%** |
| WAN 2.2 T2V-A14B (720×1280, 81 f, 40 steps, NVFP4 static, `torch.compile` only) | 2 | 2 | 1 | 206.731 | 198.311 | 8.420 | **−4.07%** |
|  | 4 | 2 | 2 | 103.031 | 100.504 | 2.527 | **−2.45%** |
|  | 8 | 4 | 2 | 55.444 | 54.114 | 1.330 | **−2.40%** |

3 timed E2E runs/cell, median reported. Per-cell range is well below the OFF→ON delta in every case (non-overlap). WAN's larger relative gain reflects more compute-per-A2A and a longer overlap window per layer (no cuda_graph overhead to amortize the baseline path).

## Files

| File | Role |
|------|------|
| `cpp/tensorrt_llm/kernels/ulyssesPermuteScatterKernel.{h,cu}` (new) | Fused pre-A2A permute + self-chunk-direct-write kernel. |
| `cpp/tensorrt_llm/kernels/ulyssesPostUnscatterKernel.{h,cu}` (new) | Fused post-A2A unscatter (5D → 4D, joint Q/K/V); NHD-stride storage, HND via transpose-view. |
| `cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp` (new) | `SendHandle` + `AsyncUlyssesOp` (symm-mem slot ring + `mCanonicalHandle` cache + CE push + barrier). Ops: `ulysses_a2a_async_prepare`, `ulysses_a2a_async_push`, `ulysses_a2a_async_barrier`. |
| `cpp/tensorrt_llm/thop/ulyssesPermuteScatterOp.cpp` (new) | Thop wrapper for permute+scatter kernel (unit test entry). |
| `cpp/tensorrt_llm/thop/ulyssesPostUnscatterOp.cpp` (new) | Thop wrapper for post-unscatter kernel. |
| `cpp/tensorrt_llm/thop/CMakeLists.txt` | Register new thop files. |
| `tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py` | `register_fake` for `trtllm::ulysses_post_unscatter_qkv`. |
| `tensorrt_llm/_torch/visual_gen/config.py` | New `async_ulysses` Pydantic field (default False). |
| `tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py` | `UlyssesAttention(async_pipeline=...)`, side-stream singleton, `_issue_async`/`_join_async` deferred-barrier pipeline, V/Q/K rolling `forward_async`. |
| `tensorrt_llm/_torch/visual_gen/modules/attention.py` | Plumb `async_ulysses` flag; **QKV `fp4_quantize` dedup in `Attention.forward_async`** (shared `Fp4QuantizedTensor` across to_q/to_k/to_v). |
| `tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py` | LTX-2 self-attn opts into async path; **mirrored QKV dedup in `LTX2Attention.forward_async`** override. |
| `tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/utils_ltx2.py` | Sigma-sync fix for cuda_graph compat. |
| `tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py` | WAN self-attn (`attn1`) reads the gate; SEPARATE_QKV when on. |
| `tests/unittest/_torch/thop/parallel_hw_agnostic/test_ulysses_permute_scatter.py` (new) | Unit test for permute+scatter kernel. |
| `tests/unittest/_torch/thop/parallel_hw_agnostic/test_ulysses_post_unscatter.py` (new) | Unit test for post-unscatter kernel. |
| `tests/unittest/_torch/visual_gen/multi_gpu/test_ulysses_async.py` (new) | Multi-rank ws=2 pipeline tests (slot-ring wraparound, cuda_graph capture, multi-PG). |

## Follow-ups

1. **`cudaMemcpyBatchAsync` under CUDA Graph capture (CUDA 13.5).** Today the capture branch falls back to a `P-1`-iteration per-peer `cudaMemcpyAsync` loop because `cudaMemcpyBatchAsync` is not graph-capture-safe. Per [NVBugs 5853376](https://nvbugspro.nvidia.com/bug/5853376) and [NVBugs 5760690](https://nvbugspro.nvidia.com/bug/5760690), targeted at **CUDA 13.5 (Q4 2026)**. We tried `cudaGraphAddMemcpyNode1D` with sibling-DAG independence as a stop-gap; nsys confirmed runtime serializes the sibling memcpys on the same HW queue (~4 µs gaps, no CE fanout), so the per-peer loop wins today.

2. **Collapse N barriers at join → 1 barrier.** Current `_join_async` drains N symm-mem barriers in a loop (one per deferred push) — kept 1:1 with the original protocol for risk-free refactor. The barriers are semantically identical (all channel-0 PG-level fences); a single barrier after `[push_V, push_Q, push_K]` should sync the recv buffers just as well (side-stream FIFO + cross-rank barrier covers all prior pushes). Saves 2 of 3 `barrier_kernel` per layer (~186 µs × N_layers per step on B200). Requires verifying that all ranks symmetrically issue 1 barrier per join (slot rotation is independent of barrier count — `nextSlotIdx` ticks in `_prepare`, not `_barrier`).

3. **Drop `@torch.compiler.disable`, native multi-stream `torch.compile`.** Stream-switch boundaries in `_issue_async`/`_join_async` are guarded by `@torch.compiler.disable(recursive=False)` because inductor doesn't yet model `with cuda.stream(...)` + `event.record/wait_event` as first-class graph ops (4 graph breaks per layer × N layers). Inline once `torch.compile`'s multi-stream support matures.

4. **Fuse pre-A2A permute into the RMSNorm+RoPE kernel.** We already own the fused RMSNorm+RoPE kernel (see `fuse_qk_norm_rope=True` path). Passing the slot's `send_buf` pointer + permute index map directly into the kernel epilogue lets it write the permuted layout straight into the symm-mem slot, eliminating the separate `ulyssesPermuteScatterKernel` launch on the critical path. Leaves the post-unscatter kernel untouched.

## Test Coverage

- **Kernel unit tests** (single-GPU) under `tests/unittest/_torch/thop/parallel_hw_agnostic/`: `test_ulysses_permute_scatter.py` and `test_ulysses_post_unscatter.py` validate the two new fused CUDA kernels against torch references in isolation.
- **Multi-rank pipeline test** (ws=2) under `tests/unittest/_torch/visual_gen/multi_gpu/test_ulysses_async.py`: three `torch.multiprocessing.spawn` tests covering the production `prepare/push/barrier` path end-to-end —
  - `test_slot_ring_wraparound`: eager loop > `kNumSlots*2` iterations, byte-exact vs `all_to_all_4d`. Catches off-by-one slot-reuse bugs.
  - `test_capture_smoke`: warm slot out-of-capture, then `torch.cuda.CUDAGraph` replay 8× with fresh inputs, byte-exact vs `all_to_all_4d`. Exercises the per-peer `cudaMemcpyAsync` path used under production cuda_graph.
  - `test_multi_pg`: two distinct PGs spanning the same ranks, alternate calls, byte-exact vs each PG's `all_to_all_4d`. Exercises PG-name caching in `getOrCreateOp` and `set_group_info` re-registration across groups.
- **Recommended CI bot stages**: `DGX_B200-4_GPUs-PyTorch-1`, `DGX_B200-8_GPUs-PyTorch-1` (covers Ulysses ws=4 and ws=8 in both LTX-2 and WAN end-to-end).

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/utils_ltx2.py
```diff
@@ -15,9 +15,11 @@ def to_velocity(
 
     velocity = (sample - denoised) / sigma
     """
+    # Tensor sigma: keep on device. `.item()` would force a D2H sync that
+    # deadlocks under nsys profiling combined with CUDA graph replay.
     if isinstance(sigma, torch.Tensor):
-        sigma = sigma.to(calc_dtype).item()
-    if sigma == 0:
+        sigma = sigma.to(calc_dtype)
+    elif sigma == 0:
```
chang-l: **[BLOCKING] Silent zero-check regression for tensor `sigma`.**

Before:
```python
if isinstance(sigma, torch.Tensor):
    sigma = sigma.to(calc_dtype).item()
if sigma == 0:
    raise ValueError("Sigma can't be 0.0")
```
After (this PR):
```python
if isinstance(sigma, torch.Tensor):
    sigma = sigma.to(calc_dtype)
elif sigma == 0:
    raise ValueError("Sigma can't be 0.0")
```
The `if → elif` short-circuits the zero check whenever `sigma` is a Tensor — previously `.item()` materialized the scalar and the `if sigma == 0` caught it; now a 0.0-valued tensor flows straight into the division on the next line, producing NaN/Inf silently.

Either (a) keep the check tensor-aware:
```python
if isinstance(sigma, torch.Tensor):
    sigma = sigma.to(calc_dtype)
    # NB: cannot check sigma == 0 here without re-syncing; rely on upstream guarantee.
elif sigma == 0:
    raise ValueError("Sigma can't be 0.0")
```
and document the upstream guarantee, or (b) use a sync-free guard like `torch.where(sigma == 0, ...)` if NaNs are unacceptable. Just leaving it as `elif` is a silent behavioral change.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+#if ENABLE_MULTI_DEVICE
+
+namespace
+{
+
+class AsyncUlyssesOp
+{
+public:
+    static constexpr int kNumSlots = 4;
```
chang-l: **[HIGH] Justify `kNumSlots = 4` and add a wrap-around test.**

3 `_issue_async` calls per `forward_async` × `kNumSlots = 4` ⇒ slot 0 is reused after ~1.33 attention calls. Walking the hazards:

- *Default-stream RAW*: prev SDPA reads slot 0 → next layer's permute writes slot 0 on the **same** stream. Safe (FIFO).
- *Side-stream peer push into local slot 0*: gated by an event from default recorded **after** the new permute. Safe.
- *Cross-rank fence*: the previous round's `h->barrier()` on slot 0 has to retire before the next round's CE push lands. This is guaranteed by FIFO on the side stream (barrier from round N is queued before CE push from round N+1 because issues 0..N+1 are all on the same `_async_side_stream`). Safe — but **only because of that FIFO discipline.**

The correctness argument depends on (a) a single side stream and (b) the implicit assumption that no SDPA on slot 0 from round N can still be in flight when slot 0 is rebound in round N+1 — which holds because both run on the default stream sequentially.

Please either:
1. Add a comment block above this constant capturing the FIFO/event argument (so the next person who tries to add a second side stream or break the FIFO doesn't silently break it), and/or
2. Add a multi-rank pytest that loops `>= kNumSlots * 2` `_prepare/_async` calls and asserts byte-exact match vs `torch.distributed.all_to_all_single`. This catches off-by-one slot reuse bugs that the kernel-only tests cannot.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
```diff
@@ -0,0 +1,450 @@
...
+                void* peerBase = reinterpret_cast<void*>(peer_recv_ptrs[p]);
+                dsts.push_back(
+                    static_cast<char*>(peerBase) + static_cast<size_t>(pgRank) * static_cast<size_t>(slot_bytes));
+                srcs.push_back(sendBase + static_cast<size_t>(p) * static_cast<size_t>(slot_bytes));
+                sizes.push_back(static_cast<size_t>(slot_bytes));
+            }
+            cudaMemcpyAttributes attrs[1];
+            std::memset(&attrs[0], 0, sizeof(attrs[0]));
+            attrs[0].srcAccessOrder = cudaMemcpySrcAccessOrderStream;
+            attrs[0].flags = 1u;
```
chang-l: **[LOW] Magic number + missing CUDA version guard.**

`attrs[0].flags = 1u` is opaque — please use the named constant from `<cuda_runtime.h>` (looks like `CU_MEMCPY_FLAG_DEFAULT` / equivalent runtime API enum), with a one-line comment on what flag it sets.

Also: `cudaMemcpyBatchAsync` was added in CUDA 12.8 (the cudart headers list it as `_v12080`). If TRT-LLM's minimum supported CUDA toolkit is older anywhere in the build matrix, this won't link there. Either:
- Document that the `asyncUlyssesOp` target requires CUDA >= 12.8 in CMakeLists.txt, or
- Wrap the batch path in `#if CUDART_VERSION >= 12080` with the per-peer `cudaMemcpyAsync` loop as the lower-version fallback (you already have that fallback for capture).

Most B200/Blackwell builds are on 13.x so this is likely fine in practice, but worth being explicit about the dependency.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
```diff
@@ -0,0 +1,450 @@
...
+            {
+                if (s.handle)
+                {
+                    h = s.handle;
+                    break;
+                }
+            }
+        }
+        TLLM_CHECK_WITH_INFO(h, "emitBarrier: no slot allocated yet — _prepare must precede the first _async barrier.");
+        h->barrier(/*channel=*/0, /*timeout_ms=*/10000);
```
chang-l: **[MEDIUM] PR description ↔ implementation mismatch on the barrier channel.**

The PR description says:
> `channel` is per slot, letting concurrent V/Q/K issues use independent channels without interfering.

But the code unconditionally uses `channel=0`. Today this is **correct** because V/Q/K barriers are FIFO-serialized on the single `_async_side_stream`, so channel multiplexing isn't required. But the description sells a property the implementation doesn't have.

Pick one:
- (a) Update the description to describe the actual FIFO design ("single side stream, channel=0"), or
- (b) Plumb `channel = slotIdx % kNumChannels` here so the implementation actually delivers per-slot channel parallelism, which would let a future PR split V/Q/K to multiple side streams without re-architecting this layer.

Also: the 10 s timeout is generous; a comment on why this number was chosen (vs the symm-mem default) would help operators interpreting any future timeout failures.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
```diff
@@ -0,0 +1,450 @@
...
+            slot.peerPtrs.clear();
+        }
+
+        int currentDev = -1;
+        TLLM_CUDA_CHECK(cudaGetDevice(&currentDev));
+        c10::Device device(c10::DeviceType::CUDA, currentDev);
+        std::string const& groupName = mPg->getGroupName();
+        int const pSize = mPg->getSize();
+
+        slot.symm_tensor = c10d::symmetric_memory::empty_strided_p2p(
```
chang-l: **[HIGH] First-time slot allocation must run out-of-capture — please enforce or document.**

This path triggers a collective `empty_strided_p2p(...)` + `rendezvous(...)` + (further down) `cudaMalloc(...)` for `sendBuf`. None of these are stream-capture-safe. If the user's first `forward_async` call happens inside a CUDA graph capture region (e.g., warm-up was skipped or a new shape appears post-capture), this will fail in an unobvious way.

The comment above `cudaMalloc` says "Capture-safe because we eagerly alloc here (out of capture)" — but this only holds if the user warmed up before capture. That contract is not enforced at this layer.

Suggestions (pick one):
1. Add a `cudaStreamCaptureStatus` check at the top of `getOrAllocSlot` and `TORCH_CHECK(!underCapture, "async-ulysses: first-time slot allocation for shape ... must run out-of-capture; warm up the model before cuda_graph capture")`.
2. Document the warm-up requirement in (a) `UlyssesAttention.__init__(async_pipeline=True)` and (b) the `forward_async` docstring.

(1) is friendlier — the hard error tells the user exactly what to do.

tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/utils_ltx2.py
chang-l, line 24: **[BLOCKING] Silent zero-check regression for tensor `sigma`.**

Before:
```python
if isinstance(sigma, torch.Tensor):
    sigma = sigma.to(calc_dtype).item()
if sigma == 0:
    raise ValueError("Sigma can't be 0.0")
```
After (this PR):
```python
if isinstance(sigma, torch.Tensor):
    sigma = sigma.to(calc_dtype)
elif sigma == 0:
    raise ValueError("Sigma can't be 0.0")
```
The `if → elif` short-circuits the zero check whenever `sigma` is a Tensor — previously `.item()` materialized the scalar and the `if sigma == 0` caught it; now a 0.0-valued tensor flows straight into the division on the next line, producing NaN/Inf silently.

Either (a) keep the check tensor-aware:
```python
if isinstance(sigma, torch.Tensor):
    sigma = sigma.to(calc_dtype)
    # NB: cannot check sigma == 0 here without re-syncing; rely on upstream guarantee.
elif sigma == 0:
    raise ValueError("Sigma can't be 0.0")
```
and document the upstream guarantee, or (b) use a sync-free guard like `torch.where(sigma == 0, ...)` if NaNs are unacceptable. Just leaving it as `elif` is a silent behavioral change.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
chang-l: **[HIGH] Justify `kNumSlots = 4` and add a wrap-around test.**

(1) Please add a code comment above this constant explaining *why* 4 — what is the slot-reuse hazard analysis (default-stream RAW, side-stream peer push, cross-rank barrier fence) that makes 4 the minimum safe depth under the current single-side-stream FIFO design? A future contributor adding a second side stream or breaking FIFO needs to see this argument.

(2) Add a multi-rank pytest that loops `>= kNumSlots * 2` `_prepare`/`_async` calls and asserts byte-exact match against `torch.distributed.all_to_all_single`. This catches off-by-one slot-reuse bugs that the single-GPU kernel tests cannot.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
chang-l: **[MEDIUM] PR description ↔ implementation mismatch on the barrier channel.**

The PR description says:
> `channel` is per slot, letting concurrent V/Q/K issues use independent channels without interfering.

But the code unconditionally uses `channel=0`. Today this is **correct** because V/Q/K barriers are FIFO-serialized on the single `_async_side_stream`, so channel multiplexing isn't required. But the description sells a property the implementation doesn't have.

Pick one:
- (a) Update the description to describe the actual FIFO design ("single side stream, channel=0"), or
- (b) Plumb `channel = slotIdx % kNumChannels` here so the implementation actually delivers per-slot channel parallelism, which would let a future PR split V/Q/K to multiple side streams without re-architecting this layer.

Also: the 10 s timeout is generous; a comment on why this number was chosen (vs the symm-mem default) would help operators interpreting any future timeout failures.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
chang-l: **First-time slot allocation must run out-of-capture.**

  `empty_strided_p2p` + rendezvous + `cudaMalloc(sendBuf)` here are not capture-safe. The eager-alloc comment below assumes the user warmed up before
  capture, but nothing enforces it — a first call (or new shape) under capture will fail unobviously.

  Please add a `cudaStreamIsCapturing` check at the top of `getOrAllocSlot` and `TORCH_CHECK(!underCapture, "async-ulysses: warm up the model
  out-of-capture before cuda_graph capture")`.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -281,19 +280,28 @@ def __init__(
...
+        _qkv_mode_self = QKVMode.SEPARATE_QKV if self._use_async_ulysses else QKVMode.FUSE_QKV
         self.attn1 = Attention(
             hidden_size=hidden_size,
             num_attention_heads=num_heads,
             head_dim=head_dim,
-            qkv_mode=QKVMode.FUSE_QKV,
+            qkv_mode=_qkv_mode_self,
             qk_norm=True,
             eps=eps,
             fuse_qk_norm_rope=True,
```
chang-l: **[MEDIUM] Async path bypasses `fuse_qk_norm_rope=True` — please leave a TODO.**

When `_use_async_ulysses = True`, `qkv_mode` is forced to `SEPARATE_QKV` above, so the carefully-tuned fused QK-Norm+RoPE CUDA kernel is no longer reachable for self-attention. The `compute_q/compute_k` closures around L417/L430 now run `to_q → norm_q → view → apply_rotary_emb` as 4 separate ops. The overlap win still beats the loss in the benchmarks, but you're leaving perf on the table.

This matches PR-description Follow-up #3 ("fuse pre-A2A permute into the fused RMSNorm+RoPE kernel"). Please leave a TODO comment near `fuse_qk_norm_rope=True` on this line, e.g.:
```python
            # NOTE: fuse_qk_norm_rope is currently inert when _use_async_ulysses is
            # set (qkv_mode forced to SEPARATE_QKV). See PR follow-up #3 for the
            # plan to feed the fused norm+rope kernel directly into the symm-mem slot.
            fuse_qk_norm_rope=True,
```
so future readers don't re-derive the issue.

tests/unittest/_torch/thop/parallel_hw_agnostic/test_ulysses_permute_scatter.py
```diff
@@ -0,0 +1,93 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
chang-l, line 1: **[HIGH] Missing multi-rank unit test for the async A2A pipeline.**

These kernel-only unit tests are great, but they don't cover the actual production path: `ulysses_a2a_async_prepare` → CE push → `h->barrier()` → recv. The PR description notes "the production caller is multi-rank and not amenable to a single-GPU pytest", but that's not quite right — we have `tests/unittest/_torch/multi_gpu/` precisely for this. A `torch.multiprocessing.spawn` test with 2 ranks would catch:
- Slot-ring wrap-around bugs (loop > `kNumSlots * 2` iterations)
- Barrier race conditions / hung ranks (the timeout is 10s)
- PG-name caching bugs in `getOrCreateOp`
- `set_group_info` re-registration races across PGs
- Capture-vs-non-capture path divergence (cudaMemcpyBatchAsync vs per-peer loop)

Proposed minimal test:
```python
# tests/unittest/_torch/multi_gpu/test_ulysses_async_a2a.py
# spawn 2 ranks, build a NCCL PG, run K times:
#   recv, send_h = torch.ops.trtllm.ulysses_a2a_async_prepare(x, pg.boxed())
#   torch.ops.trtllm.ulysses_a2a_async(send_h, pg.boxed())
#   expected = torch.distributed.all_to_all_single(...)  # or hand-rolled reference
#   assert torch.equal(recv.reshape(...), expected)
# repeat with K > kNumSlots * 2 to exercise ring wrap.
```

This is also where the cuda-graph capture path can be smoke-tested under `torch.cuda.graph()`.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -396,15 +404,45 @@ def forward(
...
+                    k = apply_rotary_emb(k, freqs[0], freqs[1])
+                return k
+
+            def compute_v():
+                return attn1.to_v(normed).view(B_, Sp, num_kv_heads, head_dim)
+
+            attn1_out_4d = attn1.attn.forward_async(compute_q, compute_k, compute_v)
+            T_ = attn1_out_4d.shape[1]
+            attn1_out = attn1_out_4d.reshape(B_, T_, num_heads * head_dim)
+            attn1_out = attn1.to_out[0](attn1_out)
```
chang-l: Can the `compute_q/k/v` closure pattern (the `self._use_async_ulysses` branch) be moved into Attention.forward_async() so LTX-2 and WAN don't duplicate it?

chang-l (commented):
Thanks for the work. 
From your microbenchmark in PR description, it seems the input/output paths of our Ulysses are diverged:
  - Input QKV A2A → new symm-mem + CE path, no NCCL.
  - Output A2A → existing path, NCCL (all_to_all_single).

And the input symm-mem path has 1.2X speedup compared to output nccl path –  maybe we can extend to use `fused_permute` + symm_ce for the output path as well in future.

tensorrt_llm/_torch/visual_gen/models/ltx2/ltx2_core/utils_ltx2.py
luyiyun1021: Applied option (a) — added a comment documenting that we rely on the scheduler's upstream guarantee that sigma > 0 inside the denoise loop. Re-checking on the tensor path would re-introduce the .item() D2H sync.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
luyiyun1021: Went with (a). PR body now describes actual design: V/Q/K share _async_side_stream, barriers FIFO-serialize on that stream, channel=0 suffices. 
Also added a code comment justifying the 10 s timeout_ms: it's a hang-diagnostic budget — on timeout the PT symm-mem barrier kernel traps with "rank X failed to send signal to rank Y on channel Z after T µs" instead of spinning silently until SLURM wall-clock kills. 10 s is generous enough to absorb first-touch IPC + first cuda_graph capture jitter.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
luyiyun1021: (1) Tightened to kNumSlots = 3. Hazard analysis as code comment: minimum is 3 because V/Q/K within one forward_async must each own a distinct slot — Phase-1 default-stream write would race Phase-2 side-stream CE push on the same slot.sendBuf otherwise. Cross-layer reuse is safe because _join_async drains the side stream and SDPA on the default stream consumes the recv view before the next attention call. Comment lays out the (a) Phase-1 write → (b) CE push → (c) barrier → (d) SDPA read sequence

(2) Added multi-rank pytest at tests/unittest/_torch/visual_gen/multi_gpu/test_ulysses_async_slot_ring.py. ws=4, loops _prepare/_async for 8 iterations (≥ 2 ring rotations at kNumSlots=3), byte-exact assert vs all_to_all_4d. Passes locally.

cpp/tensorrt_llm/thop/asyncUlyssesOp.cpp
luyiyun1021: Added the cudaStreamIsCapturing guard at the top of the alloc/realloc branch in getOrAllocSlot. Cache-hit path (warmed-up slots) still returns early, so the guard only fires when we'd actually call empty_strided_p2p / rendezvous / cudaMalloc. Failure message tells the user to warm up out-of-capture before enabling cuda_graph capture. Also tightened the stale "Capture-safe because we eagerly alloc here" comment a few lines down to point at the new guard.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
luyiyun1021: Done

tests/unittest/_torch/thop/parallel_hw_agnostic/test_ulysses_permute_scatter.py
luyiyun1021: The misleading "single-GPU pytest" wording was already deleted in my v2 PR-body rewrite. New test file: tests/unittest/_torch/visual_gen/multi_gpu/test_ulysses_async.py (ws=2, torch.multiprocessing.spawn) with three test functions covering the production path end-to-end:

- test_slot_ring_wraparound — loops _prepare/_async > kNumSlots*2 iterations in eager mode (cudaMemcpyBatchAsync path), byte-exact vs all_to_all_4d on each iter.
- test_capture_smoke — warms up kNumSlots times out-of-capture (forced by my new cudaStreamIsCapturing guard in getOrAllocSlot), captures one _prepare/_async pair into torch.cuda.CUDAGraph, replays 8× with fresh inputs, byte-exact each replay. Exercises the per-peer cudaMemcpyAsync path used under production cuda_graph.
- test_multi_pg — two distinct PGs (new_group × 2) on the same rank set, alternates calls between them, byte-exact vs each PG's all_to_all_4d. Exercises PG-name caching in getOrCreateOp and set_group_info re-registration.

All 3 pass locally. Test Coverage section in PR body updated to list them. The PR body's "single-GPU pytest" wording was already gone in v2 — agree it was misleading.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
luyiyun1021: The closure is duplicated, and I do want to hoist it into Attention.forward_async(). But it can't land cleanly until a related refactor PR (currently in review) merges.

Blocker today: the two apply_rotary_emb helpers are incompatible — WAN uses (x, cos, sin) on doubled cos with 4D-first view; LTX-2 uses (x, (cos, sin), rope_type) on plain/block-duplicated cos with 3D-first view. Any single base method either breaks one of them or needs a subclass hook.

After the refactor PR lands: the fused_dit_split_norm_rope op's INTERLEAVE template + cos-shape auto-detection covers both conventions. Closure can call self.apply_split_norm_rope(...) instead of eager rope — kernel absorbs the model differences via self.interleave + cos shape, no subclass override needed.

Design sketch for the follow-up (pseudocode):

```
# modules/attention.py — new base method (~30 lines)
def forward_async(self, hidden_states, freqs=None):
    if not hasattr(self.attn, "forward_async"):
        return None
    B, S, _ = hidden_states.shape
    H, KV, D = self.num_attention_heads, self.num_key_value_heads, self.head_dim

    def compute_q():
        q = self.to_q(hidden_states)               # GEMM (lazy, overlapped)
        if self.qk_norm and freqs is not None:
            self.apply_split_norm_rope(q, self.norm_q.weight, H, *freqs)   # fused norm+rope
        elif self.qk_norm:
            self.apply_split_norm(q, self.norm_q.weight, H)                # norm-only (LTX-2 text cross-attn)
        return q.view(B, S, H, D)

    def compute_k():
        k = self.to_k(hidden_states)
        if self.qk_norm and freqs is not None:
            self.apply_split_norm_rope(k, self.norm_k.weight, KV, *freqs)
        elif self.qk_norm:
            self.apply_split_norm(k, self.norm_k.weight, KV)
        return k.view(B, S, KV, D)

    def compute_v():
        return self.to_v(hidden_states).view(B, S, KV, D)

    return self.attn.forward_async(compute_q, compute_k, compute_v)

# transformer_wan.py — caller shrinks from ~22 to ~7 lines
out_4d = self.attn1.forward_async(normed, freqs=freqs) if self._use_async_ulysses else None
if out_4d is not None:
    B_, T_, *_ = out_4d.shape
    attn1_out = self.attn1.to_out[0](out_4d.reshape(B_, T_, num_heads * head_dim))
    x = (x.float() + attn1_out.float() * gate_msa).to(x.dtype)
else:
    x = (x.float() + self.attn1(normed, freqs=freqs).float() * gate_msa).to(x.dtype)

# transformer_ltx2.py — caller shrinks from ~38 to ~10 lines
# async path is gated on (context is None and pre_projected_kv is None);
# self-attn callers don't pass k_pe (L672/L697), and the AV-cross-attn callers
# that do pass non-None k_pe (L771/L801) also set pre_projected_kv, which the
# gate excludes — so k_pe == pe holds vacuously under this gate.
# _use_async_ulysses already implies SEPARATE_QKV (forced at __init__), so no
# extra qkv_mode check is needed.
if self._use_async_ulysses and context is None and pre_projected_kv is None:
    out_4d = self.forward_async(x, freqs=pe)
    if out_4d is not None:
        if self.to_gate_logits is not None:                       # gated attention stays at caller
            gates = 2.0 * torch.sigmoid(self.to_gate_logits(x))
            out_4d = out_4d * gates.unsqueeze(-1)
        b, t = out_4d.shape[:2]
        return self.to_out[0](out_4d.reshape(b, t, H * D))
# fall through to existing FUSE_QKV / SEPARATE_QKV branches
```
GEMMs (to_q/to_k/to_v) stay inside the closures so the V→Q→K pipeline still overlaps GEMM + fused kernel with the peer push. Per-closure kernel count drops from 3 (GEMM + eager norm + eager rope) to 2 (GEMM + fused norm+rope kernel), with HBM round-trips for norm+rope halved.

The op-level multi-rank tests at tests/unittest/_torch/visual_gen/multi_gpu/test_ulysses_async.py continue to work as-is — the refactor is at the model-attention layer (closure construction), not the op layer (slot ring + prepare/async).

Plan:

Land the unify PR.
Rebase this PR onto main.
Follow-up adds base Attention.forward_async() + removes closure duplication.

luyiyun1021:
> Thanks for the work. From your microbenchmark in PR description, it seems the input/output paths of our Ulysses are diverged:
> 
> * Input QKV A2A → new symm-mem + CE path, no NCCL.
> * Output A2A → existing path, NCCL (all_to_all_single).
> 
> And the input symm-mem path has 1.2X speedup compared to output nccl path – maybe we can extend to use `fused_permute` + symm_ce for the output path as well in future.

Regarding this comment, 1.2X speedup comes from the overlapping not using CE, but yes, I can test if there are potential perf gain using the ce path for a2a output without overlapping

cpp/tensorrt_llm/kernels/ulyssesPostUnscatterKernel.cu
```diff
@@ -0,0 +1,102 @@
...
+    default:
+        in_ptr = v_in;
+        out_ptr = v_out;
+        break;
+    }
+
+    // in[p, b, sp, h, d]: ((((p*B + b)*Sp + sp)*H + h)*D + vec_idx*VEC)
+    int const in_base = ((((p * B + b) * Sp + sp) * H + h) * D) + vec_idx * VEC;
+    // out[b, h, p*Sp+sp, d]: (((b*H + h)*PSp + psp)*D + vec_idx*VEC)
+    int const out_base = (((b * H + h) * PSp) + psp) * D + vec_idx * VEC;
```
NVShreyas: should inner most values be cast to int64_t to protect from overflows?

cpp/tensorrt_llm/thop/ulyssesPostUnscatterOp.cpp
```diff
@@ -0,0 +1,78 @@
...
+//     .transpose(1, 2).contiguous()
+// for Q, K, V in one kernel launch.
+std::tuple<torch::Tensor, torch::Tensor, torch::Tensor> ulysses_post_unscatter_qkv(
+    torch::Tensor& q_in, // [P, B, Sp, H, D]
+    torch::Tensor& k_in, // [P, B, Sp, H, D]
+    torch::Tensor& v_in) // [P, B, Sp, H, D]
+{
+    TORCH_CHECK(q_in.dim() == 5 && k_in.dim() == 5 && v_in.dim() == 5,
+        "ulysses_post_unscatter_qkv expects 5D tensors [P, B, Sp, H, D]");
+    TORCH_CHECK(q_in.sizes() == k_in.sizes() && q_in.sizes() == v_in.sizes(), "Q/K/V must share the same shape");
```
NVShreyas, line 51: why is this check necessary and what would it take to support models with GQA?

tensorrt_llm/_torch/visual_gen/modules/attention.py
```diff
@@ -181,16 +186,23 @@ def __init__(
...
-                    self.attn = RingAttention(self.attn, process_group=vgm.ring_group)
-                if ulysses_size > 1:
-                    from ..attention_backend.parallel import UlyssesAttention
-
-                    self.attn = UlyssesAttention(self.attn, process_group=vgm.ulysses_group)
+            # Ring: legacy gate (FUSE_QKV only).
+            if ring_size > 1 and self.qkv_mode != QKVMode.SEPARATE_QKV:
+                from ..attention_backend.parallel import RingAttention
+
+                self.attn = RingAttention(self.attn, process_group=vgm.ring_group)
```
NVShreyas: ring attention is silently disabled when async_ulysses=True. this is fine for now but lets add a check so its explicit

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
chang-l: Thanks @luyiyun1021, I agree w/ your plan.

chang-l (commented):
Can we update the docs for this feature, for example in `docs/source/models/visual-generation.md`, `examples/visual_gen/README.md`, and, probably the server YAML files such as `examples/visual_gen/configs/wan2.2-t2v-fp4-4gpu.yaml`?

Please also make sure the docs mention that this is currently limited to MNNVL with CUDA IPC. Also, it looks like the PR description says dit_async_ulysses, but the code uses async_ulysses.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
luyiyun1021: Hi Chang @chang-l I have moved wan's forward_async to base attn, however, since ltx2 already uses a derived LTX attn class and the post-attn handling inside attn module and rope handling still bit different from the one defined in base attn(fused rms+rope pattern is unified, however there is still use_fused_rms_rope=false condition so we need to support), so that I still implement forward_async in ltx2 attn seperately for ltx2,  otherwise we still need wrapper inside base attn, which will make the base attn unclean. Do you think the current solution is reasonable?

cpp/tensorrt_llm/kernels/ulyssesPostUnscatterKernel.cu
luyiyun1021: Agree. fixed

cpp/tensorrt_llm/thop/ulyssesPostUnscatterOp.cpp
luyiyun1021: The check enforces the kernel's single-H invariant; current callers (WAN / LTX-2 self-attn via UlyssesAttention.forward_async) are all MHA by construction, so the equal-shape constraint matches their input exactly and the assert acts as a misuse guard rather than a feature gate.

GQA support is feasible (~2 split kernel launches, separate H_q / H_kv in the launcher API) but unmotivated until a GQA caller shows up.

tensorrt_llm/_torch/visual_gen/modules/attention.py
luyiyun1021: I've added the check for now. And I think ring attn does not conflict with seperate qkv or async ulysses theoretically? Maybe we can enable it in follow up pr.

NVShreyas: yep. I synced with @chang-l and we will add it in a follow up PR

cpp/tensorrt_llm/thop/ulyssesPostUnscatterOp.cpp
NVShreyas: okay we might have to do this for cosmos3 but lets keep it in a follow up PR

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
chang-l: Thanks @luyiyun1021. I am fine w/ the current implementation

chang-l (approved):

yizhang-nv (approved):
LGTM at compiler side

QiJune (approved):

tensorrt_llm/visual_gen/args.py
```diff
@@ -205,6 +205,15 @@ class ParallelConfig(StrictBaseModel):
         status="prototype",
         description=("Ulysses head-sharding degree. Heads are sharded across ulysses_size GPUs."),
     )
+    async_ulysses: bool = Field(
```
zhenhuaw-me, line 208: I thought this feature is always on if Ulysses is enabled since it should not have any perf penalty. Did I miss any case? If that's the case, we don't need a knob.

Feel free to resolve this comment to unblock the PR merging. If we converged that we don't need the knob, we can address in upcoming PRs.

tensorrt_llm/_torch/visual_gen/modules/attention.py
```diff
@@ -210,10 +238,22 @@ def __init__(
...
+        # async_ulysses forces SEPARATE_QKV, which bypasses the FUSE_QKV-only
+        # Ring gate inside wrap_parallel_attention. Surface this conflict
+        # explicitly instead of silently dropping Ring.
+        if vgm is not None and vgm.ring_size > 1 and async_ulysses:
+            raise ValueError(
+                "ring_size > 1 is incompatible with async_ulysses=True: "
+                "async_ulysses forces SEPARATE_QKV which bypasses the "
+                "RingAttention wrapper. Set ring_size=1 or "
+                "async_ulysses=False."
+            )
```
zhenhuaw-me: If we are going to keep `async_ulysses` knob, we may check this in the VisualGen validate function.

tensorrt_llm/_torch/visual_gen/modules/attention.py
```diff
@@ -115,6 +118,20 @@ def __init__(
 
         self._init_qkv_proj()
 
+        # Structural eligibility for SEPARATE_QKV self-attn quantize dedup.
+        # When True, get_qkv() may pre-quantize hidden_states once and pass the
+        # shared Fp4QuantizedTensor to to_q/to_k/to_v (relies on Linear's
+        # Fp4QuantizedTensor shortcut). Numerical equality of input_scales is
+        # guaranteed by modelopt calibration; verified for WAN (36+36 layers)
+        # and LTX-2 (42 layers) via tmp/ulysses-e2e/scripts/check_qkv_input_scale.py.
```
zhenhuaw-me: Are these agent debugging notes?

zhenhuaw-me (approved):
I left some comments about the non-kernel part. Feel free to address in the follow up PRs if applicable to unblock the PR merging as-is.

tensorrt_llm/_torch/visual_gen/modules/attention.py
luyiyun1021: removed

tensorrt_llm/visual_gen/args.py
luyiyun1021: Flux and cosmos not supported yet. Since different models has different attn module computation so we have to customize the pipeline. This may need some extra work we may leave it to future pr.

tensorrt_llm/_torch/visual_gen/modules/attention.py
luyiyun1021: moved to ParallelConfig model_validator

tensorrt_llm/visual_gen/args.py
luyiyun1021: No perf penalty in our test cases but since currently we have cuda graph perf issues with cudaMemcpyBatchAsync. I think we may leave some flexibilty in case there comes perf penalty. How do you think?

luyiyun1021:
@zhenhuaw-me Thanks for catching the enable_ulysses kwarg — you were exactly right. It was a rebase artifact: before this PR merged, base `Attention` had already been consolidated to `enable_sequence_parallel` only (in #13944 Attention2D), and the rebase conflict resolution mistakenly re-introduced the older `enable_ulysses` kwarg in a union. No model used the SP-on-but-Ulysses-off escape hatch in practice. Opened #15141 to clean it up — would appreciate your review there.

PR #13998, opened 2026-05-11
[None][fix] sync warmup forward decision across TP ranks

qiaoxj07:
## Summary

Fix a cross-comm deadlock in `PyTorchModelEngine.warmup` that surfaces under attention-DP when ranks' KV cache pools have asymmetric capacity.

## Bug

`PyTorchModelEngine.warmup` calls `_run_autotuner_warmup` and `_general_warmup_impl`, which each invoke `self.forward` only when the per-rank warmup batch is non-None:

```python
with self._release_batch_context(warmup_request, resource_manager) as batch:
    if batch is not None:   # <- rank-divergent under attention-DP
        ...
        self.forward(batch, ...)
```

Under attention-DP each rank's KV cache pool capacity can differ, so `_create_warmup_request` may return a valid batch on some TP ranks and `None` on others. The ranks then diverge:

- Ranks with valid batches enter `forward`, which (under attention-DP) issues `tp_comm` collectives in `_get_all_rank_num_tokens` (`self.dist.tp_allgather(num_tokens)`).
- Ranks with `batch is None` skip the forward entirely, exit the `autotune()` / warmup context, advance to subsequent warmup steps, and finally reach the `COMM_WORLD.Barrier()` in `PyExecutor._set_global_steady_clock_offset`.

`tp_comm` and `COMM_WORLD` are distinct MPI communicators, so the two waiting groups never satisfy each other and the job hangs forever. Verified with py-spy + gdb on a DeepSeek-V4-Pro `trtllm-bench` ctx-only run with `--tp 4 --ep 4` (attention-DP=4), `--max_num_tokens 16384`, `--max_batch_size 16`, `--kv_cache_free_gpu_mem_fraction 0.4`:

- One rank stuck in `_run_autotuner_warmup` → `forward` → `_get_all_rank_num_tokens` → `tp_allgather` (recursive-doubling step 0, waiting on a peer).
- Three ranks stuck in `_set_global_steady_clock_offset` → `MPI_Barrier(COMM_WORLD)`.

The boundary moves with available KV memory but the deadlock pattern is the same; bumping `kv_cache_free_gpu_mem_fraction` from 0.3 to 0.4 just shifted the diverging-rank set from {0,3} to {3}.

## Fix

Add a `_all_tp_ranks_have_warmup_batch` helper that `tp_allgather`s batch validity across TP ranks. Gate the warmup forward in both `_run_autotuner_warmup` and `_general_warmup_impl` on the symmetric result, so all TP peers either run or all skip.

- No-op when `tp_size <= 1`.
- The conservative skip path leaves the autotuner cache empty when any rank lacks capacity (ops fall back to default tactics), which is preferable to hanging the job.
- The extra `tp_allgather` of a single int per warmup batch is negligible.

## Test plan

- [x] Re-run the affected `trtllm-bench` ctx-only configs (`ctx_1024 BS=16 MTPx, --max_num_tokens 16384, --kv_cache_free_gpu_mem_fraction 0.3/0.4`) on 4×GB300 and confirm warmup completes instead of hanging.
- [x] Verify pre-existing TP=4 / attention-DP benchmarks (e.g. `ctx_8192 BS=2`) still pass.
- [x] `pytest tests/unittest/_torch/` for the pyexecutor area on the CI runners.

CI may need `/bot run --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1"` since this touches `_torch/pyexecutor/model_engine.py`.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/model_engine.py
+++ b/tensorrt_llm/_torch/pyexecutor/model_engine.py
@@ -853,6 +853,36 @@ class PyTorchModelEngine(ModelEngine):
         with self.no_cuda_graph():
             self._general_warmup_impl(resource_manager, warmup_requests_configs)
 
+    def _require_warmup_batch_on_all_tp_ranks(
+            self, batch: Optional[ScheduledRequests], num_tokens: int,
+            num_gen_tokens: int, warmup_name: str) -> ScheduledRequests:
+        """Return batch iff every TP rank has a valid warmup batch.
+
+        Warmup forward passes issue tp_comm collectives (e.g.
+        ``_get_all_rank_num_tokens`` under attention-DP). If only some TP
+        ranks run the forward, they deadlock against peers that skipped, and
+        those skipped peers subsequently reach the ``COMM_WORLD`` barrier in
+        ``PyExecutor._set_global_steady_clock_offset``, producing a cross-comm
+        deadlock. Raise instead of silently skipping so a missing autotuner
+        warmup is exposed rather than turning into an unexpected perf drop.
+        """
+        local_has_batch = int(batch is not None)
+        if self.mapping.tp_size <= 1:
+            flags = [local_has_batch]
+        else:
+            flags = self.dist.tp_allgather(local_has_batch)
+        flags = [int(flag) for flag in flags]
+
+        if all(flags):
+            assert batch is not None
+            return batch
+
+        raise RuntimeError(
+            f"Cannot run {warmup_name} warmup with {num_tokens} tokens and "
+            f"{num_gen_tokens} generation tokens: at least one TP rank could "
+            "not allocate a dummy warmup batch from KV cache. "
+            f"Per-TP-rank batch availability: {list(flags)}.")
+
     def _general_warmup_impl(
             self, resource_manager: ResourceManager,
             warmup_requests_configs: List[Tuple[int, int]]) -> None:
@@ -866,8 +896,8 @@ class PyTorchModelEngine(ModelEngine):
                         self._create_warmup_request(resource_manager,
                                                     num_tokens, num_gen_tokens),
                         resource_manager) as batch:
-                    if batch is None:
-                        continue  # Not enough KV cache space
+                    batch = self._require_warmup_batch_on_all_tp_ranks(
+                        batch, num_tokens, num_gen_tokens, "general")
                     logger.info(
                         f"Run warmup with {num_tokens} tokens, include {num_gen_tokens} generation tokens"
                     )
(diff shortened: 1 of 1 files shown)


hyukn (dismissed):

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -839,7 +856,7 @@ def _run_autotuner_warmup(self, resource_manager: ResourceManager):
                 resource_manager, curr_max_num_tokens, 0)
             with self._release_batch_context(warmup_request,
                                              resource_manager) as batch:
-                if batch is not None:
+                if self._all_tp_ranks_have_warmup_batch(batch):
```
liji-nv: ideally the size should have already been practiced in previous general_warmup. I see a hang today and it pass the previous warmup but takes over 30min to run autotune.
```6: [05/12/2026-05:01:16] [TRT-LLM] [I] [_torch  ] [RANK 6] Run warmup with 128 tokens, include 0 generation tokens
2: [05/12/2026-05:01:16] [TRT-LLM] [I] [_torch  ] [RANK 2] Run warmup with 128 tokens, include 0 generation tokens
7: [05/12/2026-05:01:16] [TRT-LLM] [I] [_torch  ] [RANK 7] Run warmup with 128 tokens, include 0 generation tokens
3: [05/12/2026-05:01:16] [TRT-LLM] [I] [_torch  ] [RANK 3] Run warmup with 128 tokens, include 0 generation tokens
1: [05/12/2026-05:01:16] [TRT-LLM] [I] [_torch  ] [RANK 1] Run warmup with 128 tokens, include 0 generation tokens
0: [05/12/2026-05:01:16] [TRT-LLM] [I] [_torch  ] [RANK 0] Run warmup with 128 tokens, include 0 generation tokens
4: [05/12/2026-05:01:16] [TRT-LLM] [I] [_torch  ] [RANK 4] Run warmup with 128 tokens, include 0 generation tokens
5: [05/12/2026-05:01:16] [TRT-LLM] [I] [_torch  ] [RANK 5] Run warmup with 128 tokens, include 0 generation tokens
...
7: [05/12/2026-05:01:37] [TRT-LLM] [I] [_torch  ] [RANK 7] Running autotuner warmup...
7: [05/12/2026-05:01:37] [TRT-LLM] [I] [_torch  ] [RANK 7] [Autotuner] Autotuning process starts ...
6: [05/12/2026-05:01:37] [TRT-LLM] [I] [_torch  ] [RANK 6] Running autotuner warmup...
6: [05/12/2026-05:01:37] [TRT-LLM] [I] [_torch  ] [RANK 6] [Autotuner] Autotuning process starts ...
3: [05/12/2026-05:01:37] [TRT-LLM] [I] [_torch  ] [RANK 3] Running autotuner warmup...
3: [05/12/2026-05:01:37] [TRT-LLM] [I] [_torch  ] [RANK 3] [Autotuner] Autotuning process starts ...
2: [05/12/2026-05:01:37] [TRT-LLM] [I] [_torch  ] [RANK 2] Running autotuner warmup...
2: [05/12/2026-05:01:37] [TRT-LLM] [I] [_torch  ] [RANK 2] [Autotuner] Autotuning process starts ...
4: [05/12/2026-05:01:38] [TRT-LLM] [I] [_torch  ] [RANK 4] Running autotuner warmup...
4: [05/12/2026-05:01:38] [TRT-LLM] [I] [_torch  ] [RANK 4] [Autotuner] Autotuning process starts ...
5: [05/12/2026-05:01:38] [TRT-LLM] [I] [_torch  ] [RANK 5] Running autotuner warmup...
5: [05/12/2026-05:01:38] [TRT-LLM] [I] [_torch  ] [RANK 5] [Autotuner] Autotuning process starts ...
1: [05/12/2026-05:01:38] [TRT-LLM] [I] [_torch  ] [RANK 1] Running autotuner warmup...
1: [05/12/2026-05:01:38] [TRT-LLM] [I] [_torch  ] [RANK 1] [Autotuner] Autotuning process starts ...
0: [05/12/2026-05:01:38] [TRT-LLM] [I] [_torch  ] [RANK 0] Running autotuner warmup...
0: [05/12/2026-05:01:38] [TRT-LLM] [I] [_torch  ] [RANK 0] [Autotuner] Autotuning process starts ...
```

Will skipping the autotune warmup lead to unexpected perf drop.

hyukn (commented):
The original code skips the warmup if KV cache does not have sufficient space. Thus, this PR will not break the patterns.
However, we have not encountered any similar issues based on my experience for v1. So I am not sure whether this is a difference in behavior between the two KV cache versions, or the case itself (too high memory usage) that triggers this issue.
I suggest that we can do an early drop to expose the issue instead of skipping it, when any rank is observed to have an empty warmup batch. cc @liji-nv

Issue #1053, opened 2024-02-05
New Model Support


JinPai0218:
Could you create TensorRT LLM for bark model?

OValery16:
I am also interested

hello-11:
cc  @AdamzNV @ncomly-nvidia @laikhtewari for vis.

karljang:
@JinPai0218 , apologies for the very delayed response. We haven’t had a chance to explore the model yet. While we don’t have immediate plans to support it, if it gains traction in the community and receives strong interest or upvotes, we’ll be happy to revisit it.

Issue #1015, opened 2024-01-31
API Usage Error


nanmi:
### System Info

NVIDIA   A100 

### Who can help?

@seanprime7 
@Superjomn 
@aaronp24 
@lukeyeager 
@aflat 

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Following the example in example/llama

### Expected behavior

Can infer normally

### actual behavior

```bash
[01/31/2024-08:24:58] [TRT] [E] 3: [executionContext.cpp::resolveSlots::2991] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::resolveSlots::2991, condition: allInputDimensionsSpecified(routine) )
[01/31/2024-08:24:58] [TRT] [E] 3: [executionContext.cpp::resolveSlots::2991] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::resolveSlots::2991, condition: allInputDimensionsSpecified(routine) )
[01/31/2024-08:24:58] [TRT] [E] 3: [executionContext.cpp::resolveSlots::2991] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::resolveSlots::2991, condition: allInputDimensionsSpecified(routine) )
[01/31/2024-08:24:58] [TRT] [E] 3: [executionContext.cpp::resolveSlots::2991] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::resolveSlots::2991, condition: allInputDimensionsSpecified(routine) )
[01/31/2024-08:24:58] [TRT] [E] 3: [executionContext.cpp::resolveSlots::2991] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::resolveSlots::2991, condition: allInputDimensionsSpecified(routine) )
[01/31/2024-08:24:58] [TRT] [E] 3: [executionContext.cpp::resolveSlots::2991] Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::resolveSlots::2991, condition: allInputDimensionsSpecified(routine) )
[...]

PR #13999, opened 2026-05-11
[None][fix] CppMambaHybridCacheManager: handle ranks with zero local mamba layers

VALLIS-NERIA:
## Summary

Under PP sharding, a rank can end up with only full-attention layers and no mamba layers. The previous `CppMambaHybridCacheManager.__init__` unconditionally allocated SSM/conv state and set up linear-attention metadata, leaving these ranks in an invalid state.

This PR adds an early-exit path after `super().__init__()` when `local_num_mamba_layers == 0`:

- Forward `num_layers` (not `mamba_num_layers + num_layers`) and the **union** of mamba+attention layer masks to the parent `KVCacheManager`.
- Skip all mamba-only setup (`ssm/conv_state_shape`, `linear_attention_metadata`, MTP intermediate states, replay buffers, layer offsets, etc.).
- Initialize `self.requests = []` before the early return so the guards below are safe.

Guard `prepare_resources` / `update_mamba_states` / `_setup_state_indices` with the same condition so they no-op on these ranks (the parent's `prepare_resources` still runs).

## Test plan

- [x] New unit test `test_cpp_hybrid_zero_local_mamba_layers` in `tests/unittest/_torch/executor/test_mamba_cache_manager.py`:
  - Constructs `CppMambaHybridCacheManager` with the **real** parent `KVCacheManager` (no mocks) on `world_size=1` with an all-attention layer mask — same Python branch as a PP rank with zero local mamba layers.
  - Verifies early-exit invariants: `local_num_mamba_layers == 0`, `mamba_pp_layers == []`, `requests == []`, `pp_layers == [0,1,2,3]`, parent `self.impl` / `blocks_per_window` initialized, `is_linear_attention is False`, no mamba-only attributes set.
  - Calls the three guarded methods on the zero-mamba manager to confirm they no-op without raising on the uninitialized mamba state.
- [x] Existing `test_mamba_cache_manager.py` suite still passes (10/10).

## Summary by CodeRabbit

* **Refactor**
  * Optimized Mamba cache manager initialization to skip unnecessary state setup when ranks have no local Mamba layers, improving efficiency in distributed configurations.

* **Tests**
  * Added test coverage for Mamba cache manager edge cases in distributed tensor-parallel scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13999)

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
+++ b/tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py
@@ -997,11 +997,59 @@ class CppMambaHybridCacheManager(KVCacheManager, MambaHybridCacheManager):
         mapping: Mapping,
         dtype: DataType = DataType.HALF,
         spec_config: Optional["DecodingBaseConfig"] = None,
-        layer_mask: Optional[List[bool]] = None,
+        layer_mask: Optional[
+            List[bool]] = None,  # this is the full attention layer mask
         is_estimating_kv_cache: bool = False,
         use_replay_state_update: bool = False,
         **kwargs,
     ) -> None:
+        # 3 kinds of layers:
+        # 1) Mamba layers (mamba_layer_mask is True)
+        # 2) Full attention layers (full_attention_layer_mask is True)
+        # 3) Not managed layers (both masks are False)
+        full_attention_layer_mask = layer_mask.copy()
+        total_layers = len(mamba_layer_mask)
+        layer_mask = [
+            mamba_layer_mask[i] or full_attention_layer_mask[i]
+            for i in range(total_layers)
+        ]
+        # PP sharding is done across all layers.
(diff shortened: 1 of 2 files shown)


VALLIS-NERIA:
Addressed CodeRabbit feedback:
- **#1015**: Handle `layer_mask=None` (treat as all-False full-attention mask) and validate length against `mamba_layer_mask`.
- **#1053**: Seed `self._use_replay_state_update` and `self.ssm_state_dtype` **before** the zero-mamba early-return so `get_mamba_ssm_cache_dtype()` and the `use_replay_state_update` property don't AttributeError on ranks with no local mamba layers.

Wanli-Jiang:
Verified that it can run well for new ckpt, thanks!

Wanli-Jiang (approved):

PR #14000, opened 2026-05-11
[TRTLLM-12589][fix] Reset MoE A2A dispatch state on warmup OOM

Barry-Delaney:
The MoE all-to-all dispatch/combine state machine (MoeAlltoAll._state.phase and NVLinkOneSided._dispatch_state["phase"]) is left in "dispatched" when a forward raises between dispatch() and combine(). _general_warmup_impl's "try / except OutOfMemoryError" catches the OOM and clears CUDA cache but does not reset Python-side stateful objects, so the next warmup config's forward fails the dispatch() invariant with

  "dispatch called twice without an intervening combine"

This converts the "skip this warmup shape and try a smaller one" behavior the framework intended into "die at server startup with a confusing assert that has nothing to do with memory".

Fix:
* Add reset_state() to MoeAlltoAll and NVLinkOneSided.
* In _general_warmup_impl's OOM handler, walk self.model.modules() and call reset_state() on any moe_a2a / comm submodule that exposes it, before retrying the next warmup config.

Non-OOM exceptions are deliberately not covered: they escape the `except OutOfMemoryError` block, propagate through PyExecutor.__init__ and kill the server before any further dispatch() runs, so the stuck state machine never gets exercised again -- guarding that path would violate "don't guard against scenarios that can't happen".

This restores the documented "skip OOMing config" semantics. The underlying 4096-token warmup OOM at kv_cache_free_gpu_memory_fraction=0.9 is a separate KV pool sizing issue and is not addressed here.

### DIFF
--- a/tensorrt_llm/_torch/distributed/moe_alltoall.py
+++ b/tensorrt_llm/_torch/distributed/moe_alltoall.py
@@ -310,9 +310,19 @@ class MoeAlltoAll:
             payload_in_workspace, use_low_precision_combine)
 
         # Reset state for next round
-        self._state = _A2AState()
+        self.reset_state()
 
         return output
+
+    def reset_state(self) -> None:
+        """Reset the dispatch/combine state machine to ``idle``.
+
+        Safe to call between forward passes (or from an error handler) to
+        recover from a forward that called ``dispatch`` but did not reach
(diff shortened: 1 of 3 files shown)


Barry-Delaney:
Waiting on https://github.com/NVIDIA/TensorRT-LLM/pull/14028 for fixing CI.

PR #14003, opened 2026-05-11
[None][fix] Fix CppMambaHybridCacheManager functional and perf issues

VALLIS-NERIA:
## Summary

Fixes a set of functional and performance issues in `CppMambaHybridCacheManager` and the recurrent-state cache path.

### Functional fixes

1. **Zero local mamba layers under PP sharding** (`ac30ded849`)
   Under pipeline-parallel sharding, a rank can end up with only full-attention layers and no mamba layers. The previous constructor unconditionally allocated SSM/conv state and set up linear-attention metadata, leaving such ranks in an invalid state. Added an early-exit after `super().__init__` when `local_num_mamba_layers == 0`, forwarded the union of mamba+attention layer masks to the parent `KVCacheManager`, and guarded `prepare_resources` / `update_mamba_states` to no-op on those ranks.

2. **Reserve recurrent-state slots for CUDA graph padding and draft-len sentinels** (`992ec1bd08`)
   The recurrent-state snapshot pool (`_calculate_max_num_blocks_for_linear_attention` → `max_snapshots`) was sized at exactly `max_batch_size`, which doesn't reserve room for:
   - The CUDA graph padding sentinel (`CUDA_GRAPH_DUMMY_REQUEST_ID`) — needs 1 extra slot.
   - The per-draft-len sentinels that `CUDAGraphRunner::_get_padded_batch` issues during speculative decoding — needs one extra slot per draft length.

   Under load, this caused the padding sentinels to evict live recurrent state and surface as block-allocation failures. Added `+1` for the CUDA graph padding sentinel and `+ spec_config.max_draft_tokens` when a `spec_config` is present. The block-reuse override (`max_tokens // interval`) still wins when active.

3. **Follow-ups on the slot-reservation fix**
   - `3310de0a0e` — fixed a `max_draft_tokens` typo and added unit tests covering the slot reservation logic.
   - `5a858312ce` — store `spec_config` on `self` in `__init__` so the override path can read it.
   - `5aad39a889` — addressed CodeRabbit review comments on the above.

### Performance fixes (`516ea84eaa`)

Several host/device sync points were eliminated from the mamba-hybrid prep path. With Qwen3.5-A17B-NVFP4 + MTP Eagle one-model + CUDA graphs, nsys showed `numGenReq × (1 + max_draft_len)` `cudaMemcpyAsync` + `cudaStreamSynchronize` pairs per iteration sitting inside `_prepare_inputs`, plus a `refresh_blocks` stream sync at the tail of `prepare_resources` that blocked subsequent prep work.

- **`Mamba2Metadata.state_indices`: alias the cache manager's CUDA buffer instead of per-element D2H.**
  The previous code did `for i, idx in enumerate(indices): state_indices_cpu[i] = idx`, where `indices` is `CppMambaHybridCacheManager.cuda_state_indices` (a CUDA tensor). Iterating a CUDA tensor produces 0-d CUDA slices, and assigning each to a CPU tensor issues a `cudaMemcpyAsync` + `cudaStreamSynchronize` per element. Replaced with a direct reference assignment (`self.state_indices = indices`) when the source is on CUDA, guarded by a `data_ptr()` invariant assertion to catch future buffer reallocations that would silently break CUDA graph replays. CPU/list paths keep the original behavior.

- **`KVCacheTransferManager::copyBlock` layer-first path: one `cudaMemcpy2DAsync` instead of N async memcpys.**
  Pool layout is `{numLayers, numBlocks, kvFactor, blockSize}`. For a fixed block index, per-layer slices are equal-length rows separated by stride `numBlocks * rowBytes`. Replaced the per-layer `cudaMemcpyAsync` loop with a single pitched 2D copy.

- **Defer `refresh_blocks()` (stream sync) until just before forward.**
  Split `CppMambaHybridCacheManager._prepare_resources` into two phases: (1) fire all `copy_linear_attention_block` onboards asynchronously and record whether any actually scheduled a transfer, and (2) a new `flush_state_transfers()` method that calls `refresh_blocks()` only when there are pending transfers. The flush is invoked at the end of `Mamba2Metadata.prepare()`, so the rest of `_prepare_tp_inputs` (host work + H2D copies) overlaps with the in-flight onboard transfers.

  To support the "skip if nothing scheduled" optimization, `KVCacheManager::copyLinearAttentionBlock` now returns `bool` indicating whether at least one onboard was actually issued. The 3 layers of the call (`KVCacheManager` → `BlockManager` → `WindowBlockManager`) all return `bool`; pybind reflects the new return type automatically.

## Test plan

- [x] Unit tests: `tests/unittest/_torch/executor/test_mamba_cache_manager.py` covers slot reservation across CUDA-graph + draft-len combinations.
- [x] Integration: end-to-end CUDA graph + MTP on a mamba hybrid model no longer hits "no free recurrent-state blocks".
- [x] Perf: nsys on Qwen3.5-A17B-NVFP4 + MTP Eagle one-model + CUDA graphs shows the per-(gen_req × (draft+1)) `cudaMemcpyAsync` + sync pairs gone, and the `refresh_blocks` stream sync moved out of the critical path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory allocation calculation for linear attention operations to better account for CUDA graph padding and speculative decoding configurations, enhancing overall resource management efficiency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14003)


VALLIS-NERIA:
Addressed CodeRabbit feedback: store `spec_config` on `self` in `KVCacheManager.__init__` so the linear-attention sizing branch doesn't rely on a subclass setting it before `super().__init__`.

/bot run

nv-guomingz (approved):

SimengLiu-nv (commented):
### Concern 1: Layer-first recurrent-state transfers use the primary pool shape for secondary memory

In `cpp/tensorrt_llm/batch_manager/kvCacheTransferManager.cpp`, the new layer-first `cudaMemcpy2DAsync` path derives `numBlocks`, `layerStrideBytes`, `rowBytes`, and both source/destination pitches from `pool.primaryPtr`. That is only valid for GPU-to-GPU copies where both sides have the primary pool layout.

For recurrent-state pools with host offload, `secondaryPtr` is allocated with `mNumSecondaryBlocks`, which can differ from `mNumPrimaryBlocks`. On offload/onboard, the source and destination tensors can therefore have different layer strides, but the code indexes and pitches both as if they used the primary shape. This can copy from/to the wrong host rows and corrupt restored recurrent states.

Please compute source and destination offsets/pitches from `srcPool` and `dstPool` independently, or keep the previous per-layer slice copy for layer-first pools where primary and secondary shapes may differ.

### Concern 2: Block-reuse recurrent-state sizing still drops the live/padding slot floor

In `tensorrt_llm/_torch/pyexecutor/resource_manager.py`, `_calculate_max_num_blocks_for_linear_attention()` starts with `max_batch_size + 1` plus draft sentinel slots, but then overwrites `max_snapshots` with `max_tokens // interval` whenever block reuse is enabled.

That reintroduces the underallocation this PR is trying to fix. For example, with `max_batch_size=4`, default `mamba_state_cache_interval=256`, and `max_tokens=512`, the block-reuse branch allocates only `2` recurrent-state slots, even though the live batch plus CUDA-graph padding already needs at least `5` slots before any reusable interval snapshots are considered. Under load, the padding or draft sentinel can evict or alias live recurrent state.

Please preserve the sentinel/live-state floor in the block-reuse branch, e.g. `max(max_tokens // interval, max_batch_size + 1 + max_draft_len)`, and add a regression test with `enable_block_reuse=True`; the current new tests disable block reuse, so they do not cover this path.

SimengLiu-nv (approved):
LGTM

VALLIS-NERIA:
/bot run --disable-fail-fast --stage-list "DGX_B200-4_GPUs-PyTorch-2, DGX_B200-8_GPUs-PyTorch-1, GB200-4_GPUs-PyTorch-1, GB200-4_GPUs-PyTorch-2"

VALLIS-NERIA:
/bot run --disable-fail-fast --stage-list "DGX_B200-4_GPUs-PyTorch-2, DGX_B200-8_GPUs-PyTorch-1, GB200-4_GPUs-PyTorch-1, GB200-4_GPUs-PyTorch-2"

VALLIS-NERIA:
/bot run --disable-fail-fast --stage-list "DGX_B200-4_GPUs-PyTorch-2, DGX_B200-8_GPUs-PyTorch-1, GB200-4_GPUs-PyTorch-1, GB200-4_GPUs-PyTorch-2"

yuxianq (approved):
`tensorrt_llm/_torch/pyexecutor/_util.py` LGTM

PR #13468, opened 2026-04-25
[None][feat] Pass sequence hash into kv connector

jthomson04:
## Summary by CodeRabbit

* **New Features**
  * Added block hash chain computation and tracking for KV cache block management.
  * KV cache connector now validates beam search configuration and raises an error if beam width exceeds 1.

* **Documentation**
  * Updated KV cache connector documentation to describe block hash chain details in scheduler output.

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

To see a list of available CI bot commands, please comment `/bot help`.

PR #14005, opened 2026-05-11
[None][feat] Expose stored block-hash chain to KV cache connector

pcastonguay:
## Summary

Surface the per-block hash chain to the KV cache connector by reading each `KVCacheBlock`'s stored `mHash` directly, instead of recomputing it Python-side.

This is an alternative design to #13468. It produces the same `RequestData.block_hashes` value but avoids hashing the same bytes twice for connector users:

- C++ `KVCacheManager::commitAndGetBlockHashesForRequest` walks the request's currently-allocated blocks. For each not-yet-committed full block it sets `mBlockKey` + `mHash` once, using the same `BlockKeyHasher` construction that `storeBlocks` would later apply. Subsequent calls are pure lookups.
- The Python connector's `KvCacheConnectorSchedulerOutputRequest.update_and_build_data` no longer maintains a parallel Python hash cache or a `parent_hash` argument — it just asks the manager for the chain.
- `storeBlocks`'s existing `oldHash != newHash` check now commonly sees `oldHash == newHash` and skips the rewrite. The recompute path inside `storeBlocks` is intentionally left untouched (`searchRoot->getHash()` can differ from `mPrevBlockInSeq->getHash()` after prefix-merge restructuring).

## Behaviour change vs main

- New `RequestData.block_hashes: List[int]` field, populated from C++ stored hashes. Matches the hash that KV cache events will emit for the same block. Hash is observable to event emit sites slightly earlier than today (right after the block fills, before `storeBlocks` runs) — the same "front-running" semantic #13468 documents.
- `kv_connector_config` with `max_beam_width > 1` is rejected at startup. The chain only covers beam 0, and the C++ method asserts this invariant defensively.

## Test plan

- [x] `pytest tests/unittest/_torch/test_connector.py -k "block_hashes or mtp"` — new `test_scheduler_output_block_hashes_read_through` validates pass-through behaviour; updated MTP test mocks the new manager method.
- [ ] Full unit suite on a build with the C++ change compiled in (CI).
- [ ] Soak with a connector implementation that consumes `RequestData.block_hashes` to confirm the chain matches the KV cache event stream.

## Relation to #13468

Same connector-facing API (`RequestData.block_hashes`, beam-width-1 guard, docs entry). Different mechanism: read stored hashes instead of recomputing. Worth comparing diffs side-by-side before deciding which to land.

## Release Notes

* **New Features**
  * KV Cache Connector now tracks and retrieves cumulative block hashes for committed cache blocks.

* **Improvements**
  * Added configuration validation to prevent KV Cache Connector usage with beam search configurations exceeding single-beam width.
  * Enhanced documentation clarifying KV Cache Connector scheduler output and block hash handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14005)


jthomson04 (approved):

pcastonguay:
Closing in favor of https://github.com/NVIDIA/TensorRT-LLM/pull/14806

PR #14021, opened 2026-05-12
[TRTLLM-12527][feat] Parallelize multi-shard visual-gen checkpoint loading and pre-fetch checking

yibinl-nvidia:
## Summary by CodeRabbit
Similar to LLM [sharded loading](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/models/checkpoints/hf/weight_loader.py#L106) and [pre-fetch](https://github.com/NVIDIA/TensorRT-LLM/blob/ca876e034345f613685c66c108e3279b4a6d1a82/tensorrt_llm/_torch/models/checkpoints/hf/weight_loader.py#L89).

* **Refactor**
  * Optimized weight loading for visual generation models to process sharded files concurrently, improving performance when multiple weight files are present.
  * Enhanced error reporting to clearly identify which weight file shard failed during loading.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14021)

## Description

With this PR, the improvements are great.

| model | mode | weight GB | model/pipeline load | vs baseline |
|---|---|---:|---:|---:|
| Wan2.1-T2V-14B-Diffusers | prefetch + parallel shard | 80.39 | 88.23s | 52.3% faster, 2.09x |
| Wan2.1-T2V-14B-Diffusers | parallel shard only | 80.39 | 193.55s | 4.7% slower, 0.95x |
| Wan2.1-T2V-14B-Diffusers | no features | 80.39 | 184.82s | baseline |
| Wan2.2-T2V-A14B-Diffusers | prefetch + parallel shard | 126.18 | 146.85s | 67.5% faster, 3.08x |
| Wan2.2-T2V-A14B-Diffusers | parallel shard only | 126.18 | 344.26s | 23.9% faster, 1.31x |
| Wan2.2-T2V-A14B-Diffusers | no features | 126.18 | 452.53s | baseline |
| FLUX.1-dev | prefetch + parallel shard | 34.08 | 82.34s | 2.1% faster, 1.02x |
| FLUX.1-dev | parallel shard only | 34.08 | 88.48s | 5.2% slower, 0.95x |
| FLUX.1-dev | no features | 34.08 | 84.09s | baseline |
| FLUX.2-dev | prefetch + parallel shard | 113.14 | 193.71s | 32.8% faster, 1.49x |
| FLUX.2-dev | parallel shard only | 113.14 | 292.94s | 1.6% slower, 0.98x |
| FLUX.2-dev | no features | 113.14 | 288.39s | baseline |

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
+++ b/tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
@@ -1,17 +1,71 @@
 """Weight loader for diffusion models."""
 
 import json
+import multiprocessing
+import os
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
 from typing import Any, Dict, List, Union
 
+import psutil
 import torch
 import tqdm
 
 from tensorrt_llm._torch.models.checkpoints.base_weight_loader import BaseWeightLoader
 from tensorrt_llm._torch.visual_gen.pipeline_registry import PipelineComponent
+from tensorrt_llm._utils import local_mpi_barrier, local_mpi_rank, local_mpi_size
 from tensorrt_llm.logger import logger
 from tensorrt_llm.mapping import Mapping
+
+_PREFETCH_CHUNK_SIZE = 16 * 1024 * 1024
+
+
+def _prefetch_file(file_name: str) -> None:
+    if not os.path.exists(file_name):
+        return
+
+    logger.info(f"Prefetching checkpoint file {file_name} to host page cache...")
+    with open(file_name, "rb") as f:
+        while f.read(_PREFETCH_CHUNK_SIZE):
+            pass
+    logger.info(f"Finished prefetching checkpoint file {file_name}.")
+
+
+def _prefetch_safetensors_files(file_names: List[str]) -> None:
+    """Warm safetensors files in host page cache across local ranks."""
+    paths: List[str] = []
+    seen = set()
+    for file_name in file_names:
+        path = os.path.abspath(file_name)
+        if path not in seen:
+            paths.append(path)
+            seen.add(path)
+
+    if not paths:
+        return
+
+    try:
+        prefetch_size = sum(os.path.getsize(path) for path in paths if os.path.exists(path))
+        available_memory = psutil.virtual_memory().available
+        if prefetch_size >= available_memory * 0.9:
+            logger.info(
+                "Skipping visual-gen checkpoint prefetch because files require "
+                f"{prefetch_size / (1024**3):.2f}GB and available host memory is "
+                f"{available_memory / (1024**3):.2f}GB."
+            )
+            return
+
+        local_paths = paths[local_mpi_rank() :: local_mpi_size()]
+        if local_paths:
+            logger.info(
+                f"Prefetching {prefetch_size / (1024**3):.2f}GB visual-gen "
+                "checkpoint files across local ranks."
+            )
+            max_workers = min(multiprocessing.cpu_count() * 2, 16, len(local_paths))
+            with ThreadPoolExecutor(max_workers=max_workers) as executor:
+                list(executor.map(_prefetch_file, local_paths))
+    finally:
+        local_mpi_barrier()
 
 
 class WeightLoader(BaseWeightLoader):
@@ -87,6 +141,9 @@ class WeightLoader(BaseWeightLoader):
             weight_files = self._find_weight_files(weight_dir)
             if not weight_files:
                 raise ValueError(f"No weight files found in {weight_dir}")
+
+            if all(wf.endswith(".safetensors") for wf in weight_files):
+                _prefetch_safetensors_files(weight_files)
 
             component_weights = self._load_weight_files(weight_files, component, is_pipeline)
             all_weights[component] = component_weights
--- a/tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
+++ b/tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
@@ -1,71 +1,18 @@
 """Weight loader for diffusion models."""
 
 import json
-import multiprocessing
-import os
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
 from typing import Any, Dict, List, Union
 
-import psutil
 import torch
 import tqdm
 
 from tensorrt_llm._torch.models.checkpoints.base_weight_loader import BaseWeightLoader
+from tensorrt_llm._torch.visual_gen.checkpoints.prefetch import prefetch_files_to_host_cache
 from tensorrt_llm._torch.visual_gen.pipeline_registry import PipelineComponent
-from tensorrt_llm._utils import local_mpi_barrier, local_mpi_rank, local_mpi_size
 from tensorrt_llm.logger import logger
 from tensorrt_llm.mapping import Mapping
-
-_PREFETCH_CHUNK_SIZE = 16 * 1024 * 1024
-
-
-def _prefetch_file(file_name: str) -> None:
-    if not os.path.exists(file_name):
-        return
-
-    logger.info(f"Prefetching checkpoint file {file_name} to host page cache...")
-    with open(file_name, "rb") as f:
-        while f.read(_PREFETCH_CHUNK_SIZE):
-            pass
-    logger.info(f"Finished prefetching checkpoint file {file_name}.")
-
-
-def _prefetch_safetensors_files(file_names: List[str]) -> None:
-    """Warm safetensors files in host page cache across local ranks."""
-    paths: List[str] = []
-    seen = set()
-    for file_name in file_names:
-        path = os.path.abspath(file_name)
-        if path not in seen:
-            paths.append(path)
-            seen.add(path)
-
-    if not paths:
-        return
-
-    try:
-        prefetch_size = sum(os.path.getsize(path) for path in paths if os.path.exists(path))
-        available_memory = psutil.virtual_memory().available
-        if prefetch_size >= available_memory * 0.9:
-            logger.info(
-                "Skipping visual-gen checkpoint prefetch because files require "
-                f"{prefetch_size / (1024**3):.2f}GB and available host memory is "
-                f"{available_memory / (1024**3):.2f}GB."
-            )
-            return
-
-        local_paths = paths[local_mpi_rank() :: local_mpi_size()]
-        if local_paths:
-            logger.info(
-                f"Prefetching {prefetch_size / (1024**3):.2f}GB visual-gen "
-                "checkpoint files across local ranks."
-            )
-            max_workers = min(multiprocessing.cpu_count() * 2, 16, len(local_paths))
-            with ThreadPoolExecutor(max_workers=max_workers) as executor:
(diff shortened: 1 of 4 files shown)


tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
```diff
@@ -1,6 +1,7 @@
 """Weight loader for diffusion models."""
```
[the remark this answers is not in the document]
yibinl-nvidia: No need for copyright header here.

tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
```diff
@@ -87,12 +88,9 @@ def load_weights(
...
 
-            # Load all weights with progress bar
-            component_weights = {}
-            desc = f"Loading {component}" if is_pipeline else "Loading checkpoint"
-            for wf in tqdm.tqdm(weight_files, desc=desc):
-                component_weights.update(self._load_file(wf))
-
+            component_weights = self._load_weight_files(
+                weight_files, component, is_pipeline
+            )
```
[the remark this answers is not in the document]
yibinl-nvidia: resolved.

tensorrt_llm/_torch/visual_gen/models/ltx2/pipeline_ltx2.py
```diff
@@ -152,6 +156,66 @@ def _assert_resolution(height: int, width: int, *, is_two_stage: bool = False) -
...
+        available_memory = psutil.virtual_memory().available
+        if prefetch_size >= available_memory * 0.9:
+            logger.info(
+                "Skipping LTX-2 checkpoint prefetch because files require "
+                f"{prefetch_size / (1024**3):.2f}GB and available host memory is "
+                f"{available_memory / (1024**3):.2f}GB."
+            )
+            return
+
+        local_paths = paths[local_mpi_rank() :: local_mpi_size()]
```
chang-l: My understanding is that we are not using MPI in the TRT-LLM VisualGen workflow. Instead, we use Python multiprocessing for intra-node and torchrun/srun for inter-node. So I’m not sure what you will get from `local_mpi_rank`, but it likely isn’t the real local rank orchestrated by torch.distributed.

Could you please change the rank/world size handling to use torch.distributed?

tensorrt_llm/_torch/visual_gen/models/ltx2/pipeline_ltx2.py
```diff
@@ -152,6 +156,66 @@ def _assert_resolution(height: int, width: int, *, is_two_stage: bool = False) -
...
+                f"Prefetching {prefetch_size / (1024**3):.2f}GB LTX-2 checkpoint "
+                "files across local ranks."
+            )
+            max_workers = min(multiprocessing.cpu_count() * 2, 16, len(local_paths))
+            with ThreadPoolExecutor(max_workers=max_workers) as executor:
+                list(executor.map(_prefetch_ltx2_file, local_paths))
+    except Exception as exc:
+        logger.warning(f"LTX-2 checkpoint prefetch failed; continuing without prefetch: {exc}")
+    finally:
+        _LTX2_PREFETCHED_SAFETENSORS.update(paths)
```
chang-l: This `finally` runs even when the prefetch was skipped above — e.g., the memory-pressure `return` at line 201, the `if not paths: return` at line 191, or an exception logged at line 213. As a result `_LTX2_PREFETCHED_SAFETENSORS` is updated with paths that were never warmed, and every subsequent call dedups them out via the filter at line 188 — the prefetch is silently disabled for the rest of the process lifetime. Suggest only adding to the cache on the successful path, e.g.:

```python
prefetched = False
try:
    ...
    if prefetch_size >= available_memory * 0.9:
        ...
        return
    ...
    list(executor.map(_prefetch_ltx2_file, local_paths))
    prefetched = True
except Exception as exc:
    logger.warning(...)
finally:
    if prefetched:
        _LTX2_PREFETCHED_SAFETENSORS.update(paths)
    local_mpi_barrier()
```

tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
```diff
@@ -1,17 +1,72 @@
...
+    """Warm safetensors files in host page cache across local ranks."""
+    paths: List[str] = []
+    seen = set()
+    for file_name in file_names:
+        path = os.path.abspath(file_name)
+        if path not in seen:
+            paths.append(path)
+            seen.add(path)
+
+    if not paths:
```
chang-l: `if not paths: return` here exits *before* the `try/finally`, so the early-return path does not call `local_mpi_barrier()` while the memory-skip path at line 50–56 (inside `try`) does. In practice all local ranks see the same `file_names` so the early-return is symmetric across ranks, but mixing the two patterns is a footgun: if any caller ever invokes this from a path where rank A has files and rank B doesn't (e.g. component-specific weight discovery), the local barrier will deadlock. The LTX-2 sibling at `pipeline_ltx2.py:189` already puts `if not paths: return` *inside* the `try/finally`; please mirror that here so the barrier is always reached, or document and assert that `file_names` is identical across all local ranks.

tensorrt_llm/_torch/visual_gen/models/ltx2/pipeline_ltx2.py
```diff
@@ -152,6 +156,66 @@ def _assert_resolution(height: int, width: int, *, is_two_stage: bool = False) -
...
+        return
+
+    logger.info(f"Prefetching LTX-2 checkpoint file {file_name} to host page cache...")
+    with open(file_name, "rb") as f:
+        while f.read(_LTX2_PREFETCH_CHUNK_SIZE):
+            pass
+    logger.info(f"Finished prefetching LTX-2 checkpoint file {file_name}.")
+
+
+def _prefetch_ltx2_safetensors_files(file_names: List[str]) -> None:
```
chang-l: `_prefetch_ltx2_file` / `_prefetch_ltx2_safetensors_files` here are essentially byte-for-byte duplicates of `_prefetch_file` / `_prefetch_safetensors_files` in `tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py` — same chunk size, same psutil memory check, same stride sharding, same barrier in `finally`. The only material differences are the log strings and that the LTX-2 copy has a module-level `_LTX2_PREFETCHED_SAFETENSORS` dedup set. After the rank/size fix this should be one shared helper (e.g. in `visual_gen/checkpoints/weight_loader.py` or `visual_gen/utils.py`) taking an optional dedup set, used by both call sites — otherwise the two copies will drift (someone will fix one of the bugs above but only in one copy).

tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
```diff
@@ -1,17 +1,72 @@
...
+        return
+
+    logger.info(f"Prefetching checkpoint file {file_name} to host page cache...")
+    with open(file_name, "rb") as f:
+        while f.read(_PREFETCH_CHUNK_SIZE):
+            pass
+    logger.info(f"Finished prefetching checkpoint file {file_name}.")
+
+
+def _prefetch_safetensors_files(file_names: List[str]) -> None:
```
chang-l: The LLM weight loader at `tensorrt_llm/_torch/models/checkpoints/hf/weight_loader.py` (which the PR description cites as the model) has already solved two of the issues raised on this prefetch helper:

1. **Per-rank memory-check divergence** — `hf/weight_loader.py:46-62` defines `_get_local_available_host_memory()` which does `local_mpi_comm().allreduce(available_host_memory, op=_MPI.MIN)` so all local ranks see the *same* value before the `prefetch_size >= available_memory * 0.9` decision. The docstring at lines 51-57 explicitly calls out the deadlock hazard it's solving. The version added here (line 49) just calls `psutil.virtual_memory().available` per-rank, which can diverge under transient host-memory pressure and produce the exact partial-prefetch / barrier-deadlock failure the LLM author was protecting against.

2. **Asymmetric barrier path** — `hf/weight_loader.py:99` calls `local_mpi_barrier()` **unconditionally** outside any `if enable_prefetch`, with the comment at lines 94-98 spelling out why ("different ranks may take different branches; gating the barrier on it would deadlock"). The version added here gates the barrier on entering the `try` block, so the `if not paths: return` at line 44 exits without barrier — and the LTX2 copy has the additional dedup-on-skip path.

Given there will now be **three** near-identical copies of this prefetch logic in-tree (LLM, this file, `pipeline_ltx2.py`), suggest extracting a single shared helper — e.g. `tensorrt_llm/_torch/checkpoints/prefetch.py` with an API like:

```python
def prefetch_files_with_local_sharding(
    file_paths: List[str],
    *,
    skip_if_seen: Optional[Set[str]] = None,   # for dedup across calls (LTX2 use case)
    memory_safety_ratio: float = 0.9,
    description: str = "checkpoint",
) -> None: ...
```

Then the LLM weight loader, the visual-gen weight loader, and the LTX2 pipeline all call into one place. That way:
- The MPI→torch.distributed rank/size fix happens **once**.
- The allreduce-MIN memory check and unconditional barrier are inherited by all callers.
- Future improvements (e.g., `posix_fadvise(POSIX_FADV_WILLNEED)` instead of the 16MB read loop, or rank-0-decides-and-broadcasts) only need to land in one file.

For this PR, at minimum the new code should mirror the LLM pattern verbatim on points (1) and (2) above.

chang-l (commented):
Is it possible to reduce the code dup in the PR?

tensorrt_llm/_torch/visual_gen/models/ltx2/pipeline_ltx2.py
yibinl-nvidia: good points, added shared torch-distributed rank/barrier logic in prefetch.py.

tensorrt_llm/_torch/visual_gen/models/ltx2/pipeline_ltx2.py
yibinl-nvidia: Agreed, that could silently disable future prefetch attempts.  Now `updates` happens only after successful prefetch.

tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
yibinl-nvidia: Good point, I unified the logic so all exits pass through finally and call the distributed barrier.

tensorrt_llm/_torch/visual_gen/models/ltx2/pipeline_ltx2.py
yibinl-nvidia: Good point, refactored to `prefetch_files_to_host_cache`.

tensorrt_llm/_torch/visual_gen/checkpoints/weight_loader.py
yibinl-nvidia: Good suggestion, I added `prefetch_files_to_host_cache` to unify the prefetch in LTX-2 and other visual gen models.

chang-l (approved):

PR #14012, opened 2026-05-12
[TRTLLM-11410][feat] MoT World Model Support

NVShreyas:
## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

To see a list of available CI bot commands, please comment `/bot help`.

## Release Notes

* **New Features**
  * Added Cosmos3 video generation pipeline supporting text-to-video and image-to-video generation modes.
  * Introduced text and video guardrails with profanity filtering, content safety classification, and face anonymization capabilities.
  * Enhanced diffusion model architecture with improved attention mechanisms for visual generation tasks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14012)

PR #14044, opened 2026-05-12
[TRTLLM-12653][feat] LTX-2 Ulysses cross-attention for v2a with audio padding

luyiyun1021:
## Description

This PR adds Ulysses sequence-parallelism support for the **v2a** (video-to-audio) cross-attention in LTX-2 by reusing the existing `UlyssesAttention` wrapper. It replaces the all-gather (AG) path on the heavy video K/V side with the Ulysses 4-collective (Q + K + V + output) all-to-all, which is significantly cheaper when `T_q << T_kv`. The audio padding mode, FA4 `key_padding_mask` support, audio self-attn backend fallback, and the AG-fallback gating fix are all in service of making the Ulysses v2a path safe across all supported LTX-2 configurations.

## Why v2a, not a2v?

LTX-2 has two AV cross-attentions per block:

| Module | Q | K/V |
|---|---|---|
| `audio_to_video_attn` (a2v) | video (`T_v ≈ 6300` at 768×1280 / 121 frames) | audio (`T_a ≈ 126`, padded to 128 under `audio_pad_for_ulysses=True`) |
| `video_to_audio_attn` (v2a) | audio (`T_a`) | video (`T_v`) |

### Per-rank communication volume (cross-attn K/V path), `U = ulysses_size`

| Path | Formula (bf16, 2 B/elem) | What it ships |
|---|---|---|
| **AG K/V** (baseline) | `2 × (U-1)/U × T_kv × H × D × 2` | each rank receives full K and V across ranks |
| **Ulysses 4-a2a** (Q + K + V + output) | `(U-1)/U² × 2 × (T_q + T_kv) × H × D × 2` | each rank exchanges its 1/U head shard with the other U-1 |

Setting `Ulysses < AG` and simplifying gives the break-even:

```
T_q / T_kv  <  U − 1
```

Plug in LTX-2 with `U = 4`:

| Module | `T_q / T_kv` | vs `U − 1 = 3` | Decision |
|---|---|---|---|
| v2a | 128 / 6300 ≈ 0.02 | ≪ 3 | **Ulysses wins** — switch |
| a2v | 6300 / 128 ≈ 49 | ≫ 3 | AG wins — **keep AG path** |

Concrete byte numbers (LTX-2 v2a: `H × D = 32 × 128`, `U = 4`, bf16, per v2a layer per step, per rank):

| Path | Per-rank bytes shipped | Net per layer per step per rank |
|---|---|---|
| Baseline v2a (2 AG: K, V on full `T_v = 6300`) | `2 × 0.75 × 6300 × 32 × 128 × 2 B` | **≈ 147.6 MB** |
| OPT v2a wrapper (4 a2a: Q + K + V + out) | `0.75 × (128 + 6300) × 32 × 128 × 2 B` | ≈ 37.7 MB |
| OPT a2v new AGs (audio K/V, small) | `2 × 0.75 × 128 × 32 × 128 × 2 B` | ≈ 1.5 MB |
| **OPT total (v2a wrapper + a2v new AGs)** | | **≈ 39.2 MB** |

**OPT ships 27% of baseline** for the v2a/a2v cross-attn comm budget — a theoretical **−73% byte volume** per layer per step. This is the upper bound on wall-time savings; per-kernel launch overhead (OPT has roughly 2× the kernel count) is what eats into the realized win.

Hence only v2a gets the wrapper; a2v stays on AG. Hardcoded in `BasicAVTransformerBlock._init_av_cross_modules` via `use_ulysses_cross=True` on `video_to_audio_attn` only.

### NCCL kernel time, per rank, single denoise step (`step 5` capture, 4× B200, CFG=1 × Ulysses=4, 768×1280, 121 frames)

Captured with nsys `--capture-range=cudaProfilerApi --cuda-graph-trace=node`, `TLLM_PROFILE_VISUAL_GEN_START_STOP=5-5`:

| Path | AllGather total | SendRecv total | Per-rank NCCL total |
|---|---|---|---|
| Baseline (`audio_pad_for_ulysses=false`, AG fallback) | **26.7 ms** | 27.4 ms | **54.0 ms** |
| **OPT (`audio_pad_for_ulysses=true`, v2a wrapper)** | **3.3 ms** | 41.4 ms | **44.7 ms** |
| **Δ (OPT − Baseline)** | **−23.4 ms** | **+14.0 ms** | **−9.3 ms (−17 %)** |

`AG time drops 23.4 ms` because the dominant cost item (gathering full video K/V) is gone — the remaining AGs ship only the small audio K/V (a2v) and avg per-op cost falls 8× (138 µs → 17 µs). `SR time rises 14 ms` because the 4 a2a collectives in the v2a wrapper are realized as `ncclDevKernel_SendRecv` pairs (U-1 SR ops per a2a per rank). The net 9.3 ms / step / rank saved is well below the 73% theoretical byte savings — launch overhead from going from 385 to 770 kernel calls eats the rest.

### End-to-end wall time (8× B200, CFG=2 × Ulysses=4, 768×1280, 121 frames, 40 steps)

| | Run 0 | Run 1 | **Avg** | Δ vs baseline |
|---|---|---|---|---|
| Baseline (`audio_pad_for_ulysses=false`, AG fallback) | 9.506 s | 9.502 s | **9.504 s** | — |
| OPT (`audio_pad_for_ulysses=true`, v2a wrapper engaged) | 9.113 s | 9.121 s | **9.117 s** | **−4.07 %** |

Same prompt / seed / steps; output MP4s are visually identical (sample artifacts: `tmp/ulysses-a2v/ltx2_{baseline,opt}_refactor.mp4`).

## Implementation

- `LTX2Attention.__init__` builds a **dual-attn pair** on `video_to_audio_attn` only — both a `UlyssesAttention(inner_backend=...)` wrapper and the base-class plain attention — and toggles `self.attn` between them at runtime via `set_ulysses_active(bool)`. Symmetric getter `is_ulysses_active()` lets the block forward gate the AG fallback correctly under Attention2D / ring (where `_audio_is_sharded` can be true without a Ulysses wrapper being built).
- The dual-attn block only fires when `use_ulysses_cross=True ∧ ulysses_size > 1 ∧ cp_size == 1` (pure Ulysses). Combined modes (ring+Ulysses, attn2d+Ulysses) shard the seq axis by `seq_size = cp_size × ulysses_size`, which this wrapper has not been validated against — they fall through to the existing AG path.
- `configure_audio_ulysses(audio_seq_len)` runs once per request: when `parallel.audio_pad_for_ulysses=True` (default) it pads audio to a multiple of `ulysses_size` so the wrapper can always engage; otherwise it falls back to the natural-divisibility check.
- `LTXModel.forward` pads audio (zero-pad latent, repeat-last positions/timesteps) on entry and strips the padded tail on exit. `prepare_text_cache` pre-pads audio PE so the denoise hot path sees consistent shapes. A `[B, T_a_padded]` validity mask travels through `TransformerArgs.audio_padding_mask` and is consumed by `audio_attn1` and `audio_to_video_attn`.

### Backend safety for `key_padding_mask`

Padded audio means valid Q tokens would attend to padded K positions unless the backend honors `key_padding_mask`:

| Backend | Where it runs in LTX-2 | Honors mask? | Status |
|---|---|---|---|
| VANILLA | any | yes (PR adds `key_padding_mask` → SDPA `attn_mask`) | ✓ |
| FA4 | any | yes (PR adds `key_padding_mask` → cute `seqused_k`) | ✓ |
| TRTLLM | self-attn only (cross-attn auto-falls-back to VANILLA per existing rule) | no — silently drops `**kwargs` | `audio_attn1` overrides backend to VANILLA when `audio_pad_for_ulysses=True` (mirrors the cross-attn TRTLLM→VANILLA fallback in `modules/attention.py`; audio self-attn is small at `T_a ≈ 126` so the backend downgrade is negligible) |

v2a itself never receives the mask — padded audio Q rows are stripped on `LTXModel.forward` exit, so their attention output is discarded regardless.

## Test Coverage

- `tests/unittest/_torch/visual_gen/test_vanilla_key_padding_mask.py` — single-rank parity for `VanillaAttention.forward(key_padding_mask=...)` vs unpadded SDPA.
- `tests/unittest/_torch/visual_gen/test_fa4_key_padding_mask.py` — `FlashAttn4Attention.forward(key_padding_mask=...)` on a B200: self-attn parity, cross-attn parity, and pad-fill invariance.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
```diff
@@ -190,6 +188,113 @@ def support_fused_qkv(cls) -> bool:
         return True
 
 
+class UlyssesCrossAttention(AttentionBackend):
```
NVShreyas: @luyiyun1021 Why do we need a whole new object for this? The current UlyssesAttention already supports unfused which can be used by cross attention. The only thing missing would be to convert the 2 4D A2A over K/V to 1 5D A2A but we should check performance before we do that.
FYI I've tested ulysses for cross attention in my other [PR](https://github.com/NVIDIA/TensorRT-LLM/pull/14012/changes#diff-90156b13a4def6e1149fdc90d3a5790d6ac9837bddbddc580cad0842c15fe24a)

luyiyun1021: I agree that this is overdesigned, I've updated the pr to reuse UlyssesAttention and currently do not use fused kv for simplicity.

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -553,7 +609,14 @@ def forward(
             )
             if not skip_a_self:
                 norm_ax = rms_norm(ax, eps=self.norm_eps) * (1 + ascale_msa) + ashift_msa
-                a_self_out = self.audio_attn1(norm_ax, pe=audio.positional_embeddings) * agate_msa
+                a_self_out = (
+                    self.audio_attn1(
+                        norm_ax,
+                        pe=audio.positional_embeddings,
+                        key_padding_mask=audio.audio_padding_mask,
```
yibinl-nvidia, line 700: This relies on `key_padding_mask` being honored by `audio_attn1`, but in the common LTX-2 Ulysses configurations this module can wrap TRTLLM or FA4 self-attention. Both of those backends currently accept `**kwargs` and ignore `key_padding_mask`, so with `audio_pad_for_ulysses=True` valid audio queries can still attend to padded audio keys whenever `T_a % U != 0`. That changes the real audio stream, not just the padded tail. Maybe you can either implement mask support for the non-Vanilla backends/wrappers, force padded audio attention onto a backend that honors the mask, or only enable audio padding when the selected backend can actually mask the padded tokens.

yibinl-nvidia (commented):
Left some comments

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -642,13 +707,29 @@ def forward(
...
+                # K/Q rope are applied locally (rotate-before-gather),
+                # value-preserving under a2a. No key_padding_mask — video K/V
+                # is unpadded; padded audio Q is stripped on exit by
+                # LTXModel.forward.
+                # When _audio_is_sharded is False (no padding + non-divisible
+                # T_a), ``set_ulysses_active(False)`` switches v2a back to
+                # the plain backend, and the block falls back to the AG path.
                 k_v2a, v_v2a = self.video_to_audio_attn.project_kv(vx_scaled)
-                if self._use_seq_parallel:
+                if self._audio_is_sharded:
```
yibinl-nvidia: `_audio_is_sharded` is true not only for the v2a Ulysses case. For example, under Attention2D `configure_audio_ulysses()` can set it true because generic sequence parallelism is enabled and padding makes the audio length divisible, but `video_to_audio_attn` is not wrapped with Ulysses because `ulysses_size == 1`. 

We probably should gate this path when the actual v2a Ulysses wrapper is active or keep the all-gather path for Attention2D/non-Ulysses sequence seperate.

luyiyun1021: You're right. I added a getter `LTX2Attention.is_ulysses_active()` symmetric with the existing `set_ulysses_active(...)`. It returns true iff `self.attn` is currently the Ulysses-wrapped path,
which is the actual condition we want to gate on. The v2a forward now uses `self.video_to_audio_attn.is_ulysses_active()` instead of `self._audio_is_sharded`, so Attention2D + audio padding falls through to the AG branch as before this PR.

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
luyiyun1021: hi yibin, what backend does ltx2 currently support? I'm considering adding key_padding_mask to these backend.

luyiyun1021: I see trtllm, fa4, vanilla for now. fa4 extends easily as vanilla. For trtllm, seems already does not support cross attn and padding, I need to figure out a proper solution.

luyiyun1021: I think we may fallback audio self attn to vanilla if we open audio padding?

luyiyun1021: FA4 now honors key_padding_mask via seqused_k (FA4's cute interface accepts per-batch valid lengths on padded [B,S,H,D] input, so no varlen repack). For TRTLLM, audio_attn1 falls back to VANILLA at
construction when audio_pad_for_ulysses=True. Audio self-attn is tiny (T_a ~ 126), so the backend downgrade is negligible.

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -137,6 +138,46 @@ def __init__(
             )
             self._has_dual_attn = True
 
+        # Cross-attention Ulysses wrap (v2a in LTX-2). Active only when
+        # the configured parallelism is Ulysses (ulysses_size > 1).
+        # Attention2D mode keeps ulysses_size == 1 (mapping enforces mutual
+        # exclusion), so this branch is naturally skipped under Attention2D.
+        self._has_cross_dual_attn = False
+        if use_ulysses_cross and self._is_cross_attn and ulysses_size > 1:
```
NVShreyas: should probably add a gate here for ring/attention2d as well since ring+ulysses is enabled on main and attention2d+ulysses will be enabled soon

luyiyun1021: done, only enable cross attn ulysses for pure ulysses for now

NVShreyas: thanks!

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -137,6 +138,44 @@ def __init__(
...
+        if use_ulysses_cross and self._is_cross_attn and ulysses_size > 1 and cp_size == 1:
+            U = ulysses_size
+            H = self.num_attention_heads
+            H_kv = self.num_key_value_heads
+            if H % U != 0 or H_kv % U != 0:
+                raise ValueError(
+                    f"Ulysses cross-attn requires num_attention_heads ({H}) and "
+                    f"num_key_value_heads ({H_kv}) divisible by ulysses_size ({U})"
+                )
+            from tensorrt_llm._torch.visual_gen.attention_backend.parallel import UlyssesAttention
```
NVShreyas: could we move imports to the top please?

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -234,6 +295,10 @@ def forward(
             k_pe: Separate (cos, sin) RoPE embeddings for K (for AV cross-attn).
             pre_projected_kv: Pre-projected (k, v) tuple from project_kv().
                 When provided, skips K/V projection and K-norm (already done).
+            key_padding_mask: Optional ``[B, S_kv]`` bool tensor; True = valid,
+                False = pad. Forwarded through ``_attn_impl`` to the backend.
+                Only ``VanillaAttention`` (and ``UlyssesAttention`` wrapping it)
```
NVShreyas: FA4 too?

NVShreyas (approved):
LGTM from ulysses cross attention and attention backend perspective.
Might be good to have tests validating output with ulysses enabled for audio?

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
luyiyun1021: Done

tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
luyiyun1021: Done

luyiyun1021: Done

luyiyun1021:
> LGTM from ulysses cross attention and attention backend perspective. Might be good to have tests validating output with ulysses enabled for audio?

@NVShreyas Do you mean e2e test testing the output quality or testing the padding correctness for fa4 and vanilla backend? I think for the former one, there is no test yet and for the latter one, I've added two tests.

luyiyun1021:
Added `test_ulysses_with_key_padding_mask_parity` test to test key_padding_mask works properly with ulysses

NVShreyas:
> > LGTM from ulysses cross attention and attention backend perspective. Might be good to have tests validating output with ulysses enabled for audio?
> 
> @NVShreyas Do you mean e2e test testing the output quality or testing the padding correctness for fa4 and vanilla backend? I think for the former one, there is no test yet and for the latter one, I've added two tests.

yes I meant e2e test with a small input size. It is fine as long as there is some test ensuring output of transformer is consistent with ulysses enabled, for both self and cross attention.

luyiyun1021:
added `multi_gpu/test_ltx2_ulysses.py` for e2e test

zhenhuaw-me (approved):

PR #14066, opened 2026-05-13
[None][fix] Disable PDL by default for trtllm-gen MoE pipeline kernels

lancelly:
The trtllm-gen MoE all-to-all pipeline (prepare_dispatch, dispatch, sanitize_expert_ids, prepare_combine, combine) and the BlockScaleMoe expert-compute pipeline (convertsf, activation, finalize) all launch with PDL when ``TRTLLM_ENABLE_PDL=1`` (the workload's default). The kernels themselves call ``cudaGridDependencySynchronize`` and ``cudaTriggerProgrammaticLaunchCompletion`` inline.

Under CUDA Graph replay on disagg / attention-DP workloads, this combination deadlocks: the PDL trigger from one kernel does not reliably reach the next kernel's GridDependencySynchronize during a captured graph replay, so one random rank ends up exactly one MoE round behind every replay until the in-kernel 300s completion-flag spin-wait times out and traps with ``cudaErrorLaunchFailure``. The "DSv4-Pro disagg dispatch hang" reproduces this every run within ~1 min of bench start: every non-victim rank prints

    dispatch: ---Rank N timed out waiting for completion flag from rank X
    (read=R-2 expected=R my_flag_val=R addr=...)

and the victim's stack stays in sampler_event.synchronize while the other 15 ranks deadlock in tp_gather from the unsynced ``_update_requests`` exception path. The victim rank is random across runs (5 / 2 / 12 / 4 / 15 observed), the drift is always exactly 2.

Following PR 11994's pattern (which gated PDL off-by-default for moeRoutingRenormalize after the same family of race produced NaNs):

* Add ``getEnvEnableTrtllmgenMoePdl()`` (env ``TRTLLM_ENABLE_TRTLLMGEN_MOE_PDL``, default off).
* Add ``launchWithPdlWhenBothEnabled(name, enableSubsys, ...)`` that AND's the global PDL gate with the subsystem-specific gate.
* Switch all six A2A launch sites in ``moeAlltoAllKernels.cu`` from ``launchWithPdlWhenEnabled`` to the new wrapper, gated on the new env helper.
* Switch the three ``mUsePdl = true`` assignments in ``blockScaleMoe/runner.cu`` (convertsf, activation, finalize) to the env helper.

After this change, ``TRTLLM_ENABLE_PDL=1`` (the default) keeps PDL on everywhere except the racy MoE pipeline, so unrelated kernels (FMHA / BMM / GEMM outside MoE / non-MoE attention paths) keep their PDL benefit. Opt back into the racy path for perf comparison by setting ``TRTLLM_ENABLE_TRTLLMGEN_MOE_PDL=1``.

Validation on the DSv4-Pro disagg benchmark, ``TRTLLM_ENABLE_PDL=1``, CUDA Graphs on, default new env var (off):

    concurrency 2048: 20480/20480 prompts, 77.15 req/s, 71183 tok/s
    concurrency 2151: 21510/21510 prompts, 78.93 req/s, 72838 tok/s
    concurrency 2253: 22530/22530 prompts, 79.81 req/s, 73648 tok/s

Zero watchdog firings, zero dispatch timeouts, zero hang detections. Throughput is on par with the global ``TRTLLM_ENABLE_PDL=0`` workaround while preserving PDL for non-MoE PDL-eligible kernels.


cpp/tensorrt_llm/kernels/communicationKernels/moeAlltoAllKernels.cu
```diff
@@ -589,7 +590,8 @@ __global__ void moeA2ADispatchKernel(int32_t const* token_selected_experts, // [
 
 void moe_a2a_prepare_dispatch_launch(MoeA2ADispatchParams const& params)
 {
-    launchWithPdlWhenEnabled("moeA2APrepareDispatchKernel", moeA2APrepareDispatchKernel, 1, params.ep_size, 0,
+    launchWithPdlWhenBothEnabled("moeA2APrepareDispatchKernel",
```
bobboli, line 593: This PR gates the one-sided A2A dispatch/combine/sanitize launches on `TRTLLM_ENABLE_TRTLLMGEN_MOE_PDL`. That is a trtllm-gen MoE policy knob, but these are A2A communication kernels.

Unless the reported hang is specifically caused by the A2A kernels, please avoid changing their launch policy here. Keep the A2A launches under the existing generic PDL gate, or use a separate A2A-specific gate with a clear failure mode.

PR #14069, opened 2026-05-13
[https://nvbugs/6152892][fix] Fix Triton MOE memory free when no swizzling enabled

dongfengy:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced compatibility with NVIDIA GPU devices including H20-family processors.
  * Fixed memory management in tensor weight and scale operations to prevent unintended storage deallocation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14069)

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
@@ -1140,8 +1140,8 @@ class TritonMXFP4FusedMoEMethod(TritonUnquantizedFusedMoEMethod):
         new_weight, new_scale = swizzle_weight_and_scale(
             weight_data, scale_data)
         replacement_storage_ptrs = {
-            new_weight.storage.data.untyped_storage().data_ptr(),
-            new_scale.storage.data.untyped_storage().data_ptr(),
+            new_weight.data.untyped_storage().data_ptr(),
+            new_scale.data.untyped_storage().data_ptr(),
         }
         for name in (weight_name, scale_name):
             old_param = module._parameters.pop(name, None)
--- a/tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
@@ -1139,11 +1139,23 @@ class TritonMXFP4FusedMoEMethod(TritonUnquantizedFusedMoEMethod):
                              scale_data):
         new_weight, new_scale = swizzle_weight_and_scale(
             weight_data, scale_data)
+        replacement_storage_ptrs = {
+            new_weight.storage.data.untyped_storage().data_ptr(),
+            new_scale.storage.data.untyped_storage().data_ptr(),
+        }
         for name in (weight_name, scale_name):
             old_param = module._parameters.pop(name, None)
             assert old_param is not None, \
                 f"Expected {name} to be a registered parameter before swizzling MXFP4 weights."
-            old_param.data.storage().resize_(0)
+            old_storage = old_param.data.untyped_storage()
+            # H20 uses StridedLayout as an MXFP4 swizzle fallback. That
+            # conversion can alias the original scale storage, unlike the
+            # Hopper swizzled layout path, so only release storage that is no
+            # longer backing the replacement tensor.
+            old_storage_ptr = old_storage.data_ptr()
+            if (old_storage.nbytes() > 0
+                    and old_storage_ptr not in replacement_storage_ptrs):
+                old_storage.resize_(0)
         torch.cuda.empty_cache()
         setattr(module, weight_name, new_weight)
         setattr(module, scale_name, new_scale)
--- a/tests/integration/defs/accuracy/test_llm_api_pytorch.py
+++ b/tests/integration/defs/accuracy/test_llm_api_pytorch.py
@@ -26,8 +26,6 @@ from mpi4py.futures import MPIPoolExecutor
 
 from tensorrt_llm import LLM
 from tensorrt_llm._torch.model_config import MoeLoadBalancerConfig
-from tensorrt_llm._torch.modules.fused_moe.fused_moe_triton import \
-    is_swizzling_supported
 
 # isort: off
 from tensorrt_llm.llmapi import (
@@ -42,8 +40,9 @@ from tensorrt_llm.quantization import QuantAlgo
 
 from ..conftest import (get_device_count, get_device_memory, llm_models_root,
                         parametrize_with_ids, skip_no_hopper,
-                        skip_post_blackwell, skip_pre_ada, skip_pre_blackwell,
-                        skip_pre_hopper, skip_ray)
+                        skip_no_mxfp4_swizzle, skip_post_blackwell,
+                        skip_pre_ada, skip_pre_blackwell, skip_pre_hopper,
+                        skip_ray)
 from .accuracy_core import (GSM8K, MMLU, CnnDailymail, GPQADiamond,
                             JsonModeEval, LlmapiAccuracyTestHarness,
                             LongBenchV1, LongBenchV2)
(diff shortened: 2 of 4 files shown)


tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
```diff
@@ -1126,11 +1139,23 @@ def _swizzle_and_replace(module, weight_name, scale_name, weight_data,
...
             old_param = module._parameters.pop(name, None)
             assert old_param is not None, \
                 f"Expected {name} to be a registered parameter before swizzling MXFP4 weights."
-            old_param.data.storage().resize_(0)
+            old_storage = old_param.data.untyped_storage()
+            # H20 uses StridedLayout as an MXFP4 swizzle fallback. That
+            # conversion can alias the original scale storage, unlike the
+            # Hopper swizzled layout path, so only release storage that is no
+            # longer backing the replacement tensor.
+            old_storage_ptr = old_storage.data_ptr()
```
yuxianq, line 1155: We replace two tensors weight/scale here, for H20, don't need to replace both of them. So why don't we just skip the whole replacement in H20 case, i.e., `if not is_swizzling_supported()`?

dongfengy: The code here is more general and future-proof. For now we know that the old data is reused when not swizzling. But in the future that may change since it's Triton MOE's 3rd party code. By actually checking the storage is same or not, we ensure that all the future cases are handled. Maybe in the future swizzling will also reuse.

yuxianq: If so, we can use `if new_weight.untyped_storage().data_ptr() != weight_data.untyped_storage().data_ptr()` instead of `if not is_swizzling_supported()`, is it simpler?

tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
```diff
@@ -690,6 +690,20 @@ class TritonMXFP4FusedMoEQuantScales(NamedTuple):
     fc2_input_dequant: torch.Tensor
 
 
+def is_swizzling_supported() -> bool:
```
jieli-matrix, line 693: actually the function imply the logic -- "skip on H20 since MXFP4 swizzle not supported".
I suggest replace the current implementation with a marker in conftest:
```
skip_no_mxfp4_swizzle = pytest.mark.skipif(
    check_device_contain(["H20"]) and not check_device_contain(["H200"]),
    reason="nvbugs/5446119: MXFP4 swizzle not supported on H20")
```

tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
dongfengy: I am not sure this can be further simplified. Could you clarify? This is supposed to be a best effort clean up of not used data. The replacement itself is needed because the new_weight will be a Triton wrapper even though the underlying ptr is the same. The object is different. So it can't be skipped.

tests/integration/defs/accuracy/test_llm_api_pytorch.py
```diff
@@ -5371,7 +5367,7 @@ class TestGPTOSS(LlmapiAccuracyTestHarness):
                              ids=["v2_kv_cache", "v1_kv_cache"])
     def test_w4_1gpu(self, kv_cache_dtype, moe_backend, cuda_graph,
                      overlap_scheduler, mocker, v2_kv_cache):
-        if (moe_backend == "TRITON" and is_h20_gpu()):
+        if (moe_backend == "TRITON" and not is_swizzling_supported()):
```
jieli-matrix: I think it'd better to add test skip decorator at top of the test case instead of inline condition checks:
```
 @skip_no_hopper
    @skip_no_mxfp4_swizzle
    def test_w4_1gpu_piecewise_cuda_graph(self, mocker):
```

```
@pytest.mark.parametrize("moe_backend", [
        "CUTLASS",
        pytest.param("TRTLLM", marks=skip_pre_blackwell),
        pytest.param("TRITON",
                     marks=[skip_no_hopper, skip_no_mxfp4_swizzle])
    ],
...
    def test_w4_4gpus(
```

tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
```diff
@@ -717,8 +731,7 @@ def swizzle_weight_and_scale(w: torch.Tensor, w_scale: torch.Tensor):
         mx_axis=1)
     scale_layout, scale_layout_opts = layout.make_default_matmul_mxfp4_w_scale_layout(
         mx_axis=1, num_warps=num_warps)
-    # swizzling path is broken for H20
-    if torch.cuda.get_device_name() == "NVIDIA H20":
+    if not is_swizzling_supported():
```
jieli-matrix, line 734: Given the new test skip decorator, I think the original code is okay since only skipping on H20?

jieli-matrix (changes requested):
LGTM, but some comments from SDET perspective:
The test-skip logic for H20+TRITON MXFP4 currently uses inline `if … pytest.skip()` calls inside test method bodies.
However, the inline pattern is harder to discover, doesn't appear in `pytest --collect-only` output, and duplicates the same condition in three test methods.
actually `is_swizzling_supported` appears for two use:
1. Test code (what I request changes): skip tests entirely on H20 —declarative marker is the right pattern.
2. Runtime code (swizzle_weight_and_scale): runtime fallback to StridedLayout on H20 — this must stay inline because it's the actual behavioral branch that makes the code work on H20.
No logic breaks in current commits, but better to make it clearer & reusable for tests configs.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
yuxianq: Got it.

yuxianq (approved):

tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
dongfengy: This code is still needed. So we turn off swizzling due to H20 kernel issue here so that the test won't get accuracy = 0. This is the WAR before the kernel fix. But, sometimes, with disabled swizzling, some tests will still time out due to being too slow. So some additional skips are needed. They serve different purposes.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_triton.py
dongfengy: skip_no_mxfp4_swizzle has been added.

tests/integration/defs/accuracy/test_llm_api_pytorch.py
dongfengy: done

dongfengy:
Test pass on H20
============================================================= slowest durations =============================================================
136.93s call     accuracy/test_llm_api_pytorch.py::TestGPTOSS::test_dummy_load_format
0.85s setup    accuracy/test_llm_api_pytorch.py::TestGPTOSS::test_dummy_load_format

(1 durations < 0.005s hidden.  Use -vv to show these durations.)
================================================= 1 passed, 3 warnings in 145.35s (0:02:25) =================================================

dongfengy:
/bot run --stage-list "DGX_B200-PyTorch-1,DGX_B200-PyTorch-2,DGX_B200-PyTorch-3,DGX_B200-PyTorch-4,DGX_B200-PyTorch-5"

jieli-matrix (approved):
LGTM

jieli-matrix:
/bot run --stage-list "DGX_B200-PyTorch-1,DGX_B200-PyTorch-2,DGX_B200-PyTorch-3,DGX_B200-PyTorch-4,DGX_B200-PyTorch-5"

dongfengy:
H20 is not included in CI. The above stages all pass. Let's reuse the passing stages. (I just build the rebased commit and checked H20 H100 H200 all pass)

PR #14074, opened 2026-05-13
[None][fix] Support DeepSeekV4 routing in perfect-router planner

qiaoxj07:
## Summary

`DeepSeekV3PerfectRouterPlanner._resolve_group_config()` only knew how to introspect `DeepSeekV3MoeRoutingMethod`, which nests `n_group` / `topk_group` on `routing_impl`. `DeepSeekV4MoeRoutingMethod` stores them directly on the outer object, so passing it in raised at model-init time:

```
ValueError: DeepSeekV3 perfect-router planning requires routing_method.routing_impl to provide n_group and topk_group
```

This crashed every executor rank when `ENABLE_PERFECT_ROUTER=1` was set on DSv4 (e.g. `bm_deepseek-v4-pro-eplb-perfectrouter-sol-1024-1024-ratio08-...` benchmarks). The error path went through `DeepseekV4MoE.__init__` -> `ConfigurableMoE.__init__` -> `_init_perfect_router()` -> `precompute_common_perfect_router_logits()` -> `planner.get_cache_signature()` -> `_resolve_group_config()`, hitting the same exception on the fallback `model_loader.load()` retry.

## Change

Fall back to the routing method itself when `routing_impl` is absent. Both V3 (nested on `routing_impl`) and V4 (flat on the outer object) layouts now resolve to the same `(n_group, topk_group)` tuple, and the error message is updated to mention both supported shapes.

Add a CPU-only regression test `test_perfect_router_resolve_group_config` parametrized over both routing classes that drives the planner directly.

## Impact

* Unblocks `ENABLE_PERFECT_ROUTER=1` on DSv4 (perfect-router load-balancer benchmarks).
* Backwards-compatible: V3 still resolves through `routing_impl` first.
* No change to runtime kernels or numerics; init-time config check only.

## Test Plan

- [x] `pre-commit run` clean on both modified files
- [x] `pytest tests/unittest/_torch/modules/test_moe_routing.py::test_perfect_router_resolve_group_config` (CI)
- [x] Rerun the failing DSv4 DEP=16 perfect-router benchmark to confirm init succeeds.

### DIFF
--- a/tensorrt_llm/_torch/modules/fused_moe/routing.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/routing.py
@@ -946,8 +946,14 @@ class BasePerfectRouterPlanner:
         the learned bias must be masked out before those logits are consumed.
         This helper mutates the routing-method instance so subsequent routing
         calls observe the zero-bias callable.
+
+        Routing methods that don't carry a learned bias (e.g.
+        ``DeepSeekV4MoeRoutingMethod`` with ``is_hashed=True`` returns ``None``
+        from its bias callable) have nothing to mask, so this becomes a no-op.
         """
         bias = routing_method.e_score_correction_bias
+        if bias is None:
+            return
         zero_bias = torch.zeros_like(bias, device=device)
         routing_method.callable_e_score_correction_bias = lambda zero_bias=zero_bias: zero_bias
 
--- a/tensorrt_llm/_torch/modules/fused_moe/routing.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/routing.py
@@ -497,6 +497,14 @@ class DeepSeekV3MoeRoutingMethod(BaseMoeRoutingMethod):
         return self.routing_impl.top_k
 
     @property
+    def n_group(self) -> int:
+        return self.routing_impl.n_group
+
+    @property
+    def topk_group(self) -> int:
+        return self.routing_impl.topk_group
+
+    @property
     def routing_method_type(self):
         return RoutingMethodType.DeepSeekV3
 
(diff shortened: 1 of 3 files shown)


xxi-nv (approved):

tensorrt_llm/_torch/modules/fused_moe/routing.py
```diff
@@ -1172,18 +1178,23 @@ def _resolve_group_config(
...
-                                                   routing_impl, "topk_group"):
+        source = getattr(routing_method, "routing_impl", None) or routing_method
+        if not hasattr(source, "n_group") or not hasattr(source, "topk_group"):
             raise ValueError(
-                "DeepSeekV3 perfect-router planning requires routing_method.routing_impl "
-                "to provide n_group and topk_group")
-        return routing_impl.n_group, routing_impl.topk_group
+                "DeepSeekV3 perfect-router planning requires routing_method "
+                "(or its routing_impl) to provide n_group and topk_group")
+        return source.n_group, source.topk_group
```
xxi-nv: It is suggested to only return one value which represents the topk_group, so the user doesn't have to recognize it whether V3 or V4.

qiaoxj07: Addressed in b80ff77: pushed the layout abstraction into the routing-method classes themselves. `DeepSeekV3MoeRoutingMethod` now exposes `n_group` / `topk_group` via `@property` (delegating to `routing_impl`); `DeepSeekV4MoeRoutingMethod` already had them as direct attributes. The planner now reads `routing_method.n_group` / `routing_method.topk_group` directly and `_resolve_group_config` is removed — the planner no longer needs to recognize V3 vs V4.

PR #14077, opened 2026-05-13
[None][chore] Make poetry.lock update opt-in in flashinfer-upgrade skill

yihwang-nv:
## Summary
- Add a third user question in the `flashinfer-upgrade` skill's Step 3 asking whether to also update `security_scanning/poetry.lock`; default is **No**.
- Update Step 4's file table and the lockfile-hash subsection to only apply when the user opts in.
- Update Step 7 (branch / commit / PR) so the lockfile is dropped from the stash list, `git add`, and commit message body if the user opted out.

## Why
The `security_scanning/poetry.lock` file is typically regenerated by maintainers in a dedicated step. Editing it inside the flashinfer-upgrade workflow tends to produce spurious hash diffs and stale `metadata.content-hash` values, so making it opt-in keeps the default skill output clean.

## Test plan
- [x] Run the `/flashinfer-upgrade` skill end-to-end and confirm the new prompt appears
- [x] Confirm that selecting "No" leaves `security_scanning/poetry.lock` untouched
- [x] Confirm that selecting "Yes" still walks through the lockfile-hash + `poetry lock --no-update` flow

## Release Notes

* **New Features**
  * Users can now choose whether to update specific dependency lockfiles during upgrade operations.
  * Automated workflows respect user preferences by conditionally including or excluding selected files from version updates and commit operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14077)


yihwang-nv:
/bot skip --comment "This is skill update only, don't need CI"

wenmingw (approved):

yihwang-nv:
/bot skip --comment "Only add a skill, no need testing"

PR #14082, opened 2026-05-13
[None][fix] Gemma4 CUDA-graph test KV pre-alloc + L0 registration

Hudayday:
## Description

  Two test-side Gemma4 changes, both safe to land on `main` today (the
  current `transformers==5.3.0` pin).

  1. **Multi-step decode CUDA-graph IMA fix.**
     `TestGemma4CUDAGraph::test_cuda_graph_multi_step_decode` and
     `test_cuda_graph_decode_high_gqa` reserved only `initial+1` KV slots via
     `add_dummy_requests`, then grew `kv_len` per decode step. Once it crossed
     a `tokens_per_block` boundary, `paged_kv_indptr_decode` expected one more
     block entry than `get_batch_cache_indices` returned, leaving the trailing
     slot at uninitialized `torch.empty()` garbage → FlashInfer decode kernel
     read a stale `block_id` → IMA at `graph.replay()`. Production is immune
     because `resource_manager.update_resources` grows capacity every
     iteration; the hand-rolled test loop skipped that. One-line test fix:
     `token_nums = [t + 1 + num_decode_steps for t in initial_cached]`.

  2. **Register Gemma4 tests in `l0_b200.yml`.** Adds
     `test_modeling_gemma4.py` and
     `test_gemma4_e2e_dummy.py::test_e2e_multimodal_26b_dummy`. Both self-skip
     on transformers<5.5 (module-level `pytest.importorskip` and per-test
     `@requires_gemma4_transformers` respectively), so on `main` today they
     are collected and cleanly skipped — the only observable CI delta is two
     extra `skipped` lines. Once the transformers 5.5.x upgrade
     lands they start exercising real coverage.

  ## Test Coverage

  - `test_modeling_gemma4.py` 62/62 PASS in 83s on B200 + transformers 5.5.3,
    including both previously-IMA-ing CUDA-graph tests.
  - On `main` today (transformers 5.3.x): both registered targets skip; no
    other tests touched.

### DIFF
--- a/tests/integration/test_lists/test-db/l0_b200.yml
+++ b/tests/integration/test_lists/test-db/l0_b200.yml
@@ -132,10 +132,12 @@ l0_b200:
   - unittest/_torch/modeling -k "modeling_mixtral"
   - unittest/_torch/modeling -k "modeling_gpt_oss"
   - unittest/_torch/modeling/test_modeling_exaone_moe.py
+  - unittest/_torch/modeling/test_modeling_gemma4.py
   - unittest/_torch/modeling/test_gemma4_e2e_dummy.py::test_e2e_text_26b_dummy
   - unittest/_torch/modeling/test_gemma4_e2e_dummy.py::test_e2e_text_e2b_dummy
   - unittest/_torch/modeling/test_gemma4_e2e_dummy.py::test_e2e_text_31b_dummy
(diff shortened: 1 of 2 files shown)


lfr-0531 (approved):
LGTM~

Issue #13893, opened 2026-05-08
[Feature]: Add offloading support to visual_gen to reduce GPU memory usage


rahul-steiger-nv:
## Feature request

I would like to add offloading support to `tensorrt_llm/_torch/visual_gen`.

The goal is to reduce peak GPU memory usage by allowing selected parts of visual generation pipelines, such as model components, auxiliary modules, or guardrails, to be offloaded when they are not actively needed during generation.

## Motivation

This would help support more memory-heavy visual generation workloads on a wider range of GPUs.

In particular, it would be useful for:

- GPUs with limited VRAM, including consumer/workstation GPUs
- larger visual generation models
- longer video generation
- multi-view generation
- pipelines that include extra components such as safety or guardrail models

Initial experiments suggest that offloading can significantly reduce peak GPU memory usage with limited latency overhead.

## Proposal

Add a general offloading mechanism to `visual_gen` that can be configured by users and reused across visual generation models, rather than relying on model-specific pipeline changes.

I would be interested in contributing an initial implementation and would appreciate guidance on the preferred design/API for making this upstreamable.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and checked the [documentation](https://nvidia.github.io/TensorRT-LLM/) and [examples](https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples) for answers to frequently asked questions.

NVShreyas:
cc @chang-l this request is from our team

PR #14095, opened 2026-05-13
[TRTLLM-11412][feat] Add offloading support for visual_gen

rahul-steiger-nv:
## Description

This PR adds initial offloading support for `tensorrt_llm/_torch/visual_gen`.

The goal is to reduce peak GPU memory usage for memory-heavy visual generation pipelines by allowing selected model components, auxiliary modules, or guardrails to be offloaded when they are not actively needed during generation.

This is intended to help support larger visual generation workloads, longer video generation, multi-view generation, and lower-VRAM GPUs.

This PR is an initial implementation for the feature request in #13893.

Multi-GPU support and quantization compatibility are still being validated. There may also be some design changes after further discussion to make the implementation more general and easier to maintain upstream.

## Test Coverage

Current testing:

- Tested visual generation with offloading enabled on a single GPU.
- Verified that peak GPU memory usage is reduced compared to running without offloading.
- Verified that generation completes successfully with offloading enabled.

Planned / in-progress testing:

- Multi-GPU execution.
- Quantized model paths.
- Additional validation of latency overhead.
- Additional tests for configuration/error handling once the final design is agreed upon.

## PR Checklist

## GitHub Bot Help

## Dev Engineer Review

- Added staged CPU offloading for visual-generation components, including Wan, Cosmos3, VAE, transformer, and guardrail modules.
- Added packed CPU/GPU storage, active-stage rebinding, pinned-memory support, configuration validation, cleanup, and CUDA-graph incompatibility checks.
- Added device-aware configuration and pipeline-loader handling for CPU-offloaded modules.
- Added optional CUDA peak-memory logging.
- Added documentation and public `CpuOffloadConfig` exports.
- The offloading design must prevent inactive groups from retaining stale CUDA views. The shared GPU arena must bind only the active group.
- CI still reports failures. The LTX2 `_device` regression was fixed, but the remaining failures require confirmation.
- Multi-GPU support, quantized paths, latency validation, and additional configuration/error tests remain incomplete.

**Verdict: needs follow-up.**

## QA Engineer Review

Test code changed in:

- `tests/unittest/_torch/visual_gen/test_offloading.py`: Added unit coverage for configuration, staging, storage alignment, rebinding, inactive-stage access, Cosmos3 component discovery, VAE contexts, guardrails, and stage validation.
- `tests/unittest/_torch/visual_gen/test_wan22_t2v_offload.py`: Added Wan 2.2 offload execution, baseline comparison, FP8 comparison, and CUDA-graph incompatibility tests.
- `tests/unittest/_torch/visual_gen/test_wan22_t2v_pipeline.py`: Added offload pipeline setup and cosine-similarity coverage.
- `tests/unittest/_torch/visual_gen/test_cosmos3_t2v_offload.py`: Added Cosmos3 offload execution, output validation, baseline comparison, and CUDA-graph incompatibility tests.
- Existing Cosmos3, LTX2, and pipeline tests were updated for device and configuration changes.

Test-list changes:

- `tests/integration/test_lists/test-db/l0_b200.yml` adds `test_offloading.py` to pre-merge coverage.
- It adds two Wan 2.2 offload tests and one Cosmos3 offload test to post-merge coverage.

The added integration tests are listed in `test-db/l0_b200.yml`. CBTS coverage data is unavailable, and CI continues to report failures.

**Verdict: needs follow-up.**

## Performance results

Measured with `Wan-AI/Wan2.2-T2V-A14B-Diffusers` at 480x832 resolution and 33 frames.
- Without offloading: 73.96s total pipeline time, 68.54 GiB peak CUDA memory
- With CPU offloading: 74.28s total pipeline time, 31.62 GiB peak CUDA memory
- Offloading setup time: 15.29s

In this single-run measurement, CPU offloading reduced peak CUDA memory by 36.92 GiB, from 68.54 GiB to 31.62 GiB, which is about a 54% reduction. End-to-end pipeline time remained effectively flat, with only 0.32s additional runtime, or about 0.4% overhead, excluding the one-time offloading setup cost.

This enables Wan-AI/Wan2.2-T2V-A14B-Diffusers to run comfortably on L40 and A6000-class 48 GB GPUs. RTX 5090 support may be feasible with careful tuning, but 32 GB remains a marginal target due to allocator fragmentation and limited headroom at higher resolutions or frame counts.


NVShreyas:
Few initial thoughts - 

1. What is the main offloading method? From the code, I think its at model level. If so, the harness seems a bit overkill. Why can't we do model.to(device) before and after? 
2. the setup right now overrides the same GPU arena which could cause silent stale pointers if another group is accessed. Any code that touches a non-active group's parameter outside its own forward — cache extractor setup, state-dict export, CUDA-graph capture — silently sees or corrupts the wrong data.

tensorrt_llm/_torch/visual_gen/executor.py
NVShreyas: if we're adding this, we should add a flag to disable by default and only enable for development purposes.

NVShreyas: are syncs really needed?
From Claude - 

```
  Mostly no — and that's the kicker. The CUDA caching allocator's allocated_bytes and peak counters are host-side counters updated synchronously when torch.empty(...) is called, not when GPU kernels
   complete. By the time pipeline.infer(req) returns, every PyTorch allocation made during the request has already incremented the counter. There is no in-flight allocation that a sync would
  "settle."

  The defensive reasoning ("sync first so the reading is accurate") is folklore. It would only matter if:
  1. You allocate via raw cudaMallocAsync outside the caching allocator (rare; not the case here).
  2. You're reading mem_get_info() (driver query, reflects physical state) rather than memory_stats() / max_memory_allocated() (PyTorch counters).

  For max_memory_allocated() specifically: no sync needed for correctness. The PyTorch test suite reads it without sync routinely.
```

NVShreyas: should not all reduce here. I'd prefer to let each rank log its max individually to be consistent with other parts of the codebase

rahul-steiger-nv: udpated

rahul-steiger-nv: done

rahul-steiger-nv: fixed

rahul-steiger-nv:
@NVShreyas 
> 1. What is the main offloading method? From the code, I think its at model level. If so, the harness seems a bit overkill. Why can't we do model.to(device) before and after?

I agree that model.to(device) before/after use is simpler and is a good baseline. The reason I didn’t start there is that it has a few downsides for this use case:

- It copies each parameter/buffer separately, which can turn staging into many small H2D/D2H transfers instead of one large bandwidth-friendly transfer.
- Moving back with .to("cpu") copies GPU weights back to CPU, even though inference weights are immutable and we can keep CPU as the source of truth.
- It does not naturally give us a persistent packed pinned-memory CPU store. We can use pinned tensors in PyTorch, but once we want pinned CPU master weights plus large flattened H2D copies, we are already outside plain module-level .to().
- It may create more allocator churn because module parameters are repeatedly rebound to newly allocated device storage.
- Avoiding the D2H leg should cut the transfer work roughly in half for immutable weights, and packing the tensors should reduce copy launch overhead.

So the harness is intended to trade complexity for fewer transfers, pinned CPU backing storage, and larger contiguous copies. That said, I agree the design only makes sense if it avoids stale CUDA views and remains clearly scoped.

rahul-steiger-nv:
@NVShreyas 
> 2. the setup right now overrides the same GPU arena which could cause silent stale pointers if another group is accessed. Any code that touches a non-active group's parameter outside its own forward — cache extractor setup, state-dict export, CUDA-graph capture — silently sees or corrupts the wrong data.

This makes sense, I agree. The current invariant should be tightened so inactive/offloaded weights are either CPU-backed or fail loudly if accessed as CUDA weights. Silent stale CUDA views into a reused arena are not acceptable.

I’ll work on this next week. My current direction is to ensure only the active group is bound to the GPU arena, while inactive groups are rebound to CPU storage or otherwise guarded so accidental use errors out. Happy to take pointers on whether you’d prefer CPU-backed inactive params or an explicit sentinel/error tensors.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
NVShreyas: what is the plan to support WAN 2.2 2 transformers?

NVShreyas: why is this and _WAN_T2V_OFFLOAD_STAGES in pipeline needed? Can we just have 1 place to define whether transformer blocks should be offloaded?

tensorrt_llm/_torch/visual_gen/offloading.py
NVShreyas: the entry point used by callers doesn't have to be __call__ so it will not work for some models. Can we make it explicit instead using context manager? something like - 
```
with offload_context("transformer.blocks", enable=...):
  self.transformer(latents, t, emb)
```
since the torch.cuda.synchronize() would occur before the transformer.forward() call, this would also allow us to enable cuda graphs right?

NVShreyas: should we be raising a clearer error if this much memory isnt available?
Also have you checked `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` to avoid fragmentation?

NVShreyas: IIUC during initialization, there will be a short time when there is 2x memory allocated on the CPU before `_copy_group_to_cpu_storage` and `_bind_views`?

NVShreyas: do we need `_OffloadGroup`? Can we just use `dict[str, nn.Module]`?

NVShreyas: this method is doing 3 things - compute offsets, detect aliases, build specs. Can we refactor and simplify?

NVShreyas: can we break this up into 2 - `_rebind_to_cpu` and `_rebind_to_gpu`?

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
NVShreyas: why does the transformer have to know about whether offloading is enabled or not? Can this be moved to the pipeline?

tensorrt_llm/_torch/visual_gen/offloading.py
NVShreyas: can we merge both of these to make the code a bit simpler? I dont think we need frozen here necessarily?

NVShreyas (commented):
Overall the approach is good and I'm okay with not using module.to(device) based on the explanation you provided. Lets make sure the code is architected in a maintainable and readable manner (including docstrings). It also should be tested with all single GPU features like quantization, caching, torch compile, etc (or raise NotImplementedError explicitly). Let's leave multi-gpu offloading as a future enhancement for now.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
rahul-steiger-nv: I agree. That makes sense

rahul-steiger-nv: Added support

rahul-steiger-nv: Completely agree, removed

tensorrt_llm/_torch/visual_gen/offloading.py
rahul-steiger-nv: done

rahul-steiger-nv: good point, resolved

rahul-steiger-nv: agree

rahul-steiger-nv: Yes, that’s right. There is a short period where we keep both the original CPU weights and the packed CPU offload storage.

I kept it this way because it’s much safer for quantization: some quant flows can replace or repack parameters during post_load_weights(), so wiring offload storage directly into loading would be a bigger change.

rahul-steiger-nv: updated

rahul-steiger-nv: Good idea, I think this approach makes more sense.

rahul-steiger-nv: updated

rahul-steiger-nv:
@coderabbitai help

rahul-steiger-nv:
@coderabbitai review

rahul-steiger-nv:
@coderabbitai full review

rahul-steiger-nv:
@coderabbitai run pre-merge checks

examples/visual_gen/README.md
NVShreyas: can we remove this flag from prod examples? it should only be used as a debug tool

examples/visual_gen/visual_gen_wan_t2v.py
NVShreyas: could you also add short feature documentation in the visual-gen docs? 
https://github.com/NVIDIA/TensorRT-LLM/blob/main/docs/source/models/visual-generation.md

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
NVShreyas: there is also self.is_wan22_5b, would this branching work for it?

NVShreyas: is this method needed for each pipeline if the base pipeline has knowledge of the `default_offload_stages`?

tensorrt_llm/_torch/visual_gen/offloading.py
NVShreyas: is the method still needed? can we just use `self._rebind_to_cpu` instead?

NVShreyas: can we just use nn.Module instead?

NVShreyas, line 533: can you add a comment about why there's no exit action?

rahul-steiger-nv: added

rahul-steiger-nv: sure

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
rahul-steiger-nv: removed

rahul-steiger-nv: I made the path clearer

examples/visual_gen/visual_gen_wan_t2v.py
rahul-steiger-nv: done

examples/visual_gen/README.md
rahul-steiger-nv: good catch!

docs/source/models/visual-generation.md
NVShreyas: we wanna keep this doc lean so just 1 line here similar to teacache would work. Detailed should be found in class docstrings (feature-specific docs in the future).
Also add to the feature matrix above ^

rahul-steiger-nv:
@coderabbitai full review

examples/visual_gen/configs/wan-t2v-cpu-offload.yaml
NVShreyas: rename the file wan-2.2 so this comment can be removed and its clearer

examples/visual_gen/visual_gen_wan_t2v.py
NVShreyas: can we remove this and corresponding methods, and just use a list of comma separated strings for the offload stages?

NVShreyas: can we raise some errors here for unsupported combinations like quantization, cuda graphs, torch compile etc. This makes it easier for users to see in the example script even if it raises error inside the visual gen executor

tensorrt_llm/_torch/visual_gen/pipeline.py
NVShreyas: why another method for this if its just called in 1 place?

NVShreyas: similarly for this

NVShreyas: correct me if I'm wrong, arent `offload_pipeline_parts`  and this doing the same thing?

tensorrt_llm/_torch/visual_gen/offloading.py
NVShreyas: nit: raise error instead of asserts

tensorrt_llm/_torch/visual_gen/pipeline.py
NVShreyas: CC flagged this, please confirm - 
```
BasePipeline.device delegates to self.transformer.device, which calls get_parameter_device(transformer) →
  next(transformer.parameters()).device. After initialize_offload_pipeline(), transformer.blocks params are CPU-backed.
  If blocks happens to be the first submodule registered in WanTransformer3DModel.__init__, then
  next(transformer.parameters()).device returns CPU — and every subsequent tensor.to(self.device) call puts tensors on
  CPU instead of GPU. Silent wrong results or shape-mismatch crashes.

  # _encode_prompt:
  input_ids = text_inputs.input_ids.to(self.device)   # self.device could be CPU here
  attention_mask = text_inputs.attention_mask.to(self.device)

  This works today only because of constructor ordering in WanTransformer3DModel.__init__ — if non-block params
  (condition_embedder, patch_embed, etc.) are registered before blocks, the first param is on GPU. One refactor of the
  transformer constructor order silently breaks this.
```

rahul-steiger-nv:
@coderabbitai full review

tensorrt_llm/_torch/modules/linear.py
```diff
@@ -805,11 +805,14 @@ def rescale_fused_weights(self, module: Linear):
...
         shard_key_to_index = weight_mode.shard_key_to_index
+        compute_device = module.weight.device
+        if compute_device.type == "cpu" and torch.cuda.is_available():
+            compute_device = torch.device("cuda")
 
         # Handle input_scale
         if hasattr(module, "has_static_input_scale"):
             # Compute max and replace input_scale with a new parameter
-            max_input_scale = module.tmp_input_scales.max()
+            max_input_scale = module.tmp_input_scales.to(compute_device).max()
```
NVShreyas, line 908: why are the changes in this file needed?

tensorrt_llm/_torch/visual_gen/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -51,9 +53,30 @@
...
+    (COSMOS3_GENERATOR_OFFLOAD_PART,),
+)
+
 
 # TODO: add hf_ids
-@register_pipeline("Cosmos3OmniMoTPipeline")
+@register_pipeline(
+    "Cosmos3OmniMoTPipeline",
+    defaults={
+        "enable_cuda_memory_logging": False,
```
NVShreyas: add this arg to config so it can be used without specifying it here similar to `enable_layerwise_nvtx_marker`

tensorrt_llm/_torch/visual_gen/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -62,6 +85,86 @@ def _init_transformer(self) -> None:
...
+                modules = [m for m in runner.models if isinstance(m, nn.Module)]
+                if modules:
+                    parts[part_name] = nn.ModuleList(modules)
+
+        return parts
+
+    def _tower_offload_context(self, part_name: str) -> AbstractContextManager:
+        """Return an offload context for ``part_name`` if it is configured."""
+        return self.offload_context(part_name, enable=part_name in self.requested_offload_parts())
+
```
NVShreyas: seems like a generic method that can be used for all pipelines?

tests/integration/defs/examples/test_visual_gen.py
```diff
@@ -1207,20 +1283,14 @@ def test_wan_t2v_example(_visual_gen_deps, llm_root, llm_venv):
     output_path = os.path.join(out_dir, "wan_t2v_output.mp4")
 
     script_path = os.path.join(llm_root, "examples", "visual_gen", "models", "wan_t2v.py")
-    config_path = os.path.join(
-        llm_root, "examples", "visual_gen", "configs", "wan2.2-t2v-fp4-1gpu.yaml"
-    )
     assert os.path.isfile(script_path), f"Example script not found: {script_path}"
-    assert os.path.isfile(config_path), f"Config not found: {config_path}"
```
NVShreyas: why these changes?

tests/integration/defs/examples/test_visual_gen.py
```diff
@@ -83,6 +83,19 @@
 WAN22_LPIPS_SEED = 42
 WAN22_LPIPS_FRAME_RATE = 16.0
 
+COSMOS3_LPIPS_PROMPT = "A cute cat playing piano."
```
NVShreyas: lets revert changes in this file. cosmos3 tests will be skipped since the model isnt there yet and wan e2e unit tests are enough

tests/unittest/_torch/visual_gen/test_wan21_t2v_offload.py
```diff
@@ -0,0 +1,284 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
NVShreyas: having both wan2.1 and wan2.2 is unnecessary. lets just keep wan2.2 tests

tensorrt_llm/_torch/visual_gen/pipeline.py
```diff
@@ -396,6 +407,165 @@ def post_load_weights(self) -> None:
...
+        stages: list[str | list[str]],
+    ) -> tuple[OffloadPipelineStage, ...]:
+        """Normalize user-configured offload stages to tuple form."""
+        normalized_stages: list[OffloadPipelineStage] = []
+        for stage in stages:
+            parts = (stage,) if isinstance(stage, str) else tuple(stage)
+            if not parts:
+                raise ValueError("Offload stages must contain at least one part")
+            normalized_stages.append(parts)
+        return tuple(normalized_stages)
```
NVShreyas: can you simplify the offload helper methods in BasePipeline if possible and move it to offloading.py? BasePipeline can have required methods like `initialize_offload_pipeline` and `offload_pipeline_parts` only.
Or maybe BasePipeline can have only `initialize_offload_pipeline` and `offload_stages`, and offload pipeline factory can take pipeline as input and create an Offloadpipeline object. Upto you, basically trying to keep BasePipeline clean

tensorrt_llm/_torch/visual_gen/pipeline.py
```diff
@@ -396,6 +407,165 @@ def post_load_weights(self) -> None:
...
+        if text_encoder is not None:
+            parts[PipelineComponent.TEXT_ENCODER.value] = text_encoder
+
+        vae = getattr(self, PipelineComponent.VAE.value, None)
+        if vae is not None:
+            parts[PipelineComponent.VAE.value] = vae
+
+        for component_name in self.transformer_components:
+            transformer = getattr(self, component_name, None)
+            blocks = getattr(transformer, "blocks", None) if transformer is not None else None
```
NVShreyas, line 472: not sure if this is the best way to detect modules since models can have different names like cosmos3. Should we force every model pipeline to set it up themselves? This is non-blocking now but may need to be re-designed in the future if we need to add offloading to more models.
Maybe just add a comment for now

tensorrt_llm/_torch/modules/linear.py
rahul-steiger-nv: These .to(compute_device) additions ensure that when weights are offloaded to CPU, the FP8 re-quantization is still done on the GPU and copied back, so offloaded and non-offloaded models produce numerically identical quantized weights.

tensorrt_llm/_torch/visual_gen/models/cosmos3/pipeline_cosmos3.py
rahul-steiger-nv: done

tensorrt_llm/_torch/visual_gen/models/cosmos3/pipeline_cosmos3.py
rahul-steiger-nv: I agree

tensorrt_llm/_torch/visual_gen/pipeline.py
rahul-steiger-nv: I chose to implement an additional PipelineOffloader class.

tensorrt_llm/_torch/visual_gen/pipeline.py
rahul-steiger-nv: I left a comment

tests/integration/defs/examples/test_visual_gen.py
rahul-steiger-nv: removed

tests/unittest/_torch/visual_gen/test_wan21_t2v_offload.py
rahul-steiger-nv: sounds good!

tests/integration/defs/examples/test_visual_gen.py
rahul-steiger-nv: undid the changes

tests/unittest/_torch/visual_gen/test_cosmos3_t2v_offload.py
```diff
@@ -0,0 +1,355 @@
...
+    COSMOS3_VIDEO_GUARDRAIL_OFFLOAD_PART,
+    Cosmos3OmniMoTPipeline,
+)
+
+# ============================================================================
+# Fast unit tests for the Cosmos3 offload wiring (no GPU / no checkpoint)
+# ============================================================================
+
+
+class _FakeRunner:
```
NVShreyas: I have added Cosmos3-Nano to TRTLLM CI. Could you setup the test using the real model now? Path is just LLM_MODELS_ROOT/Cosmos3-Nano

rahul-steiger-nv: Rebased and updated tests

docs/source/helper.py
```diff
@@ -74,7 +74,6 @@ def generate_examples():
         'visual_gen_flux.py',
         'visual_gen_ltx2.py',
         'visual_gen_wan_i2v.py',
-        'visual_gen_wan_t2v.py',
```
NVShreyas, line 76: is this rebase issue?

docs/source/models/visual-generation.md
```diff
@@ -169,6 +174,17 @@ Configured under `VisualGenArgs.parallel_config`. Modes can be combined:
...
+
+### Troubleshooting
+
+**Out of memory:** enable quantization (`quant_config`), reduce resolution or `num_frames`, enable TeaCache (`cache_config`), enable CPU offloading (Wan T2V), or shard the model across more GPUs with Ulysses or Attention2D parallelism.
+
+**Parallelism constraints:**
+
+- Total GPUs must equal `cfg_size × ulysses_size` (Ulysses) or `cfg_size × attn2d_row_size × attn2d_col_size` (Attention2D).
+- Ulysses requires the model's attention-head count to be divisible by `ulysses_size`, and the sequence length to be divisible by `ulysses_size`.
+- Attention2D requires an LSE-capable backend (`FA4` or `CUTEDSL`), the sequence length to be divisible by `attn2d_row_size × attn2d_col_size`, and cannot be combined with Ulysses.
```
NVShreyas: rebase issue?

tensorrt_llm/visual_gen/args.py
```diff
@@ -400,6 +400,37 @@ class CudaGraphConfig(StrictBaseModel):
     enable: bool = Field(False, status="prototype")
 
 
+class OffloadConfig(StrictBaseModel):
```
NVShreyas: @zhenhuaw-me could you review this?

tensorrt_llm/visual_gen/args.py
```diff
@@ -524,6 +559,12 @@ class VisualGenArgs(StrictBaseModel):
         ),
     )
 
+    enable_cuda_memory_logging: bool = Field(
```
zhenhuaw-me: Does it make sense to make this as a env var in short term. For reference, the nsys trace knob is an env var.

tensorrt_llm/visual_gen/args.py
```diff
@@ -400,6 +400,37 @@ class CudaGraphConfig(StrictBaseModel):
     enable: bool = Field(False, status="prototype")
 
 
+class OffloadConfig(StrictBaseModel):
+    """Configuration for visual generation module offloading."""
```
zhenhuaw-me: Can we add more context about "offloading"?

tensorrt_llm/visual_gen/args.py
zhenhuaw-me: Can we name this config as `CpuOffloadConfig`? I wonder if `Offload` is not specific enough in some context (e.g., vs KVCacheOffloag).

tensorrt_llm/visual_gen/args.py
```diff
@@ -400,6 +400,37 @@ class CudaGraphConfig(StrictBaseModel):
...
+        "cpu",
+        status="prototype",
+        description="Device where offloaded modules are stored between active stages.",
+    )
+    pin_memory: bool = Field(
+        True,
+        status="prototype",
+        description="Allocate pinned CPU storage for faster host-to-device copies.",
+    )
+    stages: Optional[List[Union[str, List[str]]]] = Field(
```
zhenhuaw-me, line 516: In this context, does _staged_ means _offloaded_? I see _stages_/_components_, can we try to align the terms?

zhenhuaw-me (commented):
A few comments on the APIs. Thanks!

tensorrt_llm/visual_gen/args.py
rahul-steiger-nv: moved it to an env var TLLM_VISUAL_GEN_LOG_CUDA_MEMORY (defaults off), matching the TLLM_PROFILE_VISUAL_GEN_START_STOP nsys knob, and dropped the VisualGenArgs field since it's a dev-only knob.

tensorrt_llm/visual_gen/args.py
rahul-steiger-nv: added it

tensorrt_llm/visual_gen/args.py
rahul-steiger-nv: That makes sense

tensorrt_llm/visual_gen/args.py
rahul-steiger-nv: Renamed it to CpuOffloadConfig and aligned the terminology throughout so "component" = a named sub-model and "stage" = one or more co-resident components.

NVShreyas (approved):

examples/visual_gen/configs/wan2.2-t2v-cpu-offload.yaml
```diff
@@ -0,0 +1,34 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
```
zhenhuaw-me: Configs are supposed to manage the "best show case config". For such true/false, knob, prefer to leave it to API.

examples/visual_gen/models/wan_t2v.py
```diff
@@ -18,13 +18,24 @@
 Usage:
     python wan_t2v.py
     python wan_t2v.py --visual_gen_args ../configs/wan2.2-t2v-fp4-1gpu.yaml
+    python wan_t2v.py --enable_offloading \
+        --offload_stages text_encoder,denoising_transformer,denoising_transformer_2,vae
```
zhenhuaw-me: We don't add additional CLI interfaces to these modeling examples

tensorrt_llm/visual_gen/args.py
```diff
@@ -418,6 +418,58 @@ class CudaGraphConfig(StrictBaseModel):
...
+
+    enable: bool = Field(
+        False,
+        status="prototype",
+        description="Enable offloading of model components to CPU between pipeline stages.",
+    )
+    device: Literal["cpu", "cuda"] = Field(
+        "cpu",
+        status="prototype",
+        description="Device where components are held while not the active stage on the GPU.",
```
zhenhuaw-me: What does it mean if we assign `cuda` here?

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
```diff
@@ -78,6 +78,16 @@
...
 
+_WAN_SINGLE_TRANSFORMER_OFFLOAD_STAGES = (
+    ("text_encoder",),
+    ("denoising_transformer",),
+)
+_WAN_TWO_TRANSFORMER_OFFLOAD_STAGES = (
+    ("text_encoder",),
+    ("denoising_transformer",),
+    ("denoising_transformer_2",),
+)
```
zhenhuaw-me, line 109: We should align the naming with model_index.json

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
```diff
@@ -537,13 +579,15 @@ def forward_fn(
             current_t = timestep if timestep.dim() == 0 else timestep[0]
 
             # Select model based on timestep (if two-stage denoising is enabled)
+            offload_stage = "denoising_transformer"
```
zhenhuaw-me: I don't get the motivation of this change.

zhenhuaw-me (changes requested):

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
rahul-steiger-nv: aligned

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
rahul-steiger-nv: For Wan 2.2 A14B, denoising switches between `transformer` and `transformer_2` based on the timestep. When CPU offload is enabled, the offloader needs the matching component name before the forward call; otherwise it may stage the wrong transformer.

tensorrt_llm/visual_gen/args.py
rahul-steiger-nv: removed device entirely now

examples/visual_gen/models/wan_t2v.py
rahul-steiger-nv: removed

examples/visual_gen/configs/wan2.2-t2v-cpu-offload.yaml
rahul-steiger-nv: removed

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
zhenhuaw-me: If I didn't misread, the components are configurable through APIs, why need to assign the `transformer(_2)` here?

docs/source/models/visual-generation.md
```diff
@@ -226,6 +227,10 @@ args = VisualGenArgs(
 
 **Wan 2.2 dual-transformer note:** Wan 2.2 uses two expert transformers (high-noise and low-noise stacks). All `CacheDiTConfig` parameters apply to both stacks, except `max_warmup_steps` and `max_cached_steps`: the low-noise stack always uses fixed internal caps (`max_warmup_steps=2`, `max_cached_steps=20`) regardless of user config.
 
+### CPU Offloading (Wan T2V only)
+
+CPU offloading stages move selected Wan T2V pipeline components between CPU and GPU to reduce peak GPU memory usage; enable it with `cpu_offload_config.enable: true` or `--enable_offloading` in the example script.
```
zhenhuaw-me:
```suggestion
CPU offloading stages move selected Wan T2V pipeline components between CPU and GPU to reduce peak GPU memory usage; enable it with `cpu_offload_config.enable: true`
```

examples/visual_gen/models/wan_t2v.py
```diff
@@ -49,8 +49,12 @@ def main():
     args = parser.parse_args()
 
     # Engine config from shared YAML (optional); model-specific defaults apply otherwise.
-    extra_args = VisualGenArgs.from_yaml(args.visual_gen_args) if args.visual_gen_args else None
-    visual_gen = VisualGen(model=args.model, args=extra_args)
+    if args.visual_gen_args:
```
zhenhuaw-me: Why this change? By design, it should be safe to ignore `args` if not provided

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
rahul-steiger-nv: The configurable API defines the offload schedule, but the runtime still has to tell the offloader which configured component is about to execute.

For Wan 2.2 A14B, `forward_fn` dynamically chooses between `transformer` and `transformer_2` based on the current denoising timestep. When CPU offload is enabled, only one stage is resident on GPU at a time, so before calling `current_model(...)` we need to stage the matching component. This variable is that bridge: it keeps the selected module and its public component name paired.

tensorrt_llm/visual_gen/args.py
```diff
@@ -433,6 +433,53 @@ class CudaGraphConfig(StrictBaseModel):
...
+    )
+    stages: Optional[List[Union[str, List[str]]]] = Field(
+        default=None,
+        status="prototype",
+        description=(
+            "Optional ordered list of stages, named with model-specific public component "
+            "names. Each entry is a single component, or a list of components that are "
+            "co-resident on the GPU and run together as one stage, e.g. "
+            "['text_encoder', 'transformer', 'vae'] for Wan T2V, or "
+            "['reasoner', 'generator', 'text_guardrail', 'video_guardrail', 'vae'] for Cosmos3. "
```
zhenhuaw-me, line 526: Can we have an example about the "co-resident components"?

rahul-steiger-nv: added it

examples/visual_gen/models/wan_t2v.py
rahul-steiger-nv: I agree, I reverted the change

docs/source/models/visual-generation.md
```diff
@@ -43,15 +44,15 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
...
-| **Wan 2.2** | Yes | Yes | Yes [^2] | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
-| **LTX-2** | Yes | Yes | Yes [^3] | Yes | Yes | Yes | No | No | Yes | Yes | Yes | Yes | No |
-| **Qwen-Image** [^4] | Yes | Yes | No | No | No | Yes | No | Yes | Yes | Yes | Yes | Yes | No |
-| **Cosmos3** | Yes | Yes | No | No | Yes | Yes | Yes | Yes | Yes | Yes | No | No | Yes |
+| Model | FP8 blockwise | NVFP4 | TeaCache | Cache-DiT | CPU Offloading | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve | Attention2D | Ring Attention | Tensor Parallelism |
+|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
+| **FLUX.1** | Yes | Yes | Yes | Yes | No | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes |
+| **FLUX.2** | Yes | Yes | Yes | Yes | No | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes |
+| **Wan 2.1** | Yes | Yes | Yes | Yes | T2V only | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
+| **Wan 2.2** | Yes | Yes | Yes [^2] | Yes | T2V only | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes |
```
zhenhuaw-me: Yes / No with footnote.

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
```diff
@@ -235,19 +256,29 @@ def load_standard_components(
 
         if PipelineComponent.TEXT_ENCODER not in skip_components:
             logger.info("Loading text encoder...")
+            text_encoder_device = (
+                torch.device("cpu")
+                if PipelineComponent.TEXT_ENCODER.value in self.offloader.requested_components()
```
zhenhuaw-me, line 281: Out of the scope of this PR.

This _matching_ between the "TRTLLM defined component name value" with the "user provided component string" is somehow strange - I think we claim that the "components" are the ones defined in `model_index.json` but "PipelineComponent.TEXT_ENCODER.value matches the  `model_index.json`" is an assumption and I wonder it may break someday.

I think we need to find a solution for the `PipelineComponent` or we just leave to the pipeline to handle.

docs/source/models/visual-generation.md
rahul-steiger-nv: Updated

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
rahul-steiger-nv: Agreed. This is out of scope for this PR.

zhenhuaw-me:
@rahul-steiger-nv Can you rebase the PR? I see some conflicts

rahul-steiger-nv:
> @rahul-steiger-nv Can you rebase the PR? I see some conflicts

@zhenhuaw-me Done

docs/source/models/visual-generation.md
```diff
@@ -18,6 +18,7 @@ TensorRT-LLM **VisualGen** provides a unified inference stack for diffusion mode
 - Sparse attention support: see [VisualGen Sparse Attention](../visual-gen/features/sparse-attention.md).
 - Multi-GPU parallelism (CFG parallel, Ulysses sequence parallel, Tensor parallelism).
 - **Step caching** — two runtime caching backends (**TeaCache** and **Cache-DiT**) that skip transformer computation on steps where the step-to-step change is small.
+- CPU offloading for Wan text-to-video pipelines to reduce peak GPU memory usage.
```
zhenhuaw-me:
```suggestion
- CPU offloading to reduce peak GPU memory usage.
```

tests/integration/test_lists/test-db/l0_b200.yml
```diff
@@ -245,6 +245,13 @@ l0_b200:
   - unittest/_torch/visual_gen/test_wan21_t2v_teacache_user_coefficients.py
   - unittest/_torch/visual_gen/test_wan22_i2v_teacache.py
   - unittest/_torch/visual_gen/test_wan22_t2v_teacache.py
+  - unittest/_torch/visual_gen/test_offloading.py
+  # Heavyweight offload accuracy tests (double pipeline load) run in post_merge;
+  # keep only lightweight smoke + error-path coverage in pre_merge.
+  - unittest/_torch/visual_gen/test_wan22_t2v_offload.py::TestWan22_A14B_Offload::test_wan22_offload
+  - unittest/_torch/visual_gen/test_wan22_t2v_offload.py::TestWanT2VOffloadCudaGraphRaisesError::test_wan22_raises_if_cuda_graph_and_offload_enabled
+  - unittest/_torch/visual_gen/test_cosmos3_t2v_offload.py::TestCosmos3Offload::test_cosmos3_offload_forward
+  - unittest/_torch/visual_gen/test_cosmos3_t2v_offload.py::TestCosmos3OffloadCudaGraphRaisesError::test_cosmos3_raises_if_cuda_graph_and_offload_enabled
```
zhenhuaw-me:
```suggestion
```

Reflect the comment?

tests/integration/test_lists/test-db/l0_b200.yml
```diff
@@ -356,6 +363,11 @@ l0_b200:
   - examples/visual_gen/test_visual_gen.py::test_qwenimage_lpips_against_golden TIMEOUT (10)
   - examples/visual_gen/test_visual_gen.py::test_cosmos3_nano_t2i_lpips_against_golden TIMEOUT (10)
   - examples/visual_gen/test_visual_gen.py::test_cosmos3_nano_t2v_lpips_against_golden TIMEOUT (15)
+  # Heavyweight CPU-offload accuracy tests (each loads the full pipeline twice:
+  # offload + baseline). Kept in post_merge to bound the B200 pre_merge wall-time.
+  - unittest/_torch/visual_gen/test_wan22_t2v_offload.py::TestWan22_A14B_Offload::test_wan22_offload_matches_baseline
```
zhenhuaw-me, line 431: How long do it take to run these tests?

rahul-steiger-nv: On a GH200 system, test_wan22_offload_matches_baseline takes roughly 4 minutes — it loads the full Wan2.2-A14B two-stage pipeline twice (offload + baseline). test_cosmos3_nano_t2v_lpips_against_golden is around 5 minutes (per the inline note above its entry). Both are dominated by model loading rather than compute, which is why they sit in post_merge.

docs/source/models/visual-generation.md
```diff
@@ -65,6 +66,8 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
 
 [^5]: Qwen-Image ships a native BF16 implementation with per-module numerical parity against `diffusers.QwenImagePipeline` (cosine similarity >= 0.999 on the full 20B transformer) and supports `trtllm-serve` / `/v1/images/generations`. VisualGen supports FP8 blockwise and NVFP4 dynamic quantization from BF16 checkpoints, as well as direct loading of statically quantized FP8 and NVFP4 ModelOpt checkpoints.
 
+[^6]: CPU offloading is currently validated for Wan and Cosmos3 text-to-video pipelines.
```
zhenhuaw-me:
```suggestion
[^6]: CPU offloading of Cosmos3 is currently validated the text-to-video pipelines.
```

docs/source/models/visual-generation.md
```diff
@@ -44,16 +45,16 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
...
-| **LTX-2** | Yes | Yes | Yes [^4] | Yes | Yes | Yes | No | No | Yes | Yes | Yes | Yes | No | No |
-| **Qwen-Image** [^5] | Yes | Yes | No | No | No | Yes | No | Yes | Yes | Yes | Yes | Yes | No | No |
-| **Cosmos3** | Yes | Yes | No | No | Yes | Yes | Yes | Yes | Yes | Yes | No | No | Yes | No |
+| Model | FP8 blockwise | NVFP4 | TeaCache | Cache-DiT | CPU Offloading | CFG Parallelism | Ulysses Parallelism | Parallel VAE | CUDA Graph | torch.compile | trtllm-serve | Attention2D | Ring Attention | Tensor Parallelism | VSA |
+|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
+| **FLUX.1** | Yes | Yes | Yes | Yes | No | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes | No |
+| **FLUX.2** | Yes | Yes | Yes | Yes | No | No [^1] | Yes | No | Yes | Yes | Yes | Yes | Yes | Yes | No |
+| **Wan 2.1** | Yes | Yes | Yes | Yes | Yes [^6] | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | No |
+| **Wan 2.1 VSA** [^2] | Yes | Yes | Yes | Yes | Yes [^6] | Yes | Yes | Yes | Yes | Yes | Yes | No | No | Yes | Yes |
+| **Wan 2.2** | Yes | Yes | Yes [^3] | Yes | Yes [^6] | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | No |
```
zhenhuaw-me:
```suggestion
| **Wan 2.1** | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | No |
| **Wan 2.1 VSA** [^2] | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | No | No | Yes | Yes |
| **Wan 2.2** | Yes | Yes | Yes [^3] | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | Yes | No |
```

docs/source/models/visual-generation.md
```diff
@@ -267,6 +270,9 @@ attention_config:
     vsa_sparsity: 0.90
 ```
 
+### CPU Offloading (Wan and Cosmos3 T2V)
```
zhenhuaw-me:
```suggestion
### CPU Offloading
```

zhenhuaw-me (approved):

tensorrt_llm/_torch/visual_gen/models/cosmos3/transformer_cosmos3.py
```diff
@@ -1121,14 +1136,6 @@ def forward(
                     timestep=timestep,
                     real_text_lens=real_text_lens,
```
fredricz-20070104: **Possible unintended behavior change on the sharded (multi-GPU) Cosmos3 path.**

To wrap the `gen_layers` loop in `with offload_context("generator")`, the re-indentation here collapsed the original `if not self.sharder.is_active: ... else: ...` into a single branch. As a result the `layer(...)` call is now **always** invoked with `real_text_lens=real_text_lens`, whereas the previous `else` branch (i.e. when `self.sharder.is_active`, the multi-GPU sequence-parallel path) called `layer(...)` **without** `real_text_lens`.

- Limited blast radius: `real_text_lens` only changes attention when `batch_size > 1` (it slices K/V per-sample by the real text length to mask padding). For `batch_size == 1` the two paths are equivalent, so there is no change.
- The only affected configuration is therefore **multi-GPU + batch > 1** Cosmos3 — which is exactly the path this PR states is not yet validated and is not covered by the added tests. The new behavior looks more correct (no longer attending to padding), but it is a semantic change that slipped into the offload refactor rather than an intentional fix.

Could you confirm whether always passing `real_text_lens` on the `sharder.is_active` path is intentional, or a side effect of the re-indent? If intentional, it would be worth calling out (and ideally covering) separately from the offloading change.

fredricz-20070104 (approved):
Please check the commented issue I have raised before merging. Thanks.

tensorrt_llm/_torch/visual_gen/models/cosmos3/transformer_cosmos3.py
rahul-steiger-nv: good catch - fixed

mzweilz (approved):

mikeiovine (approved):
Stamp on behalf of runtime-devs

chang-l (approved):

LarryXFly (approved):

arysef (approved):

ishovkun:
Maybe already obsolete, but:
One of the failures IS this PR's fault

`test_ltx2_pipeline.py::TestLTX2TwoStageLoRAHelpers::test_run_warmup_precaptures_both_topologies` is a genuine regression. Verified empirically, not inferred:
```
has _device attr?  False
AttributeError: 'Dummy' object has no attribute 'device'
```
The chain:
1. The test builds the pipeline via `object.__new__` (line 1283), deliberately skipping `__init__`
2. It calls `_run_warmup`, which at `pipeline_ltx2.py:682` does `torch.zeros(..., device=self.device)`
3. This PR changed `BasePipeline.device` from `return self.transformer.device` → `return self._device`
4. `_device` is assigned **only** in `BasePipeline.__init__` (`pipeline.py:129`) — which never ran

A/B confirms it:

| | result |
|---|---|
| old impl `self.transformer.device` | `cuda` → **passes** |
| new impl, `_device` unset | **AttributeError** |
| new impl, `_device` set | `cuda` → passes |

Fix

The semantic change is *correct* — under offloading the transformer can sit on CPU, so `self.transformer.device` would report the wrong device. So the test should adapt, not the property revert. One line in `test_ltx2_pipeline.py`, mirroring what `__init__` does:

```python
pipeline = object.__new__(ltx2_two_stages.LTX2TwoStagesPipeline)
pipeline._device = torch.device("cuda")   # __init__ is bypassed; device property reads this

**Comment by chang-l** [comment]
@rahul-steiger-nv Please address the deterministic LTX2 unit-test regression identified in https://github.com/NVIDIA/TensorRT-LLM/pull/14095#issuecomment-5145647362 before another CI retry. The PR correctly changes `BasePipeline.device` to use `_device`, but `test_run_warmup_precaptures_both_topologies` constructs the pipeline with `object.__new__` and therefore must initialize `pipeline._device = torch.device("cuda")`. L0 build 51172 reproduced the `test_ltx2_pipeline.py` wrapper failure on current head `7617d4e`. A rerun is already in progress, but I have paused automated babysitting to prevent further retries until a fix is pushed. Please ping us after updating the PR so monitoring can resume.

**Comment by rahul-steiger-nv** [comment]
@chang-l @ishovkun Thank you for reviewing the PR. I’ve addressed the requested changes and will rerun CI/CD.

File: docs/source/models/visual-generation.md [hunk 33]
```diff
@@ -82,6 +83,8 @@ Models are auto-detected from the checkpoint directory. Diffusers-format models
 
 [^7]: `FastVideo/FastWan2.2-TI2V-5B-FullAttn-Diffusers` — a distilled version of Wan2.2-TI2V-5B with 3 denoising steps. CFG parallelism, TeaCache, and Cache-DiT are not applicable.
 
+[^8]: CPU offloading is currently validated for Wan and Cosmos3 text-to-video pipelines.
+
```
zhenhuaw-me: No need to add such footnote since it's already shown in the matrix.

Should be fine to use `/bot reuse` if the current CI and pushed any new doc commits.

rahul-steiger-nv: done

PR #14010, opened 2026-05-11
[TRTLLM-12533][refactor] Move Media IO modality loading into MediaIO Interfaces

aswinvisva:
## Release Notes

* **Improvements**
  * Standardized media loading across images, audio, and video for more consistent behavior.
  * Enhanced support for multiple input formats including bytes, base64-encoded data, URLs, and local file paths.
  * Improved reliability and error handling in multimodal content processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14010)

## Description

This [PR](https://github.com/NVIDIA/TensorRT-LLM/pull/13779) created the `MediaIO` interface for video, image and audio. It added just the modality-specific logic for merging media IO kwargs at the request level and the server level. This PR moves some more modality-specific logic for loading IO data (URLs, data, local file) and postprocessing the IO data into the `MediaIO` interface.

## Test Coverage

Tests added to the test_chat_utils.py which covers IO loading, updated the tests to load with the new MediaIO functions.

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [X] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

To see a list of available CI bot commands, please comment `/bot help`.

PR #14115, opened 2026-05-14
[#14100][fix] audio_url decodes data URI base64 like image_url

### LINKED ISSUE #14100: [Bug] trtllm-serve /v1/chat/completions: audio_url with data: URI base64 fails with "Supplied filename too long", only image_url decodes data: scheme
### System Info

- **TensorRT-LLM version:** `1.2.0rc6`
- **Container image:** `nvcr.io/nvidia/tensorrt-llm/release:1.2.0rc6`
- **Model:** `nvidia/Phi-4-multimodal-instruct-NVFP4` (FP4 weights from NVIDIA)
- **Endpoint:** `POST /v1/chat/completions`
- **GPU:** NVIDIA RTX 5080 (Blackwell, sm_120), single GPU
- **CUDA / Driver:** as shipped in the official image
- **Host OS:** Windows 11 with Docker Desktop on WSL2 (container is Linux)
- **Client:** `curl` from the host, also reproduced from Python `httpx`

### Who can help?

_No response_

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Start the server (single-GPU, default settings, no source patches):

```bash
docker run --gpus all --rm -p 8000:8000 --ipc=host \
  -e HF_HOME=/root/.cache/huggingface \
  -v $HOME/.cache/huggingface:/root/.cache/huggingface \
  nvcr.io/nvidia/tensorrt-llm/release:1.2.0rc6 \
  trtllm-serve nvidia/Phi-4-multimodal-instruct-NVFP4 \
    --host 0.0.0.0 --port 8000 \
    --trust_remote_code \
    --max_seq_len 8192 \
    --kv_cache_free_gpu_memory_fraction 0.5
```

Wait for the server to finish loading the model (a few minutes for Phi-4-MM the first time — log will show `INFO: Started server process` and `Application startup complete.`). Then send a chat-completions request with `audio_url` carrying an inline `data:audio/wav;base64,...` URI (here a ~16 s mono 16 kHz PCM WAV, ~500 KB → ~670 KB after base64):

```bash
B64=$(base64 -w 0 sample.wav)
curl -sS -X POST http://localhost:8000/v1/chat/completions \
  -d "{
    \"model\": \"nvidia/Phi-4-multimodal-instruct-NVFP4\",
    \"messages\": [{
      \"role\": \"user\",
      \"content\": [
        {\"type\": \"audio_url\", \"audio_url\": {\"url\": \"data:audio/wav;base64,${B64}\"}}
      ]
    }],
    \"max_tokens\": 512,
    \"temperature\": 0
  }"
```

The same request shape with `data:image/...;base64,...` works fine for image_url on other multimodal models.

### Expected behavior

`audio_url` with a `data:audio/<mime>;base64,...` URI should be base64-decoded into raw audio bytes, then handed off to the model's audio processor — exactly mirroring how `image_url` already handles `data:image/<mime>;base64,...` URIs in chat-completions today.

A successful response should look like:

```json
{
  "id": "chatcmpl-...",
  "choices": [{
    "index": 0,
    "message": {"role": "assistant", "content": "<transcribed text>"},
    "finish_reason": "stop"
  }],
  "usage": {...}
}
```

### actual behavior

The request returns HTTP 400:

```json
{
  "object": "error",
  "message": "cannot unpack non-iterable NoneType object",
  "type": "BadRequestError",
  "param": null,
  "code": 400
}
```

The server log shows libsndfile complaining that the data URI is a too-long *filename*, followed by `audios=None` reaching the multimodal preprocessor:

```
QBhAHQAewCGAJMApADDANYA6QD9AB8B...': Error : Supplied filename too long.
[TRT-LLM] [W] Basic input processor failed: cannot unpack non-iterable NoneType object.
[TRT-LLM] [E] Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/serve/openai_server.py", line 574, in openai_chat
    promise = self.llm.generate_async(
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/llmapi/llm.py", line 457, in generate_async
    prompt_token_ids, extra_processed_inputs = input_processor_with_hash(
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/inputs/registry.py", line 738, in input_processor_wrapper
    return input_processor(inputs, sampling_params)
  File "/usr/local/lib/python3.12/dist-packages/tensorrt_llm/_torch/models/modeling_phi4mm.py", line 915, in __call__
    audio_inputs = self.processor.audio_processor(
  File "/root/.cache/huggingface/modules/transformers_modules/.../processing_phi4mm.py", line 355, in __call__
    for audio_data, sample_rate in audios:
TypeError: cannot unpack non-iterable NoneType object
```

So:
- libsndfile receives the full `data:audio/wav;base64,<thousands of chars>` string and tries to treat it as a filesystem path (POSIX `ENAMETOOLONG`, >255 bytes), returns nothing.
- The audio loader returns `None`.
- The Phi-4-MM processor then crashes when iterating `audios`.

The same workflow with `data:image/...;base64,...` on image_url works correctly on other multimodal models — only audio is broken.

### Root cause (from source inspection)

In `tensorrt_llm/inputs/utils.py`, `async_load_audio` only branches on three URL schemes:

```python
if parsed_url.scheme in ["http", "https"]:
    resp = _safe_request_get(audio, stream=False)
    audio = BytesIO(resp.content)
elif parsed_url.scheme in ("", "file"):
    audio = _normalize_file_uri(audio)
else:
    raise ValueError(f"Unsupported URL scheme: {parsed_url.scheme!r}")
```

For a `data:` URI, `urllib.parse.urlparse(...).scheme == "data"`. There is an `is_base64` parameter on the same loader that *would* base64-decode the body, but it is never set to `True` for `data:` URIs — the caller in `tensorrt_llm/serve/chat_utils.py:parse_chat_message_content_part` (audio_url branch) simply passes the URL through.

In practice the request seems to reach the soundfile/libav-backed loader before the URL-scheme check fires (or the scheme check is silently swallowed elsewhere), which is why the user-facing error is `Supplied filename too long` from libsndfile rather than `Unsupported URL scheme`.

By contrast, the image_url branch in the same chat-completions parser auto-detects `data:image/...` and decodes it (the standard `image_b64` path documented in `examples/openai_chat_client_for_multimodal`). The asymmetry between image_url (works) and audio_url (silently fails) is the bug.

### Suggested fix

In the `audio_url` branch of `parse_chat_message_content_part` (`tensorrt_llm/serve/chat_utils.py`), mirror the existing image_url logic — detect `url.startswith("data:")`, split off the base64 payload, and call `async_load_audio` with `is_base64=True`. Pseudo-diff:

```python
if part_type == "audio_url":
    url = part["audio_url"]["url"]
    audio_kwargs = resolve_media_io_kwargs(
        mm_data_tracker._multimodal_server_config,
        mm_data_tracker.request_media_io_kwargs, "audio",
    )
    if url.startswith("data:"):
        # data:audio/<mime>;base64,<payload>  →  decode inline
        _, payload = url.split(",", 1)
        audio_kwargs["is_base64"] = True
        data = async_load_audio(payload, **audio_kwargs)
    else:
        # existing http/https/file branch
        data = async_load_audio(url, **audio_kwargs)
    return MultimodalData(modality="audio", data=data, is_embedding=False)
```

### Workarounds for users hitting this now

Either is acceptable until the fix lands:

1. Host audio over HTTP and pass an `http://...` URL to `audio_url` (TRT-LLM fetches and decodes it).
2. Bind-mount a host directory into the container and pass `file:///path/to/audio.wav` (TRT-LLM reads the file directly).

Both bypass the data-URI path entirely.

### Why this matters

OpenAI-compatible clients (LangChain, llama-index, custom eval pipelines) often send audio as inline base64 — it removes the dependency on an external file server and works the same way as image_url. The asymmetry between image_url (works) and audio_url (silently fails) is a footgun for anyone setting up multimodal eval against trtllm-serve, especially because the user-facing error message ("cannot unpack non-iterable NoneType object") doesn't hint at the actual cause ("the data URI was never decoded").

I'm happy to put together a PR with the snippet above plus a regression test if that helps.

ssam18:
The audio_url branch in trtllm-serve doesn't decode data URIs even though image_url already does. When a client sends an inline base64 WAV the long URL reaches soundfile.read as a filesystem path and libsndfile fails with Supplied filename too long, then audios None bubbles up and the Phi-4-MM processor crashes with a confusing message. This adds a data scheme branch to load_audio and async_load_audio that mirrors what load_base64_image already does on the image side, and a regression test that loads a real WAV through a data audio/wav base64 URL. Fixes #14100.

## Summary by CodeRabbit

* **New Features**
  * Added support for loading audio from base64-encoded data URLs. Audio loading functions now accept and properly decode data-scheme URLs containing base64-encoded audio content.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14115)

### DIFF
--- a/tensorrt_llm/inputs/utils.py
+++ b/tensorrt_llm/inputs/utils.py
@@ -215,6 +215,17 @@ async def async_load_video(video: str,
         raise ValueError(f"Unsupported URL scheme: {parsed_url.scheme!r}")
 
 
+def load_base64_audio(parsed_url) -> BytesIO:
+    data_spec, data = parsed_url.path.split(",", 1)
+    _, data_type = data_spec.split(";", 1)
+
+    if data_type != "base64":
+        msg = "Only base64 data URLs are supported for now."
+        raise NotImplementedError(msg)
+
+    return BytesIO(base64.b64decode(data))
+
+
 def load_audio(
     audio: str,
     format: str = "pt",
@@ -224,6 +235,8 @@ def load_audio(
     if parsed_url.scheme in ["http", "https"]:
         resp = _safe_request_get(audio, stream=False)
         audio = BytesIO(resp.content)
+    elif parsed_url.scheme == "data":
+        audio = load_base64_audio(parsed_url)
     elif parsed_url.scheme in ("", "file"):
         audio = _normalize_file_uri(
             audio) if parsed_url.scheme == "file" else audio
@@ -250,6 +263,8 @@ async def async_load_audio(
         session = await _get_aiohttp_session()
         audio_data = await _safe_aiohttp_get(audio, session=session)
         audio = BytesIO(audio_data)
+    elif parsed_url.scheme == "data":
+        audio = await asyncio.to_thread(load_base64_audio, parsed_url)
     elif parsed_url.scheme in ("", "file"):
         audio = _normalize_file_uri(
             audio) if parsed_url.scheme == "file" else audio
--- a/tensorrt_llm/inputs/utils.py
+++ b/tensorrt_llm/inputs/utils.py
@@ -216,6 +216,17 @@ async def async_load_video(video: str,
         raise ValueError(f"Unsupported URL scheme: {parsed_url.scheme!r}")
 
 
+def load_base64_audio(parsed_url) -> BytesIO:
+    data_spec, data = parsed_url.path.split(",", 1)
+    _, data_type = data_spec.split(";", 1)
+
+    if data_type != "base64":
+        msg = "Only base64 data URLs are supported for now."
+        raise NotImplementedError(msg)
+
+    return BytesIO(base64.b64decode(data))
+
+
 def load_audio(
     audio: str,
     format: str = "pt",
@@ -225,6 +236,8 @@ def load_audio(
     if parsed_url.scheme in ["http", "https"]:
         resp = _safe_request_get(audio, stream=False)
         audio = BytesIO(resp.content)
+    elif parsed_url.scheme == "data":
+        audio = load_base64_audio(parsed_url)
     elif parsed_url.scheme in ("", "file"):
         audio = _normalize_file_uri(
             audio) if parsed_url.scheme == "file" else audio
@@ -251,6 +264,8 @@ async def async_load_audio(
         session = await _get_aiohttp_session()
         audio_data = await _safe_aiohttp_get(audio, session=session)
         audio = BytesIO(audio_data)
+    elif parsed_url.scheme == "data":
+        audio = await asyncio.to_thread(load_base64_audio, parsed_url)
     elif parsed_url.scheme in ("", "file"):
         audio = _normalize_file_uri(
             audio) if parsed_url.scheme == "file" else audio
--- a/tests/unittest/inputs/test_async_media_loading.py
+++ b/tests/unittest/inputs/test_async_media_loading.py
@@ -49,6 +49,16 @@ def _make_audio_file(tmp_path: str) -> str:
     return tmp_path
 
 
+def _make_audio_data_url() -> str:
+    sample_rate = 16000
+    time_axis = np.linspace(0, 0.05, int(sample_rate * 0.05), endpoint=False)
+    audio_samples = (np.sin(2 * np.pi * 440 * time_axis) * 0.5).astype(np.float32)
+    audio_buf = BytesIO()
+    soundfile.write(audio_buf, audio_samples, sample_rate, format="WAV")
+    b64_str = base64.b64encode(audio_buf.getvalue()).decode()
+    return f"data:audio/wav;base64,{b64_str}"
+
+
 # ──────────────────────────────────────────────────────────────
 # async_load_image
 # ──────────────────────────────────────────────────────────────
@@ -112,6 +122,13 @@ class TestAsyncLoadAudio:
         audio_array, sample_rate = await async_load_audio(wav_path)
         assert isinstance(audio_array, np.ndarray)
         assert sample_rate == 16000  # matches the sr=16000 used in _make_audio_file
+
+    @pytest.mark.asyncio
+    async def test_load_audio_from_data_url(self):
+        url = _make_audio_data_url()
+        audio_array, sample_rate = await async_load_audio(url)
+        assert isinstance(audio_array, np.ndarray)
+        assert sample_rate == 16000
 
     @pytest.mark.asyncio
     async def test_cpu_work_runs_in_executor(self):
--- a/tests/unittest/inputs/test_async_media_loading.py
+++ b/tests/unittest/inputs/test_async_media_loading.py
@@ -52,6 +52,16 @@ def _make_audio_file(tmp_path: str) -> str:
     return tmp_path
 
 
+def _make_audio_data_url() -> str:
+    sample_rate = 16000
+    time_axis = np.linspace(0, 0.05, int(sample_rate * 0.05), endpoint=False)
+    audio_samples = (np.sin(2 * np.pi * 440 * time_axis) * 0.5).astype(np.float32)
+    audio_buf = BytesIO()
+    soundfile.write(audio_buf, audio_samples, sample_rate, format="WAV")
+    b64_str = base64.b64encode(audio_buf.getvalue()).decode()
+    return f"data:audio/wav;base64,{b64_str}"
+
+
 # ──────────────────────────────────────────────────────────────
 # async_load_image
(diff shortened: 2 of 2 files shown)


yechank-nvidia:
Hi @ssam18, thanks for the PR and for catching this!

We're currently refactoring the multimodal I/O paths, and I believe #14010 will already cover this scenario by routing data: URIs through the new `AudioMediaIO` loader. @aswinvisva, please correct me if I'm wrong here.

If that's the case, we'll likely close this PR in favor of #14010. That said, your regression test(test_load_audio_from_data_url) is a nice addition that #14010 doesn't have — if you'd like to keep that coverage, feel free to open a separate PR adding just the test once #14010 lands.

u7k4rs6:
Hi @ssam18  @yechank-nvidia , I was looking at this issue independently and
arrived at a similar fix. A few findings that might help unblock this PR:

1. **#14010 has merged but the legacy bug persists.** I confirmed on
   current `main` (post #14010) that the bug still reproduces:

```python
   import asyncio, base64
   from io import BytesIO
   import numpy as np, soundfile
   from tensorrt_llm.inputs.utils import async_load_audio

   buf = BytesIO()
   soundfile.write(buf, np.zeros(800, dtype=np.float32), 16000, format="WAV")
   url = "data:audio/wav;base64," + base64.b64encode(buf.getvalue()).decode()
   asyncio.run(async_load_audio(url))
   # ValueError: Unsupported URL scheme: 'data'
```

2. **The serve path is fixed** by `AudioMediaIO.async_load()` (the new
   route after #14010), but `tensorrt_llm.inputs.utils.async_load_audio`
   and `load_audio` are still part of the public API surface and any
   direct caller still hits the original bug.

3. **CodeRabbit's suggestion** about a sync-path test (`test_load_audio_from_data_url`
   for `load_audio`) seems still valid given point 2.

Happy to either help address the remaining feedback here, or open the
test-only follow-up PR you mentioned, whichever is most useful. Let me
know how you'd like to proceed.

BowenFu:
The behavior change is safe — `data:` previously fell through to `raise ValueError(f"Unsupported URL scheme: ...")` in both `load_audio` and `async_load_audio`, so this only turns an error into a success and cannot affect `http(s)`/`file`/bare-path callers.

Two things before it lands:

1. **This is now the third copy of the same data-URI parser.** `load_base64_image` (`inputs/utils.py:45`) and `load_base64_video` (`:124`) already do the identical `split(",", 1)` / `split(";", 1)` / `!= "base64"` → `NotImplementedError` dance, down to the same message string. Worth factoring the parse into one helper that returns the decoded bytes and letting the three callers wrap it (`Image.open`, `BytesIO`, `BytesIO`) — otherwise the next media type adds a fourth.

2. **`data_spec.split(";", 1)` is fragile for a real data URI.** It assumes exactly one `;`, so a legal `data:audio/wav;codecs=opus;base64,...` yields `data_type == "codecs=opus;base64"` and raises `NotImplementedError` on a well-formed input. Splitting on `;` and testing whether `"base64"` is in the parameter list would be more correct. The existing image/video copies have the same bug, which is one more argument for a single shared helper.

Also worth noting for whoever merges: the green check here is only DCO/pre-commit/title/base-freshness — blossom-ci has never run, and the head commit is from 2026-05-16, so the new test has no build evidence on current main.

u7k4rs6:
@BowenFu  @ssam18  I've opened #17168 with the shared parser from point 1.

It's scoped to the three defects that already exist in the image and video paths on
main, independent of the audio work here: the `split(";", 1)` bug, `load_base64_image`
/ `load_base64_video` raising `ValueError` instead of `NotImplementedError` on a legal
non-base64 URI, and a missing comma surfacing as an unpack error. It touches nothing
in this PR's diff, so #14115 should rebase cleanly and shrink to just the audio branch.

One deviation from your point 1: the helper returns the base64 payload as a string
rather than decoded bytes, since `BaseMediaIO.load_base64(media_type, data)` takes the
string and each subclass decodes. Happy to switch it either way.

u7k4rs6:
Confirming the #17168 side: if this lands first I'll open the follow-up folding load_base64_audio() into parse_data_uri, so audio doesn't stay on the old split(";", 1) form. If #17168 lands first, the rebase here is just the two elif branches calling the shared helper, as you describe.

PR #14121, opened 2026-05-14
[None][feat] Implement VSWA front-eviction for the Triton attention kernels

eopXD:
## Summary

Phase 2 of the Gemma-4 enablement (#13745 was Phase 1).  Under SWA
front-eviction the live window's cache length diverges from the global
`input_pos + seq_len`, so the kernel, the prepare-extra-metadata op, and the
per-group host prepare all need a window-capped `seq_len_with_cache`.  The
shim hands the kernel a window-coherent local-coord view per group;
kernel masks stay in local coordinates and need no changes.

End-to-end this unblocks long-context Gemma-4 (prompts > W=1024) where
front-eviction actually fires.  Pre-Phase-2, SWA layers silently read
stale page entries and the write-position for new tokens overshoots the
live page list, so accuracy collapses on long prompts while short-prompt
benchmarks (MMMU / MMLU / GSM8K) stay healthy.

## Source changes

- `custom_ops/attention_interface.py`: allocate per-group
  `seq_len_with_cache_g{i}`; extend `nest_sequences` with
  `seq_len_with_cache_per_group` / `last_page_len_per_group`; keep
  per-group `seq_len_with_cache` in sync inside `offset_pos_and_cache_`.
- `shim/ad_executor.py`: per `(request, window)`, query
  `get_num_front_blocks_removed`, slice `cache_loc` as
  `all_indices[front_removed : front_removed + num_active]`, compute
  window-capped per-group `seq_len_with_cache` and `last_page_len`.
  Math extracted into `_compute_window_local_view` for testability.
- `transform/library/kvcache.py`: add `"seq_len_with_cache"` to
  `vswa_swappable_bases`; build per-group `meta_nodes_extra_by_group` by
  re-invoking the backend's prepare-extra op once per non-zero group
  (generic over `vswa_swappable_bases`, not hardcoded names) so backends
  whose extra-op consumes swappable names (e.g. Triton, TrtllmAttention)
  no longer silently route every non-zero-group layer through group-0
  inputs.

No kernel changes — both `_paged_context_kernel` and
`_flash_decode_stage1_kernel` are already shift-invariant in local
coordinates.

## Test coverage

| Layer | File | What it guards |
|---|---|---|
| Transform graph shape | `tests/unittest/_torch/auto_deploy/unit/singlegpu/transformations/library/test_kvcache_vswa_metadata.py` | Per-group `seq_len_with_cache_g{i}_host` placeholders exist; the prepare-extra op is invoked once per group; each cached-attention call's `meta_nodes_extra` slot resolves to its own group's input. Fails on the pre-fix transform. CPU-only. |
| Triton kernel | `tests/unittest/auto_deploy/singlegpu/custom_ops/attention/test_triton_paged_attention_swa.py` | (a) Long-prefill SWA: prefill_len ∈ {1, 2, 4}×W, exercising both Phase-1 full-page and Phase-2 boundary-page masking; (b) Front-eviction: parametrized over `front_removed ∈ {0, 1, 4}`, `chunk ∈ {1, 64, 128}`, `page_size ∈ {16, 64}`, with/without sliding window. SDPA reference on the live window. |
| Shim integration | `tests/unittest/auto_deploy/singlegpu/shim/test_ad_executor_swa_eviction.py` | Real two-window `CachedSequenceInterface` + `KVCacheManager`; monkeypatches `get_num_front_blocks_removed` (real eviction can't be fired from Python without a model). Asserts: live-slice correctness, window-capped `seq_len_with_cache`, consistent `last_page_len`, page-table capacity invariant, spillover slot lives AFTER the live window. |
| End-to-end accuracy | `tests/integration/defs/accuracy/test_llm_api_autodeploy.py::TestGemma4MoE::test_bf16_long_context` | `PassKeyRetrieval64k` (64k prompts) on Gemma-4-26B-A4B-it — forces heavy SWA front-eviction while keeping the passkey reachable through full-attention layers. Reference accuracy registered in `passkey_retrieval_64k.yaml` (placeholder = 99, to be replaced with the measured value after the first CI run). |

## Known coverage gaps (no fast unit regression guard yet)

The following are transitively covered by the E2E test but lack
dedicated unit-level tests; flagged here so a follow-up can fill them
when convenient:

1. `nest_sequences` argument handling for `seq_len_with_cache_per_group`
   / `last_page_len_per_group` (override vs. default-replicate vs.
   partial-map guard).
2. `offset_pos_and_cache_` per-group `seq_len_with_cache` sync on the
   overlap-scheduler path.
3. Full `_prepare_inputs` shim flow integration (currently only the
   extracted helper is unit-tested; the surrounding `nest_sequences`
   plumbing is only exercised E2E).

## Out of scope (separately backlogged)

- `ad_executor.py`'s non-VSWA single-window branch still uses
  `cache_indices[:num_active_blocks_i]` without an eviction-aware
  slice.  Latent bug only if a single-window SWA model is wired up
  through AutoDeploy in the future; not Gemma-4-blocking.  Tracked in
  the project backlog.

## CI

Suggested run:

```
/bot run --extra-stage "DGX_H100-1_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1"
```

## Release Notes

* **New Features**
  * Added support for Variable Sliding Window Attention (VSWA) enabling different attention window sizes across model layers with independent memory optimization per window.
  * Introduced per-window KV cache configuration supporting heterogeneous head dimensions and data types for improved memory efficiency.
  * Extended long-context inference capabilities with 64k token passkey retrieval support for advanced models.

* **Tests**
  * Comprehensive test coverage for sliding window attention correctness and front-eviction behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14121)

### DIFF
--- a/tests/integration/defs/accuracy/test_llm_api_autodeploy.py
+++ b/tests/integration/defs/accuracy/test_llm_api_autodeploy.py
@@ -29,7 +29,7 @@ from tensorrt_llm.sampling_params import SamplingParams
 
 from ..conftest import get_device_count, llm_models_root, skip_pre_blackwell
 from .accuracy_core import (GSM8K, MMLU, MMMU, CnnDailymail,
-                            LlmapiAccuracyTestHarness)
+                            LlmapiAccuracyTestHarness, PassKeyRetrieval64k)
 
 _AD_CONFIGS_DIR = (Path(get_llm_root()) / 'examples' / 'auto_deploy' /
                    'model_registry' / 'configs')
@@ -1180,6 +1180,47 @@ class TestGemma4MoE(LlmapiAccuracyTestHarness):
             task = GSM8K(self.MODEL_NAME)
             task.evaluate(
                 llm,
+                extra_evaluator_kwargs=self.EXTRA_EVALUATOR_KWARGS,
+            )
+
+    @pytest.mark.timeout(7200)
+    @pytest.mark.skip_less_device_memory(80000)
+    def test_bf16_long_context(self):
+        """End-to-end accuracy guard for the SWA front-eviction path.
+
+        Gemma-4's SWA window is 1024 tokens; the global-attention layers see
+        the full context. A 64k-prompt PassKey retrieval task drives heavy
+        front-eviction on the SWA pool while keeping the passkey reachable
+        through the full-attention layers — the workload that exercises the
+        Phase-2 (window-coherent metadata + per-group prepare-extra) path
+        added by the accompanying source change.
+
+        Pre-Phase-2: the SWA layer reads stale/evicted page entries and the
+        write-position for new tokens overshoots the live page list, so
+        accuracy collapses well below the registered reference even though
+        MMMU / MMLU / GSM8K (all short-prompt) stay healthy.
+        """
+        yaml_paths, registry_world_size = _get_registry_yaml_extra(
+            self.MODEL_NAME)
+        if get_device_count() < registry_world_size:
+            pytest.skip("Not enough devices for world size, skipping test")
+
+        sampling_params = SamplingParams(
+            max_tokens=PassKeyRetrieval64k.MAX_OUTPUT_LEN,
+            truncate_prompt_tokens=PassKeyRetrieval64k.MAX_INPUT_LEN,
+            end_id=None,
+            pad_id=None,
+            n=1,
+            use_beam_search=False,
+        )
+        with AutoDeployLLM(model=self.MODEL_PATH,
+                           tokenizer=self.MODEL_PATH,
+                           world_size=registry_world_size,
+                           yaml_extra=yaml_paths) as llm:
+            task = PassKeyRetrieval64k(self.MODEL_NAME)
+            task.evaluate(
+                llm,
+                sampling_params=sampling_params,
                 extra_evaluator_kwargs=self.EXTRA_EVALUATOR_KWARGS,
             )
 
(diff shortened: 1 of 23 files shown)


tests/integration/defs/accuracy/test_llm_api_autodeploy.py
```diff
@@ -1181,38 +1122,102 @@ def test_nvfp4(self, ep_size, attention_dp):
...
     MODEL_PATH = hf_id_to_local_model_dir(MODEL_NAME)
     EXTRA_EVALUATOR_KWARGS = {
         "apply_chat_template": True,
     }
 
-    def get_default_sampling_params(self):
-        return SamplingParams(
-            max_tokens=MMMU.MAX_OUTPUT_LEN,  # noqa: F821
-            truncate_prompt_tokens=MMMU.MAX_INPUT_LEN,  # noqa: F821
+    @pytest.mark.skip_less_device_memory(80000)
```
crazydemo, line 1142: please add these cases to llm_function_core.txt

nvpohanh:
@thorjohnsen @SimengLiu-nv could you review this?

eopXD:
Dropping this merge request, recently-added test coverage for gemma3 running MMLU forced us to land the kernel change altogether with the multi-pool change or else we will break long-sequence infernce because previously phase 1 did not include the plumbing local view of SWA attention (front-eviction).

Now we will implement the front-eviction logic altogether in #13745.

PR #14127, opened 2026-05-14
[None][feat] Expose host/GPU per-iter time and clarify iter labeling in /metrics

eopXD:
## Summary

Add three new JSON keys to each `/metrics` `IterationStats` record and fix a
labeling artifact in the `iter` field. All changes are Python-only and ride
the existing serializer-injection pattern used for `kvCacheIterationStats`
(PR #12413) — **no C++ rebuild required**.

### New JSON keys

| Key | Type | Meaning |
|---|---|---|
| `hostStepTimeMS` | float, ms | Clean per-loop CPU wall captured by `profile_step()`. Mirrors the log line's `host_step_time`. Always single-loop, unlike `iterLatencyMS` which spans ~2 loops under the overlap scheduler. |
| `prevDeviceStepTimeMS` | float, ms | GPU forward time read via the ping-pong CUDA event pair in `profile_step()`. The `"prev_"` prefix reflects that under steady state the value lags `hostStepTimeMS` on the same record by 1 loop. |
| `schedulerMode` | `"overlap"` \| `"non_overlap"` | Per-record indicator so consumers can interpret `iterLatencyMS` without inspecting server config. PP (`pp_size > 1`) always reports `"overlap"` since `_executor_loop_pp` is structurally overlap-style (BatchStatePP is queued in iter N and consumed in a later loop), regardless of `disable_overlap_scheduler`. |

### `iter` field relabel

`stats.iter` is now stamped at batch construction time in
`_get_init_iter_stats` (the loop that BUILT the batch) instead of at
consumption time in `_update_iter_stats`. Under the overlap and PP
schedulers `_process_iter_stats` runs in a later loop than the one that
built the batch, so the live `iter_counter` mislabeled records by 1+ iters.

**Note**: this changes how `/metrics` jsonl aligns with `server_runtime.log`.
Under overlap mode the two streams are now off by 1 iter (the log line still
stamps from the live counter at print time, which is intentional and
unchanged here).

### Why

These changes address three concrete pain points users hit when joining
`/metrics` data against batch composition for performance benchmarking:

1. `numCtxTokens` (and other `inflightBatchingStats`) appeared to lag the
   `iter` label by 1+ iterations under overlap — actually the `iter` label
   itself was wrong; the batch composition was correctly attributed.
2. `iterLatencyMS` is misleading for absolute per-loop CPU cost under
   overlap (spans ~2 loops, `sum ≈ 2 × wall_time`). `hostStepTimeMS` gives
   the clean per-loop value, and `schedulerMode` lets consumers branch.
3. GPU per-iter time was only available in the text log (`prev_device_step_time`),
   not in the JSON. `prevDeviceStepTimeMS` exposes it directly to /metrics
   consumers.

## Implementation summary

- `tensorrt_llm/_torch/pyexecutor/py_executor.py`:
  - `_profiler.profile_step()`: split timing capture from log emission so
    timing is stashed on `self._latest_host_step_time_ms` /
    `self._latest_prev_device_step_time_ms` whenever
    `enable_iter_perf_stats` or `print_iter_log` is on.
  - `_get_init_iter_stats`: stamp `stats.iter` at construction.
  - `_update_iter_stats`: removed the consumption-time `stats.iter` overwrite
    (replaced with a comment).
  - `_process_iter_stats`: snapshot per-loop timings and pass through the
    append/queue path; documented `iterLatencyMS` semantics under both
    schedulers.
  - `_append_iter_stats`: extended with `host_step_time_ms` /
    `prev_device_step_time_ms` kwargs; computes `schedulerMode` with the
    PP-as-overlap rule; injects new keys into the `gather_all_ranks`
    per-rank dict and into the legacy tuple (now 7 entries, defensive
    `len() > N` access in the serializer keeps older shapes readable).
- `tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py`: `ADPIterStatsRecord`
  carries the new timing fields; `queue` / `finalize` / `_discard` updated
  to preserve them across the ADP fanout delay. Rank-0 values are broadcast
  to all rank rows (true per-rank timing would require widening the
  rank-state allgather payload — left for a future PR).
- `tensorrt_llm/executor/base_worker.py`: `_stats_serializer` reads tuple
  slots `[4]`/`[5]`/`[6]` defensively and injects the new JSON keys with
  field-level comments explaining the semantics.

## Test plan

- Build wheel and serve a model with `enable_iter_perf_stats=true` + `enable_block_reuse=true`
- Verify `GET /metrics` JSON contains `hostStepTimeMS`, `prevDeviceStepTimeMS`, and `schedulerMode` on each record
- Verify `iter` field matches construction-time iter (numbers are 1 lower under overlap mode than before)
- Validate that a PP run (`pp_size > 1`) reports `schedulerMode = "overlap"` in `/metrics` even when `disable_overlap_scheduler=True`.
- Confirm log emission still works when `print_iter_log=true` (gated condition was widened to also fire when `enable_iter_perf_stats=true`)
- Confirm legacy 4-/3-tuple stat consumers (e.g. the TRT-engine path returning 3-tuples) still serialize OK
- Run unit tests under `tests/unittest/`
- Validate ADP fanout path (multi-rank attention-DP run) still emits records

[x] Write code carefully and responsibly.


Superjomn (commented):
Are there any tests to protect this feature? Or it will be broken easily in the following iterations.

Superjomn (approved):
LGTM

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1445,12 +1478,42 @@ def _append_iter_stats(self,
...
         """
         # Non-ADP appends immediately, so the latest KV stats belong to this
         # IterationStats. ADP appends later and passes the saved iter-matched
         # KV stats explicitly.
         if not self.enable_attention_dp:
             kv_iter_stats = self._latest_kv_iter_stats
+            # Same reasoning for the per-loop timings: non-ADP captures them
+            # at append time. ADP passes the values that were saved with the
+            # IterationStats at queue time.
+            if host_step_time_ms is None:
```
lancelly, line 1552: Nit: Why kv_iter_stats gets an unconditional overwrite in the non-ADP branch while host_step_time_ms / prev_device_step_time_ms use an is None fallback?🤔

lancelly (approved):
LGTM~ Thanks!

tensorrt_llm/_torch/pyexecutor/py_executor.py
eopXD: - Non-ADP path: caller passes timings, function reads kv_iter_stats from self._latest_*. The if … is None: is dead code but harmless — it never fires in non-ADP because the caller always passes non-None values.
  - ADP path: enable_attention_dp is True, the whole if not self.enable_attention_dp: block is skipped, so all three values come from the caller (the saved record).

It works. The asymmetry is a style nit, not a correctness bug.

eopXD:
/bot skip --comment "Given that this MR has earned pre-merge CI success before. The latest merge is only on resolving conflict to the print calls. The failing pipeline is un-related to the change introduced in this MR. Skipping the CI now."

PR #12932, opened 2026-04-10
[None][feat] Add Gemma4 multimodal model support (text + vision + audio)

lfr-0531:
## Summary

- Add full Gemma4 multimodal model support (text + vision + audio) to TensorRT-LLM's PyTorch backend
- Covers all 4 Gemma4 model variants: 26B-A4B-it (MoE, K=V), E2B-it (KV sharing, PLE), 31B-it, E4B-it
- Implements all Gemma4-specific features: hybrid attention (per-layer head_dim 256/512), MoE (128 experts), K=V attention, KV sharing (via cache_layer_idx), Per-Layer Embeddings (PLE), double-wide MLP, logit softcapping, bidirectional mask, v_norm
- Adds Gemma4 tool call parser for custom non-JSON format
- Adds multimodal support: vision tower, audio tower, embedder, input processor
- **Adds Triton prefill attention fallback** for head_dim=512 layers with bidirectional (custom) masks — solves trtllm-gen's lack of Custom mask cubin for H512
- **CUDA graph support**: Fixed VSWA hybrid attention for CUDA graph capture/replay — unified trtllm-gen backend, fixed pool buffer sizing, `_block_tables` update in prepare()
- **NVFP4 support**: Q is cast to FP8 alongside KV in the FlashInfer backend so trtllm-gen context cubins (`QkvE4m3OBfloat16`) work; uniform trtllm-gen backend across all layers; CUDA graphs work for both BF16 and NVFP4 paths
- **V2 sliding-window eviction enabled**: `_util.py` uses the model's actual `sliding_window` for sliding layers' `max_attention_window`. V2 evicts old blocks beyond the window, dropping per-sequence sliding-pool block count from `ceil(max_seq_len/page_size)` to `ceil(sliding_window/page_size)` (≈4× sliding-pool memory reduction at typical configs). Unlocks much larger batch sizes on a single B200.
- **Real model E2E validated**: all 4 BF16 models + the NVFP4 quantized 31B produce correct text output; 26B/31B multimodal with Triton prefill produces image descriptions consistent with HF reference

### New Files
| File | Purpose |
|------|---------|
| `_torch/models/modeling_gemma4.py` | Text model: Attention, MoE, PLE, DecoderLayer, ForCausalLM |
| `_torch/models/modeling_gemma4mm.py` | Multimodal: vision/audio towers, embedder, input processor, VLM wrapper |
[...]

PR #14134, opened 2026-05-14
[None][feat] Add chunked prefill support for Gemma4 (text + vision multimodal)

Hudayday:
## Summary

  - Add chunked prefill support for Gemma4 (all 4 variants: E2B, E4B, 26B-A4B MoE, 31B) for both text-only and multimodal (image + audio) requests
  - Builds on top of #12932 (Gemma4 base support). Independent of any KV-cache-manager-V2 leak fixes — those will land in a separate follow-up PR
  - **SWA chunked prefill via Triton**: trtllm-gen has no `head_dim=512` + sliding-window + chunked-prefill cubin, so we route SWA-layer prefill through the existing Triton kernel (added in #12932) once a request is in chunked mode. Causal mask
  path supported (no custom mask needed)
  - **MM bidirectional block chunk alignment**: Modified V2 scheduler aligns each chunk boundary to bidirectional MM block boundaries (Gemma4 26B/31B vision blocks), so a vision block never gets split across two prefill chunks (which would break the
  bidirectional attention mask invariant)
  - **Companion `triton_prefill` int64-stride fix**: 26B/31B production trigger — `safe_page_ids * stride_buf_kpage` was int32 and overflowed once `page_id * stride > 2^31`, producing illegal memory accesses on B200 chunked MMMU. Single
  `.to(tl.int64)` cast covers the offset arithmetic.
  - **Validated end-to-end**: smoke MMMU (image) + GSM8K (text-only) across all 4 variants on B200 single-GPU, plus full-set 31B mmmu (900 samples) — all green

  ## Modified Files

  | File | Purpose |
  |------|---------|
  | `_torch/attention_backend/triton_prefill.py` | Routing knob: enter Triton path on SWA + chunked + no-custom-mask. `safe_page_ids.to(tl.int64)` overflow fix. |
  | `_torch/attention_backend/flashinfer.py` | Gate Triton-fallback selection from FlashInfer backend in chunked mode. |
  | `_torch/models/modeling_gemma4.py` | Wire chunked-prefill path through Gemma4Attention for SWA layers. |
  | `_torch/models/modeling_gemma4mm.py` | Per-instance `mm_bidirectional_blocks` flag derived from `text_config.use_bidirectional_attention == "vision"`. `get_mm_token_ids` override. Audio `get_num_tokens_per_audio` override. |
  | `_torch/models/modeling_multimodal_utils.py` | Slice `mm_embeds` per chunk before `fuse_input_embeds`. Drop no-op log. |
  | `_torch/pyexecutor/scheduler/scheduler_v2.py` | `_align_chunk_to_mm_block` — INCLUSIVE-cumsum chunk-to-bidirectional-block alignment, with impossibility raise when MM block > `max_num_tokens`. |
  | `inputs/registry.py` | `setdefault(mm_bidirectional_blocks)` before cumsum idempotency guard so the hashing path also populates the gate. `isinstance(dict)` Mock-truthy guard. |
  | `tests/unittest/_torch/executor/test_kv_cache_v2_scheduler.py` | Unit tests for `_align_chunk_to_mm_block` (INCLUSIVE cumsum convention, MM-block-split scenarios, impossibility raise, snap-down test fix). |

  ## Key Design Decisions

  - **Triton-prefill routing**: Re-use the existing Triton prefill kernel added in #12932 (originally for `head_dim=512` + custom mask). Chunked-prefill SWA layers also enter this kernel, but with `custom_mask=None` (auto-generated causal). This
   avoids requiring a new trtllm-gen cubin for `H512` + chunked + SWA.
  - **MM chunk alignment — INCLUSIVE cumsum**: The chunk aligner consumes `embed_mask.cumsum(0)` of length `P` written by `registry.py` / runtime / `ad_executor`. The scheduler now reads it as INCLUSIVE (`mask[i] = cumsum[i] - cumsum[i-1]`)
  instead of the SHIFTED `[P+1]` convention, eliminating the out-of-range `cumsum[end_abs+1]` read that previously tripped `IndexError @ scheduler_v2.py:572` when the last chunk landed on `prompt_len-1`.
  - **Per-instance `mm_bidirectional_blocks`**: Set only when `text_config.use_bidirectional_attention == "vision"` (26B/31B). E2B/E4B (fully causal text + audio-only MM) skip the aligner entirely — their large audio soft-token block (≈451)
  would otherwise exceed `max_num_tokens` and trip the impossibility raise.
  - **MM embed slicing per chunk**: `fuse_input_embeds` receives only the slice of `mm_embeds` covering the current chunk's positions, so the mm-embed-to-text-positions alignment stays valid across chunks.

  ## Companion Bug Fixes

  - **`triton_prefill` int32 overflow** — `page_id * stride` silently wrapped to negative for `page_id > 16K` on a `[N, 2, 8, 32, 256]` KV layout (stride 131072). Deterministic IMA on first `prefix_lens > 0` iter. Single `.to(tl.int64)` cast on
  `safe_page_ids` fixes both K and V offset arithmetic.
  - **Audio hashing path** — Gemma4 inherited `get_num_tokens_per_image/_video` but had no `get_num_tokens_per_audio`. After image-then-audio request sequences locked `multimodal_hashing_supported=True`, audio raised `AttributeError`. Added
  override returning `processor.audio_seq_length` (750 fixed for E2B/E4B).
  - **Mock-truthy `py_multimodal_data` gate** — was `getattr-or-{}`, which doesn't short-circuit on `Mock`. Changed to `isinstance(data, dict)`.

  ## Test Plan

  - `test_kv_cache_v2_scheduler` unit tests (INCLUSIVE-cumsum chunk-to-mm-block alignment, impossibility raise, snap-down preempted, multi-block scenarios) — **all green locally** (153 V2 + 16 MM tests).
  - B200 single-GPU MMMU smoke (NUM_SAMPLES=30, `max_num_tokens=2048`, `max_seq_len=16384`) for **all 4 variants** (E2B / E4B / 26B / 31B) — all green.
  - B200 single-GPU GSM8K smoke (NUM_SAMPLES=50, text-only) for **all 4 variants** — all green.
  - B200 single-GPU MMMU full set (NUM_SAMPLES=900, `max_num_tokens=4096`, `max_seq_len=16384`) on 31B chunked + nochunk — accuracy matches non-chunked baseline within seed noise.

  ## Known Limitations

  - KV-cache-manager-V2 high-pressure leak (overlap-scheduler + suspend-as-victim desync) is **out of scope** for this PR. Validation here is on configurations where that leak does not trigger, or with the V2-leak fix applied externally.
  - **Multimodal towers are still HF-based.** Vision tower and audio tower in `modeling_gemma4mm.py` use `AutoModel.from_config(config.vision_config)` / `AutoModel.from_config(config.audio_config)` from `transformers`, carried over from #12932.
  Per [@yechank-nvidia's review on #12932](https://github.com/NVIDIA/TensorRT-LLM/pull/12932):
    > Can you re-write the vision_tower into our TRT-LLM's style? We deprecated all direct HF vision tower import.

    As discussed in that thread, the tower port is deferred to a follow-up PR because it requires (a) a `SiglipVisionModel`-equivalent TRT-LLM module, (b) a custom weight remapper for the multimodal checkpoint, and (c) a Siglip2-config compat
  check against the existing `modeling_siglip.py`. The audio tower has the same shape and will ship in the same follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added sliding-window constraint support for Triton prefill attention operations.
  * Enhanced multimodal attention to respect bidirectional block boundaries during request scheduling.
  * Improved Gemma4 support for cached/paged prefix concatenation with multimodal tokens.
  * Added audio token counting predictions for Gemma4 multimodal processing.

* **Bug Fixes**
  * Hardened out-of-window KV cache page handling in Triton prefill.

* **Tests**
  * Added comprehensive test suite for multimodal-aware chunk alignment validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14134)

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_gemma4mm.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma4mm.py
@@ -307,6 +307,22 @@ class Gemma4InputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyInpu
             return super().get_mm_token_ids()
         return torch.tensor(ids, dtype=torch.int32)
 
+    def get_num_tokens_per_audio(self, *, audio, **kwargs) -> int:
+        # Gemma4 audio token count is dynamic: HF Gemma4Processor inserts
+        # placeholders via _compute_audio_num_tokens(waveform, sampling_rate),
+        # which mirrors the audio encoder's mel-frame + 2x SSCP-conv stride
+        # math and caps at audio_seq_length (=750, the 30s window). HF's
+        # `_get_num_multimodal_tokens` evaluates the formula at
+        # `feature_extractor.sampling_rate` (16 kHz), so the predicted count
+        # must use the post-resample length. Run the same
+        # `_normalize_audio_inputs` step that the prompt-side `_preprocess`
+        # path applies (mono downmix + scipy.signal.resample_poly to 16 kHz)
+        # so the predictor and the actual placeholder count agree exactly.
+        target_sr = self.processor.feature_extractor.sampling_rate
+        normalized = _normalize_audio_inputs([audio], target_sr=target_sr)
+        n_samples = len(normalized[0])
+        return self.get_num_multimodal_tokens(audio_lengths=[n_samples])["num_audio_tokens"][0]
+
     @property
     def dtype(self) -> torch.dtype:
         return self._dtype
--- a/tensorrt_llm/_torch/models/modeling_gemma4mm.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma4mm.py
@@ -48,7 +48,7 @@ from ...sampling_params import SamplingParams
 from ..attention_backend import AttentionMetadata
 from ..modules.linear import Linear
 from .modeling_gemma4 import Gemma4ForCausalLM
-from .modeling_multimodal_utils import fuse_input_embeds
+from .modeling_multimodal_utils import find_input_mm_embeds, fuse_input_embeds
 from .modeling_utils import ModelConfig, filter_weights, register_auto_model
 
 _MIN_TRANSFORMERS_FOR_GEMMA4 = "5.5.0"
@@ -895,6 +895,13 @@ class Gemma4ForConditionalGeneration(PreTrainedModel):
                 mm_token_ids_for_mask = fuse_token_ids.to(input_ids.device)
                 mm_mask = torch.isin(input_ids, mm_token_ids_for_mask)
                 ple_input_ids = torch.where(mm_mask, torch.full_like(input_ids, pad_id), input_ids)
+
+        # Slice mm_embeds to the current chunk window for chunked prefill /
+        # KV reuse. Without this, the full request's mm_embeds is passed to
+        # fuse_input_embeds even when input_ids is a chunk slice, which trips
+        # the count-equality check in fuse_input_embeds. Mirrors the call in
+        # modeling_qwen2vl / modeling_phi4mm / modeling_mistral.
+        mm_embeds = find_input_mm_embeds(mm_embeds, multimodal_params)
 
         input_ids, inputs_embeds = fuse_input_embeds(
             embedding_layer=self.llm.model.embed_tokens,
--- a/tensorrt_llm/_torch/models/modeling_gemma4mm.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma4mm.py
@@ -199,10 +199,12 @@ class Gemma4InputProcessor(BaseMultimodalInputProcessor, BaseMultimodalDummyInpu
     processing using the image processor saved in the model directory.
     """
 
-    # Gemma4 image/audio soft-token runs use bidirectional attention spanning
-    # the full block — chunked-prefill must keep each block intact in a single
-    # iteration. See registry.BaseMultimodalInputProcessor.mm_bidirectional_blocks.
-    mm_bidirectional_blocks = True
+    # Default class-level fallback. Real value computed per-instance below
+    # from text_config.use_bidirectional_attention. Only 26B/31B set
+    # use_bidirectional_attention="vision" — their image blocks need intact
+    # bidirectional attention. E2B/E4B set None (fully causal) — audio (and
+    # image, when present) can be split across chunks safely.
+    mm_bidirectional_blocks = False
 
     def __init__(
         self,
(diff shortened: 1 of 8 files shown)


tensorrt_llm/inputs/registry.py
```diff
@@ -136,6 +136,14 @@ class BaseMultimodalInputProcessor(ABC):
     processes the multimodal inputs.
     """
 
+    # Whether multimodal soft-token runs need bidirectional attention spanning
+    # the full block. When True, the chunked-prefill scheduler must keep each
+    # MM block intact within a single iteration (snap-up or snap-down at chunk
+    # boundaries) so the second half does not attend to a frozen first half.
+    # Default False: most VLMs (Llava, Qwen-VL) use causal attention over MM
+    # tokens and tolerate splitting. Gemma4 sets True.
+    mm_bidirectional_blocks: ClassVar[bool] = False
```
2ez4bz, line 145: I might be missing something - why do we need this here if the check is on the instance at line 840?

tensorrt_llm/_torch/models/modeling_gemma4mm.py
```diff
@@ -273,6 +290,39 @@ def get_vocab_size(self) -> Optional[int]:
...
+        # zero MM positions, producing an all-zero embed_mask cumsum and
+        # making chunked-prefill drop every MM embedding row. HF's
+        # `Gemma4Processor` does not expose `mm_token_ids` either, so we
+        # build the list from the config here.
+        ids = []
+        for name in ("image_token_id", "audio_token_id", "video_token_id"):
+            tid = getattr(self._config, name, None)
+            if tid is not None:
+                ids.append(int(tid))
+        if not ids:
```
2ez4bz, line 306: Based on the code comment, wouldn't this be wrong? If so, should we raise an error instead?

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
```diff
@@ -321,8 +318,10 @@ def find_input_mm_embeds(
 
     if len(mm_embeds) == 1:
         sliced = [mm_embeds[0][start:end] for start, end in slices]
-        return [torch.cat(sliced, dim=0)]
-    return [mm_embeds[i][start:end] for i, (start, end) in enumerate(slices)]
+        out = [torch.cat(sliced, dim=0)]
```
2ez4bz: Were these just cosmetic changes?

2ez4bz (approved):
Approving changes for the modeling files.

tensorrt_llm/_torch/models/modeling_gemma4mm.py
Hudayday: Yes, we should raise an error here. Thanks for the suggestion, fixed in the following commit.

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
Hudayday: Yes, those were cosmetic which I added when I was debugging. Reverted `tensorrt_llm/_torch/models/modeling_multimodal_utils.py` to the original.

tensorrt_llm/inputs/registry.py
Hudayday: It has been added here as a template for any MM processors. 
Putting the default on the base class lets any subclass with a static value opt in with one line (`mm_bidirectional_blocks = True`) and skip `__init__` entirely, and keeping it on the base standardizes the contract so future MM processors don't have to rediscover the flag.

Gemma4 uses the same flag and just additionally sets it per-instance in `__init__` because the value depends on `text_config.use_bidirectional_attention`.

lfr-0531 (approved):

lancelly (approved):
Scheduler changes looks good to me

PerkzZheng (approved):

PR #14140, opened 2026-05-14
[None][chore] Refactor salting support for KVCacheManagerV2

lowsfer:
## Summary
- Refactor KV cache reuse namespace into ReuseScope.
- Route LoRA task ID and cache salt through ReuseScope for KVCacheManagerV2.
- Keep salting coverage focused on ReuseScope serialization, radix-tree scoping, and scoped reuse behavior.

## Tests
- git diff --check
- python -m py_compile on touched Python files
- PYTHONPATH=/home/yaoy/tekit/tensorrt_llm/runtime/ python tests/unittest/kv_cache_manager_v2_tests/test_kv_cache_salting.py -v
- pre-commit hooks on commit

Note: targeted CUDA E2E test was attempted locally but blocked by CUDA initialization failure before test body.

## Summary by CodeRabbit

* **New Features**
  * Introduced `ReuseScope`, a unified parameter component for managing KV cache reuse configuration including LoRA and cache isolation settings.

* **API Changes**
  * `KVCacheManager.create_kv_cache()` now accepts a `reuse_scope` parameter instead of separate `lora_task_id` and `cache_salt_id` parameters, providing a more consolidated interface for cache reuse control.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14140)


eopXD (approved):
Looks good. Thank you for the refactoring.

tensorrt_llm/runtime/kv_cache_manager_v2/_common.py
```diff
@@ -57,6 +58,27 @@ class PageIndexMode(enum.IntEnum):
 #      If we do this, we can't skip the encoder.
 TokenIdExt = TokenId | bytes
 
+
+class ReuseScope(NamedTuple):
+    """Per-request namespace for prefix reuse."""
+
+    lora_id: int | None = None
+    salt: int | None = None
+    # When adding more fields, remember to update to_bytes().
```
eopXD: Slightly misleading comment.
Do you mean updating ` assert len(self) <= 8`?
I think it should be tightened to 2 given the number of current fields.

lowsfer: No, I mean if new field is not int|None, then we need additional logic for serialization. len(self)<8 is the requirement for the mask, actually, I made a mistake, it should be 64 as we are using 8-byte bitmask. But yes I can add additional assert to remind developer to update.

eopXD: I see. That resolves my confusion. Thank you.

PR #14165, opened 2026-05-15
[TRTLLM-12462][fix] Fix FP8 block scaling GEMM autotuner cache growth

cascade812:
### Issue Description

Repeated requests with `Qwen/Qwen3-30B-A3B-Instruct-2507-FP8` caused steady RSS growth (mpi4py.futures.server RSS: +~350 MB / 3000 requests) in the MPI worker process. The leak reproduced with OpenAI serving and with a direct LLM API repro, even after disabling postprocess workers and perf/stat metrics.

### Root Cause
`get_fp8_block_scaling_gemm_constraint_spec()` created fresh function objects on every call:
`lambda inputs: inputs[0][0]`and a nested `fp8_quantize_1x128_sm90_constrant()` function.

Those callables are stored inside `ConstraintSpec`, and `ConstraintSpec` participates in the key for `AutoTuner._find_nearest_profile`, which uses an unbounded `@lru_cache(maxsize=None)`. Even for repeated shapes, the cache key changed every request because the function object identity changed, causing unbounded cache misses and retained entries.

### Fix
Move the constraint functions to stable module-level helpers and cache the generated `ConstraintSpec` tuple by SM version. This keeps the constraint logic unchanged while making the autotuner cache key stable across requests.
After the fix:
```text
_find_nearest_profile currsize: 139 entries, flat through 3000 requests
traced_current_mb: ~4 MB, flat through 3000 requests
mpi worker RSS: +22.8 MB / 3000 requests
```

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
+++ b/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
@@ -1693,17 +1693,29 @@ class Fp8BlockScalingGemmRunner(TunableRunner):
             a, b, a_scale, b_scale)
 
 
-def get_fp8_block_scaling_gemm_constraint_spec():
-    # The implementation aligns with the fp8_quantize_1x128 custom op.
-    def fp8_quantize_1x128_sm90_constrant(inputs: List[List[int]]):
-        pad_m = fp4_utils.pad_up(inputs[0][0], 4)
-        blocked_n = (inputs[0][1] + 127) // 128
-        return fp4_utils.pad_up(pad_m * blocked_n * 4, 128) // 4
+def _fp8_block_scaling_gemm_sm100_constraint(inputs: List[List[int]]) -> int:
+    return inputs[0][0]
 
-    if get_sm_version() >= 100:
-        return (ConstraintSpec(2, 1, lambda inputs: inputs[0][0]), )
+
+def _fp8_quantize_1x128_sm90_constraint(inputs: List[List[int]]) -> int:
+    # The implementation aligns with the fp8_quantize_1x128 custom op.
+    pad_m = fp4_utils.pad_up(inputs[0][0], 4)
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
```diff
@@ -1693,17 +1693,29 @@ def forward(
...
+    return fp4_utils.pad_up(pad_m * blocked_n * 4, 128) // 4
+
 
-    if get_sm_version() >= 100:
-        return (ConstraintSpec(2, 1, lambda inputs: inputs[0][0]), )
+@lru_cache(maxsize=None)
+def _get_fp8_block_scaling_gemm_constraint_spec(
+        sm_version: int) -> Tuple[ConstraintSpec, ...]:
+    if sm_version >= 100:
+        return (ConstraintSpec(2, 1,
```
yuzisun, line 1711: This fixes the constraint_specs which helps the same request sending over again, would it still be a problem of variable inputs with the unbounded cache?

cascade812: @hyukn The cache was changed to unbounded in this [PR](https://github.com/NVIDIA/TensorRT-LLM/pull/4466) , is there a specific reason for doing this?

hyukn: Because the overall entries of the cache can be large in some cases. So the upper bound was removed.
This fix reminds me that the LRU result of the find_nearest_profile is not guaranteed. When a dynamic object is used instead of a static one, such as a class member or a static method, then the LRU cache will never work, and will also introduce unexpected memory growth.

I have examined other ops and seem there is no further issue.

hyukn (approved):
This might also affect perf when the LRU cache grows very large. the LRU look-up may introduce extra overhead.

PR #14213, opened 2026-05-17
[None][fix] Avoid dp_size x ep_size double-count in MegaMoEDeepGemm SymmBuffer

qiaoxj07:
## Summary

When attention DP is enabled, the DG MegaMoE ``SymmBuffer`` per-rank pool
is sized as ``dp_size * ep_size * max_num_tokens * num_topk`` instead of
``dp_size * max_num_tokens * num_topk`` — the EP factor is double-counted.
On DEP=32 DSv4 (max_num_tokens=1024, num_topk=6, hidden=7168,
moe_intermediate=3072) this inflates the SymmBuffer to **95 GiB per rank**
and OOMs the GEN ranks during model load on a 276 GiB GB300 GPU.

Root cause: ``ModelConfig.__post_init__`` pre-multiplies
``moe_max_num_tokens = max_num_tokens * dp_size``. ``MegaMoEDeepGemm.__init__``
then forwards this value to ``get_symm_buffer_for_mega_moe(...)``, whose
``layout::Workspace`` / ``get_num_max_pool_tokens`` (DeepGEMM) further
scales the pool by ``num_ranks`` (= ``ep_size``). Because the supported
full-ADP topology asserts ``ep_size == parallel_size == dp_size``, the
``dp_size`` factor is applied twice.

Fix: divide ``self.max_num_tokens`` by ``ep_size`` (ceil) when
``use_dp`` is true and ``ep_size > 1``. The DG kernel still sees a pool
sized for the worst case where every cluster token routes to one rank's
local experts — but the EP factor is no longer doubled.

Effect (theoretical, DSv4 routed expert, full ADP, max_num_tokens=1024):

| Topology | Before | After |
| --- | ---: | ---: |
| DEP=32 | 95 GiB | ~3 GiB |
| DEP=16 | 24 GiB | ~0.6 GiB |
| DEP=8  | 3 GiB  | ~0.15 GiB |
| non-ADP | unchanged | unchanged |

Non-ADP paths (``use_dp == False``) and ADP+EP=1 paths are not touched.

## Test plan

- [x] Rerun DSv4 disagg DEP=32 megamoe_deepgemm GEN that previously OOM'd
      on rank 3 / rank 7 — verify SymmBuffer line reports ~3 GiB and GEN
      workers reach the benchmark loop.
- [x] Rerun DSv4 disagg DEP=16 baseline to confirm no perf regression.
- [x] Sanity run on a non-ADP (``enable_attention_dp: false``) MegaMoE
      config to confirm ``self.max_num_tokens`` is unchanged in that
      path.
- [x] L0 / DSv4 unit tests under
      ``tests/unittest/_torch/modules/test_fused_moe.py`` and
      ``tests/unittest/_torch/multi_gpu_modeling/test_deepseek.py`` —
      none of them exercise the ADP+EP>1 path covered by the fix, but
      keep them green.

### DIFF
--- a/tensorrt_llm/_torch/modules/fused_moe/mega_moe/mega_moe_deepgemm.py
+++ b/tensorrt_llm/_torch/modules/fused_moe/mega_moe/mega_moe_deepgemm.py
@@ -298,9 +298,7 @@ class MegaMoEDeepGemm(MoE):
         # max_num_tokens``, doubling the EP factor and exploding HBM (see
         # ``layout::Workspace`` / ``get_num_max_pool_tokens`` in DG).
         if self.use_dp and self.ep_size > 1:
-            self.max_num_tokens = max(
-                1, (self.max_num_tokens + self.ep_size - 1) // self.ep_size
-            )
+            self.max_num_tokens = max(1, (self.max_num_tokens + self.ep_size - 1) // self.ep_size)
 
         # Resolve the EP ProcessGroup at module construction — creating a
         # group at forward time would be collective on a non-synchronous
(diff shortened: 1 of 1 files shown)


Barry-Delaney (approved):

lfr-0531:
/bot skip --comment "CI passed before, and only changed the code format in the latest commit."

PR #14227, opened 2026-05-17
[None][test] Add verl post-merge E2E gates (0.5B convergence + 7B Megatron smoke)

shikicloud:
## Description

Adds two E2E training tests to the existing `DGX_B200-4_GPUs-Verl-Post-Merge-1` Jenkins stage, layered directly on top of #14037.

#14037 bumps the verl pin to `d324b01` and replaces the three file-level verl wrappers with 19 per-case wrappers. The verl pin is unchanged in this PR.

Until this PR, the verl post-merge stage only ran rollout unit tests; no end-to-end training test exercised the trainer ↔ rollout boundary owned by TRT-LLM.

This PR adds:

- **`test_e2e_convergence_qwen25_05b_grpo_trtllm`**
  - Runs 30-step GRPO on Qwen2.5-0.5B-Instruct + GSM8K + TRT-LLM rollout.
  - Asserts:
    - `max(critic/rewards/mean) > 0.01` to verify a learning signal.
    - `max(actor/ppo_kl) < 0.1` to verify the trainer policy stays close to the rollout policy.
  - This catches weight-sync, dtype, and NaN-gradient regressions at the TRT-LLM rollout boundary.

- **`test_e2e_smoke_qwen25_7b_megatron_trtllm`**
  - Runs a 3-step Megatron actor + TRT-LLM rollout smoke test on Qwen2.5-7B-Instruct.
  - Configuration:
    - Megatron actor: TP=2, PP=1
    - TRT-LLM rollout: TP=2
    - Script: `examples/grpo_trainer/run_qwen3_8b_megatron.sh`
  - Pass condition: subprocess exits with code `0`.
  - Exercises the most feature-loaded combination currently shipped by verl: Megatron + TP/PP + TRT-LLM.

## Stacking

This PR is **stacked on #14037**.

GitHub will show both commits in the diff until #14037 merges. After #14037 is merged, the diff will automatically recompute to only the eight files added or modified in this PR.

Please merge #14037 first.

## Release Notes

* **New Features**
  * Added end-to-end GRPO training convergence validation tests.
  * Added smoke test for Qwen model training workflow.

* **Tests**
  * Expanded integration test suite with comprehensive coverage for placement groups, resource pools, async operations, and rollout scenarios.
  * Reorganized test structure with granular test cases for improved maintainability.

* **Chores**
  * Updated model configuration and pinned new model version.

[...]

PR #14037, opened 2026-05-12
[None][test] Split verl tests into 19 fine-grained per-case wrappers

Superjomn:
## Summary
- Replace 3 file-level verl wrappers with 19 fine-grained per-case wrappers in `tests/integration/defs/verl/test_verl_cases.py` (one wrapper per upstream verl pytest case)
- Bump verl pin: `4cda6af` -> `d324b01`
- Update `l0_verl` test list to enumerate the 19 case-level entries
- Drop the 3 blanket `nvbugs/5981833` waives that were skipping the old file-level wrappers

## Why
File-level wrappers force us to skip an entire upstream test file when a single case is flaky. Per-case wrappers let us waive precisely.

## Test plan
- [x] Local: `pytest --collect-only tests/integration/defs/verl/test_verl_cases.py` -> 19 collected
- [x] Local: 4 pure HTTP-mock adapter cases pass when run directly against a verl clone at `d324b01` (`test_make_async_request_{get,post,http_error,max_attempts_exceeded}_method`)
- [x] Verified all 19 case names are unique across the 5 verl test files - no `-k` substring collisions
- [x] CI: `/bot run --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1"` to exercise the 4xB200 `l0_verl` post-merge stage

## Notes
The `verl_setup` fixture builds gdrcopy + nvshmem + DeepEP and needs root for `/usr/local/lib`. It runs cleanly in CI containers but cannot run on a non-root dev shell, so full validation depends on CI.

## Summary by CodeRabbit

* **Tests**
  * Refactored integration test suite to use individual test functions for finer-grained control and coverage.
  * Expanded test cases with additional scenarios for placement groups, async operations, multimodal generation, and inter-node rollout.

* **Chores**
  * Updated test infrastructure configuration and removed obsolete test entries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14037)

PR #14229, opened 2026-05-17
[None][test] Add verl standalone E2E weight-sync guard

shikicloud:
Adds `test_verl_E2E_standalone.py` to the verl post-merge CI stage. This is layered on top of #14227, which itself stacks on #14037.

The test runs ~3 GRPO trainer steps via verl’s `tests/special_e2e/run_fully_async_policy.sh` on Qwen2.5-0.5B-Instruct + GSM8K, using an FSDP2 trainer on 2 GPUs and standalone TRT-LLM rollout on 2 GPUs. It asserts that each step’s `actor/ppo_kl < 0.05`, meaning the trainer and rollout policies stay within ±5% of each other under `bypass_mode=True`.

This is tighter than #14227’s convergence test, which uses a 0.1 bound. With `bypass_mode=True`, `old_log_probs` come directly from the rollout engine, so `actor/ppo_kl` directly measures `KL(pi_trainer || pi_rollout)`. With bit-correct weight sync, the ratio is theoretically 1; floating-point and dtype-cast noise stay under 5%. A weight-sync regression, such as a broken parameter group, dtype mismatch, or dropped allreduce, drives `ppo_kl` well above 0.05 on the first PPO step.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end convergence test for GRPO training with Qwen2.5-0.5B-Instruct actor and TRT-LLM rollout backend.
  * Added end-to-end smoke test for Megatron actor configuration with TRT-LLM rollout.
  * Added end-to-end standalone async-RL test with policy validation.
  * Expanded test coverage with granular per-case tests for rollout functionality, async request handling, and multimodal scenarios.

* **Chores**
  * Updated CI configuration with new test models and dependencies.
  * Refactored test infrastructure for improved test granularity and maintainability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14229?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


shikicloud:
/bot run --stage-list "DGX_B200-4_GPUs-Verl-Post-Merge-1" --disable-fail-fast

tests/integration/test_lists/test-db/l0_verl.yml
```diff
@@ -15,6 +15,25 @@ l0_verl:
       backend: verl
       orchestrator: mpi
```
hchings, line 16: @Superjomn why mpi here? Also these convergence test doesn't need b200. Do we have H100s for these test?

tests/integration/test_lists/waives.txt
```diff
@@ -494,6 +502,3 @@ unittest/llmapi/test_llm_pytorch.py::test_nemotron_nas_lora[cuda_graph_config0]
 unittest/llmapi/test_memory_profiling.py::test_profile_kvcache SKIP (https://nvbugs/5580781)
```
hchings, line 442: Why the unwaive at L505-507? Is this nvbug resolved in recent MRs or this MR already?

tests/integration/defs/verl/test_verl_E2E_convergence.py
```diff
@@ -0,0 +1,75 @@
...
+
+import os
+
+import pytest
+
+from .test_verl_cases import _check_convergence, _ensure_gsm8k, _run_verl_train
+
+
+@pytest.mark.timeout(1800)
+def test_e2e_convergence_qwen25_05b_grpo_trtllm():
```
hchings: This test is testing sync instead of async RL as we discussed. 
Can you please change it to test async RL instead? 

Refer to https://github.com/verl-project/verl/blob/main/verl/experimental/fully_async_policy/shell/grpo_8b_base_math_megatron_4_4_trtllm.sh for a config, tune it down for qwen2.5-0.5B.

tests/integration/defs/verl/test_verl_E2E_convergence.py
```diff
@@ -0,0 +1,75 @@
...
+            "trainer.logger=['console']",
+            "trainer.n_gpus_per_node=4",
+            "trainer.nnodes=1",
+            "trainer.total_training_steps=30",
+            "trainer.save_freq=-1",
+            "trainer.test_freq=-1",
+        ],
+        timeout=1800,
+    )
+    _check_convergence(log_file, target=0.01, ppo_kl_max=0.1)
```
hchings: Can we also check ppl ratio < a number? e.g., `1.1`? 

This will help catching logprob errors.

tests/integration/defs/verl/test_verl_E2E_convergence.py
```diff
@@ -0,0 +1,75 @@
...
+            "actor_rollout_ref.actor.ppo_mini_batch_size=8",
+            "actor_rollout_ref.actor.ppo_micro_batch_size_per_gpu=1",
+            "actor_rollout_ref.rollout.name=trtllm",
+            "actor_rollout_ref.rollout.tensor_model_parallel_size=1",
+            "actor_rollout_ref.rollout.n=4",
+            "actor_rollout_ref.rollout.max_num_seqs=16",
+            # 16 seqs × (512 prompt + 512 response) = 16384 worst-case
+            # batched tokens; bump from the trtllm rollout default (1024)
+            # so generation isn't artificially serialized.
+            "actor_rollout_ref.rollout.max_num_batched_tokens=16384",
```
hchings: Do we need this or it can be lower?

tests/integration/defs/verl/test_verl_E2E_convergence.py
```diff
@@ -0,0 +1,75 @@
...
+            "actor_rollout_ref.rollout.max_model_len=1024",
+            "actor_rollout_ref.rollout.gpu_memory_utilization=0.6",
+            "actor_rollout_ref.ref.log_prob_micro_batch_size_per_gpu=1",
+            "actor_rollout_ref.rollout.log_prob_micro_batch_size_per_gpu=1",
+            "trainer.logger=['console']",
+            "trainer.n_gpus_per_node=4",
+            "trainer.nnodes=1",
+            "trainer.total_training_steps=30",
```
hchings: I think 10 steps would be sufficient.

tests/integration/defs/verl/test_verl_E2E_feature.py
```diff
@@ -0,0 +1,74 @@
...
+            "ROLLOUT_TP": "2",
+            "NGPUS_PER_NODE": "4",
+            "NNODES": "1",
+            "INFER_BACKEND": "trtllm",
+            "MODEL_PATH": model,
+        }
+    )
+    cmd = [
+        "bash",
+        os.path.join(VERL_ROOT, "examples/grpo_trainer/run_qwen3_8b_megatron.sh"),
```
hchings, line 59: we are already testing this in verl CI. Why makes it needs to be tested again in TRTLLM repo? Or am I missing some context?

tests/integration/defs/verl/test_verl_E2E_standalone.py
```diff
@@ -0,0 +1,164 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES.
```
hchings: I'll prefer that we have a single test_e2e_convergence.py that covers both sync and async case with one test correspondingly.

tests/integration/defs/verl/test_verl_E2E_standalone.py
```diff
@@ -0,0 +1,164 @@
...
+# All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""E2E weight-sync correctness guard for verl STANDALONE-mode async-RL.
+
+Runs the verl-shipped ``tests/special_e2e/run_fully_async_policy.sh`` script
```
hchings: As mentioned, please use https://github.com/verl-project/verl/blob/main/verl/experimental/fully_async_policy/shell/grpo_8b_base_math_megatron_4_4_trtllm.sh instead.

tests/integration/defs/verl/test_verl_E2E_standalone.py
```diff
@@ -0,0 +1,164 @@
...
+            "MODEL_ID": "Qwen/Qwen2.5-0.5B-Instruct",
+            "MODEL_PATH": model_path,
+        }
+    )
+
+    script = os.path.join(VERL_ROOT, "tests/special_e2e/run_fully_async_policy.sh")
+    overrides = [
+        # Use the CI-resolved GSM8K parquets instead of the script's default
+        # $HOME/data/gsm8k.
+        f"data.train_files=['{data_dir}/train.parquet']",
```
hchings: Do we have these parquets available already in our CI env?

tests/integration/defs/verl/test_verl_E2E_standalone.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+    values = _parse_actor_ppo_kl(log_file)
+    assert len(values) >= EXPECTED_STEPS, (
+        f"expected >= {EXPECTED_STEPS} actor/ppo_kl readings, got {len(values)}: "
+        f"{values}; see {log_file}"
+    )
+
+    for step_idx, kl in enumerate(values[:EXPECTED_STEPS], start=1):
+        assert kl < PPO_KL_MAX, (
```
hchings: I think kl < PPO_KL_MAX is too easy to meet, and likely won't capture convergence issue. 
It'd rather that we run a real wandb of the same config for sync & async you have in these new tests, and hard-coded the ppo kl or other key convergence metric's value. and make sure the values generated from the tests don't diverge from historical values too much. 

Also same as sync test, please also test ppl ratio. For ppl ratio, I'm more comfortable using a naive check to ensure it's lower than `1.1`.

shikicloud:
/bot run --stage-list "DGX_B200-4_GPUs-Verl-Post-Merge-1" --disable-fail-fast

shikicloud:
/bot run --stage-list "DGX_B200-4_GPUs-Verl-Post-Merge-1" --disable-fail-fast

shikicloud:
/bot run --stage-list "DGX_B200-4_GPUs-Verl-Post-Merge-1" --disable-fail-fast

tests/integration/defs/verl/test_verl_E2E_convergence.py
shikicloud: Added on the latest commit.

tests/integration/defs/verl/test_verl_E2E_convergence.py
shikicloud: Changed to 1024, which aligns with veRL's ci test.

tests/integration/defs/verl/test_verl_E2E_convergence.py
shikicloud: Changed to 10 steps.

tests/integration/defs/verl/test_verl_E2E_feature.py
shikicloud: I think we still need to test on TRTLLM side since we need to check out the latest and related TRTLLM's features match the fixed veRL version.

tests/integration/defs/verl/test_verl_E2E_standalone.py
shikicloud: Combined to one file.

tests/integration/defs/verl/test_verl_E2E_standalone.py
shikicloud: The current PR only asserts max(critic/rewards/mean) > 0.01 and max(exp|ppo_kl|) < 1.05 (PPL near 1) and both thresholds are based on my local experiments.

tests/integration/defs/verl/test_verl_E2E_standalone.py
shikicloud: We need to download them, and veRL also downloads them while testing.

shikicloud:
/bot run --stage-list "DGX_B200-4_GPUs-Verl-Post-Merge-1" --disable-fail-fast

shikicloud:
/bot run --stage-list "DGX_B200-4_GPUs-Verl-Post-Merge-1" --disable-fail-fast

shikicloud:
/bot run --stage-list "DGX_B200-4_GPUs-Verl-Post-Merge-1" --disable-fail-fast

shikicloud:
/bot run --stage-list "DGX_B200-4_GPUs-Verl-Post-Merge-1" --disable-fail-fast

tests/integration/test_lists/test-db/l0_verl.yml
Superjomn: Just a dummy word here to reuse the MPI resources.

tests/integration/defs/verl/verl_config.yml
hchings, line 5: This is from May 12th. Let's bump it to latest ToT main of verl.

hchings, line 20: cc @Superjomn  - this can be in a separate MR for improvement. 

I think we should directly use our latest public TRTLLM image https://github.com/verl-project/verl/blob/main/docker/Dockerfile.stable.trtllm in TRTLLM CI, instead of these manual version control, to avoid accidentally testing two different envs.

tests/integration/defs/verl/test_verl_E2E_convergence.py
hchings, line 53: For all convergence tests, is it feasible to use `Qwen3-8B` instead in terms of still having a reasonable testing duration and using 2 B200 in this test? cc @Superjomn. Our public async script for 8B uses 2 nodes, I'm not sure whether single node will fit.

Previously, when there was weight update issues, the critic/score/mean and ppl_ratio are way off even in first few steps and it's easy to catch, so 2-3 steps would be fine as a CI check. 

I'm concerned that 0.5B model is way to small to look at its convergence or meaningfully test weight update in standalone mode against any parallel config. @shikicloud do you have the actual wandb of this model & setting for several steps so we can look at its curves?

tests/integration/test_lists/test-db/l0_verl.yml
hchings, line 10: Side question to @Superjomn  - why do we need b200 specifically for these tests? Is it to test specific envs? I'm fine with it but if CI resources is tight, I wonder whether we should downgrade to H100/H200s?

tests/integration/defs/verl/test_verl_E2E_convergence.py
hchings, line 100: Please change all convergence tests to using megatron trainer.  
Verl team has been primarily focusing on megtaron.

tests/integration/defs/verl/test_verl_cases.py
hchings, line 208: Since this dataset don't change, are we able to statically pre-downalod it and persist it somewhere in TRTLLM CI infra to reduce CI time? @Superjomn

tests/integration/defs/verl/test_verl_E2E_convergence.py
hchings, line 120: Please change to megatron.

tests/integration/defs/verl/test_verl_E2E_feature.py
hchings: Got it. But doesn't the convergence tests added in this MR also test the e2e flow for both sync and async RL already?

tests/integration/defs/verl/test_verl_E2E_standalone.py
hchings: Can we check with CI team on whether we can persistently host this dataset somewhere in CI infra? TRTLLM's CI infra is more sophisticated than verl's so there might already be a practice for this.

tests/integration/defs/verl/test_verl_E2E_convergence.py
hchings: I think once we look at the 0.5B wandb curve, we can conclude that whether testing convergence on 0.5B is sufficient for now.

PR #14241, opened 2026-05-18
[None][chore] cap RoPE cos/sin table to runtime max_seq_len

lancelly:
## Summary

Split out only the **RoPE cos/sin table cap** from #14053 / #14241 as an independent, **always-on** optimization. 

### RoPE cos/sin table cap (`attention_backend/interface.py`, `models/modeling_deepseekv4.py`)

`RopeParams.from_config` auto-caps `max_positions` to `max_seq_len` when the latter is set, positive, and smaller than `max_positions`. YaRN cos/sin at position `p` depends only on `p` and the YaRN scaling constants (which use `original_max_position_embeddings`, NOT the table size), so truncating to `max_seq_len` yields **bitwise-identical** values for indices `< cap`. The cap is applied via field mutation so it participates in the cache key (preserving dedup semantics) and downstream kernel-facing fields stay consistent.

`modeling_deepseekv4.py` plumbs `model_config.max_seq_len` onto the HF `PretrainedConfig` so `from_config` can see it (DeepSeek HF configs don't have a `max_seq_len` attribute by default). The plumb mirrors the existing pattern in `modeling_phi3.py`.

**Runtime safety net:** the existing `TrtllmAttentionWrapper.ensure_rope_table_size` expands the table on demand if a kernel asks for a larger `max_positions` (e.g., MTP draft tokens push position index above the cap). So the cap is a **memory-only optimization**, not a correctness contract. At worst the extended-MTP edge case triggers one resize per attention wrapper (≈4 MiB × num_layers, paid once); the cap still saves vs the un-capped HF default (~20× smaller table for DSv4-Pro).

For DSv4-Pro on GB300: HF default `max_position_embeddings = 163840`, runtime `max_seq_len = 8192` → table shrinks ≈20×.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_deepseekv4.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv4.py
@@ -81,7 +81,7 @@ from ..modules.mhc.hyper_connection import HCHead, HCState, mHC
 from ..modules.multi_stream_utils import maybe_execute_in_parallel
 from ..modules.rms_norm import RMSNorm
 from ..peft.lora.layer import LoraLayer
-from ..speculative import SpecMetadata
+from ..speculative import SpecMetadata, get_num_extra_kv_tokens
 from ..utils import (
     AuxStreamType,
     EventType,
@@ -149,9 +149,22 @@ def weight_dequant(x: torch.Tensor, s: torch.Tensor, block_size: int = 128) -> t
 
 
 def _deepseek_v4_pos_embd_params(
-    config: PretrainedConfig, model_config: ModelConfig, layer_idx: Optional[int]
+    config: PretrainedConfig,
+    model_config: ModelConfig,
+    predicted_tokens_per_seq: int,
+    layer_idx: Optional[int],
 ) -> PositionalEmbeddingParams:
-    rope_params = RopeParams.from_config(config)
+    # Match the spec/overlap headroom that py_executor_creator applies
+    # to model_engine_max_seq_len before sizing the KV cache. The overlap
+    # branch is included unconditionally because disable_overlap_scheduler
+    # isn't reachable here; the overshoot when overlap is actually off is
+    # at most predicted_tokens_per_seq positions (= 1 + max_total_draft_tokens,
+    # which can exceed max_draft_len for tree-shaped drafts).
+    runtime_max_seq_len = model_config.max_seq_len
+    if runtime_max_seq_len is not None and model_config.spec_config is not None:
+        runtime_max_seq_len += 2 * max(0, predicted_tokens_per_seq - 1)
+        runtime_max_seq_len += get_num_extra_kv_tokens(model_config.spec_config)
+    rope_params = RopeParams.from_config(config, runtime_max_seq_len=runtime_max_seq_len)
 
     compress_ratios = None
     if model_config.sparse_attention_config is not None:
@@ -1292,7 +1305,9 @@ class DeepseekV4Attention(MLA):
             predicted_tokens_per_seq=predicted_tokens_per_seq,
             max_position_embeddings=config.max_position_embeddings,
             bias=False,
-            pos_embd_params=_deepseek_v4_pos_embd_params(config, model_config, layer_idx),
+            pos_embd_params=_deepseek_v4_pos_embd_params(
+                config, model_config, predicted_tokens_per_seq, layer_idx
+            ),
             layer_idx=layer_idx,
             dtype=config.torch_dtype,
             config=model_config,
--- a/tensorrt_llm/_torch/models/modeling_deepseekv4.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv4.py
@@ -151,8 +151,8 @@ def weight_dequant(x: torch.Tensor, s: torch.Tensor, block_size: int = 128) -> t
 def _deepseek_v4_pos_embd_params(
     config: PretrainedConfig,
     model_config: ModelConfig,
-    predicted_tokens_per_seq: int,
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/models/modeling_deepseekv4.py
```diff
@@ -149,8 +149,18 @@ def weight_dequant(x: torch.Tensor, s: torch.Tensor, block_size: int = 128) -> t
...
+    predicted_tokens_per_seq: int,
+    layer_idx: Optional[int],
 ) -> PositionalEmbeddingParams:
+    # Pass max_seq_len to config so RopeParams.from_config can auto-cap
+    # max_positions to runtime max_seq_len. Add MTP draft headroom so
+    # speculative draft positions (which may overshoot the accepted
+    # boundary by up to predicted_tokens_per_seq - 1) still fit without
+    # triggering ensure_rope_table_size resize.
+    if model_config.max_seq_len is not None:
+        config.max_seq_len = model_config.max_seq_len + max(0, predicted_tokens_per_seq - 1)
```
pengbowang-nv: 1. This changed the pretrained config. I doubt if this would cause some consequences for other operations following RoPE param construction that also uses max_seq_len, for example following layers of Attn. I suggest change this logic and deduce rope runtime max_seq_len in `RopeParams.from_config`.

2. About previously said if max_seq_len + predicted_tokens_per_seq - 1 would be enough, I (let ai) check again and found that for overlap scheduler, we need twice the size. See `tensorrt_llm/_torch/pyexecutor/py_executor_creator.py:599-606`. And there is also a `get_num_extra_kv_tokens(spec_config)`, might be relevant.

pengbowang-nv (changes requested):

tensorrt_llm/_torch/models/modeling_deepseekv4.py
```diff
@@ -149,9 +149,28 @@ def weight_dequant(x: torch.Tensor, s: torch.Tensor, block_size: int = 128) -> t
...
+    # Derive the runtime cap for the RoPE cos/sin table from model_config
+    # without mutating the shared HF config. The cap mirrors the
+    # max_seq_len adjustments applied in py_executor_creator.py:599-606:
+    #   - Spec decoding contributes (tokens_per_gen_step - 1) draft positions
+    #     plus get_num_extra_kv_tokens(spec_config) extra KV slots.
+    #   - The overlap scheduler (when enabled alongside spec) contributes
+    #     another tokens_per_gen_step - 1 positions.
+    # disable_overlap_scheduler isn't reachable at attention construction
+    # time, so include the overlap headroom unconditionally when spec is
+    # on; the overshoot is at most max_draft_len positions, negligible
```
pengbowang-nv: NIT: should be at most predicted_tokens_per_seq . predicted_tokens_per_seq  = 1 + max_total_draft_tokens (which may > max_draft_len for tree drafts).

tensorrt_llm/_torch/models/modeling_deepseekv4.py
```diff
@@ -149,9 +149,28 @@ def weight_dequant(x: torch.Tensor, s: torch.Tensor, block_size: int = 128) -> t
...
+    # relative to the cos/sin table savings.
+    runtime_max_seq_len = model_config.max_seq_len
```
pengbowang-nv, line 163: Is this guaranteed to have? I see `getattr(config, "max_seq_len", None)` in other cases.

tensorrt_llm/_torch/models/modeling_deepseekv4.py
```diff
@@ -149,9 +149,28 @@ def weight_dequant(x: torch.Tensor, s: torch.Tensor, block_size: int = 128) -> t
...
+    if runtime_max_seq_len is not None and model_config.spec_config is not None:
+        runtime_max_seq_len += 2 * max(0, predicted_tokens_per_seq - 1)
+        runtime_max_seq_len += get_num_extra_kv_tokens(model_config.spec_config)
```
yuxianq, line 166: We have already added `predicted_tokens_per_seq`, why we need to add another `get_num_extra_kv_tokens(model_config.spec_config)`?

tensorrt_llm/_torch/models/modeling_deepseekv4.py
lancelly: Fixed, thanks

tensorrt_llm/_torch/models/modeling_deepseekv4.py
lancelly: Fixed.

tensorrt_llm/_torch/models/modeling_deepseekv4.py
lancelly: Fixed.

tensorrt_llm/_torch/models/modeling_deepseekv4.py
lancelly: Fixed

tensorrt_llm/_torch/models/modeling_deepseekv4.py
lancelly: Followed:
```
py_executor_creator.py:599-606
    if not llm_args.disable_overlap_scheduler and spec_config is not None:
        model_engine_max_seq_len += spec_config.tokens_per_gen_step - 1

    if spec_config is not None:
        model_engine_max_seq_len += get_num_extra_kv_tokens(spec_config)
        model_engine_max_seq_len += spec_config.tokens_per_gen_step - 1
```

PR #14250, opened 2026-05-18
[None][perf] fuse quant into norm and rope

mingyangHao:
@coderabbitai summary

## Description

| Kernel                                                                                                       | pre-change   | post-change | delta              |
| ------------------------------------------------------------------------------------------------------------ | ------------ | ----------- | ------------------ |
| `deepseekV4QNormKernel<bf16, 512>`                                                                            | **77.5 µs**  | —           | —                  |
| `deepseekV4QNormFusedKernel<bf16, 512, 448>` (FP8 nope STG inlined)                                           | —            | **62.1 µs** | **−15.4 µs**       |
| `applyMLARopeAndAssignQKVKernelOptContext<..., kOutputFp8Q=false>` (bf16 rope STG to q_buf)                  | **19.1 µs**  | —           | —                  |
| `applyMLARopeAndAssignQKVKernelOptContext<..., kOutputFp8Q=true>` (FP8 rope STG to quant_q_buf)              | —            | **21.6 µs** | **+2.5 µs**        |
| `quantizeCopyInputToFp8Kernel<bf16, 256, 448, 64, 512, true>` (bf16 LDG full Q → FP8 STG full Q)             | **73.3 µs**  | (eliminated) | **−73.3 µs**       |
| **Sum of kernels**                                                                                            | **169.9 µs** | **83.7 µs**  | **−86.2 µs**       |
| **End-to-end (Python event timing, incl. host launch overhead)**                                              | **212.3 µs** | **89.3 µs**  | **−123.0 µs**      |

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1734,6 +1734,41 @@ def _deepseek_v4_q_b_layernorm(self, q: torch.Tensor) -> torch.Tensor:
...
+        quant_q, q_pe = torch.ops.trtllm.deepseek_v4_q_norm_fused_fp8(
+            q_proj,
+            self.num_heads_tp,
+            self.qk_head_dim,
+            self.kv_lora_rank,  # nope_dim
+            float(self.q_b_layernorm.variance_epsilon),
+            self._fused_q_fp8_scale,
+            True,  # interleaved_quant_q -> stride=head_dim
+        )
+        dummy = torch.empty_like(q_proj)
```
lfr-0531: Can we return a `None`, not an empty tensor here?

liji-nv (approved):

tensorrt_llm/_torch/modules/attention.py
mingyangHao: Done, directly pass q_proj out here, and use a placeholder_q for clarification. This should be safe since q_proj value is never used on fused path.

heyuhhh (approved):
LGTM～ BTW, do we need to keep the unfused version? I think we can just keep the fused version for V4. Other changes are also general enough for other situations, which can be reused later

mingyangHao:
> LGTM～ BTW, do we need to keep the unfused version? I think we can just keep the fused version for V4. Other changes are also general enough for other situations, which can be reused later

I agree. Maybe lets keep this version now, and later raise another PR to deprecate the unfused path for stability. @lfr-0531 What do you think?

lfr-0531:
> I think we can just keep the fused version for V4. Other changes are also general enough for other situations, which can be reused later

Agree. For DS-V4, we can only keep the fused version.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -15,23 +15,17 @@
...
-    RopeParams,
-    merge_attention_forward_args,
-)
+from ..utils import (compute_swizzled_sf_shape, get_global_attrs,
+                     get_model_extra_attrs)
+from .interface import (AttentionBackend, AttentionForwardArgs,
+                        AttentionInputType, AttentionMask, AttentionMetadata,
+                        KVCacheParams, MLAParams, PositionalEmbeddingParams,
+                        PredefinedAttentionMask, RopeParams,
+                        merge_attention_forward_args)
```
yuxianq, line 28: Use `merge_attention_forward_args,` to recover the original style

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -3089,11 +3160,16 @@ def forward_absorption_context(
             out_scale=self.out_scale,
             output=output if self.is_deepseek_v4 else None,
             latent_cache=latent_cache,  # kvcache and k_pe
-            q_pe=q_pe,  # used by `invokeMLARopeGeneration`
+            q_pe=
+            extra_q_pe,  # used by `invokeMLARopeGeneration` / RoPE_OptContext
```
yuxianq: What is `RoPE_OptContext`? I cannot find it.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1734,6 +1734,45 @@ def _deepseek_v4_q_b_layernorm(self, q: torch.Tensor) -> torch.Tensor:
...
+        # are either overridden by the caller or discarded by the DSv4 branch.
+        # Reusing `q_proj` (q_b_proj output) avoids a ~num_tokens × hidden bf16
+        # allocation per forward.
+        assert q_proj.dim() == 2
+        assert q_proj.shape[1] == self.num_heads_tp * self.qk_head_dim
+        if getattr(self, "_fused_q_fp8_scale", None) is None:
+            self._fused_q_fp8_scale = torch.tensor([1.0],
+                                                   dtype=torch.float32,
+                                                   device=q_proj.device)
+        quant_q, q_pe = torch.ops.trtllm.deepseek_v4_q_norm_fused_fp8(
```
yuxianq: We allocate a new tensor `quant_q` here. We already have a same shape `quant_q_buffer` in MLA, can we directly write to it in `deepseek_v4_q_norm_fused_fp8`? `interleaved_quant_q` is unnecessary in this case, since we can use `quant_q_buffer`'s shape as row stride.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -3089,11 +3160,16 @@ def forward_absorption_context(
             out_scale=self.out_scale,
             output=output if self.is_deepseek_v4 else None,
             latent_cache=latent_cache,  # kvcache and k_pe
-            q_pe=q_pe,  # used by `invokeMLARopeGeneration`
+            q_pe=
+            extra_q_pe,  # used by `invokeMLARopeGeneration` / RoPE_OptContext
+            quant_q_buffer=extra_quant_q_buffer,  # fused-FP8 path only
```
yuxianq: We use 4 names `extra_quant_q_buffer/quant_q_buffer/fused_quant_q/_fused_q_fp8_quant_q` for the same tensor, can we unify it? How about use `quant_q_buffer` directly? In `if use_fused_q_fp8:` block, we can use `quant_q_buffer = quant_q_buffer[:num_tokens].view(xxx)`. We can also always use `quant_scale_qkv` and `q_pe` as name.

tensorrt_llm/_torch/attention_backend/trtllm.py
mingyangHao: Reverted trtllm.py to upstream — the PR did not actually need any changes here (an earlier yapf pass had reformatted the import wrap; gone now).

tensorrt_llm/_torch/modules/attention.py
mingyangHao: Sorry, my shorthand for applyMLARopeAndAssignQKVKernelOptContext in cpp/tensorrt_llm/kernels/mlaKernels.cu. Renamed all references to the full name in the latest push.

tensorrt_llm/_torch/modules/attention.py
mingyangHao: Good point, refactored. `deepseek_v4_q_norm_fused_fp8 `is now in-place — caller pre-allocates `quant_q_out + q_pe_out `and the op writes into them. Per-head stride (nope_dim packed vs head_dim interleaved-with-rope-slot) is inferred from `quant_q_out.shape[1] / num_heads`, so the `interleaved_quant_q` flag is gone.

tensorrt_llm/_torch/modules/attention.py
mingyangHao: Done. Unified to `quant_q_buffer` / `q_pe` / `quant_scale_qkv` throughout `forward_absorption_context`. In the` if use_fused_q_fp8:` block, `quant_q_buffer` is now reassigned to its sliced view directly instead of going through an `extra_*` alias. Instance attrs are now `self._fused_quant_q_buffer` / `self._fused_q_pe` / `self._quant_scale_qkv`.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -3078,6 +3136,26 @@ def forward_absorption_context(
 
         # Use generation_only for generation phase and context_only for context phase in DSA attention
         attention_input_type = AttentionInputType.context_only
+
+        # Fused FP8-Q path: forward the pre-quantized buffers stashed in
+        # `_q_branch`; the C++ op enables fusion when both are non-None.
+        quant_q_buffer = getattr(self, "_fused_quant_q_buffer", None)
+        q_pe_fused = getattr(self, "_fused_q_pe", None)
```
yuxianq: Keep the name consistent, we can use `fused_q_pe` instead of `q_pe_fused`.

yuxianq (approved):

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: We still need to pass `quant_scale_qkv` in trtllm.py, right?

mingyangHao: Yes you are indeed correct, I have add it in the new commit.

tensorrt_llm/_torch/modules/attention.py
mingyangHao: Updated.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1470,6 +1470,14 @@ def _run(
                 global_layer_idx=self.layer_idx,
             )
         else:
+            # The fused DSv4 FP8-Q-quant path requires both `quant_q_buffer`
+            # and `quant_scale_qkv` to reach the C++ op together. Without the
+            # scale the C++ side silently drops back to the BF16 quant kernel
+            # while Python keeps feeding it the pre-LayerNorm placeholder Q,
+            # producing garbage outputs (see CI #39174).
+            assert (forward_args.quant_q_buffer is None) == (
```
yuxianq: fp8 kv cache also use `quant_q_buffer`, but it does not require `quant_scale_qkv`, so this assertion will fail the fp8 kv cache case. This condition should be if `quant_scale_qkv` is not None then `quant_q_buffer` must be not None. In addition, please don't add this assertion specifically for thop.attention, all checks should be finished before we dispatch between trtllm_gen_attention and thop.attention.

mingyangHao: Done, move this assertion before dispatch. This assertion now check ` quant_scale_qkv is None or quant_q_buffer is not None` to allow quant_q_buffer standalone, but not scale.

PR #14254, opened 2026-05-18
[None][perf] DSv4 o_a_proj: enable fused inv-RoPE+FP8 BMM by default + optimize kernel

lishicheng1996-nv:
## Description

This PR does two things on DSv4's `o_a_proj` MLA path.

### 1. Enable fused inv-RoPE + FP8 BMM for `o_a_proj` by default

Removes `self.use_cute_dsl_blockscaling_bmm` from the two gates in `MLA._deepseek_v4_o_proj` (init-time FP8 weight retention + runtime kernel selection) so DSv4 unconditionally uses the FP8-native `o_a_proj` + fused inv-RoPE + cute-dsl FP8 BMM chain on SM100. The flag still controls K/V absorption BMM kernel selection (`k_b_proj_trans`, `v_b_proj`) for both DSv3 and DSv4 — that is a separate kernel-choice concern and is unaffected here.

### 2. Optimize the fused inv-RoPE + FP8 quant kernel

Making the fused chain the default means the kernel now runs on every DSv4 deployment. The original (vLLM-port Triton) was tuned for a small-M GEN-side regime and underperformed at the CTX-side M (≥ 2048). This PR refreshes the kernel with the changes below:

- **BTM dispatch (1 / 8 / 16 / 32 by M-threshold)** — multi-token-per-block batching closes the small-M and large-M perf gaps the original single-config kernel had.
- **Two `@triton.jit` functions, one each for BTM=1 and BTM>1** — Triton recompiles differently with an outer loop scaffold even at constexpr BTM=1, so physically splitting the two regimes avoids a small-M regression a single-kernel variant showed.
- **`num_warps` / `num_stages` retune** — `nw=1, ns=1` at BTM=1; `nw=2, ns=2` at BTM ≥ 8. Right band for memory-bound block-scale quant kernels: `nw=4` either breaks per-128 coalescing (at BTM=1) or starves SM occupancy via register pressure (at BTM=16).
- **Inner-loop `tl.range(0, BTM, num_stages=2)`** software-pipelines the next iter's load with the current iter's compute and store.

## Microbench results

Measured on **GB200 (B200, HBM3e, ~8 TB/s peak per GPU)**, DSv4-Pro DEP8 shape (`n_groups=8, heads_per_group=16, head_dim=512 (nope=448+rope=64), quant_group=128`). DeepGEMM-style timing (500-iter warmup, single `cuda.Event`-pair elapsed_time / N-iter mean, L2 flush pre-warmup, large fp32 matmul to warm host launch path). **Median across 4 independent same-allocation legs on 4 different gb200 nodes** to suppress per-chip clock-state variance.

Both legs run the same `fused_inv_rope_fp8_quant_vllm_port` op; only the BTM dispatch differs. Unoptimized = the original vLLM-port single-config BTM=1 kernel (the pre-PR baseline). Optimized = the BTM-dispatched + retuned Triton kernel from §2.

| M | Unoptimized µs | Optimized Triton µs | Optimized vs Unoptimized |
|---:|---:|---:|---:|
| 64   | 68.7  | 61.7  | tied (within noise) |
| 256  | 61.1  | 58.7  | tied (within noise) |
| 1024 | 74.2  | 64.5  | **−13.1 % faster** |
| 2048 | 142.0 | 105.3 | **−25.8 % faster** |
| 4096 | 279.8 | 204.2 | **−27.0 % faster** |
| 8192 | 554.4 | 391.7 | **−29.4 % faster** |

At **M ≥ 2048** (the CTX-side regime for DSv4-Pro ISL=8192 workloads) the optimization yields a structural **26-29 % win** over the unoptimized vLLM-port. The optimized Triton kernel saturates B200 HBM at ~4.11 TB/s = **51 %** of the 8 TB/s peak. The unoptimized BTM=1 kernel sits at ~2.9 TB/s = 37 % of peak — the gap closed by the BTM dispatch + retune is reproducible across all 4 allocations.

At **M = 1024** (BTM=8 in the optimized dispatch) the optimization yields a **13 %** win — also reproducible across allocations.

At **M ≤ 256** both Unoptimized and Optimized dispatch the BTM=1 kernel — same code path, so any difference is measurement noise. The cross-allocation spread at this regime (kernel wall ≈ 60-90 µs) is ±15 µs.

`ncu --set full` on the optimized kernel at M=8192 confirms the structural ceiling: **86.5 % Compute SM throughput** / 46.0 % memory throughput / 31.0 % DRAM throughput, **60.1 % achieved occupancy at the 62.5 % theoretical max** (register-limited), `Warp Cycles Per Issued Instruction = 10.98`. The kernel is compute-bound on per-warp instruction-issue with no occupancy headroom to hide load latency; further gains beyond this point would need either a kernel-structure change (e.g. TMA-bulk + warp-spec) or a host-side improvement.

## End-to-end accuracy (gsm8k)

Goal: confirm the PR's "FP8-native o_a_proj + fused inv-RoPE + cute-dsl FP8 BMM" chain — which this PR makes the *unconditional* default for DSv4 on SM100 — produces the same gsm8k accuracy as the pre-PR bf16-fallback path that ran whenever `use_cute_dsl_blockscaling_bmm=False` (the default).

**Setup.** DSv4-Pro on GB200, TP8 EP8, `max_batch=16`, `kv_cache_free_gpu_mem_fraction=0.7`, greedy via `trtllm-eval gsm8k --apply_chat_template`, 1319 questions, 5-shot. Each row is one full eval run.

**A/B comparison** (DEP8, attention DP on — the path most sensitive to the scale-buf padding bug fixed on this branch):

| DEP8 leg | o_a_proj kernel chain | flexible-extract | strict-match |
|---|---|---:|---:|
| **A** | `mla_rope_inplace` → bf16 `torch.bmm` (with bf16-dequant `o_a_proj`) | 96.51 | 96.59 |
| **B** | `fused_inv_rope_fp8_quant_vllm_port` → `cute_dsl_fp8_bmm_blackwell` (FP8 `o_a_proj`) | **96.44** | **96.51** |
| Δ (B − A) | | −0.07 | −0.08 |

Δ is within run-to-run noise (stderr ≈ 0.5 on each score). DEP8 sees no accuracy regression from switching to the FP8 chain.

**TEP8 sanity check** (attention DP off) — only the PR-on number measured since this config has never had the failure mode that triggered the DEP8 investigation:

| TEP8 leg | o_a_proj kernel chain | flexible-extract | strict-match |
|---|---|---:|---:|
| this PR | `fused_inv_rope_fp8_quant_vllm_port` → `cute_dsl_fp8_bmm_blackwell` (FP8 `o_a_proj`) | **96.29** | **96.36** |

Comparable to DEP8's PR-on row; matches expected DSv4-Pro gsm8k performance.

> **A related fix on this branch.** While validating DEP8, an earlier draft of the optimized kernel showed a real accuracy regression (DEP8 dropped to 89.69 flex / 83.32 strict). Root cause: the BTM>1 multi-token kernel padded the scale buffer's m-dim to `pad_up(M, max(BTM, 4))`, but `cute_dsl_fp8_bmm_blackwell` hard-codes `sf_m = pad_up(M, 4)` for the scale stride — drifting wrong scale rows whenever `M % BTM ≠ 0`. The fix aligns producer padding with the consumer contract; `<2 %` perf delta at all M (within microbench noise). The numbers in the table above are *after* that fix.

## Test Coverage

- `tests/unittest/_torch/custom_ops/test_fused_inv_rope_fp8_quant.py` covers the fused op at `num_tokens ∈ {3, 64, 257, 512, 1024, 2048, 8192}`. The op matches the legacy 2-kernel reference (`mla_rope_inplace + fp8_batched_quantize_1x128_permute102`) within ~6e-4 relative diff in dequantized BF16 space — well inside the 1 % bound the test asserts.
- The existing `tests/unittest/_torch/attention/sparse/deepseek_v4/test_deepseek_v4_o_proj.py` exercises the DSv4 `o_a_proj` BMM on SM100. With this PR, that test now hits the FP8-native + fused chain by default (no need to manually set `use_cute_dsl_blockscaling_bmm=True`).
- Manual bench plan: DSv4-Pro 8k-1k pareto sweep (TEP4 / TEP8 / DEP8) with the new default — confirms `o_a_proj` path activates without setting the flag and produces the expected fused-chain device-time profile.

## PR Checklist

## GitHub Bot Help


mingyangHao (approved):
This is a good implementation, I think we can merge this ATM.

lishicheng1996-nv:
Hold since no BW, gsm8k pass with 96+ score shown in PR description, but unnitest value diff need futher evaluate

image

lishicheng1996-nv:
Close it as this optimization is taken by #14446 and merged successfully.

PR #11173, opened 2026-02-01
[None][feat] Enable joint optimization of agent applications and TensorRT-LLM with Scaffolding

Boreas618:
## Release Notes

* **New Features**
  * Added agent-tree based request scheduling for intelligent prioritization in batch processing.
  * Implemented KV cache hints and truncation for optimized memory management in multi-turn conversations.
  * Introduced new chat and multi-turn conversation controllers with tool-calling support.
  * Added MCP (Model Context Protocol) worker integration for agentic workflows.
  * Expanded performance metrics to include KV cache utilization and block tracking.
  * Added multiround chat benchmarking with configurable traffic patterns and conversation distributions.

* **Bug Fixes**
  * Fixed input validation to handle empty lists gracefully.

* **Documentation**
  * Added comprehensive README and examples for benchmarking and deep research workflows.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

## Description

This PR addresses several issues in the Scaffolding-based implementation of Open Deep Research. It adds support for capturing agent application information via Task Collection, enables proactive KV cache eviction to reduce GPU memory pressure from agents, and introduces agent-tree–based performance isolation between agents and chatbots. It also includes benchmarks for evaluating Scaffolding-based agents.

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

`/bot [-h] ['run', 'kill', 'skip', 'reuse-pipeline'] ...`

Provide a user friendly way for developers to interact with a Jenkins server.

[...]

PR #14276, opened 2026-05-18
[None][fix] Handle unset attention_dp_relax in ADP routers

peihu-nv:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed request scheduling to correctly treat unset parameters as permitting relaxed routing mode, ensuring proper request distribution under capacity constraints.

* **Tests**
  * Added unit tests verifying request routing behavior with different parameter configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14276?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Description

Fix an ADP router crash when a request has `SchedulingParams`, but `attention_dp_relax` is unset and causing NoneType comparison. Hit this issue when doing perfwork for DS V4 using inferenceX benchmarking script but both main and DS V4 branch will have this issue. 

 `SchedulingParams.attention_dp_relax` is typed as `Optional[bool]` and defaults to `None`.

  The ADP routers were using this field directly as the sort key:

 ` return scheduling_params.attention_dp_relax`

  This was previously mostly hidden for OpenAI serving because requests often had no `py_scheduling_params`, so the router took the safe path:

```
  if scheduling_params is None:
      return True
```

  After a31d650304e / #11173, OpenAI chat and harmony paths create SchedulingParams(agent_hierarchy=...) and pass it into generation. That means `scheduling_params` is present, but `attention_dp_relax` remains `None`. With multiple such requests, Python sorting can compare `None` values and fail with:
`TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'`

Fix:
Treat only an explicit False as strict and keep None aligned with the existing missing-scheduling-params behavior. Apply the same logic to both DefaultADPRouter and KVCacheAwareADPRouter, with regression coverage for both paths.

## Test Coverage

Added regression coverage for both ADP router implementations where one request has `attention_dp_relax=None `and another has `attention_dp_relax=False`.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
+++ b/tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
@@ -304,7 +304,7 @@ class DefaultADPRouter(ADPRouter):
             scheduling_params = getattr(req_item.request, "py_scheduling_params", None)
             if scheduling_params is None:
                 return True
-            return scheduling_params.attention_dp_relax
+            return scheduling_params.attention_dp_relax is not False
 
         sorted_requests = sorted(new_requests, key=get_relax_value)
 
@@ -576,7 +576,7 @@ class KVCacheAwareADPRouter(ADPRouter):
             scheduling_params = getattr(req_item.request, "py_scheduling_params", None)
             if scheduling_params is None:
                 return True
-            return scheduling_params.attention_dp_relax
+            return scheduling_params.attention_dp_relax is not False
 
         sorted_requests = sorted(new_requests, key=get_relax_value)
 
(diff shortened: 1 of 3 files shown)


tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
```diff
@@ -304,7 +304,7 @@ def get_relax_value(req_item):
             scheduling_params = getattr(req_item.request, "py_scheduling_params", None)
             if scheduling_params is None:
                 return True
-            return scheduling_params.attention_dp_relax
+            return scheduling_params.attention_dp_relax is not False
```
mikeiovine: Would it be better to just make it default to True? Don't like how you have to dig into the code to find the default behavior

mikeiovine (approved):

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
peihu-nv: Sounds good! Made the change.

Tabrizian (approved):

PR #13721, opened 2026-05-04
[TRTLLM-12436][feat] visual_gen: add CuTe DSL attention via exported binaries

xrq-phys:
## Summary by CodeRabbit

* **New Features**
  * Added CuTe DSL attention backend support for visual generation models on Blackwell GPUs.
  * Introduced `--quant_attention_mode` option with `NO_QUANT`, `QK16PV8`, and `SAGE` modes, replacing the previous `--enable_sage_attention` flag.
  * Extended `--attention_backend` to include `CUTEDSL` choice alongside existing options.

* **Documentation**
  * Updated visual generation examples with CuTe DSL backend usage and new quantization mode configurations.

* **Tests**
  * Added new tests for CuTe DSL attention kernel execution and validation.
  * Expanded existing attention tests to cover new backends and quantization configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/13721?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Description

Adds CuTe DSL `fmha.py` exported cubins to the visual_gen workflow as backend option `CUTEDSL`, supporting:
- Ragged attention on 16-bit datatypes
- Ragged attention with Qk16Pv8 (Q & K in BF16, V in FP8, online P converted to FP8)

Currently, only `head_dim=128` is supported due to downstream model requirements.

Integrated into Wan & FLUX.

## Test Coverage

- Added `test_attention_cute_dsl.py` for smoke tests CuTe DSL
- Added tests for the integration to `test_attention_integration.py` and `test_attention_perf.py`

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
[...]

PR #14280, opened 2026-05-19
[TRTLLM-12347][feat] enable VSA in VisualGen

o-stoner:
## Summary by CodeRabbit

* **New Features**
  * Added Video Sparse Attention (VSA) algorithm for visual generation models, enabling efficient sparse attention computation on Blackwell GPUs for supported Wan pipelines.
  * Introduced `flow_shift` parameter to override scheduler configuration in Wan pipelines during inference (allows us to have an apples-to-apples quality comparison with FastVideo, where the Wan pipelines have different `flow_shift` values that what exists by default in the scheduler).
  * Added `VideoSparseAttentionConfig` for controlling VSA sparsity levels.

* **Enhancements**
  * Extended attention mechanisms to support configurable gate tensors for fine-grained attention control.
  * Improved sparse attention validation across multiple pipeline variants (FLUX, FLUX.2, LTX-2, Wan).

## Description

Adds VSA attention backend for TRT-LLM VisualGen based on the following VSA [paper](https://arxiv.org/pdf/2505.13389). Integrates the B200 CuteDSL kernel [here](https://gitlab-master.nvidia.com/cjerry/vsa/-/tree/main/csrc/attn/video_sparse_attn/vsa/blackwell?ref_type=heads). Currently, this backend is supported for Wan 2.1 using the following [fine-tuned model](https://huggingface.co/FastVideo/Wan2.1-VSA-T2V-14B-720P-Diffusers) from FastVideo. This support will be extended to Wan 2.2 T2V 14B / TI2V 5B once ModelOpt fine-tuned weights are ready.

Quality/perf findings are summarized on the page [here](https://sc.talos.nvidia.com/view/home/scratch.ostoner_coreai_1/TensorRT-LLM/examples/visual_gen/wan21_vsa_sweep/outputs/findings.html), and quality against H200 FastVideo with the same input noisy latent/flow_shift value are summarized [here](https://sc.talos.nvidia.com/view/home/scratch.ostoner_coreai_1/TensorRT-LLM/examples/visual_gen/wan21_vsa_b200_vs_h200_comparison/comparison.html).
## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -97,8 +97,17 @@ class SageAttentionConfig(StrictBaseModel):
 class AttentionConfig(StrictBaseModel):
     """Configuration for Attention layers."""
 
-    backend: Literal["VANILLA", "TRTLLM", "FA4"] = PydanticField(
-        "VANILLA", description="Attention backend: VANILLA (PyTorch SDPA), TRTLLM, FA4"
+    backend: Literal["VANILLA", "TRTLLM", "FA4", "VSA"] = PydanticField(
+        "VANILLA", description="Attention backend: VANILLA (PyTorch SDPA), TRTLLM, FA4, VSA"
```
zhenhuaw-me: VSA sounds like algo rather than a backend. @xrq-phys is adding other CuTeDSL base kernels with https://github.com/NVIDIA/TensorRT-LLM/pull/13721 and we are likely to use `CUTEDSL` as the backend keyword, which is aligned with the MoE path.

tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/video_sparse_attention/__init__.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""
+Video Sparse Attention CuTe DSL forward kernel for Blackwell (sm_100).
+
+Vendored kernel (block_sparse_attn_dsl_fwd.py, ptx.py, scheduler.py)
+from https://gitlab-master.nvidia.com/cjerry/vsa — inference-only.
```
zhenhuaw-me: Can we remove this?

xrq-phys:
**Suggested restructuring: reuse the `CUTEDSL` backend (from PR #13721) and split VSA-specific branches into a sparsity sub-config + kernel sub-directory**

Hi, @o-stoner ! I checked in with the @zhenhuaw-me today and we'd like to coordinate the VSA integration so it composes with the CuTe-DSL backend that #13721 is about to land. Below is a concrete restructuring proposal — happy to discuss alternatives if any of these don't fit your kernel's constraints.

**Context (what PR #13721 brings):**
- A new `AttentionConfig.backend = "CUTEDSL"` choice, served by `tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl.py`.
- A directory `tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/blackwell/attention/` holding the packaged cubins (`cubins/`) and a thin Python runner (`fmha.py`) that resolves and launches them.
- A new optional `AttentionConfig.quant_attention_config` sub-config (`Optional[QuantAttentionConfig]`) that turns on QK16PV8 (and, on TRTLLM, SAGE). Backend init reads it; absence means "default behavior".

**Requested changes for #14280:**

1. **Drop the new `"VSA"` backend literal; reuse `"CUTEDSL"`.**
   Given that VSA is a "CuTe DSL attention kernel with sparsity", it can be considered as the same backend. Concretely:
   - In `config.py` (after CuTe DSL lands), VSA does not need to extend `backend`'s `Literal[...]` set — it'll already include `"CUTEDSL"`.
   - Drop `VSAAttentionBackend` as a separate class registered in `get_visual_gen_attention_backend`. The factory keeps returning `CuTeDSLAttention` for `"CUTEDSL"`; that class (implemented in `tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl.py`) becomes the dispatcher between quantized attention and sparse attention.

2. **Move the VSA CuTe DSL kernel source to `tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/blackwell/video_sparse_attention/`.**
   Instead of landing the kernels in `cute_dsl_kernels/video_sparse_attention/`, convention like `cute_dsl_kernels/blackwell/<feature>/` could be better (see `tensorrt_llm/_torch/cute_dsl_kernels/blackwell/` for LLM-side directory structures), so:
   - `cute_dsl_kernels/blackwell/attention/` — packaged cubins + runner for dense / QK16PV8 **(PR 13721)**
   - `cute_dsl_kernels/blackwell/video_sparse_attention/` — VSA JIT source + interface (this PR)

3. **Add an `Optional[SparseAttentionConfig]` field on `AttentionConfig` next to `quant_attention_config`.**
   Mirror the pattern **PR 13721** introduced: a sub-config that is `None` by default; setting it switches the backend into the sparse path. Something like:

   ```python
   class SparseAttentionConfig(StrictBaseModel):
       """Sparse-attention recipe (CUTEDSL backend / VSA only)."""
       vsa_sparsity: float = Field(0.875, ge=0.0, le=1.0, ...)
       skip_softmax_threshold: float = Field(0.0, ge=0.0)

   class AttentionConfig(StrictBaseModel):
       backend: Literal["VANILLA", "TRTLLM", "FA4", "CUTEDSL"] = ...
       quant_attention_config: Optional[QuantAttentionConfig] = None     # from #13721
       sparse_attention_config: Optional[SparseAttentionConfig] = None   # new in this PR
   ```

4. **Move the VSA caller-side wrapper into `cute_dsl.py`; let `CuTeDSLAttention` dispatch between the two CuTe DSL kernels.**
   Sketch (pseudocode, names are negotiable):

   ```python
   # attention_backend/cute_dsl.py
   class CuTeDSLAttention(AttentionBackend):
       def __init__(self, ..., quant_attention_config=None, sparse_attention_config=None, **kw):
           # Mutually exclusive: at most one of quant_/sparse_attention_config is set
           # (config-level validator will enforces this).
           self.quant_attention_config = quant_attention_config
           self.sparse_attention_config = sparse_attention_config
           ...

       def forward(self, q, k, v, *, gate_compress=None, gate_fine=None, **kw):
           if self.sparse_attention_config is not None:
               # VSA path: tile / coarse-pool / topk / block-sparse JIT kernel
               return self._forward_vsa(q, k, v, gate_compress=gate_compress, gate_fine=gate_fine, **kw)
           # Standard dense / QK16PV8 path: packaged cubins
           return self._forward_dense(q, k, v, **kw)
   ```

   - `_forward_dense` is what `cute_dsl.py` already does in **PR 13721** (calls into `cute_dsl_kernels/blackwell/attention/`).
   - `_forward_vsa` holds today's `VSAAttentionBackend.forward` body and calls into `cute_dsl_kernels/blackwell/video_sparse_attention/`.
   - The tiling / metadata builder (`VSAMetadata`, `VSAMetadataBuilder`, `set_vsa_forward_context`) can stay in `cute_dsl.py` alongside the dispatcher, or live in `cute_dsl_kernels/blackwell/video_sparse_attention/interface.py` and be imported lazily — runtime call.

   This way, callers see exactly one `"CUTEDSL"` backend; the sparse vs dense decision is data-driven (presence of `sparse_attention_config`) rather than a third backend name.

Let me know what you think — or if any of these conflict with constraints I'm missing (e.g., kernel availability, config, etc.).

tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/blackwell/video_sparse_attention/ptx.py
```diff
@@ -0,0 +1,377 @@
...
+            """{\n\t
+            atom.relaxed.shared::cta.cta.max.s32 _, [$0], $1;\n\t
+            \n\t}""",
+            "l, r",
+            has_side_effects=True,
+            is_align_stack=False,
+            asm_dialect=llvm.AsmDialect.AD_ATT,
+            loc=loc,
+            ip=ip,
+        )
```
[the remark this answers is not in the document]
o-stoner: This code was not authored by me (see source [here](https://gitlab-master.nvidia.com/cjerry/vsa/-/blob/main/csrc/attn/video_sparse_attn/vsa/blackwell/ptx.py?ref_type=heads#L130)); will leave as-is.

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl.py
```diff
@@ -42,12 +48,223 @@
...
+
 
 class CuTeDSLAttention(AttentionBackend):
     """
     CuTe DSL (NVIDIA kernels) backend for diffusion models.
 
-    Uses pre-compiled cubin kernels (head_dim=128 only).
+    Dense path uses pre-compiled cubins and requires head_dim=128. The VSA
+    sparse path (sparse_attention_config set) uses a JIT-compiled CuTe kernel
+    when head_dim=128 / fp16-bf16 / sm100+, and otherwise falls back to dense SDPA.
```
chang-l: I think it makes sense to keep VSA as its own attention class, e.g. `VSAAttention(AttentionBackend)`, instead of plumbing it through the dense FMHA path. I’d also put it in a separate `vsa.py` file.

Even though VSA uses the CuTeDSL backend, that’s more of a frontend/API choice for kernel generation. Internally, we can decouple it from the dense FMHA interface, since the two don’t seem to share much in common.

In general, does it make sense to have something like: 
```
 attention_backend/cute_dsl/
    __init__.py     # re-exports CuTeDSLAttention, VSAAttention
    fmha.py        # CuTeDSLAttention   (the existing cute_dsl.py (dense), moved)
    vsa.py          # VSAAttention       (new, separate file — but in the family)
    _common.py      # shared CuTe-DSL plumbing (composition, utils)
```
and then dispatch like?
```
def create_attention(backend, *, attention_config=None,
                       sparse_attention_config=None,
                       attention_metadata_state=None, **dims):
      attn_cls = get_visual_gen_attention_backend(backend)  
      # VSA = the cutedsl family running the vsa sparse algo:
      if backend.upper() == "CUTEDSL" and sparse_attention_config is not None \
              and sparse_attention_config.algorithm == "vsa":
          from .cute_dsl.vsa import VSAAttention
          attn_cls = VSAAttention                                     
```

chang-l: also add @xrq-phys @zhenhuaw-me for discussion

o-stoner: @chang-l thank you, yes, this makes sense, I have refactored accordingly - there is actually not enough shared between fmha + vsa to have a `_common.py` file, so that does not exist under the `cute_dsl` directory as of now. @xrq-phys @zhenhuaw-me would love to hear your thoughts as well.

tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/blackwell/video_sparse_attention/block_sparse_attn_dsl_fwd.py
```diff
@@ -0,0 +1,2350 @@
...
+
+        qk_mma_pipeline.consumer_wait(qk_mma_consumer_state)
+        correction_pipeline.producer_acquire(p_producer_state)
+        cute.copy(
+            tiled_copy_t2r,
+            tCtS_ld[None, None, None, 0, which, qk_mma_consumer_state.index],
+            tCrS_ld,
+        )
+        # calculate P
+        _max = self.mask_then_cal_local_max(tCrS_ld, n_size)
```
chang-l: does it make sense to adopt the changes here: https://gitlab-master.nvidia.com/liuc/vsa/-/tree/vsa-blk64-fwd-opt?ref_type=heads – it should help improve the perf of this kernel further

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/vsa.py
```diff
@@ -0,0 +1,407 @@
...
+        scale = D**-0.5
+        scores_c = torch.einsum("bnhd,bmhd->bhnm", q_c, k_c) * scale
+        attn_probs_c = scores_c.softmax(dim=-1)
+        o_c = torch.einsum("bhnm,bmhd->bnhd", attn_probs_c, v_c)
+
+        use_cute = (
+            _vsa_import_error is None
+            and is_cute_supported(q)
+            and (q.dtype == k.dtype == v.dtype)
+            and num_cubes <= VSA_KERNEL_MAX_CUBES
```
chang-l, line 356: **[correctness] Missing even-`num_cubes` guard.** `use_cute` gates only on `num_cubes <= VSA_KERNEL_MAX_CUBES`, but the paired-block kernel also requires `num_cubes % 2 == 0` (the test asserts this and the kernel comment notes the pairing). `num_cubes = ceil(T/4)*ceil(H/4)*ceil(W/4)` can be **odd** for latent shapes other than the validated Wan-720p (e.g. Wan2.2-5B / other resolutions) → it would dispatch to a kernel that assumes even pairing and produce a wrong result or crash. Suggest adding `and num_cubes % 2 == 0` to this condition (fall back to SDPA when odd), unless the kernel pads internally — can you confirm?

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/vsa.py
```diff
@@ -0,0 +1,407 @@
...
+        o_c_full = VSAPreprocessor.untile(o_c_tiled, reverse_tile_partition_indices, non_pad_index)
+        o_f = F.scaled_dot_product_attention(
+            q.transpose(1, 2), k.transpose(1, 2), v.transpose(1, 2)
+        ).transpose(1, 2)
+        if gate_fine is not None:
+            return gate_compress * o_c_full + gate_fine * o_f
+        return gate_compress * o_c_full + o_f
+
+    @classmethod
+    def support_lse(cls) -> bool:
```
chang-l, line 403: **[correctness] Confirm Ring is gated off.** `support_lse` returns False, so Ring attention (which needs per-split LSE to combine partials) can't produce correct output for VSA. Attention2D is explicitly rejected and Ulysses is head-parallel (no LSE needed, and is tested), but I don't see an explicit guard rejecting VSA + Ring in `wrap_parallel_attention`. Can you confirm Ring is rejected? Otherwise it would silently produce wrong results.

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/vsa.py
```diff
@@ -0,0 +1,407 @@
...
+
+    Requires an active VSA forward context (set_vsa_forward_context) during
+    each forward call. Does not support LSE output.
+    """
+
+    def __init__(
+        self,
+        layer_idx: int = 0,
+        num_heads: int = 8,
+        head_dim: int = 64,
```
chang-l: Minor: default `head_dim=64` is inconsistent with the kernel's hard 128 (the path relies on `is_cute_supported` to fall back). Consider defaulting to 128 or making it required. Relatedly, the coarse mean-pool einsum assumes `num_kv_heads == num_heads` — fine for Wan MHA, but worth an explicit assert for any GQA/MQA model.

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+CuTe DSL attention backend family for visual generation models.
+
+  fmha.py  — CuTeDSLAttention  (dense cubin path, head_dim=128)
+  vsa.py   — VSAAttention       (Video Sparse Attention, CuTe JIT + SDPA fallback)
+  _common.py — shared CuTe DSL plumbing
```
chang-l: Nit: this docstring references `_common.py — shared CuTe DSL plumbing`, but that file isn't in this PR. Drop the line or add the file.

tests/integration/test_lists/test-db/l0_b200.yml
```diff
@@ -188,6 +188,7 @@ l0_b200:
   - unittest/_torch/visual_gen/test_cache_dit.py
   - unittest/_torch/visual_gen/test_quant_ops.py
   - unittest/_torch/visual_gen/test_attention_cute_dsl.py
+  - unittest/_torch/visual_gen/test_attention_cute_dsl_vsa.py
```
chang-l, line 222: can we also add your `tests/unittest/_torch/visual_gen/multi_gpu/test_wan_vsa_ulysses.py` to the CI?

chang-l (commented):
Thanks @o-stoner – can we also add some E2E test or pipeline-level test ? Meanwhile, can you also update the docs to include this new feature: `docs/source/models/visual-generation.md` (enable + constraints), and example doc (how to enable)?

tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/blackwell/video_sparse_attention/block_sparse_attn_dsl_fwd.py
o-stoner: I added the changes, and the videos visually look good - however, I am seeing about the same e2e perf for a Wan 2.1 VSA video generation as the original kernel. Do you have any thoughts?

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/vsa.py
o-stoner: The kernel handles odd num_cubes correctly; the last Q-block has no pair and is processed without one. `test_cute_kernel_odd_num_cubes_correctness` in `tests/unittest/_torch/visual_gen/test_attention_cute_dsl_vsa.py` covers this: it runs the kernel with `num_cubes` in {1, 3, 9} and asserts the output matches dense SDPA within tolerance. I've removed the even `num_cubes` assertion from the previous test; that is not a kernel constraint.

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/vsa.py
o-stoner: Ring is now explicitly rejected at construction time with a ValueError in Attention.__init__ , same place Attention2D is rejected. It raises before any forward pass, so there's no silent wrong-result path.

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/vsa.py
o-stoner: Fixed; `head_dim` now defaults to 128, and there's an explicit assert `self.num_kv_heads == self.num_heads` with an explanatory message at construction time in vsa.py:L269.

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/__init__.py
o-stoner: Dropped the line; thanks.

tests/integration/test_lists/test-db/l0_b200.yml
o-stoner: I believe `test_wan_vsa_ulysses.py` should be picked up by `l0_dgx_b200.yml` via the directory entry `unittest/_torch/visual_gen/multi_gpu`? Or no?

laikhtewari (approved):

tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/blackwell/video_sparse_attention/block_sparse_attn_dsl_fwd.py
chang-l: I see, thanks. I’ve only done microbenchmarks, but maybe in E2E the blocks are not full most of the time?

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/fmha.py
```diff
@@ -221,8 +204,7 @@ def support_lse(cls) -> bool:
 
     @property
     def preferred_layout(self) -> AttentionTensorLayout:
-        """Return the preferred tensor layout for this backend."""
-        return self._preferred_layout
+        return AttentionTensorLayout.NHD
```
chang-l: I don’t think we need to make a change here. Can we avoid making unnecessary changes to our dense FMHA file?

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
```diff
@@ -186,10 +186,23 @@ def _forward_unfused(
         v: torch.Tensor,
         **kwargs,
     ) -> torch.Tensor:
+        # gate_compress / gate_fine (VSA) must follow the same all-to-all as
+        # Q/K/V so they arrive at the inner backend in the same (full-S, sharded-H) layout.
+        gate_compress = kwargs.pop("gate_compress", None)
+        gate_fine = kwargs.pop("gate_fine", None)
+
```
chang-l, line 209: Is it possible to propagate `gate_compress` change to the `_forward_fused` path?

tensorrt_llm/_torch/visual_gen/attention_backend/utils.py
```diff
@@ -128,6 +128,13 @@ def create_attention(
                 "DiffusionModelConfig; creation path must not allocate metadata implicitly."
             )
         kwargs["attention_metadata_state"] = attention_metadata_state
+    if backend.upper() == "CUTEDSL" and attention_config is not None:
+        if attention_config.sparse_attention_config is not None:
```
chang-l: can we also add check `attention_config.sparse_attention_config.algo=="vsa"` here?

tensorrt_llm/_torch/visual_gen/attention_backend/cute_dsl/fmha.py
o-stoner: Yes, thanks; reverted almost all changes back to main version except for a few comments/import paths.

tensorrt_llm/_torch/visual_gen/attention_backend/parallel.py
o-stoner: Done, thanks; `_forward_fused` now pops `gate_compress`/`gate_fine`, applies `all_to_all_4d` on each, then re-injects them into `kwargs` before the inner backend call.

tensorrt_llm/_torch/visual_gen/attention_backend/utils.py
o-stoner: Done, thanks

chang-l (commented):
CI coverage for test_wan_vsa_ulysses.py (8-GPU, cfg=2 × ulysses=4) — please confirm it actually runs before merge.

The test is collected via the unittest/_torch/visual_gen/multi_gpu directory entry in l0_dgx_b200.yml, which lives under the system_gpu_count: 8 / stage: post_merge / backend: pytorch condition. Two problems:

--add-multi-gpu-test only adds pre-merge multi-GPU stages, so it will not trigger this. Post-merge tests need /bot run --stage-list "<stage>" (or the heavy /bot run --post-merge).

Could you run python scripts/test_to_stage_mapping.py --tests "test_wan_vsa_ulysses" on this branch and confirm which stage runs it, then trigger that stage (e.g. /bot run --stage-list "<stage>") and verify it passes before merge?

o-stoner:
> CI coverage for test_wan_vsa_ulysses.py (8-GPU, cfg=2 × ulysses=4) — please confirm it actually runs before merge.
> 
> The test is collected via the unittest/_torch/visual_gen/multi_gpu directory entry in l0_dgx_b200.yml, which lives under the system_gpu_count: 8 / stage: post_merge / backend: pytorch condition. Two problems:
> 
> --add-multi-gpu-test only adds pre-merge multi-GPU stages, so it will not trigger this. Post-merge tests need /bot run --stage-list "" (or the heavy /bot run --post-merge).
> 
> Could you run python scripts/test_to_stage_mapping.py --tests "test_wan_vsa_ulysses" on this branch and confirm which stage runs it, then trigger that stage (e.g. /bot run --stage-list "") and verify it passes before merge?

@chang-l `unittest/_torch/visual_gen/multi_gpu/test_wan_vsa_ulysses.py` is run in the CI report [here](http://tensorrt-llm.tensorrt-llm-ci-report.sc2-paas.nvidia.com/?job=%2FLLM%2Fmain%2FL0_MergeRequest_PR&build=44049) under `L0_Test-x86_64-Multi-GPU` by `DGX_B200-8_GPUs-PyTorch-1`, so IIUC I think it is being collected and triggered already, but please correct if I am misunderstanding. It failed due to a process kill in the previous run, but I will confirm on the next run whether or not it passes.

tensorrt_llm/_torch/visual_gen/executor.py
```diff
@@ -498,6 +498,19 @@ def run_diffusion_worker(
...
+        if torch.cuda.is_available() and visual_gen_args is not None:
+            _attn = visual_gen_args.attention_config
+            _sa = getattr(_attn, "sparse_attention_config", None)
+            _is_vsa = (
+                getattr(_attn, "backend", "") == "CUTEDSL"
+                and getattr(_sa, "algorithm", "") == "vsa"
+            )
+            _has_ulysses = getattr(visual_gen_args.parallel_config, "ulysses_size", 1) > 1
+            if _is_vsa and _has_ulysses:
+                os.environ.setdefault("NCCL_NVLS_ENABLE", "0")
```
chang-l, line 513: Why do we need to set this env var? Is it for VSA (within the scope of this PR), or is it a workaround for a flaky test?

tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/blackwell/video_sparse_attention/interface.py
```diff
@@ -0,0 +1,177 @@
...
+    "block_sparse_attn_from_indices_cute",
+]
+
+
+# JIT compile is multi-second; reuse aggressively.
+_COMPILE_CACHE: dict = {}
+
+
+def is_cute_supported(q: torch.Tensor) -> bool:
+    """Capability check for the CuTe path. Set TLLM_VSA_DISABLE_CUTE=1 to force off."""
```
chang-l: Are we actively using the `TLLM_VSA_DISABLE_CUTE` environment variable? If not, can we remove it?

tests/unittest/_torch/visual_gen/multi_gpu/test_wan_pipeline_parallel.py
```diff
@@ -76,20 +76,20 @@ def _cleanup_mpi_env():
...
     root = Path("/home/scratch.trt_llm_data_ci/llm-models/")
     if "LLM_MODELS_ROOT" in os.environ:
         root = Path(os.environ["LLM_MODELS_ROOT"])
     if not root.exists():
         root = Path("/scratch.trt_llm_data/llm-models/")
-    assert root.exists(), (
-        "Set LLM_MODELS_ROOT or ensure /home/scratch.trt_llm_data_ci/llm-models/ is accessible."
-    )
-    return str(root)
+    return str(root) if root.exists() else None
```
chang-l: Why are we removing this assertion? Without it, it would bypass the test silently if the model root dir cannot be found, right?

tensorrt_llm/_torch/visual_gen/modules/attention.py
```diff
@@ -96,13 +96,34 @@ def __init__(
         # Select compute backend (orthogonal to parallelism)
         vgm = config.visual_gen_mapping
         ulysses_size = vgm.ulysses_size if vgm else 1
+        ring_size = vgm.ring_size if vgm else 1
+        attn2d_size = (vgm.attn2d_row_size * vgm.attn2d_col_size) if vgm else 1
```
chang-l: we can simplify here to:
`cp_size = vgm.cp_size if vgm else 1` (context-parallel size (cp_size) covers both Attention2D and ring parallelism.)

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -357,6 +370,46 @@ def __init__(
...
+        self.to_gate_compress = None
+        self.to_gate_fine = None
+        _attn_cfg = getattr(model_config, "attention", None)
+        _sa_cfg = getattr(_attn_cfg, "sparse_attention_config", None) if _attn_cfg else None
+        _is_vsa = (
+            _attn_cfg is not None
+            and getattr(_attn_cfg, "backend", "VANILLA") == "CUTEDSL"
+            and _sa_cfg is not None
+            and getattr(_sa_cfg, "algorithm", None) == "vsa"
+        )
```
chang-l, line 384: Non-blocking: Refactor suggestion: 

Could we centralize both VSA detection and pipeline compatibility validation?

1. Since sparse_attention_config is a discriminated union and _validate_sparse_attention_config() already enforces vsa → CUTEDSL, AttentionConfig could expose a single predicate:
```
   @property
   def uses_vsa(self) -> bool:
       return isinstance(
           self.sparse_attention_config,
           VideoSparseAttentionConfig,
       )
 ```
Call sites such as WanBlock, Attention, WanPipeline, PipelineLoader, and Executor could then use attention_config.uses_vsa instead of repeating backend/null/algorithm checks.

2. Could pipeline support use a positive capability declaration instead of separate VSA blacklists in FLUX, FLUX.2, LTX-2, and Wan I2V (and possibly other new models, like cosmos3 and qwenimge)? For example, BasePipeline.supported_sparse_algorithms, with Wan T2V opting into {"vsa"}. This would remove duplicated checks and consistently reject VSA for unsupported current and future pipelines.

If you prefer and agree, this refactor can also be addressed in a follow-up PR

chang-l (approved):
Approve with some refactor suggestions

tensorrt_llm/_torch/visual_gen/executor.py
o-stoner: This is a WAR for a NCCL hang on Blackwell when VSA (CuTeDSL) + Ulysses are combined. It's gated via _is_vsa and _has_ulysses so should be confined to just the scope of this PR. This was the least invasive solution I found to resolve. I haven't root-caused it further than confirming this disabled NVLS resolves the hang, but I would be happy to dig into an NCCL_DEBUG trace if you would prefer firmer confirmation before merging.

tensorrt_llm/_torch/visual_gen/cute_dsl_kernels/blackwell/video_sparse_attention/interface.py
o-stoner: Good catch, removed

tests/unittest/_torch/visual_gen/multi_gpu/test_wan_pipeline_parallel.py
o-stoner: Yes, you are correct - reverted this file back to main version

tensorrt_llm/_torch/visual_gen/modules/attention.py
o-stoner: Good idea, done

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
o-stoner: Created a Jira task to address this refactoring PR: https://jirasw.nvidia.com/browse/TRTLLM-13944

tensorrt_llm/_torch/visual_gen/executor.py
chang-l: Thanks. Could we have an NVBug to track with repro of issue, then add the link to the comment section here?

o-stoner: Yes, filed a bug: https://nvbugspro.nvidia.com/bug/6425321

PR #14297, opened 2026-05-19
[None][fix] DSv4 indexer: stable radix aux scratch for CUDA Graph safety

longcheng-nv:
## Description

The fp32 Radix path of `indexer_topk_decode` was not CUDA-Graph-safe: it allocated its split-work scratch buffers (`aux_indices`, `aux_logits`) via `th::empty` per call when `blocks_per_row > 1`. Under CUDA Graph capture + replay these per-call pointers become stale when the caching allocator is perturbed by chunked-prefill activations at high concurrency. The captured Radix part1 / part2 kernels then write to recycled memory; the resulting `CUDA_ERROR_ILLEGAL_ADDRESS` surfaces at the next sync (commonly inside DeepGEMM's `smxx_layout.hpp:97` TMA descriptor pack), looking like a downstream FP8 GEMM crash but caused upstream by the indexer.

`IndexerTopKOp.cpp:91` already documents the pitfall for the Heuristic path:

> "Caller-owned scratch buffer for heuristic TopK output values. **Must be pre-allocated with stable address for CUDA Graph compatibility.**"

and `dsa.py:879` follows that contract for `heuristic_scratch_values`. The Radix path was the missing half of the same pattern — this PR adds it.

### Observed failure (Flash MXFP4, ISL ≈ 100K, TP = EP = 4, BS = 32, MTP = 0)

| cell | CONC | Mode | GVR | result |
|---|---|---|---|---|
| `cell_013_C008_TEP_GVRF` | 8  | TEP | OFF (Radix) | `CUDA 700` at `smxx_layout.hpp:97` |
| `cell_017_C016_TEP_GVRF` | 16 | TEP | OFF (Radix) | `CUDA 700` at `smxx_layout.hpp:97` |
| `cell_021_C032_TEP_GVRF` | 32 | TEP | OFF (Radix) | `CUDA 700` at `smxx_layout.hpp:97` |

All three crashes converge on the identical stack: `fp8_swap_ab_gemm` → `_fp8_quantize_1x128_ue8m0` → `deep_gemm.get_mn_major_tma_aligned_packed_ue8m0_tensor` → `smxx_layout.hpp:97`. The DeepGEMM line is the next sync after the offending kernel, not the offender itself. Same-point GVR-ON cells (Heuristic path, uses persistent `heuristic_scratch`) run clean — confirming the asymmetry is in the indexer.

### Change

- **`cpp/tensorrt_llm/thop/IndexerTopKOp.cpp`** — extend `indexer_topk_decode` with two optional `Tensor?` kwargs (`radix_aux_indices`, `radix_aux_logits`). When both are provided and `blocks_per_row > 1`, use the caller-owned buffers; otherwise fall back to per-call `th::empty` (back-compat for bench scripts / warmup helpers / callers not under CUDA Graph capture). Validates `is_cuda`, same device, contiguous, dtype (`int32` / `float32`), `numel ≥ num_rows × blocks_per_row × index_topk`.
- **`tensorrt_llm/_torch/attention_backend/sparse/dsa.py`** — unconditionally allocate persistent `radix_aux_indices` / `radix_aux_logits` in `DSAtrtllmAttentionMetadata` sized to the worst case (`max_gen_tokens × kMaxBlocksPerRowDecode = 10 × num_sparse_topk`). Buffers are routed through `self.get_empty(self.cuda_graph_buffers, ...)` so they have stable addresses across graph replays and survive metadata-level CUDA-graph capture. Passed at the call site.
- **`tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py`** — extend the `register_fake` Python signature to match the new schema.

### Side-effect surface

- Same Radix kernel invocation, same dispatch logic, **same numerical output**.
- DSv3.2 sparse path shares this op and metadata class — benefits transparently, no regression.
- API back-compat: new kwargs default to `None`; existing callers continue to use the `th::empty` fallback unchanged.
- Memory cost: ≈ 1.3 MB per `DSAtrtllmAttentionMetadata` instance on Flash defaults (`max_num_seqs = 32 × kMaxBlocksPerRowDecode = 10 × index_topk = 512 × 2 buffers × 4 B/elt`). Negligible.

## Test Coverage

Unit tests added to `tests/unittest/_torch/thop/parallel/test_indexer_topk.py`:

- **`test_indexer_topk_decode_radix_aux_equivalence`** — Parametrised over `(batch_size, num_tokens, compress_ratio)` shapes that force `blocks_per_row > 1` (including the exact `batch_size=8, num_tokens=16384, compress_ratio=4` that mirrors the Flash CONC=8 failure point). Verifies caller-owned-aux output is bit-identical to the legacy `th::empty` fallback.
- **`test_indexer_topk_decode_radix_aux_cuda_graph_replay`** — Captures the op into a CUDA Graph with caller-owned aux, replays 8× and asserts each replay's output matches a non-graph reference. This is the direct regression test for the original stale-pointer bug.
- **`test_indexer_topk_decode_radix_aux_validation`** — Negative tests for the new validation paths: wrong `dtype` (indices not int32 / logits not float32), undersized buffer (< `num_rows × blocks_per_row × index_topk`), non-contiguous tensor.

Existing coverage retained:
- `test_indexer_topk_decode` / `test_indexer_topk_decode_sm_saturation` / `test_indexer_topk_decode_launch_policy_transitions` continue to exercise the `th::empty` fallback path (do not pass the new kwargs).
- `bench_indexer_topk_kernel.py` continues to work unchanged (does not pass new kwargs).

End-to-end validation:
- Re-run of `pareto_v4_pro_swebench100k_b300` sweep against this binary on 8× B300 is in progress; CONC = 1 paired cells (P1 / P2) already completed cleanly with the new fix in place. High-CONC GVR-OFF cells (the ones that previously triggered G4) will be the live test.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/thop/IndexerTopKOp.cpp
+++ b/cpp/tensorrt_llm/thop/IndexerTopKOp.cpp
@@ -123,35 +123,28 @@ void indexer_topk_decode(th::Tensor const& logits, th::Tensor const& seq_lens, t
         th::Tensor aux_logits = th::empty({0}, th::TensorOptions().dtype(th::kFloat32).device(logits.device()));
         if (blocks_per_row > 1)
         {
-            // Prefer caller-owned scratch with stable address (CUDA Graph safe;
-            // matches the heuristic_scratch convention noted above). Falls
-            // back to per-call th::empty when the caller did not supply one
-            // (back-compat for bench scripts / older callers — exposes the
-            // CUDA-Graph stale-pointer hazard at high CONC, see SKILL G4).
+            // Require caller-owned scratch with stable address (CUDA Graph safe;
+            // matches the heuristic_scratch convention noted above). All
+            // in-tree callers go through dsa.py's DSAtrtllmAttentionMetadata
+            // which always pre-allocates these buffers.
             int64_t const needed_elts = static_cast<int64_t>(num_rows) * blocks_per_row * index_topk;
-            if (radix_aux_indices.has_value() && radix_aux_logits.has_value())
-            {
-                auto const& ai = radix_aux_indices.value();
-                auto const& al = radix_aux_logits.value();
-                TORCH_CHECK(ai.is_cuda() && al.is_cuda(), "radix_aux_{indices,logits} must be CUDA tensors");
-                TORCH_CHECK(ai.device() == logits.device() && al.device() == logits.device(),
-                    "radix_aux_{indices,logits} must be on the same device as logits");
-                TORCH_CHECK(ai.is_contiguous() && al.is_contiguous(), "radix_aux_{indices,logits} must be contiguous");
-                TORCH_CHECK(ai.scalar_type() == th::kInt32, "radix_aux_indices must be int32");
-                TORCH_CHECK(al.scalar_type() == th::kFloat32, "radix_aux_logits must be float32");
-                TORCH_CHECK(ai.numel() >= needed_elts && al.numel() >= needed_elts,
-                    "radix_aux_{indices,logits} must hold at least num_rows*blocks_per_row*index_topk elements (got ",
-                    ai.numel(), " / ", al.numel(), ", need ", needed_elts, ")");
-                aux_indices = ai;
-                aux_logits = al;
-            }
-            else
-            {
-                aux_indices = th::empty({num_rows, blocks_per_row, index_topk},
-                    th::TensorOptions().dtype(th::kInt32).device(logits.device()));
-                aux_logits = th::empty({num_rows, blocks_per_row, index_topk},
-                    th::TensorOptions().dtype(th::kFloat32).device(logits.device()));
-            }
+            TORCH_CHECK(radix_aux_indices.has_value() && radix_aux_logits.has_value(),
+                "radix_aux_{indices,logits} must be pre-allocated by the caller when blocks_per_row > 1 "
+                "(got blocks_per_row=",
+                blocks_per_row, "). Required for CUDA Graph safety.");
+            auto const& ai = radix_aux_indices.value();
+            auto const& al = radix_aux_logits.value();
+            TORCH_CHECK(ai.is_cuda() && al.is_cuda(), "radix_aux_{indices,logits} must be CUDA tensors");
+            TORCH_CHECK(ai.device() == logits.device() && al.device() == logits.device(),
+                "radix_aux_{indices,logits} must be on the same device as logits");
+            TORCH_CHECK(ai.is_contiguous() && al.is_contiguous(), "radix_aux_{indices,logits} must be contiguous");
+            TORCH_CHECK(ai.scalar_type() == th::kInt32, "radix_aux_indices must be int32");
+            TORCH_CHECK(al.scalar_type() == th::kFloat32, "radix_aux_logits must be float32");
+            TORCH_CHECK(ai.numel() >= needed_elts && al.numel() >= needed_elts,
+                "radix_aux_{indices,logits} must hold at least num_rows*blocks_per_row*index_topk elements (got ",
+                ai.numel(), " / ", al.numel(), ", need ", needed_elts, ")");
+            aux_indices = ai;
+            aux_logits = al;
         }
         tk::invokeIndexerTopKDecode(logits.data_ptr<float>(), seq_lens.data_ptr<int32_t>(), indices.data_ptr<int32_t>(),
             aux_logits.data_ptr<float>(), aux_indices.data_ptr<int32_t>(), splitWorkThreshold, num_rows, num_columns,
--- a/cpp/tensorrt_llm/thop/IndexerTopKOp.cpp
+++ b/cpp/tensorrt_llm/thop/IndexerTopKOp.cpp
@@ -38,7 +38,8 @@ namespace torch_ext
 
 void indexer_topk_decode(th::Tensor const& logits, th::Tensor const& seq_lens, th::Tensor const& indices,
     int64_t next_n, int64_t index_topk, std::optional<th::Tensor> const& pre_idx,
-    std::optional<th::Tensor> const& heuristic_scratch, int64_t compress_ratio)
+    std::optional<th::Tensor> const& heuristic_scratch, int64_t compress_ratio,
+    std::optional<th::Tensor> const& radix_aux_indices, std::optional<th::Tensor> const& radix_aux_logits)
 {
     TORCH_CHECK(compress_ratio > 0, "compress_ratio must be greater than 0");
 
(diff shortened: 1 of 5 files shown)


longcheng-nv:
## Local validation results

Two pieces of independent evidence collected on this host (8× B300, internal node) before requesting review:

### 1. Unit tests — all 7 pass

`pytest tests/unittest/_torch/thop/parallel/test_indexer_topk.py -k radix_aux -p no:xdist -v`

```
test_indexer_topk_decode_radix_aux_equivalence[8-1-512-16384-4]       PASSED
test_indexer_topk_decode_radix_aux_equivalence[16-1-512-16384-4]      PASSED
test_indexer_topk_decode_radix_aux_equivalence[32-1-512-16384-4]      PASSED
test_indexer_topk_decode_radix_aux_equivalence[8-1-512-8192-1]        PASSED
test_indexer_topk_decode_radix_aux_cuda_graph_replay[8-1-512-16384-4]  PASSED
test_indexer_topk_decode_radix_aux_cuda_graph_replay[32-1-512-16384-4] PASSED
test_indexer_topk_decode_radix_aux_validation                          PASSED
======================= 7 passed in 2.69s =======================
```

Test design note (baked into the helpers + docstrings of `_build_decode_inputs` and `_topk_logit_values_sorted`): the test fixture pins `seq_lens = torch.full((batch_size,), num_tokens)` to deterministically hit `blocks_per_row > 1` (otherwise `generate_seq_lens`'s 90/10 random distribution can land below the `num_columns >= 4096` threshold and silently bypass the new code path), and compares top-K logit VALUES (sorted) rather than indices to tolerate the histogram-radix sort's atomic-driven tie-breaking nondeterminism.

### 2. V4 Flash MXFP4 end-to-end reproducer — original crash config now clean

Reproduced the cell that originally crashed in `pareto_b200_swebench100k_v1_2waves/logs/cell_013_C008_TEP_GVRF.log` (model: `/home/scratch.jinshik_gpu/DeepSeek-V4-Flash`, 4 of 8 B300 via `CUDA_VISIBLE_DEVICES=0,1,2,3`):

| Knob | Value |
|---|---|
| TP × EP | 4 × 4 |
| `max_batch_size` | 32 |
| `concurrency` | **8** |
| `max_num_tokens` | 16384 |
| `kv_cache_free_gpu_mem_fraction` | 0.9 |
| ISL | 100200 (SWE-bench-100K real prompt) |
| MoE backend | TRTLLM |
| **`sparse_attention.enable_heuristic_topk`** | **`false`** ← drives the Radix path |
| `enable_attention_dp` | true (TEP mode) |
| CUDA graph batch_sizes | `[1, 2, 4, 8, 16, 32]`, padding ON |
| `TRTLLM_SCHEMEX_DEBUG` | 1 (indexer dispatch trace enabled) |

```
EXIT=0
no smxx_layout.hpp:97 / no CUDA_ERROR_ILLEGAL_ADDRESS / no Fatal error in log

PERFORMANCE OVERVIEW
Request Throughput (req/sec):                     0.1312
Total Output Throughput (tokens/sec):             537.5508
Total Token Throughput (tokens/sec):              13687.5969
Total Latency (ms):                               121915.9222
Average request latency (ms):                     60934.6482
Per User Output Throughput [w/ ctx] (tps/user):   67.4326
Per GPU Output Throughput (tps/gpu):              134.3877
Average time-to-first-token [TTFT] (ms):          8373.2775
Average time-per-output-token [TPOT] (ms):        12.8355
[TPOT] P50: 13.40   P90: 14.12   P95: 14.27   P99: 14.27
```

5586 `[Scheme X] ... -> Radix path` dispatch lines in the trace — confirming the Radix split-work path was the dominant path exercised throughout the run. Pre-fix this exact configuration produced `CUDA driver error (smxx_layout.hpp:97): 700 (CUDA_ERROR_ILLEGAL_ADDRESS)` on rank 0 within ~80 s of the bench loop. With this PR applied: 121.9 s of clean steady-state decoding across 16 prompts, P99 TPOT 14.3 ms.

## What's tested vs. what's not

- ✅ FP32 Radix-split-work path (this PR's primary target)
- ✅ Caller-owned buffer ≡ fallback `th::empty` (selected top-K logit values bit-identical)
- ✅ CUDA Graph capture + 8 replays with stable selected top-K
- ✅ Validation: dtype / size / contiguous TORCH_CHECKs reject malformed buffers
- ✅ End-to-end production workload (Flash MXFP4) on the exact pre-fix-crash configuration

Happy to add other reproducer configs or runs of the existing parametrisations if reviewers want broader coverage.

cpp/tensorrt_llm/thop/IndexerTopKOp.cpp
```diff
@@ -122,10 +123,35 @@ void indexer_topk_decode(th::Tensor const& logits, th::Tensor const& seq_lens, t
...
-                th::TensorOptions().dtype(th::kInt32).device(logits.device()));
-            aux_logits = th::empty({num_rows, blocks_per_row, index_topk},
-                th::TensorOptions().dtype(th::kFloat32).device(logits.device()));
+            // Prefer caller-owned scratch with stable address (CUDA Graph safe;
+            // matches the heuristic_scratch convention noted above). Falls
+            // back to per-call th::empty when the caller did not supply one
+            // (back-compat for bench scripts / older callers — exposes the
+            // CUDA-Graph stale-pointer hazard at high CONC, see SKILL G4).
+            int64_t const needed_elts = static_cast<int64_t>(num_rows) * blocks_per_row * index_topk;
+            if (radix_aux_indices.has_value() && radix_aux_logits.has_value())
```
heyuhhh: This should always be true because we pre-allocated the space and pass them in the function. For simplify i think we can add some assert sentence here to make sure and remove the remaining code, especially the if-else

heyuhhh (approved):
LGTM. Leave little comments for refine the code. Thanks!

longcheng-nv:
Pushed commit 383d8c51a8 (top of HEAD now) addressing two items:

**1. @heyuhhh's inline review on `IndexerTopKOp.cpp:L132`** — done. Replaced the `if (radix_aux_indices.has_value() && ...)` runtime branch with a `TORCH_CHECK(...)` precondition and dropped the dead `else` fallback (per-call `th::empty`). The success path is byte-identical; future callers that forget to supply the buffers now get a clear error instead of silent CUDA-Graph stale-pointer corruption.

**2. Unrelated CI unblock in `model_config.py`** — the previous run (PR_Github #49291, L0_MergeRequest_PR/38953) failed all 10 V4 test cases with:

```
TypeError: tensorrt_llm.llmapi.llm_args.DeepSeekV4SparseAttentionConfig()
got multiple values for keyword argument 'indexer_k_dtype'
```

Root cause was a duplicate-kwarg on the V4 construction site at `model_config.py:838` — `indexer_k_dtype` was being passed both as an explicit kwarg and via `**indexer_config` (which itself sets `indexer_config['indexer_k_dtype']` in `update_sparse_attention_indexer_config`). None of the 10 failures involved the radix-aux code itself; they all abort in `ModelConfig.from_pretrained` before reaching any indexer path. Folded the 2-line fix into this PR rather than opening a separate one for speed. Full reasoning in the commit body.

@lfr-0531 @mingyangHao Could you take a look when convenient? Local validation summary already in the PR body's "Local validation results" section (7/7 radix-aux unit tests pass + 3 high-CONC TEP+Radix CUDA-illegal-address crashes on B200 Pareto sweep are fully eliminated).

cpp/tensorrt_llm/thop/IndexerTopKOp.cpp
heyuhhh: Do we really need this two tensors? As we have already passed two aux tensors. I think we can just used them to avoid incomplete graph capture

longcheng-nv:
Pushed commit `ffbc5addce` addressing @heyuhhh's second inline review comment on `IndexerTopKOp.cpp:122`.

**The reviewer's point** — the two `th::empty({0}, ...)` placeholder Tensors at lines 122-123 became vestigial after the prior commit 383d8c5 tightened the `blocks_per_row > 1` contract. They were only reachable in the `blocks_per_row == 1` case, where the kernel never dereferences them. Reviewer suggested using the caller-passed buffers directly to avoid the leftover per-call allocation.

**The fix** — replace both Tensor placeholders with raw nullable `int32_t*` / `float*` pointers, matching the existing `heuristicScratchPtr` convention at lines 98-111. Hoist all caller-buffer validation (dtype / device / contiguity) out of the `bp > 1` branch; keep the `numel` check inside it since that bound only matters when the kernel actually uses the buffers. For `bp == 1` without caller-owned scratch, pass nullptr through; for `bp > 1` without scratch, fail early with the same TORCH_CHECK message as before.

**Same-PR-introduced regressions also fixed** — chased down two paths that the strict TORCH_CHECK in 383d8c5 would now trip but weren't caught locally (the previous run was filtered with `-k radix_aux`, and CI broke earlier at `model_config.py` so the indexer tests never executed):

1. **`warmup_heuristic_topk_decode` (dsa.py:68)** — default `num_cols=4096` falls below `kSeqSmall=12288` so canUseHeuristic is false and the call routes to Radix with `blocks_per_row=2` (`num_rows=1` sweeps `bp ∈ [2, maxByCols=2]`). Allocate worst-case `kMaxBlocksPerRowDecode=10` caller-owned scratch in the warmup. Cost ~80 KB, one-shot.
2. **`_run_indexer_topk_decode_check` (test_indexer_topk.py:229)** — covers params like `num_tokens=32768, compress_ratio=1` → numCols=32768 → bp=9–10. Same fix: pre-allocate via existing `_build_radix_aux_buffers` and pass via kwargs. Same change applied to `test_indexer_topk_decode_dist` (fp32 + `num_tokens=8192` hits Radix bp>1) and `_run_indexer_topk_decode_v4_gvr_check` (defense-in-depth — GVR almost always succeeds for that param matrix, but the kwarg is now uniform).

The radix-aux equivalence test no longer made sense in its original form (compared caller-owned vs the removed `th::empty` fallback). Rewrote it to verify allocation-pointer-independence — two invocations with freshly-allocated aux at distinct backing addresses must yield the same selected top-K logit values.

`pre-commit run` is clean on all three changed files (clang-format, ruff-format, isort, codespell, yapf).

cc @heyuhhh @lfr-0531 @mingyangHao

longcheng-nv:
@lfr-0531 Friendly ping — this PR is ready and would appreciate a merge when you have a moment.

**Current status (head `ffbc5ad`):**

- ✅ **CI green** — latest `PR_Github #49446` on `ffbc5ad` → **SUCCESS** (`L0_MergeRequest_PR` pipeline #39091 SUCCESS)
- ✅ **Approved by @heyuhhh** on 2026-05-20: *"LGTM. Leave little comments for refine the code. Thanks!"*
- ✅ **Both inline review comments resolved:**
  - `IndexerTopKOp.cpp:L132` — replaced the `if (radix_aux_indices.has_value() && ...)` runtime branch with a `TORCH_CHECK(...)` precondition and dropped the dead `else` fallback (commit `383d8c5`).
  - `IndexerTopKOp.cpp:L122` — removed the two vestigial `th::empty({0}, ...)` placeholder tensors; the caller-passed aux tensors are reused directly (commit `ffbc5ad`).
- ✅ **Mergeable: `clean`**, base = `feat/deepseek_v4`.

**Scope recap:** makes the fp32 Radix path of `indexer_topk_decode` CUDA-Graph-safe by moving `aux_indices` / `aux_logits` scratch to caller-owned stable allocations — mirroring the same contract `heuristic_scratch_values` already follows on the Heuristic path. Fixes the `CUDA 700` at `smxx_layout.hpp:97` seen at ISL ≈ 100K / BS ≥ 8 with GVR-OFF (Radix) on Flash MXFP4.

Thanks!

lfr-0531 (approved):

PR #13842, opened 2026-05-07
[None][doc] Add Claude skill for multimodal model onboarding

yechank-nvidia:
Adds .claude/skills/trtllm-model-onboard-multimodal/SKILL.md, an agent skill guiding contributors through onboarding HuggingFace multimodal models to the TensorRT-LLM PyTorch backend (tensorrt_llm/_torch/models/). Covers the aggregated and EPD-disaggregated request lifecycles, module reuse rules, three performance contracts (sync-free forward, async server-side preprocessing, shared-tensor broadcast), an onboarding phase walk-through keyed off TestModelingMultimodal, and a pre-PR checklist.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive onboarding documentation for integrating HuggingFace multimodal models into TensorRT-LLM. Includes system architecture, performance requirements, implementation phases, and pre-release verification guidelines.

PR #14300, opened 2026-05-19
[None][feat] Gemma4 MM: native vision + audio towers

Hudayday:
## Summary                                           
                                                                                                                                                                                                                    
  Per multimedia team's [PR](https://github.com/NVIDIA/TensorRT-LLM/pull/12932) feedback on the initial Gemma4 MM port, this PR replaces the temporary `AutoModel.from_config` HF fallback with **native TensorRT-LLM implementations** of the Gemma4 vision and audio
  (Conformer) towers, so that the multimodal path goes through TRT-LLM's own attention/MLP modules rather than re-entering HuggingFace at runtime.                                                                  
                                                    
  - **Native Gemma4 vision tower** (`modeling_gemma4_vision.py`) — replaces `AutoModel.from_config`; vision attention workspace now bound to the vision scale rather than inheriting the LM-side bound (avoids the  
  5589 GiB resize previously observed on 31B MMMU).         
  - **Native Gemma4 audio Conformer tower** (`modeling_gemma4_audio.py`) — replaces `AutoModel.from_config`; chunked local attention + rel-pos bias ported directly from the HF reference.
  - **HF Gemma4RMSNorm convention fix** (audio + vision) — Gemma4's MM-tower RMSNorm is the plain `w * x / rms`, **not** the Gemma3 LLM-side `(1 + w) * x / rms`. Wiring the correct norm convention is what brings 
  BLEU back into parity (E4B CoVoST2 1.90 → ~41.9 in earlier validation runs).
  - **`modeling_gemma4mm.py`** updated to import the native towers instead of the HF fallback.                                                                                                                      
  - **Test suite** `tests/unittest/_torch/multimodal/test_gemma4_multimodal.py` registered in `l0_b200.yml`, covering Tier 0/1/2 vision-tower parity plus the new RMSNorm regression tests.
  - Debug probes used during the port (`DEBUG_AUDIO_TOWER` env-gated stats) have been scrubbed before this commit; only Signed-off-by trailers in the history.                                                      
                                                                                                                                                                                                                    
  ## Alignment with HF tower baseline                       
                                                                                                                                                                                                                    
  The fix-mode (native towers) results have been **aligned with the HF-fallback baseline**. All completed fix-mode cells so far are within ±2pp of the HF baseline (well inside MMMU/CoVoST seed noise):                                                                                                                                         
                                                                                                                                                                                                                    
  | variant | task   | HF baseline | native fix | Δ      |                                                                                                                                                          
  |---------|--------|-------------|------------|--------|                                                                                                                                                          
  | E2B     | mmmu   | 42.67       | 42.11      | −0.56  |                                                                                                                                                          
  | E2B     | covost | 37.70 BLEU  | 37.78 BLEU | +0.08  |                                                                                                                                                          
  | E4B     | covost | 41.91 BLEU  | 41.91 BLEU | exact  |                                                                                                                                                          
  | 26B     | mmmu   | 58.33       | 58.56      | +0.23  |  

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Added native audio and vision tower implementations for Gemma4 multimodal model, replacing previous placeholder approach with optimized TensorRT-LLM components.

* **Tests**
  * Expanded test coverage for multimodal audio/vision components, including forward pass validation, weight loading parity, and end-to-end generation testing.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14300?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<!-- review_stack_entry_end -->

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_gemma4_audio.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma4_audio.py
@@ -15,61 +15,782 @@
 """Native TRT-LLM Gemma4 audio tower (Conformer).
 
 Replaces ``AutoModel.from_config(config.audio_config)`` in
-``modeling_gemma4mm.py``. **Greenfield**: this is the first native Conformer
-implementation in the fork (Phi4MM also uses HF reuse via ``importlib``
-at ``modeling_phi4mm.py:106``). Conformer = mel-subsample conv + macaron
-FFN + relative-position attention + depthwise conv branch -- not standard
-ViT/transformer layout.
+``modeling_gemma4mm.py``. Greenfield: first native Conformer port in the
+fork. The audio attention is **chunked local attention with relative-position
+bias + softcap tanh + per-dim query scale** — fundamentally non-standard
+shape, so it stays pure PyTorch instead of inheriting TRT-LLM ``Attention``
+(unlike vision, which routes through the standard backend dispatch).
+``Linear`` comes from TRT-LLM for layout parity with the vision tower.
+RMSNorm reuses HF's ``Gemma4RMSNorm`` directly (plain ``w * x / rms`` — not
+the Gemma3 ``(1 + w) * x / rms`` LLM variant) instead of the TRT-LLM
+``RMSNorm`` module — the TRT-LLM module dispatches to ``flashinfer_rmsnorm``
+whose CUTE kernel rejects non-contiguous row strides, and
+``Gemma4AudioLightConv1d`` feeds it exactly such tensors via
+``depthwise_conv1d(x.transpose(1,2)).transpose(1,2)``. Audio is prefill-only;
+the unfused path's overhead is negligible. The subsample conv stack uses
+plain ``nn.LayerNorm`` (HF uses real LN there, not RMSNorm).
 
-Architecture references (HF transformers 5.5.3, verified 2026-05-14 spike):
-- ``Gemma4AudioRelPositionalEncoding`` @ ``transformers/models/gemma4/modeling_gemma4.py:178``
-- ``Gemma4AudioAttention`` @ ``:209``
-- ``Gemma4AudioLayer`` @ ``:485``
-
-Migration plan: ``GEMMA4_MM_TOWER_MIGRATION.md`` §3 Phase 2.
-Gated behind vision Tier 3 parity per acceptance gate matrix (§1.6.10).
+Architecture references (HF transformers 5.5.3 ``modeling_gemma4.py``):
+- ``Gemma4ClippableLinear``                     @ ``:128``
+- ``Gemma4RMSNorm``                             @ ``:157``
+- ``Gemma4AudioRelPositionalEncoding``          @ ``:178``
+- ``Gemma4AudioAttention``                      @ ``:209``
+- ``Gemma4AudioSubSampleConvProjectionLayer``   @ ``:317``
+- ``Gemma4AudioSubSampleConvProjection``        @ ``:345``
+- ``Gemma4AudioFeedForward``                    @ ``:375``
+- ``Gemma4AudioCausalConv1d``                   @ ``:411``
+- ``Gemma4AudioLightConv1d``                    @ ``:444``
+- ``Gemma4AudioLayer``                          @ ``:485``
+- ``Gemma4AudioModel``                          @ ``:1800``
 """
 
-from typing import Dict, Optional
+import math
+import os
+from dataclasses import dataclass
+from functools import cached_property
+from typing import Dict, Optional, Tuple
 
 import torch
 import torch.nn as nn
+import torch.nn.functional as F
+from transformers.activations import ACT2FN
+from transformers.models.gemma4.modeling_gemma4 import Gemma4RMSNorm as _HFGemma4RMSNorm
 
 from ..model_config import ModelConfig
+from ..modules.linear import Linear
+from .modeling_utils import _load_weights_impl
+
+# DEBUG_PROBE_BEGIN: gated by env var, fires only when DEBUG_AUDIO_TOWER=1.
+_DEBUG_AUDIO = os.environ.get("DEBUG_AUDIO_TOWER", "0") == "1"
+
+
+def _audio_stats(name: str, t: Optional[torch.Tensor]) -> None:
+    if not _DEBUG_AUDIO or t is None:
+        return
+    f = t.detach().float()
+    finite_frac = torch.isfinite(f).float().mean().item()
+    print(
+        f"  [audio] {name:32s} shape={tuple(t.shape)} dtype={t.dtype} "
+        f"mean={f.mean().item():+.4g} std={f.std().item():.4g} "
+        f"absmax={f.abs().max().item():.4g} finite={finite_frac:.4f}",
+        flush=True,
+    )
+
+
+# DEBUG_PROBE_END
+
+
+@dataclass
+class _AudioOutput:
+    """Stand-in for ``transformers.Gemma4AudioModelOutput`` (fields used by caller)."""
+
+    last_hidden_state: torch.Tensor
+    attention_mask: Optional[torch.Tensor] = None
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


tensorrt_llm/_torch/models/modeling_gemma4_audio.py
```diff
@@ -0,0 +1,713 @@
...
+    attention_mask: Optional[torch.Tensor] = None
+
+
+# ---------------------------------------------------------------------------
+# RMSNorm: thin kwarg/dtype adapter over HF's ``Gemma4RMSNorm`` (see module
+# docstring for why this isn't the TRT-LLM ``RMSNorm`` module).
+# ---------------------------------------------------------------------------
+
+
+class _Gemma4AudioRMSNorm(_HFGemma4RMSNorm):
```
yechank-nvidia: Can you remove '_' from the class name please?

tensorrt_llm/_torch/models/modeling_gemma4_audio.py
```diff
@@ -0,0 +1,713 @@
...
+from transformers.activations import ACT2FN
+from transformers.models.gemma4.modeling_gemma4 import Gemma4RMSNorm as _HFGemma4RMSNorm
+
+from ..model_config import ModelConfig
+from ..modules.linear import Linear
+from .modeling_utils import _load_weights_impl
+
+
+@dataclass
+class _AudioOutput:
```
yechank-nvidia: Also, for this one.

tensorrt_llm/_torch/models/modeling_gemma4_vision.py
```diff
@@ -0,0 +1,819 @@
...
+        position_embeddings_btn = self.encoder.rotary_emb(inputs_embeds, pixel_position_ids)
+
+        # Flatten to (sum_valid, H) and drop padded tokens — encoder runs
+        # FULL attention with per-image cu_seqlens.
+        flat_embeds = inputs_embeds[valid]
+        cos_btn, sin_btn = position_embeddings_btn
+        flat_cos = cos_btn[valid]
+        flat_sin = sin_btn[valid]
+        flat_position_ids = pixel_position_ids[valid]
+        seq_lens_cpu = valid.sum(dim=-1).to(torch.int).cpu()
```
yechank-nvidia: Hi, this one can incure CPU-GPU sync. and CPU-GPU sync can hurt the performance. Is there any way we can do it other way?

tensorrt_llm/_torch/models/modeling_gemma4_vision.py
```diff
@@ -0,0 +1,819 @@
...
+
+        # SigLip-style context-only metadata: kv_cache_manager=None, no decode
+        # phase. Re-prepared each forward with the actual per-image seq lens.
+        # Vision tower runs one image per call (mm forward loops per-image),
+        # so bounds are vision-scale, not LLM-scale. Using LLM max_num_tokens
+        # here makes the C++ attn workspace resize to TB-range and OOM
+        # (sized as max_num_requests * heads * max_num_tokens^2 * bytes).
+        self.metadata_cls = get_attention_backend(model_config.attn_backend).Metadata
+        self.attn_metadata = self.metadata_cls(
+            max_num_requests=1,
```
yechank-nvidia: Can I ask why you set 1?

tensorrt_llm/_torch/models/modeling_gemma4_vision.py
```diff
@@ -0,0 +1,819 @@
...
+        )
+        self.pooler = Gemma4VisionPooler(self.config)
+        self.standardize = bool(getattr(self.config, "standardize", False))
+        if self.standardize:
+            self.register_buffer("std_bias", torch.zeros(self.config.hidden_size, dtype=dtype))
+            self.register_buffer("std_scale", torch.ones(self.config.hidden_size, dtype=dtype))
+
+        # SigLip-style context-only metadata: kv_cache_manager=None, no decode
+        # phase. Re-prepared each forward with the actual per-image seq lens.
+        # Vision tower runs one image per call (mm forward loops per-image),
```
yechank-nvidia: Can I ask why is that? We should do cross-requests batching on images.

tensorrt_llm/_torch/models/modeling_gemma4_audio.py
Hudayday: Done. Renamed to Gemma4AudioRMSNorm.

tensorrt_llm/_torch/models/modeling_gemma4_audio.py
Hudayday: Done. Renamed to AudioOutput

tensorrt_llm/_torch/models/modeling_gemma4_vision.py
Hudayday: Sync is API-mandated because it is using the TRTLLM-based Attention, which AttentionMetadata.prompt_lens is `Optional[List[int]] (host tensor)`, same in `modeling_siglip.py`, `modeling_qwen2vl.py`, `modeling_nemotron_nano.extract_feature_dynamic`. Even if we delayed it, `attn_metadata.prepare()` forces the materialisation.

tensorrt_llm/_torch/models/modeling_gemma4_vision.py
Hudayday: Now scales with `model_config.max_num_requests` (like other mm models). The `=1` was a workaround for the unfused MHA path's O(B × T²) workspace; with the FMHA-on fix in the latest commit, the bump is safe.

tensorrt_llm/_torch/models/modeling_gemma4_vision.py
Hudayday: It was set that way for the same reason as the previous comment. The pattern now matches Gemma3vl / Qwen2VL / Qwen3VL / LlavaNext, supporting cross-request batching.

tensorrt_llm/_torch/models/modeling_gemma4_vision.py
yechank-nvidia: I mean, yes in other modeling files, those are already on the cpu, so pinned_tensor to GPU would not occur CPU-GPU sync.

But it seems Gemma4 derives those values from GPU tensor to CPU which occurs CPU-GPU sync. and hurt performance very badly.

Is there anyway to derive those numbers from host tensors?

Hudayday: Thanks for the catch!

`Gemma4InputProcessor` now emits `image_seq_lens` as a CPU `List[int]` at preprocessing time (host-only, same shape as `Qwen2VL`'s `grid_thw.tolist()` path). The previous `valid.sum(-1).cpu()` GPU→CPU sync is gone; the tower now builds `seq_lens_cpu purely` from the host list and GPU is untouched.

E2E on B200 / 30-sample MMMU shows a small but measurable improvement

yechank-nvidia: Thx for the update.

FYI, I just merged the claude skills for multimodal-models([https://github.com/NVIDIA/TensorRT-LLM/pull/13842](https://github.com/NVIDIA/TensorRT-LLM/pull/13842)). If you are okay, can you try?

yechank-nvidia (approved):
Thanks all for the hard work!

tensorrt_llm/_torch/models/modeling_gemma4_vision.py
Hudayday: Hi, I just used the agent skill to check this PR and fixed a few formatting and import mismatches. The agent check and CI should both be green now. Thanks for the suggestion!

PR #14305, opened 2026-05-19
[None][chore] Route MHC fused-HC workspace through global Buffers pool

lancelly:
## Summary

Split out the **MHC fused-HC workspace pool routing** from PR #14053 / #14241 as an independent, **opt-in** memory optimization.

Activated via the env switch `TRTLLM_DSV4_MEM_OPTS=1`; default off → behavior is byte-identical to current source.

## Change

`_FusedHcWorkspaceCache.get` (`tensorrt_llm/_torch/modules/mhc/mhc_cuda.py`) gains an opt-in path that routes the four output buffers (`residual_cur`, `post_mix_cur`, `comb_mix_cur`, `layer_input_cur`) and the three internal scratch buffers (`y_acc_ws`, `r_acc_ws`, `done_counter_ws`) through `get_memory_buffers()` so MHC scratch can share backing memory with other layer-transient buffers (e.g., the attention workspace).

The env switch is read **once** at class definition (cached on `_FusedHcWorkspaceCache._USE_POOL`) so the lookup doesn't run on every workspace request.

### Aliasing invariant

Pool routing only applies for `B <= _CACHE_MAX_B` (decode-sized shapes). For `B > _CACHE_MAX_B` (prefill), the code falls through to fresh `torch.empty` per call, matching the upstream uncached path; this preserves prefill's no-aliasing semantics where layer N's `*_prev` inputs and layer N+1's `*_cur` outputs live in distinct tensors.

At decode, upstream's existing per-runner LRU already returns the same tensor across layers, so the kernel is already in-place safe; pool routing inherits that aliasing without widening it.

## Files touched

```
tensorrt_llm/_torch/modules/mhc/mhc_cuda.py | 55 +++++++++++++++++++++++++++++
1 file changed, 55 insertions(+)
```

## Required env

```bash
export TRTLLM_DSV4_MEM_OPTS=1
```

Optional (recommended for parallel-job stability, orthogonal to this PR):
```bash
export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
```


mingyangHao:
Concern: with TRTLLM_DSV4_MEM_OPTS=1, this may break CUDA graph address stability.

  The old cache is keyed by (B, ws_ks, m_batches, device). The new pool path uses fixed
  buffer names and reserve_buffer=False. If a later larger decode bucket/splitK shape
  requests the same name, Buffers.get_buffer() can delete the smaller block and allocate a
  new one, invalidating pointers captured by an earlier CUDA graph.

  Suggested fix:
  1. Include shape/device in the buffer names, e.g. mhc_residual_cur_{device}_{B}_{n}
  _{hidden_size}, mhc_y_acc_ws_{device}_{ws_ks}_{B}_{shape_n}, mhc_done_counter_ws_{device}
  _{m_batches}.
  2. Or keep the existing runner-local LRU for CUDA graph decode buckets, and only use the
  global pool for non-graph eager decode.
  3. If using the global pool for graph decode, reserve graph-captured buffers so they
  cannot be deleted/replaced by later larger shapes.

  Also add a regression test with TRTLLM_DSV4_MEM_OPTS=1:
  capture/replay one decode bucket, allocate/use a larger decode bucket or larger splitK
  workspace, then replay the earlier graph and compare with eager.

  There is also a device-keying issue: old cache includes device; the new pool path does
  not, and memory_buffer_utils allocates on current cuda device. The fix should preserve
  per-device separation.

longlee0622:

PR #14307, opened 2026-05-19
[None][fix] cold-start warmup for KV-aware ADP router

lancelly:
## Summary

Cold-start mitigation for the KV-aware ADP router: round-robin the first relaxed requests across all ranks so every rank caches the (assumed shared) system prompt before cache-affinity scoring would otherwise pin all traffic to the first warm rank.

## Problem

With `match_rate_threshold=0.1` the cache-affinity gate stays ON for any system prompt longer than ~10% of a request. Once a single rank caches that prompt, the `(req_tokens - match_len)` term dominates the load term:

- Low-traffic sequential arrivals: each new request pins to the warm rank; other ranks never see the prompt.
- Eventual traffic spike: cold ranks pay the full prefill cost re-computing the same shared prompt.

## Fix

`KVCacheAwareADPRouter` tracks `_pending_warmup_ranks: Set[int]`:

- Initialised to `{0, ..., tp_size-1}` only when `cold_start_warmup=True`; otherwise an empty set.
- In `route_requests`, any `target_dp_rank` (explicit strict or synthesised warmup) discards its rank from the set. Relaxed requests with no explicit target synthesise `min(_pending_warmup_ranks)` while the set is non-empty.
- Strict pre-assignments count toward warmup naturally; cap-saturated ranks also discard so the synthesiser never loops on a busy rank.
- Once the set is empty, warmup is dormant forever and routing is pure scoring.

## Opt-in (default False)

The behaviour is gated by `AttentionDpConfig.kv_cache_routing_cold_start_warmup` (default `False`). The warmup is a win for workloads that share a long system prompt, but for diverse-prompt workloads it can scatter requests onto cold ranks and bypass cache-affinity scoring entirely for the first `tp_size` requests — wasting prefill work the scoring path would otherwise consolidate. Default `False` preserves pre-warmup routing; users who hit cold-start pinning opt in explicitly.

Test layout:

- `TestKVCacheAwareADPRouterWarmup` constructs its router with `cold_start_warmup=True` to exercise the new code path, plus a new `test_disabled_by_default` asserting the empty pending set when the flag is absent.
- Pre-existing scoring tests no longer need the `_pending_warmup_ranks = set()` opt-out (default is already off).

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
+++ b/tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
@@ -23,7 +23,7 @@ import random
 from abc import ABC, abstractmethod
 from collections import namedtuple
 from dataclasses import MISSING, astuple, dataclass, field, fields, replace
-from typing import TYPE_CHECKING, Dict, List, Tuple
+from typing import TYPE_CHECKING, Dict, List, Set, Tuple
 
 from tensorrt_llm.logger import logger
 
@@ -450,6 +450,9 @@ class KVCacheAwareADPRouter(ADPRouter):
         self.match_rate_threshold = match_rate_threshold
         self.fair_share_multiplier = fair_share_multiplier
         self._all_ranks_prefix_matches: List[Dict[int, int]] = []
+        # Cold-start warmup: ranks not yet targeted through this router.
+        # See ``route_requests``.
+        self._pending_warmup_ranks: Set[int] = set(range(self.dist.tp_size))
         # Requests still sending KV to GEN are invisible in active_requests;
         # fold them back in via the transfer manager (see create_rank_state).
         self.async_transfer_manager = async_transfer_manager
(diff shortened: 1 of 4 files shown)


lishicheng1996-nv (approved):

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
```diff
@@ -450,6 +452,11 @@ def __init__(
         self.match_rate_threshold = match_rate_threshold
         self.fair_share_multiplier = fair_share_multiplier
         self._all_ranks_prefix_matches: List[Dict[int, int]] = []
+        # Cold-start warmup: ranks not yet targeted through this router.
+        # Empty set disables warmup; see ``route_requests``.
+        self._pending_warmup_ranks: Set[int] = (
+            set(range(self.dist.tp_size)) if cold_start_warmup else set()
```
QiJune: `self.dist.tp_size` -> `self.dist.mapping.dp_size`, which may be easier to understand.  It takes me a few time to catch up that if enable_attention_dp=True, dp_size = tp_size :).

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
```diff
@@ -580,25 +587,37 @@ def get_relax_value(req_item):
...
+                    self._req_tokens(req_item) - self._match_len(target_dp_rank, req_item.id),
+                    0,
+                )
+                all_ranks_num_active_tokens[target_dp_rank] += effective
+                scheduled = True
+                all_ranks_new_requests[target_dp_rank].append(req_item)
+            # Any targeted rank counts as warmed; cap-saturation also
+            # discards to avoid the synthesiser looping on a busy rank.
+            if target_dp_rank is not None:
+                self._pending_warmup_ranks.discard(target_dp_rank)
```
QiJune, line 619: A saturated warmup target should not be removed from `_pending_warmup_ranks` unless the request is actually scheduled there. Otherwise a rank can be marked as warmed without receiving any request, so its shared prefix is never seeded. Please move `self._pending_warmup_ranks.discard(target_dp_rank)` under the `scheduled` path; if needed, saturated ranks can be skipped only for the current routing call.

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
lancelly: Sure, fixed.

tensorrt_llm/_torch/pyexecutor/scheduler/adp_router.py
lancelly: Thanks! Fixed.

QiJune (approved):
LGTM

PR #14308, opened 2026-05-19
[None][chore] memory observability probes for DSv4 mem-opts validation

lancelly:
Add some memory profile log.
Moe related workspace and activation tensor should be traced in a seperate PR, as well as attention active tensor.


pengbowang-nv (approved):
Attention change LGTM

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -637,6 +637,15 @@ def create_rope_const_params(self, interleave: bool = True):
                 if rope_inv_freq is not None else None,
                 weakref.ref(rope_cos_sin),
             )
+        # One-shot log on cache miss (typically 2-4 times per model load).
+        from ..pyexecutor.trace_log_utils import log_tensor_size
```
pengbowang-nv: NIT: Can we move this to the start of file?

lancelly: Sure, fixed.

PR #14312, opened 2026-05-19
[None][fix] slice incremental decode from last emitted text

ltalal:
## Summary
We observed missing spaces in streaming chunks after detokenizations with Kimi-K2.5 when used stream_interval=8. Example (missing space between 'logs' and 'and'):
`{"id":"chatcmpl-c5ba8e658c5e48db94ae3308819586ba","choices":[{"delta":{"content":"Investigating snapshot build and verification: locating logs","reasoning_content":""},"index":0,"avg_decoded_tokens_per_iter":2.9772727489471436}],"created":1778888341,"model":"Kimi-K2.5-NVFP4","object":"chat.completion.chunk"}
{"id":"chatcmpl-c5ba8e658c5e48db94ae3308819586ba","choices":[{"delta":{"content":"and exploring the ml-infra-server snapshot pipeline.","reasoning_content":""},"index":0,"avg_decoded_tokens_per_iter":3.0833332538604736}],"created":1778888341,"model":"Kimi-K2.5-NVFP4","object":"chat.completion.chunk"}`

- Fix `trtllm_decode_incrementally` to slice new text from the last emitted decoded string (`last_decoded_text`) instead of re-decoding prior tokens in isolation.
- Prevents BPE/tiktoken streaming bugs where a token decodes differently alone vs in context (e.g. `"thecat"` instead of `"the cat"`).
- Add a unit test with a prefix-unstable mock tokenizer that reproduces the failure mode.

## Test plan
- [x] `pytest tests/unittest/llmapi/test_llm.py::test_trtllm_decode_incrementally_slices_from_previous_decoded_text -v`
- [x] `pytest tests/unittest/llmapi/test_llm.py::test_tokenizer_decode_incrementally -k "TinyLlama and TRTLLM" -v`
- [x] pre-commit on modified files

### DIFF
--- a/tensorrt_llm/tokenizer/tokenizer.py
+++ b/tensorrt_llm/tokenizer/tokenizer.py
@@ -412,14 +412,14 @@ class TransformersTokenizer(TokenizerBase):
             states = {}
         last_new_tokens = states.pop('last_new_tokens', [])
         pending_tokens = states.pop('pending_tokens', [])
-
-        if len(last_new_tokens) > 0:
-            last_new_text = self.convert_tokens_to_string(
-                last_new_tokens,
-                skip_special_tokens=skip_special_tokens,
-                spaces_between_special_tokens=spaces_between_special_tokens)
-        else:
-            last_new_text = ""
+        # Some BPE/tiktoken decoders can render a token differently by itself than
+        # in a longer token span. E.g. the tokenizer decodes token "B" as " the",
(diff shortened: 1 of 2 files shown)


yibinl-nvidia:
Hi @ltalal , thanks for your findings and the fix PR, I will take a look. Could you share the TRTLLM command to repro missing spaces in streaming chunks with Kimi-K2.5?

ltalal:
This fix breaks tool calling

PR #12801, opened 2026-04-07
[None][feat] Add llm.encode() fast path for encoder-only models

tingyangk:
<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

## New Features
* Added encoder-only execution mode via new `encode_only` parameter for models like BERT
* Introduced `encode()` API method for efficient encode-only inference
* Added `EncoderOutput` dataclass containing logits and tokenized prompts
* Generation APIs now raise errors when encode-only mode is active

## Tests
* Added test coverage for encode-only inference, batch processing, and API validation
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

<!--
Please write the PR title by following this template:

**[JIRA ticket/NVBugs ID/GitHub issue/None][type] Summary**

Valid ticket formats:
  - JIRA ticket: [TRTLLM-1234] or [FOOBAR-123] for other FOOBAR project
  - NVBugs ID: [https://nvbugs/1234567]
  - GitHub issue: [#1234]
  - No ticket: [None]

Valid types (lowercase): [fix], [feat], [doc], [infra], [chore], etc.

Examples:
  - [TRTLLM-1234][feat] Add new feature
  - [https://nvbugs/1234567][fix] Fix some bugs
  - [#1234][doc] Update documentation
  - [None][chore] Minor clean-up

Alternative (faster) way using CodeRabbit AI:

**[JIRA ticket/NVBugs ID/GitHub issue/None] @coderabbitai title**

NOTE: "@coderabbitai title" will be replaced by the title generated by CodeRabbit AI, that includes the "[type]" and title.
For more info, see /.coderabbit.yaml.

-->

 ## Summary                                                                                                                                                             
   
  Adds a dedicated `llm.encode()` API for encode-only paths that bypasses the decoder-oriented PyExecutor loop entirely. Works for encoder models and decoder models running a "single-prefill" path.
                                                         
 ### Problem

  The current LLM API routes encoder models through the same `PyExecutor` designed for autoregressive decoders, introducing significant CPU overhead per batch from scheduler, KV cache management, sampling, and request state machine — none of which apply to encoders. **Encoder models need a simple, direct path to the model’s forward call with batch inference executed in a single pass.**
                                                                                                                                                                         
  ### Solution
[...]

PR #14326, opened 2026-05-19
[None][feat] Add encoder CUDA graph support to llm.encode()

tingyangk:
## Release Notes

* **New Features**
  * Added CUDA graph acceleration support for encoder workloads, enabling faster inference through graph caching and replay.
  * Added `return_raw_logits` parameter to encoder API for direct logits output.
  * Extended CUDA graph configuration with encoder-specific sizing options.

* **Improvements**
  * Enhanced encoder execution stability and performance optimization.
  * Added stricter validation for encoder and multimodal encoder configurations.

* **Tests**
  * Expanded encoder testing to include CUDA graph equivalence checks against eager execution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14326?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Description

The PyTorch backend encoder forward path [`llm.encode()`](https://github.com/NVIDIA/TensorRT-LLM/pull/12801) currently runs in pure eager mode. Prefill-only encoder workloads (BERT classification, decoder-as-encoder) are dominated by launch overhead at small batch sizes or short sequence lengths.

This PR introduces encoder CUDA graph infrastructure to the encode-only path. The main advantage is that, unlike piecewise CUDA graphs, **it captures the entire model, including the attention layers, into one CUDA graph per bucket**.

It adds a dedicated `EncoderCUDAGraphRunner` alongside the decoder runner in `cuda_graph_runner.py`, wires graph capture/replay into `PyTorchModelEngine.warmup_encoder()` and the encode forward step, and surfaces the configuration through the existing `CudaGraphConfig` on `TorchLlmArgs`.

### Key design points:

- **Composite key.** Encoder graphs use a 3-tuple `(padded_batch_size, padded_num_tokens, padded_max_seq_len)` because two batches with the same `num_tokens` can still differ in `max(seq_lens)` for context-only batches, and FMHA launch parameters depend on the latter. A new `max_context_q_len_override` field — plumbed through to the `thop.attention` op — locks the FMHA kernel launch parameters across replays even when the actual batch has shorter sequences.

- **Stable attention metadata.** `TrtllmAttentionMetadata` gains `bind_encoder_cuda_graph_seq_lens()` and `prepare_encoder_cuda_graph_replay()`. When CC (confidential computing) is off, pinned host buffers are used: per-replay `seq_lens` updates and model inputs are CPU memcpys into pinned host buffers, and the host-to-device copies are captured in the graph to minimize CPU work in the replay path. When CC is on, pinned host buffers are not preferred, so the replay path updates the persistent GPU buffers directly before graph replay.

- **Attention backend constraints.** Only `TRTLLM` attention metadata is eligible. Non-TRTLLM backends, including FlashInfer, fall back to eager because their planner/wrapper state is per batch and is not replay-safe.

- **Input constraints.** Encoder CUDA graphs currently support text-token inputs only. Additional encoder inputs such as `inputs_embeds` or other model-specific tensors remain supported by the eager `llm.encode()` path, but are not captured/replayed by `EncoderCUDAGraphRunner` yet.

## New API surface

`CudaGraphConfig` (`tensorrt_llm/llmapi/llm_args.py`) gains four encoder-only fields. They mirror the existing `batch_sizes / max_batch_size` pair:

| Field            | Type                | Default | Meaning                                                                |
| ---------------- | ------------------- | ------- | ---------------------------------------------------------------------- |
| `num_tokens`     | `Optional[List[int]]` | `None`  | Total token counts (Σ per-request lengths) to capture graphs for.       |
| `max_num_token`  | `NonNegativeInt`    | `0`     | If `num_tokens` is provided, must equal `max(num_tokens)`; otherwise the list is auto-generated from this. |
| `seq_lens`       | `Optional[List[int]]` | `None`  | Per-request max sequence lengths to capture graphs for.                 |
| `max_seq_len`    | `NonNegativeInt`    | `0`     | Same auto-derivation rule as `max_num_token`, for `seq_lens`.            |

Leaving all four at their defaults keeps encoder CUDA graphs disabled, even when `batch_sizes` is set — graphs only activate when the user explicitly opts in to encoder bucketing.

## Usage

Minimal example — BERT classification with encoder CUDA graphs enabled:

```python
from tensorrt_llm import LLM
from tensorrt_llm.llmapi import EncodeCudaGraphConfig

cgc = EncodeCudaGraphConfig(
    batch_sizes=[1, 4],          # batch buckets, same as the decoder path
    num_tokens=[16, 64],         # encode-only: total tokens per batch
    seq_lens=[8, 32],            # encode-only: per-request max seq lens
    enable_padding=True,         # round inputs up to the nearest bucket
)

llm = LLM(
    model="bert-base-uncased-yelp-polarity",
    encode_only=True,
    cuda_graph_config=cgc,
)

outputs = llm.encode([
    "Hello, my name is",
    "The capital of France is",
])
for out in outputs:
    print(out.logits)

llm.shutdown()
```

The grid here captures graphs for the feasible graph buckets of the cross-product `{1, 4} × {16, 64} × {8, 32}`; at inference time `enable_padding=True` rounds each incoming batch up to the nearest supported bucket. Decoder-only models can also use the encode-only path with `encode_only=True` and the same config.

### BERT 110M benchmark

- GPU: L4
- Batch size: 1
- ISL: ~10
- Column definition:
  - `baseline_generate`: runs with old `llm.generate()` path
  - `baseline_encode`: runs with new `llm.encode()` path in eager mode
  - `cuda-graph`: runs with new `llm.encode()` path + encoder CUDA graph enabled
  - `torch-compile+cuda-graph`: runs with new `llm.encode()` path + TorchCompileConfig & encoder CUDA graph enabled
- Row definition:
  - speedup `generate`: uses original `llm.generate()` latency as baseline; shows the speedup from moving to `llm.encode()`.
  - speedup `encode`: uses eager `llm.encode()` latency as baseline; shows the speedup from encoder CUDA graphs.

| Metric             | `baseline_generate` | `baseline_encode` | `cuda-graph` | `torch-compile+cuda-graph` |
| ------------------ | ------------------- | ----------------- | ------------ | ----------- |
| `p50`              | 6.744 ms            | 4.337 ms          | 1.292 ms     | 1.202 ms    |
| `p99`              | 7.323 ms            | 4.455 ms          | 1.486 ms     | 1.309 ms    |
| mean             | 6.789 ms            | 4.340 ms          | 1.314 ms     | 1.218 ms    |
| throughput       | 147.29 b/sec        | 230.42 b/sec      | 760.84 b/sec | 821.13 b/sec |
| speedup `generate` | 1.00x               | 1.56x             | 5.17x        | **5.57x**       |
| speedup `encode`   | 0.64x               | 1.00x             | 3.30x        | **3.56x**       |

### Qwen3-0.6B benchmark: PyTorch `llm.encode()` variants vs legacy TRT path

- GPU: L4
- Batch size: 1
- TRT engine configs (built 2 engines):
  - `max_num_tokens=768` for shorter (ISL <= 512) sequence lengths
  - `max_num_tokens=16384` for longer (ISL > 512) sequence lengths
- Column definition:
  - latency delta vs TRT: uses the best latency from `eager`, `cuda-graph`, or `torch-compile+cuda-graph` and compares it with TRT; **negative means the PyTorch variant is faster than TRT**.

| ISL   | TRT (ms)  | eager (ms) | `cuda-graph` (ms)   | `torch-compile+cuda-graph` (ms) | latency delta vs TRT |
| ----: | ------: | -------: | ------: | ------------: | ----------: |
| 16    | 7.65    | 16.39    | 5.98    | 5.97          | **-21.9%**      |
| 64    | 7.65    | 16.60    | 6.39    | 6.43          | **-16.5%**      |
| 128   | 8.25    | 16.60    | 7.01    | 6.99          | **-15.3%**      |
| 512   | 14.70   | 16.89    | 14.76   | 14.65         | -0.4%       |
| 1024  | 28.20   | 27.93    | 29.35   | 29.15         | -0.9%       |
| 2048  | 62.11   | 61.19    | 61.59   | 61.49         | -1.5%       |
| 4096  | 154.56  | 142.04   | 143.57  | 142.85        | -8.1%       |
| 8192  | 411.93  | 375.75   | 378.39  | 376.89        | -8.8%       |
| 12288 | 767.49  | 696.46   | 699.78  | 698.69        | -9.3%       |
| 16000 | 1172.01 | 1064.07  | 1065.81 | 1067.61       | -9.2%       |

### Integration tests

Add new tests to `tests/integration/defs/accuracy/test_llm_api_pytorch_encode.py`:

- `TestEncoderEncode` (encoder-only architectures):
  - `test_encoder_encode_cuda_graph_matches_eager_logits[bert-yelp]` — Tight numerical bound (`rtol=atol=1e-3`) between graph replay and eager output on the same model, to catch subtle launch-parameter or metadata drift that HF-tolerance wouldn't flag.

- `TestDecoderEncode` (decoder-as-encoder):
  - `test_decoder_encode_cuda_graph_matches_eager_logits[tinyllama-1.1b]` — Tight graph-vs-eager bound for the single-prefill decoder path.

### Unit tests

Add new tests to `tests/unittest/llmapi/test_llm_encode.py`:

- CUDA graph path: `test_encode_cuda_graph_and_return_raw_logits` — exercises capture + replay end-to-end and verifies `return_raw_logits` output wrapping on the graph path.
- Backend guard: `test_trtllm_gen_is_supported_rejects_none_kv_cache_manager` — asserts the `trtllm_gen` backend refuses to be selected without a `kv_cache_manager`.

`tests/unittest/llmapi/test_llm_args.py` adds coverage for the new `CudaGraphConfig` fields (auto-derivation of `max_num_token` / `max_seq_len`, mismatch error paths, sorted-list normalization).

### CI wiring

The new integration tests are added to:
- `tests/integration/test_lists/test-db/l0_a100.yml`
- `tests/integration/test_lists/qa/llm_function_core.txt`
- `tests/integration/test_lists/qa/llm_function_l20.txt`
- `tests/integration/test_lists/qa/llm_function_rtx6k.txt`

CC: @schetlur-nv @nvrohanv @amukkara @symphonylyh 

## PR Checklist

- [X] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help


tensorrt_llm/llmapi/llm_args.py
```diff
@@ -107,9 +107,7 @@ def Field(default: Any = ...,
 
 
 class CudaGraphConfig(StrictBaseModel):
```
2ez4bz: Should we make use of[ discriminated unions](https://pydantic.dev/docs/validation/latest/concepts/unions/) for this? We could have decode and encode specific models that have their own fields (some could even be shared through a small base class), and the `LlmArgs` object could have a union type instead.

tensorrt_llm/llmapi/llm.py
```diff
@@ -855,20 +847,30 @@ def encode(
             k: v
             for k, v in model_kwargs.items() if k not in _RESERVED_KEYS
         }
+
+        if filtered_kwargs and engine.encoder_cuda_graph_runner.enabled:
+            raise NotImplementedError(
+                "LLM.encode(..., **model_kwargs) is not supported when encoder CUDA "
+                "graphs are enabled. Disable encoder CUDA graphs or omit model_kwargs. "
+                f"Unsupported keys: {sorted(filtered_kwargs)}")
```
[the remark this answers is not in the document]
tingyangk: The plan is to support this in a following PR.

tensorrt_llm/_torch/pyexecutor/cuda_graph_runner.py
```diff
@@ -562,3 +566,432 @@ def clear(self):
...
+                    capture_inputs["input_ids"].copy_(
+                        sliced_static_tensors_cpu["input_ids"],
+                        non_blocking=True)
+                    capture_inputs["position_ids"].copy_(
+                        sliced_static_tensors_cpu["position_ids"],
+                        non_blocking=True)
+                    attn_md._seq_lens_cuda.copy_(attn_md._seq_lens,
+                                                 non_blocking=True)
+                output = forward_fn(capture_inputs)
+
```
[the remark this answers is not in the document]
tingyangk: The plan is to support this in a following PR.

yingguo-trt (approved):

jieli-matrix (approved):

tensorrt_llm/_torch/pyexecutor/cuda_graph_runner.py
```diff
@@ -562,3 +566,432 @@ def clear(self):
...
+    Restricted to `TrtllmAttentionMetadata` — FlashInfer's per-batch planner state is not compatible with CUDA graph capture/replay.
+    """
+
+    WARMUP_STEPS = 2
+
+    def __init__(self, config: EncoderCUDAGraphRunnerConfig):
+        self.config = config
+
+        self.enabled = config.use_cuda_graph
```
2ez4bz, line 604: Just checking: do these need to be public attributes?

tensorrt_llm/llmapi/llm_args.py
tingyangk: Good call, thanks. I’ll split this into encode/decode-specific CUDA graph configs with a discriminated union, while keeping the existing decode config path backward-compatible.

tensorrt_llm/_torch/pyexecutor/cuda_graph_runner.py
tingyangk: I think most attributes can actually be private except for `enabled`, which is used in `model_engine.py`. I was following the existing `CudaGraphRunner` pattern here.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -1503,9 +1503,13 @@ def is_supported(
     Returns:
         Tuple of (is_supported, reason_if_not_supported).
     """
-    kv_cache_dtype = torch_dtype_to_binding(q.dtype)
-    if kv_cache_manager is not None:
-        kv_cache_dtype = kv_cache_manager.dtype
+    if kv_cache_manager is None:
```
yuxianq: @yihwang-nv Please review this part, thanks~

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1463,6 +1513,13 @@ def _run(
                 global_layer_idx=self.layer_idx,
             )
         else:
+            # max_context_q_len_override is only set when encoder CUDA graphs are enabled.
+            max_context_q_len_override = metadata.max_context_q_len_override
+            if max_context_q_len_override is not None:
+                assert metadata.is_cuda_graph
+                assert metadata.num_generations == 0
+                assert metadata.kv_cache_manager is None
+                assert metadata.num_contexts == metadata.num_seqs
```
yuxianq: All checks should happen before we dispatch between trtllm_gen_attention and thop.attention, so that we can switch between them seemlessly once they support the same feature, please move these assertions to outer scope.

tingyangk: Done. Thank you!

tensorrt_llm/llmapi/llm_args.py
tingyangk: Finished the fix. Thank you!

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
tingyangk: This check is done by the attention refactor.

2ez4bz (approved):
Approving changes to `tensorrt_llm/llmapi/mm_encoder.py`.

govind-ramnarayan (approved):

tensorrt_llm/llmapi/llm.py
```diff
@@ -729,8 +729,9 @@ def encode(
         self,
         inputs: Union[PromptInputs, Sequence[PromptInputs]],
         add_special_tokens: bool = True,
+        return_raw_logits: bool = False,
```
ixlmar, line 732: Could we change this to use the `batch_indexed_model_output` option from #14693 instead?

tensorrt_llm/_torch/pyexecutor/cuda_graph_runner.py
```diff
@@ -562,3 +566,432 @@ def clear(self):
         del self.memory_pool
         self.memory_pool = None
         torch.cuda.empty_cache()
+
+
+EncoderKeyType: TypeAlias = Tuple[int, int, int]
+
+
+@dataclass
+class EncoderCUDAGraphRunnerConfig:
```
Superjomn, line 575: Is this holding identical options from `llm_args.EncodeCudaGraphConfig`? If so, how about reusing that here directly?

tensorrt_llm/llmapi/llm.py
tingyangk: The main reason why I added `return_raw_logits` is to skip the loop of packing the EncoderOutput (to reduce CPU work). In some very tiny models, skipping the loop can reduce the latency a bit.
I suggest to keep. both - my `return_raw_logits` and your `batch_indexed_model_output`. Thanks!

tensorrt_llm/_torch/pyexecutor/cuda_graph_runner.py
tingyangk: No, they are different. `llm_args.EncodeCudaGraphConfig` only contains user-facing config, where `EncoderCUDAGraphRunnerConfig` holds more information needed for `EncoderCUDAGraphRunner`.

This structure is the same as generation CUDA Graphs - `llm_args.CudaGraphConfig`, `CUDAGraphRunnerConfig`, and `CUDAGraphRunner`!

juney-nvidia (approved):

PR #14348, opened 2026-05-20
[None][perf] Skip host zero_() of FP8 quant scale buffer by writing 0 in kernel

mingyangHao:
## Description

`fp8_quantize_1x128_packed_ue8m0` is used as the SM100 (Blackwell) entry point for FP8
block-scale GEMM inputs in DSv4 and other FP8 paths. Before this PR, the op zeroed the
entire packed scale buffer on the host before launching the kernel:

```cpp
at::Tensor packedBuf = at::detail::empty_cuda({num_packed_sf_k, m_aligned}, ...);
packedBuf.zero_();   // <-- emits a vectorized_elementwise_kernel<FillFunctor<int>>
```

The zero-fill exists because the kernel previously did an early `return` for padded
`m_idx >= m` rows, leaving the `[m, m_aligned)` tail unwritten. Downstream `deep_gemm`
reads `m_aligned` rows per K block, so the tail had to be deterministic 0.

In a real DSv4 generation worker the op runs ~17,800 times per benchmark, and each call
adds a separate `FillFunctor<int>` kernel launch (~1.1 µs GPU + ~2-3 µs host dispatch),
totalling ~20 ms of pure overhead per worker — see the 0514 ctx-only-8k-mtp3 nsys trace.

## Solution

Move the zero-fill into the kernel. Padded rows now compute `packed = 0u` explicitly
and write through the normal STG path; the host `zero_()` is removed:

```cpp
bool const row_in_range = (m_idx < m);
uint32_t packed = 0u;
if (row_in_range) {
    // existing load → amax → ue8m0 → fp8 STG → pack 4 scales into `packed`
}
if (lane_id == 0) {
    // grid covers exactly [0, m_aligned), so this STG is always in-bounds.
    packed_scale_output[packed_sf_k_idx * scale_leading_dim_uint32 + m_idx] = packed;
}
```

The grid is `(num_packed_sf_k, ceil(m / WarpsPerBlock))`, so `m_idx ∈ [0, m_aligned)`
exactly — the extra STG for padded rows is in-bounds and adds at most a few int32
writes per launch.

## Test Coverage

Two tests added under `tests/unittest/_torch/thop/parallel/test_fp8_quantize.py`:

- `test_fp8_quantize_1x128_packed_ue8m0_matches_legacy`
  Compares the fused op output (FP8 values + packed UE8M0 scales) against the legacy
  `(fp8_quantize_1x128 use_ue8m0=True → get_col_major_tma_aligned_packed_tensor)` path
  for `m ∈ {1, 3, 4, 7, 16, 127, 1024}`, `k ∈ {128, 256, 512, 4096, 7168}`. Must be
  bit-identical.

- `test_fp8_quantize_1x128_packed_ue8m0_padded_rows_are_zero`
  Allocates a sentinel-filled buffer the same size as the op's scale output, frees it,
  calls the op (CUDA caching allocator likely returns the same slab), and asserts the
  `[m, m_aligned)` padded rows are all zero — the property that previously required
  the host-side `zero_()`.

Both tests pass on B300 SM100.

### Microbench (B300, single op invocation, p50 of 200 iters)

| Shape (m, k)          | Baseline (µs) | Fix (µs) | Δ      |
|-----------------------|--------------:|---------:|-------:|
| 1, 7168               | 17.22         | 13.44    | -22%   |
| 8, 7168               | 17.02         | 13.31    | -22%   |
| 64, 7168              | 17.12         | 13.28    | -22%   |
| 256, 7168             | 16.77         | 13.22    | -21%   |
| 1024, 7168            | 17.18         | 13.73    | -20%   |
| 256, 1536 (gate proj) | 16.96         | 13.34    | -21%   |
| 3, 256 (padded)       | 16.86         | 13.38    | -21%   |
| 127, 4096 (padded)    | 17.34         | 13.28    | -23%   |

Saving is ~3.7 µs per call (one launch + the FillFunctor GPU time).

### E2E (DSv4-Pro DEP8 on b300, same container, single run)

| Config         | Metric                    | Baseline   | Fix        | Δ        |
|----------------|---------------------------|-----------:|-----------:|---------:|
| conc=512 mtp=3 | output throughput (tok/s) | 12,769.94  | 13,300.43  | **+4.15%** |
| conc=512 mtp=3 | per-user tok/s            | 27.93      | 29.48      | **+5.56%** |
| conc=512 mtp=3 | avg request latency (ms)  | 38,996.01  | 37,510.53  | -3.81%   |

`conc=8 mtp=3` was also run but the single-run delta sat inside the run-to-run noise
band, so the meaningful e2e signal is the `conc=512` row above. The
`conc=512` measurement is fully single-node, same docker container,
swapping only `libth_common.so` between runs.

## PR Checklist

- [x] PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- [x] Test cases are provided for new code paths.

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/fp8_blockscale_quant_packed.cu
+++ b/cpp/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/fp8_blockscale_quant_packed.cu
@@ -183,9 +183,9 @@ __global__ void fp8_quantize_1x128_packed_kernel_impl(__nv_fp8_e4m3* __restrict_
     }
 
     // Always write the packed scale — `packed` is 0 for padded rows. The grid
-    // covers m_idx ∈ [0, m_blocks * WarpsPerBlock), i.e. exactly the physical
-    // [0, m_aligned) extent of packed_scale_output, so the STG is in bounds.
-    if (lane_id == 0)
+    // covers the full [0, scale_leading_dim_uint32) leading dim (rounded up to
+    // WarpsPerBlock), and the m_idx guard drops the few rows past the buffer end.
+    if (lane_id == 0 && m_idx < scale_leading_dim_uint32)
     {
         packed_scale_output[static_cast<int64_t>(packed_sf_k_idx) * scale_leading_dim_uint32 + m_idx] = packed;
     }
@@ -202,7 +202,10 @@ void launch_fp8_quantize_1x128_packed_bf16_e4m3(__nv_fp8_e4m3* fp8_output, int32
 {
     constexpr int kWarpsPerBlock = 4;
     int const num_packed_sf_k = (((k + 127) / 128) + 3) / 4;
-    int const m_blocks = (m + kWarpsPerBlock - 1) / kWarpsPerBlock;
+    // Cover the full TMA-padded leading dim so the entire [0, scale_leading_dim_uint32)
+    // extent of packed_scale_output is written (in-kernel zero for rows past `m`),
+    // regardless of how the caller padded the input.
+    int const m_blocks = (scale_leading_dim_uint32 + kWarpsPerBlock - 1) / kWarpsPerBlock;
     dim3 const grid(num_packed_sf_k, m_blocks, 1);
     dim3 const block(kWarpsPerBlock * 32, 1, 1);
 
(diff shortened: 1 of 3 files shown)


lishicheng1996-nv (approved):

cpp/tensorrt_llm/kernels/cutlass_kernels/fp8_blockscale_gemm/fp8_blockscale_quant_packed.cu
liji-nv: My agent change m to scale_leading_dim_uint32 to make sure all value within scale_leading_dim_uint32 is filled no matter what the padding size is. https://github.com/NVIDIA/TensorRT-LLM/pull/14358/changes

mingyangHao: Good catch — fixed in bcf2d3f17f to match #14358: grid sized on scale_leading_dim_uint32, store guarded by m_idx < scale_leading_dim_uint32.

liji-nv (approved):

PR #14351, opened 2026-05-20
[None][chore] Replace cache_salt_id impl with main's TreeTaskId (cp #13793)

lancelly:
## Summary
Replaces feat/deepseek_v4's parallel `cache_salt_id` impl (introduced as a side effect of PR #13624 / commit 07adb42545, which mainly added V1↔V2 KV event hash matching) with main's `TreeTaskId` abstraction from PR #13793.

Both implementations achieve identical user-visible behavior (per-request salt-based KV cache reuse isolation). The differences are internal:

| | DSv4 (before) | this PR (after) |
|---|---|---|
| Radix tree | knows `lora_task_id` + `cache_salt_id` as 2 fields | opaque `tree_task_id: int \| bytes \| None` |
| Hash mixing | inside the tree, via `_make_root_key` + tag bytes | above the tree, in `_KVCache._make_tree_task_id` |
| `RootBlock` | stores both fields | stores single `tree_task_id` |

The opaque-seed shape allows mixing future dimensions (e.g. multi-modal digest) without changing the tree. It is also the prerequisite for cherry-picking PR #14140 (ReuseScope NamedTuple refactor) and PR #14333 (KV cache reuse probe), which will follow in subsequent PRs.

## Conflict resolution notes
- **Dummy/warmup requests**: adopted main's design — no salt on dummies. Rationale: dummy paths `stop_committing()` anyway so the salt choice is inert, but explicitly passing `None` avoids coupling synthetic data to any salted branch.
- **IndexMapper saturation guard**: kept DSv4's `None`-return path in `_create_kv_cache` (needed for disagg gen-trans-in-progress requests).
- **Event manager / try_get_tree / other DSv4 hooks**: preserved verbatim in the touched files — git auto-merge handled these because #13793 does not touch the same lines.


lancelly:
@peihu-nv and @eopXD  Could you please take a look at this PR? Currently the salt implementation seems to diverge on feat/deepseek_v4 and main branch. To cherry-pick other kv manager related code onto feat/deepseek_v4, we need to resolve this conflict first.

peihu-nv (approved):
Direction LGTM, approving to unblock the stack assuming follow up PR to land right after.

- Seems to me that test_v2_root_key_distinguishes_lora_from_cache_salt_id still calls RootBlock.make_key(None, None), now it is a single arg.
- _event_manager._root_attrs_from_root_block still reads lora_task_id/cache_salt_id off RootBlock, but this PR removes both → silently (None, None) for every event → V1-compat hashes break for any LoRA/salt request and Dynamo routing degrades. Not sure if dropping V1 compat is intentional or might need to address this in the ReuseScope follow-up (root.reuse_scope.lora_id/.salt).

PR #14353, opened 2026-05-20
[None][chore] ReuseScope refactor (cp #14140)

lancelly:
## Summary
Stacked on top of #14351. Cherry-picks #14140 onto `feat/deepseek_v4` and fixes two regressions that surface once that refactor lands on the DSv4 branch.

1. **#14140 (cherry-picked)** — Refactor `cache_salt_id` / `lora_task_id` into a `ReuseScope` NamedTuple. Eliminates positional `(int | None, int | None)` parameters across the radix tree API. Conflict resolution against the DSv4 branch:
   - `_kv_cache_manager.py`: collapse the public `create_kv_cache` / private `_create_kv_cache` split into a single entry point (#14140's shape), but keep the DSv4-only `expected_prompt_length` parameter.
   - `_kv_cache.py`: drop `cache_salt_id` from `__init__` (now folded into `reuse_scope`), keep `expected_prompt_length`.
   - `resource_manager.py`: switch caller from `impl._create_kv_cache(lora_task_id, ..., cache_salt_id=, expected_prompt_length=)` to `impl.create_kv_cache(ReuseScope(lora_id=..., salt=...), input_tokens, id=request_id, expected_prompt_length=...)`. Preserves the DSv4 `IndexMapper` free-slot guard.

2. **Re-migrate router callers from `TreeTaskId` → `ReuseScope`.** PR #14351's `eaddc9ac22` migrated callers to a `TreeTaskId` seed (`_KVCache._make_tree_task_id(None, cache_salt_id)`); #14140 supersedes that API, so two sites in `tensorrt_llm/serve/router.py` and four sites in `tests/unittest/disaggregated/test_router.py` need to be re-migrated to `ReuseScope(salt=cache_salt_id)` (or `ReuseScope()` for the unsalted path). Drops the now-unused `_KVCache` import.

3. **Event-hash V1 compat fix** (addresses @peihu-nv's review feedback on #14351). After #14140, `RootBlock` no longer exposes `lora_task_id` / `cache_salt_id` directly — both fields live on `root.reuse_scope`. `KVCacheEventManager._root_attrs_from_root_block` was still reading them via `getattr(root, "lora_task_id", None)` / `getattr(root, "cache_salt_id", None)`, so every emitted V1-compat event silently received `(None, None)` and the hash collapsed to the same value for any LoRA/salt request. Downstream Dynamo prefix-cache routing depends on those hashes, so this regression would materially degrade hit rate. Fix reads `root.reuse_scope` first and falls back to the legacy attributes (keeps any in-flight pre-refactor `RootBlock` instances working). Includes a regression test (`test_v2_kv_cache_event_manager_v1_hash_reads_root_reuse_scope`) covering: (a) ReuseScope-shaped vs legacy root produce identical V1 hashes for the same lora/salt, (b) different scopes still produce different hashes (no silent collapse), (c) empty `ReuseScope` matches a legacy root with no salt/lora.


peihu-nv (approved):
LGTM. 

Minor follow-up later: a couple tests not in CI still use stale APIs (`cache_salt_id=` / `make_key(None)`).

pcastonguay (approved):

PR #14360, opened 2026-05-20
Features/cascade attention

Fallenpetal:
## Summary by CodeRabbit

* **New Features**
  * Added experimental cascade attention kernel with environment variable configuration support (`TRTLLM_ENABLE_CASCADE_MMHA`, `TRTLLM_CASCADE_MMHA_MIN_BEAM`, `TRTLLM_CASCADE_MMHA_MIN_PREFIX`)

  * Optimized beam search finalization and token path reconstruction
  * Added small-vocabulary-optimized top-k kernel for faster token selection

* **Tests**
  * Added unit tests for cascade attention kernel validation

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14360?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

**[None][feat] Optimize beam-search decode path: Cascade Attention + finalizeKernel + insertUnfinishedPathKernel + radix_topk small-vocab fast path**

## Description

This PR introduces a set of GPU-side optimizations for the beam-search decode pipeline. All changes are gated behind env vars / fall back to the original implementations when preconditions are not met, so existing models keep bit-identical behavior by default.

### 1. Cascade Attention (new kernels: `cascade_prefix_mqa_kernel` + `cascade_suffix_decode_kernel`)

Implements the three-stage shared-prefix attention from
*"Cascade Inference: Memory Bandwidth Efficient Shared Prefix Batch Decoding"*
(https://flashinfer.ai/2024/02/02/cascade-inference.html), specialized for TRT-LLM's MMHA path.

**Motivation.** The baseline `masked_multihead_attention_kernel` launches one block per (head, beam) and re-loads the **shared prompt KV** from HBM `beam_width` times per request — wasteful for beam search where every beam shares the same prefix `[0, L_p)`.

**Design — two kernels (Phase 3 fused into Phase 2):**

| Phase | Kernel | Grid | What it does |
|---|---|---|---|
| 1 | `cascade_prefix_mqa_kernel` | `(num_heads, num_requests, ⌈beam/16⌉)` | All beams of a request attend to the **shared** prefix KV. KV is loaded **once** into SMEM; the beam dimension is packed as MMA-M=16 so we drive **m16n8k16 Tensor Core MMAs** for both `Q·Kᵀ` and `P·V`. Online softmax `(m, l)` and FP32 `O_acc` are kept across tile iterations; cp.async double-buffers K/V tiles so HBM transfer overlaps with compute. |
| 2 (fused) | `cascade_suffix_decode_kernel` | `(num_heads, total_seqs)` | Per-(head, beam) suffix decode following `cache_indir`. **Three things fused into one launch**: (a) computes the current step's K/V (bias + RoPE) and the GQA leader Q-head writes them to KV cache; (b) walks suffix tokens `[L_p, T)` with online softmax; (c) merges with the prefix partial **in registers** at the epilogue, eliminating the standalone merge kernel and its HBM round-trip. |

**Engineering details:**
- `PersistentWorkspace`: single `cudaMalloc` on first call (during TRT warmup, before graph capture); reused for every subsequent call → **zero runtime allocation**, fully **CUDA Graph capture-compatible**.
- Prefix length is **never copied to host** — `params.input_lengths` device pointer is read in-kernel via `__ldg`, so no D2H/stream sync on the launch path.
- SMEM banks: each Q/K/V row carries 16B padding (`Dh + 8` bf16/half) to break the 8-way bank conflict on `Q·Kᵀ` LDS.
- `block_sum`: warp-shuffle + 1-step cross-warp reduce (3 barriers vs 9 in the tree version).
- `cascade_eligible(params)`: strict gating (env var `TRTLLM_ENABLE_CASCADE_MMHA`, beam ≥ `MIN_BEAM`, no cross-attn / position-shift / block-sparse / softcap / int8-fp8 KV / non-NeoX RoPE / partial RoPE / Dh ∉ {64, 128} → falls back to baseline MMHA).

**Expected gains** (beam=16, prefix=1000, suffix=10, Dh=128):
- Prefix HBM traffic: `beam · L_p` → `1 · L_p` (≈ 16× less)
- Tensor Core utilization on prefix path (was gemv M=1, now gemm M=16)
- Kernel launches per step on the cascade path: 4 → **2** (Phase 0 KV-write + Phase 3 merge fused into Phase 2)

### 2. `finalizeKernel` → `finalizeKernelOptimized`

Same task (sort `≤ 2·nBM` CBA candidates by normalized score, then move `outputIds / sequenceLengths / cumLogProbs / logProbs` to the final outputs), but rebuilt for parallelism when `nCBA > 32` (i.e. larger beam widths):

- **Sort** (`nCBA > 32` branch): replace the serial `for i in [0, nBM)` block-reduce-max + single-thread scan loop with a **parallel bitonic sort** over `nPow2 = next_pow2(nCBA)` elements, with simultaneous `(score, index)` swaps so the original CBA position is preserved. Sentinel `-FLT_MAX` is filled into `[nCBA, 2·nBM)` to keep the network well-defined when `nCBA` is not a power of two. Synchronization cost drops from `O(nBM)` to `O(log² nPow2)`.
- **Data movement**: the per-beam outer loop is replaced by a **flat 1-D parallel loop** over `globalIdx ∈ [0, nBM·nMSL)`, yielding even per-beam load-balance (no long-tail when sequence lengths differ).
- **SMEM reuse**: `tempIdx` for the bitonic index buffer is overlaid on `smemSL`, which is only used after sorting.
- The `nCBA ≤ 32` warp-shuffle path is unchanged (already optimal for small beams).

### 3. `insertUnfinishedPathKernel` → `insertUnfinishedPathKernelOptimized`

The original kernel launched **one thread per (batch)** and serially processed all `nBM` beams inside it, including a serial `parentIds` walk for each beam — making it a strict serialization point in the beam-search path.

New kernel:
- Launches `bh.nBatchSize` blocks × `min(nBM, 1024)` threads, **one thread per beam**.
- Each thread independently walks its own `parentChain` (path reconstruction is now `O(step)` per thread, fully parallel across beams).
- The `outputIds` and `logProbs` copies for each beam happen in parallel; `numBeamsCBA` is updated once per block via `atomicAdd(..., nBM)` from `tid==0`.
- Boundary handling preserved (`batchDones`, `step < 0` guard, optional `bOutputLogProbs` path).

For typical `beam_width = 4..16`, per-batch latency drops by roughly the beam-width factor.

### 4. `radix_topk_one_block_kernel` → small-vocab fast path: `small_vocab_topk_kernel`

For beam-search Stage-1 Top-K with **small vocab** (`N ≤ 4096`, e.g. draft / speculative heads, embedding-shared small heads, classifier heads), the general radix kernel is over-engineered. New specialized kernel:

| Aspect | General `radix_topk_one_block_kernel` | New `small_vocab_topk_kernel` |
|---|---|---|
| Input residence | Re-read from global per pass | Loaded once into `s_values[MaxVocab]`; **all passes hit SMEM** |
| `BitsPerPass` | 11 (2048 bins) | **8** (256 bins) — pass count for fp16 stays at 2, but histogram drops from 8 KB → 1 KB |
| Histogram scan | Block-wide multi-step | Serial inclusive scan by `tid==0` over 256 bins (faster at this size, no warp synchronization) |
| Stability machinery | `IS_STABLE` atomicMin loops | Removed — beam-search Stage 1 doesn't require sorted output |
| BlockSize | 512 | **256** (better SM occupancy with the still-modest SMEM, sweet spot for 4096 / 256 = 16 elems/thread, low atomic contention vs 256 buckets) |
| External CUB sort | Required | **Not needed** (output is unsorted, downstream beam logic doesn't care) |
| Output | One pass | Two-phase (strictly-better elements + boundary fill) — keeps semantics identical for ties |

Net effect: the small-vocab path skips one kernel launch, cuts SMEM by ~7 KB, and stays in shared memory for the entire radix pipeline.

## Test Coverage

- **Numerical correctness**
  - Cascade Attention path is gated by `TRTLLM_ENABLE_CASCADE_MMHA`. Comparison harness toggles the env var and asserts max abs/rel error vs the baseline MMHA path on `(Dh ∈ {64, 128}) × (T ∈ {half, bf16}) × (beam ∈ {4, 8, 16}) × (prefix_len ∈ {32, 256, 1024}) × (suffix_len ∈ {0, 1, 8})`.
  - `finalizeKernelOptimized` validated against `finalizeKernel` on `nBM ∈ {4, 8, 16, 24, 32}`, `nCBA ∈ {nBM, 2·nBM}`; checks the moved `outputIds / sequenceLengths / cumLogProbs / logProbs` and the resulting rank order.
  - `insertUnfinishedPathKernelOptimized` validated against the serial version on randomized `parentIds` chains across `beam ∈ {1..16}`, with and without `logProbs`.
  - `small_vocab_topk_kernel` validated against the general `radix_topk_one_block_kernel` on `N ∈ {64, 512, 1024, 2048, 4096}`, `K ∈ {1, 4, 16, 64}`, fp16 / bf16, `select_min ∈ {true, false}`, including ties on the boundary value.
- **End-to-end smoke**: beam-search decode (Llama-3 / Qwen-style configs) with `beam ∈ {4, 8, 16}` produces identical `outputIds` to the baseline (cascade off) and improves per-step latency when cascade is on.
- **CUDA Graph**: Cascade path is captured + replayed without re-allocation (verified via `cudaGraphInstantiate` + repeated `cudaGraphLaunch`); `cascade_eligible` short-circuits before any allocation when not eligible.
- **Fallback**: when `TRTLLM_ENABLE_CASCADE_MMHA` is unset or any precondition fails, the dispatcher falls back to the original MMHA kernel — confirmed by unit tests that exercise the rejection paths in `cascade_eligible`.
- **Existing tests**: `tests/unittest/_torch/...` beam-search and decoder kernel tests pass unchanged.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/common/envUtils.cpp
+++ b/cpp/tensorrt_llm/common/envUtils.cpp
@@ -274,6 +274,60 @@ bool getEnvEnablePDL()
     return enablePDL;
 }
 
+bool getEnvEnableCascadeMmha()
+{
+    static bool const enable = getBoolEnv("TRTLLM_ENABLE_CASCADE_MMHA");
+    return enable;
+}
+
+int getEnvCascadeMmhaMinBeam()
+{
+    static std::once_flag flag;
+    static int minBeam = 2;
+    std::call_once(flag,
+        [&]()
+        {
+            char const* env = std::getenv("TRTLLM_CASCADE_MMHA_MIN_BEAM");
+            if (env)
+            {
+                int v = std::atoi(env);
+                if (v >= 1)
+                {
+                    minBeam = v;
+                }
+                else
+                {
+                    TLLM_LOG_WARNING("Invalid TRTLLM_CASCADE_MMHA_MIN_BEAM, using default %d", minBeam);
+                }
+            }
+        });
+    return minBeam;
+}
+
+int getEnvCascadeMmhaMinPrefix()
+{
+    static std::once_flag flag;
+    static int minPrefix = 256;
+    std::call_once(flag,
+        [&]()
+        {
+            char const* env = std::getenv("TRTLLM_CASCADE_MMHA_MIN_PREFIX");
+            if (env)
+            {
+                int v = std::atoi(env);
+                if (v >= 0)
+                {
+                    minPrefix = v;
+                }
+                else
+                {
+                    TLLM_LOG_WARNING("Invalid TRTLLM_CASCADE_MMHA_MIN_PREFIX, using default %d", minPrefix);
+                }
+            }
+        });
+    return minPrefix;
+}
+
 bool getEnvUseUCXKvCache()
 {
     static bool const useUCXKVCache = getBoolEnv("TRTLLM_USE_UCX_KVCACHE");
--- a/cpp/tensorrt_llm/common/envUtils.h
+++ b/cpp/tensorrt_llm/common/envUtils.h
@@ -60,6 +60,19 @@ int getEnvMmhaKernelBlockSize();
 // Whether PDL is enabled.
 bool getEnvEnablePDL();
 
+// Whether the experimental cascade attention kernel is enabled (replaces
+// masked_multihead_attention_kernel for beam-search decoding).
+// Controlled by env var TRTLLM_ENABLE_CASCADE_MMHA (default: false).
+bool getEnvEnableCascadeMmha();
+
+// Minimum beam_width required to dispatch the cascade kernel.
+// Controlled by TRTLLM_CASCADE_MMHA_MIN_BEAM (default: 2).
+int getEnvCascadeMmhaMinBeam();
+
+// Minimum shared prefix length required to dispatch the cascade kernel.
+// Controlled by TRTLLM_CASCADE_MMHA_MIN_PREFIX (default: 256).
+int getEnvCascadeMmhaMinPrefix();
+
 template <typename KernelFn, typename... Args>
 inline void launchWithPdlWhenEnabled(char const* name, KernelFn kernelFn, dim3 grid, dim3 block, size_t dynamicShmSize,
     cudaStream_t stream, Args&&... args)
--- a/cpp/tensorrt_llm/kernels/beamSearchKernels/beamSearchKernelsTemplate.h
+++ b/cpp/tensorrt_llm/kernels/beamSearchKernels/beamSearchKernelsTemplate.h
@@ -626,7 +626,15 @@ void beamSearchKernelLauncher(
         void* pTopK = reinterpret_cast<void*>(reinterpret_cast<char*>(workspace) + offset);
 
         // Stage 1
-        invokeTopkLastDim<T>(nBS * nBMIn, nV, nBMOut * 2, true, logProbs, pStage1LogProbs, pStage1Ids, pTopK, stream);
+        if (nV <= 4096)
+        {
+            // Small-vocab optimized path: smem-resident radix, no workspace needed
+            invokeSmallVocabTopkLastDim<T>(nBS * nBMIn, nV, nBMOut * 2, true, logProbs, pStage1LogProbs, pStage1Ids, stream);
+        }
+        else
+        {
+            invokeTopkLastDim<T>(nBS * nBMIn, nV, nBMOut * 2, true, logProbs, pStage1LogProbs, pStage1Ids, pTopK, stream);
+        }
         sync_check_cuda_error(stream);
 
         int nThread = min(roundUp(nBMIn * nBMOut * 2, 32), 1024);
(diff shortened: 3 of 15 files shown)


StanleySun639:
Want to confirm: Are you pushing to the wrong branch？Does this feature have to be in release/1.2?  If not,  please submit to main branch instead.

StanleySun639 (changes requested):
Does this feature have to be in release/1.2? If not, please submit to main branch instead.

jieli-matrix (changes requested):
Seems an external dev to contribute new feats when they're working on release/1.2. plz rebase on main branch to see if it works in local.

Fallenpetal:
> Does this feature have to be in release/1.2? If not, please submit to main branch instead.

Sorry, thanks for your reply. it doesn't need to be in release/1.2 , I just based my work on that branch. I'll switch the PR to target main. Thanks for the guidance!

Fallenpetal:
> Seems an external dev to contribute new feats when they're working on release/1.2. plz rebase on main branch to see if it works in local.

Sorry, thanks for your reply. it doesn't need to be in release/1.2 , I just based my work on that branch. I'll switch the PR to target main. Thanks for the guidance!

ruodil (approved):

PR #14363, opened 2026-05-20
[None][fix] Time-weighted _avg_history_length to avoid commit-burst contamination

eopXD:
## Summary

Follow-up to #14098 that addresses the same `_avg_history_length` contamination problem via a different mechanism.

#14098 removed `self._avg_history_length.update(history_length)` from `_KVCache.resize()`'s explicit-history branch because that branch is fired transitively by `commit()` → `history_length` setter → `resize(None, X)` (`_kv_cache.py:717-718` → `:678-682` → `:485`), oversampling commit-driven bumps and biasing the per-cache unweighted average. The side effect is that `_avg_history_length.value` is now frozen at the init seed for the rest of the cache's lifetime, which makes the per-cache stat redundant with `manager._avg_reused_length` (already updated with the same init value at line 283).

This PR keeps the update call in `resize()` and instead switches `_avg_history_length` from `Average` (unweighted append-mean) to a new `TimeWeightedAverage` that computes a time-weighted lifetime mean of `history_length`. Rapid commit-driven `update()` calls now contribute proportionally less weight because the elapsed wall-clock between them is small, so the metric is safe from commit-burst contamination while still tracking `history_length` over the cache's life.

## Time source

`TimeWeightedAverage` calls Python stdlib **`time.monotonic()`** in three places: `__init__` (records `t_init`), `update()` (closes the prior segment), and the `value` property (closes the open segment up to "now"). No new import dependency at the cache layer — `KVCacheManager.__init__` already calls `time.monotonic()` for `_last_adjustment_time`.

Measured cost:

| Op | Pure-Python | Notes |
|---|---|---|
| `time.monotonic()` | ~62 ns | VDSO on Linux, no syscall |
| `Average.update()` (PR baseline) | ~87 ns | for reference |
| `TimeWeightedAverage.update()` | ~133 ns | +46 ns vs `Average` |
| `TimeWeightedAverage.value` | ~127 ns | once per cache at `close()` |
| `TimeWeightedAverage` ctor | ~179 ns | once per cache at `__init__` |

Per request that generates ~1000 tokens this adds <50 µs total, <0.1% of request latency. Under mypyc compilation the surrounding Python overhead drops and `time.monotonic()` will dominate, with total `update()` cost dropping closer to ~80 ns.

## Inherited from #14098 (verbatim)

- Init seed for `_avg_capacity` in `_KVCache.__init__`
- Removal of redundant close-time `_avg_capacity.update()`
- New `KVCacheManager._num_closed_kv_caches` counter (slot, annotation, init=0), bumped at the tail of `_KVCache.close()`

## Summary by CodeRabbit

* **Refactor**
  * Improved KV cache statistics tracking to use time-weighted averaging for more accurate history length measurements.
  * Enhanced cache closure accounting for better lifecycle management and resource monitoring.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14363?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
+++ b/tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
@@ -82,7 +82,7 @@ from .._utils import (
     unwrap_rawref,
     value_or,
 )
-from ._moving_average import Average
+from ._moving_average import Average, TimeWeightedAverage
 
 if TYPE_CHECKING:
     from ._kv_cache_manager import KVCacheManager, ScratchDesc
@@ -224,7 +224,7 @@ class _KVCache:
     _finish_event: CachedCudaEvent | None
 
     _tokens_per_block: int
-    _avg_history_length: Average
+    _avg_history_length: TimeWeightedAverage
     _avg_capacity: Average
 
     _ssm_blocks: TypedIndexList[BeamIndex, TypedIndexList[LifeCycleId, BlockPage]]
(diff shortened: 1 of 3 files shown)


eopXD:
Dropping, Yao has future planning.

PR #14370, opened 2026-05-20
[TRTLLM-12535][chore] Refactor fast path (token ID space) preprocessing logic out to the input preprocessor methods only

moraxu:
## Summary by CodeRabbit

* **Refactor**
  * Unified text-prompt multimodal input interfaces across many vision/multimodal model processors (Gemma3VL, Gemma4MM, HyperClovaX, Kimi K25, Llama4, LLaVA-Next, Mistral3, Nemotron Nano, Phi4MM, Qwen2VL, Qwen3VL, VILA).
  * Reorganized input-processor registry to simplify multimodal prompt handling and dispatch logic for tokenized and text-based inputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14370?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Summary

Encapsulate the VLM tokenized+MM fast path inside `BaseMultimodalInputProcessor` so the registry no longer hosts model-specific stitching logic, and so `llm.py` no longer carries the tokenized-input detokenization guard.

## Why

Previously the fast path (handling `prompt_token_ids + multi_modal_data` without detokenizing) was spread across three layers: a detokenization guard in `llm.py`, a `tokenized_multimodal_process` block in `registry.py`, and per-model hooks (`get_text_with_mm_placeholders`, `expand_prompt_token_ids_for_mm`) on the input processors. The registry - a generic, model-agnostic component - ended up merging model-specific auxiliary streams (e.g. Nano V3 Omni's video `evs_ids`) and carrying a three-armed dispatch in `multimodal_hashing_process` for precomputed-vs-not inputs.

## What changed

- **`BaseMultimodalInputProcessor`** now owns the dispatch:
  - A concrete `__call__` that routes `prompt_token_ids + multi_modal_data` (no `prompt`) to the fast path when the subclass opts in, and otherwise detokenizes and falls through to the text path.
  - An abstract `call_with_text_prompt` - the renamed existing `__call__` on every subclass (the per-model HF-processor path).
  - A concrete `call_with_token_ids` that builds the dummy placeholder text, runs `call_with_text_prompt` on it to populate the multimodal payload, calls `expand_prompt_token_ids_for_mm`, and merges any returned `mm_data_updates` into `extra_processed_inputs["multimodal_data"]`. Raises `NotImplementedError` if the subclass hasn't opted into the fast path.
  - A `supports_token_id_mm_expansion` class flag (`ClassVar[bool]`, default `False`; `True` on LlavaNext + NanoV2VL) that gates the fast path - replacing the old `hasattr`-based probe.
  - A `_detokenize_to_text_prompt` helper (decode `prompt_token_ids` → text, preserve `multi_modal_data` / `mm_processor_kwargs`, flip `add_special_tokens` off) used by both `__call__` and `attach_multimodal_embeddings`.
  - `attach_multimodal_embeddings` is now a Template Method wrapper: it detokenizes for non-fast-path VLMs, then delegates to a new `_attach_multimodal_embeddings_impl` hook that subclasses override.
- **`create_input_processor_with_hash`** is simplified: `tokenized_multimodal_process` and `process_tokenized_prompt_maybe_hash` are deleted, `multimodal_hashing_process` collapses to a single `input_processor(...)` call (and now raises rather than returning an empty prompt when MM token lengths can't be determined, so the caller falls back instead of silently caching hashing-supported), and `input_processor_wrapper` no longer branches on tokenized-MM input shape.
- **All twelve `BaseMultimodalInputProcessor` subclasses** (LlavaNext, NanoV2VL, Llama4, Mistral3, Vila, Qwen2VL, Qwen3VL, Gemma3, Gemma4, Phi4MM, KimiK25, HCXVision) have `__call__` renamed to `call_with_text_prompt`. LlavaNext and Llama4 additionally rename their `attach_multimodal_embeddings` override to `_attach_multimodal_embeddings_impl`.
- **`llm.py`** no longer carries the tokenized-input detokenization guard - the fallback now lives in `BaseMultimodalInputProcessor` (`__call__` and `attach_multimodal_embeddings`), gated on `supports_token_id_mm_expansion`. This keeps the detokenize decision in one place and applies it uniformly to both entry points.

## Test Coverage

Ran individual `curl` requests against Nano Omni V3 for the following scenarios:
- image
- audio
- video
- audio-from-video

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_gemma3vl.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma3vl.py
@@ -102,7 +102,7 @@ class Gemma3InputProcessor(BaseMultimodalInputProcessor,
         return input_ids, pixel_values
 
     @torch.inference_mode()
-    def call_with_txt_prompt(
+    def call_with_text_prompt(
         self, inputs: TextPrompt, sampling_params: SamplingParams
     ) -> Tuple[List[int], Optional[ExtraProcessedInputs]]:
         input_ids, pixel_values = self._preprocess(inputs)
--- a/tensorrt_llm/_torch/models/modeling_gemma3vl.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma3vl.py
@@ -102,7 +102,7 @@ class Gemma3InputProcessor(BaseMultimodalInputProcessor,
         return input_ids, pixel_values
 
     @torch.inference_mode()
-    def __call__(
+    def call_with_txt_prompt(
         self, inputs: TextPrompt, sampling_params: SamplingParams
     ) -> Tuple[List[int], Optional[ExtraProcessedInputs]]:
         input_ids, pixel_values = self._preprocess(inputs)
--- a/tensorrt_llm/_torch/models/modeling_vila.py
+++ b/tensorrt_llm/_torch/models/modeling_vila.py
@@ -1103,7 +1103,7 @@ class VilaInputProcessor(BaseMultimodalInputProcessor,
         return fused_input_ids, mm_features
 
     @torch.inference_mode()  # critical to minimize memory footprint
-    def call_with_txt_prompt(
+    def call_with_text_prompt(
         self, inputs: TextPrompt, sampling_params: SamplingParams
     ) -> Tuple[List[int], Optional[ExtraProcessedInputs]]:
         """
--- a/tensorrt_llm/_torch/models/modeling_vila.py
+++ b/tensorrt_llm/_torch/models/modeling_vila.py
@@ -1103,7 +1103,7 @@ class VilaInputProcessor(BaseMultimodalInputProcessor,
         return fused_input_ids, mm_features
 
     @torch.inference_mode()  # critical to minimize memory footprint
-    def __call__(
+    def call_with_txt_prompt(
         self, inputs: TextPrompt, sampling_params: SamplingParams
     ) -> Tuple[List[int], Optional[ExtraProcessedInputs]]:
         """
--- a/tensorrt_llm/inputs/registry.py
+++ b/tensorrt_llm/inputs/registry.py
@@ -171,14 +171,64 @@ class BaseMultimodalInputProcessor(ABC):
         multimodal_embedding: Dict[str, List[torch.Tensor]],
         sampling_params: SamplingParams,
     ) -> Tuple[List[int], Optional[ExtraProcessedInputs]]:
-        """
-        Handle externally provided multimodal input embeddings.
+        """Handle externally provided multimodal input embeddings.
 
-        While inputs["multi_modal_data"] is handled by __call__, this method is intended to process
-        inputs["multi_modal_embeddings"].
+        While inputs["multi_modal_data"] is handled by __call__, this method is
+        intended to process inputs["multi_modal_embeddings"].
+
+        Auto-detokenizes `prompt_token_ids` --> `prompt` for non-fast-path VLMs
+        before delegating to `_attach_multimodal_embeddings_impl`. Subclasses
+        override `_attach_multimodal_embeddings_impl` (not this wrapper).
+        """
+        if (inputs.get("prompt_token_ids") is not None
+                and inputs.get("prompt") is None
+                and not self.supports_token_id_mm_expansion):
+            inputs = self._detokenize_to_text_prompt(inputs, sampling_params)
+        return self._attach_multimodal_embeddings_impl(inputs,
+                                                       multimodal_embedding,
+                                                       sampling_params)
+
+    def _attach_multimodal_embeddings_impl(
+        self,
+        inputs: TextPrompt,
+        multimodal_embedding: Dict[str, List[torch.Tensor]],
+        sampling_params: SamplingParams,
+    ) -> Tuple[List[int], Optional[ExtraProcessedInputs]]:
+        """Subclass hook for `attach_multimodal_embeddings`.
+
+        Subclasses that support `multi_modal_embeddings` override this method
+        rather than the wrapper `attach_multimodal_embeddings` itself.
+        Default raises NotImplementedError.
         """
         raise NotImplementedError(
             "Input processor does not support multimodal embedding input")
+
+    def _detokenize_to_text_prompt(
+        self,
+        inputs: TextPrompt,
+        sampling_params: SamplingParams,
+    ) -> TextPrompt:
+        """Decode `prompt_token_ids` to text and rebuild `inputs` as a TextPrompt.
+
+        Used by `__call__` and `attach_multimodal_embeddings` to normalize
+        tokenized inputs for non-fast-path VLMs before delegating to the
+        subclass-specific text-prompt path.
+
+        Also flips `sampling_params.add_special_tokens` to False so the
+        subclass's re-tokenization doesn't double-add BOS/EOS on top of the
+        ones already encoded in `prompt_token_ids`.
+        """
+        prompt = self.tokenizer.decode(inputs["prompt_token_ids"])
+        if sampling_params is not None and sampling_params.add_special_tokens:
+            logger.debug(
+                "Setting add_special_tokens to False because prompt_token_ids "
+                "were provided to generate. VLMs will re-encode the prompt.")
+            sampling_params.add_special_tokens = False
+        return TextPrompt(
+            prompt=prompt,
+            multi_modal_data=inputs.get("multi_modal_data"),
+            mm_processor_kwargs=inputs.get("mm_processor_kwargs") or {},
+        )
 
     @property
     @abstractmethod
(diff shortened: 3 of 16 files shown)


moraxu:
@coderabbitai full review

tensorrt_llm/inputs/registry.py
```diff
@@ -975,30 +992,19 @@ def multimodal_hashing_process(
...
+            mm_data,
+            input_processor,
+            multimodal_data=(extra_processed_inputs
+                             or {}).get("multimodal_data"),
+        )
+        if not num_mm_tokens_by_key:
+            return [], None
+        num_mm_tokens = next(iter(num_mm_tokens_by_key.values()))
         if len(num_mm_tokens) <= 0:
             return [], None
```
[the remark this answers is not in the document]
moraxu: Committed the fix, I think it would only fire if a user submitted e.g. `multi_modal_data={"image": []}`

tensorrt_llm/_torch/models/modeling_gemma3vl.py
2ez4bz: Nit: might as well be `text`?

tensorrt_llm/_torch/models/modeling_vila.py
2ez4bz: How do you feel about keeping these as `__call__` (since the signature is the same), and instead of `abstractmethod`s in the base class, have them raise `NotImplementedError`?

That would minimize the amount of modeling files touched.

I don't feel strongly about it, so up you.

tensorrt_llm/inputs/registry.py
2ez4bz: Nit: I think this should not be mentioned here, but in `llm.py` at the call site(s).

2ez4bz: Nit: the technical details of `call_with_token_ids` should sit in its own docstring, IMO.

2ez4bz: Do you think it's a good idea to have `get_text_with_mm_placeholders` and `expand_prompt_token_ids_for_mm` declared in this class?

Main pro is that it documents the expected signature clearly - right now, it's completely undocumented, and one has to go inspect the private helpers `_process_multimodal_with_dummy_placeholders` + the innards of this method to find out.

Main con is that it kind of feels like we're starting to bloat the base class.

2ez4bz: This is still weird UX, IMO, since children still need to override 2 methods `get_text_with_mm_placeholders` and `expand_prompt_token_ids_for_mm` to get the final `extra_processed_inputs.multimodal_data` dict populated in its final form.

Could they somehow be combined?

2ez4bz, line 962: @venkywonka is this a restriction / assumption you're working on lifting separately?

tensorrt_llm/_torch/models/modeling_vila.py
moraxu: I don't fully get this part: "have them raise `NotImplementedError?`" - `call_with_txt_prompt` logic (or if we put it under `__call__`, or whatever you want to call it) has to be defined for every input processor anyway? Only the `call_with_token_ids` optimization is not yet implemented for every VLM?

tensorrt_llm/inputs/registry.py
moraxu: Well, I think they do belong in this class since they can be thought of as private helper methods that in the end fulfill the duties of the input **preprocessor**.

I even considered moving `_process_multimodal_with_dummy_placeholders` to this class as well, but wasn't sure if it's not too much.

moraxu: So, I think the split is mostly reflecting the HF/vLLM conventions rather than being something arbitrary in this PR.

1. vLLM has the same conceptual separation for tokenized prompt + MM data. Their docs describe HF processors as doing: tokenize text, process MM inputs, then perform prompt updates. For tokenized inputs, the text has already bypassed the HF processor, so vLLM uses dummy text to let HF process the MM payload, then applies the prompt update step separately to keep token IDs and processed MM data consistent:
https://docs.vllm.ai/en/latest/design/mm_processing/#tokenized-prompt-inputs

That is basically what `get_text_with_mm_placeholders` and `expand_prompt_token_ids_for_mm` represent here:
- `get_text_with_mm_placeholders`: let us run the HF/MM preprocessing path without the real prompt.
- `expand_prompt_token_ids_for_mm`: apply the token-expansion logic to the already-tokenized real prompt.

If `transformers` eventually has native support for some of this preprocessing logic, as we discussed, I think sticking to this convention makes us more compatible with HF/vLLM API conventions.

2. I think combining the two into one method would be possible, but it would make each model reimplement more boilerplate code **in a single high level method**: 1) build dummy text, 2) run MM preprocessing, 3) infer feature-token counts, 4) expand the real token IDs. With steps 1) and 4) being only applicable if that method happens to be called from within `call_with_token_ids`.

Currently, the only tiny "downside" of the present approach is that both paths - `call_with_txt_prompt` and `call_with_token_ids` - call the same lightweight helper methods for shape-related computations for more complex modalities only. For example, in Nemotron's video case, the current split lets both the text-prompt path and the tokenized fast path reuse the same token-accounting helpers. E.g., `_compute_token_numbers_per_video` is used by the normal MM preprocessing path when building video prompt tokens, and also by the token-ID expansion path, but those computations are light, it's not like we re-computing MM tensors as discussed.

So given 1) and 2), I'd be in favor of keeping everything as is.

tensorrt_llm/_torch/models/modeling_vila.py
2ez4bz: My point was that the amount of changes would be smaller if you kept each existing model's `__call__` method instead of renaming it to `call_with_txt_prompts`. But this is fine as is.

tensorrt_llm/inputs/registry.py
2ez4bz: No, I meant, declared explicitly as part of the class itself, instead of `hasattr` / `getattr`'ed.

2ez4bz: > Currently, the only tiny "downside" of the present approach is that both paths 

I disagree - that part is fine, actually, IMO. The main downside is how unnatural the `mm_data_updates` part still feels. At least now it's contained to `expand_prompt_token_ids_for_mm` + this caller method, but it's still not great. It even "violates" the vLLM principle you mentioned:

> vLLM has the same conceptual separation for tokenized prompt + MM data

since `expand_prompt_token_ids_for_mm` is acutely aware of both.

tensorrt_llm/_torch/models/modeling_vila.py
moraxu: Discussed offline, will keep as is

tensorrt_llm/inputs/registry.py
moraxu: Discussed offline, @2ez4bz to circle back

moraxu: If not possible to move to base class and raise notImplemented by default, add their signature etc. to call_with_token_ids's docstring

2ez4bz, line 147: Could add `supports_token_id_mm_expansion: ClassVar[bool] = False` (name TBD).

moraxu: I went with the latter option, i.e. elaborated the existing docstring.

moraxu: Will keep it as is, as discussed offline. Having the `call_with_token_ids` defined here, in the base class, without the `if mm_data_updates` part and have the overridden `call_with_token_ids` in child classes do that update would require introducing some intermediate signature changes.

tensorrt_llm/llmapi/llm.py
moraxu, line 550: @2ez4bz , I made the change we discussed in https://github.com/NVIDIA/TensorRT-LLM/pull/14370/changes/105c0a44344ecdad4792678fc058fadcc5e79626.


If we fold the detokenize-fallback into `BaseMultimodalInputProcessor.__call__`, out of `llm.py` where it currently lives, we only catch the `multi_modal_data` path. Requests that arrive with `multi_modal_embeddings` + `prompt_token_ids` would skip the new fallback entirely because they're routed to `attach_multimodal_embeddings`. To fully replicate the current `llm.py` behavior, we'd need to introduce a slight refactor of the `BaseMultimodalInputProcessor`s `attach_multimodal_embeddings`.

Two `attach_multimodal_embeddings` overrides exist today: `LlavaNext` and `Llama4`. With this new change:

| Model | `supports_token_id_mm_expansion` | New detokenize fires in `attach_multimodal_embeddings`? |
|---|---|---|
| LlavaNext | `True` | **No** - gate is `not supports_token_id_mm_expansion`, so it stays False. Existing behavior. |
| Llama4 | `False` (default) | **Yes** - Llama4's `_attach_multimodal_embeddings_impl` now receives a detokenized `prompt` instead of `prompt_token_ids`. |

So the change would only require testing this path: Llama4 + `prompt_token_ids` + `multi_modal_embeddings`.

Let me know what you think about this refactor first, though?

2ez4bz (approved):

venkywonka (approved):

PR #14373, opened 2026-05-20
[https://nvbugs/6035425][fix] Fix KV cache host splitting logic

mikeiovine:
## Description

The fix in https://github.com/NVIDIA/TensorRT-LLM/pull/13130 only applied to the V2 KV cache manager.

Make it so that the KV cache host budget is always split proportionally across the draft and target. I did not mess with the GPU side splitting at all. 

Note that V1 (non-VSWA) does not use the same budget splitting logic for the GPU side memory. It instead does it "implicitly" using a single `max_num_tokens` value. I did not want to mess with it since this is a sensitive part of the system, so the behavior there is unchanged (a simpler way to fix this is to simply remove the VSWA restriction on the call to `_split_kv_cache_budget_for_draft`, but the changes existing behavior).

## Test Coverage

New tests added and manually verified that host budget is not doubled.

## PR Checklist

## GitHub Bot Help

## Summary by CodeRabbit

* **Refactor**
  * Optimized KV cache memory management for speculative decoding. Improved internal budget allocation enables more efficient GPU and host memory distribution, providing better resource utilization and enhanced support for speculative decoding operations across various hardware configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14373?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/_util.py
+++ b/tensorrt_llm/_torch/pyexecutor/_util.py
@@ -925,21 +925,9 @@ class KvCacheCreator:
             is_disagg=self._is_disagg,
         )
 
-    def _split_kv_cache_budget_for_draft(self) -> Optional[KvCacheConfig]:
-        """Split KV cache budgets between target and draft KV caches.
-
-        When using KVCacheManagerV2 with a separate draft KV cache,
-        max_gpu_total_bytes and host_cache_size each represent the total
-        budget for both target and draft combined.  This method splits both
-        budgets proportionally based on their per-token KV cache sizes.
-
-        Returns a cloned KvCacheConfig for the draft, or None if no split is
-        needed.  Also modifies self._kv_cache_config in-place for the target.
-        """
-        total_budget = self._kv_cache_config.max_gpu_total_bytes
-        if total_budget is None or total_budget <= 0:
-            return None
-
+    def _get_target_and_draft_cache_costs(
+        self, ) -> Optional[tuple[CacheCost, CacheCost]]:
+        """Per-manager KV cache costs for target and draft layers."""
         total_kv = self._get_kv_size_per_token()
         target_kv = self._per_manager_cache_cost(
             self._kv_cache_manager_cls, self._model_engine.model.model_config)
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -931,62 +924,129 @@ def _split_kv_cache_budget_for_draft(self) -> Optional[KvCacheConfig]:
         draft_slope_share = int(slope_budget * draft_kv.slope / slope_total)
         draft_budget = draft_kv.intercept + draft_slope_share
         target_budget = total_budget - draft_budget
+        return target_budget, draft_budget
+
+    def _draft_byte_fraction(self, target_kv: CacheCost,
+                             draft_kv: CacheCost) -> Optional[float]:
+        """Fraction of a byte budget allocated to the draft manager."""
+        total_gpu_budget = self._kv_cache_config.max_gpu_total_bytes or 0
```
ziyixiong-nv: This might be incorrect. The value has been updated to `target_budget` in the function `_split_kv_cache_gpu_budget_for_draft` since `_draft_byte_fraction` is called after `_split_kv_cache_gpu_budget_for_draft` in `build_managers`.

mikeiovine: Nice catch. Let me see if I can simplify this a bit. The logic has gotten a bit too complicated

ziyixiong-nv (approved):

xinhe-nv:
@coderabbitai review the pr as a QA engineer, check the coverage.

mikeiovine:
I attempted to address the first comment from CodeRabbit as follows:

1. If the combined fixed cost for the target and draft exceeds the total budget on device, you'd have no space left for any cache at runtime. The run would OOM. If this somehow happens, we should hard fail during the split.
2. If this happens on the host, you can just set the host allocation to zero since host offloading is optional. Added some defensive code paths to do this.
3. However, this **cannot happen in practice**. Mamba states only reside on the GPU. Therefore, we can zero all intercepts when doing the host split.

All other suggested tests from CodeRabbit do not look helpful to me.

Tabrizian (approved):

PR #14384, opened 2026-05-21
[None][perf] Eliminate torch.where host sync in multimodal fuse_input_embeds path

yechank-nvidia:
@coderabbitai summary

  ## Description

  VLM `fuse_input_embeds` falls back to `filter_mm_token_from_input_ids` when the MM/text token indices are missing. On GPU `input_ids` that filter's `torch.where` forces a **host-device synchronization on every VLM forward step**. This PR removes that sync and the legacy out-of-vocab (OOV) placeholder workaround that made it necessary.

  **What changed**

  1. **Sync-free index plumbing (`model_engine`)**
     - `_prepare_multimodal_indices` computes `text_token_indices` / `mm_token_indices`
       on a **CPU** `input_ids` copy.
     - New `_ship_multimodal_indices` helper (shared by `_prepare_tp_inputs` and
       `_prepare_tp_inputs_no_cache`) pins and async-H2D-copies them into the inputs
       dict. The KV-cache path reuses the CPU `text_token_indices` and extends it with
       an `arange` for the post-context range instead of allocating `torch.ones(N, bool)`
       + rebuilding via `torch.where`.
     - The `mm_encoder_only` flag guard is replaced with a structural
       `hasattr(model.config, 'vocab_size')` check.

  2. **In-vocab `mm_token_ids` contract (drop OOV remap)**
     - Removes the `token_ids[mask] = vocab_size + 1` remap in **Llama4, LLaVA-Next,
       Qwen2-VL, Qwen2.5-VL, Qwen3-VL, VILA** input processors.
     - Each model exposes `mm_token_ids` so the executor's `_prepare_multimodal_indices`
       and `maybe_compute_mm_embed_cumsum` locate MM positions via
       `torch.isin(input_ids, mm_token_ids)` instead of the OOV `>= vocab_size` predicate.
     - **Kimi K25 / Gemma4-MM** cache the tensor (`_mm_token_ids`) behind an `@property`
       to avoid per-step `torch.tensor` / `torch.cat` allocations.

  3. **`fuse_input_embeds` fast path** — when `mm_token_ids` is provided (in-vocab
     caller), embed the full `input_ids` once and overwrite the MM rows instead of the
     gather + `torch.empty` + text-scatter path.

  4. VLM forwards now extract the indices from `kwargs` and pass them **explicitly** to
     `fuse_input_embeds` rather than relying on a `**kwargs` splat.

  **Not converted:** Hyperclovax stays on the OOV path (no in-vocab content placeholder
  in its config); the OOV branch in `filter_mm_token_from_input_ids` /
  `fuse_input_embeds` is retained as the Hyperclovax-only fallback.

  ### Performance — per VLM forward step

  The change trades a small **CPU** predicate cost for the removal of GPU work and, more
  importantly, a host-device sync on the critical path:

  | Path | Before | After |
  |------|--------|-------|
  | MM/text index compute | GPU `torch.where` on `input_ids` → **host sync** | CPU predicate + async H2D (no sync) |
  | KV-cache text complement | `torch.ones(N, bool)` + `torch.where` | reuse CPU indices + `arange`/`cat` |
  | `fuse_input_embeds` | gather + `torch.empty` alloc + text-scatter | single embed + MM-row overwrite |

  Net per VLM forward step: **−1 host-device sync, −2 GPU kernels, −1 device alloc**, at
  the cost of a CPU `torch.isin` predicate over `input_ids` whose result is shipped async.

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_gemma4mm.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma4mm.py
@@ -1034,7 +1034,8 @@ class Gemma4ForConditionalGeneration(PreTrainedModel):
             input_ids=input_ids,
             mm_embeds=mm_embeds,
             mm_token_ids=fuse_token_ids,
-            **kwargs,
+            mm_token_indices=kwargs.get("mm_token_indices"),
+            text_token_indices=kwargs.get("text_token_indices"),
         )
         logits = self.llm.forward(
             attn_metadata=attn_metadata,
--- a/tensorrt_llm/_torch/models/modeling_gemma4mm.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma4mm.py
@@ -638,6 +638,8 @@ class Gemma4ForConditionalGeneration(PreTrainedModel):
         _mm_ids = [self.image_token_ids]
         if self.audio_token_ids is not None:
             _mm_ids.append(self.audio_token_ids)
+        if self.video_token_ids is not None:
+            _mm_ids.append(self.video_token_ids)
         self._mm_token_ids = torch.cat(_mm_ids) if len(_mm_ids) > 1 else self.image_token_ids
 
         model_config_cp = copy.deepcopy(model_config)
--- a/tensorrt_llm/_torch/models/modeling_gemma4mm.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma4mm.py
@@ -635,6 +635,11 @@ class Gemma4ForConditionalGeneration(PreTrainedModel):
             else None
         )
 
+        _mm_ids = [self.image_token_ids]
+        if self.audio_token_ids is not None:
+            _mm_ids.append(self.audio_token_ids)
+        self._mm_token_ids = torch.cat(_mm_ids) if len(_mm_ids) > 1 else self.image_token_ids
+
         model_config_cp = copy.deepcopy(model_config)
         self.model_config = model_config_cp
 
@@ -1050,8 +1055,5 @@ class Gemma4ForConditionalGeneration(PreTrainedModel):
         return logits
 
     @property
-    def mm_token_ids(self):
-        ids = [self.image_token_ids]
-        if self.audio_token_ids is not None:
-            ids.append(self.audio_token_ids)
-        return torch.cat(ids) if len(ids) > 1 else self.image_token_ids
+    def mm_token_ids(self) -> torch.Tensor:
+        return self._mm_token_ids
--- a/tensorrt_llm/_torch/models/modeling_gemma4mm.py
+++ b/tensorrt_llm/_torch/models/modeling_gemma4mm.py
@@ -635,12 +635,12 @@ class Gemma4ForConditionalGeneration(PreTrainedModel):
             else None
         )
 
-        _mm_ids = [self.image_token_ids]
-        if self.audio_token_ids is not None:
-            _mm_ids.append(self.audio_token_ids)
-        if self.video_token_ids is not None:
-            _mm_ids.append(self.video_token_ids)
-        self._mm_token_ids = torch.cat(_mm_ids) if len(_mm_ids) > 1 else self.image_token_ids
+        _mm_ids = [config.image_token_id]
+        if getattr(config, "audio_token_id", None) is not None:
+            _mm_ids.append(config.audio_token_id)
+        if getattr(config, "video_token_id", None) is not None:
+            _mm_ids.append(config.video_token_id)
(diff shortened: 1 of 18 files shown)


tensorrt_llm/_torch/models/modeling_llava_next.py
2ez4bz: Nit: I think this type of comment can be left out of the model docstrings.

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
2ez4bz, line 356: Nit:

```bash
sed -i 's/`/``/g'
```

on the changed lines.

2ez4bz (approved):

tensorrt_llm/_torch/models/modeling_qwen3vl.py
```diff
@@ -1013,6 +1039,23 @@ def __init__(
...
+            )
+            if tid is not None
+        ]
+        self._mm_token_ids = torch.tensor(_mm_ids, dtype=torch.int32)
+
+    @property
+    def mm_token_ids(self) -> torch.Tensor:
+        return self._mm_token_ids
+
         self.post_config()
```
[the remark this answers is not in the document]
yechank-nvidia: Addressed

tensorrt_llm/_torch/models/modeling_gemma4mm.py
```diff
@@ -613,6 +613,11 @@ def __init__(self, model_config: ModelConfig[Gemma4Config]):
             else None
         )
 
+        _mm_ids = [self.image_token_ids]
+        if self.audio_token_ids is not None:
+            _mm_ids.append(self.audio_token_ids)
+        self._mm_token_ids = torch.cat(_mm_ids) if len(_mm_ids) > 1 else self.image_token_ids
+
```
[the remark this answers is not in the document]
yechank-nvidia: Addressed

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -2461,13 +2490,19 @@ def _prepare_tp_inputs(
...
+            # input_ids holds only context tokens here; extend/draft tokens are
+            # appended below and are by construction text, so we reuse the
+            # CPU-side text_token_indices and just extend it with the
+            # post-context arange instead of recomputing via a bool mask +
+            # torch.where over the full range.
+            text_token_indices_ctx, mm_token_indices = \
+                self._prepare_multimodal_indices(input_ids)
         else:
+            text_token_indices_ctx = None
             mm_token_indices = None
```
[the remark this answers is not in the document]
yechank-nvidia: Addressed

tensorrt_llm/_torch/models/modeling_vila.py
```diff
@@ -1069,15 +1079,19 @@ def _postprocess(self, input_ids, mm_features):
...
         )))  # len(input_ids_splits) = num_segments after mm tokens are isolated
+        # Use the first in-vocab media token id repeated mm_total_length times
+        # instead of unique arange(vocab_size, vocab_size + L) OOV IDs — only
+        # the predicate that distinguishes mm vs text matters downstream
+        # (fuse_input_embeds uses ``torch.isin(input_ids, mm_token_ids)``).
         mm_ids_splits = list(
-            torch.arange(vocab_size,
-                         vocab_size + mm_total_length,
-                         device=input_ids.device).split(mm_lengths_per_split)
+            torch.full((mm_total_length, ),
```
QiJune, line 1094: Could you help double-check the VILA path here?

My reading is that this changes the expanded MM ids from OOV ids to the in-vocab media token. That seems consistent with the new `fuse_input_embeds(..., mm_token_ids=...)` path, but I noticed `_validate_vila_mm_alignment()` still appears to classify MM positions via the OOV fallback (`input_ids >= vocab_size`). If so, it may find zero MM positions after this change and raise a token/embed count mismatch before reaching `fuse_input_embeds`.

QiJune (commented):
LGTM for the runtime changes

tensorrt_llm/_torch/models/modeling_vila.py
yechank-nvidia: Thx for the catch! Fixed it now.

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
yechank-nvidia: Applied.

amukkara (approved):
Approving phi modelling changes

QiJune (approved):

2ez4bz:
first try

PR #14395, opened 2026-05-21
[None][feat] KV reuse probe

lancelly:
## Summary

Stacked on top of #14353. Adds the KV cache reuse probe API from #14333 by cherry-picking `f9968c6bd3` onto the `feat/dsv4-refactor-and-probe` stack.

- Picks up the full content of #14353 (ReuseScope refactor + DSv4 follow-ups).
- Cherry-picks `Add KV cache reuse probe` (#14333, `f9968c6bd3`) — adds `KVCacheManager.probe_reuse()` plus the `BlockRadixTree.match` → `ReuseMatch` refactor that splits the prune logic out of `_KVCache._setup_for_reuse`.

## Conflict resolution notes (`tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py`)

#14333 was based on `main`, which doesn't have DSv4's reuse-stats tracking. The cherry-pick conflicted in two regions; resolved as follows:

1. `_KVCache.__init__` — keep `_pending_stats = _PendingStats()` and the post-setup `_refresh_generation_alloc_ready()` call, while adopting the new `reuse_match: ReuseMatch | None` parameter from #14333.
2. `_setup_for_reuse` — adopt #14333's lean shape (use `match.blocks` / `match.num_tokens` from a pre-pruned `ReuseMatch`), then re-insert the DSv4 stats tracking inside the per-life-cycle loop (`full_reused_blocks`, `partial_reused_blocks`, `_pending_stats.record_reuse`, `mark_stats_dirty`) using `num_tokens` instead of the removed `get_num_matched_tokens(matched)` helper. Indexing was migrated from `matched[ordinal][0]` to `matched[ordinal]` since `matched` is now `list[Block]`.
3. `_KVCache.__init__` previously defaulted `expected_prompt_length` from `len(input_tokens)`. After #14333 removes the `input_tokens` parameter, that fallback was moved up one level into `KVCacheManager.create_kv_cache` so the public-API behavior is preserved.

### DIFF
--- a/tensorrt_llm/runtime/kv_cache_manager_v2/__init__.pyi
+++ b/tensorrt_llm/runtime/kv_cache_manager_v2/__init__.pyi
@@ -284,7 +284,7 @@ class _KVCache:
         self,
         manager: "KVCacheManager",
         reuse_scope: ReuseScope,
-        input_tokens: Sequence[TokenIdExt] | None,
+        reuse_match: Any | None,
         id: Any,
         custom_priority_callback: Callable[[int, Any], Priority],
         cache_salt_id: int | None = None,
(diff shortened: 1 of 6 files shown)


peihu-nv (approved):
LGTM. 

One follow up needed for full E2E is wiring the probe reuse to the existing `probe_prefix_match_length(...)` I think.

PR #14398, opened 2026-05-21
[TRTLLM-12842][feat] Maximal LLMAPI capture in usage telemetry

venkywonka:
## Motivation

Usage telemetry had coarse feature flags, not effective LLM API runtime-shape config. Needed: know how deployed configs are shaped without collecting private or user-controlled data.

Hard rule: no PII, no prompts, no outputs, no model/tokenizer paths, no env vars, no raw YAML, no arbitrary fields, no arbitrary strings.

## Design

Full list and related TRTLLM documentation preview: https://venkywonka.github.io/TensorRT-LLM/developer-guide/telemetry.html 

- Add field-local `Field(..., telemetry=...)` opt-in metadata for approved LLM args fields.
- Walk only declared Pydantic fields from validated effective args. Unmarked fields stay invisible.
- Emit deterministic `llmApiConfigJson` for sanitized config values.
- Emit `llmApiConfigMetaJson` with capture version, policy version, args class, digests, counts, source, and unsafe-exclusion status.
- Fail closed in sanitizer: reject paths, raw strings without allowlists, dict/object payloads, `Any`, callables, unsafe unions, and heterogeneous tuples.
- Keep `featuresJson` for legacy summary consumers. Add TODO to dedupe with `llmApiConfigJson` later.
- Generate telemetry docs during Sphinx build from `llm_args.py` annotations so exhaustive captured-key docs do not drift manually.
- Preserve existing usage opt-out behavior and precompiled editable validation path for Python/schema-only changes.

## Testing

- `python3 -m json.tool tensorrt_llm/usage/schemas/trtllm_usage_event_schema.json`
- `python3 -m py_compile` on telemetry usage modules and docs generator
- `ruff check` / `ruff format --check` on touched Python files
- `git diff --check`
- Usage unit coverage: `199 passed, 5 skipped`
- Forced real-GPU usage capture smoke: `2 passed`
- Generated-docs unit coverage: `2 passed`
- Full Sphinx HTML build completed; telemetry page renders README link, schema link, privacy invariant, and exhaustive LLM API config field tables.
- Real GPU precomputed-wheel validation path: `54 passed`
- Real-CUDA smoke payloads passed for single-GPU, TP=2 compile, and scheduler/KV-cache privacy shapes.
- TRT-LLM MCP runtime contract check: `status=ready`

## Summary by CodeRabbit

* **Documentation**
  * Enhanced README telemetry section with explicit enumeration of excluded data categories (outputs, model paths, PII, raw configurations, persistent identifiers).
  * Expanded developer guide with detailed telemetry capture reference and configuration field allowlists.

* **New Features**
  * Introduced sanitized, opt-in LLM API configuration telemetry with deterministic capture and validation; schema version bumped to 0.2.
  * Added disaggregated launcher telemetry environment propagation for deployment tracking.

* **Tests**
  * Comprehensive test coverage for configuration telemetry capture, sanitization, and end-to-end validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,9 +298,10 @@ Deprecation is used to inform developers that some APIs and tools are no longer 
 TensorRT-LLM collects anonymous telemetry data by default. This data is used
 in aggregate to understand usage patterns and prioritize engineering efforts.
 **This data cannot be traced back to any individual user.** No prompts,
-user-identifying information, or persistent identifiers are collected. Any
-deployment identifiers are ephemeral, randomly generated per deployment, and
-not linked to users. The data we collect includes:
+outputs, model weights, model paths, tokenizer paths, user-identifying
+information, raw free-form configuration strings, or persistent identifiers are
+collected. Any deployment identifiers are ephemeral, randomly generated per
+deployment, and not linked to users. The data we collect includes:
 
 - Ingress point (e.g., LLM API, CLI, serve command)
 - Deployment duration (via periodic heartbeats)
@@ -309,8 +310,10 @@ not linked to users. The data we collect includes:
 - Parallelism configuration (TP/PP/CP/MoE-EP/MoE-TP sizes), quantization algorithm, dtype, KV cache dtype
 - System information (OS platform, Python version, CPU architecture, CPU count)
 - TRT-LLM version and backend
-- Feature flags (LoRA, speculative decoding, prefix caching, CUDA graphs, chunked context, data parallelism)
+- Feature summary flags (LoRA, speculative decoding, prefix caching, CUDA graphs, chunked context, data parallelism)
 - Disaggregated serving metadata (role and deployment ID)
+- Selected LLM API configuration values: parallelism, dtype, KV cache, scheduler, CUDA graph, and compile settings
+- Capture diagnostics for that payload: a schema checksum (for provenance), the count of captured fields, and whether any free-form value was skipped
 
 Telemetry is automatically disabled in CI and test environments.
 
--- a/docs/source/_ext/llmapi_config_telemetry.py
+++ b/docs/source/_ext/llmapi_config_telemetry.py
@@ -1,0 +1,102 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+_GOLDEN_REL = "tensorrt_llm/usage/llm_args_golden_manifest.json"
+
+_REFERENCE_PREAMBLE = """\
+# Telemetry
+
+This page documents TensorRT-LLM usage telemetry. It is generated during the
+Sphinx docs build by rendering the committed telemetry manifest
+(`tensorrt_llm/usage/llm_args_golden_manifest.json`).
+
+Start with the
+[Telemetry Data Collection section in the root README](source:README.md#telemetry-data-collection)
+for the user-facing collection and opt-out overview, and the
+[telemetry schema reference](source:tensorrt_llm/usage/schemas/README.md)
+for the wire schema.
+
+**No PII or free-form fields are captured.** LLM API configuration capture is
+*type-driven*: fields whose type is categorical (`Literal`/`Enum`/`bool`) or
+numeric (`int`/`float`), plus safe collections of those, are captured
+automatically. Free-form `str`/`Any`/`Path`/`dict`/`Callable` are never captured
+unless a field carries an explicit allowlist (`TelemetryField.categorical(...)`),
+and any field may opt out with `telemetry=False`. Every captured field is listed
+below; the runtime can capture nothing absent from this list.
+
+## LLM API Configuration Fields
+
+A field can still be absent from a specific payload when its parent config is
+unset or when the safety sanitizer rejects the runtime value.
+"""
+
+
+def _escape(text: str) -> str:
+    return text.replace("|", "\\|").replace("\n", " ")
+
+
+def _format_values(values: list[str]) -> str:
+    return ", ".join(f"`{_escape(v)}`" for v in values) if values else ""
+
+
+def _table(rows: list[dict]) -> str:
+    lines = [
+        "| Captured key | Annotation | Kind | Converter | Allowed values |",
+        "|--------------|------------|------|-----------|----------------|",
+    ]
+    for row in rows:
+        lines.append(
+            f"| `{_escape(row['path'])}` | `{_escape(row['annotation'])}` | "
+            f"`{_escape(row['kind'])}` | {_escape(row['converter'])} | "
+            f"{_format_values(row['allowed_values'])} |"
+        )
+    return "\n".join(lines)
+
+
+def generate_telemetry_reference(repo_root: Path | str, output_path: Path | str) -> None:
+    repo_root = Path(repo_root)
+    golden = json.loads((repo_root / _GOLDEN_REL).read_text())
+    content = [_REFERENCE_PREAMBLE]
+    for args_class in ("TorchLlmArgs", "TrtLlmArgs"):
+        rows = golden.get(args_class, [])
+        content.extend(
+            [
+                f"### `{args_class}`",
+                "",
+                f"{len(rows)} captured fields.",
+                "",
+                _table(rows),
+                "",
+            ]
+        )
+    output = Path(output_path)
+    output.parent.mkdir(parents=True, exist_ok=True)
+    output.write_text("\n".join(content))
+
+
+def _on_builder_inited(app) -> None:
+    docs_source = Path(app.confdir)
+    repo_root = docs_source.parents[1]
+    generate_telemetry_reference(repo_root, docs_source / "developer-guide/telemetry.md")
+
+
+def setup(app) -> dict[str, object]:
+    app.connect("builder-inited", _on_builder_inited)
+    return {"version": "0.2", "parallel_read_safe": True, "parallel_write_safe": True}
--- a/docs/source/developer-guide/telemetry.md
+++ b/docs/source/developer-guide/telemetry.md
@@ -1,0 +1,532 @@
+# Telemetry
+
+This page documents TensorRT-LLM usage telemetry. It is generated during the
+Sphinx docs build by rendering the committed telemetry manifest
+(`tensorrt_llm/usage/llm_args_golden_manifest.json`).
+
+Start with the
+[Telemetry Data Collection section in the root README](source:README.md#telemetry-data-collection)
+for the user-facing collection and opt-out overview, and the
+[telemetry schema reference](source:tensorrt_llm/usage/schemas/README.md)
+for the wire schema.
+
+**No PII or free-form fields are captured.** LLM API configuration capture is
+*type-driven*: fields whose type is categorical (`Literal`/`Enum`/`bool`) or
+numeric (`int`/`float`), plus safe collections of those, are captured
+automatically. Free-form `str`/`Any`/`Path`/`dict`/`Callable` are never captured
+unless a field carries an explicit allowlist (`TelemetryField.categorical(...)`),
+and any field may opt out with `telemetry=False`. Every captured field is listed
+below; the runtime can capture nothing absent from this list.
+
+## LLM API Configuration Fields
+
+A field can still be absent from a specific payload when its parent config is
+unset or when the safety sanitizer rejects the runtime value.
+
+### `TorchLlmArgs`
+
+234 captured fields.
+
+| Captured key | Annotation | Kind | Converter | Allowed values |
+|--------------|------------|------|-----------|----------------|
+| `allreduce_strategy` | `Optional[Literal['AUTO', 'NCCL', 'UB', 'MINLATENCY', 'ONESHOT', 'TWOSHOT', 'LOWPRECISION', 'MNNVL', 'NCCL_SYMMETRIC']]` | `categorical` |  | `AUTO`, `NCCL`, `UB`, `MINLATENCY`, `ONESHOT`, `TWOSHOT`, `LOWPRECISION`, `MNNVL`, `NCCL_SYMMETRIC` |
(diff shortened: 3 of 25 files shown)


tburt-nv (approved):
I'm only reviewing the setup.py changes here, which look good to me.

docs/source/developer-guide/telemetry.md
```diff
@@ -0,0 +1,128 @@
...
+Model paths, tokenizer paths, prompts, outputs, raw free-form strings,
+dict/object payloads, callables, and raw `Any` values are not captured.
+
+### `TorchLlmArgs`
+
+54 telemetry-marked fields.
+
+| Captured key | Defining field | Annotation | Kind | Converter | Allowed values |
+|--------------|----------------|------------|------|-----------|----------------|
+| `attn_backend` | `TorchLlmArgs.attn_backend` | `str` | `categorical` | allowlist | `TRTLLM`, `FLASHINFER`, `FLASH_ATTENTION`, `VANILLA` |
```
Superjomn: Is this table generated from LlmArgs? It seems the hardcoded table isn't necessary and is easy to be out-of-date, maybe let some script generate an intermediate during doc building.

docs/source/developer-guide/telemetry.md
```diff
@@ -0,0 +1,128 @@
...
+| `torch_compile_config.enable_inductor` | `TorchCompileConfig.enable_inductor` | `bool` | `value` |  |  |
+| `torch_compile_config.enable_piecewise_cuda_graph` | `TorchCompileConfig.enable_piecewise_cuda_graph` | `bool` | `value` |  |  |
+| `torch_compile_config.enable_userbuffers` | `TorchCompileConfig.enable_userbuffers` | `bool` | `value` |  |  |
+| `torch_compile_config.max_num_streams` | `TorchCompileConfig.max_num_streams` | `PositiveInt` | `value` |  |  |
+
+### `TrtLlmArgs`
+
+35 telemetry-marked fields.
+
+| Captured key | Defining field | Annotation | Kind | Converter | Allowed values |
```
Superjomn: So is this. I saw the "categorical"/"value" is registered for each Field.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -111,16 +134,19 @@ class CudaGraphConfig(StrictBaseModel):
...
         default=None,
-        description="List of batch sizes to create CUDA graphs for.")
+        description="List of batch sizes to create CUDA graphs for.",
+        telemetry={"kind": "value"})
 
     max_batch_size: NonNegativeInt = Field(
-        default=0, description="Maximum batch size for CUDA graphs.")
+        default=0,
+        description="Maximum batch size for CUDA graphs.",
+        telemetry={"kind": "value"})
```
Superjomn: Is it true that for each Field, we need to register a `telemetry.value`? If so, I saw most of them are "value", how about making "value" the default value, and only let developers register the "categorial" register explicitly.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -2594,8 +2630,11 @@ class KvCacheConfig(StrictBaseModel, PybindMirror):
     dtype: str = Field(
         default="auto",
         description=
-        "The data type to use for the KV cache. Use 'auto' to follow checkpoint metadata, otherwise force the specified dtype."
-    )
+        "The data type to use for the KV cache. Use 'auto' to follow checkpoint metadata, otherwise force the specified dtype.",
+        telemetry=TelemetryField(kind="categorical",
+                                 converter="allowlist",
+                                 allowed_values=("auto", "float16", "bfloat16",
```
Superjomn: Regarding the allowed_values, I think it should not be a part of telemetry field registration; instead, it should originate from the original Field.
e.g. We may amend the dtype's Field to have `Literal["auto", "float16", ...]` and let the Telemetry field deduce from it.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -2901,10 +2942,19 @@ class BaseLlmArgs(StrictBaseModel):
...
 
     dtype: str = Field(default="auto",
-                       description="The data type to use for the model.")
+                       description="The data type to use for the model.",
+                       telemetry={
+                           "kind":
+                           "categorical",
+                           "converter":
+                           "allowlist",
+                           "allowed_values":
```
Superjomn: So, is this `allowed_values`, I doubt if we can remove all such allowed_values and let telemetry deduce from the Typing.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -2970,7 +3025,8 @@ class BaseLlmArgs(StrictBaseModel):
     load_format: Literal['auto', 'dummy'] = Field(
         default='auto',
         description="The format to load the model.",
-        json_schema_extra={"type": "Literal['auto', 'dummy']"})
+        json_schema_extra={"type": "Literal['auto', 'dummy']"},
+        telemetry={"kind": "categorical"})
```
Superjomn: One question, when will it be `categorical`? Can we let the Field deduce automatically rather than register manually? Ideally, the Field should get all the metadata, right? We may make the telemetry deducing as a `__post_init__` like to get all the metadata.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -111,16 +134,19 @@ class CudaGraphConfig(StrictBaseModel):
...
+        description="Maximum batch size for CUDA graphs.",
+        telemetry={"kind": "value"})
 
     enable_padding: bool = Field(
         default=False,
         description=
-        "If true, batches are rounded up to the nearest cuda_graph_batch_size. This is usually a net win for performance."
-    )
+        "If true, batches are rounded up to the nearest cuda_graph_batch_size. This is usually a net win for performance.",
+        telemetry={"kind": "value"})
```
QiJune: Suggestion: the telemetry opt-in model makes sense, but the field annotations are becoming verbose. I think we can reduce the boilerplate in two small ways while keeping the same explicit privacy boundary.

1. For safe primitive fields, prefer `telemetry=True` instead of repeating `telemetry={"kind": "value"}` everywhere. `None` remains the default, so fields are still not captured unless explicitly opted in.

2. For categorical string / enum-like fields, add a helper such as

```python
  `TelemetryField.categorical(...)`, e.g.

      telemetry=TelemetryField.categorical(
          "TRTLLM",
          "FLASHINFER",
          "FLASHINFER_STAR_ATTENTION",
          "FLASH_ATTENTION",
          "VANILLA",
      )
```

This avoids repeating `kind/converter/allowed_values`, reduces typo risk, and makes it easier to review allowlist drift when valid config values change.

QiJune (approved):
LGTM

docs/source/_ext/llmapi_config_telemetry.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def generate_telemetry_reference(repo_root: Path | str, output_path: Path | str) -> None:
+    rows = collect_llm_api_config_telemetry_fields(repo_root)
+    by_args_class = {
+        args_class: [row for row in rows if row.args_class == args_class]
+        for args_class in ("TorchLlmArgs", "TrtLlmArgs")
+    }
+
+    content = [
```
Superjomn: It seems that using a multi-line string can make the text easier to read and maintain. For example:

content = """
# Telemetry

This page documents TensorRT-LLM usage telemetry. It is generated during the Sphinx docs build from source annotations in
""".strip()

Superjomn (commented):
It may be possible to make this change less invasive to the existing LlmArgs implementation and derive most of the telemetry metadata from information that is already available.

venkywonka (commented):
@coderabbitai full review

mikeiovine (approved):

venkywonka:
@coderabbit full review

docs/source/_ext/llmapi_config_telemetry.py
venkywonka: have addressed it

venkywonka: the prose block is now a single triple-quoted _REFERENCE_PREAMBLE; the per-class
tables stay generated programmatically.

tensorrt_llm/llmapi/llm_args.py
venkywonka: Done in b38a090531 — `kind` is now derived from the annotation (`Literal/Enum/allowlist` → categorical,
else value) via a shared `derive_kind()`, so it's no longer registered per-field.

tensorrt_llm/llmapi/llm_args.py
venkywonka: done

tensorrt_llm/llmapi/llm_args.py
venkywonka: agreed, for bare-str fields that don't have annotations we can work with, the allowlist is a deliberate
privacy subset, so I kept an explicit `.categorical(...)` as suggested by @QiJune and verified the
values against the consuming code. Literal/Enum fields elsewhere now auto-derive; str→Literal
tightening in actual LLMAPI code is beyond scope.

tensorrt_llm/llmapi/llm_args.py
venkywonka: Done in b38a090531 — `telemetry=True` is now the shorthand for a plain value field, and kind is
derived from the annotation rather than registered, so the {"kind": "value"} boilerplate is gone. Most
fields are now just `telemetry=True`.

tensorrt_llm/llmapi/llm_args.py
venkywonka: Both done in b38a090531: `telemetry=True` is now the shorthand for safe primitives, and there's a
`TelemetryField.categorical(*values)` classmethod for the allowlist cases (when the annotation isn't deducable by itself, eg: `attn_backend`, `dtype` fields, etc.). Tightening dtype's type to a Literal is a separate API validation-contract change I'd rather do in a focused follow-up. ideally, all categorical fields must be annotation with either enum or literal, but dtype , attn_backend have bare-str.

docs/source/developer-guide/telemetry.md
venkywonka: its auto-generated now

tensorrt_llm/llmapi/llm_args.py
Superjomn: Good progress.

Regarding telemetry=True, I noticed it has been added to many Fields. Would it make sense to make telemetry=True the default value in Field instead? That could avoid the O(n) explicit updates across all fields and leave only the relatively few cases that require telemetry=False.

venkywonka:
@coderabbitai full review

Superjomn (approved):
LGTM

README.md
```diff
@@ -309,8 +310,10 @@ not linked to users. The data we collect includes:
 - Parallelism configuration (TP/PP/CP/MoE-EP/MoE-TP sizes), quantization algorithm, dtype, KV cache dtype
 - System information (OS platform, Python version, CPU architecture, CPU count)
 - TRT-LLM version and backend
-- Feature flags (LoRA, speculative decoding, prefix caching, CUDA graphs, chunked context, data parallelism)
+- Legacy feature summary flags (LoRA, speculative decoding, prefix caching, CUDA graphs, chunked context, data parallelism)
 - Disaggregated serving metadata (role and deployment ID)
+- Sanitized, type-driven effective LLM API configuration fields (categorical and numeric inference knobs; never prompts, paths, secrets, or free-form text), such as selected parallelism, dtype, KV cache, scheduler, CUDA graph, and compile settings
+- Capture metadata for the LLM API configuration payload, including schema digests, field counts, capture success, and unsafe-exclusion signals
```
arysef: This seems a bit vague to me - what is a schema digest, capture success, and unsafe-exclusion signals?

README.md
```diff
@@ -309,8 +310,10 @@ not linked to users. The data we collect includes:
 - Parallelism configuration (TP/PP/CP/MoE-EP/MoE-TP sizes), quantization algorithm, dtype, KV cache dtype
 - System information (OS platform, Python version, CPU architecture, CPU count)
 - TRT-LLM version and backend
-- Feature flags (LoRA, speculative decoding, prefix caching, CUDA graphs, chunked context, data parallelism)
+- Legacy feature summary flags (LoRA, speculative decoding, prefix caching, CUDA graphs, chunked context, data parallelism)
```
arysef: Why are these marked legacy?

README.md
```diff
@@ -309,8 +310,10 @@ not linked to users. The data we collect includes:
...
 - Disaggregated serving metadata (role and deployment ID)
+- Sanitized, type-driven effective LLM API configuration fields (categorical and numeric inference knobs; never prompts, paths, secrets, or free-form text), such as selected parallelism, dtype, KV cache, scheduler, CUDA graph, and compile settings
```
arysef: I don't think we need to include the first sentence - we can probably just stick to specifically what we are including

README.md
venkywonka: schema digest is a checksum of the schema - its just to exactly track the provenance. the capture success is just a boolean telling if we were able to sucessfully walk through the pydantic instance and capture the full config without errors. 
unsafe exclusion signals are situations when a free-form string that isn't part of the enum/declared categorical type is passed for the knob - in which case we DON'T want to collect it even by mistake. 

i shall clear those up. thanks!

README.md
venkywonka: as now that we maximally capture the llmapi, the above feature flags are a subset/derivative of that, so its not new information.... i have added a todo for this - but maybe we'd rather keep it as a task item instead of advertising that its legacy. good point.

README.md
venkywonka: gotca

venkywonka:
nvtelemetry v0.2 schema has been promoted, so this PR is safe to merge.

tensorrt_llm/usage/usage_lib.py
```diff
@@ -590,6 +596,17 @@ def _background_reporter(
...
+            llm_api_config_json, llm_api_config_meta_json = _collect_llm_api_config_payloads(
+                llm_args
+            )
+        except Exception:
+            # Double net: reporter must not die if collector call breaks.
+            # Shared failure payload keeps metadata shape in one place.
+            args_class = type(llm_args).__name__ if llm_args is not None else ""
+            llm_api_config_json, llm_api_config_meta_json = _failure_llm_api_config_payloads(
+                args_class=args_class
+            )
```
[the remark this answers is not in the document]
venkywonka: Keeping the outer daemon catch broad intentionally: `collect_llm_api_config_payloads()` now catches only `(AttributeError, TypeError, ValueError, KeyError)` and propagates unexpected collector bugs, while `_background_reporter()` remains the final fail-silent guard so telemetry cannot crash inference. Covered by `test_collect_llm_api_config_propagates_unexpected_errors` and `test_background_reporter_keeps_initial_report_when_config_capture_fails`.

tensorrt_llm/llmapi/llm_args.py
venkywonka: Current head implements this as type-driven auto-enroll: safe categorical/numeric annotations are selected by default in `build_capture_manifest`, `telemetry=False` opts out, and `TelemetryField.categorical(...)` remains for allowlisted bare strings like dtype/attn_backend. Resolving the stale thread.

PR #14416, opened 2026-05-21
[None][fix] Force slow tokenizer for Kimi K2.5 (TikTokenTokenizer)

qiaoxj07:
## Summary

- Force `use_fast=False` when loading the Kimi K2.5 tokenizer so the model's own slow `TikTokenTokenizer` (`tokenization_kimi.py`) is used instead of transformers 5.x's auto-converted Fast tokenizer.
- In `tensorrt_llm/bench/utils/data.py:initialize_tokenizer`, gate `use_fast=False` on `tokenizer_class == "TikTokenTokenizer"` via a new helper that reads `tokenizer_config.json` without instantiating, so behavior is unchanged for every other model.
- In `tensorrt_llm/_torch/models/modeling_kimi_k25.py:KimiK25InputProcessor.__init__`, hardcode `use_fast=False` for `AutoTokenizer` (the file is K2.5-specific) and pin `self._processor.tokenizer = self._tokenizer` so any text path through `self._processor(text=...)` uses the same slow tokenizer. `AutoProcessor`'s `use_fast` is left as-is because it only controls image processor selection.

## Why

Kimi K2.5 ships only a slow `TikTokenTokenizer(PreTrainedTokenizer)` with a custom CJK-aware `pat_str` regex; there is no Fast equivalent and no `tokenizer.json` upstream. transformers >= 5 newly tries to auto-convert it to a Fast tokenizer via `TokenizersBackend.convert_to_native_format`, which:

1. Sees `vocab_file = tiktoken.model` exists.
2. Tries to extract it as SentencePiece -> fails ("Wrong wire type in tag").
3. Logs `Could not extract SentencePiece model … Falling back to TikToken extractor`.
4. Builds a Fast tokenizer from the BPE merges with a generic pre-tokenization regex (not Kimi's `pat_str`).

The resulting Fast tokenizer mis-segments multilingual / CJK content, producing roughly +20% more tokens for the same input string compared to the slow path. In one benchmark sweep (kimi-k2-sol-8192-1024-ratio08), this caused dataset prompts that were generated at <=8192 tokens to be re-tokenized by `trtllm-bench` at up to 9820 tokens, overflowing the server's `max_seq_len` and rejecting requests with HTTP 400.

Upstream's own constraints corroborate this: `moonshotai/Kimi-K2.5` README pins `transformers >= 4.57.1`, and the sibling `Kimi-K2.6` (same architecture) caps to `< 5.0.0`. This PR makes trtllm work correctly under either transformers 4.x or 5.x without requiring a transformers downgrade.

## Test plan

- [x] `trtllm-bench` throughput on a Kimi-K2.5 checkpoint under transformers >= 5: confirm the DATASET DETAILS percentiles match the slow tokenizer (e.g. for the random-v2 ratio-0.8 ISL=8192 dataset, P100 should be <= 8192 + chat-template overhead, not ~9820).
- [x] `trtllm-bench` throughput on a non-Kimi model (e.g. Llama, DeepSeek-V3) under transformers >= 5: confirm `use_fast=True` is still selected (no perf regression for the Fast-tokenizer path) by inspecting the tokenizer type after init.
- [x] Kimi K2.5 multimodal serving smoke test: confirm `self._processor(text=...)` and `self._tokenizer(text)` yield the same `input_ids` for a sample prompt.

## References

- https://huggingface.co/moonshotai/Kimi-K2.5/discussions/7 — slow tokenizer issue thread
- https://huggingface.co/moonshotai/Kimi-K2.5/blob/main/README.md — `transformers >= 4.57.1`
- https://huggingface.co/moonshotai/Kimi-K2.6 — sibling model caps `transformers < 5.0.0`
- https://github.com/huggingface/transformers/blob/v5.1.0/src/transformers/tokenization_utils_tokenizers.py — `convert_to_native_format` SentencePiece-then-TikToken fallback

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tokenizer initialization to ensure consistent tokenizer behavior across all text and image preprocessing paths, significantly improving model reliability and performance predictability across different scenarios
  * Enhanced tokenizer selection logic to automatically determine and apply the correct tokenizer implementation for each specific model, improving system compatibility, stability, and overall robustness

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14416?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


qiaoxj07:
Updated approach: `use_fast=False` is a no-op in transformers >= 5 for this tokenizer (verified empirically — `AutoTokenizer.from_pretrained(model_path, use_fast=False)` on a transformers 5.5.3 container still returned `TokenizersBackend, is_fast=True`, and the dataset MAX percentile remained inflated at 9820).

Switched to loading the slow `TikTokenTokenizer` class directly via `transformers.dynamic_module_utils.get_class_from_dynamic_module` from the model repo's remote code, bypassing `AutoTokenizer`'s conversion entirely. The hot-patched bench job on OCI GB200 with transformers 5.5.3 now produces dataset percentiles MIN 6571 / MAX 8156 / AVG 7415 / P99 8155 — exactly matching the transformers 4.57.3 baseline from 2026-05-19.

qiaoxj07:
v3: moved the fix into the runtime tokenizer layer so **trtllm-serve** is also covered.

The previous v2 only fixed the bench-side dataset prep path (`initialize_tokenizer`) and the K2.5 multimodal `KimiK25InputProcessor.__init__` for the `tokenizer is None` branch. trtllm-serve was still broken because `LLM._try_load_tokenizer` -> `ModelLoader.load_hf_tokenizer` -> `TransformersTokenizer.from_pretrained` -> `AutoTokenizer.from_pretrained` still produced the broken Fast tokenizer, which was then handed to `KimiK25InputProcessor` via `create_input_processor` -- bypassing the `tokenizer is None` branch entirely. The original CTX worker's `default_max_tokens (-746) = max_seq_len (8232) - splited_prompt_len (8978)` HTTP 400 errors at the top of the thread were caused by exactly this path.

v3 consolidates:
- New public helper `try_load_dynamic_slow_tokenizer(model_dir, **kwargs)` in `tensorrt_llm/tokenizer/tokenizer.py`, driven by `DYNAMIC_SLOW_TOKENIZER_MAP = {"TikTokenTokenizer": "tokenization_kimi.TikTokenTokenizer"}`.
- `TransformersTokenizer.from_pretrained` now tries the helper before falling back to `AutoTokenizer.from_pretrained`. This fixes every runtime call site (LLM API, trtllm-serve, PyExecutor model engine) at once.
- `bench/utils/data.py:initialize_tokenizer` calls the same helper for the dataset-prep path.
- `modeling_kimi_k25.py` reverts to the original `AutoTokenizer` call (the upstream `_tokenizer` is now slow already); the only retained tweak is pinning `self._processor.tokenizer = self._tokenizer` so the K2.5 processor's internal tokenizer (built independently by `AutoProcessor`) doesn't diverge.

Re-verified on OCI GB200 / transformers 5.5.3 via `trtllm-bench throughput` on Kimi K2.5-NVFP4 with the same 8192/1/ratio-08 dataset: dataset MIN/MAX/AVG/P99 = 6571 / 8156 / 7415 / 8155 (matches the 4.57.3 baseline) and zero HTTP 400 errors during the run.

Wanli-Jiang (approved):

PR #14478, opened 2026-05-23
[None][fix] Release stale-range holder after _commit_block in KVCache…

Tabrizian:
…ManagerV2

When _commit_block finalizes a block that is already inside stale_range[lc] for some finite-window lifecycle (e.g. DSv4 DSA / SWA), the freshly assigned holder (_PageHolder or _SharedPageLock from line 1268 / 1312-1325) was left in place. _check_sanity then asserted on the precondition of suspend():

  /workspaces/tensorrt_llm/.../sampler.../suspend()
    -> _kv_cache.py:931 assert self._check_sanity()
    -> _kv_cache.py:1531 assert holder is None   # for committed stale block

This trips deterministically on DSv4-Pro disagg GEN runs under MAX_UTILIZATION scheduling: when the V2 scheduler tries to suspend an in-flight request to make room, the precondition check aborts the worker. The previously-known release paths (_on_stop_committing, _unlock_stale_blocks) never revisit a committed-and-already-stale ordinal, so a 'commit-into-stale-range' case slipped through.

Fix: at the tail of _commit_block, for the committed block, iterate each non-SSM lifecycle and clear beam_block[lc] when the ordinal lies inside that lifecycle's stale_range. The committed page itself remains alive in the radix tree (committed pages are shareable across caches); only this _KVCache's redundant strong reference is dropped, restoring the 'committed AND stale -> holder is None' invariant.

Validated end-to-end on DSv4-Pro disagg sweep on Lyris GB300 (c=4/8/16/32/64, DEP8/TEP8): zero _check_sanity AssertionError across ~160k commit-into-stale events under real traffic.

@coderabbitai summary

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
+++ b/tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
@@ -1337,16 +1337,11 @@ class _KVCache:
             self._commit_state = self.CommitState.VIRTUAL_STOP
 
         if seq_block.is_committed:
-            ssm_lc_id_fix = self.manager._life_cycles.ssm_life_cycle_id
-            for beam_block_fix in seq_block.pages:
-                for lc_fix_idx, lc_fix_obj in self.manager._life_cycles.items():
-                    if lc_fix_idx == ssm_lc_id_fix:
-                        continue
-                    s_fix, e_fix = _KVCache._get_stale_range(
-                        tokens_per_block, self.history_length, lc_fix_obj
-                    )
-                    if s_fix <= ordinal < e_fix and beam_block_fix[lc_fix_idx] is not None:
-                        beam_block_fix[lc_fix_idx] = None
+            for lc_idx, lc in self.manager._life_cycles.attention_life_cycles():
+                stale_range = _KVCache._get_stale_range(tokens_per_block, self.history_length, lc)
+                if ordinal in stale_range:
+                    for beam_block in seq_block.pages:
+                        beam_block[lc_idx] = None
 
         if is_last or self._commit_state == self.CommitState.VIRTUAL_STOP:
             self._commit_state = self.CommitState.USER_STOP
--- a/tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
+++ b/tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
@@ -1336,6 +1336,18 @@ class _KVCache:
             # We can't commit and can't reuse existing block. Just stop committing.
             self._commit_state = self.CommitState.VIRTUAL_STOP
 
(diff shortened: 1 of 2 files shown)


yizhang-nv:
Could we add a focused regression UT for this state transition?

I locally reproduced the pre-fix failure on a B200 node with a minimal KVCacheManagerV2 sequence:

1. create an SWA KV cache without scratch reuse,
2. `resize(capacity=N, history_length=0)` to allocate real uncommitted pages,
3. `resize(capacity=N, history_length=N)` so early SWA blocks become stale but are still held as `_PageHolder` for future commit,
4. `commit(tokens)`,
5. `suspend()`.

The observed state was:

- after initial resize: `block0/lc0 = _SharedPageLock`
- after advancing history length: `block0/lc0 = _PageHolder`
- after commit: `block0/lc0` was still `_PageHolder`
- `suspend()` then failed in `_check_sanity()` at `assert holder is None`

So my read is that existing SWA tests cover the normal commit-before-stale path and the scratch-reuse path, but not this history-before-commit path. A focused test in `tests/unittest/kv_cache_manager_v2_tests/test_kv_cache_manager_v2.py` would make the intended "history length can advance before KV-cache-reuse commit" behavior explicit and should fail before this patch / pass after it.

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
```diff
@@ -1330,6 +1330,18 @@ def _commit_block(self, ordinal: BlockOrdinal, is_last: bool) -> None:
             # We can't commit and can't reuse existing block. Just stop committing.
             self._commit_state = self.CommitState.VIRTUAL_STOP
 
+        if seq_block.is_committed:
```
lowsfer, line 1339: This could be simplified:
```python
        if seq_block.is_committed:
            for lc_idx, lc in self.manager._life_cycles.attention_life_cycles():
                stale_range = _KVCache._get_stale_range(
                    tokens_per_block, self.history_length, lc
                )
                if ordinal in stale_range:
                    for beam_block in seq_block.pages:
                        beam_block[lc_idx] = None````

lowsfer:
For unit test, I think we can cover it by updating existing tests instead of creating a new one. We can add a parameter delay_commit to TestNoBatching.run_naive defaulting to False, then if True, only commit all tokens at end. And add that arg to TestNoBatching.test_naive().

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
Tabrizian: I simplified it.

yizhang-nv (approved):

lowsfer (approved):

Issue #14225, opened 2026-05-17
[AutoDeploy]: Investigate SuperV3 with MTP not scaling well


galagam:
### 🚀 The feature, motivation and pitch

AD surpasses PT at WS=1 across all concurrencies — 18% faster at c=1, near-parity (1%) at c=64.
AD fails to scale at WS=4: PT leads by 20–35%. The gap is driven by poor WS=1→4 scaling on the AD side: PT scales 1.52× from WS=1 to WS=4 at c=1 (4.67ms→3.07ms), while AD only gains 1.04× (3.82ms→3.67ms). 

Investigate and resolve the poor scaling on AutoDeploy side.
Baseline branch: `nv-auto-deploy:gagam/super-mtp-perf-2-replay` (see #13725)

### Additional context

See SuperV3 MTP ticket #12359 
Scripts, configs and experiment data: 
https://gitlab-master.nvidia.com/ghubaraagam/agent-reports/-/tree/main/260428_superv3_mtp?ref_type=heads

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and checked the [documentation](https://nvidia.github.io/TensorRT-LLM/) and [examples](https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples) for answers to frequently asked questions.

galagam:
@MrGeva Pushing anything to main seems hopeless at this point. I cleaned up and rebased, hopefully you can take it from here while I keep battling CI issues.

MrGeva:
@galagam ack

MrGeva:
<img width="940" height="202" alt="Image" src="https://github.com/user-attachments/assets/f91587b3-cf67-458c-b7c3-a8fcee1174cb" />

MrGeva:
<img width="2964" height="1764" alt="Image" src="https://github.com/user-attachments/assets/54b9d063-b9be-4292-a839-a10a5dbdf1a7" />
sha1: 5504cf859c22500a603a3084e1e9916a828341c2

MrGeva:
PR: https://github.com/NVIDIA/TensorRT-LLM/pull/14505
sha1: 307f3c7cbbabc279bcdc42cec9c947b868904c4a
WIP analyzing the c>32 perf gap
<img width="1040" height="715" alt="Image" src="https://github.com/user-attachments/assets/8abcd657-2954-49e0-be8e-85ce465ce604" />

[...]

PR #13723, opened 2026-05-04
[#13718][feat] AutoDeploy MoE all-to-all: cache + runtime max-tokens

greg-kwasniewski1:
## Summary

Closes #13718.
Closes #13360.

 
 
 # Implementation

- **Module-level `_GlobalMoeA2ACache`** in `tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe/trtllm_moe.py` caches `MoeAlltoAll` instances and parsed `DistConfig` mappings, eliminating per-call Python ctor + JSON deserialization (mirrors `_GlobalTrtllmPlanner` in `trtllm_attention.py`).
- **Runtime max-tokens-per-rank** via `BatchInfo` slot 13: pre-forward `tp_allgather` in `ADEngine` writes the cross-rank `max(total_num_tokens)` to the slot; the MoE custom op reads it at capture/eager time. Replaces the static `max_num_tokens` over-padding with the per-iteration cross-rank max — matches base TRT-LLM's `runtime_max_tokens_per_rank` from `fused_moe_cutlass.py:764`.

Files touched:
- `tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py` — `BatchInfo._NUM_ELEMENTS = 14`, slot 13 helpers
- `tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py` — pre-forward `tp_allgather` override
- `tensorrt_llm/_torch/auto_deploy/transform/library/sharding.py` — thread `batch_info_host` graph input into MoE op call sites
- `tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe/trtllm_moe.py` — `_GlobalMoeA2ACache` + slot 13 read in dispatch path
- `tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe/torch_moe.py` — `batch_info_host` kwarg threading for sharding-time targets

## Why this is safe under CUDA-graph capture

The `int(batch_info_host[13].item())` read is on a pinned-host CPU tensor — pure Python read, no GPU sync. Under graph capture the value is baked in at capture time, so each captured graph corresponds to the `cg_batch_size` bucket it was warmed at. AD's existing `maybe_pad_for_cuda_graph` (`ad_executor.py`) syncs `batch_size` across ranks via `tp_allgather` and pads up to `cg_batch_size` — so for any decode replay path, all DP ranks have the same `total_num_tokens`, and the captured constant equals the runtime cross-rank max.

[...]

Issue #13718, opened 2026-05-03
[AutoDeploy][Feature]: Stateful MoE all-to-all (instance cache + runtime max-tokens)


greg-kwasniewski1:
## 🚀 The feature, motivation and pitch

Auto-deploy's MoE all-to-all path under attention-DP currently lags base TensorRT-LLM in two ways:

1. **Per-call `MoeAlltoAll` re-instantiation.** The current `_run_moe_with_alltoall` helper in `tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe/trtllm_moe.py` re-instantiates `MoeAlltoAll` on every layer call, re-deserializes the `DistConfig` JSON via `_check_moe_alltoall`, and re-runs `MoeAlltoAll.calculate_required_workspace_size`. The class-level `MoeAlltoAll._WORKSPACE` already deduplicates the GPU workspace allocation, but the Python ctor + JSON parse + 6 dict-equality asserts run per-call. With ~30 MoE layers × ~N forward steps per decode iteration, this is hundreds of redundant ctor invocations per process.

2. **Static `max_num_tokens` for dispatch padding.** The current code uses `runtime_max_tokens_per_rank = max_num_tokens` (the config-level upper bound) to size MoE all-to-all dispatch padding. For typical decode batches with `local_num_tokens << max_num_tokens`, this wastes substantial NVLink bandwidth on padded zero-rows. Base TensorRT-LLM uses `max(all_rank_num_tokens)` from a per-forward `tp_allgather` (see `tensorrt_llm/_torch/pyexecutor/model_engine.py::_get_all_rank_num_tokens`), which is structurally constant within a captured CUDA graph thanks to cross-rank batch-size sync.

This issue tracks two related changes that together bring auto-deploy's MoE all-to-all path in line with base TRT-LLM:

**Phase 1: `_GlobalMoeA2ACache`** — module-level singleton cache for `MoeAlltoAll` instances + parsed `DistConfig` mappings, restricted to `trtllm_moe.py`. Mirrors the `_GlobalTrtllmPlanner` precedent in `tensorrt_llm/_torch/auto_deploy/custom_ops/attention/trtllm_attention.py`. Lazy first-call init, keyed on workspace-defining ctor parameters.

[...]

PR #14505, opened 2026-05-24
[#14225][feat] AutoDeploy MTP-eagle attention-DP support

MrGeva:
This PR is rebased on #13723

## 1. MTP-eagle one-model correctness under attention-DP

Files: `models/custom/modeling_eagle.py`, `models/eagle.py`, `pyexecutor/py_executor.py`,
`shim/ad_executor.py`, `transform/library/sharding.py`, test `singlegpu/models/test_eagle.py`.

Getting MTP-eagle (one-model speculative decoding) to produce **correct** output together with
attention-DP required several coordinated fixes:

- **Seed `py_draft_tokens` on the attention-DP padding dummy request.** Under attention-DP,
  ranks with no real work are padded with a dummy generation request. With an empty
  `py_draft_tokens` that dummy was classified as a *decode* sequence, which trips the eagle
  wrapper's invariant `assert num_decode == 0` (MTP one-model has no pure-decode sequences —
  every generation sequence is an *extend* of `1 + max_draft_len` tokens). The dummy is now
  seeded with `py_draft_tokens = [1] * max_total_draft_tokens` so it materializes as extend.
- **Forward the draft sizes to PyExecutor** so the dummy is built with the correct draft length
  (`max_draft_len` / `max_total_draft_tokens` are passed through at executor creation).
- **ADP-aware sample broadcast + draft EP revert.** The eagle wrapper's sampling/broadcast is
  made consistent across DP ranks, and the draft model is kept off the EP-sharding path
  (`sharding.py`) so the drafter's small projections aren't expert-parallel-sharded.
- Added a unit test for the broadcast/draft gate (`test_eagle.py`).

Net effect: MTP-eagle one-model + attention-DP runs without NaNs / shape asserts and passes the
gsm8k accuracy gate (~92.x %, acceptance ≥ 0.45).

## 2. MoE all-to-all: no input over-padding

`custom_ops/fused_moe/trtllm_moe.py`.

The dispatch kernel writes **only the real `local_num_tokens` rows** into the one-sided
workspace, and `moe_a2a_sanitize_expert_ids` marks the unfilled slots as invalid so the
grouped MoE GEMM skips them. We therefore **stopped padding the input `x`** up to the per-rank
budget before dispatch. Only the workspace *stride* / recv-view / GEMM row count is governed by
the budget (see §4); the actual data moved is the true token count. This removed redundant
copy/compute on the padding region (1.5–3× OTPS in the low/mid-concurrency regime).

## 3. MoE all-to-all token-budget selection policy

`custom_ops/fused_moe/trtllm_moe.py` — `_hybrid_runtime_max_tokens_per_rank(...)`.

This is the value-selection logic for the per-rank budget (the budget itself is explained in
§4). It evolved through several commits to its current "hybrid" form:

```python
if torch.cuda.is_current_stream_capturing() or cuda_graph_state.in_warm_up():
    budget = local_num_tokens
    if budget > 0 and budget * ep_size * 4 <= max_num_tokens:
        return budget            # tight stride (decode/extend under cudagraph capture)
return max_num_tokens            # static fallback (eager, or large budgets)
```

Rationale:
- **Tight under capture (decode/extend):** use the actual `local_num_tokens` (a *shape*, taken
  from `x.shape[0]`), so the all-to-all stride / GEMM rows are small → faster dispatch and a
  small-M `bmm_E2m1` GEMM tactic. Because it's a shape (not a device scalar read with `.item()`),
  there is **no per-layer host sync** — that host sync was what made an earlier slot-13-based
  version regress at low concurrency.
- **`× ep_size × 4 ≤ max_num_tokens` gate:** only take the tight stride while the resulting MoE
  GEMM row count (`budget × ep_size`) stays in the fast small-M tactic band; above it the
  trtllm-gen autotuner picks a launch-latency-bound tactic (a c=32 regression), so fall back to
  the static budget.
- **Static `max_num_tokens` otherwise (eager / large budgets):** a constant every rank computes
  identically, so the stride is consistent across ranks with no host sync.

Cross-rank consistency holds in both branches: under attention-DP every rank pads its decode
batch to the same cudagraph batch size, so the captured tight value equals the cross-rank value;
the eager branch is a constant. (Mixed prefill/decode — where a captured value would be stale —
is routed to eager by the `#13718` force-eager guard, which is gk's, not ours.)

## 4. The "budget" (`runtime_max_tokens_per_rank`) — what it is

The MoE layer runs **expert parallel + attention-DP**, so each forward step it does a one-sided
MNNVL **all-to-all**: every rank scatters its tokens to the ranks that own the selected experts
(dispatch), the experts run, then results are gathered back (combine). The shared workspace is a
fixed `[ep_size, max_tokens_per_rank, hidden]` buffer in peer memory; a rank writes token `t`
destined for `target_rank` at offset `target_rank * max_tokens_per_rank + slot`.

`runtime_max_tokens_per_rank` (the **budget**) is the **per-rank stride** of that workspace for a
given step — i.e. how many token-slots each rank's region spans. It drives:
- the dispatch/sanitize/combine **addressing stride** (where writes/reads land),
- the **recv-view shape** the MoE GEMM consumes (`[ep_size, budget, hidden]`),
- and therefore the **GEMM row count** (`budget × ep_size`).

It does **not** change how many real tokens are moved — dispatch writes only `local_num_tokens`
rows and sanitize invalidates the rest. The workspace is always *allocated* at the static
`max_num_tokens`; the budget just picks how much of that stride is actually used per step (the
guard `assert runtime_max ≤ max_num_tokens` always holds).

Why it matters: the budget is a **cross-rank-shared** quantity — all ranks must agree on it or
the one-sided writes land at mismatched offsets and corrupt each other. A *smaller* budget means
a tighter stride and a smaller MoE GEMM (faster), but it must stay consistent across ranks and,
under CUDA graphs, must not be a value that gets baked at capture time and then diverges at
replay. §3 is exactly the policy that picks a small budget when it is safe (uniform decode under
capture) and a safe constant otherwise.

gk's force-eager (_bypass_captured_graphs, #13718) forces all ranks to the eager path the moment any rank can't graph (mixed). The eager branch of _hybrid returns the constant max_num_tokens, so all ranks agree. That's the entire reason force-eager exists — to guarantee this equality. (Without it: the decode rank, graphed, would have baked its small local_num_tokens; the prefill rank, eager, would use max_num_tokens → diverge → the NaN the commit fixed.)

## 5. attention-DP request balancer (the c≤128 throughput win)

`shim/ad_executor.py` (commit `233231ab7a`).

AutoDeploy was hardcoding `self.llm_args.attention_dp_config = None`, which disabled
PyExecutor's own `_balance_adp_requests` scheduler. We enable it:

```python
self.llm_args.attention_dp_config = AttentionDpConfig(
    enable_balance=True,
    batching_wait_iters=10,
    timeout_iters=16,
)
```

Why: under attention-DP a rank holding a full prefill while its siblings decode is a
**straggler** — the MoE all-to-all is a collective, so the decode ranks reach the dispatch
barrier and spin-wait for the prefill rank (these stalls dominated GPU time at high
concurrency). The balancer coordinates **context (prefill) scheduling across ranks** so prefills
land in the same iteration instead of one rank stalling the rest. `timeout_iters=16` caps how
long a prefill is deferred while waiting for the other ranks to also have context (kept small so
it does not starve the decode batch at low/mid concurrency); `batching_wait_iters=10` controls
the cross-rank prefill batching once all ranks have context. It is a no-op when attention-DP is
off.

Result (eosfixed, OTPS vs no-balancer): c=1 +1 %, c=16 +7 %, c=32 +13 %, c=64 +25 %,
**c=128 +39 % (5171 → 7190)**; gsm8k unchanged (92.30 %). At c=128 AutoDeploy goes from 0.69× to
~0.95× of the PyTorch backend, winning at every concurrency 1–128.

## 6. Config

Files: `examples/auto_deploy/.../super_v3_mtp*.yaml` (registry configs) and a low-latency
SuperV3 yaml. These wire attention-DP on, the manual EP/bmm sharding plan, the MTP-eagle
one-model spec config, the trtllm-gen NVFP4 MoE backend, and the triton SSM / causal-conv
backends (the only ones that currently support speculative Mamba state caching).

@coderabbitai summary

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/auto_deploy/transform/library/sharding.py
```diff
@@ -1233,6 +1233,27 @@ def _apply(
...
+                    _is_draft
+                    and config.dist_config.enable_attention_dp
+                    and transform_container.ep_transforms
+                ):
+                    ad_logger.info(
+                        "Reverting %d EP sharding transform(s) on draft submodel "
+                        "under attention_dp (replicating instead).",
+                        len(transform_container.ep_transforms),
+                    )
+                    transform_container.ep_transforms.clear()
```
MrGeva, line 1256: check perf impact

MrGeva:

tensorrt_llm/_torch/pyexecutor/py_executor.py
MrGeva: revisit this. can it be avoided?

tensorrt_llm/_torch/auto_deploy/transform/library/sharding.py
MrGeva: opt#1 _ sharding.py (revert draft EP-sharding under ADP): REQUIRED (functionality).
  - opt#1 OFF faults under ADP serving load (c=32): dispatch: Rank 1/2/3 timed out waiting for completion flag from rank 0 + Sampling failed _ the exact draft-alltoall workspace corruption the comment predicts.
  - It did run at low load (build_and_run, bs=2), so the bug is load-dependent _ which is why it's a real, non-obvious requirement, not removable.

PR #14550, opened 2026-05-26
[None][chore] Remove one-warp-per-token policy from MoE A2A kernels

bobboli:
## Summary

Drop the `WarpPolicy` (one-warp-per-token) code path from MoE A2A dispatch and combine kernels, keeping only `BlockPolicy` (one-block-per-token).

## Motivation

`BlockPolicy` was already the default. `WarpPolicy` is only beneficial when each token has very little work — i.e. very small hidden sizes — which is not representative of modern MoE models. The "many CTAs" concern that motivates `WarpPolicy` only triggers at batch sizes well above production. Keeping a code path that wins only outside the target regime — and silently regresses small/medium batches via the combine kernel — is not worth the maintenance cost.

## Supporting data

B300x8, EP=8, NVLink one-sided, blog18 repro (`tests/microbenchmarks/bench_moe_comm.py`, `--perfect_router`, median dispatch+combine µs across ranks).

Modern hidden sizes — `BlockPolicy` wins across the full batch sweep:

| config | batch | block (µs) | warp (µs) | block speedup |
|---|---|---|---|---|
| h=7168, FP8_BLOCK (DeepSeek-V3) | 4096 | 1388 | 1427 | +2.8% |
| h=4096, FP8_BLOCK | 4096 | 816 | 858 | +5.2% |
| h=4096, NVFP4 | 4096 | 576 | 603 | +4.8% |
| h=4096, NO_QUANT | 4096 | 818 | 856 | +4.7% |

`WarpPolicy` only catches up in a narrow corner — small hidden + aggressive quant + large batch:

| config | batch | block (µs) | warp (µs) | warp speedup |
|---|---|---|---|---|
| h=1024, NVFP4 | 2048 | 132 | 104 | +20.9% |
| h=1024, NVFP4 | 4096 | 229 | 189 | +17.7% |
| h=2880, MXFP4 (gpt_oss) | 4096 | 271 | 237 | +12.7% |

Combine-side `WarpPolicy` is multiple × slower than `BlockPolicy` at small/medium batches due to grid sizing (e.g. h=7168 FP8 b=1: 23 µs block vs 208 µs warp; h=4096 FP8 b=64: 23 µs vs 149 µs). This drags total latency far below block until b≈1024 across all configs.

## What changes

- Remove `WarpPolicy` struct, `SWITCH_POLICY` macro, and all `grid_size_warp` paths in dispatch / prepare-combine / combine launches.
- Remove `one_block_per_token` field from `MoeA2ADispatchParams` and `MoeA2ACombineParams`.
- Remove `getEnvMoeA2AOneBlockPerToken()` and the `TLLM_MOE_A2A_ONE_BLOCK_PER_TOKEN` env knob.

No behavior change for users not overriding the env var.

### DIFF
--- a/cpp/tensorrt_llm/common/envUtils.cpp
+++ b/cpp/tensorrt_llm/common/envUtils.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -530,17 +530,6 @@ bool getEnvDisableChunkedAttentionInGenPhase()
     return getBoolEnv("TRTLLM_DISABLE_CHUNKED_ATTENTION_IN_GEN_PHASE");
 }
 
-bool getEnvMoeA2AOneBlockPerToken()
-{
-    // Default true; return false only if env set to "0"
-    static std::optional<int32_t> const val = getIntEnv("TLLM_MOE_A2A_ONE_BLOCK_PER_TOKEN");
-    if (!val.has_value())
-    {
-        return true;
-    }
-    return val.value() != 0;
-}
-
 static int sanitizeBlockSize(std::optional<int32_t> const& val)
 {
     // Default 256 when not set or invalid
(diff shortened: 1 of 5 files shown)


bobboli:
Addressed CodeRabbit feedback in `ec26ba0f76`:

1. `clang-format` applied to `moeAlltoAllKernels.cu` launch block.
2. Wrapped `sanitizeBlockSize(getIntEnv(...))` in `getSanitizedBlockSizeFromEnv()` helper that catches `std::stoi` exceptions, logs a warning, and falls back to default — malformed block-size env vars no longer crash.

djns99 (approved):

PR #14555, opened 2026-05-26
[TRTLLM-12963][refactor] LTX-2 attention: drop dead k_pe parameter; require cached cross-attn

luyiyun1021:
@coderabbitai summary

## Description

Drop the now-dead `k_pe` parameter from `LTX2Attention.forward` and `_forward_unfused`, and enforce a clean cached-cross-attn invariant: cross-attn requires `pre_projected_kv` from `project_kv`. The uncached SEPARATE_QKV branch becomes self-attn-only.

## Why

`k_pe` was the optional per-K rope override on `LTX2Attention.forward` — it let cross-attn pass a different rope for K than for Q. It became dead code after the earlier norm+rope fusion work moved K-rope into `project_kv`. That move had two goals:

1. **Eliminate the cos/sin all-gather under Ulysses** — rope is per-token element-wise and commutes with seq-dim concat, so rotating K on the local shard before all-gather is bit-identical and saves a collective.
2. **Enable K-norm + K-rope to go through a single fused kernel** (`fused_dit_split_norm_rope`) inside `project_kv`.

After that, every production cross-attn caller pre-rotates K via `project_kv` and then calls `forward` with `pre_projected_kv=(k, v)` and `k_pe=None`. Grep confirms zero non-`None` `k_pe` call sites. The "inline fallback when `text_kv_video`/`text_kv_audio` is None" path documented on `BasicAVTransformerBlock.forward` is also unreachable in production: `prepare_text_cache` always populates the KV list when its modality is enabled, and the block only invokes the cross-attn when the corresponding modality is enabled — so the cached KV is always present.

The old uncached SEPARATE_QKV cross-attn branch additionally had a latent silent footgun: `k_pe if k_pe is not None else pe` applied Q's rope to K when both came from different sources with potentially different sequence lengths.

## What changed (+40 / −44, 2 files)

`tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py`:
- Drop `k_pe` from `LTX2Attention.forward` and `_forward_unfused`.
- Drop the `k_pe=None` kwargs from `BasicAVTransformerBlock`'s a2v / v2a call sites.
- Replace `k_pe if k_pe is not None else pe` in the uncached SEPARATE_QKV branch with just `pe`.
- Enforce: uncached SEPARATE_QKV is self-attn-only — `if context is not None: raise ValueError(...)` (survives `python -O`, unlike `assert`). Cross-attn must use `pre_projected_kv` from `project_kv`. Same invariant in `_forward_unfused`.
- Update `BasicAVTransformerBlock.forward` docstring: `text_kv_video` / `text_kv_audio` are required when the corresponding stream runs cross-attn (built by `LTXModel.prepare_text_cache`) — drop the "falls back to inline computation" wording.
- Update class + `project_kv` docstrings to drop `k_pe` references.

`tests/unittest/_torch/visual_gen/test_ltx2_attention.py`:
- `TestLTX2CrossAttention.test_cross_attention_sanity` and `test_cross_attention_different_dims` now use the production cached pattern: `k, v = attn.project_kv(ctx, pe=...)` → `attn(x, pre_projected_kv=(k, v), pe=...)`. The previous `attn(x, context=ctx, pe=None)` shape was exercising the unreachable inline-fallback path and is now rejected.

## Test Coverage

Local on B200, wt-3 container:
- `tests/unittest/_torch/visual_gen/test_ltx2_attention.py` — **6/6 PASS**
- `tests/unittest/_torch/visual_gen/test_ltx2_transformer.py` — **14/14 PASS**

No production behavior change.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
+++ b/tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
@@ -72,7 +72,7 @@ class LTX2Attention(Attention):
     - Output projection (to_out)
 
     Adds LTX-2 specifics:
-    - LTX 3D RoPE (INTERLEAVED / SPLIT) with separate k_pe support
+    - LTX 3D RoPE (INTERLEAVED / SPLIT)
     - Gated attention (to_gate_logits)
     - Cross-attention with different context_dim for K/V input
     """
@@ -275,8 +275,8 @@ class LTX2Attention(Attention):
         before all-gather. RoPE is per-token element-wise so it commutes with
         seq-dim concat — bit-identical to the post-gather rope while saving
         the cos/sin all-gather collective and reducing K-rope compute by U×.
-        The forward() consumer should pass ``k_pe=None`` to signal that K is
-        already rotated.
+        After this, K is already rotated, and the forward() consumer passes
+        ``pre_projected_kv=(k, v)`` to skip re-rotation.
         """
         k = self.to_k(context)
         v = self.to_v(context)
@@ -299,19 +299,23 @@ class LTX2Attention(Attention):
         x: torch.Tensor,
         context: torch.Tensor | None = None,
         pe: tuple[torch.Tensor, torch.Tensor] | None = None,
-        k_pe: tuple[torch.Tensor, torch.Tensor] | None = None,
         pre_projected_kv: tuple[torch.Tensor, torch.Tensor] | None = None,
         key_padding_mask: torch.Tensor | None = None,
     ) -> torch.Tensor:
         """Forward pass.
 
         Caller contract:
-          - FUSE_QKV (self-attn): pe must be set; k_pe and pre_projected_kv unused.
-          - SEPARATE_QKV (cross-attn): cached path requires pre_projected_kv;
-            uncached path uses ``context`` (may be None when the async-Ulysses
-            inner backend was swapped to a non-async one — falls back to
-            self-attn via kv_source=x). pe optional (None = norm-only).
-            k_pe overrides pe for K (e.g. AV cross-attn) when provided.
+          - FUSE_QKV (self-attn): pe must be set; pre_projected_kv unused.
+          - SEPARATE_QKV self-attn (async-Ulysses): pre_projected_kv=None,
+            context=None — routed to ``forward_async`` (V/Q/K rolling A2A).
+            Falls through to the sync SEPARATE_QKV self-attn path when the
+            inner backend lacks ``forward_async`` (Ulysses-inactive swap).
+          - SEPARATE_QKV cross-attn: pre_projected_kv must be set (K already
+            norm+rope'd by ``project_kv`` upstream — text cache or AV
+            project-before-gather). pe optional (None = norm-only on Q).
+            Uncached cross-attn (context != None without pre_projected_kv) is
+            rejected: Q/K may have different lengths so sharing pe would
+            mis-rotate K. Caller must use project_kv + pre_projected_kv.
 
         Args:
             key_padding_mask: Optional ``[B, S_kv]`` bool tensor; True = valid,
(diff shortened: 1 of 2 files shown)


tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py
```diff
@@ -267,29 +262,27 @@ def forward(
...
-                # The caller is responsible for any K-norm + K-rope on the
-                # cached tensor; we only fuse Q here.
                 k, v = pre_projected_kv
                 q = self.to_q(x)
                 self.apply_split_norm_or_norm_rope(
                     q, self.norm_q.weight, self.num_attention_heads, pe
                 )
             else:
-                # Uncached cross-attn (not exercised by LTX-2 in practice; kept for fuse-dispatch consistency).
+                assert context is None, (
```
yibinl-nvidia: This removes the uncached cross-attn path, but `BasicAVTransformerBlock.forward()` still accepts `text_kv_video=None` / `text_kv_audio=None` and documents that those fall back to inline computation [here](https://github.com/NVIDIA/TensorRT-LLM/blob/f6ba936179b19e9dc8d3dcc05eefa0039cfdbe56/tensorrt_llm/_torch/visual_gen/models/ltx2/transformer_ltx2.py#L578). With this assert, a caller that omits the text cache now reaches `attn2(..., context=..., pre_projected_kv=None)` and fails here. We would need to ensure that cached KV is always required and raise a clear error if it is None.

yibinl-nvidia: optimized python code (`-O`) ignores `assert`, consider raising an error instead `if context is not None: raise ValueError(...)`

luyiyun1021: Good catch — fixed in 3575972c3a. The `assert context is None` was too strict and would have broken the `BasicAVTransformerBlock.forward()` inline fallback path you flagged (when `text_kv_video=None` / `text_kv_audio=None`, `attn2(rms_norm(vx), context=video.context, pre_projected_kv=None)` reaches that branch with `context != None`).

The actual unsafe case is only `uncached cross-attn + pe != None` — applying Q's pe to K when Q and K can have different sequence lengths. In LTX-2 production, the inline text cross-attn fallback always has `pe=None` (text positions don't carry rope), so the path is safe.

New guard: `if context is not None and pe is not None: raise ValueError(...)`. This preserves the documented inline-fallback path (`pe=None`) and rejects only the silent-mis-rotation footgun (`pe!=None`).

luyiyun1021: Right — switched to `raise ValueError` in 3575972c3a so the check survives `python -O`.

luyiyun1021: Update — went stricter in b3b5e22aa4 (the previous 3575972c3a was correct but conservative).

On further trace, the `BasicAVTransformerBlock.forward()` inline fallback path (when `text_kv_video=None` / `text_kv_audio=None`) is **never actually reachable** in production: `prepare_text_cache` always populates the KV list when its modality is enabled, and the block only invokes cross-attn when that modality is enabled — so cached KV is always present whenever cross-attn runs. The documented fallback was defensive code with no live caller.

New invariant (matches your second sentence — 'ensure that cached KV is always required and raise a clear error if it is None'):
- uncached SEPARATE_QKV is self-attn-only (`context` must be None)
- cross-attn (`context != None`) without `pre_projected_kv` → `raise ValueError(...)`

This kills both (a) the silent-mis-rotation footgun and (b) the unreachable inline-fallback path. `BasicAVTransformerBlock.forward` docstring updated to drop the 'falls back to inline computation' wording. `TestLTX2CrossAttention` tests updated to use the production `project_kv` + `pre_projected_kv` pattern. 20/20 LTX-2 unit tests still PASS.

yibinl-nvidia (approved):
LGTM

zhenhuaw-me (approved):

PR #14560, opened 2026-05-26
[None][fix] py_executor: lift conditional tp_allgather sites out of per-rank-divergent gates

lancelly:
## Summary

Fix the GEN-side ADP-router desync that crashes all 8 ranks of DSv4-Pro DEP8+DEP8 disagg with one of:

```
TypeError: RankState.__init__() takes from 2 to 4 positional arguments but 27 were given     # TLLM_METRICS_ALL_RANKS=1, doc-variant 1
TypeError: argument of type 'bool' is not iterable                                           # TLLM_METRICS_ALL_RANKS unset,  doc-variant 2
TypeError: 'int' object is not iterable     at tp_cp_allgather                               # observed in this PR's A/B verify (variant 3)
```

Repro doc: `ape-repo/astra-projects/wwfo/artificial-analysis@DSV4-disagg:docs/knowledge/dsv4_disagg_dep8_adp_router_desync.md`.

## Root cause

Two conditional `tp_allgather` collectives in `tensorrt_llm/_torch/pyexecutor/py_executor.py` live inside per-rank-divergent gates:

1. **`_handle_responses` line 4438** — `tp_allgather(bool(timed_out_requests))`. Reached from `_process_previous_batch` (overlap loop), which is gated on `should_process_previous_batch = can_queue or not can_queue_this_rank` plus `previous_batch is not None`. When one ADP rank's `scheduled_batch.batch_size == 0` and its `previous_batch` was cleared the iter before, that rank skips `_process_previous_batch` while peers enter it. Peers issue the `tp_allgather`, the empty-batch rank does not — the next collective in the queue (the very next `gather_all_rank_states`) then receives the bool payload positionally as `RankState` fields and crashes on every rank.

2. **`_append_iter_stats` line 1436** — `tp_allgather(local_dict)` (27-key payload under `TLLM_METRICS_ALL_RANKS=1`). Same outer gate via `_process_iter_stats`. With the env var set, the per-rank-divergent branch fires both the timeout gather AND the iter-stats gather, so the misalignment moves by two positions and the 27-key dict lands in the `RankState.deserialize` slot.

The three observed variants (dict, bool, int) are the same bug, differing only in how many extra collectives the divergent branch issued and which downstream collective consumes the misaligned payload.

## Fix

Mirror the existing `_flush_pending_transfer_responses` pattern: buffer the payload on `self` and drain at a rank-symmetric flush point.

- `_handle_responses` → push to `self._pending_timed_out_requests`. New helper `_handle_kv_transfer_timeouts_synced()` does the ADP-safe drain.
- `_append_iter_stats` → push to `self._pending_iter_stats_dict`. New helper `_flush_iter_stats_synced()` does the ADP-safe drain (gathers `None` from idle ranks so the queue stays aligned).
- Both helpers are called from a rank-symmetric position in each executor loop:
  - **Overlap loop**: right after the `if previous_batch is not None and should_process_previous_batch` block (immediately adjacent to the existing `_enqueue_responses([])` sync sibling in the `else` branch).
  - **Non-overlap loop**: right after the rank-symmetric `if can_queue` block that calls `_handle_responses` (outside the if to be defensive against future restructuring).
  - **PP loop / `_handle_executed_batch`**: at the bottom of every call, outside the `executed_batch is not None` conditional. `handle_executed_batches(executed_batch_num)` calls this helper an equal number of times on every PP rank (broadcast `executed_batch_num`).

No new collectives introduced — the existing two are just moved out of the divergent branches to the same flush point the codebase already uses for `_enqueue_responses`.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -424,6 +424,12 @@ class PyExecutor:
         # responses and flushing them at a synchronised point in the executor
         # loop avoids the mismatch.
         self._pending_transfer_responses: List[Tuple[int, LlmResponse]] = []
+        # Same buffer-then-synced-flush pattern as _pending_transfer_responses
+        # above: _handle_responses and _append_iter_stats are reached from
+        # per-rank-divergent gates, so their tp_allgather collectives are
+        # lifted to _handle_kv_transfer_timeouts_synced / _flush_iter_stats_synced.
+        self._pending_timed_out_requests: List[LlmRequest] = []
+        self._pending_iter_stats_dict: Optional[Dict] = None
         # ADP dummy role for _pad_attention_dp_dummy_request. Default is gen;
         # updated from observed request types.
         self._adp_dummy_is_gen: bool = True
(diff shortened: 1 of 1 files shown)


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -707,6 +713,53 @@ def _flush_pending_transfer_responses(self):
...
+        gathered = self.dist.tp_allgather(local_dict)
+        if self.dist.tp_rank != 0:
+            return
+        rank_dicts = [d for d in gathered if d is not None]
+        if not rank_dicts:
+            return
+        with self.stats_lock:
+            cap = self.max_stats_len * tp_size
+            overflow = max(0, len(self.stats) + len(rank_dicts) - cap)
+            if overflow:
```
reasonsolo: This is awkward, appending all rank_dicts and trimming by cap is much more human readable

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -707,6 +713,53 @@ def _flush_pending_transfer_responses(self):
...
+        """ADP-safe drain of the TLLM_METRICS_ALL_RANKS dict-gather collective.
+
+        Lifted out of _append_iter_stats (same divergent-gate problem as
+        _handle_kv_transfer_timeouts_synced).  Under ADP every rank sends
+        either the local dict or ``None`` every iter so the gather stays in
+        lockstep with peer collectives.  When the gate doesn't hold,
+        _append_iter_stats takes its legacy path and never populates the
+        buffer.
+        """
+        tp_size = getattr(self.dist, "tp_size", 1)
```
reasonsolo: `self.dist.tp_size` is accessible for sure. Agents use `getattr` toooo much.

reasonsolo (approved):

tensorrt_llm/_torch/pyexecutor/py_executor.py
lancelly: Yes. This function is extracted out of the origin code.
I will fix this.

tensorrt_llm/_torch/pyexecutor/py_executor.py
lancelly: This is aligned with origin code, but I think you are right. Fixed.

zheyuf (approved):
LGTM

PR #14566, opened 2026-05-26
[None][fix] gate NCCL window autotune pre-hook

liji-nv:
The allreduce autotuner can profile with dummy tensors. When the real input is not an NCCL symmetric window buffer, registering a dummy window input during autotune lets some fallback tactic get incorrect perf result.

Add an NCCL-window buffer detector and include the real input window state in the allreduce runner identity. Only install the autotune pre-hook that allocates a symmetric window input when the real allreduce input already uses the NCCL symmetric memory window.

@coderabbitai summary

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/tensorrt_llm/thop/allreduceOp.cpp
+++ b/cpp/tensorrt_llm/thop/allreduceOp.cpp
@@ -1541,6 +1541,47 @@ void preallocateNCCLWindowBuffer(
 #endif
 }
 
+bool isNCCLWindowBuffer(torch::Tensor const& input, torch::List<int64_t> const& group)
+{
+#if ENABLE_MULTI_DEVICE
+    if (!input.is_cuda() || input.numel() == 0 || group.size() == 0)
+    {
+        return false;
+    }
+
+    std::set<int> groupSet;
+    for (auto const& rank : group)
+    {
+        groupSet.insert(static_cast<int>(rank));
+    }
+
+    std::shared_ptr<ncclComm_t> commPtr;
+    try
+    {
+        commPtr = getComm(groupSet);
+    }
+    catch (std::exception const& e)
+    {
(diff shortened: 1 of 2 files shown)


cpp/tensorrt_llm/thop/allreduceOp.cpp
```diff
@@ -1541,6 +1541,47 @@ void preallocateNCCLWindowBuffer(
...
+    }
+
+    if (!commPtr || *commPtr == nullptr)
+    {
+        TLLM_LOG_DEBUG("[isNCCLWindowBuffer] NCCL comm is null");
+        return false;
+    }
+
+    using tensorrt_llm::common::nccl_util::NCCLWindowAllocator;
+    auto const buffer = NCCLWindowAllocator::getInstance().searchBuffer(*commPtr, input.data_ptr());
```
nv-lschneider, line 1576: If this causes problems, we may have to upgrade searchBuffer from a linear search to a hashmap search. Since this lookup is now in the hot CPU path for every AR call.

nv-lschneider (approved):
Thanks for upgrading the logic for the auto-tuning.

I think we should strive to get more inputs to be window backed if possible, but this is the correct logic for all inputs we haven't upgraded yet.

cpp/tensorrt_llm/thop/allreduceOp.cpp
liji-nv: I guess the host overhead is still under control given limited number of nccl window buffer in use. Previously for the UB manager, my assumption is there will be up to 3 UB buffers registered if tensor released correctly. When the number of buffers is small using O(N) linear algo may not be too bad comparing with hash map.

nv-lschneider: Agreed, I only flagged that for the future. For the current situation, this is non-problematic.

hyukn (approved):

PR #14576, opened 2026-05-26
[#14502][fix] nano-v3: swap leaked reasoning into content when enable_thinking=False

### LINKED ISSUE #14502: [llmapi] NemotronV3ReasoningParser returns empty content when enable_thinking=False
## System Info

- TensorRT-LLM version: 1.3.0rc9
- Model: `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4` (revision `4f0cf9d`)
- Reasoning parser: `--reasoning_parser nano-v3`
- Tool parser: `--tool_parser qwen3_coder`
- Backend: pytorch
- GPU: NVIDIA RTX PRO 6000 Blackwell (96 GB, SM120)
- Inference cmd: `trtllm-serve serve <model> --backend pytorch --reasoning_parser nano-v3 --tool_parser qwen3_coder --enable_chunked_prefill --trust_remote_code ...`

## Description

`NemotronV3ReasoningParser` (registered as `nano-v3` in `tensorrt_llm/llmapi/reasoning_parser.py`) returns empty `content` when a request is sent with `chat_template_kwargs: {"enable_thinking": false}` AND the model leaks tokens into the reasoning stream OR fails to emit a closing `</think>` tag.

The parser inherits `DeepSeekR1Parser`'s default behavior of routing tokens to `reasoning_content` until a `</think>` is seen. In thinking-disabled mode the model shouldn't be producing reasoning at all — but in practice it occasionally does (model behavior, not parser behavior). The current parser then returns:

- `content`: empty string
- `reasoning_content`: the actual answer

OpenAI-compatible chat clients render only `content`, so the user sees a blank response.

## Why this is a bug, not a feature

`NemotronV3ReasoningParser` already has a `force_nonempty_content` mechanism that does exactly the right thing — but it requires the caller to opt in via `chat_template_kwargs.force_nonempty_content=True`. That's a fine power-user toggle, but it doesn't match the documented intent of `enable_thinking=False`.

For comparison, the HuggingFace model card for `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4` ships a vLLM reasoning parser (`super_v3_reasoning_parser.py`, revision `4f0cf9d`) that triggers the same swap when EITHER `force_nonempty_content=True` OR `enable_thinking=False`. So the TRT-LLM parser is out of sync with the model's documented inference behavior.

## Reproduction (occurs ~5-15% of the time in our environment)

```bash
# Spin up trtllm-serve with Nemotron-3-Super-120B-NVFP4 + --reasoning_parser nano-v3
for i in $(seq 1 20); do
  curl -ksf https://localhost:8443/v1/chat/completions \
    -d '{"model":"nemotron-super-120b-nvfp4",
         "messages":[{"role":"user","content":"What is 17 * 23?"}],
         "max_tokens":80,
         "chat_template_kwargs":{"enable_thinking":false}}' \
    | jq -r '.choices[0].message.content' &
done
wait
# Count empty responses
```

Without the fix, 1-3 of 20 responses come back with empty `content` while `reasoning_content` has the answer. With the proposed fix, 20/20 return content.

## Proposed fix

Extend the swap-gate in `NemotronV3ReasoningParser` so it ALSO fires when `chat_template_kwargs.enable_thinking is False`. Same semantics as `force_nonempty_content=True`, just gated on an additional condition.

Four touch points, all inside the same class in `tensorrt_llm/llmapi/reasoning_parser.py`:

1. `__init__`: read `enable_thinking` flag, store as `self._enable_thinking_is_false`
2. `_maybe_swap_content`: extend gate to `(self._force_nonempty_content or self._enable_thinking_is_false)`
3. `finish`: same gate extension on the missing-closing-tag branch
4. `parse_delta`: same gate extension on the accumulator

Behavior when thinking is ENABLED — unchanged.
Behavior when `force_nonempty_content=True` — unchanged.
The new branch only fires when `enable_thinking=False` AND the parser would otherwise return empty content.

47-line unified diff (no incidental edits — strictly within `NemotronV3ReasoningParser`) ready to submit. Happy to open a PR with unit tests as soon as this issue is approved.

## Worth considering during review

- Should the fix also extend to `KimiK2ReasoningParser` (also inherits `DeepSeekR1Parser`)? Kimi K2 doesn't expose an `enable_thinking` chat-template flag today, so unclear if it has the equivalent problem — happy to investigate as a follow-up.
- Alternative API: instead of adding a new instance var, the constructor could just default `self._force_nonempty_content = True` when `enable_thinking is False`. Less code, same effect, but `force_nonempty_content` then becomes overloaded with two meanings. Open to either direction.

## Environment evidence

Patched our local install of `tensorrt_llm/llmapi/reasoning_parser.py` with the proposed fix on 2026-05-21. Smoke verified: `PING`/`PONG` and a 17×23 prompt both return non-empty content with `enable_thinking=False`. Behavior with `enable_thinking=True` unchanged (clean `<think>...</think>` framing, correct math answer). No regression in `force_nonempty_content=True` path.

zentradev-rabih:
## Description

Fixes #14502.

`NemotronV3ReasoningParser` returns empty `content` when the caller asks for `enable_thinking=False` but the model still emits reasoning tokens (either a full `<think>…</think>` block, or unterminated reasoning that the stream-end has to flush). OpenAI-compatible chat clients render only `content`, so the user sees a blank response.

The existing `force_nonempty_content=True` knob already does the right swap — this change extends that gate to also fire when `chat_template_kwargs.enable_thinking is False`. That matches the HuggingFace `SuperV3ReasoningParser` (the one shipped on the model card for `nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-NVFP4`), which gates its swap on either condition.

### What the change does

Four touch points, all inside `NemotronV3ReasoningParser` in `tensorrt_llm/llmapi/reasoning_parser.py`:

1. `__init__` — read `enable_thinking` from `chat_template_kwargs`, store `self._enable_thinking_is_false`
2. `_maybe_swap_content` — extend the swap gate to `(self._force_nonempty_content or self._enable_thinking_is_false)`
3. `parse_delta` — extend the accumulator gate (both the explicit-tool-call branch and the general-streaming branch) so reasoning deltas accumulate when `enable_thinking=False`
4. `finish` — extend the no-closing-tag swap gate, so a stream that ends mid-reasoning still produces non-empty content

The swap only fires when content is empty (or whitespace-only, per the existing NVBug 6060281 behavior). A model that emits real content after a leaked think-block keeps the original split.

### Behavior matrix

| Caller setting | Model behavior | Before this PR | After this PR |
|---|---|---|---|
| `enable_thinking=True` (default) | normal `<think>…</think>` | content / reasoning split | unchanged |
| `enable_thinking=False` | no `<think>` tag | `content` = full text | unchanged |
| `enable_thinking=False` | leaks `<think>X</think>` | `content=""`, `reasoning_content="X"` | `content="X"`, `reasoning_content=""` |
| `enable_thinking=False` | leaks `<think>X` (stream ends mid-reasoning) | `content=""`, `reasoning_content="X"` | `content="X"`, `reasoning_content=""` |
| `enable_thinking=False` | `<think>leak</think> real` | `content=" real"`, `reasoning_content="leak"` | unchanged (swap correctly does NOT fire) |
| `force_nonempty_content=True` | any | swap fires (existing) | unchanged (existing) |

## Test Coverage

`tests/unittest/llmapi/test_reasoning_parser.py`:

- **`test_nano_v3_reasoning_parser`** — 4 new parametrized cases covering the non-streaming swap for `enable_thinking=False` (complete `<think>…</think>`, unterminated `<think>…`, longer body, and the "don't over-fire when content already exists" guard)
- **`test_nano_v3_reasoning_parser_finish`** — 1 existing case (`[<think>a, b]` + `enable_thinking=False`) updated from the pre-fix expectation of empty content to `"ab"` (that case was documenting the bug); 1 new multi-delta case (`[<think>answer, " is", " 391"]` → `"answer is 391"`)

All existing `force_nonempty_content` paths and tool-call interactions remain unchanged.

## PR Checklist

## Related

- Issue #14502 — original bug report (HF SuperV3ReasoningParser comparison, repro script)
- NVBug 6060281 — `_maybe_swap_content` whitespace-only handling; preserved
- NVBug 6082303 — `<tool_call>` as implicit end-of-reasoning; extended same gate

### DIFF
--- a/tensorrt_llm/llmapi/reasoning_parser.py
+++ b/tensorrt_llm/llmapi/reasoning_parser.py
@@ -308,12 +308,31 @@ class NemotronV3ReasoningParser(DeepSeekR1Parser):
                  *,
                  reasoning_at_start: bool = True,
                  chat_template_kwargs: Optional[dict[str, Any]] = None) -> None:
+        """Initialize the parser.
+
+        Two boolean flags are derived from ``chat_template_kwargs`` and gate
+        the "swap leaked reasoning into content" behavior in
+        ``_maybe_swap_content`` / ``parse_delta`` / ``finish``:
+
+        - ``_force_nonempty_content``: set when the caller passed
+          ``force_nonempty_content=True``. Power-user opt-in that forces the
+          swap whenever ``content`` is empty.
+        - ``_enable_thinking_is_false``: set when the caller passed
+          ``enable_thinking=False``. Triggers the same swap so OpenAI-style
+          clients (which render only ``content``) see a non-empty response
+          when the model leaks reasoning despite the thinking-off request.
+          Matches the HuggingFace ``SuperV3ReasoningParser`` shipped with
+          the model.
+        """
         self._force_nonempty_content = False
+        self._enable_thinking_is_false = False
         if isinstance(chat_template_kwargs, dict):
             reasoning_at_start = chat_template_kwargs.get(
                 "enable_thinking", reasoning_at_start)
             self._force_nonempty_content = chat_template_kwargs.get(
                 "force_nonempty_content", False) is True
+            self._enable_thinking_is_false = (
+                chat_template_kwargs.get("enable_thinking") is False)
         super().__init__(reasoning_at_start=reasoning_at_start,
                          chat_template_kwargs=chat_template_kwargs)
         self._tool_call_start = "<tool_call>"
@@ -326,14 +345,16 @@ class NemotronV3ReasoningParser(DeepSeekR1Parser):
 
     def _maybe_swap_content(
             self, result: ReasoningParserResult) -> ReasoningParserResult:
-        """When force_nonempty_content is set and content is empty, move
-        reasoning_content into content so the response always has content.
+        """When force_nonempty_content is set, or the caller requested
+        ``enable_thinking=False`` and the model leaked into reasoning anyway,
+        move reasoning_content into content so the response always has
+        content.
 
         Whitespace-only content (e.g. a newline after the closing think tag) is
         treated as empty so the swap still runs (NVBug 6060281)."""
         content = result.content or ""
-        if self._force_nonempty_content and not content.strip(
-        ) and result.reasoning_content:
+        if ((self._force_nonempty_content or self._enable_thinking_is_false)
+                and not content.strip() and result.reasoning_content):
             return ReasoningParserResult(content=result.reasoning_content,
                                          reasoning_content="")
         return result
@@ -357,7 +378,7 @@ class NemotronV3ReasoningParser(DeepSeekR1Parser):
             tool_idx = delta_text.find(self._tool_call_start)
             reasoning = remaining + delta_text[:tool_idx]
             content = delta_text[tool_idx:]
-            if self._force_nonempty_content:
+            if self._force_nonempty_content or self._enable_thinking_is_false:
                 self._found_closing_tag = True
                 self._accumulated_reasoning = ""
             return ReasoningParserResult(content=content,
@@ -365,7 +386,7 @@ class NemotronV3ReasoningParser(DeepSeekR1Parser):
 
         was_in_reasoning = self.in_reasoning
         result = super().parse_delta(delta_text)
-        if self._force_nonempty_content:
+        if self._force_nonempty_content or self._enable_thinking_is_false:
             if result.reasoning_content:
                 self._accumulated_reasoning += result.reasoning_content
             if was_in_reasoning and not self.in_reasoning:
@@ -387,7 +408,7 @@ class NemotronV3ReasoningParser(DeepSeekR1Parser):
         if self.in_reasoning and not self._found_closing_tag:
             remaining = self._buffer
             self._buffer = ""
-            if self._force_nonempty_content:
+            if self._force_nonempty_content or self._enable_thinking_is_false:
                 all_content = self._accumulated_reasoning + remaining
                 self._accumulated_reasoning = ""
                 self.in_reasoning = False
--- a/tests/unittest/llmapi/test_reasoning_parser.py
+++ b/tests/unittest/llmapi/test_reasoning_parser.py
@@ -349,6 +349,29 @@ TOOL_CALL_END = "</tool_call>"
          f"{TOOL_CALL}get_weather{TOOL_CALL_END}", "reasoning", {
              "enable_thinking": False
          }),
+        # GH issue #14502: when the caller asked for enable_thinking=False but
+        # the model leaks a complete <think>...</think> block anyway, the
+        # accidental reasoning should be swapped into content so OpenAI-style
+        # clients (which render only `content`) don't see an empty response.
+        (f"{R1_START}17 * 23 = 391{R1_END}", "17 * 23 = 391", "", {
+            "enable_thinking": False
+        }),
+        # Same fix, with a non-trivial reasoning body.
+        (f"{R1_START}some reasoning here{R1_END}", "some reasoning here", "", {
+            "enable_thinking": False
+        }),
+        # If the model leaks <think> but never emits </think>, partition leaves
+        # an empty `content` and the full body in `reasoning_content` — the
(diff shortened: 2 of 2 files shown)


zentradev-rabih:
Quick note for whoever triages — addressed the two `PR Checks` failures from the auto-checks above:

- **PR title:** updated to `[#14502][fix] nano-v3: swap leaked reasoning into content when enable_thinking=False` (matches the `[ticket][type] Summary` regex in `.github/workflows/pr-check.yml`)
- **PR body:** added the full template structure (Description / Test Coverage / PR Checklist) with the checklist item checked, so `pr_checklist_check.py` is happy
- **DCO:** sign-off added on the previous push (already green)

The new workflow runs against commit `9c4f7f9` are sitting in `action_required` — would appreciate an approve-and-run when convenient. Happy to address any review feedback that surfaces.

tensorrt_llm/llmapi/reasoning_parser.py
```diff
@@ -308,12 +308,31 @@ def __init__(self,
...
+          swap whenever ``content`` is empty.
+        - ``_enable_thinking_is_false``: set when the caller passed
+          ``enable_thinking=False``. Triggers the same swap so OpenAI-style
+          clients (which render only ``content``) see a non-empty response
+          when the model leaks reasoning despite the thinking-off request.
+          Matches the HuggingFace ``SuperV3ReasoningParser`` shipped with
+          the model.
+        """
         self._force_nonempty_content = False
+        self._enable_thinking_is_false = False
```
JunyiXu-nv, line 328: `_enable_thinking_is_false` is not very good naming. Maybe change to:


```suggestion
        self._thinking_disabled = False
```

or just get `_enable_thinking` here and use `not self._enable_thinking` in the following if condition.

tests/unittest/llmapi/test_reasoning_parser.py
```diff
@@ -349,6 +349,29 @@ def test_laguna_reasoning_parser_stream(delta_texts: list, content: list,
...
+        # the model leaks a complete <think>...</think> block anyway, the
+        # accidental reasoning should be swapped into content so OpenAI-style
+        # clients (which render only `content`) don't see an empty response.
+        (f"{R1_START}17 * 23 = 391{R1_END}", "17 * 23 = 391", "", {
+            "enable_thinking": False
+        }),
+        # Same fix, with a non-trivial reasoning body.
+        (f"{R1_START}some reasoning here{R1_END}", "some reasoning here", "", {
+            "enable_thinking": False
+        }),
```
JunyiXu-nv, line 362: This case looks the same as the first one.

tests/unittest/llmapi/test_reasoning_parser.py
JunyiXu-nv, line 489: Could we add the cases for streaming?

JunyiXu-nv (approved):
Overall LGTM. Left some comments. Thanks for contributing!

zentradev-rabih:
Gentle nudge for a maintainer on @NVIDIA/trt-llm-llmapi-devs (CODEOWNER for `tensorrt_llm/llmapi/`):

This PR is approved by @JunyiXu-nv, all pre-merge checks are green (Title Format, Checklist Resolution, DCO, Pre-commit, Base Freshness, API-compat label), and it's currently `MERGEABLE` with no conflicts. It's a small, well-tested fix (one parser file + unit tests) that brings `NemotronV3ReasoningParser` in line with the HF `SuperV3ReasoningParser` behavior for `enable_thinking=False`.

It looks like it's only waiting on (1) a CODEOWNER review and (2) a `/bot run` CI trigger, neither of which a community contributor can self-serve. Could someone kick the pipeline and take a look when you have a moment? Happy to rebase if base has moved. Thanks!

karljang:
@zentradev-rabih, thank you for your contribution. 
Yes, @JunyiXu-nv approved it but also made a few comments. Could you please address and resolve those comments as well?

PR #14591, opened 2026-05-26
[TRTLLM-12214][perf] DeepGemmFusedMoE: skip redundant data expand via fused expand+quant Triton kernel

xwang233:
# [TRTLLM-12214][perf] DeepGemmFusedMoE: skip redundant data expand via fused expand+quant Triton kernel

## Summary

The pre-GEMM pipeline in `DeepGemmFusedMoE` was:

```
moe_permute_op  (writes expanded buffer: ~3.5 MB at bs=32 × topk=8 × hidden=7168 × bf16)
    └── expandInputRowsKernel
  → _masked_index_copy_group_quant_fp8  (reads expanded buffer, writes FP8 + scales)
```

The expanded buffer is a `topk`-x materialization of the source activations that is written by the C++ `expandInputRowsKernel` and re-read one step later by the quant kernel. **Both reads and writes go through HBM.**

This PR:

1. **Adds a Triton kernel `fused_expand_group_quant_fp8`** that re-derives the source row analytically from the permutation map
   ```
   source_row = perm_to_unperm[i] % num_source_tokens
   ```
   (the `expandInputRows` encoding has `k_rank` as the slow index, not `token_id` — handled explicitly) and applies the same group-wise FP8 quant `exp2(ceil(log2(absmax/fp8_max)))`.

2. **Adds a `skip_data_expand: bool = False` parameter** to `moe_permute_op` / `runPermute<T>`. When `True`, the C++ launcher computes the permutation maps but skips the `expandInputRowsKernel` call (the data-copy step) entirely.

3. **Wires `DeepGemmFusedMoE`** to call `moe_permute_op(skip_data_expand=True)` plus the new fused kernel — the materialized expanded buffer is eliminated from the DeepGemm path.

`skip_data_expand` defaults to `False`, so cutlass MoE and any other consumers of `moe_permute_op` are unaffected.

## Why this PR is a "C++ flag PR with a supporting Triton refactor"

Per-call cost of the **Triton half** is essentially unchanged after the refactor — the fused kernel does the same per-element work, it just reads from the compact source via the permutation map instead of from a pre-expanded buffer. **The win is the eliminated `expandInputRowsKernel` launch** (the entire data-copy step) and the bandwidth of the expanded buffer.

## Microbenchmark (B200, sm_100)

`triton.testing.do_bench(warmup=50, rep=300)` over 40 cells: `num_source_tokens ∈ {1, 4, 16, 32, 64} × topK ∈ {4, 8} × experts ∈ {64, 128} × hidden ∈ {4096, 7168}`.

### Triton half — fused vs un-fused quant kernel

The fused kernel is the same per-element work; the refactor is what enables the C++ skip.

| Shape | Un-fused (µs) | Fused (µs) | Δ |
|---|---:|---:|---:|
| `N=1,  K=4, E=64,  H=4096` | 9.22 | 9.12 | −1.0% |
| **anchor** `N=32, K=8, E=128, H=4096` | 11.10 | 10.34 | **−6.9%** |
| `N=64, K=8, E=128, H=7168` | 15.26 | 13.87 | −9.1% |

| Aggregate (40 cells) | Count |
|---|---:|
| Regress (>+2%) | 2 / 40 |
| Win (<−2%) | 6 / 40 |
| Neutral | 32 / 40 |
| Median Δ | −0.33% |

### C++ half — `moe_permute_op` default vs `skip_data_expand=True`

This is where the win lives.

| Shape | Baseline (µs) | `skip_data_expand=True` (µs) | Δ |
|---|---:|---:|---:|
| `N=1,  K=4, E=64,  H=4096` |  9.31 |  6.18 | **−33.7%** |
| **anchor** `N=32, K=8, E=128, H=4096` | 11.26 |  8.16 | **−27.6%** |
| `N=64, K=8, E=128, H=7168` | 13.31 |  8.16 | **−38.7%** |

| Aggregate (40 cells) | Count |
|---|---:|
| Regress (>+2%) | **0 / 40** |
| Win (<−2%) | **40 / 40** |
| Neutral | 0 / 40 |
| Median Δ | **−36.0%** |
| Worst Δ | −23.4% |
| Best Δ | −46.6% |

**Mechanism**: HBM traffic for the pre-GEMM pair drops from ~9.2 MB → ~2.2 MB (4.2× reduction at the model anchor shape) — the compact source fits in L2; the expanded buffer does not.

## Full-bench context

Qwen3-235B-A22B-FP8 + EAGLE3 dyntree, 4×GB200 NVL72, TP=4, EP=1, bs=32, mtbench-32, output_tokens=512, `--ignore_eos`, all PR-1..PR-3 stacked:

| | Baseline | Patched | Δ |
|---|---:|---:|---:|
| Replay median (µs) | 35 887 | 32 587 | **−9.2%** |
| End-to-end TPS | 1 447 | 1 583 | **+9.4%** |
| Acceptance rate | 0.555 | +0.0038 | neutral |
| Output length | 1465 | −0.0031 | neutral |

## Risk callouts

- **`skip_data_expand` is a behaviour change** in `moe_permute_op`. Default is `False`, so cutlass MoE and every other `TORCH_LIBRARY_FRAGMENT` consumer is unaffected.
- The fused Triton kernel is wired **only** into the `DeepGemmFusedMoE` path. The DeepGemm path no longer consumes `permuted_data_tensor` on the pre-GEMM side, so the buffer is safe to skip there.
- **Bit-exact** — verified on 7 shape configurations including the actual model shape. The off-by-one trap on the `expandInputRows` index encoding (`unpermuted_idx = k_rank × num_tokens + token_id`, i.e. `k_rank` is the *slow* index) is handled explicitly in the kernel.

## Files changed

- `cpp/tensorrt_llm/thop/moeUtilOp.cpp` — add `skip_data_expand` arg through `runPermute<T>` and `moe_permute_op`; default `False`; update `TORCH_LIBRARY_FRAGMENT` signature.
- `tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py` — add `skip_data_expand: bool = False` to the fake op signature.
- `tensorrt_llm/_torch/modules/fused_moe/fused_moe_deepgemm.py` — new `_fused_expand_group_quant_fp8` JIT kernel + `fused_expand_group_quant_fp8` launcher; replace the pre-GEMM call site.
- `tensorrt_llm/_torch/modules/fused_moe/ops/moe_op_deepgemm.py` — import + call site swap.

## Test plan

- [x] DeepGemm MoE unit tests for the new fused kernel (bit-exact vs old pair at the 7 shape configs)
- [x] Qwen3-235B-A22B-FP8 + DeepSeek-V3 integration accuracy check (no regression on AR or OSL)

## Release Notes

* **New Features**
  * Added fused expand + group-quantize FP8 kernel for improved MoE performance
  * Introduced `skip_data_expand` option to optimize permutation operations

* **Tests**
  * Added comprehensive unit tests validating fused expand + quantize correctness across multiple configurations

* **Chores**
  * Updated test matrices to include new MoE fused operations tests


liji-nv (approved):

tensorrt_llm/_torch/modules/fused_moe/fused_moe_deepgemm.py
```diff
@@ -795,6 +987,7 @@ def run_moe(
             cluster_rank=self.cluster_rank,
             min_latency_mode=False,
             use_fp8_block_scaling=True,
+            skip_data_expand=True,
```
leslie-fang25, line 1002: Based on my understanding, enabling this flag leaves certain returned tensors, such as `permuted_data_tensor` and `permuted_token_final_scales_tensor`, empty. Could we add a comment to prevent them from being used mistakenly later on?

tensorrt_llm/_torch/modules/fused_moe/fused_moe_deepgemm.py
```diff
@@ -821,12 +1014,14 @@ def run_moe(
                                    self.expert_size_per_partition,
                                    scale_k_padded // 4, m_padded)
 
-        act_input_sf = masked_index_copy_group_quant_fp8(
+        act_input_sf = fused_expand_group_quant_fp8(
             act_input_fp8,
             act_input_sf,
-            permuted_data_tensor,
```
leslie-fang25, line 827: Can we avoid using `permuted_data_tensor` elsewhere, since its value is left uninitialized after this optimization?

leslie-fang25 (approved):
Some small comments. Others LGTM.

tensorrt_llm/_torch/modules/fused_moe/fused_moe_deepgemm.py
xwang233: added comments and replaced the outputs with unused `_`

tensorrt_llm/_torch/modules/fused_moe/fused_moe_deepgemm.py
xwang233: replaced the usage of that with `num_permuted_tokens = permuted_row_to_unpermuted_row_tensor.shape[0]`

xwang233:
/bot skip --comment "test failures are irrelevant, all of our new tests passed"

PR #14279, opened 2026-05-19
[None][refactor] Add derived properties for the thop.attention call site

yuxianq:
## Summary by CodeRabbit

* **New Features**
  * Enhanced attention backend with improved sparse attention support and optimized metadata handling for better performance.
  * Added configuration properties for RoPE embeddings and skip-softmax thresholds.

* **Tests**
  * Added comprehensive sync validation tests ensuring attention kernel bindings remain consistent with the Python call-site implementation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14279?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Description

Convert the inline `thop.attention(...)` call in `TrtllmAttention._run` from a ~75-arg positional call into an **explicit-kwarg** call where every kwarg is sourced as `source.attribute` from one of the rich objects (`self` / `metadata` / `forward_args` / `sparse_args`). Section comments at the call site mirror the rich-object boundaries.

This is a foundation PR — additive and behavior-preserving — extracted from a larger ongoing refactor of the attention dispatch path. It does **not** extract a `_call_thop_attention` wrapper method (that comes in a follow-up); the explicit-kwarg call remains inline in `_run` with section-comment organization.

#### 1. `interface.py` — properties + dataclass + constants

- `AttentionForwardArgs`:
  - `mask_type` `@property` — replaces the legacy `TrtllmAttention._get_mask_type`.
  - `effective_out_scale` `@property` — replaces the inline `out_scale_sf if output_sf is not None else out_scale` ternary.
  - `mrope_rotary_cos_sin` / `mrope_position_deltas` `@property` — pull from the `mrope_config` dict.
  - Drop the dead `enable_attn_nvfp4_output` field (no readers).
- New `AttentionSparseArgs` dataclass for the sparse-attention args bag (`sparse_kv_indices`, `sparse_kv_offsets`, `sparse_attn_indices`, `sparse_attn_offsets`, `sparse_attn_indices_block_size`).
- Module constants `_THOP_EXCLUDED_FIELDS` and `_THOP_LITERAL_NONE` for the sync test.

#### 2. `trtllm.py` — metadata field, ~13 @property accessors, call-site rewrite

[...]

PR #14618, opened 2026-05-27
[None][feat] Default on FlashInferTrtllmGenAttention

yihwang-nv:
## Description

Built on top of #14279 (`[None][refactor] Add derived properties for the thop.attention call site`). That PR refactors the `thop.attention(...)` call site to use explicit kwargs sourced from rich objects; this PR stacks the following on top of it:

### Default on `FlashInferTrtllmGenAttention`

Switches `TrtllmGenAttention` from an opt-in (env-gated) to the default decode backend on Blackwell+, replacing the previous default with `FlashInferTrtllmGenAttention`.

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/trtllm.py
+++ b/tensorrt_llm/_torch/attention_backend/trtllm.py
@@ -1522,20 +1522,19 @@ class TrtllmAttention(AttentionBackend[TrtllmAttentionMetadata]):
             trtllm_gen_backend = self._get_trtllm_gen_backend()
             use_trtllm_gen = trtllm_gen_backend.is_supported(
                 q,
-                metadata=metadata,
-                forward_args=forward_args,
-                mask_type=int(forward_args.mask_type),
-                active_helix=helix_active,
-                use_sage_attn=use_sage_attn,
+                attn=self,
+                meta=metadata,
+                fwd=forward_args,
             )[0]
 
         if use_trtllm_gen:
-            trtllm_gen_backend.attention(
+            trtllm_gen_backend.forward(
                 q,
-                metadata=metadata,
-                forward_args=forward_args,
-                mask_type=int(forward_args.mask_type),
-                use_paged_context_fmha=metadata.use_paged_context_fmha,
+                k,
+                v,
+                attn=self,
+                meta=metadata,
+                fwd=forward_args,
             )
         else:
             # Every kwarg sources from ``self`` / ``metadata`` /
--- a/tensorrt_llm/_torch/attention_backend/trtllm.py
+++ b/tensorrt_llm/_torch/attention_backend/trtllm.py
@@ -29,7 +29,7 @@ from .interface import (AttentionBackend, AttentionForwardArgs,
 # Enable TRTLLM-Gen attention backend by default. Set
 # TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION=0 to force the thop.attention path.
 _TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION = (os.environ.get(
-    "TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION", "0") == "1")
+    "TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION", "1") == "1")
 
 # ``AttentionForwardArgs`` fields that this backend does not consume.
 # Sync test (test_attention_op_sync.py) requires every other field to map to a
@@ -1511,11 +1511,6 @@ class TrtllmAttention(AttentionBackend[TrtllmAttentionMetadata]):
             assert metadata.num_generations == 0
             assert metadata.kv_cache_manager is None
             assert metadata.num_contexts == metadata.num_seqs
-
-        helix_active = metadata.helix_position_offsets is not None
-        use_sage_attn = (forward_args.sage_attn_num_elts_per_blk_q > 0
-                         or forward_args.sage_attn_num_elts_per_blk_k > 0
-                         or forward_args.sage_attn_num_elts_per_blk_v > 0)
 
         use_trtllm_gen = False
         if _TRTLLM_ENABLE_TRTLLM_GEN_ATTENTION:
--- a/tensorrt_llm/_torch/attention_backend/trtllm.py
+++ b/tensorrt_llm/_torch/attention_backend/trtllm.py
@@ -1517,6 +1517,8 @@ class TrtllmAttention(AttentionBackend[TrtllmAttentionMetadata]):
             trtllm_gen_backend = self._get_trtllm_gen_backend()
             use_trtllm_gen = trtllm_gen_backend.is_supported(
                 q,
+                k,
+                v,
                 attn=self,
                 meta=metadata,
                 fwd=forward_args,
--- a/tensorrt_llm/_torch/attention_backend/trtllm_gen.py
+++ b/tensorrt_llm/_torch/attention_backend/trtllm_gen.py
@@ -615,6 +615,11 @@ class FlashInferTrtllmGenAttention:
             return False, "trtllm-gen does not support sparse attention."
         if has_skip_softmax:
             return False, "trtllm-gen does not support skip-softmax attention."
+        if meta.use_spec_decoding and meta.is_spec_dec_tree:
+            return (
+                False,
+                "FlashInfer trtllm-gen does not support spec-dec tree/custom masks.",
+            )
         if is_mla_enable and fwd.attention_input_type != AttentionInputType.generation_only:
             return False, "trtllm-gen MLA supports generation-only attention."
 
--- a/tensorrt_llm/_torch/attention_backend/trtllm_gen.py
+++ b/tensorrt_llm/_torch/attention_backend/trtllm_gen.py
@@ -442,8 +442,6 @@ class FlashInferTrtllmGenAttention:
         self._support_results: dict[tuple[object, ...], Tuple[bool, str]] = {}
         # Lazily set on the first forward() call from the query device.
         self._multi_processor_count: Optional[int] = None
-        self._multi_ctas_kv_counter_workspace_ptr: Optional[int] = None
-        self._multi_ctas_kv_counter_size: int = 0
 
     @property
     def layout(self) -> str:
@@ -776,33 +774,6 @@ class FlashInferTrtllmGenAttention:
         if device_index is None:
             device_index = torch.cuda.current_device()
         return self._get_multi_processor_count_for_device(device_index)
-
-    def _clear_multi_ctas_kv_counter_workspace_if_needed(
-        self,
-        fmha_workspace: torch.Tensor,
-        num_heads: int,
-        max_num_requests: int,
-    ) -> None:
-        counter_size = _get_multi_ctas_kv_counter_size(
-            num_heads,
-            max_num_requests,
-            self._multi_processor_count,
-        )
-        workspace_ptr = fmha_workspace.data_ptr()
-        if (
-            workspace_ptr == self._multi_ctas_kv_counter_workspace_ptr
-            and counter_size <= self._multi_ctas_kv_counter_size
-        ):
-            return
-
-        _clear_multi_ctas_kv_counter_workspace(
-            fmha_workspace,
-            num_heads,
-            max_num_requests,
-            self._multi_processor_count,
-        )
-        self._multi_ctas_kv_counter_workspace_ptr = workspace_ptr
-        self._multi_ctas_kv_counter_size = counter_size
 
     def forward(
         self,
@@ -1229,10 +1200,20 @@ class FlashInferTrtllmGenAttention:
             True,  # need_build_kv_cache_metadata
         )
 
-        self._clear_multi_ctas_kv_counter_workspace_if_needed(
-            fmha_workspace,
-            attn.num_heads,
-            meta.max_num_requests,
+        # FIXME: Flashinfer trtllm-gen API doesn't support a separate
+        # multi CTAs counter buffer. We have to clear a small buffer
+        # before trtllm_gen_batch_decode_with_kv_cache.
+        #
+        # We must also avoid clearing the workspace only when it is
+        # resized. The warmup phase may have already cached the workspace
+        # pointer; if the capture phase skips the zeroing step, the
+        # CUDA graph will not include the counter initialization. We
+        # have already verified—specifically in the context of the GPTOSS-20B
+        # test graph replay scenario—that this skipping logic is unsafe.
+        #
+        # https://github.com/flashinfer-ai/flashinfer/issues/3433
+        _clear_multi_ctas_kv_counter_workspace(
+            fmha_workspace, attn.num_heads, meta.max_num_requests, self._multi_processor_count
         )
 
         q_len_per_req = None if is_multi_token_gen else params.input_seq_length
--- a/tensorrt_llm/_torch/attention_backend/trtllm_gen.py
+++ b/tensorrt_llm/_torch/attention_backend/trtllm_gen.py
@@ -60,6 +60,16 @@ def _has_sparse_tensor(tensor: torch.Tensor) -> bool:
     return tensor is not None and tensor.numel() > 0
 
 
+def _clear_multi_ctas_kv_counter_workspace(
+    fmha_workspace: torch.Tensor,
+    num_heads: int,
+    max_num_requests: int,
+    multi_processor_count: Optional[int],
+) -> None:
+    counter_size = max(num_heads * max_num_requests, multi_processor_count or 0)
+    fmha_workspace.narrow(0, 0, counter_size).zero_()
+
+
 @lru_cache(maxsize=128)
 def _get_context_workspace_layout(
     dtype: torch.dtype,
@@ -222,6 +232,7 @@ class FmhaParams:
     # Context-only fields
     batch_size: int = 0
     # Generation-only fields
+    max_num_requests: int = 0
     beam_width: int = 1
     num_requests: int = 0
     predicted_tokens_per_seq: int = 1
@@ -754,6 +765,7 @@ class FlashInferTrtllmGenAttention:
             paged_context_fmha=meta.use_paged_context_fmha,
             kv_factor=self._kv_factor,
             total_num_blocks=total_num_blocks,
+            max_num_requests=max_num_requests,
         )
 
         sequence_length = meta.kv_lens_cuda_runtime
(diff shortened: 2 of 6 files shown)


tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1485,20 +1485,19 @@ def _run(
             trtllm_gen_backend = self._get_trtllm_gen_backend()
             use_trtllm_gen = trtllm_gen_backend.is_supported(
                 q,
```
yuxianq, line 1519: Use the same argument list with `forward`, add `k/v` here although we don't use them.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -54,277 +50,24 @@
...
-            if is_mla_enable:
-                supported, reason = cls._check_mla_generation_support(
-                    head_size=head_size,
-                    tokens_per_block=tokens_per_block,
-                    kv_lora_rank=kv_lora_rank,
-                    qk_rope_head_dim=qk_rope_head_dim,
-                )
-                if not supported:
-                    return False, reason
+def _has_sparse_tensor(tensor: torch.Tensor) -> bool:
```
yuxianq: `_optional_tensor_list_value/_has_sparse_tensor/_clear_multi_ctas_kv_counter_workspace` are only used inside one method, can we make them local functions?

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: It seems that `helix_active/use_sage_attn` are unused, can we remove them?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -1198,20 +1149,25 @@ def run_mla_generation(
...
-            seq_lens=params.sequence_lengths,
-            max_seq_len=params.max_past_kv_length,
-            out=params.context_buf.view(batch_beam, q_len_per_req, self._num_heads, kv_lora_rank),
-            bmm1_scale=bmm1_scale,
-            bmm2_scale=1.0,
-            sinks=params.forward.attention_sinks,
-            uses_shared_paged_kv_idx=self.USE_SHARED_PAGED_KV_IDX,
-            enable_pdl=self._enable_pdl,
-            backend="trtllm-gen",
+            query,  # query
```
yuxianq, line 1306: Can we still use kwargs instead of comments?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -1164,19 +1115,19 @@ def run_mla_generation(
...
-            num_kv_heads=self._num_kv_heads,
-            tokens_per_block=params.tokens_per_block,
-            head_dim=self._head_dim,
-            kv_factor=params.kv_factor,
-            total_num_blocks=params.total_num_blocks,
-            kv_cache_quant_mode=params.kv_cache_quant_mode,
-            batch_start=params.seq_offset,
-            batch_size=batch_beam,
-            dtype=params.attention_input.dtype,
+            params.host_kv_cache_pool_pointers,  # host_kv_cache_pool_pointers
```
yuxianq: Can we still use kwargs instead of comments?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -1128,29 +1072,36 @@ def run_generation(
...
-            window_left=window_left,
-            kv_layout=self._layout,
-            sinks=params.forward.attention_sinks,
-            q_len_per_req=q_len_per_req,
-            max_q_len=decode_max_q_len,
-            cum_seq_lens_q=decode_cu_seqlens,
-            uses_shared_paged_kv_idx=self.USE_SHARED_PAGED_KV_IDX,
-            enable_pdl=self._enable_pdl,
-            backend="trtllm-gen",
+            q_processed,  # query
```
yuxianq, line 1234: Can we still use kwargs instead of comments?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -1089,37 +1016,54 @@ def run_generation(
...
-            max_past_kv_length=params.max_past_kv_length,
-            bmm2_scale=1.0,
-            fp8_context_fmha=params.fp8_context_fmha,
-            predicted_tokens_per_seq=params.predicted_tokens_per_seq,
-            multi_processor_count=self._multi_processor_count,
-            total_num_blocks=params.total_num_blocks,
-            kv_factor=params.kv_factor,
-            need_build_kv_cache_metadata=True,
-            **self._static_kw,
+            params.qkv_input,  # qkv_input
```
yuxianq, line 1155: Can we still use kwargs instead of comments? There are many similar refactor, why we refactor them?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -697,64 +645,53 @@ def _get_multi_processor_count(self, device: torch.device) -> int:
...
         self,
         q: torch.Tensor,
-        *,
-        metadata: "TrtllmAttentionMetadata",
-        forward_args: AttentionForwardArgs,
-        mask_type: int,
-        use_paged_context_fmha: bool,
+        k: Optional[torch.Tensor],
+        v: Optional[torch.Tensor],
+        attn: "TrtllmAttention",
```
yuxianq, line 778: We have already pass `attn` to `FlashInferTrtllmGenAttention` in `__init__`, can we remove it from `forward/is_supported` to avoid duplicate?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -810,42 +739,51 @@ def attention(
...
         cyclic_attn_window_size = attention_window_size
-        kv_factor, total_num_blocks = self._get_kv_cache_metadata(metadata, is_mla_enable)
-        params = EnqueueParams(
-            forward=forward_args,
+        tokens_per_block = tokens_per_block if tokens_per_block is not None else 64
+        _, total_num_blocks = self._get_kv_cache_dtype_and_total_blocks(meta, is_mla_enable)
+
+        params = FmhaParams(
             workspace=workspace,
+            kv_scale_orig_quant=fwd.kv_scale_orig_quant,
```
yuxianq: We use FmhaParams to repack many fields from fwd/meta/attn, if we only store new fields created in `FlashInferTrtllmGenAttention` to FmhaParams, will the code be clearer and less host overhead?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -525,18 +304,28 @@ def __init__(
             )
 
         # Cache layer-static properties to avoid repeated attribute lookups
-        # through the weakref on every layer forward call.
+        # on every layer forward call.
         self._num_heads = attention_layer.num_heads
```
yuxianq: We reassign many fields from `attention_layer` to `FlashInferTrtllmGenAttention`, can we just name `attention_layer` as `attn` to keep consistent and store it as `self.attn`, don't reassign any field, will the code be clearer?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: I have tested six more graph-on models with the temporary direct kwarg calls. Every graph-on kwarg row regressed output throughput by 2.48% to 5.66%, with E2EL up 2.54% to 6.00%. That is enough perf evidence to keep positional calls on these hot-path APIs.

image

To make it easier to understand, I've used comments instead of kwargs to achieve the same effect. If you feel it's necessary, I can add comments explaining why we do it this way. WDYT?

tensorrt_llm/_torch/attention_backend/trtllm.py
yihwang-nv: Sure, I'll remove it.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Sure, I'll inline these functions.

chenfeiz0326:
/bot run  --disable-fail-fast --stage-list "\*GB200\*PerfSanity\*,\*GB300\*PerfSanity\*"

chenfeiz0326:
/bot run  --disable-fail-fast --stage-list "\*GB200\*PerfSanity\*,\*GB300\*PerfSanity\*"

yihwang-nv:
/bot run  --disable-fail-fast --stage-list "\*GB200\*PerfSanity\*,\*GB300\*PerfSanity\*"

yihwang-nv:
/bot run  --disable-fail-fast --stage-list "\*GB200\*PerfSanity\*,\*GB300\*PerfSanity\*"

tensorrt_llm/_torch/models/modeling_gemma3vl.py
```diff
@@ -184,6 +184,14 @@ def forward(self, vision_outputs: torch.Tensor):
     ))
 class Gemma3VLM(PreTrainedModel):
 
+    @classmethod
+    def get_model_defaults(cls, llm_args) -> dict:
+        # Gemma3 VLM image requests need FlashInfer for custom bidirectional
+        # image-token masks. Explicit user overrides still take precedence.
+        return {
+            "attn_backend": "FLASHINFER",
```
yuxianq: Why we need to update this file? Seems that the logic is unchanged.

tests/unittest/_torch/modeling/test_modeling_gemma3.py
```diff
@@ -106,6 +107,23 @@ def __repr__(self) -> str:
 
 class TestGemma3(unittest.TestCase):
 
+    def test_gemma3_vlm_defaults_to_flashinfer(self):
+        defaults = Gemma3VLM.get_model_defaults(None)
+        self.assertEqual(defaults["attn_backend"], "FLASHINFER")
+
+    def test_gemma3_vlm_text_subconfig_keeps_parent_attention_backend(self):
```
yuxianq: Why we need to add tests for Gemma3? Seems out-of-scope.

yihwang-nv: Revert

yihwang-nv: This test changed because I have ever tried to test gemma model performance. But I hit failure, so I made these changes.

tensorrt_llm/_torch/models/modeling_gemma3vl.py
yihwang-nv: Reverted, same as below.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Discuss offline, use kwarg may cause host performance drop.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Same as above

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Same as above

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: I have remove the attn arg in __init__, because we don't need to cache the static layer config when create the backend anymore.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: The static layer config cache was removed.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: I have removed most unnecessary fields in `FmhaParams` (The field will be removed if it's directly come from attn/fwd/meta).

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: same as above

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -945,42 +985,16 @@ def _missing_fused_nanobind_ops() -> List[str]:
...
-        params: EnqueueParams,
-    ):
-        kv_scale_orig_quant, kv_scale_quant_orig = self._get_kv_scale_params(
-            params.forward, params.kv_cache_quant_mode
-        )
-        attention_output_orig_quant = params.forward.out_scale
-        mrope_rotary_cos_sin = params.forward.mrope_rotary_cos_sin
+        params: FmhaParams,
+    ) -> None:
+        attn = params.attn
```
yuxianq, line 989: Should we flatten attn/meta/fwd as top-level arguments instead of packing them into `FmhaParams`?

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -616,75 +469,294 @@ def _get_kv_scale_params(
...
+            fwd.mask_type if check_context_phase else None,
+            meta.beam_width,
+            tokens_per_block,
+            meta.kv_cache_block_offsets is not None,
+            meta.is_spec_decoding_enabled,
         )
-        if result[0]:
-            self._support_result = result
+        if support_key in self._support_results:
+            return self._support_results[support_key]
```
yuxianq: It seems that this `_support_results` cache does not speed up the check. We should move this cache lookup before all `return False` cases to shortcut them. We may also need to cache all false/true cases instead of only true cases.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Hmm, I'd prefer to keep them in `FmhaParam`. The attn/fwd/meta also is a part of param in semantic. WDYT?

yuxianq: Fine for me.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -996,118 +1012,137 @@ def run_context(
...
+            cu_kv_seqlens,  # cum_seq_lens_kv
+            window_left,  # window_left
+            params.context_buf,  # out
+            fwd.attention_sinks,  # sinks
+            self._enable_pdl,  # enable_pdl
+            kv_scale_pool,  # kv_scale_pool
+            self.USE_SHARED_PAGED_KV_IDX,  # uses_shared_paged_kv_idx
         )
 
         thop.trtllm_gen_context_postprocess(
```
yuxianq, line 1090: Should we move pre-process and post-process out of `run_context/run_generation/run_mla_generation` so that if we use FlashInferTrtllmGenAttention as BaseFmha, other Fmha libs can reuse them and only implement fmha kernels in `run_context/run_generation/run_mla_generation`? Only for discussion, not necessary part of this PR.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Fixed.

yuxianq (approved):

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Agree 100%, we can split run_context/run_generation, and only left FMHA kernel call in it.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Removed

wenmingw (approved):

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -1122,110 +1152,136 @@ def run_generation(
...
+        #
+        # We must also avoid clearing the workspace only when it is
+        # resized. The warmup phase may have already cached the workspace
+        # pointer; if the capture phase skips the zeroing step, the
+        # CUDA graph will not include the counter initialization. We
+        # have already verified—specifically in the context of the GPTOSS-20B
+        # test graph replay scenario—that this skipping logic is unsafe.
+        #
+        # https://github.com/flashinfer-ai/flashinfer/issues/3433
+        _clear_multi_ctas_kv_counter_workspace(
```
PerkzZheng, line 1210: we only need to zero it once.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -1236,31 +1292,34 @@ def run_mla_generation(
...
-            out=params.context_buf.view(batch_beam, q_len_per_req, self._num_heads, kv_lora_rank),
-            bmm1_scale=bmm1_scale,
-            bmm2_scale=1.0,
-            sinks=params.forward.attention_sinks,
-            uses_shared_paged_kv_idx=self.USE_SHARED_PAGED_KV_IDX,
-            enable_pdl=self._enable_pdl,
-            backend="trtllm-gen",
+            query,  # query
+            kv_cache,  # kv_cache
+            params.workspace.view(-1, 4),  # workspace_buffer
```
PerkzZheng, line 1308: should we also clean with _clear_multi_ctas_kv_counter_workspace ?

PerkzZheng (approved):

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: Yeah, zero once is our expectation. Current implementation is a workaround.

Flashinfer's API multi CTAs kv counter buffer combine with workspace in one buffer. Ideally, it should be reset to zero whenever the workspace is resized. But I encountered an issue when I run GPTOSS 20B model locally, the cuda graph capture seem miss a zero in generation phase, and it's cause IMA.

So I zero the small buffer every time. It does not have a significant impact on performance.

I have submitted an issue for Flashinfer team: https://github.com/flashinfer-ai/flashinfer/issues/3433

juney-nvidia (approved):

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
yihwang-nv: That's a potential issue, but wired, it's not caused any failures in CI. I'll fix it in a separate PR.

PR #14636, opened 2026-05-27
[DEP-950][feat] add multi-rank sleep/wakeup support to MPI executor path

hhzhang16:
Extends `sleep()` / `wakeup()` to work correctly when `tensor_parallel_size > 1`.

Previously the MPI/IPC path rejected multi-rank sleep/wakeup because `control_action()` only pauses the local rank-0 executor loop; it does not run code in peer rank processes. Since `release_with_tag()` / `materialize_with_tag()` must execute in each rank's own process, this MR adds a dedicated control communicator and listener thread for non-rank-0 workers.

### Control-Listener Infrastructure (`py_executor.py`)

- Duplicates a secondary MPI communicator (`_sleep_wakeup_comm`) in `start_worker()` after PP broadcast-comm setup. This is collective and runs on the main thread before the worker thread starts.
- Spawns a `_sleep_wakeup_listener_thread` daemon on non-rank-0 ranks.
  - The listener blocks on sleep/wakeup control messages.
  - It participates in a `PREPARE` / `COMMIT` / `ABORT` protocol.
  - `PREPARE` quiesces the rank and ACKs readiness without changing VMM state.
  - `COMMIT` executes the requested VMM operation locally.
  - `ABORT` releases or records the matching control request without mutating VMM state.
  - ACKs include `status`, `error`, `op_id`, and protocol phase.
- Adds `_sleep_wakeup_lock` to serialize concurrent callers.
  - `control_action()` uses an Event barrier, not a mutex.
  - Without the lock, two concurrent RPCs can both pass the barrier and interleave on `_sleep_wakeup_comm` or mutate VMM out of order.
  - The lock covers both single-rank and multi-rank sleep/wakeup.
- During shutdown, rank 0 sends a `"shutdown"` sentinel to all non-rank-0 listeners and drains listener ACKs with a bounded nonblocking poll loop.

### Coordination (`base_worker.py`)

- Removes the `world_size > 1` `NotImplementedError` from `_check_sleep_wakeup_preconditions()`.
- Adds `_multi_rank_sleep_wakeup()`:
  - Acquires `_sleep_wakeup_lock`.
  - Enters `control_action(control_id=<op_id>)`.
  - Sends `PREPARE` to non-rank-0 workers.
  - Drains prepare ACKs from peers.
  - Runs the local VMM operation on rank 0 only after peers are prepared.
  - Sends `COMMIT` so peers execute the same VMM operation.
  - Sends `ABORT` if prepare or rank-0 local execution fails.
  - Drains all peer ACKs before raising any aggregate error.
- ACK draining is bounded and filters by `op_id` and phase so stale or wrong-phase ACKs cannot corrupt a later sleep/wakeup operation.
- Updates `sleep()` and `wakeup()` so both branches are serialized:
  - `world_size == 1`
  - `world_size > 1`

### Proxy Allowlist (`proxy.py`, `rpc_proxy.py`)

- Adds `_MULTI_RANK_ALLOWED_METHODS = {"sleep", "wakeup"}`.
- Updates `_check_collective_rpc_guard()` to accept `method=`.
- Multi-rank `collective_rpc()` still rejects general methods, but allows `sleep()` / `wakeup()` because those now coordinate internally across ranks.

### Sleep/Wakeup Sequence

```mermaid
sequenceDiagram
    participant API as API caller
    participant R0 as Rank 0 worker
    participant Loop0 as Rank 0 executor loop
    participant C as _sleep_wakeup_comm
    participant R1 as Rank 1 listener
    participant R2 as Rank 2 listener

    API->>R0: sleep(tags) or wakeup(tags)
    R0->>R0: acquire _sleep_wakeup_lock
    R0->>Loop0: enter control_action(op_id)
    Loop0-->>R0: drained and paused

    R0->>C: PREPARE op_id/action/tags to rank 1
    C->>R1: deliver PREPARE
    R1->>R1: quiesce, no VMM mutation
    R1-->>C: ACK PREPARE

    R0->>C: PREPARE op_id/action/tags to rank 2
    C->>R2: deliver PREPARE
    R2->>R2: quiesce, no VMM mutation
    R2-->>C: ACK PREPARE

    C-->>R0: prepare ACKs

    R0->>R0: release/materialize local VMM

    R0->>C: COMMIT op_id/action/tags to rank 1
    C->>R1: deliver COMMIT
    R1->>R1: release/materialize local VMM
    R1-->>C: ACK COMMIT

    R0->>C: COMMIT op_id/action/tags to rank 2
    C->>R2: deliver COMMIT
    R2->>R2: release/materialize local VMM
    R2-->>C: ACK COMMIT

    C-->>R0: commit ACKs

    R0->>Loop0: exit control_action()
    R0->>R0: release _sleep_wakeup_lock
    R0-->>API: success or aggregate error
```

### Error Handling

```mermaid
flowchart TD
    Start["Rank 0 sends PREPARE"]
    Prepared{"All peers prepared?"}
    AbortPrepare["Send ABORT to peers"]
    DrainPrepare["Drain PREPARE / ABORT ACKs"]
    Local["Rank 0 local VMM op"]
    LocalOk{"Rank 0 local op ok?"}
    AbortLocal["Send ABORT to prepared peers"]
    Commit["Send COMMIT to prepared peers"]
    PeerVmm["Peers run VMM op on COMMIT"]
    DrainCommit["Drain COMMIT ACKs"]
    AnyError{"Any local or peer error?"}
    Raise["Raise aggregate RuntimeError"]
    Return["Return success"]

    Start --> Prepared
    Prepared -->|no| AbortPrepare --> DrainPrepare --> Raise
    Prepared -->|yes| Local
    Local --> LocalOk
    LocalOk -->|no| AbortLocal --> DrainPrepare --> Raise
    LocalOk -->|yes| Commit --> PeerVmm --> DrainCommit
    DrainCommit --> AnyError
    AnyError -->|yes| Raise
    AnyError -->|no| Return
```

The important guarantee is that peers do not mutate VMM state during `PREPARE`. They only run `release_with_tag()` / `materialize_with_tag()` after rank 0 has successfully completed its local VMM operation and sends `COMMIT`. Rank 0 also drains bounded, phase-tagged ACKs before returning or raising, which prevents stale ACKs from being consumed by the next sleep/wakeup operation.

### Tests

- `test_sleep_collective_rpc_guards.py`
  - Replaces removed-contract tests with routing and allowlist tests.
  - Adds `TestMultiRankSleepWakeupLock`.
    - Uses `SpyLock` to verify acquire/release ordering.
    - Adds a concurrent-overlap test.
    - Patches CUDA/VMM globals once in the main thread so both worker threads share the same mock scope.

- `TestMultiRankAckErrorPropagation`
  - Drives `_multi_rank_sleep_wakeup()` directly with fake ACKs.
  - `test_peer_error_ack_raises` verifies a single non-ok ACK surfaces its error detail.
  - `test_multiple_peer_errors_aggregated` uses `world_size=3` with two error ACKs and verifies both messages appear in the aggregate error.

- `TestMultiRankSendFailureRecovery`
  - Covers partial PREPARE send failure.
  - Verifies rank 0 does not run local VMM when prepare fails.
  - Verifies peers receive ABORT and ACKs are drained.
  - Covers bounded ACK draining when an ACK never arrives.
  - Covers COMMIT send failure and verifies the uncommitted rank receives ABORT.

- `TestListenerAbortAndShutdown`
  - Verifies PREPARE does not execute VMM and does not release the control barrier.
  - Verifies ABORT releases an active matching control request.
  - Verifies early ABORT is recorded and later consumed by the matching control request.
  - Verifies shutdown sends an ACK before listener exit.
  - Verifies rank-0 shutdown ACK draining is bounded if listener ACKs never arrive.

- `TestMultiRankRank0LocalFailureDrainsAcks`
  - Patches `release_with_tag()` to raise.
  - Asserts prepare and abort ACKs are still received before the method raises.
  - Covers the combined failure case where rank 0 fails and a peer also returns an error; both messages must appear.

- `TestSingleRankLockAcquired`
  - Uses a minimal `SpyLock` to verify single-rank `sleep()` / `wakeup()` also acquire `_sleep_wakeup_lock`.
  - Parametrized over both methods.
  - The engine stub does not need `_sleep_wakeup_comm` because the single-rank path never touches it.

- `_make_proto_worker`
  - Shared by the ACK/error-path test classes.
  - Defaults to `world_size=3` so tests cover multiple peers instead of only the degenerate single-peer case.

- `test_mpi_sleep_wakeup.py` (new, `@pytest.mark.gpu2`)
  - Adds TP=2 integration coverage using TinyLlama.
  - `_per_device_gpu_memory()` queries NVML per device.
  - This ensures a rank-1 leak on GPU 1 cannot be masked by rank 0 releasing enough memory on GPU 0.

## Release Notes

* **New Features**
  * Sleep/wakeup memory management operations now support multi-rank distributed deployments with coordinated control across all GPU ranks

* **Bug Fixes**
  * Multi-rank configurations can now use sleep/wakeup functionality (previously unavailable for distributed setups)

* **Tests**
  * Added comprehensive test coverage for multi-rank sleep/wakeup operations, including memory monitoring validation

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14636?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -100,6 +100,13 @@ class PPCommTag(IntEnum):
     SAMPLE_STATE = 20003
 
 
+# Tags for the dedicated control communicator used by multi-rank sleep/wakeup.
+# Because _control_comm is a duplicated communicator isolated from all other
+# MPI traffic, small tag values are safe and do not conflict with PPCommTag.
+_CONTROL_ACTION_TAG = 0  # rank-0 -> non-rank-0: sleep/wakeup/shutdown msg
```
chienchunhung: nit: I see many `*control*` variables introduced in this PR. The term `control` appears generic, while the logic for those variables are very specific to the sleep/wake mechanism. I'd suggest rename the variables to be more reflective for better readability.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2036,6 +2082,85 @@ def _broadcast_sample_state_loop(self):
...
+        CUASSERT(cudart.cudaSetDevice(self.device_id))
+        set_thread_local_mpi_comm(self._control_comm)
+        try:
+            while True:
+                msg = self._control_comm.recv(
+                    source=0, tag=_CONTROL_ACTION_TAG)
+                # Check for shutdown before entering the ACK-guarded block so
+                # we can break cleanly without sending a spurious ACK.
+                if msg.get("action") == "shutdown":
+                    break
```
chienchunhung, line 2856: nit: add a log for shutdown

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2036,6 +2082,85 @@ def _broadcast_sample_state_loop(self):
...
+                # Use .get() so a missing "action" key surfaces as an unknown
+                # action inside the try rather than a bare KeyError here.
+                action = msg.get("action", "<unknown>")
```
chienchunhung, line 2859: Create a enum class that defines the legitimate action values: `sleep`, `wakeup`, `unknown`. Use the enum value instead of hardcoding the values.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2036,6 +2084,85 @@ def _broadcast_sample_state_loop(self):
...
+        message breaks the loop cleanly.
+
+        Safety notes:
+        - ``_control_comm`` is a dedicated duplicated MPI communicator, isolated
+          from all regular executor MPI traffic.
+        - ``torch.cuda.synchronize()`` is called before each VMM operation so
+          that in-flight CUDA kernels from the event-loop thread complete before
+          memory mappings are changed.  The event-loop thread is effectively
+          idle at this point (it is blocked in a collective waiting for rank-0,
+          whose event loop is paused via ``control_action()``).
```
chienchunhung: Claude Code raised a concern about potential race:

```
1. Rank-0 drains its loop via control_action().
2. Rank-0 sends the control message via _control_comm.
3. The message arrives at rank-N's listener thread.
4. The listener runs torch.cuda.synchronize() + release_with_tag(*tags).

Between (1) and (3), rank-N's executor loop is becoming-idle because rank-0 isn't issuing collectives. But "becoming idle" isn't "idle." Concrete races:

- Rank-N is mid-iteration when (1) starts. It launches a CUDA kernel that doesn't depend on rank-0 (e.g., quantization scale fusion, weight transform). By the time (3) arrives, that kernel may still be running.
- torch.cuda.synchronize() at line 2132 waits for in-flight kernels — but the executor loop's next iteration could launch new kernels in the window between synchronize() and release_with_tag().
- Worse: if the executor loop is mid-NCCL when (2) arrives, the listener's cuda.synchronize() followed by VMM unmap could remap memory that the in-progress NCCL kernel is still reading.
```

Claude also suggested we stress test this assumption, e.g,. multiple sleep/wakeup iterations while the executor is processing some requests on TP=4+.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -628,6 +635,16 @@ def on_detected():
         self.worker_started = False
         self.worker_lock = threading.Lock()
         self._broadcast_mpi_comm = None
+        # Secondary MPI communicator and listener thread for multi-rank
+        # sleep/wakeup control messages.  Both are None until start_worker()
+        # calls Dup() (a collective) on the main thread.
+        self._control_comm = None
+        self._control_listener_thread: Optional[threading.Thread] = None
```
chienchunhung: We need to make sure this thread on non-rank-0 thread gets properly cleaned up in all (corner) cases, not just for the happy path, i.e., receiving shutdown signal from rank-0. Some examples (but not limited to): rank-0 never sends shutdown, non-rank-0 executor exits before shutdown arrives, ...etc.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2036,6 +2084,85 @@ def _broadcast_sample_state_loop(self):
...
+                        gc.collect()
+                        torch.cuda.empty_cache()
+                    elif action == "wakeup":
+                        materialize_with_tag(*tags)
+                        torch.cuda.synchronize()
+                    else:
+                        error_msg = f"unknown action '{action}'"
+                        logger.warning(
+                            f"Control listener: {error_msg}, ignoring.")
+                except (KeyError, TypeError, ValueError, RuntimeError,
```
chienchunhung, line 2933: Could the silent death of this thread cause the rank-0 to hang indefinitely?

chienchunhung (commented):
Thanks for the PR. Overall approach looks good. I took the first pass and left a couple of comments.

Question: what are the expected behavior if the sleep/wakeup actions succeeded partially, e.g., some ranks succeed while some fail? Do we reconcile?

tensorrt_llm/_torch/pyexecutor/py_executor.py
hhzhang16: Renamed to `*sleep_wakeup*`!

hhzhang16:
> Question: what are the expected behavior if the sleep/wakeup actions succeeded partially, e.g., some ranks succeed while some fail? Do we reconcile?

There's no reconciliation -- this is an intentional design decision matching how the rest of the executor treats unrecoverable errors. If any rank fails its VMM op, then the system is in an inconsistent state due to some ranks having released/materialized their memory and some not. Attempting to reconcile would require another coordinated round trip and could itself fail. Instead, all errors are aggregated and a single `RuntimeError` is raised on rank 0 listing all failing ranks.

tensorrt_llm/_torch/pyexecutor/py_executor.py
hhzhang16: Two mechanisms ensure safety:
1. `control_action()` on rank 0 drains all in-flight requests before proceeding. Rank 0's event loop won't exit `control_action()` until the request queue is empty and the current iteration finishes. Non-rank-0 event loops are driven entirely by NCCL collectives initiated from rank 0 — once rank 0's iteration is done, rank N's event loop blocks waiting for the next NCCL op that never comes, so it's effectively idle.
2. `torch.cuda.synchronize()` on the listener thread flushes any CUDA work that the event-loop thread submitted before it went idle. The listener thread then has exclusive access to the CUDA context from that point.

That said, this reasoning assumes NCCL + the event loop as the only sources of CUDA work on non-rank-0 ranks, which is true for our current architecture but worth stress-testing under heavy concurrent load. Happy to add a stress test if that gives more confidence!

tensorrt_llm/_torch/pyexecutor/py_executor.py
hhzhang16: The thread is daemon=True, so it will be terminated when the main process exits -- this covers the case where we have an unhandled exception in teh worker thread that propagates and kills the process. For cases where the executor shuts down without calling the `shutdown` path, the worker thread joins first then the shutdown message gets sent to each listener. 

There is a gap where the executor object is garbage-collected without shutdown being called and without the process exiting, in which case the listener thread would block on `_sleep_wakeup_comm.recv`. This is a known limitation and in the same class of issue as other background threads in TRT-LLM. I think the right answer for now is to document that the LLM context manager (`with llm:`) or an explicit `llm.shutdown()` is required; we can track this as a follow-up if needed

tensorrt_llm/_torch/pyexecutor/py_executor.py
hhzhang16: The `finally` block always runs even if the `except` clause doesn't catch an exception. So this guarantees that rank 0 gets an error ACK regardless of what the Python exception files. The outer `finally` also runs and it propagates out of the loop and kills the thread. For MPI-level failures, MPI will rainse an error and rank 0's `recv` will raise rather than hang.

chienchunhung: IIUC if there's an exception not caught by the except-list, the `finally` block would still return the ACK but falsely specify `status=ok` since `error_msg is None`. Wouldn't this cause the inconsistency view between rank-0 and others? Also, in this case, the non-zero-rank may silently die. 

It'd be a good idea to make sure the error message always reflect the real state. For example, Claude suggests the following edits:

```python3
finally:
    # If we're unwinding an uncaught exception, the narrow except above
    # missed it but the ACK to rank-0 must still reflect failure.
    if error_msg is None and sys.exc_info()[0] is not None:
        exc = sys.exc_info()[1]
        error_msg = (f"rank {self.dist.rank} '{action}' failed with "
                     f"uncaught {type(exc).__name__}: {exc!r}")
    self._sleep_wakeup_comm.send(
        {"status": "ok" if error_msg is None else "error",
         "error": error_msg},
        dest=0,
        tag=_SLEEP_WAKEUP_ACK_TAG,
    )
```

You can probably widen the except list, depending under what conditions you want to keep it alive.

tensorrt_llm/_torch/pyexecutor/py_executor.py
chienchunhung: Right, I think locking the current invariant with a test would be a good idea for future-proof, so that it guards against the modification that adds non-NCCL CUDA work on rank-N.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -923,6 +964,21 @@ def shutdown(self):
         if self.dist.pp_size > 1:
             self.executed_batch_queue.put(None)
             self.broadcast_sample_state_handler.join()
+        # Signal non-rank-0 sleep/wakeup listener threads to exit.  This runs
+        # after the worker thread has joined, which guarantees that the non-rank-0
+        # executor loops have already processed the shutdown broadcast and are
+        # no longer driving NCCL, so the send cannot deadlock.
+        if (self.dist.rank == 0 and self._sleep_wakeup_comm is not None
+                and self.dist.world_size > 1):
```
chienchunhung: nit: We should probably add a log for the shutdown broadcast.

tests/unittest/executor/test_sleep_collective_rpc_guards.py
```diff
@@ -93,28 +93,422 @@ def test_missing_sleep_config_raises(self, method):
...
+
+    w.engine = SimpleNamespace(
+        _sleep_wakeup_lock=threading.Lock(),
+        _sleep_wakeup_comm=SimpleNamespace(send=lambda *a, **kw: None, recv=fake_recv),
+        control_action=_noop_control_action,
+    )
+    return w, recv_calls
+
+
+class TestMultiRankAckErrorPropagation:
```
chienchunhung, line 370: We should also cover the scenario where an exception is not caught.

tensorrt_llm/_torch/pyexecutor/py_executor.py
hhzhang16: Good catch, added a guard in the finally block and a test

tensorrt_llm/_torch/pyexecutor/py_executor.py
hhzhang16: I can add a stress test to `test_mpi_sleep_wakeup.py` that runs sleep/wakeup in a tight loop while the executor is handling requests under TP=4, and checks for correctness throughout.

tests/unittest/executor/test_sleep_collective_rpc_guards.py
hhzhang16: Added `TestListenerUncaughtExceptionSendsErrorAck`

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -100,6 +100,21 @@ class PPCommTag(IntEnum):
     SAMPLE_STATE = 20003
 
 
+# Tags for the dedicated sleep/wakeup communicator used by multi-rank sleep/wakeup.
+# Because _sleep_wakeup_comm is a duplicated communicator isolated from all other
+# MPI traffic, small tag values are safe and do not conflict with PPCommTag.
+_SLEEP_WAKEUP_ACTION_TAG = 0  # rank-0 -> non-rank-0: sleep/wakeup/shutdown msg
```
Superjomn: How about grouping both values to an Enum?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -100,6 +100,21 @@ class PPCommTag(IntEnum):
...
+_SLEEP_WAKEUP_ACK_TAG = 1  # non-rank-0 -> rank-0: ACK after action completes
+
+
+class _SleepWakeupAction(str, Enum):
```
Superjomn: There is a StrEnum with `from enum import StrEnum` for such case.

chienchunhung (approved):
Overall LGTM. It'd be great to address the following in the follow-up PRs  (1) add the load tests, (2) harden the corner case thread cleanup.

Superjomn (approved):
LGTM

pcastonguay (approved):

chienchunhung:
The real PR-caused failure is the new MPI sleep/wakeup path:
- `unittest/_torch/multi_gpu/test_mpi_sleep_wakeup.py::test_mpi_sleep_wakeup_tp2`
- `unittest/_torch/multi_gpu/test_mpi_sleep_wakeup.py::test_mpi_sleep_wakeup_kv_cache_only_tp2`

Both fail on `DGX_H100-2_GPUs-PyTorch-Others-1` with `RequestError: Hang detected on rank 0 in PyExecutor.`

The hang happens after `sleep()` / `wakeup()`, when the test calls llm.generate() again. Log evidence shows rank 1’s executor loop stuck in `_handle_control_request()` waiting on `control_action_done`, while the new sleep/wakeup listener thread is blocked in `_sleep_wakeup_comm.recv(...)`. That matches the old guard this PR removed: non-rank-0 processes can enter the control-action pause with no matching caller to release their local control_action_done.

So the listener executes/receives sleep-wakeup messages, but it does not solve the executor control barrier on peer ranks. The PR’s rank-0-only `_multi_rank_sleep_wakeup()` enters `engine.control_action()` on rank 0, while peer event loops can still pause and never resume.

chienchunhung:
@hhzhang16 One recommendation: Initialize the sleep/wakeup communicator/listener only when multi-rank sleep/wakeup can actually be used, e.g. PyTorch backend + sleep_config enabled + MPI executor path, rather than all multi-rank PyExecutor workers. 

I believe the current implementation would initiate the listener thread even when Ray orchestration is used, which is not necessary.

hhzhang16:
@chienchunhung good catch, added gate

chienchunhung:
/bot run --disable-fail-fast --stage-list "DGX_B200-4_GPUs-AutoDeploy-1,DGX_H100-4_GPUs-AutoDeploy-1"

chienchunhung:
@hhzhang16 

2 failed tests seem highly correlated to this PR: `test_mxfp4_gsm8k[120b-tp2]`, `test_attention_dp[4]`. Most recent test failures came from this PR. 

After taking a closer look, I suspect the new sleep/wakeup initialization gate assumes `llm_args.sleep_config` always exists. AutoDeploy’s ADEngine appears to provide a minimal SimpleNamespace as llm_args and may not define sleep_config, so multi-GPU AutoDeploy can crash while evaluating the gate before the listener is even launched. 

Suggestion: Consider using `getattr(self.llm_args, "sleep_config", None) is not None` or explicitly setting `sleep_config = None` in the AutoDeploy shim.

chienchunhung:
/bot run --disable-fail-fast --stage-list "DGX_B200-4_GPUs-AutoDeploy-1,DGX_H100-4_GPUs-AutoDeploy-1"

chienchunhung:
/bot run --disable-fail-fast --stage-list "DGX_B200-4_GPUs-AutoDeploy-1,DGX_H100-4_GPUs-AutoDeploy-1"

chienchunhung (changes requested):
A couple of things flagged by Codex:

1. **Partial MPI send failure can leave ranks wedged.** `_multi_rank_sleep_wakeup()` sends control messages before entering the local try/finally; if `send()` fails after one peer received the message, rank 0 will skip ACK draining and other ranks may remain stuck in the control barrier. This probably needs send/recv error handling or at least a test where `send()` fails mid-broadcast to verify the behavior.
2. **Listener shutdown is daemon-only and not joined.** Rank 0 sends shutdown messages after worker join, but non-rank listener threads are daemon threads and there is no ACK/join before MPI/session teardown. This is probably okay in happy path, but risky around MPI finalize, process exit, or listener exceptions.
3. `_multi_rank_sleep_wakeup()` asserts rank 0 and communicator initialization. For production/user-facing paths, this should probably be explicit `RuntimeError` / `ValueError`, especially since the proxy now allowlists multi-rank sleep/wakeup.

hhzhang16:
@chienchunhung addressed the comments.
Changed:

- tensorrt_llm/executor/base_worker.py: replaced asserts with explicit RuntimeError / ValueError, moved broadcast into guarded error handling, sends best-effort aborts after partial broadcast failure, and drains ACKs from action/abort recipients.
- tensorrt_llm/_torch/pyexecutor/py_executor.py: added ABORT, shutdown ACK draining on rank 0, bounded non-rank listener join, and listener-side abort handling.
- tests/unittest/executor/test_sleep_collective_rpc_guards.py: added regression coverage for mid-broadcast send failure, explicit precondition errors, abort ACK/unblock, and shutdown ACK.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1200,21 +1209,60 @@ def shutdown(self):
...
+                    except Exception as exc:
+                        shutdown_errors.append(
+                            f"rank {dest} shutdown send failed: {exc}")
+                        logger.warning(
+                            "Failed to send sleep/wakeup listener shutdown to "
+                            "rank %d: %s", dest, exc)
+                for src in shutdown_ranks:
+                    try:
+                        ack = self._sleep_wakeup_comm.recv(
+                            source=src, tag=_SleepWakeupTag.ACK)
```
chienchunhung: If a listener died or never ACKs, rank 0 can hang during shutdown. Can we make shutdown ACK draining bounded/nonblocking, or add a regression where one shutdown ACK never arrives?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2623,8 +2680,23 @@ def _sleep_wakeup_listener_loop(self) -> None:
...
+                        # abort is stale, this bounded wait prevents the
+                        # listener from hanging forever on a message that no
+                        # longer has a matching control request.
+                        release_control_request = (
+                            self.control_request_barrier.wait(
+                                timeout=_SLEEP_WAKEUP_ABORT_BARRIER_TIMEOUT_S))
+                        error_msg = (
+                            "rank 0 aborted sleep/wakeup before local "
+                            f"execution: {msg.get('reason', 'unknown error')}")
+                        logger.warning("Sleep/wakeup listener: %s", error_msg)
```
chienchunhung, line 2894: If `ABORT` times out before `control_request_barrier` is set, it sends an ACK without setting `control_action_done.` If the executor loop later reaches the matching control request, it can wedge. 

For ABORT, should timeout be fatal or should it still arrange to unblock any later-arriving control request? A test where ABORT arrives before the barrier is set would cover this edge case.

chienchunhung (commented):
Thanks for addressing the previous comments! I left a few follow-ups.

tensorrt_llm/_torch/pyexecutor/py_executor.py
hhzhang16: Fixed. Shutdown ACK draining now uses a bounded nonblocking probe loop, so rank 0 will warn and continue instead of hanging if a listener never ACKs.

tensorrt_llm/_torch/pyexecutor/py_executor.py
hhzhang16: Fixed: I removed the timeout-before-release behavior, so the listener now waits for the matching control barrier and only ACKs after setting control_action_done, which prevents the later control request wedge.

tests/unittest/executor/test_sleep_collective_rpc_guards.py
```diff
@@ -93,28 +93,1030 @@ def test_missing_sleep_config_raises(self, method):
...
+        release.assert_not_called()
+        assert executor.control_request_barrier.is_set()
+        assert not executor.control_action_done.is_set()
+        assert sent_acks == [
+            {
+                "status": "ok",
+                "error": None,
+                "op_id": op_id,
+            }
+        ]
```
chienchunhung, line 804: The ACK payload now includes `phase`, but these tests still assert the older exact dict shape. For PREPARE the listener sends `{"status": "ok", "error": None, "op_id": op_id, "phase": _SleepWakeupAction.PREPARE}`, and shutdown sends `{"status": "ok", "error": None, "phase": _SleepWakeupAction.SHUTDOWN}`. 

Could you update these assertions to include `phase`, or assert the relevant fields individually?

chienchunhung (commented):
Thanks for address the comments and the improvement!

tests/unittest/executor/test_sleep_collective_rpc_guards.py
hhzhang16: Updated!

chienchunhung (approved):

taylor-yb-lee (approved):
LGTM for auto deploy part.

PR #14651, opened 2026-05-27
[#14712][feat] AutoDeploy: SA Enhancer for Spec Dec

### LINKED ISSUE #14712: [Feature]: AutoDeploy: Suffix Automaton Enhancement for Speculative Decoding

Should just basically use TRTLLM's Suffix Automaton framework.

### Alternatives

_No response_

### Additional context

_No response_

govind-ramnarayan:
fixes: #14712 

## Release Notes

* **New Features**
  * Added optional Suffix Automaton (SA) speculative decoding enhancement to Eagle models for improved draft token generation quality.
  * SA enhancer configuration now available with threshold-based draft refinement capabilities integrated into the execution pipeline.

* **Tests**
  * Added comprehensive unit and smoke tests validating SA speculative decoding with Eagle3 model configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14651?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/llm_args.py
+++ b/tensorrt_llm/_torch/auto_deploy/llm_args.py
@@ -202,6 +202,30 @@ class LlmArgs(DynamicYamlMixInForSettings, TorchLlmArgs, BaseSettings):
         return self
 
     @model_validator(mode="after")
+    def validate_sa_enhancer_requires_max_seq_len(self):
+        """Require an explicit max_seq_len when the SA (suffix-automaton) draft enhancer is used.
+
+        AutoDeploy builds the SA SuffixAutomatonManager and reserves its GPU workspace *before* the
+        KV-cache resize pass, so the resize sizes the cache leaving room for the SA pool. That
+        workspace is sized from max_seq_len and must be reserved before the model factory can infer
+        an unset max_seq_len from the model config, so for SA it must be set explicitly rather than
+        inferred.
+        """
+        spec_config = self.speculative_config
+        if (
+            spec_config is not None
+            and getattr(spec_config, "sa_config", None) is not None
+            and self.max_seq_len is None
+        ):
+            raise ValueError(
+                "AutoDeploy SA enhancer (speculative_config.sa_config) requires an explicit "
+                "max_seq_len: it sizes the suffix-automaton GPU workspace and is built before the "
+                "model factory can infer max_seq_len from the model config. Set max_seq_len to your "
+                "target context length."
+            )
+        return self
+
+    @model_validator(mode="after")
     def validate_parallel_config(self):
         """Setup parallel config according to world_size.
 
--- a/tensorrt_llm/_torch/auto_deploy/llm_args.py
+++ b/tensorrt_llm/_torch/auto_deploy/llm_args.py
@@ -202,30 +202,6 @@ class LlmArgs(DynamicYamlMixInForSettings, TorchLlmArgs, BaseSettings):
         return self
 
     @model_validator(mode="after")
-    def validate_sa_enhancer_requires_max_seq_len(self):
-        """Require an explicit max_seq_len when the SA (suffix-automaton) draft enhancer is used.
-
-        AutoDeploy builds the SA SuffixAutomatonManager and reserves its GPU workspace *before* the
-        KV-cache resize pass, so the resize sizes the cache leaving room for the SA pool. That
-        workspace is sized from max_seq_len and must be reserved before the model factory can infer
-        an unset max_seq_len from the model config, so for SA it must be set explicitly rather than
-        inferred.
-        """
-        spec_config = self.speculative_config
-        if (
-            spec_config is not None
-            and getattr(spec_config, "sa_config", None) is not None
-            and (self.max_seq_len is None or self.max_seq_len <= 0)
-        ):
-            raise ValueError(
-                "AutoDeploy SA enhancer (speculative_config.sa_config) requires a positive explicit "
-                "max_seq_len: it sizes the suffix-automaton GPU workspace and is built before the "
-                "model factory can infer max_seq_len from the model config. Set max_seq_len to your "
-                "target context length."
-            )
-        return self
-
-    @model_validator(mode="after")
     def validate_parallel_config(self):
         """Setup parallel config according to world_size.
 
--- a/tensorrt_llm/_torch/auto_deploy/llm_args.py
+++ b/tensorrt_llm/_torch/auto_deploy/llm_args.py
@@ -215,10 +215,10 @@ class LlmArgs(DynamicYamlMixInForSettings, TorchLlmArgs, BaseSettings):
         if (
             spec_config is not None
             and getattr(spec_config, "sa_config", None) is not None
-            and self.max_seq_len is None
+            and (self.max_seq_len is None or self.max_seq_len <= 0)
         ):
             raise ValueError(
-                "AutoDeploy SA enhancer (speculative_config.sa_config) requires an explicit "
+                "AutoDeploy SA enhancer (speculative_config.sa_config) requires a positive explicit "
                 "max_seq_len: it sizes the suffix-automaton GPU workspace and is built before the "
                 "model factory can infer max_seq_len from the model config. Set max_seq_len to your "
                 "target context length."
--- a/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
+++ b/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
@@ -43,9 +43,10 @@ from transformers.utils import ModelOutput
 
 from tensorrt_llm._torch.pyexecutor.mamba_cache_manager import MambaHybridCacheManager
 from tensorrt_llm._torch.speculative.sa_enhancer import SADraftEnhancer
+from tensorrt_llm.llmapi import SAEnhancerConfig
 
 from ...distributed.common import broadcast
-from ...shim.interface import CachedSequenceInterface
+from ...shim.interface import CachedSequenceInterface, SpeculativeDecodingModelArgs
 from ...utils._config import deep_merge_dicts
 from ...utils.logger import ad_logger
 from .modeling_nemotron_h import build_nemotron_eagle_layers
@@ -747,7 +748,7 @@ class EagleWrapperConfig:
     load_lm_head_from_target: bool
     normalize_target_hidden_state: bool = False
     sync_before_hidden_state_capture: bool = False
-    sa_config: Optional[Any] = None
+    sa_config: Optional[SAEnhancerConfig] = None
 
 
 class EagleWrapper(nn.Module):
@@ -758,6 +759,17 @@ class EagleWrapper(nn.Module):
     2. KV-cache mode (inference): runs graph-captured models with cached attention and
        in-place metadata updates. Hidden states flow through layer-prefixed
        `*_hidden_states_cache` kwargs.
+
+    SA draft-enhancement invariant:
+        `self.sa_enhancer` is set at construction iff SA is configured (`config.sa_config` is not
+        None); it is a stateless override policy. The suffix-automaton data structure it operates
+        on (`sa_manager`) is a per-iteration runtime object passed into `forward` and is only
+        available during executor inference. It is intentionally None during the KV-cache resize
+        transform, which calls `forward` purely to size memory before the executor's resource
+        managers exist. SA work therefore requires *both*: every SA call site guards on
+        `self.sa_enhancer is not None and sa_manager is not None`. The two checks are not
+        redundant and must not be collapsed — `sa_enhancer` set with `sa_manager` None is the
+        expected resize-time state and correctly no-ops SA.
     """
 
     def __init__(self, config: EagleWrapperConfig, target_model: nn.Module, draft_model: nn.Module):
@@ -845,15 +857,18 @@ class EagleWrapper(nn.Module):
         broadcast(ret, src=0)
         return ret
 
-    def forward(self, cache_seq_interface: Optional[CachedSequenceInterface] = None, **kwargs):
+    def forward(self, spec_dec_args: Optional[SpeculativeDecodingModelArgs] = None, **kwargs):
         """Dispatch to appropriate forward implementation.
 
-        - If seq_info is provided: inference mode (after graph transforms + cache init).
+        - If spec_dec_args is provided: inference mode (after graph transforms + cache init, or
+          the KV-cache resize transform). Its cache_seq_interface drives the cached path; its
+          sa_manager (possibly None) drives optional SA draft enhancement.
         - Otherwise: prefill-only mode (export time, before caches are inserted).
         """
-        sa_manager = kwargs.pop("sa_manager", None)
-        if cache_seq_interface is not None:
-            return self._forward_with_kv_cache(cache_seq_interface, sa_manager=sa_manager)
+        if spec_dec_args is not None:
+            return self._forward_with_kv_cache(
+                spec_dec_args.cache_seq_interface, sa_manager=spec_dec_args.sa_manager
+            )
         else:
             return self._forward_prefill_only(**kwargs)
 
@@ -958,9 +973,16 @@ class EagleWrapper(nn.Module):
         num_prefill: int,
         sa_manager,
     ) -> None:
+        # Both required (see class docstring): sa_manager is None during the resize transform.
         if self.sa_enhancer is None or sa_manager is None:
             return
 
+        # next_new_tokens is [batch, 1 + max_draft_len]: column 0 is the golden/bonus token and
+        # columns 1: are the draft proposals. Slice to generation rows (num_prefill:) and to the
+        # draft columns (1:) so the enhancer only sees and overrides draft tokens, leaving the
+        # golden token untouched. The PyTorch backend passes a separate pure-draft-token tensor
+        # (tensorrt_llm/_torch/speculative/eagle3.py), so the extra `1:` here only reflects AD
+        # packing the golden token alongside the drafts — it is not a behavioral difference.
         next_draft_tokens = self.sa_enhancer.maybe_override_all_draft_tokens(
             next_new_tokens[num_prefill:, 1:]
         )
@@ -1070,12 +1092,16 @@ class EagleWrapper(nn.Module):
             if num_extend > 0:
                 new_tokens_lens[num_prefill:] = new_tokens_lens_extend
 
+        # Both required (see class docstring): sa_manager is None during the resize transform.
         if self.sa_enhancer is not None and sa_manager is not None:
             self.sa_enhancer.extend_and_prepare(
                 sa_manager=sa_manager,
-                # Match the PyTorch backend's full-batch call shape. ADEngine has
-                # already prepared real manager slots for the extend requests, and
-                # the ID values are not read during the enhancer's extend path.
+                # extend_and_prepare takes request_ids but only consumes their count here: the
+                # real request->slot binding was already done by sa_manager.prepare() in the
+                # executor (with the actual IDs), so these values are unused. It is a small
+                # host-side Python list (not a device arange/tensor), so it is cheap even at
+                # batch=1. TODO: pass num_sequences instead of a dummy list once extend_and_prepare
+                # accepts a count (avoid touching the shared PyTorch enhancer signature in this PR).
                 request_ids=list(range(num_sequences)),
                 accepted_tokens=new_tokens_2d,
                 num_accepted_tokens=new_tokens_lens,
--- a/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
+++ b/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
@@ -764,12 +764,9 @@ class EagleWrapper(nn.Module):
         `self.sa_enhancer` is set at construction iff SA is configured (`config.sa_config` is not
         None); it is a stateless override policy. The suffix-automaton data structure it operates
         on (`sa_manager`) is a per-iteration runtime object passed into `forward` and is only
-        available during executor inference. It is intentionally None during the KV-cache resize
-        transform, which calls `forward` purely to size memory before the executor's resource
-        managers exist. SA work therefore requires *both*: every SA call site guards on
-        `self.sa_enhancer is not None and sa_manager is not None`. The two checks are not
-        redundant and must not be collapsed — `sa_enhancer` set with `sa_manager` None is the
-        expected resize-time state and correctly no-ops SA.
+        available when executor inference passes speculative-decoding runtime args. If cached
+        inference reaches this wrapper with an SA enhancer but no manager, the executor/model
+        contract has been broken and the wrapper fails loudly.
     """
 
     def __init__(self, config: EagleWrapperConfig, target_model: nn.Module, draft_model: nn.Module):
@@ -973,9 +970,10 @@ class EagleWrapper(nn.Module):
         num_prefill: int,
         sa_manager,
     ) -> None:
-        # Both required (see class docstring): sa_manager is None during the resize transform.
-        if self.sa_enhancer is None or sa_manager is None:
+        if self.sa_enhancer is None:
             return
+        if sa_manager is None:
+            raise RuntimeError("SA enhancer is enabled but no SA manager was provided.")
 
         # next_new_tokens is [batch, 1 + max_draft_len]: column 0 is the golden/bonus token and
         # columns 1: are the draft proposals. Slice to generation rows (num_prefill:) and to the
@@ -997,6 +995,9 @@ class EagleWrapper(nn.Module):
         Expected kwargs that are accessed directly are described in the required_kwargs property.
         Additional kwargs are forwarded to target/draft submodules (kv caches, etc.).
         """
+        if self.sa_enhancer is not None and sa_manager is None:
+            raise RuntimeError("SA enhancer is enabled but no SA manager was provided.")
+
         # ---- Phase 0: Check batch information ----
         # determine batch information from batch_info_host
         batch_info = csi.info.batch_info
@@ -1092,8 +1093,7 @@ class EagleWrapper(nn.Module):
             if num_extend > 0:
                 new_tokens_lens[num_prefill:] = new_tokens_lens_extend
 
-        # Both required (see class docstring): sa_manager is None during the resize transform.
-        if self.sa_enhancer is not None and sa_manager is not None:
+        if self.sa_enhancer is not None:
             self.sa_enhancer.extend_and_prepare(
                 sa_manager=sa_manager,
                 # extend_and_prepare takes request_ids but only consumes their count here: the
--- a/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
+++ b/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
@@ -763,9 +763,8 @@ class EagleWrapper(nn.Module):
     SA draft-enhancement invariant:
         `self.sa_enhancer` is set at construction iff SA is configured (`config.sa_config` is not
         None); it is a stateless override policy. The suffix-automaton data structure it operates
-        on (`sa_manager`) is a per-iteration runtime object passed into `forward` and is only
-        available when executor inference passes speculative-decoding runtime args. If cached
-        inference reaches this wrapper with an SA enhancer but no manager, the executor/model
+        on (`sa_manager`) is passed into `forward` through speculative-decoding runtime args. If
+        cached inference reaches this wrapper with an SA enhancer but no manager, the executor/model
         contract has been broken and the wrapper fails loudly.
     """
 
@@ -864,7 +863,9 @@ class EagleWrapper(nn.Module):
         """
         if spec_dec_args is not None:
             return self._forward_with_kv_cache(
-                spec_dec_args.cache_seq_interface, sa_manager=spec_dec_args.sa_manager
+                spec_dec_args.cache_seq_interface,
+                sa_manager=spec_dec_args.sa_manager,
+                require_sa_manager=spec_dec_args.require_sa_manager,
             )
         else:
             return self._forward_prefill_only(**kwargs)
@@ -969,24 +970,30 @@ class EagleWrapper(nn.Module):
         next_new_tokens: torch.Tensor,
         num_prefill: int,
         sa_manager,
+        require_sa_manager: bool = True,
     ) -> None:
         if self.sa_enhancer is None:
             return
         if sa_manager is None:
+            if not require_sa_manager and num_prefill == next_new_tokens.shape[0]:
+                return
             raise RuntimeError("SA enhancer is enabled but no SA manager was provided.")
 
         # next_new_tokens is [batch, 1 + max_draft_len]: column 0 is the golden/bonus token and
         # columns 1: are the draft proposals. Slice to generation rows (num_prefill:) and to the
         # draft columns (1:) so the enhancer only sees and overrides draft tokens, leaving the
-        # golden token untouched. The PyTorch backend passes a separate pure-draft-token tensor
-        # (tensorrt_llm/_torch/speculative/eagle3.py), so the extra `1:` here only reflects AD
-        # packing the golden token alongside the drafts — it is not a behavioral difference.
+        # golden token untouched.
         next_draft_tokens = self.sa_enhancer.maybe_override_all_draft_tokens(
             next_new_tokens[num_prefill:, 1:]
         )
         next_new_tokens[num_prefill:, 1:] = next_draft_tokens
 
-    def _forward_with_kv_cache(self, csi: CachedSequenceInterface, sa_manager=None):
+    def _forward_with_kv_cache(
+        self,
+        csi: CachedSequenceInterface,
+        sa_manager=None,
+        require_sa_manager: bool = True,
+    ):
         """Forward pass with KV cache (inference after graph transforms).
 
         Phases: target forward -> collect hidden states -> gather + sample + verify ->
@@ -995,7 +1002,7 @@ class EagleWrapper(nn.Module):
         Expected kwargs that are accessed directly are described in the required_kwargs property.
         Additional kwargs are forwarded to target/draft submodules (kv caches, etc.).
         """
-        if self.sa_enhancer is not None and sa_manager is None:
+        if self.sa_enhancer is not None and sa_manager is None and require_sa_manager:
             raise RuntimeError("SA enhancer is enabled but no SA manager was provided.")
 
         # ---- Phase 0: Check batch information ----
@@ -1005,6 +1012,11 @@ class EagleWrapper(nn.Module):
         num_prefill_tokens, num_extend_tokens, num_decode_tokens = batch_info.get_num_tokens()
         num_sequences = num_prefill + num_extend + num_decode
         num_total_tokens = num_prefill_tokens + num_extend_tokens + num_decode_tokens
+        skip_sa_enhancer = False
+        if self.sa_enhancer is not None and sa_manager is None:
+            if num_extend > 0:
+                raise RuntimeError("SA enhancer is enabled but no SA manager was provided.")
+            skip_sa_enhancer = True
 
         # some sanity checks on the batch
         assert num_decode == 0, "decode without drafting is not supported inside the eagle wrapper"
@@ -1093,15 +1105,11 @@ class EagleWrapper(nn.Module):
             if num_extend > 0:
                 new_tokens_lens[num_prefill:] = new_tokens_lens_extend
 
-        if self.sa_enhancer is not None:
+        if self.sa_enhancer is not None and not skip_sa_enhancer:
             self.sa_enhancer.extend_and_prepare(
                 sa_manager=sa_manager,
-                # extend_and_prepare takes request_ids but only consumes their count here: the
-                # real request->slot binding was already done by sa_manager.prepare() in the
-                # executor (with the actual IDs), so these values are unused. It is a small
-                # host-side Python list (not a device arange/tensor), so it is cheap even at
-                # batch=1. TODO: pass num_sequences instead of a dummy list once extend_and_prepare
-                # accepts a count (avoid touching the shared PyTorch enhancer signature in this PR).
+                # extend_and_prepare only needs the request count here; the real request->slot
+                # binding was already done by sa_manager.prepare() in the executor.
                 request_ids=list(range(num_sequences)),
                 accepted_tokens=new_tokens_2d,
                 num_accepted_tokens=new_tokens_lens,
@@ -1192,7 +1200,12 @@ class EagleWrapper(nn.Module):
                 csi.info.copy_("input_ids", draft_tokens)
                 csi.info.offset_pos_and_cache_(c_offset, active_args_override=draft_arg_names)
 
-        self._maybe_apply_sa_draft_override(next_new_tokens, num_prefill, sa_manager)
+        self._maybe_apply_sa_draft_override(
+            next_new_tokens,
+            num_prefill,
+            sa_manager,
+            require_sa_manager=require_sa_manager,
+        )
 
         # ---- Phase 6: Package output ----
         return EagleWrapperOutput(
--- a/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
+++ b/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
@@ -819,7 +819,6 @@ class EagleWrapper(nn.Module):
             hidden_states = fc(hidden_states)
         return hidden_states
 
-    # TODO: go through this logic, not sure if it is correct at the moment
     def apply_d2t(self, draft_output_ids: torch.Tensor) -> torch.Tensor:
         """Apply draft-to-target token mapping if available."""
         d2t = getattr(self._draft_inner_model, "d2t", None)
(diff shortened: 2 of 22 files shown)


tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -731,6 +732,16 @@ def _prepare_inputs(
...
 
+        sa_manager = resource_manager.get_resource_manager(
+            ResourceManagerType.SPEC_RESOURCE_MANAGER
+        )
+        if sa_manager is not None and getattr(self.spec_config, "sa_config", None) is not None:
+            gen_request_ids = [request.request_id for request in gen_requests]
+            sa_manager.prepare(gen_request_ids, self.spec_config.max_draft_len)
+            self.iter_states["sa_manager"] = sa_manager
+        else:
+            self.iter_states["sa_manager"] = None
```
govind-ramnarayan: nit: I don't think this is necessary - just don't insert it into iter_states?

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -731,6 +732,16 @@ def _prepare_inputs(
             **extra_args,
         )
 
+        sa_manager = resource_manager.get_resource_manager(
+            ResourceManagerType.SPEC_RESOURCE_MANAGER
+        )
+        if sa_manager is not None and getattr(self.spec_config, "sa_config", None) is not None:
+            gen_request_ids = [request.request_id for request in gen_requests]
+            sa_manager.prepare(gen_request_ids, self.spec_config.max_draft_len)
+            self.iter_states["sa_manager"] = sa_manager
```
govind-ramnarayan: is there precedent for this in the pytorch backend - does it really need to go in every iteration unconditionally?

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -937,6 +951,15 @@ def create_autodeploy_executor(
             ResourceManagerType.SEQ_SLOT_MANAGER: seq_slot_manager,
         }
     )
+    if spec_config is not None and getattr(spec_config, "sa_config", None) is not None:
```
govind-ramnarayan: wait, do we really need to register the resource manager here? Can't you just add it to the set of resource managers above and then let the executor creation do this?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1139,6 +1139,29 @@ def supports_backend(self, backend: str) -> bool:
         return backend not in ("pytorch", "_autodeploy")
 
 
+class SAEnhancerConfig(StrictBaseModel):
```
govind-ramnarayan: Why is this needed? Can't we just reuse whatever currently exists for SAEnhancer?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1185,6 +1208,11 @@ class EagleDecodingConfig(DecodingBaseConfig):
     eagle3_model_arch: Literal["llama3", "mistral_large3"] = Field(
         default="llama3",
         description="The model architecture of the eagle3 model.")
+    sa_config: Optional[SAEnhancerConfig] = Field(
```
govind-ramnarayan: ditto, I don't think we need to add anything here for support in AutoDeploy since it is already in the PyT backend...

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1329,29 +1357,6 @@ def is_linear_tree(self) -> bool:
         return False
 
 
-class SAEnhancerConfig(StrictBaseModel):
```
govind-ramnarayan, line 1504: Ah it looks like you just moved this block... can I ask why? It should be fine as it is and we don't need to touch this file right?

tests/unittest/auto_deploy/singlegpu/models/test_eagle.py
```diff
@@ -180,6 +182,54 @@ def test_eagle_rmsnorm_keeps_fp32_weights():
     assert norm.weight.dtype == torch.float32
 
 
+def test_eagle_wrapper_instantiates_sa_enhancer():
```
govind-ramnarayan: let's add a test that the EagleWrapper gets constructed correctly by the factory / the EagleWrapperConfig is populated correctly when going through the spec_config for Eagle + SAEnhancer.

tests/unittest/auto_deploy/singlegpu/models/test_eagle.py
```diff
@@ -180,6 +182,54 @@ def test_eagle_rmsnorm_keeps_fp32_weights():
...
+        ),
+        target_model=torch.nn.Module(),
+        draft_model=torch.nn.Module(),
+    )
+
+    assert wrapper.sa_enhancer is not None
+    assert wrapper.sa_enhancer.threshold == 2
+
+
+def test_eagle_wrapper_sa_override_updates_next_new_tokens():
```
govind-ramnarayan: Maybe can also include a smoke test where we run an existing Eagle/MTP smoke test with the FakeSAEnhancer, and we record the draft tokens with mocking, and we check that indeed the FakeSAEnhancer is overriding.

tests/unittest/auto_deploy/singlegpu/models/test_eagle.py
```diff
@@ -180,6 +182,54 @@ def test_eagle_rmsnorm_keeps_fp32_weights():
     assert norm.weight.dtype == torch.float32
```
govind-ramnarayan, line 201: let us also add a mock unit test for the autodeploy executor that the config is properly getting to the executor.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -731,6 +732,21 @@ def _prepare_inputs(
...
 
+        self.iter_states.pop("sa_manager", None)
+        self.iter_states.pop("sa_request_ids", None)
+        sa_manager = resource_manager.get_resource_manager(
+            ResourceManagerType.SPEC_RESOURCE_MANAGER
+        )
+        if (
+            sa_manager is not None
+            and getattr(self.spec_config, "sa_config", None) is not None
+            and num_extend > 0
```
govind-ramnarayan: Are you sure we only use the SAManager when num_extend > 0? Wouldn't it make sense to also apply it on prefill requests? Can you check the PyTorch backend?

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -744,7 +760,14 @@ def _run_forward(self) -> Dict[str, Optional[torch.Tensor]]:
         csi = self.cache_seq_interface
 
         if self.spec_config is not None:
-            model_output = self.model(**csi.named_args, cache_seq_interface=csi)
+            sa_manager = self.iter_states.get("sa_manager")
```
govind-ramnarayan: This is bad. Let's not store them in iter_states - isn't this just a way to get information about each iteration afterwards, like telemetry?

If the speculative decoding config comes with SA, maybe we should store the SAManager as a field in the ADEngine upon initialization. Then we should get it like self.sa_manager just like we get self.cache_seq_interface. 

If this is not possible because the sa_manager is stored by the executor, I still think it makes sense to do something like 

if self.sa_manager is None:
- store sa_manager in self.sa_manager

in prepare_inputs().

The SA Manager should not be changing every iteration - but it is managed by the executor.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -846,8 +851,12 @@ def forward(self, cache_seq_interface: Optional[CachedSequenceInterface] = None,
         - If seq_info is provided: inference mode (after graph transforms + cache init).
         - Otherwise: prefill-only mode (export time, before caches are inserted).
         """
+        sa_manager = kwargs.pop("sa_manager", None)
+        sa_request_ids = kwargs.pop("sa_request_ids", None)
         if cache_seq_interface is not None:
-            return self._forward_with_kv_cache(cache_seq_interface)
+            return self._forward_with_kv_cache(
+                cache_seq_interface, sa_manager=sa_manager, sa_request_ids=sa_request_ids
```
govind-ramnarayan: I thought we specifically said these would not be needed by the forward function, and that we could instead use dummy request IDs. This is preferable rather than passing unnecessary info into this function. We want to minimize the amount of info we pass to our forward functions - the SA Manager is necessary because it contains the Suffix Automaton data structure, but the actual request IDs did not seem necessary. When I looked into the function signature, it looked like these were basically just immediately used for the number of request IDs rather than anything else. Can you confirm? That's not to say they're not needed elsewhere, just it seems not in the forward itself...

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -942,7 +948,22 @@ def _collect_hidden_states(kwargs: dict, num_tokens: int) -> torch.Tensor:
...
 
-    def _forward_with_kv_cache(self, csi: CachedSequenceInterface):
+    def _maybe_apply_sa_draft_override(
+        self,
+        next_new_tokens: torch.Tensor,
+        num_prefill: int,
+        num_extend: int,
+        sa_manager,
+    ) -> None:
+        if self.sa_enhancer is None or sa_manager is None or num_extend == 0:
```
govind-ramnarayan: again I don't think we need the num_extend == 0 check here... right? As we discussed we want to call this even with the prefill requests even if it doesn't do anything?

govind-ramnarayan: But look at the PyTorch backend and let me know. Please cite your sources.

tests/unittest/auto_deploy/singlegpu/shim/test_create_ad_executor.py
```diff
@@ -152,3 +153,72 @@ def test_create_autodeploy_executor_with_guided_decoding(
...
+    )
+
+    mock_engine = Mock()
+    mock_engine.cache_seq_interface.info.max_num_tokens = 512
+    mock_engine.cache_seq_interface.info.vocab_size_padded = 1000
+    mock_engine.cache_seq_interface.max_num_state_slots = ad_config.max_batch_size
+    mock_engine.cache_seq_interface.kv_cache_manager = Mock()
+    mock_engine.cache_seq_interface.kv_cache_manager.impl = Mock()
+    mock_engine.cache_seq_interface.kv_cache_config_tuned = SimpleNamespace(tokens_per_block=64)
+    mock_dist = Mock()
```
govind-ramnarayan: wait why do we need mock_dist? Is it needed for this? We're only creating the AD Executor, feels like it should just be exactly the same as the other ones e.g. guided decoding.  Let's simplify our mocking given this.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
```diff
@@ -339,13 +339,15 @@ class TestLlama3_1_8B_Instruct_Eagle3(LlmapiAccuracyTestHarness):
     EAGLE_MODEL_PATH = hf_id_to_local_model_dir(
         "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B")
 
-    def get_default_kwargs(self, attn_backend="flashinfer"):
+    def get_default_kwargs(self, attn_backend="flashinfer", enable_sa=False):
         yaml_paths, _ = _get_registry_yaml_extra(self.MODEL_NAME)
+        sa_config = SAEnhancerConfig() if enable_sa else None
```
govind-ramnarayan: Use some reasonable SAConfig here, not completely empty. Use PyTorch backend for inspiration.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: Confirmed against the PyTorch backend and updated AD to match.

Sources:
- PyTorch Eagle3 calls `extend_and_prepare(...)` whenever both `self.sa_enhancer` and `sa_manager` exist, with no `num_gens > 0` guard: `tensorrt_llm/_torch/speculative/eagle3.py`, currently around lines 585-596.
- `SADraftEnhancer.extend_and_prepare(...)` itself handles zero-gen batches: it always resets buffers/state, then only launches the SA `extend`/`extend_global` path under `if num_gens > 0`: `tensorrt_llm/_torch/speculative/sa_enhancer.py`, currently around lines 101-137.
- PyTorch Eagle3 also calls `maybe_override_all_draft_tokens(...)` after slicing to generation rows whenever SA is enabled: `tensorrt_llm/_torch/speculative/eagle3.py`, currently around lines 753-759. With a pure-prefill iteration that slice is empty, and the shared enhancer no-ops because `_num_gens == 0`: `tensorrt_llm/_torch/speculative/sa_enhancer.py`, currently around lines 154-163.

So the AD flow now does the same thing: call `extend_and_prepare` for pure prefill, and call the override hook with the generation slice even when that slice is empty.

yingguo-trt (approved):

jieli-matrix (approved):

govind-ramnarayan:

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -746,6 +747,7 @@ class EagleWrapperConfig:
     load_lm_head_from_target: bool
     normalize_target_hidden_state: bool = False
     sync_before_hidden_state_capture: bool = False
+    sa_config: Optional[Any] = None
```
hnover-nv: We can't type this more tightly?

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
hnover-nv, line 1: in torch backend, there's Eagle and MTPEagle I think.  I'm not sure if SA is supported for both or not.  Are we at least matching them, and could we do both if they don't?  I think Nemotron models use MTPEagle so it's a useful path I think.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -942,7 +948,21 @@ def _collect_hidden_states(kwargs: dict, num_tokens: int) -> torch.Tensor:
...
         return torch.cat(layer_hidden_states, dim=1)
 
-    def _forward_with_kv_cache(self, csi: CachedSequenceInterface):
+    def _maybe_apply_sa_draft_override(
+        self,
+        next_new_tokens: torch.Tensor,
+        num_prefill: int,
+        sa_manager,
+    ) -> None:
+        if self.sa_enhancer is None or sa_manager is None:
```
hnover-nv: can these ever disagree?  would it be better to either only check one, or assert they go together?

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -942,7 +948,21 @@ def _collect_hidden_states(kwargs: dict, num_tokens: int) -> torch.Tensor:
...
+            return
+
+        next_draft_tokens = self.sa_enhancer.maybe_override_all_draft_tokens(
+            next_new_tokens[num_prefill:, 1:]
```
hnover-nv, line 980: the torch eagle implementation doesn't have the 1: here.  Are we keeping the golden token around, while they don't?

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -1046,6 +1066,20 @@ def _forward_with_kv_cache(self, csi: CachedSequenceInterface):
             if num_extend > 0:
                 new_tokens_lens[num_prefill:] = new_tokens_lens_extend
 
+        if self.sa_enhancer is not None and sa_manager is not None:
```
hnover-nv: same question about what our guards should be

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -1046,6 +1066,20 @@ def _forward_with_kv_cache(self, csi: CachedSequenceInterface):
...
+            self.sa_enhancer.extend_and_prepare(
+                sa_manager=sa_manager,
+                # Match the PyTorch backend's full-batch call shape. ADEngine has
+                # already prepared real manager slots for the extend requests, and
+                # the ID values are not read during the enhancer's extend path.
+                request_ids=list(range(num_sequences)),
```
hnover-nv, line 1095: This comment makes its ound like we're doing somethign non-standard in AD.  I feel like it'd be better to say something like (if true?):
extend_and_prepare takes in the request_ids here but only uses it for a length.  The actual slots are determined during the prepare call in ....

But also, is this inside a cuda graph.  Are we instantiating range(num_sequences) into a tensor every time, or is this all host-side code?  I think host-side but owrth checking, just an arange is a little pricey during a batch=1 invocation.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -879,6 +881,16 @@ def _prepare_inputs(
             **extra_args,
         )
 
+        if getattr(self.spec_config, "sa_config", None) is not None:
+            if self.sa_manager is None:
```
hnover-nv: Is this lazy init of the resource manager?  do other resource managers get lazy init?

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -892,7 +904,11 @@ def _run_forward(self) -> Dict[str, Optional[torch.Tensor]]:
         csi = self.cache_seq_interface
 
         if self.spec_config is not None:
-            model_output = self.model(**csi.named_args, cache_seq_interface=csi)
+            model_output = self.model(
+                **csi.named_args,
+                cache_seq_interface=csi,
+                sa_manager=self.sa_manager,
```
hnover-nv, line 1005: this seems a little unclean, why are we passing the sa manager baically uniquely.  we don't pass anything special for eagle or whatever.  We pass a generic-seeming cache seq itnerfface and the args for it, and then the sa manager?

tests/integration/test_lists/qa/llm_function_core.txt
```diff
@@ -111,6 +111,8 @@ accuracy/test_llm_api_autodeploy.py::TestLlama3_1_8B::test_auto_dtype[trtllm-Fal
 accuracy/test_llm_api_autodeploy.py::TestLlama3_1_8B::test_auto_dtype[trtllm-True-1]
 accuracy/test_llm_api_autodeploy.py::TestLlama3_1_8B_Instruct_Eagle3::test_eagle3_one_model[flashinfer]
 accuracy/test_llm_api_autodeploy.py::TestLlama3_1_8B_Instruct_Eagle3::test_eagle3_one_model[trtllm]
+accuracy/test_llm_api_autodeploy.py::TestLlama3_1_8B_Instruct_Eagle3::test_eagle3_one_model_sa[flashinfer]
```
hnover-nv: do we actually need both, or are the code paths orthogonal, so it'd be enough to use one of the backends with sa?

hnover-nv (commented):
I need to look at the last few files but thought I'd send what I've got.

hnover-nv:
read over the rest, and it's fine, so just look at the prior partial review.  thanks!

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: SA should be supported for everything - this is our only path which covers both Eagle and MTPEagle, so we're doing both.

tests/integration/test_lists/qa/llm_function_core.txt
govind-ramnarayan: We probably don't need both, down to make this just trtllm.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: We pass the CachedSequenceInterface actually uniquely for Eagle and MTP. That's just the only spec dec we support right now.

I didn't really know where the SAManager belongs. It's a resource that is not managed by us, but instead by the PyExecutor, so it doesn't really fit in with any of our existing classes well. It needs to be passed into this method somehow, so if we don't put into any pre-existing class then it needs to come on its own.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: your comment is better. That is what is intended.

I think it has to be host-side code, though I'll check. The number of sequences can vary between iterations. Maybe there's a cheaper way to make a list of this size that we can use? Ultimately would like to avoid needing to modify PyTorch backend code in this PR (e.g. to make that parameter just the batch_size instead of the request_ids in the extend_and_prepare() function)

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: TODO: comment @ top of class about invariant - if sa_enhancer is not None, make sure sa_manager is not None where needed. Namely, if we think we are doing SA enhancement, we better get the right data structure for it provided to us.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: take a look + clarify w/ a comment

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -892,7 +904,11 @@ def _run_forward(self) -> Dict[str, Optional[torch.Tensor]]:
         csi = self.cache_seq_interface
 
         if self.spec_config is not None:
-            model_output = self.model(**csi.named_args, cache_seq_interface=csi)
+            model_output = self.model(
+                **csi.named_args,
+                cache_seq_interface=csi,
```
govind-ramnarayan, line 1004: make struct for SpecDecArgs that contains special things that need to be passed for spec dec. Then we can also document this struct to make it clear why each thing is passed.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: resolution: let's make a struct called SpeculativeDecodingModelArgs{} that contains everything we need specially for speculative decoding. Then, we can have the cached sequence interface (existing) and the SAManager both in here (the latter as optional). And unpack it in the EagleWrapper.forward().

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: This is a good point; we're sort of lazy initializing in prepare_inputs() because that's where we have access to resource managers coming from the executor, but it's the same structure every time. It gets updated across iterations, but we don't actually do anything with the update - we just pass it through to the ADEngine.

Ideally, this would actually belong in the initialization of the ADEngine, the only problem is that currently these things are kind of backwards - ADEngine gets initialized before resource managers are initialized.

Let's evaluate some possibilities for resolving this - either initialize this resource manager before the ADEngine so we can pass it in at build time, or simply have a "setter" in the ADEngine to set the SAManager after initialization, something like that. The first seems preferable to me if it is possible and relatively clean...

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -946,7 +964,29 @@ def _collect_hidden_states(kwargs: dict, num_tokens: int) -> torch.Tensor:
...
+    ) -> None:
+        if self.sa_enhancer is None:
+            return
+        if sa_manager is None:
+            raise RuntimeError("SA enhancer is enabled but no SA manager was provided.")
+
+        # next_new_tokens is [batch, 1 + max_draft_len]: column 0 is the golden/bonus token and
+        # columns 1: are the draft proposals. Slice to generation rows (num_prefill:) and to the
+        # draft columns (1:) so the enhancer only sees and overrides draft tokens, leaving the
+        # golden token untouched. The PyTorch backend passes a separate pure-draft-token tensor
```
govind-ramnarayan: let's remove the reference for pytorch backend, we don't need this. The prior clarifying comment is enough.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
```diff
@@ -1050,6 +1093,23 @@ def _forward_with_kv_cache(self, csi: CachedSequenceInterface):
...
                 new_tokens_lens[num_prefill:] = new_tokens_lens_extend
 
+        if self.sa_enhancer is not None:
+            self.sa_enhancer.extend_and_prepare(
+                sa_manager=sa_manager,
+                # extend_and_prepare takes request_ids but only consumes their count here: the
+                # real request->slot binding was already done by sa_manager.prepare() in the
+                # executor (with the actual IDs), so these values are unused. It is a small
+                # host-side Python list (not a device arange/tensor), so it is cheap even at
+                # batch=1. TODO: pass num_sequences instead of a dummy list once extend_and_prepare
```
govind-ramnarayan: nit: remove the TODO - it should be clear when we hit an error :)

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -444,6 +451,11 @@ def __init__(
         self.llm_args.max_seq_len = cache_seq_interface.info.max_seq_len
         self.iter_counter = 0
         self.iter_states = {}
+        # SA suffix-automaton manager, set at construction (not lazily fetched per-iteration). Its
```
govind-ramnarayan: This information is not needed here... just get rid of the comment

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -1148,31 +1169,51 @@ def create_autodeploy_executor(
...
-    )
-
     spec_config = ad_config.speculative_config
 
     if spec_config is not None and ad_config.guided_decoding_backend is not None:
         raise ValueError(
             "Guided decoding is not currently supported for speculative decoding in AutoDeploy."
         )
 
+    # Create the SA suffix-automaton manager (if enabled) and eagerly allocate its GPU workspace
```
govind-ramnarayan: Let's shorten this - I think the justification about wanting to allocate its GPU workspace before entering the optimization pipeline that has resize_kv_cache is useful. The rest is sort of incidental and not meaningful to people outside of this PR I think.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -15,7 +15,8 @@
 
 import copy
 import functools
-from typing import Callable, Dict, List, Optional, Tuple, Union, final
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Tuple, Union, final
```
govind-ramnarayan: nit: remove TYPE_CHECKING.

govind-ramnarayan: import it if you can, don't if you can't. We don't need the type check that badly

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
```diff
@@ -338,14 +338,18 @@ class TestLlama3_1_8B_Instruct_Eagle3(LlmapiAccuracyTestHarness):
...
+            threshold=self.MAX_DRAFT_LEN) if enable_sa else None)
         speculative_config = Eagle3DecodingConfig(
-            max_draft_len=3,
+            max_draft_len=self.MAX_DRAFT_LEN,
             speculative_model=self.EAGLE_MODEL_PATH,
             eagle3_one_model=True,
             eagle3_layers_to_capture={1, 15, 28},
+            sa_config=sa_config,
         )
         # Note: Test crashes with trtllm attn_backend + torch-simple
```
govind-ramnarayan: nit: this comment is now out-of-date, let's remove it. The test no longer crashes thankfully :) Also remove the mention of this issue, fairly sure it's resolved.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
```diff
@@ -400,6 +404,24 @@ def test_eagle3_one_model(self, attn_backend):
...
+
+        with AutoDeployLLM(
+                model=self.MODEL_PATH,
+                tokenizer=self.MODEL_PATH,
+                **kwargs,
+        ) as llm:
+            task = GSM8K(self.MODEL_NAME)
+            task.evaluate(llm, extra_acc_spec="use_sa_spec")
+
+            self.check_acceptance_rate(llm, min_acceptance_rate=0.18)
```
govind-ramnarayan: nit: let's use acceptance rate 0.15... might be safer as the SA could decrease acceptance rate going below the threshold on some hardwares...

tests/torch/speculative/test_suffix_automaton.py
```diff
@@ -21,6 +41,41 @@ def test_manager_creation(self):
         assert manager is not None
         manager.shutdown()
 
+    def test_prepare_allocates_expected_gpu_workspace(self):
```
govind-ramnarayan: While this is a good test, it's actually kind of about AutoDeploy correctness afaict (not sure how kv cache resize / gpu memory accounting works for the pytorch backend). So I'd rather put something like this in the a test_suffix_automaton.py unit test file for AutoDeploy

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -482,6 +494,7 @@ def __init__(
 
         # Store the cache sequence interface
         self.cache_seq_interface = cache_seq_interface
+        self.cache_seq_interface.sa_manager = self.sa_manager
```
govind-ramnarayan: why are we storing the SA Manager in the cached sequence interface? Is this really necessary? If this is the case then we don't need the special speculative args struct do we? I think it makes more sense to have it owned by the ADExecutor, unless the reason is that we need to have it for resize kv cache. But I don't think we really do, we just need the gpu memory to be allocated and we need it to be in the cuda graph warmup calls.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
```diff
@@ -1359,3 +1360,35 @@ def shutdown(self) -> None:
...
+    receive this object; they are still invoked with just ``cache_seq_interface.named_args``.
+
+    Attributes:
+        cache_seq_interface: Per-iteration sequence and KV-cache state. Always present whenever a
+            spec-dec model forward runs with caches (inference and the resize transform).
+        sa_manager: Suffix-automaton manager backing the SA draft enhancer. Owned by the
+            ``PyExecutor`` resource managers and held by the AutoDeploy engine as a reference. Set
+            only when SA enhancement is enabled (``spec_config.sa_config`` is not ``None``) and the
+            manager exists; ``None`` otherwise (e.g. non-SA speculation).
+        require_sa_manager: Whether a configured SA draft enhancer must receive ``sa_manager``.
```
govind-ramnarayan: I think this is only necessary if there is some good reason that we don't want to pass the SA Manager to the KV Cache resize pass. Additionally, I guess by putting the SA Manager on "warm-up" passes like for cuda graph capture and kv cache resize, are we possibly polluting the SA Manager with nonsense suffixes that should not actually match when we do our real requests? Or is this handled automatically by the SAManager? I guess it should be looking for matching suffixes within the same request, so it should not get contaminated by other requests.

1) Can you check this?
2) Does it make sense for us to "free" memory from the SA Manager after building the engine so that it is not keeping around a Suffix Automaton data structure for requests that don't exist? Though these are probably tiny...

govind-ramnarayan: I guess what I'm saying is, yes, kv cache resize pass does not need the sa manager strictly speaking. But it also shouldn't be hurt by it, right?

tests/unittest/auto_deploy/singlegpu/shim/test_create_ad_executor.py
```diff
@@ -74,6 +75,59 @@ class MockFactory:
 """Unit tests for create_autodeploy_executor function."""
 
 
+def test_ad_engine_owns_sa_manager_without_mutating_cache_interface():
```
govind-ramnarayan: nit: get rid of this test, unnecessary

tests/unittest/auto_deploy/singlegpu/shim/test_create_ad_executor.py
```diff
@@ -74,6 +75,59 @@ class MockFactory:
...
+        cache_seq_interface=cache_seq_interface,
+        sa_manager=sa_manager,
+    )
+
+    assert model_build_state["has_sa_manager"] is False
+    assert not hasattr(cache_seq_interface, "sa_manager")
+    assert engine.sa_manager is sa_manager
+
+
+def test_ad_engine_build_from_config_forwards_sa_manager_to_optimizer():
```
govind-ramnarayan: nit: add a comment about why this test is meaningful. Why does it need to be in the optimizer? Then rename the test around this idea

tensorrt_llm/_torch/auto_deploy/shim/interface.py
govind-ramnarayan: Updated in ce2c0cabb9: I removed the resize-only `require_sa_manager=False` path and now pass `shared_config.sa_manager` through `ResizeKVCache` into `SpeculativeDecodingModelArgs`.

I checked the SAManager warmup behavior: unknown synthetic IDs are mapped to the reserved dummy slot in `prepare()`, and `extend()` masks dummy entries by zeroing their accepted-token counts before invoking the kernel. I added `test_warmup_dummy_request_ids_do_not_create_request_state` to cover that resize/cudagraph warmup IDs do not populate `_request_to_slot` or `_active_slots`.

I did not add a post-build free/reset: with dummy IDs there is no real request SA state to free, and the GPU workspace remains intentionally reserved for KV-cache sizing and cudagraph warmup.

tests/unittest/auto_deploy/singlegpu/shim/test_create_ad_executor.py
govind-ramnarayan: Done in ce2c0cabb9. Removed `test_ad_engine_owns_sa_manager_without_mutating_cache_interface`.

tests/unittest/auto_deploy/singlegpu/shim/test_create_ad_executor.py
govind-ramnarayan: Done in ce2c0cabb9. Renamed this to `test_ad_engine_build_passes_sa_manager_to_optimizer_for_resize_accounting` and added a short comment explaining why this belongs at the optimizer boundary: `resize_kv_cache` runs inside `InferenceOptimizer`, so SA-enabled builds need the manager in optimizer shared config for resize/cudagraph workspace accounting.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: Done in ce2c0cabb9. Removed the PyTorch-backend reference and kept the local MTP normalization comment shorter.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_eagle.py
govind-ramnarayan: Done in ce2c0cabb9. Removed the TODO comment.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: Done in ce2c0cabb9. Removed the extra explanatory detail from the `sa_manager` constructor arg docs.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: Done. The comment is now limited to the useful rationale: reserve SA workspace before engine build so `resize_kv_cache` sizes KV cache with room for the SA pool.

tensorrt_llm/_torch/auto_deploy/shim/interface.py
govind-ramnarayan: Done. Removed the `TYPE_CHECKING` path; the struct keeps the manager as `Optional[object]` so we do not need a special type-only import.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
govind-ramnarayan: Done. Removed the stale crash/issue note from the AutoDeploy accuracy test.

tests/integration/defs/accuracy/test_llm_api_autodeploy.py
govind-ramnarayan: Done. The SA Eagle3 accuracy check now uses `min_acceptance_rate=0.15`.

tests/torch/speculative/test_suffix_automaton.py
govind-ramnarayan: Done. Moved the workspace/memory-accounting coverage into AutoDeploy's `tests/unittest/auto_deploy/singlegpu/shim/test_suffix_automaton.py` instead of the PyTorch suffix-automaton test.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: Done. The SA manager is not stored on `CachedSequenceInterface`; `create_autodeploy_executor()` creates it up front, passes it into `ADEngine.build_from_config()`, and registers the same manager as the spec resource manager. `SpeculativeDecodingModelArgs` remains the executor/model handoff for cached spec-dec forward calls.

tensorrt_llm/_torch/auto_deploy/llm_args.py
```diff
@@ -201,6 +201,30 @@ def validate_ssm_replay_requires_spec(self):
             )
         return self
 
+    @model_validator(mode="after")
```
govind-ramnarayan: note to Harris; this is a little weird because we often rely on defaults for max_seq_len, but it turns out it gets defaulted during the AutoDeploy build-and-optimize pipeline. We initialize the SA Manager before this pipeline runs, because later stages of the pipeline require it, so we don't pick up this default. Technically we could insert it in between the stage where max_seq_len gets defaulted and where the SA Manager is needed, but this seemed cleaner...

tests/unittest/auto_deploy/singlegpu/models/test_eagle.py
```diff
@@ -180,6 +184,221 @@ def test_eagle_rmsnorm_keeps_fp32_weights():
...
 
 
+def test_eagle_wrapper_instantiates_sa_enhancer():
```
govind-ramnarayan: let's have unit tests of the eagle wrapper moved to a new test_modeling_eagle.py in the modeling tests. Tests that test the eagle factory can remain here.

tests/unittest/auto_deploy/singlegpu/shim/test_create_ad_executor.py
```diff
@@ -74,6 +75,40 @@ class MockFactory:
 """Unit tests for create_autodeploy_executor function."""
 
 
+def test_ad_engine_build_passes_sa_manager_to_optimizer_for_resize_accounting():
```
govind-ramnarayan: This is not necessary for resize accounting. It's needed to be passed to the optimizer so that it is present during cuda graph capture. Let's make the name 

test_inference_optimizer_sees_sa_manager()

tests/unittest/auto_deploy/singlegpu/shim/test_engine.py
```diff
@@ -824,6 +825,71 @@ def get_draft_token_length(self):
     cache_seq_interface.shutdown()
 
 
+def test_ad_engine_prepare_inputs_prepares_sa_manager_with_extend_request_ids():
```
govind-ramnarayan: nit: name this 

test_prepare_inputs_prepares_sa_manager()

tests/unittest/auto_deploy/singlegpu/shim/test_engine.py
```diff
@@ -824,6 +825,71 @@ def get_draft_token_length(self):
...
+    engine.sa_manager = Mock()
+
+    resource_manager = _DummyResourceManager(_DummyKVCacheManager(tokens_per_block))
+
+    prefill_scheduled = ScheduledRequests()
+    prefill_scheduled.context_requests_last_chunk.append(
+        _DummyRequest(tokens=[1, 2, 3], begin=0, size=3, seq_slot=0)
+    )
+    engine._prepare_inputs(prefill_scheduled, resource_manager, new_tokens=None)
+    engine.sa_manager.prepare.assert_not_called()
```
govind-ramnarayan, line 858: Explain why this should not be called here? What exactly was this testing?

govind-ramnarayan: Also I don't understand - doesn't it make sense that we should prepare with prefill requests too? Please double check this with the pytorch backend. From what I understand we should be inputting every token into the Suffix Automaton struct so that it can query for suffixes of the entire sequence so far, not just the extend requests alone. Is this not correct?

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -953,6 +962,10 @@ def _prepare_inputs(
             **extra_args,
         )
 
+        if self.sa_manager is not None and num_extend > 0:
+            gen_request_ids = [request.py_request_id for request in extend_requests]
+            self.sa_manager.prepare(gen_request_ids, self.spec_config.max_draft_len)
```
govind-ramnarayan, line 988: see my comment on the tests - why do we not prepare for prefill requests? It seems to me at a high level that we want to pass every token from each request into the Suffix Automaton struct so that we can query matching suffixes from the entire string *including* prefill. Or is that only handled by prepare_and_extend() and not by prepare() in isolation? Please describe to me what tokens we store in the SA data structure and reference the PyTorch backend in your explanation.

PR #14687, opened 2026-05-28
[None][refactor] Refactor Skip Softmax Attention Interface

bobboli:
## Summary

1. Split sparse attention user configs from backend runtime params so `AttentionBackend` instances store lowered `SparseParams`, not public config objects.
2. Centralize Skip Softmax Attention backend helpers: ModelOpt checkpoint parsing, calibration formula evaluation, scheduler state, kernel params, and `SkipSoftmaxParams`.
3. Lower LLM and VisualGen Skip Softmax Attention configs through `to_sparse_params()` before backend construction, including ModelOpt `ignore` handling.
4. Keep VisualGen checkpoint calibration per model component, so multi-transformer pipelines resolve each component from its own `config.json`.
5. Add VisualGen normalized timestep cutoff support through `disabled_until_timestep` and include the resulting Skip Softmax Attention phase in CUDA graph keys.
6. Refresh sparse attention docs and reorganize LLM/VisualGen tests around Python API, YAML API, checkpoint metadata, layer filtering, and timestep cutoff behavior.

## Details

1. Sparse runtime params boundary
   - Adds the sparse backend base carriers `SparseParams` and `SparseMetadataParams`: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/attention_backend/sparse/params.py#L25
   - `TrtllmAttention` accepts and stores lowered sparse runtime params: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/attention_backend/trtllm.py#L1114
   - `create_attention()` forwards `sparse_params` into backend constructors instead of exposing user config to backend instances: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/attention_backend/utils.py#L77

2. Shared Skip Softmax Attention backend logic
   - Centralizes ModelOpt checkpoint parsing for `sparse_attention_config.config_groups`, allowing at most one Skip Softmax Attention group: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py#L60
   - Implements formula parsing and evaluation for calibrated `target_sparsity` to `threshold_scale_factor` lowering: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py#L124
   - `SkipSoftmaxScheduler` resolves concrete kernel params and handles VisualGen timestep-based disablement: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py#L282
   - `SkipSoftmaxParams` is the runtime object passed to the attention backend: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py#L420

3. LLM lowering path
   - LLM `SkipSoftmaxAttentionConfig` validates `target_sparsity`, resolves checkpoint defaults, lowers to `SkipSoftmaxParams`, and applies checkpoint `ignore` patterns: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/llmapi/llm_args.py#L668
   - LLM attention modules lower config before backend construction and pass layer identity for checkpoint `ignore` matching: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/modules/attention.py#L531
   - ModelEngine still owns the global sparse config, but metadata/cache construction receives lowered metadata params rather than backend params: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/pyexecutor/model_engine.py#L1661

4. VisualGen per-layer lowering and checkpoint calibration
   - VisualGen keeps a separate user-facing `SkipSoftmaxAttentionConfig` for runtime knobs such as `threshold_scale_factor`, `target_sparsity`, and `disabled_until_timestep`: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/visual_gen/sparse_attention.py#L46
   - VisualGen lowering combines user config with each component model's checkpoint metadata and returns either `None` for ModelOpt-ignored layers or `SkipSoftmaxParams` for enabled layers: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/visual_gen/sparse_attention.py#L71
   - VisualGen attention modules pass the layer `module_name` and component `pretrained_config` into lowering: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/visual_gen/modules/attention.py#L203
   - Component-relative `ignore` patterns are supported for multi-transformer checkpoints such as `transformer` and `transformer_2`: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/visual_gen/sparse_attention.py#L107

5. VisualGen timestep cutoff and CUDA graph keys
   - VisualGen CUDA graph keys are structured as named key parts and can include registered non-shape key functions: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/visual_gen/cuda_graph_runner.py#L12
   - `BaseDiffusionModel` registers the Skip Softmax Attention graph phase from the model-forward `timestep`: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/visual_gen/models/modeling.py#L38
   - VisualGen TRTLLM attention backend consumes `timestep` as an internal scheduling signal and does not pass it to the raw attention op: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/visual_gen/attention_backend/trtllm.py#L291
   - WAN and Cosmos normalize scheduler timesteps before model forward, while WAN rescales internally for its embedding convention: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py#L554 and https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py#L691

6. Docs and tests
   - LLM sparse-attention docs were refocused around the public LLM API and ModelOpt checkpoint config shape: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/docs/source/features/sparse-attention.md#L1
   - Framework-level sparse-attention implementation notes live in the developer guide: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/docs/source/developer-guide/sparse-attention-development-guide.md#L1
   - VisualGen sparse-attention docs are added as an unindexed draft page for the future VisualGen doc hub: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/docs/source/visual-gen/features/sparse-attention.md#L1
   - VisualGen Skip Softmax Attention tests now cover User API parsing, checkpoint lowering, ModelOpt `ignore` patterns, timestep cutoff behavior, and per-model checkpoint metadata: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tests/unittest/_torch/visual_gen/sparse_attention/test_skip_softmax.py#L78
   - LLM tests now cover Python/YAML API parsing, target sparsity validation, checkpoint formulas/defaults, `ignore` patterns, and direct sparse checkpoint config lowering: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/tests/unittest/llmapi/test_llm_args.py#L2359

## Follow-ups

- ModelOpt sparse-attention checkpoints use `ignore` in `config.json`; this PR treats that as the supported sparse layer-disable path. Existing `hf_quant_config.json` / `exclude_modules` handling should remain a quantization compatibility concern, not the primary sparse-attention interface.
- The VisualGen sparse-attention doc is intentionally not indexed yet. It is a draft location for colleagues to continue VisualGen-specific sparse-attention documentation after this PR lands.

## Test plan

- Passed: `pre-commit run --files tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py tensorrt_llm/_torch/modules/attention.py tensorrt_llm/llmapi/llm_args.py tests/unittest/_torch/visual_gen/sparse_attention/test_skip_softmax.py tests/unittest/llmapi/test_llm_args.py`
- Passed: commit-time pre-commit hooks for the latest commit.
- Earlier syntax checks passed on the updated test files with `python3 -m py_compile` and `git diff --check`.
- Not run locally: targeted pytest suites, because this local shell does not have the full TensorRT-LLM pytest/runtime dependency stack.
- CI on this branch is pending.


lfr-0531 (commented):
The sparse attention doc changes LGTM.

tensorrt_llm/visual_gen/sparse_attention.py
```diff
@@ -12,114 +12,121 @@
...
+    )
+
+    _formula: Optional["SkipSoftmaxFormula"] = PrivateAttr(default=None)
     _layer_overrides: Optional[Dict[str, float]] = PrivateAttr(default=None)
-    _component_configs: Optional[Dict[str, "SkipSoftmaxConfig"]] = PrivateAttr(default=None)
-    _resolved_threshold_prefill: Optional[float] = PrivateAttr(default=None)
+    _component_configs: Optional[Dict[str, "SkipSoftmaxAttentionConfig"]] = PrivateAttr(
+        default=None
+    )
+    _resolved_threshold: Optional[float] = PrivateAttr(default=None)
```
zhenhuaw-me: For LLM Sparse Attention, do we have some "internal class" that parse the config file and bake into the "internal fields"? I was expecting that we do the ckpt config parsing internally (inside `tensorrt_llm/_torch`) and make the code under `tensorrt_llm/visual_gen` to be public.

zhenhuaw-me (commented):
Thanks for the fix!

docs/source/developer-guide/sparse-attention-development.md
```diff
@@ -0,0 +1,282 @@
...
+  drives the core attention kernels. The op already understands
+  page-level sparsity for MQA/GQA in the generation phase, token-level
+  sparsity for MLA in both phases, and token-level KV compression in
+  the context phase for MQA/GQA.
+- **Auxiliary memory subsystem** — manages any extra pools (KT cache,
+  indexer K cache, …) alongside the main KV cache.
+
+<div align="center">
+<figure>
+  <img src="https://github.com/NVIDIA/TensorRT-LLM/raw/main/docs/source/media/sparse_attention_framework.png" width="800">
```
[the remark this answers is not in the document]
bobboli: Done in 058baf1188 — added descriptive `alt` text to both figures.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -411,67 +411,82 @@ class SkipSoftmaxAttentionConfig(BaseSparseAttentionConfig):
...
+    def to_kernel_params(self):
+        """Resolve to the kernel-facing ``SkipSoftmaxKernelParams``."""
+        from tensorrt_llm._torch.attention_backend.sparse.skip_softmax import \
+            SkipSoftmaxKernelParams
+        return SkipSoftmaxKernelParams(
+            threshold_scale_factor_prefill=self._phase(
+                self.threshold_scale_factor, 'prefill'),
+            threshold_scale_factor_decode=self._phase(
+                self.threshold_scale_factor, 'decode'),
+        )
```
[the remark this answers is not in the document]
bobboli: Fixed in 058baf1188. `to_kernel_params` now maps a missing phase to the kernel default (`0.0` = skip-softmax off), and `resolve_from_ckpt_config` only emits the phases `target_sparsity` requested (`{...} or None`), so it no longer writes `None` into a `Dict[str, float]`.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -411,67 +411,82 @@ class SkipSoftmaxAttentionConfig(BaseSparseAttentionConfig):
...
-        If threshold_scale_factor is already set, it takes precedence and
-        this method returns self unchanged.
+            {"formula": "a * exp(b * target_sparsity)",
+             "prefill": {"a": 7e-5, "b": 7.929109},
+             "decode":  {"a": 7e-5, "b": 16.9025}}
+
+        If ``formula`` is omitted, the canonical exponential form is
+        synthesized from the coefficient names (``{a, b}`` →
+        ``a * exp(b * target_sparsity)``; ``{log_a, b}`` →
+        ``exp(log_a + b * target_sparsity)``).
```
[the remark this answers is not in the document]
bobboli: Already addressed — the synthesized-fallback paragraph was removed when the method was renamed to `resolve_from_ckpt_config`. The current docstring no longer claims a fallback; an absent `formula` raises.

tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+
+    @model_validator(mode="after")
+    def _validate_formula(self):
+        try:
+            parsed = numexpr.NumExpr(self.formula)
+        except Exception as exc:
+            raise ValueError(
+                f"SkipSoftmaxFormula: invalid formula {self.formula!r}: {exc}"
+            ) from exc
```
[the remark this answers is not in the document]
bobboli: Intentionally keeping the broad `except`. numexpr's exception surface for invalid expressions is under-documented (your own web query notes additional internal error paths may exist), and this validator's sole job is to turn any unparseable formula into a clear `ValueError`. Narrowing risks letting an exotic numexpr error escape unwrapped.

tensorrt_llm/_torch/attention_backend/sparse/__init__.py
```diff
@@ -1,11 +1,25 @@
-from .utils import (get_flashinfer_sparse_attn_attention_backend,
-                    get_sparse_attn_kv_cache_manager,
-                    get_trtllm_sparse_attn_attention_backend,
-                    get_vanilla_sparse_attn_attention_backend)
+from typing import TYPE_CHECKING
+
+# ``.utils`` imports the attention backend and KV-cache manager, which would
```
yuxianq: We should use local import in `.utils` to WAR this issue instead of this hack.

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -752,7 +765,12 @@ class AttentionForwardArgs:
     is_fused_qkv: bool = False
     update_kv_cache: bool = True
 
-    sparse: AttentionSparseArgs = field(default_factory=AttentionSparseArgs)
+    sparse_prediction: SparsePrediction = field(
+        default_factory=SparsePrediction)
+
+    # Skip-softmax thresholds passed into the attention op.
+    skip_softmax_kernel_params: SkipSoftmaxKernelParams = field(
```
yuxianq: We don't need to add `skip_softmax_kernel_params` here, we can directly get skip softmax config from `TrtllmAttention`. `AttentionForwardArgs` is for per-call arguments.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1333,20 +1333,12 @@ def rope_original_max_positions(self) -> int:
...
-        return None
-
-    @property
-    def skip_softmax_threshold_scale_factor_decode(self) -> Optional[float]:
-        """Decode skip-softmax threshold; ``None`` unless
-        ``sparse_attention_config`` is a ``SkipSoftmaxAttentionConfig``."""
-        if isinstance(self.sparse_attention_config, SkipSoftmaxAttentionConfig):
-            return self.sparse_attention_config.threshold_scale_factor_decode
-        return None
+    def _skip_softmax_kernel_params(self) -> Optional[SkipSoftmaxKernelParams]:
```
yuxianq: We can directly create `self.skip_softmax_kernel_params` in TrtllmAttention's init instead of using property.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1591,18 +1583,22 @@ def _run(
...
                 rope_append=self.rope_append,
                 attention_chunk_size=self.attention_chunk_size,
-                skip_softmax_threshold_scale_factor_prefill=self.
-                skip_softmax_threshold_scale_factor_prefill,
-                skip_softmax_threshold_scale_factor_decode=self.
-                skip_softmax_threshold_scale_factor_decode,
+                skip_softmax_threshold_scale_factor_prefill=forward_args.
+                skip_softmax_kernel_params.threshold_scale_factor_prefill,
+                skip_softmax_threshold_scale_factor_decode=forward_args.
+                skip_softmax_kernel_params.threshold_scale_factor_decode,
```
yuxianq, line 1706: Use `self.skip_softmax_kernel_params.threshold_scale_factor_prefill` and `self.skip_softmax_kernel_params.threshold_scale_factor_decode` instead.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1690,6 +1686,12 @@ def forward(
                 get_indices_block_size(),
             )
 
+        # Resolve skip-softmax thresholds from the config. Any per-step
+        # adjustment would go here.
+        config_params = self._skip_softmax_kernel_params
+        if config_params is not None:
+            forward_args.skip_softmax_kernel_params = config_params
```
yuxianq: Don't set `forward_args.skip_softmax_kernel_params`, we can directly use the `self`'s one.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -542,9 +544,10 @@ def __init__(
...
-            formula = hf_sparse.get('threshold_scale_factor', {})
-            sparse_attn_cfg = sparse_attn_cfg.resolve_for_target_sparsity(
-                formula)
+            threshold_scale_factor_config = hf_sparse.get(
+                'threshold_scale_factor', {})
+            sparse_attn_cfg = sparse_attn_cfg.resolve_from_ckpt_config(
+                threshold_scale_factor_config)
 
         attn_cls = get_attention_backend(self.attn_backend,
                                          sparse_attn_config=sparse_attn_cfg)
```
yuxianq: `get_attention_backend`'s `sparse_attn_config`'s type is `Optional["SparseAttentionConfig"]`, but we do not import `SparseAttentionConfig` in `tensorrt_llm/_torch/attention_backend/utils.py`, please add import. Please also fix all other places that we use `SparseAttentionConfig` as type annotation but never import it.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: `forward_args.sparse` should be `forward_args.sparse_prediction` now. Please also check whether there are other out-of-date comments.

tensorrt_llm/_torch/attention_backend/interface.py
bobboli: I intentionally did this, since in the future we may want to have step-variant values passed into the kernel, e.g., in DiT we may want weaker threshold at initial denoising steps.

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: See https://github.com/NVIDIA/TensorRT-LLM/pull/14687#discussion_r3339644249

tensorrt_llm/visual_gen/sparse_attention.py
```diff
@@ -134,22 +141,37 @@ def _with_public_overrides(self, user_cfg: "SkipSoftmaxConfig") -> "SkipSoftmaxC
...
+        """Return the resolved scalar threshold, caching after the first call."""
+        if self._resolved_threshold is not None:
+            return self._resolved_threshold
         if (
             self._component_configs
             and self._formula is None
-            and self.threshold_scale_factor_prefill is None
+            and self.threshold_scale_factor is None
         ):
             return None
```
yuxianq, line 104: Seems unreachable. If `self._component_configs is not None`, `resolve_threshold` will go to L179 instead of `get_or_resolve_threshold`: https://github.com/NVIDIA/TensorRT-LLM/blob/d7cfb479a2c7b519ead386b4c69d5c178fb069cd/tensorrt_llm/visual_gen/sparse_attention.py#L179

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Kept as a property on purpose: VisualGen reassigns `self.sparse_attention_config` after construction (`apply_skip_softmax_overrides` sets it per layer), so a value cached in `__init__` would go stale. The property recomputes, and the cost is just one `to_kernel_params()` per forward.

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Intentional — routing through `forward_args.skip_softmax_kernel_params` is the seam for future per-step thresholds (a caller can set it to vary thresholds across steps), and co-locates skip-softmax inputs with the other per-forward sparse inputs. Today no caller overrides it, so `forward()` just fills it from the config — functionally the same as reading `self`, but it keeps that seam open.

tensorrt_llm/visual_gen/sparse_attention.py
bobboli: I am cleaning up visual_gen sparse_attention configs right now. To be revisited.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: If its value change after init, let's name it as `get_skip_softmax_kernel_params` like `get_quantize_output_dtype/get_local_layer_idx`, so that we won't assume it a static value and can add runtime value as its argument in the future.

tensorrt_llm/_torch/attention_backend/trtllm.py
yuxianq: Add comment so we or agents won't accidently remove it.

tensorrt_llm/_torch/attention_backend/sparse/__init__.py
bobboli: Solved in https://github.com/NVIDIA/TensorRT-LLM/pull/14687/commits/a0bdd019281b5bab6b727a130cd6f7c45ae2c5d7 — localized the backend/cache-manager imports inside the `.utils` functions and reverted `sparse/__init__.py` to an eager import (dropped the lazy `__getattr__`).

tensorrt_llm/_torch/modules/attention.py
bobboli: Solved in https://github.com/NVIDIA/TensorRT-LLM/pull/14687/commits/a0bdd019281b5bab6b727a130cd6f7c45ae2c5d7 — added the `SparseAttentionConfig` import where it was used only as an annotation: `attention_backend/utils.py`, `sparse/rocket.py`, `sparse/utils.py` (TYPE_CHECKING), and `model_loader.py` (existing `llm_args` import).

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Solved in https://github.com/NVIDIA/TensorRT-LLM/pull/14687/commits/a0bdd019281b5bab6b727a130cd6f7c45ae2c5d7 — fixed the stale `forward_args.sparse` comment to `forward_args.sparse_prediction`; grep confirms no other stale references.

tensorrt_llm/_torch/attention_backend/interface.py
zhenhuaw-me: @karljang can you point @bobboli and @yuxianq how VisualGen SkipSoftmax handles the step-variant values in https://github.com/NVIDIA/TensorRT-LLM/pull/12947? To be specific, for the first N steps we don't run SkipSoftmax, and for the rest ones we run SkipSoftmax.

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Added in https://github.com/NVIDIA/TensorRT-LLM/pull/14687/commits/98e021ecb3c5da451591f37b8d6ba55880a21ad4 — comment now explains the thresholds route through `forward_args` for per-step override and that the fill must stay even though it currently mirrors the config.

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Done in https://github.com/NVIDIA/TensorRT-LLM/pull/14687/commits/5b5666e7b3bc11db404659418a3e55207e3f72c6 — converted to a method `get_skip_softmax_kernel_params()` (matching `get_local_layer_idx`/`get_quantize_output_dtype`). Agreed it shouldn't be a property: it allocates a fresh params object each call and its value can change after init (VG swap), and this leaves room for a runtime arg later.

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Refined the comment in https://github.com/NVIDIA/TensorRT-LLM/pull/14687/commits/aac36a423f — now states the kernel-params/`sparse_attention_config` split is intentional (today a mirror; leaves room for per-step adjustment later).

tensorrt_llm/_torch/attention_backend/utils.py
```diff
@@ -1,9 +1,12 @@
-from typing import Optional, Type
+from typing import TYPE_CHECKING, Optional, Type
 
 import torch
 
 from tensorrt_llm.logger import logger
 
+if TYPE_CHECKING:
```
yuxianq: Should we add TYPE_CHECKING after all normal import blocks? Or each import block has its own TYPE_CHECKING block, so that we add this block only after absolute import block.

tensorrt_llm/visual_gen/args.py
zhenhuaw-me:
```suggestion
```

We should drop this API. cc @karljang

bobboli: I will drop it in subsequent commits.

tensorrt_llm/visual_gen/sparse_attention.py
```diff
@@ -12,370 +12,157 @@
...
+    target_sparsity: Optional[float] = PydanticField(
+        default=None,
+        ge=0.0,
+        le=1.0,
+        description="Semantic target sparsity in [0, 1]; requires a calibration formula.",
+    )
+    disabled_layers: Optional[list[str]] = PydanticField(
+        default=None,
+        description="Fnmatch patterns for layers where skip-softmax is disabled.",
+    )
```
zhenhuaw-me: Does it make sense to align with `QuantConfig.exclude_module`?

tensorrt_llm/visual_gen/sparse_attention.py
```diff
@@ -12,370 +12,157 @@
...
+    initial_disabled_steps: Optional[int] = PydanticField(
```
zhenhuaw-me: maybe name like `exclude_steps` or `skip_steps`?

tensorrt_llm/visual_gen/sparse_attention.py
```diff
@@ -12,370 +12,157 @@
...
+        default=None,
+        ge=0,
+        description="Number of initial denoising steps with skip-softmax disabled.",
+    )
+
+    def to_sparse_params(self, **kwargs):
```
zhenhuaw-me, line 71: I am not sure about the general TRTLLM convention, but I would prefer to have an internal utils that can merge `SkipSoftmaxAttentionConfig` and the ckpt config into something like SkipSoftmaxParams.

tensorrt_llm/visual_gen/sparse_attention.py
bobboli: Addressed in the current design. The public config remains the user-facing VG surface, but checkpoint parsing/calibration primitives live under `_torch/attention_backend/sparse/skip_softmax.py`, and the backend receives lowered `SkipSoftmaxParams` only. I also removed the unused `_torch/visual_gen/sparse_attention.py` compatibility wrapper in 6ca60eb1ac.

tensorrt_llm/_torch/attention_backend/interface.py
bobboli: Resolved. We no longer pass skip-softmax kernel params through `AttentionForwardArgs`; `TrtllmAttention` derives them from its own `self.sparse_params`. In 9ed7191d1a the per-call signal is the normalized `timestep`, used only to select the scheduler phase.

tensorrt_llm/visual_gen/sparse_attention.py
bobboli: Resolved by the VisualGen sparse config cleanup. The old component-config path is gone; per-component checkpoint calibration now comes from each model's `pretrained_config`, and lowering returns either `None` for excluded modules or `SkipSoftmaxParams` for enabled modules.

tensorrt_llm/_torch/attention_backend/utils.py
bobboli: Leaving this as-is intentionally. These imports are annotation-only and guarded by `TYPE_CHECKING` to avoid runtime coupling to both the LLM and VisualGen public config classes. No runtime dependency is introduced here.

tensorrt_llm/visual_gen/args.py
bobboli: Resolved. The standalone VisualGen sparse config path API has been dropped; sparse calibration is now discovered from checkpoint `config.json` metadata or supplied directly through the sparse attention config.

tensorrt_llm/visual_gen/sparse_attention.py
bobboli: Done in 5c09b363a8. The VisualGen public API field is now `exclude_modules`, aligned with the established `QuantConfig.exclude_modules` naming. ModelOpt checkpoint metadata still reads `ignore` from `sparse_attention_config.config_groups`.

tensorrt_llm/visual_gen/sparse_attention.py
bobboli: Done in 9ed7191d1a. I replaced `initial_disabled_steps` with `disabled_until_timestep`, a normalized `[0, 1]` timestep cutoff. Skip-softmax is disabled while `timestep >= disabled_until_timestep`; this removes the skip-softmax-specific `step_index` attention plumbing.

tensorrt_llm/visual_gen/sparse_attention.py
bobboli: Settled in the current design. `to_sparse_params()` stays on the LLM/VG `SkipSoftmaxAttentionConfig` classes because their user-facing config surfaces differ and can lower differently. The actual lowered backend object is still `SkipSoftmaxParams`; the unused `_torch/visual_gen/sparse_attention.py` wrapper was removed in 6ca60eb1ac rather than kept as an empty merge helper.

tensorrt_llm/visual_gen/sparse_attention.py
```diff
@@ -12,370 +12,188 @@
...
+        default=None,
+        description="Raw per-block threshold; takes precedence over target_sparsity.",
+    )
+    target_sparsity: Optional[float] = PydanticField(
+        default=None,
+        ge=0.0,
+        le=1.0,
+        description="Semantic target sparsity in [0, 1]; requires a calibration formula.",
+    )
+    disabled_until_timestep: Optional[float] = PydanticField(
```
zhenhuaw-me, line 64: LGTM in general but leave a comment here to block the merging thus we can align on the API.

zhenhuaw-me (approved):
Thanks for driving the refactoring! LGTM from the VisualGen changes. Leave a comment here to block the merging in case any unexpected operation.

karljang:
@bobboli, thank you for driving this effort!
The Visual Gen part looks good overall, except for the `disabled_until_timestep` naming convention.
I’m trying the PR branch on the checkpoint that still uses `initial_disabled_steps`.

docs/source/visual-gen/features/sparse-attention.md
```diff
@@ -0,0 +1,160 @@
+# VisualGen Sparse Attention
```
chang-l, line 1: Normally, we put all visual generation-related feature docs in `docs/source/models/visual-generation.md`, including parallelization, attention quantization, etc.

If we think sparse attention deserves a separate doc, we should add a note or link there to direct users to this page.

tensorrt_llm/_torch/visual_gen/models/cosmos3/pipeline_cosmos3.py
```diff
@@ -581,7 +586,7 @@ def forward_fn(
             """
             noise_pred = self.transformer(
                 hidden_states=latent_input,
-                timestep=timestep,
+                timestep=timestep / self.scheduler.config.num_train_timesteps,
```
chang-l, line 589: why do we make this change?
@NVShreyas  can you please take a look?

tensorrt_llm/_torch/visual_gen/pipeline.py
```diff
@@ -193,7 +193,11 @@ def _setup_cuda_graphs(self):
             if model is None:
                 continue
 
-            runner = CUDAGraphRunner(CUDAGraphRunnerConfig(use_cuda_graph=True), shared_pool)
+            runner = CUDAGraphRunner(
+                CUDAGraphRunnerConfig(use_cuda_graph=True),
+                shared_pool,
+            )
+            model.register_cuda_graph_extra_key_fns(runner)
```
chang-l, line 200: can you explain how cuda graph work in this case? to capture both graphs, do we need to run pass across the dense-sparse boundary?

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -678,17 +687,19 @@ def forward(
         if rope is not None:
             freqs_cos, freqs_sin = rope
 
-        # Time and text/image embeddings
+        # Time and text/image embeddings. WAN timestep embeddings use the
+        # scheduler's 1000-step scale internally.
+        timestep_for_embedding = timestep * 1000
```
chang-l, line 692: **Timestep-normalization divisor is inconsistent across models.** All VG transformers now expect a normalized `[0,1]` timestep, but the normalization differs:
- Flux2 & Qwen-Image hardcode `timestep / 1000` (`pipeline_flux2.py:452`, `pipeline_qwen_image.py:493`).
- WAN & Cosmos use `timestep / self.scheduler.config.num_train_timesteps` (`pipeline_wan.py:554`, `pipeline_cosmos3.py:586`).

And here WAN divides by `num_train_timesteps` in the pipeline but multiplies back by a hardcoded `* 1000` — these only round-trip when `num_train_timesteps == 1000`. If a checkpoint sets `num_train_timesteps != 1000`, WAN's embedding timestep would be wrong and the `[0,1]` skip-softmax cutoff would drift across models.

Suggest unifying on a single source (a shared `normalize_timestep()` helper / always `num_train_timesteps`) and documenting the `[0,1]` contract in one place.

chang-l: i think it seems related to my previous comment – why hard code to 1000 for some models?

it seems you are using `num_train_steps` (not actual inference steps) to normalize and determine the wamrup steps, right? if so — say `disabled_until_timestep=0.2` with `num_inference_steps=40` —  does it translate to warmup (dense) steps is 8 (40*0.2)?

also, keep @karljang informed in this discussion

docs/source/visual-gen/features/sparse-attention.md
```diff
@@ -0,0 +1,160 @@
...
+
+Each component reads its own `config.json`, so formulas and `ignore` patterns can differ between `transformer` and `transformer_2`.
+
+### User Configuration
+
+User configuration is supplied through Python or YAML and controls how the checkpoint metadata is consumed:
+
+- Set `threshold_scale_factor` directly to pass a concrete threshold to the kernel. This does not require checkpoint calibration metadata.
+- Set `target_sparsity` to request a sparsity target. The runtime resolves it to `threshold_scale_factor` using the checkpoint calibration formula. If the checkpoint does not provide the required Skip Softmax Attention metadata, the runtime raises an error.
+- Set `disabled_until_timestep` to disable Skip Softmax Attention at the beginning of denoising. This cutoff is normalized and therefore independent of the user-selected number of denoising steps.
```
chang-l, line 89: I guess if we choose to use a normalized scale [0, 1], we should make it clear what it is normalized against and how it translates to the actual timesteps (which most users understand better).

karljang (commented):
_Surfaced by an AI coding agent (Claude) while testing this branch against a real ModelOpt checkpoint; repro and both fixes verified manually on an H200._

Two issues hit when running the decoupled VisualGen skip-softmax path on a phase-keyed WAN 2.2 checkpoint (flagged inline). Both are pure-Python; with both fixes the checkpoint generates end-to-end (720p / 81 frames, `target_sparsity=0.5`).

1. **Scalar formula reader ignores phase-keyed coefficients** (`skip_softmax.py`) — reads a flat `coefficients` key, but ModelOpt checkpoints nest them under the phase key (`prefill`). Fails config resolution with `ValueError: target_sparsity requires calibration formula coefficients`.
2. **Incomplete rename** — `AttentionForwardArgs.sparse` -> `sparse_prediction` (`interface.py`) leaves a stale caller in `trtllm_gen.py` `is_supported()` (`fwd.sparse...`), raising `AttributeError` at warmup.

**Repro:** VisualGen run on a diffusers WAN 2.2 checkpoint whose per-component `config.json` carries `sparse_attention_config.config_groups.*.threshold_scale_factor` with phase-keyed `prefill: {a, b}` coefficients; `SkipSoftmaxAttentionConfig(algorithm="skip_softmax", target_sparsity=0.5)`, TRTLLM backend. Without fix #1 it fails at config resolution; with #1 only, it fails at warmup with #2.

tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py
```diff
@@ -0,0 +1,425 @@
...
+        return None
+
+    threshold_config = skip_softmax_threshold_scale_factor_config_from_ckpt_sparse_attention_config(
+        ckpt_sparse_attention_config
+    )
+    if not isinstance(threshold_config, dict) and "formula" in ckpt_sparse_attention_config:
+        threshold_config = ckpt_sparse_attention_config
+    if isinstance(threshold_config, dict):
+        return SkipSoftmaxFormula.parse_from_dict(
+            threshold_config.get("coefficients"),
```
karljang, line 265: This reads only the flat `coefficients` key, but ModelOpt checkpoints nest the coefficients under the **phase key**, mirroring their phase-keyed `target_sparsity`:

```jsonc
"target_sparsity": { "prefill": 0.5 },
"threshold_scale_factor": { "formula": "a * exp(b * target_sparsity)", "prefill": { "a": 1443.49, "b": 4.30 } }
```

So `get("coefficients")` -> `None` -> formula `None` -> the VisualGen resolve path raises `ValueError: target_sparsity requires calibration formula coefficients` on a fully-calibrated checkpoint.

The scheduler path below (`SkipSoftmaxScheduler.from_threshold_scale_factor`) already handles this (`threshold_config.get(phase)` then flat fallback) — the scalar reader just needs the same:

```python
coefficients = (
    threshold_config.get("coefficients")
    or threshold_config.get("prefill")
    or threshold_config.get("decode")
)
return SkipSoftmaxFormula.parse_from_dict(coefficients, formula=threshold_config.get("formula"))
```

(Related: the VG path also rejects the checkpoint's own phase-dict `target_sparsity` as "prefill/decode phase dictionaries are LLM-only", so a scalar `target_sparsity` must be passed in the user config — worth documenting or relaxing.)

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -766,8 +769,11 @@ class AttentionForwardArgs:
 
     is_fused_qkv: bool = False
     update_kv_cache: bool = True
+    # Optional normalized diffusion timestep for timestep-varying sparse attention.
+    timestep: Optional[torch.Tensor] = None
 
-    sparse: AttentionSparseArgs = field(default_factory=AttentionSparseArgs)
+    sparse_prediction: SparsePrediction = field(
```
karljang, line 873: This rename (`AttentionForwardArgs.sparse` -> `sparse_prediction`, and `AttentionSparseArgs` -> `SparsePrediction`) misses a caller: `trtllm_gen.py` `is_supported()` still reads the old name (unchanged by this PR, so it's outside the diff):

```python
# tensorrt_llm/_torch/attention_backend/trtllm_gen.py  (~L608)
sparse_kv_indices = fwd.sparse.sparse_kv_indices
sparse_attn_indices = fwd.sparse.sparse_attn_indices
```

At warmup this raises:

```
AttributeError: 'AttentionForwardArgs' object has no attribute 'sparse'
```

Fix the stale caller to match the new field name:

```python
sparse_kv_indices = fwd.sparse_prediction.sparse_kv_indices
sparse_attn_indices = fwd.sparse_prediction.sparse_attn_indices
```

tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py
karljang: It’s possible that the checkpoint I used is outdated. If that’s the case, please provide the path, and I’ll test it again. :)

docs/source/developer-guide/sparse-attention-development-guide.md
```diff
@@ -0,0 +1,282 @@
...
+```python
+class MySparseAttentionConfig(BaseSparseAttentionConfig):
+    algorithm: Literal["my_algo"] = "my_algo"
+    topk: int = 64
+    # ... other parameters
+```
+
+Add the new class to the discriminated `SparseAttentionConfig` union at
+the bottom of the file.
+
```
heyuhhh, line 226: Could you please add an introduction about new added lower sparse params here? Thanks!

heyuhhh (approved):
Thanks for the great refactoring efforts! The sparse attention related changes totally LGTM.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
bobboli: Good catch. The WAN num_train_timesteps vs hardcoded 1000 round-trip is exactly one of the issues in the normalized-timestep design discussion we are starting internally. I will not treat the current wording/implementation as final until we align on whether the forward contract should be normalized and how each model should convert to its embedding scale.

docs/source/visual-gen/features/sparse-attention.md
bobboli: Yes, the docs need to match the final semantic choice. We are starting an internal discussion on whether VisualGen should use normalized timestep as the transformer-forward contract, and what the user-facing cutoff name should be. I will revise this section after that decision instead of patching the wording piecemeal.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
bobboli: Agreed that this needs an explicit design decision rather than another one-off model rule. We are initiating an internal discussion on the VisualGen timestep contract: normalized [0, 1] at transformer.forward vs raw scheduler/model-specific timesteps. I will update the PR once we align on the contract and naming.

tensorrt_llm/_torch/visual_gen/models/cosmos3/pipeline_cosmos3.py
bobboli: This is part of a broader VisualGen timestep-contract question: whether transformer.forward should consistently receive normalized timesteps in [0, 1], with each model converting to its own embedding scale internally. We are initiating an internal discussion with VisualGen owners before locking this down. Depending on that decision, I will either keep this with clearer docs/contract, or narrow the PR and adjust the Cosmos change accordingly.

docs/source/visual-gen/features/sparse-attention.md
bobboli: Added a link from `docs/source/models/visual-generation.md` to the VisualGen sparse attention page in commit `4e1cd15da9`. The broader VisualGen doc hierarchy/structure is still being discussed internally, so I kept this PR to a lightweight cross-link rather than moving the draft page.

tensorrt_llm/_torch/visual_gen/pipeline.py
bobboli: Added a dedicated VisualGen CUDA Graphs doc that explains capture scope, graph keys, extra keys, and lazy capture/replay behavior. It also uses Skip Softmax Attention as the concrete extra-key example for the dense/sparse boundary: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/docs/source/visual-gen/features/cuda-graph.md

docs/source/developer-guide/sparse-attention-development-guide.md
bobboli: Added in commit `3d3f3a9321`: the developer guide now has a `Lowered sparse parameters` section that explains the user config -> `SparseParams` / `SparseMetadataParams` boundary, including `SparseParams` for `AttentionBackend` and the corresponding `SparseMetadataParams` for `AttentionMetadata`.

tensorrt_llm/_torch/attention_backend/interface.py
bobboli: Fixed in commit `8eed8dcf41`: `trtllm_gen.py` now reads `fwd.sparse_prediction.sparse_kv_indices` and `fwd.sparse_prediction.sparse_attn_indices`.

tensorrt_llm/_torch/attention_backend/sparse/skip_softmax.py
bobboli: For the current VisualGen checkpoint format, the source of truth is the VisualGen sparse attention doc: https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/docs/source/visual-gen/features/sparse-attention.md#checkpoint-config

VisualGen uses scalar `target_sparsity` / `threshold_scale_factor` metadata rather than LLM `prefill` / `decode` phase dictionaries. If the checkpoint used for testing still has phase-keyed `prefill` coefficients, that checkpoint shape is outdated for the current VG contract.

docs/source/visual-gen/features/sparse-attention.md
chang-l: thanks @bobboli

docs/source/visual-gen/features/sparse-attention.md
chang-l: Thanks @bobboli  for helping with the refactoring and for leading the overall skipsoftmax effort!

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -593,10 +593,9 @@ def is_supported(
         fwd: AttentionForwardArgs,
     ) -> Tuple[bool, str]:
         is_mla_enable = attn.is_mla_enable
-        has_skip_softmax = (
-            attn.skip_softmax_threshold_scale_factor_prefill is not None
-            or attn.skip_softmax_threshold_scale_factor_decode is not None
-        )
+        sparse_params = attn.sparse_params
+        has_skip_softmax = getattr(sparse_params, "algorithm", None) == "skip_softmax"
+        has_sparse_attention = sparse_params is not None and not has_skip_softmax
```
yuxianq, line 598: @yihwang-nv Does flashinfer trtllm-gen support skip softmax? If not, we should also exclude skip softmax.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1471,7 +1467,8 @@ def yarn_get_mscale(scale=1, mscale=1):
             hidden_size=self.hidden_size,
             predicted_tokens_per_seq=self.predicted_tokens_per_seq,
             skip_create_weights_in_init=config.skip_create_weights_in_init,
-            sparse_attention_config=config.sparse_attention_config,
+            attn_cls=mqa_cls,
```
yuxianq, line 1498: Instead of calling `get_attention_backend` to get `attn_cls` and pass it to `create_attention`, can we make `get_attention_backend` accepts `sparse_params` instead of `sparse_attention_config`, so that we don't need to pass `attn_cls` here and can call `get_attention_backend` inside `create_attention`.

tensorrt_llm/_torch/attention_backend/utils.py
```diff
@@ -61,15 +73,19 @@ def create_attention(
...
 ):
     if attention_chunk_size is not None and backend_name.upper() != "TRTLLM":
         raise ValueError(
             f"Backend {backend_name} does not support chunked attention.")
+    if sparse_params is not None and attn_cls is None:
+        raise ValueError("attn_cls is required when sparse_params is set.")
 
-    attn_cls = get_attention_backend(backend_name, sparse_attention_config)
+    if attn_cls is None:
+        attn_cls = get_attention_backend(backend_name)
```
yuxianq, line 88: We can make `get_attention_backend` accepts `sparse_params` and remove `attn_cls` argument. `get_attention_backend` just uses `algorithm`, which can also get from `SparseParams`.

tensorrt_llm/_torch/attention_backend/utils.py
```diff
@@ -89,18 +105,23 @@ def create_attention(
...
         pos_embd_params=pos_embd_params,
         mla_params=mla_params,
         skip_create_weights_in_init=skip_create_weights_in_init,
         attention_chunk_size=attention_chunk_size,
-        sparse_attention_config=sparse_attention_config,
         dtype=dtype,
         aux_stream=aux_stream,
     )
+    if sparse_params is not None:
+        kwargs["sparse_params"] = sparse_params
```
yuxianq, line 119: Can we still pass all arguments in any cases? attention backends just ignore it if unsupported.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1279,14 +1267,19 @@ def __init__(
...
 
+        config = config or ModelConfig()
+        sparse_attn_cfg = config.sparse_attention_config
+        sparse_params = (sparse_attn_cfg.to_sparse_params(
+            pretrained_config=config.pretrained_config,
+            layer_idx=self.layer_idx) if sparse_attn_cfg is not None else None)
+
         # Currently only DSA sparse attention is supported.
-        if config is not None and config.sparse_attention_config is not None and config.sparse_attention_config.algorithm == "dsa":
+        if getattr(sparse_params, "algorithm", None) == "dsa":
```
yuxianq, line 1305: Use `sparse_params.algorithm` is enough, `algorithm` is required fields.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
```diff
@@ -593,10 +593,9 @@ def is_supported(
         fwd: AttentionForwardArgs,
     ) -> Tuple[bool, str]:
         is_mla_enable = attn.is_mla_enable
-        has_skip_softmax = (
-            attn.skip_softmax_threshold_scale_factor_prefill is not None
-            or attn.skip_softmax_threshold_scale_factor_decode is not None
-        )
+        sparse_params = attn.sparse_params
+        has_skip_softmax = getattr(sparse_params, "algorithm", None) == "skip_softmax"
```
yuxianq, line 597: Use `sparse_params.algorithm` is enough, `algorithm` is required fields.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1731,11 +1732,17 @@ def _set_up_attn_metadata(
...
+            self.sparse_attention_config.to_sparse_metadata_params(
+                pretrained_config=config)
+            if self.sparse_attention_config is not None else None)
+
         if kv_cache_manager is None:
             # Cache the no-cache metadata.
             if self.encoder_attn_metadata is not None:
                 return self.encoder_attn_metadata
-            self.encoder_attn_metadata = self.attn_backend.Metadata(
+            metadata_kwargs = dict(
```
yuxianq, line 1751: Unnecessary to pack and unpack `metadata_kwargs` here, we can call `metadata_cls` directly.

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1758,7 +1766,7 @@ def _set_up_attn_metadata(
             assert self.attn_metadata.kv_cache_manager is kv_cache_manager
             return self.attn_metadata
 
-        self.attn_metadata = self.attn_backend.Metadata(
+        metadata_kwargs = dict(
```
yuxianq, line 1775: Unnecessary to pack and unpack `metadata_kwargs` here, we can call `metadata_cls` directly.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1588,10 +1575,18 @@ def _run(
                 fwd=forward_args,
             )
         else:
+            sparse_params = self.sparse_params
+            skip_softmax_kernel_params = (
```
yuxianq, line 1577: Please put `skip_softmax_kernel_params` into `forward_args`. We should avoid pass any forward local variable to `thop.attention`. We are refactoring `trtllm_gen_backend` and `thop.attention` to Fmha class (see https://github.com/NVIDIA/TensorRT-LLM/pull/15204), put `skip_softmax_kernel_params` in `forward_args` so that every fmha lib can access it. We don't have `AttentionForwardArgs` subclass for each attention backend, so backend specific info like `skip_softmax_kernel_params` will leak to shared `AttentionForwardArgs`, I may create `TrtllmAttentionForwardArgs` in a follow-up PR.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -660,7 +665,11 @@ def forward(
             3. All-gather output sequence: [B, S/P] -> [B, S]
 
         When TeaCache is enabled, TeaCacheHook intercepts and replaces this call.
+
+        Args:
+            timestep: Normalized scheduler timestep tensor in [0, 1].
         """
+        del kwargs  # Kept for diffusers API compatibility.
```
yuxianq, line 672: Don't need to del unused arguments.

tensorrt_llm/tools/layer_wise_benchmarks/runner.py
```diff
@@ -606,9 +606,17 @@ def create_run_pack(
...
+        )
+        metadata_cls = AttentionCls.Metadata
+        sparse_attention_config = self.model_config.sparse_attention_config
+        sparse_metadata_params = (
+            sparse_attention_config.to_sparse_metadata_params(pretrained_config=pretrained_config)
+            if sparse_attention_config is not None
+            else None
         )
-        attn_metadata = AttentionCls.Metadata(
+        metadata_kwargs = dict(
```
yuxianq, line 619: Unnecessary to pack and unpack `metadata_kwargs`, directly call `metadata_cls`

tests/unittest/_torch/attention_backend/test_attention_op_sync.py
```diff
@@ -56,6 +57,7 @@
     "self": TrtllmAttention,
     "metadata": TrtllmAttentionMetadata,
     "forward_args": AttentionForwardArgs,
+    "skip_softmax_kernel_params": SkipSoftmaxKernelParams,
```
yuxianq, line 60: Don't need to add new source if we put `skip_softmax_kernel_params` in `forward_args`

tensorrt_llm/visual_gen/sparse_attention.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 1 (VisualGen timestep/API contract). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/visual_gen/models/cosmos3/pipeline_cosmos3.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 1 (VisualGen timestep/API contract). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 1 (VisualGen timestep/API contract). I will address it there so this PR can stay focused on the current refactor.

docs/source/visual-gen/features/sparse-attention.md
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 1 (VisualGen timestep/API contract). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 2 (Sparse backend selection and forward-args cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/modules/attention.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 2 (Sparse backend selection and forward-args cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/attention_backend/utils.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 2 (Sparse backend selection and forward-args cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/attention_backend/utils.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 2 (Sparse backend selection and forward-args cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/modules/attention.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 2 (Sparse backend selection and forward-args cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/attention_backend/trtllm_gen.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 2 (Sparse backend selection and forward-args cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/pyexecutor/model_engine.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 3 (Metadata construction cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/pyexecutor/model_engine.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 3 (Metadata construction cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/attention_backend/trtllm.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 2 (Sparse backend selection and forward-args cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 4 (WAN unused-argument cleanup). I will address it there so this PR can stay focused on the current refactor.

tensorrt_llm/tools/layer_wise_benchmarks/runner.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 3 (Metadata construction cleanup). I will address it there so this PR can stay focused on the current refactor.

tests/unittest/_torch/attention_backend/test_attention_op_sync.py
bobboli: Tracked in follow-up PR https://github.com/NVIDIA/TensorRT-LLM/pull/15417, item 2 (Sparse backend selection and forward-args cleanup). I will address it there so this PR can stay focused on the current refactor.

karljang (approved):
LGTM; I couldn't test it out because we don't have a checkpoint with new config.json~

chang-l (approved):

yuxianq (approved):

PR #14708, opened 2026-05-28
[None][perf] Reduce OpenAI stream postprocess overhead

2ez4bz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming responses now include consistent IDs and timestamps across all message chunks for improved data consistency.

* **Tests**
  * Added test coverage for streaming response metadata consistency validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14708?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Description

Streaming responses created fresh metadata for each chunk when callers did not pass it explicitly. High-concurrency workloads can emit hundreds of thousands of chunks, which makes UUID generation, and time lookups part of the CPU hot path.

Reuse stream metadata per request.

### Details

Prior to this, the stream postprocessors created each SSE chunk (e.g. `ChatCompletionStreamResponse`) without passing `id` or `created`, thus calling their factories `uuid.uuid64` and `time.time`, respectively, so every streamed chunk got fresh metadata.

Semantically, this was wrong - all chunks for one streamed OpenAI response should have the same `id` and `created` timestamp.

As a side-effect, this also benefits performance: at high concurrency on lower-tier CPUs, this change can by itself lead to a ~5% ITL decrease.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/serve/postprocess_handlers.py
+++ b/tensorrt_llm/serve/postprocess_handlers.py
@@ -90,6 +90,8 @@ class ChatPostprocArgs(PostprocArgs):
     tool_call_id_type: str = "random"
     chat_template_kwargs: Optional[dict[str, Any]] = None
     ctx_usage: Optional[UsageInfo] = None
+    # Cache per-request stream metadata so every chunk reuses the same response
+    # id and created timestamp instead of regenerating them for each chunk.
     stream_response_id: Optional[str] = None
     stream_created: Optional[int] = None
 
@@ -466,6 +468,8 @@ class CompletionPostprocArgs(PostprocArgs):
     return_logprobs: bool = False
     stream_options: Optional[StreamOptions] = None
     ctx_usage: Optional[UsageInfo] = None
+    # Cache per-request stream metadata so every chunk reuses the same response
+    # id and created timestamp instead of regenerating them for each chunk.
     stream_response_id: Optional[str] = None
     stream_created: Optional[int] = None
 
--- a/tensorrt_llm/serve/postprocess_handlers.py
+++ b/tensorrt_llm/serve/postprocess_handlers.py
@@ -649,6 +649,8 @@ class ChatCompletionPostprocArgs(PostprocArgs):
     stream_options: Optional[StreamOptions] = None
     chat_template_kwargs: Optional[dict[str, Any]] = None
     ctx_usage: Optional[UsageInfo] = None
+    stream_response_id: Optional[str] = None
+    stream_created: Optional[int] = None
 
     @classmethod
     def from_request(cls, request: ChatCompletionRequest):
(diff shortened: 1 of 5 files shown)


tensorrt_llm/serve/postprocess_handlers.py
```diff
@@ -89,6 +90,8 @@ class ChatPostprocArgs(PostprocArgs):
     tool_call_id_type: str = "random"
     chat_template_kwargs: Optional[dict[str, Any]] = None
     ctx_usage: Optional[UsageInfo] = None
+    stream_response_id: Optional[str] = None
```
hchings, line 95: nit: a brief comment on what these two fields are for, similar to the MR description would be helpful.

2ez4bz: Done.

tensorrt_llm/serve/postprocess_handlers.py
```diff
@@ -500,6 +526,8 @@ def completion_stream_post_processor(rsp: DetokenizedGenerationResultBase,
     res: List[str] = []
     prompt_tokens = args.num_prompt_tokens
     ctx_usage = _ctx_usage_for_postproc(args, rsp.outputs)
+    stream_response_id, stream_created = _ensure_stream_metadata(
```
yechank-nvidia, line 529: It looks like multi-prompt streaming completions can still emit multiple stream ids/created timestamps within one HTTP response.

In `openai_completion`, when request.prompt is a list, we create a separate `CompletionPostprocArgs` for each prompt and then merge all per-prompt generators into one StreamingResponse via merge_generators(). Since `_ensure_stream_metadata()` caches metadata on each args object and derives the id from each sub-request rsp.id, a request like prompt=["A", "B"], stream=true can produce chunks with ids like cmpl-<rsp0.id> and cmpl-<rsp1.id> in the same SSE stream.

If the intended invariant is “all chunks for one streamed OpenAI response share the same id/created”, we should seed one stream_response_id/stream_created before the prompt loop and copy it into every per-prompt `CompletionPostprocArgs`. It  would also be useful to add a test for list-prompt streaming completions.

This does not look like a new regression from this PR, but it seems like the fix is incomplete for multi-prompt streaming completions.

tensorrt_llm/serve/postprocess_handlers.py
```diff
@@ -229,6 +245,8 @@ def yield_first_chat(num_tokens: int,
     finish_reason_sent = [False] * args.num_choices
     prompt_tokens = args.num_prompt_tokens
     ctx_usage = _ctx_usage_for_postproc(args, rsp.outputs)
+    stream_response_id, stream_created = _ensure_stream_metadata(
```
yechank-nvidia, line 248: This fixes the standard chat stream processor, but the Harmony chat path still constructs ChatCompletionStreamResponse without id/created, so GPT-OSS streaming keeps the old behavior. Should this helper/shared metadata be applied there too, or should the PR scope/description be narrowed to the non-Harmony path?

tensorrt_llm/serve/postprocess_handlers.py
2ez4bz: Good catch - should be fixed now.

tensorrt_llm/serve/postprocess_handlers.py
2ez4bz: Ditto.

hchings (approved):

PR #14711, opened 2026-05-28
[None][feat] Support Step-3.7-Flash model

kaiyux:
## Release Notes

* **New Features**
  * Added support for Step3p7 models (text-only and multimodal variants)
  * Extended MTP support to Step3p7 and Step3p5 models and other architectures with native MTP modules
  * Added MMMU evaluation post-processing options

* **Bug Fixes**
  * Fixed CUDA graph attention workspace handling to prevent crashes during replay
  * Improved speculative decoding configuration loading from model checkpoints

* **Documentation**
  * Updated supported models documentation and MTP compatibility information

* **Tests**
  * Added Step3p7 accuracy, parity, and LLM API integration tests
  * Added unit tests for Step3p7 configuration and weight processing

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14711?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Description

Adds support for **StepFun Step-3.7-Flash** to the PyTorch backend, covering both the text-only and vision-language variants, plus native MTP speculative decoding and a few supporting fixes.

### New model

- **Text decoder** — `tensorrt_llm/_torch/models/modeling_step3p7.py`
  - `Step3p7ForCausalLM` (also registered as `Step3p5ForCausalLM` for standalone text use). Hybrid architecture: per-layer attention geometry with partial RoPE, plus a routed MoE block with a shared expert and a sigmoid router with learnable bias (`routed_scaling_factor` applied in `Step3p7MoE.forward`).
  - Native MTP module `Step3p7MTP` / `Step3p7MTPHead` for speculative decoding, built on `SpecDecOneEngineForCausalLM`.
- **Multimodal** — `tensorrt_llm/_torch/models/modeling_step3p7vl.py`
  - `Step3p7VLForConditionalGeneration` (registered as the HF arch `Step3p7ForConditionalGeneration`) with its own `Step3p7VLInputProcessor`. Supports text + image (`L + I`). The PerceptionEncoder vision tower runs in BF16 even when the text decoder is quantized (FP8 block-scale or NVFP4), and can load from either the FP8 or NVFP4 export. The input processor exposes multimodal-hashing hooks so image embeddings participate in KV-cache reuse.

### Supporting changes

- **MTP / speculative config**
  - `model_loader.py`: whenever a `speculative_config` is set, call `update_spec_config_from_model_config(...)` so MTP resolves its layer count and draft length from the loaded checkpoint config before the model runs.
  - `speculative/utils.py` (`update_spec_config_from_model_config`): set `num_nextn_predict_layers` from the model config; when the user left `max_draft_len` unset (`None`), resolve it from the checkpoint (vanilla MTP → all MTP heads, MTP-Eagle → 1); when the user set it but it disagrees with the checkpoint for vanilla MTP, clamp to the smaller value and warn; mirror the result into `max_total_draft_tokens`.
  - `llm_args.py` (`MTPDecodingConfig`): stop defaulting `max_draft_len` to `1`. Leave it `None` as the "use the checkpoint's `num_nextn_predict_layers`" sentinel, only validating (`> 0`) and mirroring to `max_total_draft_tokens` when the user supplies it — so downstream code can distinguish user-set from defaulted values.
  - `modeling_speculative.py`: wire `step3p7` / `step3p5` model types into `MTPForCausalLM` (→ `Step3p7MTP`).
  - `speculative/mtp.py` (`MTPWorker`): clamp generation-request `kv_lens_cuda` to at least `mtp_num_modules` in place (CUDA-graph safe). The tiny dummy sequences used during generation-step warmup could otherwise drive `kv_len < q_len`, giving the attention kernel a negative effective KV range and an illegal memory access (seen with Step3p7 MTP + dense sliding-window attention).
- **CUDA graph attention workspace fix** (`model_engine.py`): add `_presize_attention_cuda_graph_workspace()`, run after warmup and before capture. The TRTLLM C++ attention op resizes its workspace in place when the requested size grows; inside a CUDA graph capture, `resize_()` frees the old storage so kernels captured earlier hold a dangling pointer and segfault on replay. Pre-sizing `cuda_graph_workspace` to the warmed-up eager `workspace` size eliminates the in-capture resize.
- **MMMU evaluation post-processing** (`lm_eval.py`): add `--post_process_fn strip_thinking_mmmu` (resolves to `strip_thinking_and_extract_mmmu_answer`, which strips `<think>...</think>` before the MMMU answer extractor) and `--preserve_caller_max_tokens` (keeps a larger `--max_output_length` than MMMU's default `max_gen_toks=512`) for thinking models (e.g. Kimi K2.5, Step3p7) whose chain-of-thought breaks the default lm-eval regex.
- **flashinfer MoE op backend** (`moe_op_backend.py`): map `MiniMax2` / `SigmoidRenorm` routing-method types, used by Step3p7's sigmoid router.
- **Docs**: add `Step3p7ForConditionalGeneration` to the supported-models table and to both the model-feature and multimodal-feature support matrices, and update the speculative-decoding MTP note (MTP is no longer DeepSeek-only — it now covers any architecture shipping native MTP modules, including Step-3.x).

### Unit tests

- `tests/unittest/_torch/modeling/test_modeling_step3p7.py` — text decoder + MTP: stacked MoE weight splitting (per-expert key expansion, NVFP4 suffixes, no-op fallback), MTP weight rewriting to the MTP-block layout, multimodal `language_model.*` key flattening / prefix stripping, NVFP4 batched dequant round-trip, MTP-head pre-output normalization, AutoModel registration, plus real-checkpoint config/weight accounting, per-layer attention geometry helpers, and MTP spec-config defaulting to the checkpoint layer count. (Runs in L0 on A10.)
- `tests/unittest/_torch/modeling/test_modeling_step3p7vl.py` — vision tower + VLM wiring: RoPE-2D frequency cache / identity / dynamic-subgrid selection, LayerScale, fused-QKV vision attention layout (+ head-divisibility guard), encoder forward shape and positional-embedding interpolation, projector geometry and encode→text-hidden projection, weight-loading prefix splitting, VLM / vision-encoder / input-processor registration, vision-config geometry, and input-processor hooks (special and multimodal token ids, tokens-per-image, contiguous image span). (Runs in L0 on L40S.)
- `tests/unittest/llmapi/test_llm_args.py::test_MTPDecodingConfig_default_draft_len_is_not_user_set` — verifies an unset `max_draft_len` stays `None` (and out of `model_fields_set`), while an explicit value mirrors to `max_total_draft_tokens`.

### Integration accuracy tests

- `tests/integration/defs/accuracy/test_llm_api_pytorch.py::TestStep3_7` — text-decoder GSM8K: `test_auto_dtype` (BF16, TP8/EP8), `test_fp8_block_scales` (TP4/EP4, `mtp_nextn ∈ {0, 3}`, KVCacheManagerV2, TRTLLM attention + MoE backends), and `test_nvfp4` (Blackwell-only, TP4/EP4, FP8 KV cache). Reference scores in `tests/integration/defs/accuracy/references/gsm8k.yaml`.
- `tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py::TestStep3_7` — multimodal MMMU: `test_fp8_block_scales` and `test_nvfp4` (Blackwell), using the strip-thinking post-processor + `preserve_caller_max_tokens`. Reference scores in `tests/integration/defs/accuracy/references/mmmu.yaml`.

### CI registration

- **L0**: the FP8 + MTP GSM8K case (`test_fp8_block_scales[...-mtp_nextn=3]`) on `l0_dgx_h100.yml`; the two unit suites on `l0_a10.yml` (text) and `l0_l40s.yml` (vision).
- **QA** (`llm_function_core.txt`): the remaining integration cases — text BF16 (TP8), FP8 without MTP, NVFP4, and both multimodal cases.

## PR Checklist

## GitHub Bot Help


tests/integration/defs/accuracy/test_step3p7_gsm8k_canary.py
```diff
@@ -0,0 +1,477 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+r"""Step3p7 GSM8K accuracy canary driver.
```
QiJune: Please remove this file, and add a class `class TestStep3_7(LlmapiAccuracyTestHarness):`in test_llm_api_pytorch.py instead.

tests/integration/defs/models/test_step3p7_llm_api.py
```diff
@@ -0,0 +1,316 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Step3p7 LLM API smoke-test driver.
```
QiJune: Please remove this `tests/integration/defs/models` directory.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -1665,11 +1665,17 @@ def set_max_total_draft_tokens(self):
         # Default max_draft_len to 1 if not set by the user.
         # For vanilla MTP, update_spec_config_from_model_config will override this
         # with the actual num_nextn_predict_layers from the model.
+        # Keep model_fields_set limited to fields the user supplied.
         if self.max_draft_len is None:
-            self.max_draft_len = 1
+            object.__setattr__(self, "max_draft_len", 1)
```
QiJune: tensorrt_llm/llmapi/llm_args.py:1665-1674 — consider removing the validator-side default instead of bypassing __setattr__                                                                                                      
                                                                                                                   
The object.__setattr__ trick here exists to keep model_fields_set clean so that _prepare_step3p7_mtp_spec_config (modeling_step3p7.py:1433) can distinguish "user explicitly set max_draft_len=1" from "validator defaulted it to  1". That's a real distinction, but the root cause is that the validator defaults None → 1 eagerly — once you do that, you need a workaround to recover the "is this user-set?" signal you just destroyed.                        
                                                                                                                 
This is a behavior change for every MTP config, not just Step3p7 (DeepSeek-V3 also runs through MTPDecodingConfig). Anything downstream that inspects model_fields_set after the validator runs will silently see a different set. I'd prefer not to ship that as a side effect of a model-add PR.                                
                                                                                                                 
Suggested alternative: drop the defaulting from the validator, and let the only consumer that needs a concrete value (the model-specific spec-config prep) do it explicitly:

```python                                                                                                                  
  # llm_args.py — validator only validates, no defaulting                                                        
  @model_validator(mode="after")                                                                                   
  def set_max_total_draft_tokens(self):
      if self.max_draft_len is not None:                                                                           
          if self.max_draft_len <= 0:                                                                            
              raise ValueError("max_draft_len must be > 0 for MTP")                                                
          self.max_total_draft_tokens = self.max_draft_len
      return self                                                                                                  
                                                                                                                 
  # modeling_step3p7.py::_prepare_step3p7_mtp_spec_config
  if spec_config.max_draft_len is None:                                                                            
      spec_config.max_draft_len = model_layers
  elif spec_config.max_draft_len > model_layers:                                                                   
      spec_config.max_draft_len = model_layers                                                                   
  spec_config.max_total_draft_tokens = spec_config.max_draft_len                                                   
```                                                                                                                   
  Benefits:
  - No object.__setattr__ bypass — model_fields_set carries its natural meaning.                                   
  - None becomes the explicit "use model default" sentinel, no need for "max_draft_len" in model_fields_set        
  anywhere.                                                                                                
  - DeepSeek's update_spec_config_from_model_config already follows the same "override when not user-set" pattern; aligning here keeps both models on one mechanism.                                                               
  - The new test_MTPDecodingConfig_default_draft_len_is_not_user_set becomes simpler:  MTPDecodingConfig().max_draft_len is None.

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -297,6 +297,14 @@ def load_config_and_apply_defaults(
 
         config = checkpoint_loader.load_config(checkpoint_dir, **config_kwargs)
 
+        if (llm_args.speculative_config is not None and hasattr(
```
QiJune: tensorrt_llm/_torch/pyexecutor/model_loader.py:300-307 — drop this; it duplicates the existing call at model_engine.py:415


update_spec_config_from_model_config is already invoked unconditionally for every spec-decode path at model_engine.py:415 (inside if self.is_spec_decode:). Adding it here makes it run twice for every MTP model, DeepSeek-V3 included. 

The call timeline:

```python
  py_executor_creator.py:238 → load_config_and_apply_defaults    ← this PR adds the call here
  model_engine.py:288        → ModelLoader.load(...)              ← model is constructed
  model_engine.py:415        → update_spec_config_from_model_config  ← existing call
```

  The new early call doesn't enable anything downstream:

```python
  - Step3p7ForCausalLM.__init__ already has its own _prepare_step3p7_mtp_spec_config which reads
  model_config.pretrained_config.num_nextn_predict_layers directly — it doesn't depend on
  spec_config.num_nextn_predict_layers being populated first.
  - The model_config.py:841 assert (num_nextn_predict_layers must be set ... before building ModelConfig) is
  checked only inside get_bindings_model_config, which runs after model_engine.__init__ — i.e. after the existing
  call at line 415 — so no early-fill is needed there either.
```

Side effects of the double call:

1. Two model_fields_set pollutions (the function uses normal assignment), which interacts badly with the model_fields_set-based "user_set" check that _prepare_step3p7_mtp_spec_config also does.
2. For non-Step3p7 MTP models, an extra silently-added cross-model behaviour change in a PR scoped to "add Step3p7".

Suggested fix: remove this block entirely. If there's a specific downstream reader between load_config_and_apply_defaults and model_engine.py:415 that needs spec_config.num_nextn_predict_layers populated, please name it in the PR description and gate the call on that consumer rather than on hasattr(pretrained_config, "num_nextn_predict_layers") — the hasattr trigger applies to every native-MTP model, not just Step3p7.

nv-guomingz (approved):

tensorrt_llm/llmapi/llm_args.py
kaiyux: Updated.

tensorrt_llm/_torch/pyexecutor/model_loader.py
kaiyux: This is not the case after addressing the previous comment https://github.com/NVIDIA/TensorRT-LLM/pull/14711#discussion_r3321472460, if remove the block, there will be `TypeError`
```
File "/xxx/TensorRT-LLM/tensorrt_llm/llmapi/llm_args.py", line 1030, in tokens_per_gen_step
    return 1 + self.max_total_draft_tokens
           ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

tensorrt_llm/_torch/speculative/utils.py
```diff
@@ -401,9 +404,12 @@ def update_spec_config_from_model_config(spec_config, model_config):
...
             spec_config.max_draft_len = model_layers
-        spec_config.max_total_draft_tokens = spec_config.max_draft_len
-    # For Eagle MTP (1 MTP layer): max_draft_len controls how many times the
-    # single layer is run. It was already set by the user (defaults to 1).
+    else:
+        # Eagle MTP (1 MTP layer): max_draft_len controls how many times the
+        # single layer is run; default to 1 when the user didn't set it.
+        if spec_config.max_draft_len is None:
+            spec_config.max_draft_len = 1
+    spec_config.max_total_draft_tokens = spec_config.max_draft_len
```
ziyixiong-nv:
```suggestion
      is_vanilla = spec_config.spec_dec_mode.is_mtp_vanilla()                                                                                                                                                                                      
                                                                                                                                                                                                                                                   
      # Resolve max_draft_len when the user didn't set it:                                                                                                                                                                                         
      #   vanilla MTP -> use all checkpoint MTP heads                                                                                                                                                                                              
      #   MTP-Eagle   -> replay the single head once
      if spec_config.max_draft_len is None:
          spec_config.max_draft_len = (
              spec_config.num_nextn_predict_layers if is_vanilla else 1)                                                                                                                                                                           
      elif is_vanilla and spec_config.max_draft_len != spec_config.num_nextn_predict_layers:                                                                                                                                                        
          effective_draft_len = min(spec_config.max_draft_len,                                                                                                                                                                                      
                                    spec_config.num_nextn_predict_layers)                                                                                                                                                                           
          logger.warning(                                                                                                                                                                                                                           
              f"MTP: max_draft_len ({spec_config.max_draft_len}) does not match "
              f"num_nextn_predict_layers ({spec_config.num_nextn_predict_layers}); "                                                                                                                                                                
              f"using max_draft_len={effective_draft_len} draft tokens."                                                                                                                                                                            
          )                                                                                                                                                                                                                                         
          spec_config.max_draft_len = effective_draft_len                                                                                                                                                                                           
                                                                                                                                                                                                                                                    
      spec_config.max_total_draft_tokens = spec_config.max_draft_len
```

tensorrt_llm/_torch/pyexecutor/model_loader.py
```diff
@@ -297,6 +297,14 @@ def load_config_and_apply_defaults(
...
+                config.pretrained_config, "num_nextn_predict_layers")):
```
ziyixiong-nv:
```suggestion
        if llm_args.speculative_config is not None:
```

tensorrt_llm/_torch/models/modeling_step3p7vl.py
```diff
@@ -0,0 +1,946 @@
...
+def _apply_rotary_emb(freqs: torch.Tensor, t: torch.Tensor) -> torch.Tensor:
+    dtype = t.dtype
+    rot_dim = freqs.shape[-1]
+    t_rot = t[..., :rot_dim]
+    t_pass = t[..., rot_dim:]
+    t_rot = (t_rot * freqs.cos()) + (_rotate_half(t_rot) * freqs.sin())
+    return torch.cat((t_rot, t_pass), dim=-1).to(dtype)
+
+
+class _Step3VisionRope2D(nn.Module):
```
yechank-nvidia: Nit: I'm not a fan of adding _ on the class name. Can you remove all _ from the class name for VL?

yechank-nvidia (commented):
This PR adds and documents image support for `Step3p7ForConditionalGeneration`, but the test coverage is still text/config/weight-accounting only.

Do you have plan to add multimodal-related tests? 

Also, multimodal hashing looks unsupported/broken for `Step3p7` right now. `Step3p7VLInputProcessor` does not implement `get_num_tokens_per_image`(), `get_mm_token_ids`(), or `get_mm_special_token_ids`(), which are needed by the generic mm hashing path. Since Step3 image spans include structural tokens such as <patch_start>, <patch_end>, <patch_newline>, <im_start>, and <im_end>, the processor should either implement these hooks correctly.

kaiyux:
@yechank-nvidia Thanks for help review.

> Do you have plan to add multimodal-related tests?

Yes will do.

> Also, multimodal hashing looks unsupported/broken for Step3p7 right now. Step3p7VLInputProcessor does not implement get_num_tokens_per_image(), get_mm_token_ids(), or get_mm_special_token_ids(), which are needed by the generic mm hashing path. Since Step3 image spans include structural tokens such as <patch_start>, <patch_end>, <patch_newline>, <im_start>, and <im_end>, the processor should either implement these hooks correctly.

Let me take a look at that, however if that's the case, is there a way to add some kind of checkers? Otherwise I guess developers like me who are not aware of that context will miss, but the model can still run.

tensorrt_llm/_torch/models/modeling_step3p7vl.py
kaiyux: Updated.

tests/integration/defs/accuracy/test_step3p7_gsm8k_canary.py
kaiyux: Updated.

yechank-nvidia:
Hi @kaiyux, we have multimodal claude [skills.md](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.claude/skills/trtllm-model-onboard-multimodal/SKILL.md) you can take a look at.

Also, for the test-related, we have MMMU test(e.g. [EXAONE-4.5](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py#L104-L126)) and with MMMU, you can explicitly set with chunked_prefill.

Also, we have the skeleton [test_modeling_multimodal.py](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/_torch/modeling/test_modeling_multimodal.py) which you can do logits matching with HF or sanity check on cuda_graph, kv_cache_reuse, and then multiple modalities.

kaiyux:
> Hi @kaiyux, we have multimodal claude [skills.md](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.claude/skills/trtllm-model-onboard-multimodal/SKILL.md) you can take a look at.
> 
> Also, for the test-related, we have MMMU test(e.g. [EXAONE-4.5](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/defs/accuracy/test_llm_api_pytorch_multimodal.py#L104-L126)) and with MMMU, you can explicitly set with chunked_prefill.
> 
> Also, we have the skeleton [test_modeling_multimodal.py](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/_torch/modeling/test_modeling_multimodal.py) which you can do logits matching with HF or sanity check on cuda_graph, kv_cache_reuse, and then multiple modalities.

@yechank-nvidia Thanks for the information, I'll take a look. By saying add some kind of checkers, I meant if there is a way to raise errors if the required "get_num_tokens_per_image(), get_mm_token_ids(), or get_mm_special_token_ids()" hooks are not implemented, that's another topic though.

tensorrt_llm/_torch/speculative/utils.py
kaiyux: this way removes the clamp that enforces `max_draft_len <= num_nextn_predict_layers` for vanilla MTP, would that be an issue?

ziyixiong-nv: You're right. Updated to use `!=` rather than `<`.

kaiyux: Updated. Thanks.

tensorrt_llm/_torch/pyexecutor/model_loader.py
kaiyux: Updated.

ziyixiong-nv (approved):
Looks good for the speculative decoding changes.

tests/integration/test_lists/test-db/l0_dgx_b200.yml
```diff
@@ -52,6 +52,9 @@ l0_dgx_b200:
   - unittest/_torch/visual_gen/multi_gpu
   - unittest/_torch/visual_gen/test_flux_pipeline.py::TestFluxParallelism::test_ulysses_2gpu_correctness
   - unittest/_torch/visual_gen/test_flux_pipeline.py::TestFluxCombinedOptimizations::test_all_optimizations_combined
+  # ------------- Step-3.7-Flash (MoE VLM): text GSM8K + multimodal MMMU (NVFP4) ---------------
+  - accuracy/test_llm_api_pytorch.py::TestStep3_7::test_nvfp4[tp_size=4-ep_size=4] TIMEOUT (90)
```
StanleySun639: @kaiyux could you please add into [QA test list](https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/integration/test_lists/qa/llm_function_core.txt) to gain more HW coverage?

StanleySun639 (approved):

tensorrt_llm/_torch/models/modeling_step3p7vl.py
```diff
@@ -0,0 +1,1026 @@
...
+            norm_key = key[len("model.") :] if key.startswith("model.") else key
+            if norm_key.startswith("vision_model."):
+                sub = norm_key[len("vision_model.") :]
+                vision_state[sub] = weights[key]
+            elif norm_key.startswith("vit_large_projector."):
+                sub = norm_key[len("vit_large_projector.") :]
+                projector_state[sub] = weights[key]
+
+        if vision_state:
+            missing, unexpected = self.vision_model.load_state_dict(vision_state, strict=False)
```
yechank-nvidia: Confirming question) Is there a checkpoint variant that requires `strict=False` here? In the `Step3p7ForConditionalGeneration` path, the vision tower is required, so if vision keys are missing or prefix remapping is wrong, continuing with randomly initialized vision weights could cause a silent accuracy failure.

tensorrt_llm/_torch/models/modeling_step3p7vl.py
```diff
@@ -0,0 +1,1026 @@
...
+        """Project post-downsample vision features to the text hidden size."""
+        return self.vit_large_projector(image_features.to(self._dtype))
+
+    def _encode(self, pixel_values: torch.Tensor) -> torch.Tensor:
+        feats = self.vision_model(pixel_values)
+        return self._process_image_features(feats)
+
+    @nvtx_range("Step3p7VisionTower forward()")
+    @torch.inference_mode()
+    def forward(self, multimodal_params: List[MultimodalParams]) -> List[torch.Tensor]:
```
yechank-nvidia, line 577: `Step3p7VisionTower.forward()` loops over `multimodal_params` and runs `_encode()` per request, and separately for full images vs patches.

Can we batch full images across all requests, batch patch images across all requests where shapes match, and then split/reassemble according to `num_patches`?

tensorrt_llm/_torch/models/modeling_step3p7vl.py
```diff
@@ -0,0 +1,1026 @@
...
+        q = q.view(bsz, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
+        k = k.view(bsz, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
+        v = v.view(bsz, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
+        if self.rope is not None:
+            q, k = self.rope(q, k, grid_hw=grid_hw)
+        attn_output = F.scaled_dot_product_attention(q, k, v, is_causal=False, scale=self.scale)
+        attn_output = attn_output.transpose(1, 2).reshape(
+            bsz, seq_len, self.num_heads * self.head_dim
+        )
+        return self.out_proj(attn_output)
```
yechank-nvidia, line 252: This implementation keeps the Step3 vision tower on `F.linear` / `F.scaled_dot_product_attention` / plain `nn.Linear` (`Step3VisionAttention` and `vit_large_projector`). 

Since this is a 47-layer PerceptionEncoder that runs during multimodal prefill, this likely leaves TRT-LLM attention, TP/quantized Linear, and CUDA-graph/static-shape optimizations on the table. 

Can we either port the vision attention/projector to TRT-LLM modules, or document why `Step3p7` needs to be an exception and include perf validation?

tensorrt_llm/_torch/models/modeling_step3p7vl.py
kaiyux: Updated to `strict=True`.

tensorrt_llm/_torch/models/modeling_step3p7vl.py
kaiyux: Added a `TODO` for the following PR.

tensorrt_llm/_torch/models/modeling_step3p7vl.py
kaiyux: Added a `TODO` for the following PR.

tests/integration/test_lists/test-db/l0_dgx_b200.yml
kaiyux: Confirmed offline that no need to add for now. Thanks.

tests/integration/test_lists/test-db/l0_l40s.yml
```diff
@@ -23,6 +23,7 @@ l0_l40s:
   - unittest/_torch/modeling/test_modeling_qwen2_5vl.py::TestQwen2_5_VL::test_all
   - unittest/_torch/modeling/test_modeling_qwen3vl_moe.py::TestQwen3VLMoe::test_all
   - unittest/_torch/modeling/test_modeling_qwen3vl.py::TestQwen3VL::test_all
+  - unittest/_torch/modeling/test_modeling_step3p7vl.py
```
QiJune, line 27: Is this test duplicated with a10? If so, let's remove this one.

tests/integration/test_lists/test-db/l0_dgx_b200.yml
```diff
@@ -52,6 +52,9 @@ l0_dgx_b200:
...
+  - accuracy/test_llm_api_pytorch_multimodal.py::TestStep3_7::test_nvfp4 TIMEOUT (120)
```
QiJune: Considering the key model list defined by product team, I suggest we only keep one test case(e.g. enable MTP) in dgx_h200/dgx_h100/dgx_h200, and put all others to post merge(or drop directly).

QiJune (approved):
LGTM

tests/integration/test_lists/test-db/l0_l40s.yml
kaiyux: That one is `test_modeling_step3p7.py` for the text part, so not duplicated.

tests/integration/test_lists/test-db/l0_dgx_b200.yml
kaiyux: Moved to QA test list

zhaoyangwang-nvidia (approved):
LGTM

kaiyux:
/bot skip --comment "skipping since the only failing test case was moved to qa in the latest commit"

StanleySun639 (approved):

PR #14714, opened 2026-05-29
[https://nvbugs/6221483][fix] AutoDeploy: Fix Eagle metadata host syncs

### LINKED ISSUE #14838: [Bug]: AutoDeploy: Fix race from non-blocking D2H copy in Eagle + make copy finer-grained.
### Background

This issue originally surfaced as **NVBUG-6221483**. That bug was mitigated by reverting a different PR, but the revert only papered over the symptom — it removed the unrelated commit that exposed the latent problem rather than fixing the problem itself. This issue tracks the **underlying root cause** so the bug does not resurface the next time an unrelated change perturbs the metadata-sync path.

### Root cause

In the AutoDeploy Eagle/MTP one-model draft loop, the in-graph metadata update helpers (`SequenceInfo.switch_to_generate_` and `SequenceInfo.offset_pos_and_cache_`) mirror **all** active host-side args back to host after updating device tensors. In the draft loop this is both incorrect and fragile:

- Target-only host mirrors (not read by the draft submodule) can be left in stale staging values or trigger unnecessary D2H copies.
- The draft loop needs host mirroring scoped to only the args the draft submodule's `forward` actually consumes. Because the broad sync happened to produce correct-enough state in the original ordering, an unrelated commit that shifted that ordering was enough to surface the latent bug.

Additionally, the D2H `InputBuffer.copy_to_host` used `non_blocking=True` unconditionally, which can race when the host values are read on the CPU immediately after the copy.

### Fix

- Add an `active_args_override` filter to `switch_to_generate_` / `offset_pos_and_cache_` so host mirroring is limited to the next consumer's graph-input placeholders. Override names that are not active graph args (e.g. inter-module tensors `inputs_embeds` / `hidden_states`) are tolerated and ignored.
- Make `copy_to_host` accept a `non_blocking` flag and invoke it synchronously (`non_blocking=False`) from the metadata-update path.
- Convert the unsupported FlashInfer + CUDA-graph + speculative-decoding combination from a silent fallback to `compile_backend='torch-simple'` into an explicit `ValueError`.

### Component

AutoDeploy (`tensorrt_llm/_torch/auto_deploy`)

### Tracking

Originally surfaced in NVBUG-6221483 (previously mitigated by a revert of an unrelated PR).

govind-ramnarayan:
fixes: #14838 

## Summary
- Make AutoDeploy D2H metadata mirroring blocking so reused pinned host buffers are safe under overlap scheduling.
- Add `active_args_override` for metadata mutation helpers so callers can narrow host mirroring to the next consumer's graph inputs.
- Use the draft model placeholders in the Eagle draft loop, avoiding target-only Mamba host metadata syncs during drafting.

## Validation
- `CUDA_VISIBLE_DEVICES=0 PYTHONPATH=$PWD pytest -sv tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py`
- `PYTHONPATH=$PWD LLM_MODELS_ROOT=/lustre/fs1/portfolios/coreai/projects/coreai_comparch_autodeploy/autodeploy_data/llm-models-fake CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 pytest -sv tests/integration/defs/accuracy/test_llm_api_autodeploy.py::TestNemotronSuperV3::test_mtp[fp8_ws8_80gb-trtllm]`

## Summary by CodeRabbit

* **Refactor**
  * Optimized device-to-host synchronization in KV-cache inference to selectively sync only required buffers, improving performance for graph-transformed model inference.

* **Tests**
  * Added test coverage to verify selective synchronization behavior during inference operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14714?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
+++ b/tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
@@ -295,7 +295,7 @@ class InputBuffer:
                     trunc_h_buf = self._trunc_host_bufs[name][:copy_bytes]
                     trunc_d_buf.copy_(trunc_h_buf, non_blocking=True)
 
-    def copy_to_host(self) -> None:
+    def copy_to_host(self, non_blocking: bool = False) -> None:
         """Copy from device buffer to host buffer.
 
         Mirrors ``copy_to_device``: uses the current length of the truncatable tensor
@@ -306,7 +306,7 @@ class InputBuffer:
             if self._total_bytes > 0:
                 h_buffer = self._host_buffer[: self._total_bytes]
                 d_buffer = self._device_buffer[: self._total_bytes]
-                h_buffer.copy_(d_buffer, non_blocking=True)
+                h_buffer.copy_(d_buffer, non_blocking=non_blocking)
 
             # Copy each truncatable tensor independently, truncated to current length
             for name in self._truncatable_names:
@@ -316,7 +316,7 @@ class InputBuffer:
                     copy_bytes = length * dtype.itemsize
                     trunc_d_buf = self._trunc_device_bufs[name][:copy_bytes]
                     trunc_h_buf = self._trunc_host_bufs[name][:copy_bytes]
-                    trunc_h_buf.copy_(trunc_d_buf, non_blocking=True)
+                    trunc_h_buf.copy_(trunc_d_buf, non_blocking=non_blocking)
 
     def resize(self, name: str, new_capacity: int) -> None:
         """Resize a truncatable tensor's capacity.
@@ -1175,6 +1175,27 @@ class SequenceInfo:
         """
         return self._is_active(name, check_both) or self._is_active_host_prep(name, check_both)
 
+    def _active_host_update_args(
+        self, arg_names: Set[str], active_args_override: Optional[Set[str]] = None
+    ) -> List[str]:
+        """Return host args that need mirroring after an in-graph metadata update.
+
+        ``active_args_override`` lets a caller narrow host mirroring to the graph inputs the next
+        consumer actually reads. It is treated as a filter: only active host args whose names appear
+        in the override are mirrored. The override may contain names that are not active graph args
+        (e.g. a submodule's full placeholder set, which also includes inter-module tensors such as
+        ``inputs_embeds``/``hidden_states``); such entries are simply ignored. The caller is
+        responsible for including every host argument the next consumer may read.
+        """
+        needs_d2h_sync = [
+            k + self._host_suffix
+            for k in arg_names
+            if self._is_active(k + self._host_suffix, check_both=False)
+        ]
+        if active_args_override is None:
+            return needs_d2h_sync
+        return [arg_name for arg_name in needs_d2h_sync if arg_name in active_args_override]
+
     def _stage_arg(
         self,
         name: str,
@@ -1583,11 +1604,16 @@ class SequenceInfo:
             host_function(**{arg: self.get_arg(arg) for arg in args})
 
     @nvtx_range("ad_offset_pos_and_cache_")
-    def offset_pos_and_cache_(self, offset: torch.Tensor) -> None:
+    def offset_pos_and_cache_(
+        self, offset: torch.Tensor, active_args_override: Optional[Set[str]] = None
+    ) -> None:
         """Offset position and cache-related metadata for active arguments.
 
         Args:
             offset: 1D tensor [batch_size] with per-sequence position offsets.
+            active_args_override: Optional graph-input names for the next in-forward consumer. When
+                provided, host mirroring is limited to those active host args. The caller is
+                responsible for including every host argument the next consumer may read.
         """
         # check if we need a d2h sync
         _REQUIRES_UPDATE = {
@@ -1599,11 +1625,7 @@ class SequenceInfo:
             "seq_len_with_cache",
             "use_initial_states",
         }
-        needs_d2h_sync = [
-            k + self._host_suffix
-            for k in _REQUIRES_UPDATE
-            if self._is_active(k + self._host_suffix, check_both=False)
-        ]
+        needs_d2h_sync = self._active_host_update_args(_REQUIRES_UPDATE, active_args_override)
         sync_to_host = any(needs_d2h_sync)
         if sync_to_host:
             ad_logger.debug(f"d2h sync required in offset_pos_and_cache_ for {needs_d2h_sync}")
@@ -1694,7 +1716,7 @@ class SequenceInfo:
         # TODO: May need to dissect what fields are needed in the forward pass to reduce
         # data movement.
         if sync_to_host:
-            self._input_buffer.copy_to_host()
+            self._input_buffer.copy_to_host(non_blocking=False)
 
     @nvtx_range("ad_offset_with_new_lens_")
     def offset_with_new_lens_(self, new_lens_ungathered: torch.Tensor) -> None:
@@ -1718,12 +1740,17 @@ class SequenceInfo:
         self.offset_pos_and_cache_(increment)
 
     @nvtx_range("ad_switch_to_generate_")
-    def switch_to_generate_(self) -> None:
+    def switch_to_generate_(self, active_args_override: Optional[Set[str]] = None) -> None:
         """Switch all sequences metadata to generate (decode) mode.
 
         Transitions the batch from any layout (prefill/extend/decode or mixed) to
         an all-decode layout where each sequence has exactly 1 token. We assume that we just take
         the last position of each sequence for the metadata.
+
+        Args:
+            active_args_override: Optional graph-input names for the next in-forward consumer. When
+                provided, host mirroring is limited to those active host args. The caller is
+                responsible for including every host argument the next consumer may read.
 
         NOTE: update device tensors first and mirror back to host only when an updated host-side
         argument is active.
@@ -1757,11 +1784,7 @@ class SequenceInfo:
             "position_ids",
             "use_initial_states",
         }
-        needs_d2h_sync = [
-            k + self._host_suffix
-            for k in _REQUIRES_UPDATE
-            if self._is_active(k + self._host_suffix, check_both=False)
-        ]
+        needs_d2h_sync = self._active_host_update_args(_REQUIRES_UPDATE, active_args_override)
         sync_to_host = any(needs_d2h_sync)
 
         # --- input_ids (device) ---
@@ -1790,7 +1813,7 @@ class SequenceInfo:
         # TODO: May need to dissect what fields are needed in the forward pass to reduce
         # data movement.
         if sync_to_host:
-            self._input_buffer.copy_to_host()
+            self._input_buffer.copy_to_host(non_blocking=False)
 
     def copy_(self, name: str, src: torch.Tensor, strict: bool = True) -> None:
         """Copy a tensor into the buffer. USE WITH CAUTION!
--- a/tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
+++ b/tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
@@ -54,7 +54,11 @@ def _make_seq_info(extra_activate=()) -> SequenceInfo:
 
 
 def _nest_prefill(si: SequenceInfo, input_ids, pages_per_seq, cache_loc, **kw):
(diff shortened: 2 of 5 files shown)


govind-ramnarayan:
Note: fold in an explicit error when trying to run flashinfer + Eagle + cudagraph (instead of silently changing to torch-simple)

tests/unittest/auto_deploy/singlegpu/shim/test_llm_config.py
```diff
@@ -235,6 +235,45 @@ def test_accepts_mtp_eagle_one_model(self):
...
+
+        spec_config = EagleDecodingConfig(
+            max_draft_len=3,
+            speculative_model="some/model",
+            eagle3_one_model=True,
+        )
+
+        with pytest.raises(
+            pydantic.ValidationError,
+            match="Speculative decoding with FlashInfer attention does not currently support CUDA graph replay",
```
govind-ramnarayan: this sort of matching is too brittle - don't do this.

tests/unittest/auto_deploy/singlegpu/shim/test_llm_config.py
```diff
@@ -235,6 +235,45 @@ def test_accepts_mtp_eagle_one_model(self):
...
+        )
+
+        args = LlmArgs(
+            model="test-model",
+            speculative_config=spec_config,
+            attn_backend="flashinfer",
+            compile_backend="torch-simple",
+        )
+
+        assert args.compile_backend == "torch-simple"
```
govind-ramnarayan: the assertion is not necessary

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
```diff
@@ -423,6 +440,25 @@ def test_non_native_host_seq_len_with_cache_syncs(self):
         assert host_swc[0].item() == 4
         assert host_swc[1].item() == 5
 
+    def test_out_of_scope_active_host_mirror_not_synced_by_offset_pos_and_cache(self):
```
govind-ramnarayan, line 532: let's add another test that the host mirror is synced when it is in scope, for both functions.

tests/unittest/auto_deploy/singlegpu/shim/test_llm_config.py
govind-ramnarayan: Done. Removed the brittle `match=...` assertion and now only assert that the invalid config raises `pydantic.ValidationError`.

tests/unittest/auto_deploy/singlegpu/shim/test_llm_config.py
govind-ramnarayan: Done. Removed the unnecessary assertion; the test now just verifies that the explicit `torch-simple` configuration is accepted.

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
govind-ramnarayan: Done. Added positive scoped-sync coverage for both `switch_to_generate_()` and `offset_pos_and_cache_()` when the relevant host placeholder is included in `active_args_override`. Also added coverage that override names must be active graph args.

govind-ramnarayan:
Review follow-up pushed in `7cff2196d0`.

Updates:
- Removed brittle error-message matching in the FlashInfer + speculative + CUDA graph config test.
- Removed the redundant `torch-simple` assertion.
- Added positive scoped-sync tests for both `switch_to_generate_()` and `offset_pos_and_cache_()`.
- Added validation that `active_args_override` is a subset of active graph args, with unit coverage for both mutation APIs.

QA list update not required: this PR adds focused unittest coverage only and does not add, remove, or rename integration tests/test-list entries.

govind-ramnarayan:
/bot run --stage-list "A10-Build_Docs, A10-PackageSanityCheck-PY310-UB2204, A100X-PackageSanityCheck-PY312-UB2404, A30-AutoDeploy-1, H100_PCIe-AutoDeploy-1, DGX_B200-AutoDeploy-1, A100X-PyTorch-1, DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-Post-Merge-1"

govind-ramnarayan:
/bot run --stage-list "A10-Build_Docs, A10-PackageSanityCheck-PY310-UB2204, A100X-PackageSanityCheck-PY312-UB2404, A30-AutoDeploy-1, H100_PCIe-AutoDeploy-1, DGX_B200-AutoDeploy-1, A100X-PyTorch-1, DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-Post-Merge-1"

govind-ramnarayan:
/bot run --stage-list "A10-Build_Docs, A10-PackageSanityCheck-PY310-UB2204, A100X-PackageSanityCheck-PY312-UB2404, A30-AutoDeploy-1, H100_PCIe-AutoDeploy-1, DGX_B200-AutoDeploy-1, A100X-PyTorch-1, DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-Post-Merge-1"

govind-ramnarayan:
/bot run --stage-list "A10-Build_Docs, A10-PackageSanityCheck-PY310-UB2204, A100X-PackageSanityCheck-PY312-UB2404, A30-AutoDeploy-1, H100_PCIe-AutoDeploy-1, DGX_B200-AutoDeploy-1, A100X-PyTorch-1, DGX_H100-4_GPUs-AutoDeploy-1, DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-Post-Merge-1"

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -295,7 +295,7 @@ def copy_to_device(self) -> None:
                     trunc_h_buf = self._trunc_host_bufs[name][:copy_bytes]
                     trunc_d_buf.copy_(trunc_h_buf, non_blocking=True)
 
-    def copy_to_host(self) -> None:
+    def copy_to_host(self, non_blocking: bool = False) -> None:
```
hnover-nv, line 298: there's just the one use case, where we pass non-blocking False?  That's fine, just checking you didn't change the behavior here at other call sites.

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
```diff
@@ -375,6 +375,68 @@ def test_active_host_mirror_synced_by_switch_to_generate(self):
...
+        _nest_prefill(
+            si,
+            input_ids=[[1, 2, 3], [4, 5, 6, 7]],
+            pages_per_seq=[1, 1],
+            cache_loc=[10, 20],
+        )
+
+        si.switch_to_generate_(active_args_override={"input_ids"})
+
+        device_cu = si._input_buffer.get_view("cu_seqlen")
```
hnover-nv, line 433: I'm confused by  this test. I see why we would not expect host here to be the device 0,1,2, but can you explain the 0,3,7? Also the 0,1,2, where does that come from?  If this is implicit in the helper functiosn then just tell me that and I'll look more carefully. otherwise coudl be worth a comment.

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
```diff
@@ -375,6 +375,68 @@ def test_active_host_mirror_synced_by_switch_to_generate(self):
...
+            input_ids=[[1, 2, 3]],
+            pages_per_seq=[1],
+            cache_loc=[10],
+        )
+
+        # Mirrors the draft-model placeholder set: active host arg + non-active inter-module inputs.
+        draft_placeholders = {"input_ids", "cu_seqlen_host", "inputs_embeds", "hidden_states"}
+        si.switch_to_generate_(active_args_override=draft_placeholders)
+
+        increment = torch.tensor([1], dtype=torch.int32, device=si.device)
```
hnover-nv: hmm, why are we doing increment and offset_pos_and_cache_ here, vs above we just do switch_to_generate_?  I'd thinjk we are testing the same behavior about what the active_args_override does in all 3 tests, just with nothing in the set, the unique thing we care about, or also with placeholders?

hnover-nv (commented):
Good but I am confused by the tests.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: Yup just the one use case. I thought it made sense to share a signature with torch.copy_() with a non blocking param, and then also make it explicit at the call sites.

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
govind-ramnarayan: 0,3,7 is the cumulative sequence lengths that we input at the beginning - the first sequence has length 3, second has length 4, so cu_seq_len is 0, 0+3, 0+3+4

When we use switch_to_generate(), the sequence lengths all become 1. So after that, cu_seq_len is 0, 1, 2

But since we are using input_ids which is not on host and doesn't trigger the host sync, only the device-side cu_seq_len should get updated, not the host-side.

(The test is vibe coded but I think makes sense with our metadata, it's the right test except maybe the args override could be more realistic).

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
govind-ramnarayan: Pretty much. It could just be switch_to_generate_ here and that would be fine and we'd probably want a separate one for offset_pos_and_cache_(). 

The reason I made this test is I initially had a strict check within these functions that active_args_override consisted of args that were "active" in the first place - so that this would restrict to a subset of the the original active_args. But this caused a crash because the natural way to call this is "get all the arguments provided to the draft model, some of which are *not* args to the wrapper model (like hidden states), and feed them to this function". So this is just a unit test that this calling convention (where we provide random other stuff in `active_args_override`) does not cause a crash and produces expected behavior.

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
hnover-nv: got it, thanks.  the magic is _nest_prefill made cu_seqlen I guess.  Should have looked there :)

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
hnover-nv: Per discussion elsewhere, let's have tests that are just switch_to_generate, and then other tests that test the offset_pos_and_cache.

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
```diff
@@ -76,6 +80,25 @@ def _nest_prefill(si: SequenceInfo, input_ids, pages_per_seq, cache_loc, **kw):
...
+    """Convenience wrapper: nest a two-sequence all-decode batch (seq_len=1 per sequence).
+
+    cu_seqlen is the decode layout [0, 1, 2] (1 token per sequence). With the default
+    input_pos=(5, 10): seq_len_with_cache stages to [6, 11] (input_pos + 1), and
+    cache_loc/cu_num_pages give 2 pages for seq 0 and 3 for seq 1.
+    """
+    si.nest_sequences(
+        [1, 2],
+        cu_seqlen=[0, 1, 2],
+        input_pos=list(input_pos),
```
govind-ramnarayan, line 93: assert input_pos length is 2.

hnover-nv (approved):

PR #14722, opened 2026-05-29
[None][feat] serve: expose is_first_turn from routers and propagate conversation_id to workers

lancelly:
This PR makes the multi-turn conversation id usable end to end in disaggregated serving and exposes a first-turn signal from the routers.

## 1. Expose `is_first_turn` from the disagg routers

`get_next_server(...)` now returns `is_first_turn` in its metadata dict (and logs it at debug level) for:
- **`ConversationRouter`** — derived from session-table membership (no new state).
- **`KvCacheAwareRouter`** — which has no conversation tracking, so a small LRU-bounded `_seen_conversations` set (`max_sessions`, default 100000) + `_record_conversation_turn()` flags the first turn with matching semantics.

The conversation-id lookup is factored into a shared `BlockHashMixin._get_conversation_id` staticmethod (reads the serve-layer `disaggregated_params.conversation_id`, populated upstream from session headers such as `X-Session-ID` by `OpenAIDisaggServer._extract_conversation_id`); `ConversationRouter`'s duplicate copy is dropped.

`is_first_turn` is `False` when no conversation id is available (e.g. the client sent no `X-Session-ID` header) — i.e. "not known to be a first turn". Semantics match across both routers.

## 2. Propagate `conversation_id` to the executor / worker layer

Previously `conversation_id` existed **only** on the serve-layer `DisaggregatedParams` (`serve/openai_protocol.py`). The executor-layer `DisaggregatedParams` (`tensorrt_llm/disaggregated_params.py`, == `LlmDisaggregatedParams`) had no such field, and `to_llm_disaggregated_params()` did not copy it — so when a worker converted the incoming request (`openai_server.py` → `to_llm_disaggregated_params`), the id was silently dropped.

Net effect: worker-side consumers that read `request.py_disaggregated_params.conversation_id` (e.g. the ADP router) **always saw `None`**, even though the disagg orchestrator routed on a real conversation id.

Fix:
- Add `conversation_id` to the executor-layer `DisaggregatedParams`.
- Copy it through both `to_llm_disaggregated_params` and `to_disaggregated_params` so the serve↔executor round-trip preserves it.

Orchestrator-level routers are unaffected (they already read the serve layer directly, before this conversion).

## Test

- `tests/unittest/disaggregated/test_router.py`: LRU seen-set helper (first/subsequent/empty/eviction) and the `is_first_turn` output of `get_next_server` for both routers across turns / with no conversation id. Existing router tests are unaffected (they discard the metadata dict).
- `tests/unittest/disaggregated/test_disaggregated_params.py`: extend the converter tests to assert `conversation_id` propagates, plus a serve↔executor round-trip test.

> Opened as **draft**. Targets `feat/deepseek_v4`.

### DIFF
--- a/tensorrt_llm/disaggregated_params.py
+++ b/tensorrt_llm/disaggregated_params.py
@@ -51,6 +51,10 @@ class DisaggregatedParams:
     ctx_dp_rank: Optional[int] = None
     ctx_info_endpoint: Optional[str] = None
     schedule_style: Optional[DisaggScheduleStyle] = None
+    # Multi-turn conversation id (from session headers such as X-Session-ID),
+    # carried through so worker-side consumers (e.g. the ADP router) can see
+    # the same id the disagg orchestrator routed on.
+    conversation_id: Optional[str] = None
 
     # E-P Disaggregated Params
     multimodal_embedding_handles: Optional[List[Dict[str, Any]]] = (
--- a/tensorrt_llm/serve/openai_protocol.py
+++ b/tensorrt_llm/serve/openai_protocol.py
@@ -1243,6 +1243,7 @@ def to_disaggregated_params(
         ctx_dp_rank=tllm_disagg_params.ctx_dp_rank,
         ctx_info_endpoint=tllm_disagg_params.ctx_info_endpoint,
         schedule_style=tllm_disagg_params.schedule_style,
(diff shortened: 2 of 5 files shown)


longlee0622:

PR #14733, opened 2026-05-29
[TRTLLM-12491][feat] Align VisualGen serve request schema with VisualGenParams

zhenhuaw-me:
## Description

Aligns the `trtllm-serve` visual-generation routes (image / sync video / async video / image-edit) with the public `VisualGenParams` Python API, and tightens the Python API surface around parameter validation, error propagation, output saving, and seed handling.

- `VisualGenParams` (under `tensorrt_llm/visual_gen/params.py`) is now the single source of truth for per-request generation parameters. Universal fields default to `None` ("use the loaded pipeline's default"); model-specific knobs live in `extra_params` (LTX-2 exposes `image_cond_strength` / `output_type` / `guidance_rescale` / `stg_*` / `modality_scale` / `rescale_scale` / `guidance_skip_step` / `enhance_prompt`; Wan 2.2 A14B exposes `guidance_scale_2` / `boundary_ratio`; Wan I2V exposes `last_image`).
- `validate_visual_gen_params` runs synchronously at `VisualGen.generate_async` entry — checking `VisualGenParams` against the loaded pipeline's `default_generation_params` and `extra_param_specs` (cached on the executor from the worker's READY signal). Bad-request failures raise `ValueError` in the caller's process; engine-side failures raise `RuntimeError` when the result is awaited. Matches the `tensorrt_llm.LLM` convention and how vllm-omni / SGLang multimodal_gen behave.
- The OpenAI request schemas (`ImageGenerationRequest`, `VideoGenerationRequest`) map 1:1 to `VisualGenParams`. Unknown top-level fields return HTTP 422 via `extra="forbid"`; per-request denoise controls (`seed` / `num_inference_steps` / `guidance_scale` / `max_sequence_length` / `negative_prompt`); model-specific knobs flow through `extra_params` and are validated against the loaded pipeline's specs. The `/v1/images/edits` route stays registered but returns HTTP 501 (no in-tree pipeline implements editing) and no longer parses a typed request body.
- `format` is a `Literal` covering image/video encoders plus tensor formats `"safetensors"` / `"pt"`. The tensor formats carry every populated modality (image / video / audio) plus scalar metadata in one payload (`tensorrt_llm/media/tensor_payload.py`) so consumers get a single file per logical output. Routes catch `ValueError` from validation and `pipeline.infer()` and render the LLM error envelope at HTTP 400; `RuntimeError` becomes HTTP 500.
- Pipeline `forward` signatures (flux / flux2 / wan / wan_i2v / ltx2 / ltx2_two_stages / cosmos3) take `seed: int` as the 2nd positional parameter (after `prompt`, or after `image` for `WanImageToVideoPipeline`). Engine fallback materializes a fresh seed via `secrets.randbits(63)` on the coordinator rank before `torch.distributed.broadcast_object_list`, so every rank sees the same resolved seed and determinism survives multi-rank runs.
- `VisualGenOutput.save(path)` writes one logical output: unbatched and `batch == 1` tensors save as-is; `batch > 1` paired with a single path raises `ValueError` (matching diffusers' caller-iterates contract and vllm-omni's codec-layer rule). Callers who want N files pass a list of N paths; callers who want one item out of a batch use `output.save_bytes(format=..., batch_index=i)`. The legacy implicit-slice-to-first behavior in `save_image` / `save_video` / tensor save is removed.
- The async video tensor-format path writes batched outputs as `{video_id}_{i}.{fmt}`; `get_video_content`'s recovery probe accepts both the bare and batch-indexed filename shapes.

One open thread on the PR is deferred to a follow-up (TRTLLM-13143): the media-transport contract for `response_format` (raw bytes vs. b64) when paired with tensor formats.

## Test Coverage

- ~290 focused VisualGen unit tests in `tests/unittest/_torch/visual_gen/` cover the public API (params validation, output transport, save dispatch, default merging), the route-level rendering (image / sync video / async video / image-edit / get-content / delete), schema 1:1 mapping with `VisualGenParams`, and tensor payload roundtrip (`safetensors` and `pt`, batched and unbatched, with metadata overrides).
- Verified on GB200 with a clean Release build (sm_100 aarch64).

<\!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **New Features**
  * Added tensor output format support (`safetensors`, `pt`) for image and video generation.
  * Added new `format` parameter for specifying output encoding across generation requests.
  * Added image content retrieval endpoint for serving generated images.

* **Bug Fixes**
  * Improved request validation with structured error messages for better debugging.

* **Documentation**
  * Updated Visual Generation API documentation with detailed request schemas and parameter specifications.

<\!-- review_stack_entry_start -->

[\Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14733?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

<\!-- review_stack_entry_end -->

<\!-- end of auto-generated comment: release notes by coderabbit.ai -->


tensorrt_llm/visual_gen/visual_gen.py
```diff
@@ -664,11 +664,29 @@ def _build_resolved(self, response: "DiffusionResponse"):
...
+        #   pre-existing contract for non-validation engine errors.
+        # For batches, return the list as-is so callers can iterate
+        # per-item ``error`` (and ``error_reason``/``param``/``details``).
         if self._batch_size is None and isinstance(self._resolved, VisualGenOutput):
-            if self._resolved.error is not None:
-                raise RuntimeError(f"Generation failed: {self._resolved.error}")
+            out = self._resolved
+            if out.error is not None:
+                if out.error_reason is not None:
+                    from tensorrt_llm._torch.visual_gen.executor import VisualGenValidationError
```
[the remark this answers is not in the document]
zhenhuaw-me: [AI] Same answer as the visual_gen.py:795 thread — dropped in commit d72eaa1c83. See https://github.com/NVIDIA/TensorRT-LLM/pull/14733#discussion_r3322563431.

tensorrt_llm/serve/openai_protocol.py
```diff
@@ -1447,61 +1486,82 @@ def validate_size(cls, v):
...
-    This is an extension to the OpenAI API for video generation support.
+    Universal per-request fields map 1:1 to :class:`VisualGenParams`.
+    Model-specific knobs travel through ``extra_params``. Unknown
+    top-level fields are rejected with HTTP 422 via the inherited
+    ``extra="forbid"`` policy.
     """
+
+    # Prompt + transport
     prompt: str
+    response_format: Literal["url", "b64_json"] = "url"
```
zhenhuaw-me, line 1461: For this response format, what's the industry and community standard of passing the content of encoded videos, images and raw tensor content? And that's the content type? I am checking if the conversion to base64 encoding is required for image and videos.

zhenhuaw-me: [AI] Survey of how the ecosystem transports generated media:

- **OpenAI Images / DALL-E** (`/v1/images/generations`): `response_format = "url" | "b64_json"`. `url` → temporary CDN URL; `b64_json` → base64-inlined JSON. Default `url`.
- **OpenAI Sora** (`/v1/videos`): async job model — create call returns metadata, raw video bytes fetched via separate `GET /v1/videos/{id}/content` returning `application/octet-stream`. No base64.
- **Replicate / Together / Fireworks**: CDN URLs to time-limited blobs.
- **Stability AI**: base64-inlined in JSON.
- **Self-hosted (AUTOMATIC1111 etc.)**: mostly base64 JSON, some support raw-bytes endpoints.

Two transport modes are standard: raw bytes via `Content-Type: image/png` / `video/mp4` / `application/octet-stream` (efficient, RESTful, no encoding cost) and base64 inline in JSON (Pythonic, ~33% larger). URLs require persistent storage so they only make sense with a CDN/blob store.

For tensor formats specifically, raw bytes is overwhelmingly preferred — consumers `safetensors.load(...)` directly and b64 just adds a decode step. Your direction ("default b64_json for OpenAI compat, raw bytes for tensor formats") matches that: keep the `response_format` mode for image/video, but for `format in {safetensors, pt}` always stream raw bytes via `application/octet-stream` regardless of `response_format`. I'd defer the actual implementation to a follow-up — there are a couple of design knobs (do we reject `response_format="b64_json"` + tensor format with a 400, or just ignore the b64 request and stream bytes anyway?) that are worth thinking through before changing the public contract. Leaving this thread open as the marker.

tensorrt_llm/serve/openai_server.py
```diff
@@ -1931,8 +1971,12 @@ async def openai_image_generation(self, request: ImageGenerationRequest,
                                       raw_request: Request) -> Response:
         """OpenAI-compatible image generation endpoint.
 
-        Follows the OpenAI Images API specification for image generation.
+        Follows the OpenAI Images API specification for image generation,
+        with ``request.format`` extended to accept tensor payloads
+        (``"safetensors"``/``"pt"``) alongside the PNG/WebP/JPEG encoders.
         """
+        from tensorrt_llm.media.tensor_payload import is_tensor_format
```
JunyiXu-nv: Maybe import at the top of the file.

tensorrt_llm/serve/openai_server.py
```diff
@@ -1985,9 +2078,59 @@ async def openai_image_generation(self, request: ImageGenerationRequest,
...
+    ) -> Response:
+        """Serve a generated image by ID and batch index.
+
+        ``GET /v1/images/{image_id}/content?i=<batch_index>`` returns
+        the image file the corresponding ``POST /v1/images/generations``
+        wrote into ``media_storage_path``. ``i`` defaults to ``0`` for
+        single-image requests; URL responses for ``n > 1`` requests
+        carry ``?i=N`` per item.
+        """
+        from fastapi.responses import FileResponse
```
JunyiXu-nv: Same for this.

JunyiXu-nv (approved):
LGTM for serve side. Left some NIT questions.

tensorrt_llm/serve/openai_server.py
zhenhuaw-me: Hoisted `is_tensor_format` next to the existing `tensorrt_llm.media.encoding` import at the top of the file.

tensorrt_llm/serve/openai_server.py
zhenhuaw-me: Hoisted `FileResponse` into the existing `from fastapi.responses import ...` at the top of the file.

tensorrt_llm/serve/openai_protocol.py
```diff
@@ -1346,57 +1346,79 @@ def to_llm_disaggregated_params(
...
-    user: Optional[str] = None
+    format: Literal["png", "webp", "jpeg", "safetensors", "pt"] = Field(
+        default="png",
+        description=(
+            "Generation content encoding format. Image encoders write "
+            "``png``/``webp``/``jpeg``; tensor encoders write "
+            "``safetensors``/``pt`` for programmatic post-processing."),
+    )
+    seed: Optional[int] = Field(default=None,
+                                ge=0,
```
chang-l, line 1371: **Seed upper-bound clamp claimed but not enforced.** The PR description and README state `seed` is "clamped to `[0, 2**32-1]`", but the field only has `ge=0` — there is **no `le`** (same on the ImageEdit seed L1482 and Video seed L1516). So `seed=10**30` passes validation and is forwarded to the pipeline, while the engine fallback draws `secrets.randbits(63)` — a different range. Suggest adding `le=2**32 - 1` (validate → 422, don't silently clamp, which would break reproducibility), and reconciling the fallback bit-width with the documented range.

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
```diff
@@ -381,6 +381,7 @@ def infer(self, req):
     def forward(
         self,
         prompt: Union[str, List[str]],
+        seed: int,
```
chang-l, line 391: **[minor] `seed` is now required, but only the `generate*` paths resolve it — public `enqueue_requests` can pass `None`.** Making `seed` a required positional (`seed: int = 42` → `seed: int`) is fine for `generate()`/`generate_async()`, which resolve a `None` seed to a concrete int on rank-0 (`visual_gen.py:906`) before enqueue. But that resolution lives *only* there; `executor._merge_defaults` doesn't fill `seed`. So the public `enqueue_requests(...)` entry point, called with a default `VisualGenParams()` (`seed=None`), flows to `forward(seed=None)` → `torch.Generator(...).manual_seed(None)` (`pipeline_wan.py:413`) → `TypeError`.

Latent today (the only in-tree caller of `enqueue_requests` is `generate_async`, which pre-resolves), so not a live bug — but it's a foot-gun for external API users / future callers / tests, since `enqueue_requests` is public. Cheap guard: fill `seed` in `_merge_defaults` (or keep a default on `forward`) so the worker doesn't silently depend on every caller pre-resolving it.

tensorrt_llm/serve/openai_server.py
```diff
@@ -1961,29 +2005,78 @@ async def openai_image_generation(self, request: ImageGenerationRequest,
...
-                            image_to_bytes(image)).decode('utf-8'),
-                        revised_prompt=request.prompt,
-                    ) for image in output_images
-                ]
-
+            if is_tensor_format(request.format):
+                # Tensor payloads carry every populated modality in a
+                # single file. Match the image-encoder fan-out by
+                # emitting one ``ImageObject`` per batch item.
+                from tensorrt_llm.media.tensor_payload import infer_batch_size
```
chang-l, line 2028: **Server-side `ValueError`s get rendered as client HTTP 400.** This try-block spans `infer_batch_size(output)` (L2015) and `output.save_bytes(...)` (L2020), and the outer `except ValueError` (L2092) maps to `HTTPStatus.BAD_REQUEST`. But `infer_batch_size` raises `ValueError` for **server-side** conditions — `tensor_payload.py:84` ("carries no media tensor") and `:88` ("Inconsistent batch sizes across modalities") — i.e. a failure to serialize the server's *own* output. The client is then told its request was malformed (400) when it should be 500. Suggest scoping `ValueError → 400` to the parse/validate region only and letting post-generation serialization errors fall through to 500 (or raise a distinct exception type from `tensor_payload`).

tensorrt_llm/media/tensor_payload.py
```diff
@@ -0,0 +1,232 @@
...
+    )
+    if not tensors:
+        raise ValueError(
+            f"Cannot serialize output: request {output.request_id} carries no media tensor."
+        )
+
+    if fmt == "safetensors":
+        from safetensors.torch import save as safetensors_save
+
+        return safetensors_save(tensors, metadata={k: str(v) for k, v in metadata.items()})
```
chang-l: **safetensors metadata round-trip is lossy.** Scalars (`frame_rate`, `audio_sample_rate`) are stringified into the safetensors *header* via `metadata={k: str(v)}`. The canonical `safetensors.torch.load(bytes)` used on the consumer side returns **only the tensors and drops the header**, so a saved video loses its fps (unplayable) unless the consumer specifically uses `safe_open().metadata()` and re-parses the strings back to float/int. Consider storing these scalars as 0-dim tensors so they survive `load()`, or document the header-only contract loudly.

tensorrt_llm/media/tensor_payload.py
```diff
@@ -0,0 +1,232 @@
...
+    metadata.
+    """
+    tensors: Dict[str, torch.Tensor] = {}
+    metadata: Dict[str, Any] = {}
+
+    for name, src in _modalities(output):
+        if src is None:
+            continue
+        if batch_index is not None and src.dim() == _BATCHED_RANKS[name]:
+            sliced = src[batch_index]
```
chang-l, line 131: **`batch_index` slicing has no bounds / consistency validation** (unlike the file `save()` path, which calls `infer_batch_size` first). Here `src[batch_index]` is sliced directly per modality: an out-of-range index raises a raw `IndexError` (not the module's clean `ValueError`), and a mismatched multi-modal output (e.g. image batch 2 + video batch 3) is silently mis-sliced with no error. Since `save_bytes(batch_index=...)` is reachable from request input, suggest validating `0 <= batch_index < batch_size` (and running the cross-modality batch check) before slicing.

tensorrt_llm/serve/openai_protocol.py
```diff
@@ -1450,6 +1479,7 @@ class ImageEditRequest(OpenAIBaseModel):
         default=None,
         description="Text describing what to avoid in the edited image.")
     seed: Optional[int] = Field(default=None,
+                                ge=0,
```
chang-l: **`ImageEditRequest` is only half-migrated to the 1:1 `VisualGenParams` schema.** Unlike `ImageGenerationRequest`/`VideoGenerationRequest`, this model still carries legacy shape: `guidance_rescale` (removed from the other two), the custom `validate_size` field-validator (note: `re.match(pattern, None)` raises `TypeError` on explicit `size: null` → 500, not 422), required `n`, and no `extra_params` / `format` / `width` / `height` / positivity guards. With `extra="forbid"`, an OpenAI-SDK edit client sending the new fields gets 422. Either migrate it to the same field set or explicitly document it as intentionally-legacy (the route returns 501 today, but the schema is committed API surface).

tensorrt_llm/serve/openai_protocol.py
```diff
@@ -1465,61 +1495,81 @@ def validate_size(cls, v):
...
+    format: Literal["mp4", "avi", "auto", "safetensors", "pt"] = Field(
+        default="auto",
+        description=(
+            "Generation content encoding format. Video encoders write "
+            "``mp4``/``avi``/``auto``; tensor encoders write "
+            "``safetensors``/``pt`` and carry video, audio, and scalar "
+            "metadata (frame rate, audio sample rate) in one payload."),
+    )
+    seed: Optional[int] = Field(default=None,
+                                ge=0,
```
chang-l, line 1471: **Undocumented breaking API changes — please call out in release notes.** Relative to the previous schema: (1) top-level `guidance_rescale` removed from Image+Video requests → now **422** (was honored); (2) `n` removed from `VideoGenerationRequest` → clients sending `n` get **422**, multi-video is no longer expressible, and the per-batch save fan-out (`_0`/`_1`/…) becomes dead code (batch is always 1); (3) `quality="hd"` no longer auto-bumps `num_inference_steps` (a test confirms the bump was deliberately dropped). These are reasonable tightenings but are silent breaking changes for existing callers.

tensorrt_llm/serve/visual_gen_utils.py
```diff
@@ -4,59 +4,155 @@
...
-            params.num_inference_steps = request.num_inference_steps
-        if request.n is not None:
-            params.num_images_per_prompt = request.n
+        if request.frame_rate is not None:
+            params.frame_rate = request.frame_rate
+        # num_frames wins; otherwise derive from seconds * frame_rate
+        # (using whichever frame_rate is now in effect on params).
+        if request.num_frames is not None:
+            params.num_frames = request.num_frames
+        elif request.seconds is not None and params.frame_rate is not None:
```
chang-l: **`seconds` silently ignored when `frame_rate` is unresolved.** This branch only derives `num_frames` when *both* `request.seconds` and `params.frame_rate` are set. If the client passes `seconds` but the pipeline default has no `frame_rate` (or it wasn't declared), the duration request is dropped with no error — the user gets the default `num_frames` and no indication their `seconds` was ignored. Suggest: if `request.seconds is not None` but `frame_rate` can't be resolved, raise a `ValueError` (→ 400) rather than silently dropping it.

tensorrt_llm/serve/openai_server.py
chang-l: @JunyiXu-nv does above comment make sense?

tensorrt_llm/media/tensor_payload.py
chang-l: It seems our video output in saved tensor path not working/playable?

tensorrt_llm/serve/openai_protocol.py
chang-l: ^^ @zhenhuaw-me can you double check the above comment?  Since we dont support any image edit model yet, if we dont want to drop `ImageEditRequest` from API surface, maybe we can document it somehow?

tensorrt_llm/media/tensor_payload.py
chang-l: since we dont support batching in visual gen yet, can we add some checks for the batch index?

tensorrt_llm/serve/openai_protocol.py
chang-l: Maybe not on release notes... but please check if it breaks any of our doc?

chang-l (approved):

tensorrt_llm/serve/openai_protocol.py
zhenhuaw-me: I should update the related description. Decided to drop that restriction since it's only related to an OpenAI mode. Let me fix it.

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
zhenhuaw-me: The `enqueue_requests` is an interface of "internal class". Moved the class to `tensorrt_llm/_torch/visual_gen/executor.py` so it now sits next to `DiffusionExecutor`. The seed-on-`forward` requirement is fine because every in-tree path resolves it at the public `VisualGen.generate_async` entry.

tensorrt_llm/serve/openai_server.py
zhenhuaw-me: Refactored. Both `openai_image_generation` and `openai_video_generation_sync` now wrap parse+generate in an inner `try/except ValueError → 400`; the serialization region (`infer_batch_size`, `_save_bytes`, `output.save`) sits outside, so server-side `ValueError`s fall through to the outer `except Exception → 500 InternalServerError`.

tensorrt_llm/media/tensor_payload.py
zhenhuaw-me: Good catch — the test suite already encoded the asymmetry (`safetensors.torch.load` returning only tensors). Went with the hybrid: scalars are now written as both 0-d tensors under the same key AND in the file header. `safetensors.torch.load(bytes)` returns `loaded["frame_rate"]` / `loaded["audio_sample_rate"]` as 0-d tensors (call `.item()` to unbox), and `safe_open(...).metadata()` still exposes the string form for header-only consumers. Updated `tests/unittest/_torch/visual_gen/test_tensor_payload.py` to assert both retrieval paths and added a tensor-format consumer contract section to `examples/visual_gen/serve/README.md`.

tensorrt_llm/media/tensor_payload.py
zhenhuaw-me: Marked `save_bytes` as internal (renamed to `_save_bytes`); the public output API now exposes only `save()`. The single in-tree caller in `openai_server.py` already derives `batch_index` from `infer_batch_size`.

tensorrt_llm/serve/openai_protocol.py
zhenhuaw-me: Dropped `ImageEditRequest` entirely. Didn't notice that it's still there...

tensorrt_llm/serve/openai_protocol.py
zhenhuaw-me: Yeah all docs are aligned

tensorrt_llm/serve/visual_gen_utils.py
zhenhuaw-me: Good catch! `parse_visual_gen_params` now raises `ValueError` when `request.seconds` is set but `params.frame_rate` cannot be resolved.

PR #14746, opened 2026-05-29
[None][fix] Make disagg timeout cancellation rank-consistent

Shixiaowei02:
@coderabbitai summary

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -715,6 +715,70 @@ class PyExecutor:
             # can complete.
             self._enqueue_responses(responses)
 
+    def _sync_kv_transfer_timed_out_flags(self):
+        """Reach cross-rank union consensus on which in-transfer requests hit
+        the KV-transfer timeout, and promote them to the actionable
+        py_kv_transfer_timed_out flag together.
+
+        The timeout is detected per rank via wall-clock (elapsed since
+        py_kv_transfer_start_time).  Acting on that directly diverges across
+        ranks: clock skew flags the same request in different iterations, and
+        the resulting cancel_request + terminate bypasses the per-rid consensus
+        that every other disagg transfer-state transition goes through (see
+        KvCacheTransceiverV2._ctx/_gen_consensus_outcome) -> req.state diverges
+        across ranks -> collective deadlock.
+
+        Here we union the local intents across the instance so every rank flips
+        the actionable flag for the same rids in the same iteration.  The actual
+        cancellation (with its mid-write retry) still happens per rank at the
+        existing sites, now keyed off a rank-consistent flag.
+
+        Must be called from a rank-symmetric (lockstep) point; the callers of
+        _handle_kv_transfer_timeouts_synced satisfy this.
+        """
+        if self.kv_cache_transceiver is None:
+            return
+        timeout_ms = self.kv_cache_transceiver.kv_transfer_timeout_ms
+        if timeout_ms is None:
+            return
+
+        def in_transfer_requests():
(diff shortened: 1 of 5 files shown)


chienchunhung (commented):
Thanks for creating this PR and bringing the issue to awareness. We also discovered the cross-rank divergence risk during PR#13713 development and testing. 

IIUC this PR syncs on the timeout request ids across the ranks in order to mitigate the cross-rank divergence. My understanding is in V2 transceiver there's a particular consensus mechanism that coordinates the request state setting across the ranks, while in V1 transceiver each ranks still sets the request state transition **locally**, even after this PR. This means V1 transceiver still has the risks of cross-rank divergence on the request state. Does this PR plan to address the V1 state transition (e.g., similar to the consensus mechanism in V2 transceiver)?

IIRC currently we have a similar sync on whether any rank timed out or not, e.g., a boolean rather than the request ids, only for _multi-rank ADP_. Once PR#14746 lands which applies across the board, the multi-rank ADP sync becomes redundant and should be removed to reduce the overhead.

Lastly, I am wondering if you have some measurement for how much overhead adding the extra collective brings. PS: Earlier when I was trying to do something similar to V2's consensus in PR#13713 the overhead could be significant in some CP tests.

Shixiaowei02:
> Thanks for creating this PR and bringing the issue to awareness. We also discovered the cross-rank divergence risk during PR#13713 development and testing.
> 
> IIUC this PR syncs on the timeout request ids across the ranks in order to mitigate the cross-rank divergence. My understanding is in V2 transceiver there's a particular consensus mechanism that coordinates the request state setting across the ranks, while in V1 transceiver each ranks still sets the request state transition **locally**, even after this PR. This means V1 transceiver still has the risks of cross-rank divergence on the request state. Does this PR plan to address the V1 state transition (e.g., similar to the consensus mechanism in V2 transceiver)?
> 
> IIRC currently we have a similar sync on whether any rank timed out or not, e.g., a boolean rather than the request ids, only for _multi-rank ADP_. Once PR#14746 lands which applies across the board, the multi-rank ADP sync becomes redundant and should be removed to reduce the overhead.
> 
> Lastly, I am wondering if you have some measurement for how much overhead adding the extra collective brings. PS: Earlier when I was trying to do something similar to V2's consensus in PR#13713 the overhead could be significant in some CP tests.

Thanks for the review! The sync this PR adds lives in `PyExecutor`, so it covers both V1 and V2. On the ADP boolean: agreed the info is now duplicated, but it isn't a clean deletion since it keeps `_handle_errors` in lockstep and gates on the previous iteration's cancelled buffer. This sync issue currently affects correctness — I think the `all_gather` overhead isn't significant, but that needs to be measured. If later profiling shows it's a hotspot, I'll follow up with further optimization.

chienchunhung (approved):
LGTM for the current scope. Thanks!

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -715,13 +715,83 @@ def _flush_pending_transfer_responses(self):
...
+        local_ids = [
+            req.py_request_id for req in in_transfer_requests()
+            if locally_timed_out(req)
+        ]
+        # Union across the instance: a timeout on ANY rank cancels the request
+        # on ALL ranks that hold it, atomically within this iteration.
+        timed_out_ids = set()
+        for ids in self.dist.allgather(local_ids):
+            if ids:
+                timed_out_ids.update(ids)
```
pcastonguay, line 775: I don't think we can have an `allgather` because it will cause PP ranks to synchronize. See `RequestBroadcaster` for example on how PP is handled.

longlee0622:

Issue #14676, opened 2026-05-28
[Bug]: AutoDeploy model-coverage hits NVRTC compilation failed across multiple models


tcherckez-nvidia:
### System Info

- CPU architecture: x86_64
- GPU properties: not captured from CI artifact
- Libraries
  - TensorRT-LLM branch or tag: main (pipeline context)
  - Container used: AutoDeploy model-coverage CI container
- OS: CI environment
- Additional context: terminal pipeline 52785950

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Pipeline `52785720` resolved to terminal model-coverage pipeline `52785950`.

Affected registry entries currently disabled in `examples/auto_deploy/model_registry/models.yaml` include:
- `zai-org/GLM-5-FP8`
- `MiniMaxAI/MiniMax-M2.5`
- `MiniMaxAI/MiniMax-M2.7`
- `deepseek-ai/DeepSeek-V3.2`
- `deepseek-ai/DeepSeek-V3.2-Speciale`
- `MiniMaxAI/MiniMax-M2`
- `deepseek-ai/DeepSeek-Prover-V2-671B`
- `deepseek-ai/DeepSeek-V3`
- `Qwen/Qwen3-0.6B-FP8`

Representative evidence from per-test stdout logs ends with:
```text
RuntimeError: NVRTC compilation failed
```

One concrete example is `MiniMaxAI/MiniMax-M2.7`, where the traceback tail runs through `tensorrt_llm/_torch/auto_deploy/custom_ops/quantization/trtllm_quant.py` into `torch.ops.trtllm.fp8_block_scaling_gemm(...)` before failing with `RuntimeError: NVRTC compilation failed`.

### Expected behavior

These model-coverage entries should either compile and run successfully, or fail with a more specific unsupported-configuration signal if FP8 block-scaling GEMM is not supported in the current path.

### actual behavior

TensorRT-LLM fails during AutoDeploy execution with `RuntimeError: NVRTC compilation failed` across multiple distinct models, suggesting a shared runtime or codegen failure rather than a model-specific problem.

### additional notes

This looks like a TensorRT-LLM / AutoDeploy runtime issue, not a Hugging Face access problem. The same failure string appears across nine disabled entries, which makes it a good candidate for one general issue

[...]

PR #14762, opened 2026-05-29
auto_deploy: Guard FP8 block-scaling GEMM with explicit compilation fallback

Priyanshu31102003:
Addresses #14676 by wrapping the Hopper/Blackwell explicit FP8 block-scaling GEMM operation inside an NVRTC compilation guard. Rather than dropping an unhandled RuntimeError on unsupported configurations, this ensures the runtime explicitly catches the compilation artifact and allows for proper tracking of unsupported environments.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling and diagnostic messages for FP8 quantization operations to better communicate unsupported hardware configurations and compilation issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14762?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tensorrt_llm/_torch/auto_deploy/custom_ops/quantization/trtllm_quant.py
+++ b/tensorrt_llm/_torch/auto_deploy/custom_ops/quantization/trtllm_quant.py
@@ -167,14 +167,22 @@ def trtllm_finegrained_fp8_linear(
         act_fp8, act_sf = per_token_quant_and_transform(input_2d)
     else:
         # Hopper (SM90) and Blackwell (SM100+) share the same path
-        act_fp8, act_sf = torch.ops.trtllm.fp8_quantize_1x128(input_2d)
-    output = torch.ops.trtllm.fp8_block_scaling_gemm(act_fp8, weight, act_sf, weight_scale)
+        try:
+            act_fp8, act_sf = torch.ops.trtllm.fp8_quantize_1x128(input_2d)
+            output = torch.ops.trtllm.fp8_block_scaling_gemm(act_fp8, weight, act_sf, weight_scale)
+        except RuntimeError as e:
+            if "NVRTC compilation failed" in str(e) or "not supported" in str(e):
+                raise RuntimeError(
+                    f"Unsupported configuration: FP8 block-scaling GEMM compilation failed in the current environment. "
+                    f"Details: {str(e)}"
+                ) from e
+            raise e
 
-    if bias is not None:
-        output = output + bias
+        if bias is not None:
+            output = output + bias
 
-    # Reshape back to original batch dimensions: [M, N] -> [..., N]
-    return output.reshape(*input_shape[:-1], weight.shape[0])
+        # Reshape back to original batch dimensions: [M, N] -> [..., N]
+        return output.reshape(*input_shape[:-1], weight.shape[0])
 
 
 @trtllm_finegrained_fp8_linear.register_fake


tensorrt_llm/_torch/auto_deploy/custom_ops/quantization/trtllm_quant.py
```diff
@@ -167,14 +167,22 @@ def trtllm_finegrained_fp8_linear(
...
-    if bias is not None:
-        output = output + bias
-
-    # Reshape back to original batch dimensions: [M, N] -> [..., N]
-    return output.reshape(*input_shape[:-1], weight.shape[0])
+        try:
+            act_fp8, act_sf = torch.ops.trtllm.fp8_quantize_1x128(input_2d)
+            output = torch.ops.trtllm.fp8_block_scaling_gemm(act_fp8, weight, act_sf, weight_scale)
+        except RuntimeError as e:
+            if "NVRTC compilation failed" in str(e) or "not supported" in str(e):
```
bmarimuthu-nv, line 174: @Priyanshu31102003 could you provide details to reproduce this error caught in the try block?

Priyanshu31102003: Hi @bmarimuthu-nv,

The target error condition was primarily structured based on the NVRTC compilation failure patterns highlighted in issue #14676 across the dynamic AutoDeploy model-coverage entry runs (such as the stdout failure ends for Qwen3-8B-FP8 / DeepSeek variants during block-scaling execution). 

Since this runtime exception stems from specific non-supported fallback environments or NVRTC backend environment constraints encountered within the AutoDeploy pipeline rather than a local unit-test asset, this try-block acts as a defensive guard to cleanly raise an explicit unsupported configuration mapping.

bmarimuthu-nv: Thanks for the details. The error needs a proper debug instead of a defensive try-catch guard. If you encounter this issue at your end, please share the steps to reproduce. Thanks! Closing this PR now because this is not the correct solution and needs proper debug to find the root cause.

bmarimuthu-nv:
Thanks for the details @Priyanshu31102003  Closing this PR now because wrapping it inside try-catch is not the correct solution and needs proper debug to find the root cause.

tensorrt_llm/_torch/auto_deploy/custom_ops/quantization/trtllm_quant.py
Priyanshu31102003: Understood, @bmarimuthu-nv. Completely makes sense to prioritize finding the actual root cause instead of using a defensive fallback guard. 

I would love to take up the challenge of debugging this properly to locate the exact compiler issue behind the NVRTC failures. To make sure I am aligned with your expectations and don't hit a roadblock, could you please guide me on:
1. What exact architectural solution or component fix would you prefer to see here once the root cause is found?
2. Are there any specific internal environment specs, Docker configurations, or verbose logs from the AutoDeploy pipeline that I can reference to replicate this failure locally?

I am highly motivated to fix this the right way. Any pointers or initial guidance to help me kickstart the deep debug would be greatly appreciated!

bmarimuthu-nv: I appreciate the interest in helping fix this issue!

> What exact architectural solution or component fix would you prefer to see here once the root cause is found?

This is something we can think only upon root causing the issue.



> Are there any specific internal environment specs, Docker configurations, or verbose logs from the AutoDeploy pipeline that I can reference to replicate this failure locally?

I would suggest to try referring to the issue https://github.com/NVIDIA/TensorRT-LLM/issues/14676 you had pointed out and follow the instructions/ask instructions to reproduce in the ticket. It'll likely need to be reproduced and tested on a H100 machine.

Priyanshu31102003: Thank you so much for the guidance and pointers, @bmarimuthu-nv! I really appreciate your time. I'm moving over to issue #14676 now to head down the root-cause analysis path.

PR #14773, opened 2026-05-30
[None][feat] Wan 2.2: fuse NVFP4 quantization with preceding LayerNorm/AdaLN and GELU-tanh

anikaj-eng:
## Summary by CodeRabbit

* **New Features**
  * Added fused GELU (tanh) + NVFP4 quantization operations for improved inference performance
  * Added fused LayerNorm + NVFP4 quantization with optional adaptive normalization support
  * Extended MLP and attention modules to leverage optimized quantization kernels

* **Improvements**
  * Enhanced support for Blackwell (SM100) GPU architecture with optimized kernel paths

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14773?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Description

Fuses NVFP4 quantization with the layers preceding the NVFP4 GEMMs in
the Wan 2.2 T2V DiT
([`tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py`](../blob/main/tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py)),
mirroring the approach in #11473 (Mamba2 / Nemotron-H NVFP4 layernorm +
activation fusion).

**Problem.** Each LayerNorm / GELU-tanh feeding into an NVFP4 GEMM
dispatches three CUDA kernels (norm-or-activation, cast,
`fp4_quantize`) before the GEMM can launch. Across the 34 unignored
Wan 2.2 blocks this is a measurable kernel-launch + memory-bandwidth
overhead on B200.

**Solution.** Two new CUDA kernels that emit NVFP4-packed activations
and swizzled per-16-element FP8 (e4m3) scale factors directly into the
layout the CUTLASS NVFP4 GEMM kernels in `Linear` expect:

- `fusedLayerNormQuant` — LayerNorm with optional affine weight/bias
  and optional AdaLN modulation `LN(x) * (1 + scale_msa) + shift_msa`,
  fused with NVFP4 quant. Feeds `attn1.qkv_proj` / `attn2.qkv_proj` /
  `mlp.up_proj`.
- `fusedGeluTanhQuantize` — GELU-tanh + NVFP4 quant. Feeds MLP
  `down_proj`.

Both kernels consume a per-tensor calibrated `input_scale`. The fused
paths only activate when the loaded checkpoint provides such scales
(static NVFP4) and the layer is on the unignored list from the
checkpoint's `quantization_config`; otherwise they fall back to the
existing unfused path (norm/activation → cast → standalone
`fp4_quantize` → GEMM). No correctness or behavior change for
non-calibrated checkpoints.

**Kill switches.**
- `TRTLLM_DISABLE_NVFP4_LAYERNORM_FUSION=1` forces the unfused path for
  all LayerNorms (debug / A-B).
- MLP fused path is gated by `_use_fused_gelu_tanh_quant`, which can be
  cleared per-module for the same purpose.

**Performance.** End-to-end transformer forward pass on B200, A/B with
fusion forced ON vs forced OFF (n=20, after warmup, fp16 condition,
480p latent, 1560 tokens, 77-token text condition), validated against
the ModelOpt checkpoint
[`nvidia/Wan2.2-T2V-A14B-Diffusers-NVFP4`](https://huggingface.co/nvidia/Wan2.2-T2V-A14B-Diffusers-NVFP4):

| Path        | Latency (ms)         |
| ----------- | -------------------- |
| Unfused     | 96.17                |
| Fused       | 81.80                |
| **Speedup** | **1.176x (-14.94%)** |

nsys NVTX breakdown attributes the wins to fewer kernel launches per
norm/activation site (3 → 1).

## Nsys NVTX summary (480p latent / B=1 / 5 warmup + 20 timed)
Captured one nsys run that covers both A/B configurations in a single process
(loads the 14B Wan 2.2 NVFP4 checkpoint once, then runs baseline-first /
fused-second with the same input). Raw NVTX summary from `nsys stats --report
nvtx_sum`:
```
 ** NVTX Range Summary (nvtx_sum):
 Time (%)  Total Time (ns)  Instances    Avg (ns)      Med (ns)     Min (ns)    Max (ns)   StdDev (ns)   Style                       Range
 --------  ---------------  ---------  ------------  ------------  ----------  ----------  -----------  -------  -----------------------------------------------------
     25.6       2634144160          1  2634144160.0  2634144160.0  2634144160  2634144160          0.0  PushPop  TensorRT-LLM:Wan22 NVFP4 baseline (fusion=off)
     25.0       2576735104          1  2576735104.0  2576735104.0  2576735104  2576735104          0.0  PushPop  TensorRT-LLM:Wan22 NVFP4 baseline (fusion=off) warmup
     22.6       2327886688          1  2327886688.0  2327886688.0  2327886688  2327886688          0.0  PushPop  TensorRT-LLM:Wan22 NVFP4 fused (fusion=on)
     17.6       1811006208       3500      517430.3      194912.0      176896   912812416   15625806.1  PushPop  TensorRT-LLM:LN unfused
      5.6        579063232          1   579063232.0   579063232.0   579063232   579063232          0.0  PushPop  TensorRT-LLM:Wan22 NVFP4 fused (fusion=on) warmup
      3.3        341910176       2550      134082.4      133008.0      104352     1045760      21148.4  PushPop  TensorRT-LLM:LN+NVFP4 fused
      0.4         38670816        850       45495.1       44512.0       37408      488992      15586.4  PushPop  TensorRT-LLM:gelu_tanh+NVFP4 fused
```
### Per-iteration latency (the headline number)
The two `(fusion=off/on)` ranges wrap a 20-iter timed loop each (post-warmup).
Dividing total time by iteration count:
| Configuration            | Latency / iter |
| ------------------------ | --------------:|
| Unfused (fusion=off)     | 131.7 ms       |
| Fused   (fusion=on)      | 116.4 ms       |
| **Speedup**              | **1.132x**     |
| **Latency drop**         | **−11.6%**     |
(Wall-time A/B without nsys instrumentation gives ~1.18x / −14.9%; nsys adds a
constant per-launch overhead that compresses the relative win. Both are in the
same band.)
### Per-kernel attribution (median per call)
The inner NVTX ranges count individual kernel-site firings, so the median per
call shows how much faster each fusion site is:
| Kernel                      | Median per call | Speedup at this site |

**Related issue:** 
**Related PR:** #11473 (reference design this adapts).

## Test Coverage

Three new test files, all green on B200 with the calibrated checkpoint
mounted via `WAN22_T2V_NVFP4_PATH`:

**Kernel unit tests (byte-equal vs `fp4_quantize` reference):**
- `tests/unittest/_torch/modules/test_fused_activation_quant.py` —
  GELU-tanh + NVFP4 plus MLP gate logic. 8 tests.
- `tests/unittest/_torch/modules/test_fused_layernorm_quant.py` —
  plain LN, affine LN, and AdaLN forms. Parametric.

**End-to-end integration tests on the calibrated checkpoint:**
- `tests/unittest/_torch/visual_gen/test_wan22_nvfp4_fusion_integration.py`
  — 7 tests, ~6 s total, module-scoped fixture loads the checkpoint
  once:
  1. Quant config resolves to static NVFP4
     (`force_dynamic_quantization=False`, `group_size=16`,
     `exclude_modules` honored).
  2. Every unignored NVFP4 `Linear` has a finite non-zero `input_scale`
     after `load_weights` (guards against ModelOpt safetensors key
     drift).
  3. `post_load_weights` attaches `nvfp4_scale` to the 102 expected
     `norm{1,2,3}` modules on the 34 unignored blocks; the 18 norms in
     the 6 ignored blocks do NOT carry it.
  4. Exactly 34 MLP blocks report `_use_fused_gelu_tanh_quant == True`.
  5. ≥ 102 LayerNorms advertise `is_nvfp4 == True` and non-None
     `nvfp4_scale`.
  6. **Numerical A/B.** Same model, same input, two forward passes
     (fused paths forced OFF vs ON), cosine similarity **0.99756**
     (≥ 0.995 threshold).
  7. **Smoke test.** Full forward pass on a 480p/1560-token/77-condition
     input produces finite, correctly-shaped output.

All seven gated by `@skip_if_no_checkpoint` and `@skip_if_no_sm100`,
so the file no-ops cleanly in CI environments that don't have the
NVFP4 ckpt or are on pre-Blackwell hardware, and activates
automatically when both are available.

## PR Checklist

- [ x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help


tensorrt_llm/_torch/modules/layer_norm.py
```diff
@@ -66,57 +82,228 @@ def __init__(
...
+            # (vs the auto-generated `LayerNorm.forward` range from
+            # `--enable_layerwise_nvtx_marker`). Counting these in
+            # `nsys stats --report nvtxsum` gives a per-block confirmation
+            # that the fast path actually fired.
+            with nvtx_range("LN+NVFP4 fused", color="green"):
+                return self._forward_nvfp4_fused(hidden_states, scale_msa,
+                                                 shift_msa, seq_len_per_batch,
+                                                 nvfp4_scale)
+
+        with nvtx_range("LN unfused", color="grey"):
```
luyiyun1021: **[Concern · perf / shared hot-path]** `nvtx_range()` resolves to `nvtx.annotate()`, which emits an NVTX push/pop on **every** call even when no profiler is attached (`tensorrt_llm/_utils.py:953-972`). This `forward` is the shared `LayerNorm` used by StarCoder2, Cohere2, Qwen3VL, Kimi-K2.5, DSA, Flux and Flux2 — none of which set `quantize_type`, so they all take this default branch and pay the cost (2× per layer × every decode step). Sibling modules (`rms_norm.py`, `linear.py`) carry no always-on `nvtx_range`, and there's a Dynamo graph-break risk if a caller ever `torch.compile`s through this `forward`.

Suggest gating it: use the env-gated `nvtx_range_debug(...)` (`TLLM_NVTX_DEBUG` / `TLLM_LLMAPI_ENABLE_NVTX`), or only wrap when `self.is_nvfp4`, so the default path stays byte-identical to baseline. Same for the `"LN+NVFP4 fused"` label.

tensorrt_llm/_torch/modules/layer_norm.py
```diff
@@ -66,57 +82,228 @@ def __init__(
...
+            # scale_msa / shift_msa typically arrive as [B, 1, N]; reshape to [B, N].
+            # Cast to input dtype (callers often keep modulation in FP32 for the
+            # unfused path's precision; the fused kernel reads bf16/fp16).
+            s_2d = scale_msa.reshape(-1, n).to(hs_2d.dtype).contiguous()
+            sh_2d = shift_msa.reshape(-1, n).to(hs_2d.dtype).contiguous()
+            ln_w = None
+            ln_b = None
+            # Validate the row-to-batch ratio so the kernel's
+            # batch_idx = row / seq_len_per_batch indexing is correct.
+            if hs_2d.shape[0] != s_2d.shape[0] * seq_len_per_batch:
```
luyiyun1021, line 303: **[Concern · robustness / coverage]** This guard hard-raises for the per-token-timestep case. When `temb.ndim==4` (per-patch timestep, e.g. Wan 2.2 TI2V-5B), `scale_msa` is `[B, S, N]`, so `s_2d = scale_msa.reshape(-1, n)` → `[B*S, N]` while `seq_len_per_batch = S`; then `hs_2d.shape[0] (= B*S)` vs `s_2d.shape[0] * seq_len_per_batch (= B*S*S)` raises `ValueError` for any `S > 1`. `is_nvfp4` is set purely from `quant_config` + SM (not from N or timestep rank), so a static-NVFP4 model driving a 2-D timestep crashes instead of falling back. It's currently unreachable (the 2-D-timestep 5B model is hidden_size ≠ 5120 and ships no NVFP4 ckpt) and the kill switch defaults OFF, but it's fragile and untested — the tests only use a 1-D `timestep=[500.]`.

Suggest falling back to `_forward_unfused` when `scale_msa.ndim == 3 and scale_msa.shape[1] > 1`, or asserting + documenting the "1-D timestep only" V1 limitation, and adding a `temb.ndim==4` test.

tensorrt_llm/_torch/modules/layer_norm.py
```diff
@@ -66,57 +82,228 @@ def __init__(
...
+        n = int(orig_shape[-1])
+        hs_2d = hidden_states.reshape(-1, n).contiguous()
+        sf_scale = nvfp4_scale.contiguous()
+
+        if scale_msa is not None and shift_msa is not None:
+            # AdaLN case (norm1, norm3 in Wan 2.2): no LN affine, just modulation.
+            # scale_msa / shift_msa typically arrive as [B, 1, N]; reshape to [B, N].
+            # Cast to input dtype (callers often keep modulation in FP32 for the
+            # unfused path's precision; the fused kernel reads bf16/fp16).
+            s_2d = scale_msa.reshape(-1, n).to(hs_2d.dtype).contiguous()
```
luyiyun1021, line 297: **[Nit · numerics]** The fused path casts `scale_msa`/`shift_msa` to the input bf16/fp16 dtype here, but the unfused fallback (`_forward_unfused`, ~line 216) applies modulation in **fp32**. So a model that falls back (non-SM100, or no `nvfp4_scale`) gets fp32-precision modulation while the fused path gets bf16-precision modulation feeding the FP4 quantizer — a small, undocumented fused-vs-unfused delta. The unit-test reference deliberately downcasts modulation to match the kernel, so the `cos ≥ 0.995` check doesn't guard it. Either document that the fused path intentionally uses native-dtype modulation, or keep `scale_msa`/`shift_msa` in fp32 in the kernel signature to match the unfused contract.

cpp/tensorrt_llm/thop/fusedLayerNormQuant.cpp
```diff
@@ -0,0 +1,201 @@
...
+    int64_t const sfSize = tensorrt_llm::computeSwizzledLayoutSFSize(m, n / sf_vec_size);
+    at::Tensor output_sf = at::detail::empty_cuda({sfSize}, SF_DTYPE, input.device(), std::nullopt);
+
+    auto stream = at::cuda::getCurrentCUDAStream(input.get_device());
+    // Re-query SM count every call. Function-static caching is unsafe in
+    // multi-GPU processes: the first call latches whatever device the
+    // process happens to be on, then subsequent calls from other devices
+    // see a stale value (different B200 variants in the same node, MIG
+    // partitions, etc.).
+    int const multiProcessorCount = tensorrt_llm::common::getMultiProcessorCount();
```
luyiyun1021: **[Nit · dead code]** `multiProcessorCount` is computed here and threaded into `invokeFusedLayerNormQuant`, but the launcher's parameter is `int /*multiProcessorCount*/` (unused) — the grid is a fixed `dim3 grid(params.M)` with no occupancy tuning. So this value is computed and discarded, and the "re-query SM count every call / static caching is unsafe" comment defends a dead value. Suggest dropping the parameter + this call (and the `.cuh` extern-template arg), or leaving a one-line TODO if it's kept for future occupancy tuning.

luyiyun1021 (commented):
Comments from a structured review pass (CUDA kernels + shared-module impact). All non-blocking — the implementation is careful and the fused path defaults OFF — but a couple are worth addressing before merge.

**Inline:** 2 Concerns — unconditional NVTX on the shared `LayerNorm` default path (affects all non-Wan models), and a `temb.ndim==4` (per-token-timestep) hard-crash that's untested — plus several nits (dead `multiProcessorCount`, gelu-kernel DRY vs relu2, a stale FP8 rationale in the mlp gate comment, a wrong `modeling_llama.py` line citation, a personal scratch path in a test, and a fused-vs-unfused modulation-precision delta).

**PR-level (no single line):** the headline **1.176× speedup has no committed benchmark or repro script** — only the description table and the nsys A/B recipe in the `_ln_quant_type` docstring. For a perf-motivated PR, consider adding a small reproducible benchmark (or stating it's a one-off local nsys run with the exact config) so reviewers can verify the number.

cpp/tensorrt_llm/kernels/fusedActivationQuant.cu
```diff
@@ -182,6 +193,150 @@ template void invokeFusedRelu2Quantize<__nv_bfloat16>(
...
 #endif
 
+// Fused gelu(tanh approx) + NVFP4 quantization kernel.
+//
+// Mirrors fusedRelu2QuantizeKernel exactly; only the elementwise activation
+// swaps relu2_f32 for gelu_tanh_f32. Activation is computed in f32 then rounded
+// back to native precision (bf16/fp16) so absmax / SF / FP4 packing math is
+// byte-for-byte identical to the relu2 kernel and to cvt_warp_fp16_to_fp4.
+template <typename T>
+__global__ void fusedGeluTanhQuantizeKernel(T const* __restrict__ input, float const* __restrict__ sfScale,
```
luyiyun1021: **[Nit · DRY]** `fusedGeluTanhQuantizeKernel` (+ its launcher) is a ~140-line near-verbatim copy of `fusedRelu2QuantizeKernel`, differing only in the activation call (`relu2_f32` → `gelu_tanh_f32`); the whole NVFP4 epilogue (absmax, XOR-1 shuffle, SF swizzle, packing, padded-column handling) is duplicated, so any future SF-layout/absmax fix must be applied twice. Suggest templating on an activation functor, e.g. `template <typename T, typename Act> __global__ void fusedActQuantizeKernel(...)` calling `Act{}(x)`, and instantiating both relu2 and gelu_tanh from it.

tensorrt_llm/_torch/modules/mlp.py
```diff
@@ -99,11 +101,32 @@ def create_weights(self):
         has_nvfp4 = hasattr(self.down_proj,
                             'has_nvfp4') and self.down_proj.has_nvfp4
         has_kernel = hasattr(torch.ops.trtllm, 'fused_relu2_quantize')
-        has_scale = hasattr(self.down_proj, 'input_scale')
+        # NVFP4LinearMethod.create_weights always allocates input_scale as a
+        # Parameter (linear.py:1295), but a layer that opts into dynamic quant
+        # at load time will reset it to None (linear.py:821). The fused kernel
+        # needs a real scalar tensor, so guard against the None case explicitly.
+        has_scale = getattr(self.down_proj, 'input_scale', None) is not None
```
luyiyun1021, line 128: **[Nit · stale comment]** The comment above (line 106) justifies this `input_scale is not None` guard via "a layer that opts into dynamic quant at load time will reset it to None (linear.py:821)" — but `linear.py:821` is in the **FP8** `LinearMethod` path, which NVFP4 never executes. For NVFP4, `input_scale` is always an allocated Parameter (`linear.py:1295`), so for the layers this gate actually controls (NVFP4 relu2/gelu) the `hasattr → getattr is not None` change is behavior-preserving — good, no relu2/Nemotron-H regression. Only the cited rationale is wrong for NVFP4. Suggest dropping the `linear.py:821` citation, or rephrasing to "for NVFP4 `input_scale` is always allocated; this is purely defensive against a hypothetical future None-reset.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -12,12 +13,74 @@
...
+
+def _ln_quant_type(quant_config, force_dynamic_quant: bool):
+    """Return ``"nvfp4"`` if the model is configured for static (calibrated)
+    NVFP4 quantization, otherwise ``None``. The fused LayerNorm + NVFP4
+    quant fast path requires per-layer calibrated ``input_scale`` from the
+    downstream Linear, so dynamic-quant runs fall back to the unfused path
+    silently (no ``nvfp4_scale`` attached -> LayerNorm uses _forward_unfused).
+
+    A/B kill switch: ``TRTLLM_DISABLE_NVFP4_LAYERNORM_FUSION`` mirrors the
+    flag used by Llama (modeling_llama.py:673). Default is ``"1"`` (fusion
```
luyiyun1021: **[Nit]** The env-var read in `modeling_llama.py` is at ~line 704, not 673 (673 is unrelated config setup). The flag name and default (`"1"`) do match. Hard-coded line citations rot quickly — suggest citing just `modeling_llama.py`.

tests/unittest/_torch/visual_gen/test_wan22_nvfp4_fusion_integration.py
```diff
@@ -0,0 +1,530 @@
...
+
+
+def _checkpoint_path() -> str:
+    """Resolve the calibrated NVFP4 checkpoint path. Prefer the env var; fall
+    back to the conventional ``/models`` bind-mount used by the build
+    container, and finally the host scratch path."""
+    candidates = [
+        os.environ.get("WAN22_T2V_NVFP4_PATH"),
+        "/models/Wan2.2-T2V-A14B-Diffusers-NVFP4",
+        "/home/scratch.anikaj_libs/trunk_08042025/models/Wan2.2-T2V-A14B-Diffusers-NVFP4",
```
luyiyun1021: **[Nit]** This hard-codes a personal developer scratch path (`/home/scratch.anikaj_libs/...`) as a fallback candidate — it leaks a username into the public repo and is dead for every other contributor / CI runner. The env var `WAN22_T2V_NVFP4_PATH` and the `/models/...` bind-mount already cover legitimate resolution (and the skip reason only references those two). Suggest removing this candidate. The repo convention for shared test data is the `/home/scratch.trt_llm_data_ci/llm-models/` mount.

cpp/tensorrt_llm/kernels/fusedLayerNormQuant/fusedLayerNormQuant.cu
```diff
@@ -0,0 +1,408 @@
...
+        params.N == kN_HARDCODED, "fusedLayerNormQuant v1 supports N=%d only (got %d).", kN_HARDCODED, params.N);
+    TLLM_CHECK_WITH_INFO(
+        !(params.has_ln_affine && params.has_modulation), "has_ln_affine and has_modulation are mutually exclusive.");
+    if (params.has_modulation)
+    {
+        TLLM_CHECK_WITH_INFO(
+            params.seq_len_per_batch > 0, "seq_len_per_batch must be positive when has_modulation is true.");
+        TLLM_CHECK_WITH_INFO(params.M % params.seq_len_per_batch == 0,
+            "M (%d) must be divisible by seq_len_per_batch (%d).", params.M, params.seq_len_per_batch);
+    }
```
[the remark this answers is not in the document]
anikaj-eng: check my last commit and resolve if addressed.

cpp/tensorrt_llm/thop/fusedActivationQuant.cpp
```diff
@@ -79,16 +79,66 @@ std::tuple<at::Tensor, at::Tensor> fused_relu2_quantize(
     return std::make_tuple(output_fp4, output_sf);
 }
 
+std::tuple<at::Tensor, at::Tensor> fused_gelu_tanh_quantize(
+    at::Tensor const& input, at::Tensor const& sf_scale, int64_t sf_vec_size)
+{
+    CHECK_TH_CUDA(input);
+    CHECK_CONTIGUOUS(input);
+    CHECK_INPUT(sf_scale, torch::kFloat32);
```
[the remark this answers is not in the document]
anikaj-eng: addressed in last commit.

cpp/tensorrt_llm/thop/fusedLayerNormQuant.cpp
```diff
@@ -0,0 +1,188 @@
...
+ *   ln_bias:    [N]      bf16 or fp16. Optional.
+ *   scale_msa:  [B, N]   bf16 or fp16. Optional. Broadcasts: row r reads batch r / seq_len_per_batch.
+ *   shift_msa:  [B, N]   bf16 or fp16. Optional.
+ *   sf_scale:   scalar   fp32 (calibrated module.input_scale of the downstream Linear).
+ * Returns:
+ *   output_fp4: [M, N/2] uint8 (each byte = 2 packed FP4 nibbles).
+ *   output_sf:  [sf_size] uint8, swizzled layout (each byte is one FP8 e4m3 scale).
+ *
+ * v1 supports only N = 5120 (Wan 2.2 hidden_size). Other N values will trigger a TORCH_CHECK.
+ */
```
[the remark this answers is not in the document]
anikaj-eng: addressed in last commit.

cpp/tensorrt_llm/thop/fusedLayerNormQuant.cpp
```diff
@@ -0,0 +1,188 @@
...
+        TORCH_CHECK(s.sizes()[0] == m / seq_len_per_batch, "scale_msa batch dim must equal M / seq_len_per_batch.");
+    }
+
+    // Allocate output tensors.
+    at::Tensor output_fp4 = at::detail::empty_cuda({m, n / 2}, torch::kUInt8, input.device(), std::nullopt);
+    int64_t const sfSize = tensorrt_llm::computeSwizzledLayoutSFSize(m, n / sf_vec_size);
+    at::Tensor output_sf = at::detail::empty_cuda({sfSize}, SF_DTYPE, input.device(), std::nullopt);
+
+    auto stream = at::cuda::getCurrentCUDAStream(input.get_device());
+    static int const multiProcessorCount = tensorrt_llm::common::getMultiProcessorCount();
```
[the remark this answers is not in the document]
anikaj-eng: addressed in last commit.

tensorrt_llm/_torch/modules/layer_norm.py
```diff
@@ -66,57 +82,202 @@ def __init__(
...
+            # `nsys stats --report nvtxsum` gives a per-block confirmation
+            # that the fast path actually fired.
+            with nvtx_range("LN+NVFP4 fused", color="green"):
+                return self._forward_nvfp4_fused(hidden_states, scale_msa,
+                                                 shift_msa, seq_len_per_batch,
+                                                 nvfp4_scale)
 
+        with nvtx_range("LN unfused", color="grey"):
+            return self._forward_unfused(hidden_states, residual, scale_msa,
+                                         shift_msa)
```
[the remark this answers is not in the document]
anikaj-eng: addressed in last commit

tensorrt_llm/_torch/modules/mlp.py
```diff
@@ -115,7 +138,14 @@ def forward(
...
         if self._use_fused_relu2_quant:
-            x_act = self._fused_relu2_quant(x_up)
+            # Distinct NVTX label so the chunk-1 fused activation+quant fast
+            # path is visible in nsys traces (separate from the auto layerwise
+            # marker on `MLP.forward`).
+            with nvtx_range("relu2+NVFP4 fused", color="green"):
+                x_act = self._fused_relu2_quant(x_up)
+        elif self._use_fused_gelu_tanh_quant:
+            with nvtx_range("gelu_tanh+NVFP4 fused", color="green"):
+                x_act = self._fused_gelu_tanh_quant(x_up)
```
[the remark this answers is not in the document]
anikaj-eng: addressd in last commit

tests/unittest/_torch/visual_gen/test_wan22_nvfp4_fusion_integration.py
```diff
@@ -0,0 +1,498 @@
...
+
+import fnmatch
+import os
+
+os.environ.setdefault("TLLM_DISABLE_MPI", "1")
+# Enable the LayerNorm fusion path for this test; the production default is
+# off (TRTLLM_DISABLE_NVFP4_LAYERNORM_FUSION=1) so that mainline behavior is
+# unchanged until benchmarks gate it on.
+os.environ.setdefault("TRTLLM_DISABLE_NVFP4_LAYERNORM_FUSION", "0")
+
```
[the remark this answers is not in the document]
anikaj-eng: addressed in last commit

tests/unittest/_torch/visual_gen/test_wan22_nvfp4_fusion_integration.py
```diff
@@ -0,0 +1,498 @@
...
+)
+
+
+def _has_sm100() -> bool:
+    if not torch.cuda.is_available():
+        return False
+    try:
+        sm = get_sm_version()
+    except Exception:
+        return False
```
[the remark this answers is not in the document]
anikaj-eng: addressed in last commit

anikaj-eng:
Thanks @luyiyun1021. I rechecked the current head (`32e2ab08`) against the review feedback; the comments are addressed:

- CodeRabbit batch: added mode-specific fused LayerNorm pointer checks, `sf_scale.numel() == 1` validation for both fused activation ops, strict `N == 5120` validation, AdaLN pair validation, rank-preserving FP4 helper outputs, scoped env-var fixture setup, narrowed `_has_sm100()` exception handling, and `sf_fused` vs `sf_separate` coverage.
- Shared LayerNorm hot path: NVTX ranges are now only emitted for NVFP4-configured layers, so default LayerNorm callers no longer pay unconditional push/pop overhead.
- Per-token timestep modulation: `[B, S, N]` modulation with `S > 1` now falls back to `_forward_unfused` instead of hitting the fused kernel shape guard, with a Python-only regression test.
- Structured-review nits: modulation precision delta is documented, dead `multiProcessorCount` plumbing is removed, relu2/GELU-tanh share the templated `fusedActQuantizeKernel`, the MLP `input_scale` comment is corrected, the hard-coded `modeling_llama.py` line citation is removed, and the personal scratch path fallback was replaced with the repo-conventional CI model path.
- PR-level perf/docstring notes: the PR description now calls out the B200 A/B/nsys run as the exact measured config, and docstrings were added for the PR diff coverage warning.

Ready for re-review.

tests/unittest/_torch/visual_gen/test_wan22_nvfp4_fusion_integration.py
```diff
@@ -0,0 +1,559 @@
...
+    Prefer the env var ``WAN22_T2V_NVFP4_PATH``; fall back to the
+    conventional ``/models`` bind-mount used by the build container.
+    Per repo convention, shared test data also lives under
+    ``/home/scratch.trt_llm_data_ci/llm-models/`` on CI runners.
+    """
+    candidates = [
+        os.environ.get("WAN22_T2V_NVFP4_PATH"),
+        "/models/Wan2.2-T2V-A14B-Diffusers-NVFP4",
+        "/home/scratch.trt_llm_data_ci/llm-models/Wan2.2-T2V-A14B-Diffusers-NVFP4",
+    ]
```
zhenhuaw-me, line 119: Use utils to get the model dir

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -812,3 +932,30 @@ def post_load_weights(self) -> None:
...
+        total_blocks = len(self.blocks)
+        logger.info(
+            "Wan NVFP4 fused-LayerNorm scale attach: norm1=%d/%d, norm2=%d/%d, norm3=%d/%d",
+            n_attached_n1,
+            total_blocks,
+            n_attached_n2,
+            total_blocks,
+            n_attached_n3,
+            total_blocks,
+        )
```
zhenhuaw-me, line 961: Do we need this log?

tensorrt_llm/_torch/modules/layer_norm.py
```diff
@@ -66,57 +89,265 @@ def __init__(
...
+                return self._forward_nvfp4_fused(hidden_states, scale_msa,
+                                                 shift_msa, seq_len_per_batch,
+                                                 nvfp4_scale)
+
+        if self.is_nvfp4:
+            # Counterpart range only emitted when fusion was disabled via the
+            # kill switch on an NVFP4-configured layer (the A/B case). Plain
+            # LayerNorm users skip this entirely so no NVTX push/pop overhead
+            # leaks into the shared hot path.
+            with nvtx_range("LN unfused", color="grey"):
```
zhenhuaw-me, line 207: For such fine-grain nvtx trace, we should leave to debug code locally.

tensorrt_llm/_torch/modules/layer_norm.py
```diff
@@ -66,57 +89,265 @@ def __init__(
...
+        nvfp4_scale = getattr(self, "nvfp4_scale", None)
+        if (self.is_nvfp4 and nvfp4_scale is not None and residual is ...
+                and not per_token_modulation):
+            # NVTX label kept only on the NVFP4 path so the unfused fallback
+            # is byte-identical to the pre-fusion baseline for the many
+            # non-NVFP4 LayerNorm callers (StarCoder2, Cohere2, Qwen3VL,
+            # Kimi-K2.5, Flux, Flux2, ...). The label is what lets
+            # `nsys stats --report nvtx_sum` confirm the fast path fired
+            # per block.
+            with nvtx_range("LN+NVFP4 fused", color="green"):
```
zhenhuaw-me, line 197: For such fine-grain nvtx trace, we should leave to debug code locally.

tensorrt_llm/_torch/visual_gen/models/wan/transformer_wan.py
```diff
@@ -11,13 +12,79 @@
...
+            --model_path <ckpt> --enable_layerwise_nvtx_marker ...
+        # fusion on
+        TRTLLM_DISABLE_NVFP4_LAYERNORM_FUSION=0 nsys profile -o on.nsys-rep \\
+            python examples/.../visual_gen_wan_t2v.py \\
+            --model_path <ckpt> --enable_layerwise_nvtx_marker ...
+        nsys stats --report nvtxsum on.nsys-rep off.nsys-rep
+    """
+    if quant_config is None or force_dynamic_quant:
+        return None
+    if os.environ.get("TRTLLM_DISABLE_NVFP4_LAYERNORM_FUSION", "1") == "1":
```
luyiyun1021, line 52: could you make fuse_nvfp4_ln a block-level local flag, like fuse_qk_norm_rope, not in config or using env variable.

tests/unittest/_torch/visual_gen/bench_wan22_nvfp4_fusion.py
luyiyun1021, line 1: Bench files should not be in unittest folder. Do we have bench tests in visual gen? @zhenhuaw-me @chang-l

tensorrt_llm/_torch/modules/layer_norm.py
luyiyun1021, line 1: I'm wondering we may not modify the gerenal layer norm module and expose the quantize_type / is_nvfp4 / nvfp4_scale / _forward_nvfp4_fused / AdaLN scale_msa,shift_msa,seq_len_per_batch args to it. I suggest using class like DiTLayerNorm(LayerNorm) to do the extra work here.

luyiyun1021 (commented):
Thanks for the work. Left some comments. Could you please trim your comments seems lots of comments added to the generl module like mlp or layernorm and some are dev comments and verbosed.

tensorrt_llm/_torch/modules/mlp.py
```diff
@@ -91,31 +102,79 @@ def __init__(
...
+        # `not force_dynamic_quantization` to ensure the GEMM sees a valid
+        # calibrated alpha. NVFP4 layers do not set `has_static_input_scale`
+        # (that flag is FP8-only), so `force_dynamic_quantization` plus the
+        # `input_scale is not None` check above are the canonical signals.
+        has_kernel_gelu = hasattr(torch.ops.trtllm, 'fused_gelu_tanh_quantize')
+        is_gelu_tanh = self.activation is gelu_tanh
+        not_dynamic = not getattr(self.down_proj, "force_dynamic_quantization",
+                                  False)
+
+        self._use_fused_gelu_tanh_quant = (has_nvfp4 and has_kernel_gelu
```
luyiyun1021, line 146: **[Nit · mergeable-branches + gate inconsistency]**

The two parallel `_use_fused_*` booleans here + the matching `if/elif` in `forward` differ only by which helper they call and the NVTX label. Consider collapsing to a single resolved op handle, mirroring how `GatedMLP` does the same NVFP4 activation+quant fusion (`gated_mlp.py`: `_can_fuse_gate_up_swiglu*` -> one parameterized `_fused_gate_up_swiglu(x, fp4_out=...)`):

```python
# create_weights
self._fused_act_quant_op = None
if has_nvfp4 and has_scale and not_dynamic:
    if self.activation is relu2 and has_kernel:
        self._fused_act_quant_op = torch.ops.trtllm.fused_relu2_quantize
    elif self.activation is gelu_tanh and has_kernel_gelu:
        self._fused_act_quant_op = torch.ops.trtllm.fused_gelu_tanh_quantize

# forward
if self._fused_act_quant_op is not None:
    with nvtx_range("act+NVFP4 fused", color="green"):
        x_act = self._fused_act_quant(x_up, self._fused_act_quant_op)
else:
    x_act = self.activation(x_up)
```

Separately, the two gates are **inconsistent**: this gelu gate requires `not_dynamic` (the comment explains it guards a stale `module.alpha` when an `Fp4QuantizedTensor` is fed under dynamic quant), but the relu2 gate just above (`self._use_fused_relu2_quant = ...`) does **not** include `not_dynamic`. relu2 also returns an `Fp4QuantizedTensor` and hits the same alpha path -- should relu2 also gate on `not_dynamic`, or is the gelu `not_dynamic` over-cautious? Unifying the gates would force this to be decided in one place.

tensorrt_llm/_torch/modules/mlp.py
```diff
@@ -135,6 +206,36 @@ def _fused_relu2_quant(self, x: torch.Tensor) -> Fp4QuantizedTensor:
...
+        if len(orig_shape) > 2:
+            fp4_tensor = fp4_tensor.view(*orig_shape[:-1], orig_shape[-1] // 2)
+
+        return Fp4QuantizedTensor(
+            fp4_tensor=fp4_tensor,
+            scaling_factor=sf_tensor,
+            is_sf_swizzled=True,
+        )
+
+    def _fused_gelu_tanh_quant(self, x: torch.Tensor) -> Fp4QuantizedTensor:
```
luyiyun1021, line 218: **[Nit · DRY]** `_fused_gelu_tanh_quant` is identical to `_fused_relu2_quant` (just above) except for the op symbol (`fused_gelu_tanh_quantize` vs `fused_relu2_quantize`). Worth collapsing into one op-parameterized helper:

```python
def _fused_act_quant(self, x, op):
    orig = x.shape
    xf = x.view(-1, x.shape[-1]).contiguous()
    if xf.dtype not in (torch.float16, torch.bfloat16):
        xf = xf.to(torch.bfloat16)
    fp4, sf = op(xf, self.down_proj.input_scale, 16)
    if len(orig) > 2:
        fp4 = fp4.view(*orig[:-1], orig[-1] // 2)
    return Fp4QuantizedTensor(fp4, sf, is_sf_swizzled=True)
```

This mirrors `GatedMLP._fused_gate_up_swiglu(x, fp4_out=...)` (one parameterized helper). Note the CUDA side was already unified into the templated `fusedActQuantizeKernel<T, Act>` -- these Python helpers are the remaining duplicated half.

luyiyun1021 (commented):
Another pass focused on the MLP fused path -- structure/DRY suggestions (non-blocking) plus one gate-consistency question. The fusion *approach* matches the existing `GatedMLP` / relu2 precedent; these are about tightening the *implementation* to that same single-parameterized-helper style.

tests/unittest/_torch/modules/test_fused_layernorm_quant.py
```diff
@@ -0,0 +1,267 @@
...
+    """Return True iff the ``trtllm::fused_layernorm_quantize`` op is registered."""
+    return hasattr(torch.ops, "trtllm") and hasattr(torch.ops.trtllm, "fused_layernorm_quantize")
+
+
+def fp4_quantize_available():
+    """Return True iff the reference ``trtllm::fp4_quantize`` op is registered."""
+    return hasattr(torch.ops, "trtllm") and hasattr(torch.ops.trtllm, "fp4_quantize")
+
+
+skip_unless_fused_layernorm_and_fp4 = pytest.mark.skipif(
```
luyiyun1021, line 49: **[Nit · follow-existing-patterns]** The `getSMVersion() < 100` half here re-rolls the repo-standard `skip_pre_blackwell` (`tests/unittest/utils/util.py:113`), which is what sibling kernel tests use (`test_fused_moe`, `test_awq_quantization`, `test_mla_helix`, `test_modeling_kimi_k25`, the dsa tests, ...). The op-availability AND (`not fused_layernorm_quantize_available()`) is a legitimate extra condition, but the SM part should compose the shared decorator instead of duplicating the `< 100` literal:

```python
from tests.unittest.utils.util import skip_pre_blackwell

@skip_pre_blackwell
@pytest.mark.skipif(
    not (fused_layernorm_quantize_available() and fp4_quantize_available()),
    reason="requires trtllm fused_layernorm_quantize + fp4_quantize ops",
)
def test_...():
    ...
```

Same applies to the gelu gates in `test_fused_activation_quant.py` (`skip_unless_fused_gelu_tanh_*`). Keeps the Blackwell-version policy in one place so a future SM-range change doesn't have to be chased across several hand-rolled `getSMVersion() < 100` literals.

tests/unittest/_torch/visual_gen/test_wan22_nvfp4_fusion_integration.py
```diff
@@ -0,0 +1,559 @@
...
+    try:
+        sm = get_sm_version()
+    except RuntimeError:
+        # get_sm_version() raises RuntimeError when CUDA device query fails
+        # (e.g., driver not loaded). Per the TRT-LLM coding guideline that
+        # except clauses should be narrowed to expected errors, only swallow
+        # that specific failure; anything else (e.g., ImportError) should
+        # propagate so it isn't masked.
+        return False
+    return 100 <= sm < 120
```
luyiyun1021, line 149: **[Nit · follow-existing-patterns + a question for the code]** `_has_sm100()` reimplements two decorators that already exist in `tests/unittest/utils/util.py`: `100 <= sm` is `skip_pre_blackwell` (:113), and the `< 120` upper bound is exactly `skip_blackwell_geforce` (`getSMVersion() == 120`, :119). Consider composing them instead of a hand-rolled helper (drop-in, since `skip_if_no_sm100` is used as a plain decorator):

```python
from tests.unittest.utils.util import skip_pre_blackwell, skip_blackwell_geforce

@skip_pre_blackwell
@skip_blackwell_geforce
@skip_if_no_checkpoint
def test_...():
    ...
```

The `< 120` bound mirrors the production gate in `layer_norm.py` (`is_nvfp4` is cleared unless `100 <= sm < 120`), so excluding SM120 *here* is consistent with the code under test. But it raises a real question about the code itself: **why is SM120 excluded from the fused path at all?** SM120 is Blackwell and supports `cvt.rn.satfinite.e2m1x2.f32`, so the kernel should run there. If the exclusion is intentional (e.g., the kernel was only validated on SM100/SM10x), a one-line comment at the `is_nvfp4` gate would help; if not, both the gate and this test should allow `sm >= 100`.

luyiyun1021 (commented):
Test-convention nits: prefer the shared `skip_pre_blackwell` / `skip_blackwell_geforce` decorators (`tests/unittest/utils/util.py`) over hand-rolled `getSMVersion() < 100` / `_has_sm100()` gates, so the Blackwell-version policy stays in one place. The integration-test one also surfaces a question about the `< 120` (SM120) exclusion in the production `is_nvfp4` gate.

PR #14778, opened 2026-05-30
[None][feat] Port 13 AutoDeploy custom models to sharding IR + opt them in via registry

greg-kwasniewski1:
## Summary
Ports 13 AutoDeploy custom modeling files (`tensorrt_llm/_torch/auto_deploy/models/custom/modeling_*.py`) to use sharding IR, validated by the offline equivalence test introduced in #13963. Adds `enable_sharder_ir.yaml` to 70 model_registry entries so they actually exercise `apply_sharding_hints` at runtime — same opt-in pattern PR #13478 used for its 4 families.

This is **scope-restricted**: the default sharding pipeline in `default.yaml` stays on legacy (`detect_sharding`); the flip to `apply_sharding_hints` as default is deferred to a follow-up PR (see [Future work](#future-work)).

## Ported models (13)

| Family | File | Architecture |
|---|---|---|
| Dense | `modeling_gemma.py` | Gemma 1 — MHA + SwiGLU |
| Dense | `modeling_mistral.py` | Mistral — GQA + SwiGLU + sliding window |
| Dense | `modeling_granite.py` | Granite — GQA + SwiGLU + scaling multipliers |
| Dense | `modeling_hunyuan_dense.py` | HunYuan Dense V1 — GQA + per-head QK norm (post-RoPE) |
| Dense | `modeling_seed_oss.py` | Seed-OSS — GQA + SwiGLU |
| Dense | `modeling_gemma2.py` | Gemma 2 — sliding window + attn/final logit softcap |
| Dense | `modeling_cohere.py` | Cohere — parallel attn+MLP residual; LayerNorm; interleaved RoPE |
| Dense | `modeling_exaone.py` | EXAONE 3.5 — non-standard naming; bundled config |
| Dense | `modeling_llama3.py` | Llama 3 (3, 3.1, 3.2, 3.3) — GQA + SwiGLU |
| MoE | `modeling_deepseek_v2.py` | DeepSeek-V2 — MLA + MoE w/ shared expert |
| MoE | `modeling_glm4_moe.py` | GLM4 MoE — GQA + partial RoPE + QK norm |
| MoE | `modeling_qwen3_moe.py` | Qwen3 MoE — GQA + per-head QK norm |
| MoE | `modeling_llama4.py` | Llama 4 — GQA + L2 QK norm + NoPE + BMM-style MoE |

[...]

PR #14786, opened 2026-05-30
[None][feat] Sharding IR default (auto-detect) + Eagle/MTP draft sharding + qwen3_next

### LINKED ISSUE #14642: [Feature Request] Make sharding IR the AutoDeploy default; port remaining text-only custom models
## Summary
Transition AutoDeploy from the legacy detect-and-execute sharding pipeline to the sharding-IR pipeline as the project-wide default. Port all remaining text-only `models/custom/*` models to sharding IR (validated by the offline equivalence test from #13963), promote `modeling_llama3_ir.py` to canonical, then flip `default.yaml` so the legacy pipeline becomes opt-in via a new `enable_legacy_sharding.yaml`.

## Proposed approach
- Branch off PR #13963 to inherit the offline equivalence test infrastructure; rebase onto upstream/main once #13963 merges.
- Port 18 dense / standard text-only models (Tier A): llama3, mistral, gemma, gemma2, granite, internlm3, openelm, cohere, decilm, exaone, smollm3, eagle, nemotron_flash, olmo3, seed_oss, starcoder2, minimax_m2, hunyuan_dense.
- Port 10 MoE-only text models (Tier B): deepseek_v2, glm4_moe, glm4_moe_lite, glm_moe_dsa, gpt_oss, granite_moe_hybrid, hunyuan_moe, qwen3_moe, qwen3_next, llama4.
- Promote `modeling_llama3_ir.py` to canonical `modeling_llama3.py` (with a merge-forward of any ToT updates landed on the legacy file since the IR variant was forked).
- Edit `tensorrt_llm/_torch/auto_deploy/config/default.yaml`: remove `detect_sharding` and `sharding_transform_executor` blocks; enable `apply_sharding_hints` by default.
- Create new `tensorrt_llm/_torch/auto_deploy/config/enable_legacy_sharding.yaml` override that re-enables the legacy pipeline for opt-in use.
- Update model registry: drop now-redundant `enable_sharder_ir.yaml` references; add `enable_legacy_sharding.yaml` to the 5 deferred MoE+multimodal models (gemma4, gemma3n, kimi_k2, mistral3, skywork_r1v2).
- Delete obsolete `examples/auto_deploy/model_registry/configs/enable_sharder_ir.yaml`.

## Scope
- Surface: `tensorrt_llm/_torch/auto_deploy/models/custom/`, `tensorrt_llm/_torch/auto_deploy/config/`, `examples/auto_deploy/model_registry/`.
- Affected components: AutoDeploy sharding pipeline, model registry, default config.
- Out of scope: the 5 MoE+multimodal models (gemma4, gemma3n, kimi_k2, mistral3, skywork_r1v2) — deferred to a follow-up; they will stay on the legacy pipeline via `enable_legacy_sharding.yaml` until ported separately.

## Tests / validation
- Per-port: offline equivalence test (`pytest tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_ir_equivalence.py --sharding-ir-modeling-file <modeling_X.py> -s -v`) across tp / ep / tep / attn-dp configs.
- Full AutoDeploy unit suite: `pytest tests/unittest/auto_deploy -x`.
- Multi-GPU TP suite: `pytest tests/unittest/auto_deploy/multigpu -x`.
- `trtllm-serve` smoke tests at TP=2: one text-only model (new default path) + one deferred MoE+multimodal model (legacy opt-in path) — to verify both code paths work end-to-end.

## Risk
**Library-visible**: changes default sharding behavior for every model not in the deferred-multimodal exception list. Multi-GPU AutoDeploy CI stages (`DGX_B200-4_GPUs-AutoDeploy-1`, `DGX_H100-4_GPUs-AutoDeploy-1`) mandatory. No public API changes.

## Links
- Depends on / inherits: #13963 (offline equivalence test infrastructure, still open)
- Builds on: #13478 (sharding IR default for 4 model families, merged)
- Related PR: (filled in once PR is opened)

### LINKED ISSUE #14784: [Feature Request] Make sharding IR the AutoDeploy default with auto-detect fallback to legacy
## Summary
Make `apply_sharding_hints` the AutoDeploy default sharding pipeline, with a runtime auto-detection mechanism that falls back to legacy `detect_sharding` + `sharding_transform_executor` for models whose FX graphs lack sharding-IR markers. Add an explicit `enable_legacy_sharding.yaml` opt-out for users who need to force legacy on otherwise-IR-marked models. Drop the now-redundant per-entry `enable_sharder_ir.yaml` opt-ins added by #14778 (PR Fixes #14642).

## Why
Today, sharding IR is opt-in per registry entry via `enable_sharder_ir.yaml`. With 17 IR-ported modeling files (4 from #13478, 13 from #14778), enumerating opt-ins in `models.yaml` is repetitive and error-prone — every new HF model checkpoint that uses one of those 17 architectures has to be manually added. Conversely, ~30+ non-custom HF models still need legacy because their FX graphs have no IR hints. The current design pushes that bookkeeping to every registry entry.

A runtime auto-detect inside the sharding pipeline eliminates the bookkeeping: the FX graph itself declares which pipeline it wants, via the presence of `torch.ops.auto_deploy.all_reduce` nodes — a marker that legacy `detect_sharding` never inserts and that the IR-port recipe (skill rule A3) always emits.

## Proposed approach
1. **Default flip in `tensorrt_llm/_torch/auto_deploy/config/default.yaml`**: enable `apply_sharding_hints` by default; keep `detect_sharding` + `sharding_transform_executor` enabled as fallback.
2. **Auto-detect dispatcher**: inside `apply_sharding_hints`, check whether the exported FX graph contains any `torch.ops.auto_deploy.all_reduce` node. If yes → apply hints and mark `gm.meta["sharding_ir_applied"] = True`. If no → no-op (let the legacy passes handle it). In `detect_sharding` / `sharding_transform_executor`, short-circuit when `gm.meta.get("sharding_ir_applied", False)` is set.
3. **Explicit opt-out**: create `tensorrt_llm/_torch/auto_deploy/config/enable_legacy_sharding.yaml` that disables `apply_sharding_hints` and forces legacy. Useful when a model has IR hints but they are known to be broken / under investigation.
4. **Registry cleanup**: remove `enable_sharder_ir.yaml` from all `yaml_extra` lists in `examples/auto_deploy/model_registry/models.yaml` (those entries are now redundant — auto-detect handles them). Delete the now-obsolete `examples/auto_deploy/model_registry/configs/enable_sharder_ir.yaml`.

## Why the all_reduce marker is reliable
- **No false positives**: `torch.ops.auto_deploy.all_reduce` is a sharding-IR-only op. Legacy `detect_sharding` inserts plain `dist.all_reduce` nodes; it never emits `auto_deploy.all_reduce`.
- **No false negatives**: skill rule A3 of `ad-sharding-ir-port` requires `all_reduce` after every rowwise projection / MoE merge point. Every IR-ported file in `models/custom/` has at least one such node (dense MHA + SwiGLU MLP alone produce two — after `o_proj` and after `down_proj`).

## Scope
- Surface: `tensorrt_llm/_torch/auto_deploy/config/`, `tensorrt_llm/_torch/auto_deploy/transform/library/sharding.py` (skip-when-IR guard), `tensorrt_llm/_torch/auto_deploy/transform/library/sharding_ir.py` (auto-detect + meta flag), `examples/auto_deploy/model_registry/`.
- Affected components: AutoDeploy sharding pipeline (legacy and IR both retained, behaviorally exclusive), model registry.
- Out of scope: porting the remaining 20 legacy modeling files (separate follow-up); the 4 OTHER_FAIL models (eagle, nemotron_flash, qwen3_next, minimax_m2) need individual debug.

## Tests / validation
- Existing `tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_ir_equivalence.py` must still pass on all 17 IR-ported models (4 from #13478 + 13 from #14778) at all 4 dist configs.
- Smoke run a non-IR-ported model at TP=2 via `trtllm-serve` / `build_and_run_ad` to verify the legacy auto-fallback works without `enable_legacy_sharding.yaml` being explicitly applied.
- Smoke run a non-IR-ported model at TP=2 with `enable_legacy_sharding.yaml` to verify the explicit opt-out path.
- Sabotage-style negative control: pick an IR-ported model and confirm that `enable_legacy_sharding.yaml` forces it through the legacy path (output should match TP=1 eager).

## Risk
**Library-visible, medium**. Changes default sharding behavior for every model loaded via AutoDeploy. The auto-detect mechanism is safe in principle (no false positives or negatives from the marker) but exercises the legacy and IR pipelines via a new dispatch path that is itself untested at PR-open time. Multi-GPU AutoDeploy CI stages mandatory.

## Links
- Builds on (must merge first): #14778 (Fixes #14642 — 13 ports + skill cleanup + registry opt-ins)
- Builds on (already merged): #13478 (initial 4 IR-ported families + `enable_sharder_ir.yaml`), #13963 (offline equivalence test)
- Related PR: (filled in once PR is opened)

greg-kwasniewski1:
## Summary
Three related advances toward making sharding IR the universal AutoDeploy sharding path. All build on #14778 (`Fixes #14642` — 13 text-only ports + registry opt-ins); **this PR is stacked on #14778 and its diff includes those commits until #14778 merges**, after which it will be rebased to a clean diff.

1. **Auto-detect default flip.** `apply_sharding_hints` becomes the default and self-detects whether a graph carries sharding-IR markers (`torch.ops.auto_deploy.all_reduce`); if not, the legacy `detect_sharding` + `sharding_transform_executor` pipeline runs as a transparent fallback. A new `enable_legacy_sharding.yaml` forces legacy explicitly. The 93 per-entry `enable_sharder_ir.yaml` opt-ins are dropped (now redundant) and that override file is deleted.
2. **Eagle / MTP draft sharding (Llama + NemotronH).** The Eagle draft model is ported to sharding IR so the draft GraphModule shards deterministically via `apply_sharding_hints` (same path as base models), with an offline draft-equivalence test harness.
3. **qwen3_next sharding-IR port.** GatedDeltaNet + Attention + MLP + MoE all converted and validated (see below).

## Dispatcher design
`has_sharding_ir_markers(gm)` returns True iff the FX graph contains any `torch.ops.auto_deploy.all_reduce` node — emitted only by sharding-IR-authored modeling files (skill rule A3), never by legacy `detect_sharding`. So it is a sufficient, false-positive-free marker. `apply_sharding_hints` skips (deferring to legacy) when absent; when present it applies and sets `gm.meta["sharding_ir_applied"]`, which `detect_sharding` / `sharding_transform_executor` honor by short-circuiting. Both pipelines stay registered; exactly one runs per graph.

## Eagle draft sharding
- `Eagle3Attention` / `EagleMLP` (Llama draft) get the standard MHA / SwiGLU IR hints. The q/k/v `2*hidden_size` input is the replicated concat of normed embeds + hidden; only the head output shards, so colwise is unchanged.
- NemotronH draft needs no change: `NemotronHEagleLayer` reuses the already-IR-ported `NemotronHAttention` / `NemotronHMOE` mixers.
- The Eagle fusion linears (`fc`, `eh_proj`) and `embed`/`lm_head`/`d2t` stay plain `nn.Linear` → `ShardableNode.from_node` returns `None` → correctly replicated (their I/O is the replicated residual stream).
- Test infra: `spec_from_modeling_file` now walks `AutoModelForCausalLMFactory` subclasses (so models registered on `EagleDrafterFactory` / `NemotronFlashForCausalLMFactory` resolve); new `--sharding-ir-eagle-draft <model_type>` option + `test_sharding_ir_eagle_draft_equivalence` build the draft on a tiny config and feed its `(inputs_embeds, position_ids, hidden_states)` contract, reusing the same export/shard/compare tail.

## Validation
- Eagle Llama + NemotronH drafts: pass `tp-only / ep-only / tep / attn-dp` (rel_rmse ~1e-3, tol 0.02); both correctly FAIL under `SHARDING_IR_SABOTAGE=1` (rel_rmse ~0.07).
- Auto-detect fallback: base-model IR path unchanged (gemma re-checked green); non-marked graphs fall through to legacy.

## qwen3_next sharding (now correct)
`modeling_qwen3_next.py` (GatedDeltaNet + Attention + MLP + MoE) is ported to sharding IR and passes the offline equivalence test on `tp-only / ep-only / tep / attn-dp` (rel_rmse ~0.011, tol 0.02), failing correctly under `SHARDING_IR_SABOTAGE=1` (rel_rmse ~1.04). Two bugs were fixed during the port: (1) a post-gated-norm reshape that baked the unsharded `num_v_heads` (now `auto_deploy.view` with `tp_scaled_dim=2`); (2) `in_proj_qkvz` is an interleaved-per-head fused projection, so it must use plain colwise — the earlier `output_sizes=[...]` treated it as contiguous Q|K|V|Z blocks and scrambled the layout.

## Risk
**Library-visible, medium.** Changes the default AutoDeploy sharding pipeline for every model. Mitigated by the conservative marker check (no false positives by construction) and the fully-preserved, re-enterable legacy pipeline (implicit fallback + explicit `enable_legacy_sharding.yaml`).

## Out of scope
- nemotron_flash / minimax_m2 standalone ports (separate diagnostics).
- Vocab-parallel sharding of embed / lm_head / d2t.
- MTP-via-manual-tp_plan SuperV3 config (`super_v3_mtp.yaml`) — different mechanism.

Fixes #14784


greg-kwasniewski1:

greg-kwasniewski1:

greg-kwasniewski1:
/bot run --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1" --disable-fail-fast --post-merge

greg-kwasniewski1:
/bot run --extra-stage "DGX_B200-4_GPUs-AutoDeploy-1, DGX_H100-4_GPUs-AutoDeploy-1" --disable-fail-fast --post-merge

jieli-matrix (approved):

greg-kwasniewski1:

examples/auto_deploy/model_registry/configs/nano_v3.yaml
```diff
@@ -7,6 +7,15 @@ cuda_graph_config:
 kv_cache_config:
   free_gpu_memory_fraction: 0.88
 transforms:
+  apply_sharding_hints:
+    # Mamba (layer_type="ssm") in_proj is a fused [z|x|B|C|dt] weight. Its dt
+    # slice equals mamba_num_heads (=64 here), which cannot be NVFP4-aligned
+    # (multiple of 32) when TP-split at this world size (64 < 4*32), and it must
```
govind-ramnarayan, line 13: nit: There is no world_size in this config right - it's separate? And I guess this is an issue for world_size=4 but would work with world_size=2? I wonder if it makes sense to do "best effort" sharding in this case to reduce coupling between the model config and the world_size, although of course performant configs will generally be tuned for a world_size...

greg-kwasniewski1: Yes and no... the problem is that it is not as clean to fix it as in GQA with small number of kv heads. For this, we we have `min_local_shape`. To make mamba shardable with small number of mamba heads, it cannot be done cleanly with the current architecture, because that would require partial replication that would have to propagate to the out_linear layer, which breaks sharding IR locality founding assumption.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_gpt_oss.py
```diff
@@ -425,6 +425,9 @@ def forward(
 
         # ``torch_attention`` handles GQA natively; sinks / sliding_window are
         # per-call kwargs. Causal mask is applied internally for prefill.
+        # ``enable_sharding=True`` triggers ``WeightedParamShardableNode`` to slice
```
govind-ramnarayan, line 428: Q: was this a bug in gpt oss sharding that's being fixed?

tensorrt_llm/_torch/auto_deploy/transform/library/quantization.py
```diff
@@ -462,8 +462,17 @@ def load_hook(self, state_dict, prefix, *args, weight_name):
                     TRTLLM_NVFP4_SCALING_VECTOR_SIZE,
                     False,
                 )
+                # ``fp4_quantize`` returns the (swizzled) block scale as a flat
```
govind-ramnarayan, line 465: What does this have to do with this PR? Is it something needed to make qwen3_next bf16 work or something? Does it have any affect on correctness for nvfp4 models? Maybe we can make sure to check some of these (not sure what nvfp4 tests we have in pre-merge now) before merging?

tests/unittest/auto_deploy/_utils_test/_deterministic_routing.py
```diff
@@ -0,0 +1,255 @@
...
+"""Context-manager intercept that forces MoE routing to pick experts ``[0..top_k-1]``.
+
+Why this exists
+---------------
+
+The sharding-IR equivalence test runs an unsharded and a sharded copy of the
+same exported graph and compares per-token logits. MoE top-k routing is a
+non-smooth argmax: under bf16 reduction-order noise the two sides can pick
+*different* experts for the same token, producing per-token errors that look
+like sharding bugs but are really finite-precision artifacts.
```
govind-ramnarayan, line 18: This is a really good docstring. Are there other operations that are this noise-sensitive? I'm a bit worried about how we track down errors that are due to this for a new model where we don't understand the ops and implementations in it deeply. Maybe agents can help with tracking down these ops in the future or we can come up with some skills / unit tests...

tests/unittest/auto_deploy/_utils_test/_sharding_ir_helpers.py
```diff
@@ -432,6 +385,23 @@ class *name* (the registration key) -- works for IR files that import
 # -----------------------------------------------------------------------------
 
 
+def build_tiny_config(cfg_cls: type) -> Any:
```
govind-ramnarayan, line 388: nice!

tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_num_correctness.py
```diff
@@ -0,0 +1,910 @@
...
+The test is skipped (not failed) when ``--sharding-ir-modeling-file`` is
+absent. No filename pattern is assumed -- works for the canonical
+``modeling_<name>.py`` files that became the IR-aware default in #13478
+(deepseek, nemotron_h, qwen3, qwen3_5_moe), and for any future modeling
+file that opts into the sharding-IR path. See
+``_sharding_ir_helpers.spec_from_modeling_file`` for how class / config
+inference works, and ``_apply_layer_count_dependent_quirks`` for the
+hasattr-driven layer-count-dependent config patches.
+
+The ``SHARDING_IR_SABOTAGE=1`` env var is a negative-control switch: it
```
govind-ramnarayan, line 46: Is this something you used to manually test? Is it exercised at all in CI / these tests?

tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_num_correctness.py
```diff
@@ -0,0 +1,910 @@
...
+    if rank == 0:
+        print(
+            f"[sharding-ir-eq] sharding_kind={sharding_kind}: "
+            f"inserted {n_collectives} collective op(s)",
+            flush=True,
+        )
+    return gm_sharded, sharding_kind, n_collectives
+
+
+# IR-vs-legacy dispatch uses the SAME helper the production pipeline uses
```
govind-ramnarayan, line 156: nit: Maybe belongs in the docstring above or something? Took me a bit to understand that this is resolved because you are importing `is_shardingIR_enabled` above, I wasn't sure if this is a current issue or not.

tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_num_correctness.py
```diff
@@ -0,0 +1,910 @@
...
+    rank: int,
+    world_size: int,
+) -> None:
+    """Worker entry; mirrors the per-rank traceback to a log file on failure."""
+    import traceback
+
+    try:
+        _run_eagle_draft_equivalence_job_impl(model_type, rank, world_size, dist_config_name)
+    except BaseException:
+        with open(f"/tmp/sharding_ir_eagle_draft_rank{rank}.log", "w") as f:
```
govind-ramnarayan, line 888: Q: I guess this is for helping us debug when we run locally?

tensorrt_llm/_torch/auto_deploy/transform/library/sharding_ir.py
```diff
@@ -60,16 +80,312 @@
...
+
+def _load_hook(
+    state_dict,
+    prefix,
+    *args,
+    f_split: Callable[[torch.Tensor, int], torch.Tensor],
+    param_key: str,
+    param_shape: torch.Size,
+):
+    # TODO: we need to support loading either a sharded or unsharded checkpoint.
```
govind-ramnarayan, line 194: Is this TODO addressed? It kind of sounded from the tests that it's an identity when unsharded? How urgent is this issue to resolve?

tensorrt_llm/_torch/auto_deploy/transform/library/sharding_ir.py
```diff
@@ -778,19 +1130,38 @@ def _shard_scales(self, gm, dc, weight_nodes):
...
+    # FP4LinearShardableNode); apply the same floor to the fused SwiGLU split.
+    MIN_LOCAL_SHAPE: int = 32
+
     def _shard_scales(self, gm, dc, weight_nodes):
-        weight_shape = weight_nodes.weights[0].tensor.shape if weight_nodes.weights else None
-        if weight_shape is None:
+        if not weight_nodes.weights:
             return
+
+        # Each NVFP4 weight_scale must be de-swizzled with ITS OWN weight's uint8
```
govind-ramnarayan, line 1141: Can you help me generally understand what's going on with NVFP4 here? Are you needing to canonicalize the scales before sharding it? Is it because the weight_scale is in the kernel-specific representation and you need to normalize it, or vice-versa? Why does this need to happen for sharding?

govind-ramnarayan (approved):
Thanks! A lot of random nits and questions, but I am fine with you merging this if the CI is green. In particular I want to understand the interaction with nvfp4 a bit better

examples/auto_deploy/model_registry/configs/nano_v3.yaml
govind-ramnarayan: Got it - I don't fully understand but it sort of makes sense that combinations of replication and sharding will be difficult because you will need a normalization somewhere, and I guess if you can't place that immediately after the sharded op then sharding is not "local" - is that roughly the idea?

greg-kwasniewski1: Yes. 
The core principle behind sharding IR is that whatever heuristic, global information, algorithmic sharding sophistication you do, it should be while generating sharding hints. Once you reach `apply_sharding_hints`, the transformation should be mechanical, deterministic, and local.

But in this case, "opening" linear layer, that do colwise sharding, has to make a decision "do I have enough heads to do TP sharding, or do I have to do partial replication? And if `world_size=8`, but `num_heads=4`, then do I do 4-fold sharding and 2-fold replication?" That would be fine, but the problem is that the "closing" linear layer `o_proj` that does rowwise + all_reduce, doesn't have the information what decision `in_proj` did. `o_proj` doesn't have `reshape`, doesn't even know that there are some splits, fusions, and heads involved. From `o_proj` perspective, this is just a single, large linear layer, and from its local perspective, its totally fine to fit many many more ranks comfortably. So, `in_proj` would do partial sharding, `o_proj` would do full sharding, and suddenly shapes mismatch.

The only solution would be that there is a way to pass the information from `in_proj` to `o_proj`: "hey, do only 4-fold sharding, because I could not fit it, and i had to do partial replication". But this breaks the locality assumption of sharding IR.

There ARE ways to basically hack it, but I consider that this is just not worth an effort, because for the sake of saving TP-sharding of nano-v3 model, that would break the sharding IR contract and could potentially backfire later.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_gpt_oss.py
greg-kwasniewski1: Yes. The problem is the attention sinks. 

So as long as `torch.ops.auto_deploy.torch_attention` was "stateless" op, it just consumed q,k,v, it was not a "shardable" op. It should just naturally infer shapes from the sharded q,k,v. But with the introduction of a shardable parameter `sinks`, the situation is similar to `mla` op which absorbs `kv` inside the kernel.

General rule: if an op has a weight parameter as an input, it should be explicitly labeled as a shardable op with sharding hints, as we need to know whether this weight parameter should be sharded or replicated, and if sharded, then how.

And `gpt_oss` is the first model i worked with that `torch.ops.auto_deploy.torch_attention` op become "weighted", so i had to add it. Without it, it breaks, because sinks would get replicated.

tensorrt_llm/_torch/auto_deploy/transform/library/quantization.py
greg-kwasniewski1: Without it, nvfp4 multi-gpu tests fail on sharing, so it is needed to make CI green.

tests/unittest/auto_deploy/_utils_test/_deterministic_routing.py
greg-kwasniewski1: There is no good answer, to the best of my knowledge...

All my linear algebra knowledge point to one crucial observation: I am too stupid to deeply understand numerical correctness. The theory is just too hard for me. As long as you have only contractions, that is ok-ish, but e.g. factorizations, other non-linear properties, spectral analysis, convergence, I just gave up on understanding the implications of different metrics to calculate matrix norms, etc. And something like top-k is just spitting in my face. Top-k is telling me "on your need and beg for my numerical mercy".

TL;DR:
I don't know how to do it correctly other than trial-and-error setting proper error bounds. This test is fragile and there are possibilities of both false positives and false negatives.

tests/unittest/auto_deploy/_utils_test/_sharding_ir_helpers.py
greg-kwasniewski1: thanks!

tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_num_correctness.py
greg-kwasniewski1: It is something I manually test. As i mentioned in the other comment, there is a risk of false positives and false negatives, so I would definitely not put this as CI blocker test. Maybe we could consider suggesting that a porting agent could use it, but I see it as a part of an "expert" flow - basically, you need to know what you are doing. And I used it quite a lot when manually sharding the "complicated" layers.

tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_num_correctness.py
greg-kwasniewski1: This looks like an auto-generated docstring from a "cleaning" agent, yeah, we can consider completely removing it. I don't have a strong opinion whether to keep it as is, move it, or remove it, but lately I see that these "verbose" comments in the source file are usually generated by agents for agents, so I tend to keep them. But your call.

tests/unittest/auto_deploy/multigpu/transformations/library/test_sharding_num_correctness.py
greg-kwasniewski1: yeap. I hate relying on remote test infrastructure.

tensorrt_llm/_torch/auto_deploy/transform/library/sharding_ir.py
greg-kwasniewski1: Ohboi, honestly, I don't even fully remember the context why I had to apply this fix. If I recall correctly, that is related to my `test_sharding_num_correctness` test, as normally, we do not have a "pre-sharded" checkpoints, but they are sharded during the load hook from a normal checkpoint. 

For these numerical correctness tests, I had several local tests that for example weights are not randomly initialized, but they have some "checkerboard" pattern of integers like -1,0,1,2, etc, so that on the tiny config, I could infer what got wrong where. 

TL;DR:
This change is NOT needed for true e2e flow, it is a no-op and load hook proceeds as expected. It's more a utility to help local testing if something goes wrong.

tensorrt_llm/_torch/auto_deploy/transform/library/sharding_ir.py
greg-kwasniewski1: Compare it with `FP4LinearShardableNode`:
It has only a single weight node, so `weight_shape = weight_nodes.weights[0].tensor.shape if weight_nodes.weights else None` is correct. `weight_shape` is unambiguous. and you are sharding the weight together with its corresponding scale.

The problem is that `FP4SwiGLUShardableNode` ships with 3 weights with 2 different shapes, so `weight_shape = weight_nodes.weights[0].tensor.shape if weight_nodes.weights else None` is correct for gate and up, but this is an incorrect shape for down, because it has its weight transposed. Without this, you would be using `[m,n]` `weight_shape` for the weight that has actually shape `[n,m]` which results in incorrect deswizzling of quantized weights and numerical errors.

You could ask a question: why we are doing weight sharding separately and scale sharding separately. And that would be a good question. And maybe that would be the best answer:

```
zipped_weights_and_scales = [(gate_weight, gate_scale), (up_weight, up_scale), (down_weight, down_scale)]
for weight, scale in zipped_weights_and_scales: 
    shard_weight(weight)
    shard_scale(scale, weight.shape)
```

but because of legacy reasons sharding weights and scales is detached, so basically, if you have multiple scales of different shapes, you need to know what shape to use.


NOTE:
I HATE  these kind of bugs. It took a lot of time to fix, because shape propagation would be ok. It's just you have small numerical errors because quantization scales got slightly misaligned. But all you see is that it looks "almost" correct but somewhere you see that the accuracy slightly dropped....

Issue #14603, opened 2026-05-27
aa


3em0:

venkywonka:
@3em0 , may i ask if the issue has been resolved?

3em0:
> [@3em0](https://github.com/3em0) , may i ask if the issue has been resolved?

didn't. But i have send the report throuh NVIDIA PSIRT.  report: "I have opened a ticket with the following tracking number: 6007052" , do you need me to fix it?

3em0:
> > [@3em0](https://github.com/3em0) , may i ask if the issue has been resolved?
> 
> didn't. But i have send the report throuh NVIDIA PSIRT. report: "I have opened a ticket with the following tracking number: 6007052" , do you need me to fix it?

two month ago

PR #14800, opened 2026-05-31
[https://nvbugs/6226933][fix] canonicalize multimodal cache-key serialization to prevent hash collisions

venkywonka:
## Problem
Multimodal KV-cache block reuse (`enable_block_reuse=True` by default) keys cache blocks on a BLAKE3 content hash of each multimodal item (`apply_mm_hashes`). The serialization feeding that hash omitted semantic structure, so semantically different inputs produced identical bytes and therefore identical cache keys, which can cause incorrect cross-request KV-cache reuse / cache poisoning for multimodal serving. Reported in #14603.

On `main`, the following distinct inputs collided to the same hash:
- PIL images with transposed dimensions (e.g. 30×100 vs 100×30)
- ndarrays/tensors sharing raw bytes but differing in shape/dtype (e.g. `(3,224,224)` vs `(1,3,224,224)`)
- `VideoData` differing only in sampling metadata (fps / duration / frames_indices / total_num_frames)
- a `VideoData` (no audio) vs a plain list of the same frames

## Fix
Make multimodal serialization canonical and self-describing — every value is encoded as `[type tag][typed metadata][length-prefixed payload]`:
- images encode mode + width + height
- tensors and ndarrays encode dtype + shape
- lists, tuples and dicts get distinct, length-framed encodings (dict keys sorted); ints are width/sign-unambiguous and `bool` is distinct from `int`
- `VideoData` carries a modality tag and includes its sampling metadata in the hash; `AudioData` binds samples to sample_rate
- the per-item hash is prefixed with a versioned scheme tag, and the two previously-divergent list-framing paths are unified

## Tests
Adds CPU-only regression tests in `tests/unittest/llmapi/test_llm_kv_cache_events.py` covering each collision class (transposed image dims, reshaped ndarray, reshaped tensor, video-metadata sensitivity, and VideoData-vs-frame-list), plus a positive determinism test (identical inputs still match). Updates one existing assertion that had encoded the VideoData-vs-list collision as expected behavior.

## Release Notes

* **Refactor**
  * Enhanced multimodal content caching with improved, collision-resistant hashing for better consistency across cache key evolution.

* **Tests**
  * Expanded test coverage for multimodal hashing to validate correct behavior across various data types and metadata scenarios.

### DIFF
--- a/tensorrt_llm/inputs/multimodal_data.py
+++ b/tensorrt_llm/inputs/multimodal_data.py
@@ -117,6 +117,13 @@ def serialize_item(obj: object) -> bytes:
             parts.append(serialize_item(obj[key]))
         return b"".join(parts)
 
+    if isinstance(obj, np.generic):
+        # numpy scalar (e.g. np.int64 / np.float32 / np.bool_): normalize to the
+        # equivalent Python scalar and recurse, so numpy-typed values hash
+        # identically to their Python counterparts. In numpy 2.x these are not
+        # subclasses of Python int/float/bool, so they bypass the checks above.
+        return serialize_item(obj.item())
+
     raise ValueError(f"Unsupported object type: {type(obj)}")
 
 
--- a/tensorrt_llm/inputs/multimodal_data.py
+++ b/tensorrt_llm/inputs/multimodal_data.py
@@ -50,7 +50,7 @@ def serialize_item(obj: object) -> bytes:
     """Serialize a supported multimodal hash leaf to bytes.
 
     The encoding is canonical and self-describing: every value is
-    ``[1-byte type tag][typed metadata][length-prefixed payload]`` with all
+    `[1-byte type tag][typed metadata][length-prefixed payload]` with all
     multi-byte integers big-endian. This prevents cache-key hash collisions
     between distinct values that happen to share a raw byte payload (for
     example transposed image dimensions or reshaped arrays).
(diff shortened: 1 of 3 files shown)


tests/unittest/llmapi/test_llm_kv_cache_events.py
```diff
@@ -320,6 +320,20 @@ def test_apply_mm_hashes_uuid_content_combined():
         "Different UUID + same content should produce different hashes"
 
 
+def _make_video(frames, audio_samples=None, sample_rate=16000, metadata=None):
+    """Build a VideoData from frames with optional audio and metadata.
+
+    Module-level helper shared by the mm-hash tests (lifted from the former
+    local closure in test_apply_mm_hashes_video_audio_affects_hash).
```
venkywonka:
```suggestion
    Module-level helper shared by the mm-hash tests
```

tensorrt_llm/inputs/multimodal_data.py
```diff
@@ -16,29 +26,103 @@ def update(self, data: bytes) -> None:
...
+    """Encode a byte payload prefixed with its u64 length."""
+    return _u64(len(payload)) + payload
+
+
 def serialize_item(obj: object) -> bytes:
-    """Serialize a supported multimodal hash leaf to bytes."""
+    """Serialize a supported multimodal hash leaf to bytes.
+
+    The encoding is canonical and self-describing: every value is
+    ``[1-byte type tag][typed metadata][length-prefixed payload]`` with all
```
2ez4bz: Nit: double-backticks -> single-backticks on changed lines.

tensorrt_llm/inputs/multimodal_data.py
```diff
@@ -16,29 +26,103 @@ def update(self, data: bytes) -> None:
...
+        return (
+            _u8(0x10)
+            + _len_prefixed(obj.mode.encode("utf-8"))
+            + _u32(width)
+            + _u32(height)
+            + _len_prefixed(payload)
+        )
     if isinstance(obj, torch.Tensor):
-        return obj.numpy().tobytes()
+        tensor = obj.detach().cpu().contiguous()
```
2ez4bz: 2 thoughts:

1. Do you think it would make sense to instead raise errors on non-contiguous inputs? That way, we can flag this undesirable property in whatever's producing these inputs (I assume we would want them to be contiguous regardless). This also means, if other places re calling `contiguous()`, we're not doing duplicated work.
2. Is there value in combining the `torch.Tensor` / `np.ndarray` paths by having this block just convert to numpy?

tensorrt_llm/inputs/multimodal_data.py
```diff
@@ -16,29 +26,103 @@ def update(self, data: bytes) -> None:
...
         return b"".join(parts)
+    if isinstance(obj, tuple):
+        parts = [_u8(0x21), _u64(len(obj))]
+        parts.extend(serialize_item(item) for item in obj)
+        return b"".join(parts)
+    if isinstance(obj, list):
+        parts = [_u8(0x20), _u64(len(obj))]
+        parts.extend(serialize_item(item) for item in obj)
+        return b"".join(parts)
+    if isinstance(obj, dict):
```
2ez4bz, line 102: Just checking: were there bugs related to us not handling dicts before, or is this a feature-add?

tensorrt_llm/inputs/multimodal_data.py
```diff
@@ -16,29 +26,103 @@ def update(self, data: bytes) -> None:
...
+        ]
+        parts.extend(_u64(dim) for dim in shape)
+        parts.append(_len_prefixed(payload))
         return b"".join(parts)
+    if isinstance(obj, tuple):
+        parts = [_u8(0x21), _u64(len(obj))]
+        parts.extend(serialize_item(item) for item in obj)
+        return b"".join(parts)
+    if isinstance(obj, list):
+        parts = [_u8(0x20), _u64(len(obj))]
```
2ez4bz, line 99: Just checking: are we intentionally differentiating between `tuple` and `list` here with `0x21` vs `0x20`? I would imagine that we don't care too much whether it's a tuple / list for hashing purposes, but maybe I'm missing some scenario where it matters.

tests/unittest/llmapi/test_llm_kv_cache_events.py
```diff
@@ -330,33 +344,183 @@ def test_apply_mm_hashes_video_audio_affects_hash():
...
-                                       mm_uuids)
+    hashes_uuid_a, _ = apply_mm_hashes(
+        {"video": [_make_video(frames, audio_a)]}, mm_uuids)
+    hashes_uuid_b, _ = apply_mm_hashes(
+        {"video": [_make_video(frames, audio_b)]}, mm_uuids)
     assert hashes_uuid_a["video"][0] != hashes_uuid_b["video"][0]
 
 
+def test_apply_mm_hashes_image_dims_distinguished():
+    """Nvbug 6226933: same-fill images of transposed dims must not collide."""
```
2ez4bz: Nit: I think these docstrings can be removed.

2ez4bz (approved):

moraxu (approved):

tensorrt_llm/inputs/multimodal_data.py
venkywonka: 1. i'm not sure what's the runtime scenarios when we face non-contiguous inputs here... so keeping it defensive. anyways if its contiguous, then `.contiguous()` is a no-op. if we raise, we might regress hence...
2. yep that makes sense, only metadata + content must be part of hash identity not the container.

tensorrt_llm/inputs/multimodal_data.py
venkywonka: yea that's over-defensive.

tensorrt_llm/inputs/multimodal_data.py
venkywonka: yep seems like we weren't capturing dicts and therefore metdata of the payload

PR #14811, opened 2026-06-01
[None][perf] Add WAN skip-softmax recipe config

karljang:
## Summary
- Add an 8-GPU BF16 WAN skip-softmax recipe config using TRTLLM attention, target sparsity 0.75, and 16 dense warmup steps.
- Wire `first_dense_steps` into the VisualGen skip-softmax path so early denoising steps can run dense before sparse attention is enabled.
- Keep the recipe on the GI path for WAN GELU: block-level torch.compile remains enabled and GELU stays in the compiled transformer block path. Self-attention uses the existing fused QK-Norm + RoPE path.
- Keep the recipe config aligned with the current ModelOpt flow: use a checkpoint-root `sparse.yaml` for calibration/disabled-layer data, with `sparse_config_path` remaining as an advanced fallback.

## Follow-up measurement
- Detailed LPIPS and end-to-end timing results for this recipe will be added after the final validation sweep/report is ready.

## Testing
- `pre-commit run --files <changed files>`
- `python3 -m py_compile <changed Python files>`
- `git diff --check origin/main...HEAD`

Focused pytest was not run in this non-Docker shell because `mpi4py` could not load `libmpi`.

### DIFF
--- a/examples/visual_gen/configs/wan2.2-t2v-bf16-skip-softmax-8gpu.yaml
+++ b/examples/visual_gen/configs/wan2.2-t2v-bf16-skip-softmax-8gpu.yaml
@@ -1,0 +1,36 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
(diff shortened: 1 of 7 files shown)


karljang:
Closing in favor of #14687

PR #14817, opened 2026-06-01
[None][perf] DSv4 MegaMoE: quantize activations directly into the DG SymmBuffer

Barry-Delaney:
MegaMoEDeepGemm.run_moe stages the per-token MXFP8 activations and scale factors into the DeepGEMM NVLink SymmBuffer with a buf.x.copy_(x) / buf.x_sf.copy_(x_sf) on every MoE layer — after quantize_input has already produced them in freshly-allocated tensors.

This PR adds an out-variant op trtllm::mxfp8_quantize_out that writes the quantizer result directly into caller-provided output tensors (same kernel as mxfp8_quantize, only the output buffer differs). quantize_input now quantizes straight into buf.x / buf.x_sf, and run_moe skips the staging copy when its input already aliases the SymmBuffer (detected via data_ptr), falling back to the previous quantize+copy path otherwise (e.g. on a binary that predates the op).

### DIFF
--- a/cpp/tensorrt_llm/thop/mxFp8Quantize.cpp
+++ b/cpp/tensorrt_llm/thop/mxFp8Quantize.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2023, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2020-2026, NVIDIA CORPORATION.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
```diff
@@ -361,6 +361,20 @@ def pad_up(x, m: int):
         scale_fp8_sf = input.new_empty((sf_size, ), dtype=torch.uint8)
         return val_mxfp8, scale_fp8_sf
 
+    # Register only when the op is in the loaded binary (the source tree may predate it).
+    if hasattr(torch.ops.trtllm, "mxfp8_quantize_out"):
+
+        @torch.library.register_fake("trtllm::mxfp8_quantize_out")
```
liji-nv, line 367: Not needed for pure inplace custom op. But better add to inplace_info in tensorrt_llm/_torch/compilation/utils.py (Although this op is only used inside MOE custom op)

cpp/tensorrt_llm/thop/mxFp8Quantize.cpp
```diff
@@ -111,9 +179,13 @@ TORCH_LIBRARY_FRAGMENT(trtllm, m)
...
         "-> (Tensor, Tensor)");
+    m.def(
+        "mxfp8_quantize_out(Tensor input, Tensor(a!) out_val, Tensor(b!) out_sf, "
+        "bool swizzedLayout=True, int alignment=32) -> ()");
 }
 
 TORCH_LIBRARY_IMPL(trtllm, CUDA, m)
 {
     m.impl("mxfp8_quantize", &tensorrt_llm::torch_ext::mxfp8_quantize);
+    m.impl("mxfp8_quantize_out", &tensorrt_llm::torch_ext::mxfp8_quantize_out);
```
lfr-0531, line 190: It'd be better to add tests for this new op `mxfp8_quantize_out`.

Barry-Delaney:
Will not merge after offline discussion, closing this PR.

PR #12829, opened 2026-04-08
[None][feat] Upgrade transformers dependency to 5.3.0

longlee0622:
# [None][feat] Upgrade transformers dependency to 5.3.0

## Summary

Bumps `transformers` from 4.x to **5.3.0** and adapts TensorRT-LLM to the wide set of API removals, renames, and behavioral changes in the v5 release. This is a large cross-cutting change (116 files, +1521 / -560). The branch is intentionally split into ~96 small, single-concern commits to make review tractable; the plan is to **squash on merge**, so individual commit messages do not need to be final.

Reviewers: this PR description is the canonical log of what changed and why. The per-commit history is a review aid — please use it to step through one concrete issue at a time rather than reading the full diff in one pass.

---

## What changed in transformers 5.3.0

The v5 release is a breaking release. The classes of changes we hit:

### 1. RoPE config refactor
- `config.rope_scaling` is now a **computed property** that delegates to a new `rope_parameters` dict.
- The dict uses `"rope_type"` instead of `"type"`.
- `"rope_type": "default"` now explicitly means "no scaling" — many models' custom `_init_rope` code doesn't recognise it.
- `rope_theta` can live inside `rope_parameters` only (not as a top-level attribute).
- Some models (Gemma3) moved to **per-layer-type** RoPE config.
- VL models keep RoPE fields on `text_config`, not the outer config.

### 2. Tokenizer API changes
- `batch_encode_plus` removed — `__call__` has the same signature.
- `vocab_size` property and `__len__` may raise `NotImplementedError` on some tokenizers (Qwen2Tokenizer).
- `apply_chat_template` default return type changed (needs `return_dict=False`).
- **LlamaTokenizer in v5 silently replaces `tokenizer.json`'s `ByteLevel` pre-tokenizer with `Metaspace`** when the repo declares `tokenizer_class: LlamaTokenizer` — this strips spaces from prompts and caused a catastrophic accuracy drop on DeepSeek-V3 (~25% vs ~63% on GSM8K) until fixed.

### 3. Fused MoE expert weights
- Mixtral, Qwen3Moe, Qwen3Next, Qwen3VL now ship experts as **fused tensors** rather than per-expert weights.
- `gate_up_proj` is a single tensor that must be split into `w1` / `w3` for TRT-LLM.
- FP8 scale tensors are also fused and must be unfused.
- Expert weight keys dropped the `.weight` suffix in some places.
- MoE gate `router_logits` is now a tuple in v5.
[...]

PR #14832, opened 2026-06-01
[None][fix] Install processor-output validation filter at module import

aswinvisva:
Avoids the validate_typed_dict re-patching race in bypass_processor_output_validation under concurrent trtllm-serve workers.

## Summary by CodeRabbit

* **Breaking Changes**
  * The `bypass_processor_output_validation()` function has been removed.

* **Refactoring**
  * Processor output validation for multimodal models now uses a global filter approach instead of context manager-based validation.

## PR Checklist

- [ x ] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_multimodal_utils.py
+++ b/tensorrt_llm/_torch/models/modeling_multimodal_utils.py
@@ -16,7 +16,6 @@
 # This file is based on official VILA: https://github.com/NVlabs/VILA/
 # and s2wrapper: https://github.com/bfshi/scaling_on_scales
 
-import contextlib
 import math
 import os
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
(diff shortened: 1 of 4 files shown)


tensorrt_llm/_torch/models/modeling_multimodal_utils.py
```diff
@@ -110,11 +113,9 @@ def _filtered_validate(schema, data):
...
     for b in binders:
         b.validate_typed_dict = _filtered_validate
-    try:
-        yield
-    finally:
-        for b, orig in originals.items():
-            b.validate_typed_dict = orig
+
+
+_install_processor_output_validation_filter()
```
2ez4bz: Do you think it would be preferable to have this be set to run at most once, and then called from our input processors' `__init__` methods for the qwen models?

2ez4bz (approved):

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
```diff
@@ -55,31 +54,36 @@ def _is_disagg() -> bool:
...
-    validate_typed_dict``, so patching only one is insufficient to cover
-    sub-processor validation paths. The set of binder modules differs
-    across transformers versions (5.3.x re-binds it on
-    ``image_processing_utils_fast``; 5.5.x dropped that module and re-binds
-    it on ``image_processing_utils`` instead), so we discover the binders
-    by ``hasattr`` rather than hard-coding the list. The originals are
-    restored on exit.
+    ``ProcessorMixin._merge_kwargs``. The keys in ``_PROCESSOR_OUTPUT_KEYS``
+    are processor *outputs* that leak into ``output_kwargs[<modality>]`` via
+    upstream bugs — e.g. ``Qwen2_5_VLProcessor._get_num_multimodal_tokens``
```
moraxu, line 73: Just for my understanding: this `Qwen2_5_VLProcessor._get_num_multimodal_tokens` is on HF side, right? (Can't find it in our codebase: https://github.com/search?q=repo%3ANVIDIA%2FTensorRT-LLM+Qwen2_5_VLProcessor&type=code)

If so, is there an open bug on HF side for this? I'm trying to understand why we have to do such filtering on our side.

moraxu (approved):

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
aswinvisva: Yes - `Qwen2_5_VLProcessor._get_num_multimodal_tokens` lives in the transformers lib. The mismatch is that the processor's returned dict carries keys that aren't declared in its own TypedDict schema, and HF's later validate_typed_dict call then rejects those extra keys. This PR is more of a quick fix to prevent the race condition issue from causing the server to crash after a few hundred requests

It was originally added in this PR: https://github.com/NVIDIA/TensorRT-LLM/pull/12829/changes#diff-b83961714a70ad6e90099d6fc6a41cd4e32fff8945f2b6045e91460fbbfd8332R59

tensorrt_llm/_torch/models/modeling_multimodal_utils.py
aswinvisva: That way would work too, I can switch it out if you prefer. The 'only get called at most once' property right now is enforced by the [per-module import lock](https://docs.python.org/3/whatsnew/3.3.html) in python so only one thread will execute this module's body and the other threads will read the initialized one from `sys.modules`. That said happy to switch it to your version to make it more explicit if you prefer

2ez4bz: I think since it's a qwen-model specific monkey-patch, we should try to minimize the blast radius if it's reasonably easy to do so.

aswinvisva: updated to be called from within the input processor's `__init__` and used a lru cache guard so that it would only get called once.

2ez4bz:
first try

PR #14841, opened 2026-06-02
[None][fix] Stabilize Mamba replay state update

sunnyqgg:
## Summary
- explicitly zero fresh Mamba prefill state slots in the mixer so recycled SSM/conv state is not reused
- mask replay old-state loads to accepted rows so dirty-tail cache rows are ignored

## Tests
- `pre-commit run --files tensorrt_llm/_torch/modules/mamba/replay_selective_state_update.py`: passed after applying reviewer comment wording
- earlier `pre-commit run --files` on the PR files: passed
- `python3 -m pytest -q tests/unittest/_torch/modules/mamba/test_mamba_ssm_rand_seed.py tests/unittest/_torch/executor/test_mamba_cache_manager.py`: blocked in this shell during collection because the local environment is missing TRT-LLM runtime dependencies (`mpi4py`; after installing missing `nvtx`)

Excluded from this PR per request: `tensorrt_llm/_torch/pyexecutor/resource_manager.py`, `examples/llm-api/quickstart_advanced.py`, `tensorrt_llm/_torch/modules/mamba/softplus.py`, `tests/unittest/_torch/modules/mamba/test_mamba_softplus.py`, `tensorrt_llm/_torch/pyexecutor/_util.py`, `tensorrt_llm/_torch/pyexecutor/mamba_cache_manager.py`, and `tests/unittest/_torch/executor/test_mamba_cache_manager.py`.

### DIFF
--- a/tensorrt_llm/_torch/modules/mamba/mamba2_mixer.py
+++ b/tensorrt_llm/_torch/modules/mamba/mamba2_mixer.py
@@ -348,12 +348,6 @@ class Mamba2Mixer(nn.Module):
             has_initial_states = mamba_metadata.has_initial_states[:
                                                                    num_prefills]
 
-            has_initial_states_p = has_initial_states[:num_prefills]
-            # Explicitly zero fresh context slots. Recycled slots may still
-            # carry a prior request's SSM/conv state, which kernels read.
-            fresh_p = state_indices_p[~has_initial_states_p]
-            conv_states[fresh_p] = 0
-            ssm_states[fresh_p] = 0
             # Fused kernel to avoid expensive .contiguous() call in causal_conv1d_fn.
             xbc_p_t = extract_transpose_xbc_prefill(zxbcdt, num_prefill_tokens,
                                                     self.tp_d_inner,
@@ -380,6 +374,7 @@ class Mamba2Mixer(nn.Module):
 
             initial_states = None
             if mamba_metadata.use_initial_states:
+                # Rows without cached prefix state start SSM from zero.
                 initial_states = torch.where(
                     has_initial_states[:, None, None, None],
                     ssm_states[state_indices_p], 0)
--- a/tensorrt_llm/_torch/modules/mamba/mamba2_mixer.py
+++ b/tensorrt_llm/_torch/modules/mamba/mamba2_mixer.py
@@ -349,7 +349,11 @@ class Mamba2Mixer(nn.Module):
                                                                    num_prefills]
 
             has_initial_states_p = has_initial_states[:num_prefills]
-            conv_states[state_indices_p[~has_initial_states_p]].zero_()
+            # Explicitly zero fresh context slots. Recycled slots may still
+            # carry a prior request's SSM/conv state, which kernels read.
+            fresh_p = state_indices_p[~has_initial_states_p]
+            conv_states[fresh_p] = 0
+            ssm_states[fresh_p] = 0
             # Fused kernel to avoid expensive .contiguous() call in causal_conv1d_fn.
             xbc_p_t = extract_transpose_xbc_prefill(zxbcdt, num_prefill_tokens,
                                                     self.tp_d_inner,
--- a/tensorrt_llm/_torch/modules/mamba/replay_selective_state_update.py
+++ b/tensorrt_llm/_torch/modules/mamba/replay_selective_state_update.py
@@ -226,7 +226,7 @@ def _replay_precompute_kernel(
         other=0.0,
     )
 
-    # Compute raw CB once — shared across all heads in this block
+    # Compute raw CB once, shared across all heads in this block.
     raw_CB = tl.dot(C_all.to(tl.bfloat16), tl.trans(B_all).to(tl.bfloat16))
 
     # Store B to cache (once per group, only if this block covers the first heads)
@@ -458,17 +458,18 @@ def _replay_state_update_kernel(
     # two back).  coeff is all-zero (offs_t < 0), total_decay is 1.0, so the
     # replay leaves `state` unchanged — cache contents don't matter on step 0.
     coeff = tl.exp(total_dA_cumsum - old_dA_cumsum_all) * old_dt_all
-    coeff = tl.where(offs_t < prev_num_accepted_tokens, coeff, 0.0)
+    accepted_mask = t_mask & (offs_t < prev_num_accepted_tokens)
+    coeff = tl.where(accepted_mask, coeff, 0.0)
 
-    # Load old_x: (BLOCK_SIZE_T, BLOCK_SIZE_M) — single-buffered
+    # Zero stale rows beyond PNAT to prevent Inf/NaN from reaching tl.dot.
     old_x_base = old_x_ptr + cache_batch_idx * stride_old_x_cache + pid_h * stride_old_x_head
     old_x_all = tl.load(
         old_x_base + offs_t[:, None] * stride_old_x_T + offs_m[None, :] * stride_old_x_dim,
-        mask=t_mask[:, None] & m_mask[None, :],
+        mask=accepted_mask[:, None] & m_mask[None, :],
         other=0.0,
     )
 
-    # Load old_B from READ buffer: (BLOCK_SIZE_T, BLOCK_SIZE_DSTATE)
+    # Apply the same accepted-row mask to old_B.
     old_B_base = (
         old_B_ptr
         + cache_batch_idx * stride_old_B_cache
@@ -477,7 +478,7 @@ def _replay_state_update_kernel(
     )
     old_B_all = tl.load(
         old_B_base + offs_t[:, None] * stride_old_B_T + offs_n[None, :] * stride_old_B_dstate,
-        mask=t_mask[:, None] & n_mask[None, :],
+        mask=accepted_mask[:, None] & n_mask[None, :],
         other=0.0,
     ).to(tl.float32)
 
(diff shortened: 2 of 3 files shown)


tensorrt_llm/_torch/modules/mamba/mamba2_mixer.py
```diff
@@ -349,7 +349,11 @@ def forward(
                                                                    num_prefills]
 
             has_initial_states_p = has_initial_states[:num_prefills]
-            conv_states[state_indices_p[~has_initial_states_p]].zero_()
+            # Explicitly zero fresh context slots. Recycled slots may still
+            # carry a prior request's SSM/conv state, which kernels read.
+            fresh_p = state_indices_p[~has_initial_states_p]
+            conv_states[fresh_p] = 0
```
achartier: Why switch from `zero_()` to scalar assignment?

tensorrt_llm/_torch/modules/mamba/replay_selective_state_update.py
```diff
@@ -458,17 +458,19 @@ def _replay_state_update_kernel(
     # two back).  coeff is all-zero (offs_t < 0), total_decay is 1.0, so the
     # replay leaves `state` unchanged — cache contents don't matter on step 0.
     coeff = tl.exp(total_dA_cumsum - old_dA_cumsum_all) * old_dt_all
-    coeff = tl.where(offs_t < prev_num_accepted_tokens, coeff, 0.0)
+    pnat_t_mask = t_mask & (offs_t < prev_num_accepted_tokens)
```
achartier: Nit: I assume `pnat` short for `prev_num_accepted_tokens`? Would `accepted_mask` be okay as name? Easier to read.

tensorrt_llm/_torch/modules/mamba/replay_selective_state_update.py
```diff
@@ -458,17 +458,19 @@ def _replay_state_update_kernel(
...
+    coeff = tl.where(pnat_t_mask, coeff, 0.0)
 
-    # Load old_x: (BLOCK_SIZE_T, BLOCK_SIZE_M) — single-buffered
+    # Load only accepted replay rows. Mask stale rows in [PNAT, T) to 0.0 so
+    # tl.dot cannot consume dirty-tail Inf/NaN values from old cache contents.
```
achartier: This could be a simpler comment:
```
# Zero stale rows beyond PNAT to prevent Inf/NaN from reaching tl.dot.
```

Wanli-Jiang (approved):

tensorrt_llm/_torch/modules/mamba/mamba2_mixer.py
sunnyqgg: Because `fresh_p` is a tensor index, this is advanced indexing. `conv_states[fresh_p].zero_()` can operate on the indexed temporary instead of reliably writing back to `conv_states`, while `conv_states[fresh_p] = 0` goes through `__setitem__` and updates the original tensor. I also reuse the same `fresh_p` for the new `ssm_states` initialization, so both state tensors get the same explicit writeback semantics.

tensorrt_llm/_torch/modules/mamba/replay_selective_state_update.py
sunnyqgg: Done, renamed it to `accepted_mask`.

tensorrt_llm/_torch/modules/mamba/replay_selective_state_update.py
sunnyqgg: Done, simplified the comment in the same update.

tensorrt_llm/_torch/modules/mamba/mamba2_mixer.py
achartier: Thanks for the explanation!

achartier (approved):

hnover-nv:
I'm confused by what this PR is fixing.    There are two changes, both it seems purely defensive.  One is masking off loads to # of accepted tokens vs all of T.  That will just stop NaNs from spreading via tl.dot, it won't stop them from getting produced.  Upcoming checkpointing replay does seomthign similar (https://github.com/NVIDIA/TensorRT-LLM/pull/14203).
Clearing out the SSM state, unless our prefill kernels are misconfigured or just wrong, we should not be reading that?  Seems in a similar boat to pro-actively clearly the old_* caches.
Do we know which change is observed to fix things?  I would be very surprised if the SSM clearing fixes things, and if it does it seems like we have bigger problems that need solving.  For the replay changes, I could imagine it stops the NaNs from spreading.  This could result in fixing 0/unk while hurting acceptance rate.  Because if the NaNs don't spread out of a draft position, then eventually the sampler might reject it (I'm not sure how NaN poisoning would work in the sampler), but the earlier draft positions are untouched.  Unless something causes NaNs to show up in only later positions, though, then 1 / (DL+1) of the time a Nan comes in, you'd expect it to hit the golden token and things would break.  That's assuming NaNs just land in certain positions randomly, but that's plausible as you'd think a NaN at in_proj or conv or earlier would just poison that token.

I am worried that this PR, and also the prior PR with cache clearing, are just making it harder to find the NaNs, but they are still present at a low rate in our production system.  IMO, our priority should be isolating what in the MoE or other non-ADP layers cause NaNs, not covering them up.

sunnyqgg:
> I'm confused by what this PR is fixing. There are two changes, both it seems purely defensive. One is masking off loads to # of accepted tokens vs all of T. That will just stop NaNs from spreading via tl.dot, it won't stop them from getting produced. Upcoming checkpointing replay does seomthign similar (#14203). Clearing out the SSM state, unless our prefill kernels are misconfigured or just wrong, we should not be reading that? Seems in a similar boat to pro-actively clearly the old_* caches. Do we know which change is observed to fix things? I would be very surprised if the SSM clearing fixes things, and if it does it seems like we have bigger problems that need solving. For the replay changes, I could imagine it stops the NaNs from spreading. This could result in fixing 0/unk while hurting acceptance rate. Because if the NaNs don't spread out of a draft position, then eventually the sampler might reject it (I'm not sure how NaN poisoning would work in the sampler), but the earlier draft positions are untouched. Unless something causes NaNs to show up in only later positions, though, then 1 / (DL+1) of the time a Nan comes in, you'd expect it to hit the golden token and things would break. That's assuming NaNs just land in certain positions randomly, but that's plausible as you'd think a NaN at in_proj or conv or earlier would just poison that token.
> 
> I am worried that this PR, and also the prior PR with cache clearing, are just making it harder to find the NaNs, but they are still present at a low rate in our production system. IMO, our priority should be isolating what in the MoE or other non-ADP layers cause NaNs, not covering them up.

As mentioned in the comment, `Zero stale rows beyond PNAT to prevent Inf/NaN from reaching tl.dot` it only prevents NaN propagation, and the reset operation makes the codes more agent friendly, I think those changes are meaningful and feel free to debug it continuously if you can still reproduce NaN without those modifications

hnover-nv:
Sorry, why does reset make it more agent friendly?

sunnyqgg:
> Sorry, why does reset make it more agent friendly?

During my debugging, the Agent consistently assumes a reset-to-zero is required here. In reality, resetting to zero at this spot has no impact on precision, as the value is already zeroed during kernel execution and resource preparation. However, I’ve noticed the Agent fails to recognize this quickly. Though these two lines of code look redundant, they help the Agent confirm the reset has been done and steer its checks toward more valuable points. Since they incur no performance overhead, I believe this code is worthwhile to keep.

hnover-nv:
> > Sorry, why does reset make it more agent friendly?
> 
> During my debugging, the Agent consistently assumes a reset-to-zero is required here. In reality, resetting to zero at this spot has no impact on precision, as the value is already zeroed during kernel execution and resource preparation. However, I’ve noticed the Agent fails to recognize this quickly. Though these two lines of code look redundant, they help the Agent confirm the reset has been done and steer its checks toward more valuable points. Since they incur no performance overhead, I believe this code is worthwhile to keep.

Hmm.  There is a tiny penalty here, clearing it twice?  And I feel like a comment could achieve the same goal.  but I guess it's ok?  Also, are we sure it's safe with disagg?

sunnyqgg:
> And I feel like a comment could achieve the same goal

And I feel like a comment could achieve the same goal ==>Good suggestion,
Also, are we sure it's safe with disagg?===》I didn't get why it will affect disagg, could you explain it more clearly?

hnover-nv:
> > And I feel like a comment could achieve the same goal
> 
> And I feel like a comment could achieve the same goal ==>Good suggestion,
> Also, are we sure it's safe with disagg?===》I didn't get why it will affect disagg, could you explain it more clearly?

You're clearing the state at certain times related to ptefill.  There's an implicit life cycle in agg that we do ptefill, and then decode.  We break that in disagg so it's possible there is some interaction.  Not something I thought through, just something to check.

But I've slept on it and I really think we should not do this.  I do worry they cover up real problems, and they should not be needed.  If agents get confused, well, people do to, and that's what comments are for. We shouldn't do redundant compute.  The old_* clearances are at least not redundant, but I'd still take them out.  IIf something is leaking around, we would rather amplify it so we can catch it.  And it does burn a few cycles.

I'm not the code owner, but I think we should focus on making sure we don't produce NaNs, or in general covering up bad interactions, which when were lucky express themselves as NaNs.  Thinking about that further, cuda graph padding and the ADP dummy slot can use uninitialized memory as it doesn't go through prefill, so we should clear those at init. I mean just don't allocate as torch.empty.  Also I'm a little worried about the shared slot interactions for padding.  I will look into adding fixes for those to the checkpointing PR, although if you want to change the allocation that'd be great.

The tl.dot stuff is fine with me, it's a performance win independent of any the non NaN spreading.
, and mutatis mutandis is already in ckpt PR.

PR #14860, opened 2026-06-02
[None][fix] adp router: gate in-transfer load accounting behind a config flag

lancelly:
## Summary

`KVCacheAwareADPRouter.create_rank_state` folds requests still mid-KV-transfer to the GEN worker (tracked by `AsyncTransferManager`, removed from `active_requests`) back into the per-rank `num_active_tokens` / `num_active_requests` reported via `RankState`. This was added to improve cross-rank load balance on V3.2.

@JinLi-NVIDIA spotted the side effect: the inflated `num_active_requests` propagates up to the PyExecutor inference loop, which uses `total_num_active_requests == 0 and len(waiting_queue) == 0` as the gate between a 1200s heartbeat timeout (block) and a 0s non-blocking poll. With in-transfer requests counted, the loop never sees the truly-idle state, so `fetch_new_requests` stops blocking; rank 0 then reaches the broadcast path with an empty batch and every other rank gets padded with dummy requests on each idle iteration.

On V4 the load-balance benefit from this accounting is negligible (verified by @lancelly), so make it opt-in.

### DIFF
--- a/tensorrt_llm/llmapi/llm_args.py
+++ b/tensorrt_llm/llmapi/llm_args.py
@@ -744,6 +744,19 @@ class AttentionDpConfig(StrictBaseModel):
         "scatter requests that would otherwise consolidate on a single warm "
         "rank, wasting prefill. Default False preserves pre-warmup routing. "
         "Only used when enable_kv_cache_aware_routing is True.")
+    kv_cache_routing_account_for_in_transfer: bool = Field(
+        default=False,
+        description=
+        "In-transfer load accounting in KV cache-aware routing. When True, "
+        "requests still streaming KV to the GEN worker (tracked by the "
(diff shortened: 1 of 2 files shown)


tensorrt_llm/llmapi/llm_args.py
```diff
@@ -744,6 +744,19 @@ class AttentionDpConfig(StrictBaseModel):
         "scatter requests that would otherwise consolidate on a single warm "
         "rank, wasting prefill. Default False preserves pre-warmup routing. "
         "Only used when enable_kv_cache_aware_routing is True.")
+    kv_cache_routing_account_for_in_transfer: bool = Field(
```
achartier, line 747: I think you'll need to update the YAML reference files if you add a new LLM arg.

achartier (approved):

PR #14868, opened 2026-06-02
[None][perf] disagg ctx: order prefill by uncached tokens, reusing the ADP router's per-rank prefix match

lancelly:
## What

Two pieces, on the **disaggregated context server**:

1. **Expose the routed-rank prefix match on the request** (`adp_router.py`).
   `KVCacheAwareADPRouter` already probes every new request's prefix-match length
   on every rank (`gather_prefix_matches`) and uses it to route. This stashes that
   value onto the request as `py_prefix_match_length` **at the routing decision** —
   using the match for the rank the request is *actually routed to* (`target_dp_rank`
   in the warmup path, `best_rank` in the scoring path). The match is **per-rank**, so
   a request must carry the match for the rank it lands on.

2. **Order context prefill by uncached tokens** (`py_executor.py`).
   In `_fetch_and_activate_new_requests`, sort newly-activated context requests by
   `uncached = prompt_len - py_prefix_match_length`, heaviest-prefill first, before
   extending `active_requests` (the V2 scheduler walks the list in order). No extra
   probe — it reuses the value from (1).

No-op for generation-only requests and for the non-KV-aware router (falls back to
full prompt length).

## Why

On a high-cache-hit disagg workload (DeepSeek-V4-Pro agentic-coding, ~96% prefix-cache
hit), the *uncached* token count — the work a prefill actually does — varies widely
per request. The hypothesis: letting the heaviest-prefill requests into the context
phase first improves time-to-first-token under chunked-prefill / token-budget pressure.

**Key implementation point (why this PR exists):** an earlier naive version probed the
radix tree a *second* time inside the activation hot path to get the uncached count.
On the context server — which is **host-bound** — that redundant probe measurably
slowed every prefill iter (CTX `host_step_time` mean +5–11%, p99 +13–27%) and **raised**
TTFT, swamping any ordering benefit. The router already computes this match for free,
so the fix is to **reuse the per-rank value** instead of recomputing it.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -3346,7 +3346,33 @@ class PyExecutor:
             if not _respond_if_invalid(request)
         ]
 
-        self.active_requests.extend(validated_requests)
+        # Order newly-activated context requests by uncached (non-prefix-cached)
+        # token count, heaviest-prefill first. The scheduler walks
+        # active_requests in list order, so this lets the requests that need the
+        # most prefill compute enter the context phase ahead of light,
+        # mostly-cached ones. The uncached estimate reuses
+        # ``py_prefix_match_length`` -- the prefix match the KV-cache-aware ADP
+        # router already computed for the rank this request was routed to -- so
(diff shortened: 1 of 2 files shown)


longlee0622:

PR #14880, opened 2026-06-03
[None][fix] Fix host-RAM OOM during MEGAMOE weight load on small-host nodes

qiaoxj07:
## Problem

DeepSeek-V4-Pro with the `MEGAMOE_DEEPGEMM` MoE backend OOMs the **host** during weight load on memory-constrained nodes (e.g. GB300: ~900 GiB host shared by 4 ranks, tp4/ep4). There are two independent host-RAM issues, in two different phases — each alone still OOMs (verified), so both fixes are needed:

1. **Model construction.** `MegaMoEDeepGemm.create_weights()` runs the DeepGEMM SymmBuffer rendezvous (`symm_mem.rendezvous` + `c10d.barrier`, inside `get_symm_buffer_for_mega_moe`). Under `MetaInitMode` the barrier touches a meta tensor → `MetaInitException` → meta-init is aborted for the whole model and it falls back to regular init, materializing ~221 GiB/rank on host → OOM (~3.5 min in).

2. **Weight load.** Reading the safetensors mmap faults file pages into the page cache faster than they are reclaimed; 4 ranks × ~205–246 GiB resident crosses the cgroup limit → OOM (~5.7 min in). The per-tensor `MADV_DONTNEED` already in `ConsumableWeightsDict.mark_consumed` isn't aggressive enough against the interleaved per-expert read pattern.

## Fixes

- **`mega_moe_deepgemm.py`** — defer the SymmBuffer rendezvous out of `create_weights` (which runs under `MetaInitMode`) into `post_load_weights` (same build-time, all-EP-rank lockstep window, after MetaInitMode has exited). Keeps meta-init alive so weights stream straight to GPU. Idempotent → the standalone `init_load_balancer` path is unaffected. Mirrors the deferral idiom used by the vision/sound encoders in `modeling_nemotron_nano.py`.
- **`modeling_deepseekv4.py`** — after each MoE layer in `DeepseekV4WeightLoader.load_weights`, `MADV_DONTNEED` the full set of `.safetensors` mmap regions so resident file cache can't accumulate. The mappings are read-only, so dropped pages are simply re-faulted on demand (data-safe).

## Validation (GB300, single node, 4 ranks, tp4/ep4, offline EPLB, ISL 8192 / OSL 1)

| | before | after |
|---|---|---|
| result | OOM | COMPLETED |
| MaxRSS / rank | 246 GiB | ~103 GiB |
| host-anon peak | climbs to 217 GiB | plateaus ~9.7 GiB |
| CTX throughput | — | ~64,200 tok/s (reproducible across runs) |

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_deepseekv4.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv4.py
@@ -819,7 +819,32 @@ class DeepseekV4WeightLoader:
                 getattr(module, attr).data.copy_(weights[key][:])
             return True
 
+        # On host-memory-constrained nodes (e.g. cmh GB300: 900 GiB shared by 4
+        # ranks), reading the 805 GiB safetensors during load faults the model
+        # into the file-cache faster than it is reclaimed; 4 ranks' resident
+        # file pages (~205-246 GiB/rank) cross the cgroup limit and OOM-kill the
+        # step. After each MoE layer we MADV_DONTNEED the whole set of
+        # safetensors mmap regions so the resident file-cache cannot accumulate
+        # (read-only mmap -> pages re-faulted on demand, data-safe). Always on:
+        # on large-host nodes it just re-reads a few pages (negligible), on
+        # small-host nodes it is what keeps the load under the cgroup limit.
+
+        def _pageout_safetensors():
+            import ctypes as _ct
+            torch.cuda.synchronize()
+            try:
+                _libc = _ct.CDLL("libc.so.6", use_errno=True)
+                for _ln in open("/proc/self/maps"):
+                    if ".safetensors" in _ln:
+                        _a, _b = _ln.split()[0].split("-")
+                        _libc.madvise(_ct.c_void_p(int(_a, 16)),
+                                      _ct.c_size_t(int(_b, 16) - int(_a, 16)), 4)  # MADV_DONTNEED
+            except Exception:
+                pass
+
         for name, module in tqdm(all_named_modules.items(), desc="Loading weights"):
+            if name.endswith("experts.backend"):
+                _pageout_safetensors()
             if name.startswith("draft_model"):
                 continue
             names = name.split(".")
--- a/tensorrt_llm/_torch/models/modeling_deepseekv4.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv4.py
@@ -793,8 +793,8 @@ class DeepseekV4WeightLoader:
         # ConsumableWeightsDict: deletes mmap-backed source tensors right
         # after each module finishes consuming them. Drops the OS page cache
         # pressure on safetensors files during the per-module loop, which is
-        # the dominant host RSS contributor on tight Grace LPDDR nodes (lyris
-        # GB300, 240 GiB) where 4-rank MEGAMOE mnt=16k otherwise blows past
+        # the dominant host RSS contributor on tight Grace LPDDR nodes (GB300,
+        # 240 GiB/rank) where 4-rank MEGAMOE mnt=16k otherwise blows past
         # the host budget. Mirrors DSv3's pattern at modeling_deepseekv3.py.
         can_mark_consumed = hasattr(weights, "mark_consumed")
 
@@ -819,7 +819,7 @@ class DeepseekV4WeightLoader:
                 getattr(module, attr).data.copy_(weights[key][:])
             return True
 
-        # On host-memory-constrained nodes (e.g. cmh GB300: 900 GiB shared by 4
+        # On host-memory-constrained nodes (e.g. GB300: 900 GiB shared by 4
         # ranks), reading the 805 GiB safetensors during load faults the model
         # into the file-cache faster than it is reclaimed; 4 ranks' resident
         # file pages (~205-246 GiB/rank) cross the cgroup limit and OOM-kill the
--- a/tensorrt_llm/_torch/models/modeling_deepseekv4.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv4.py
@@ -831,14 +831,16 @@ class DeepseekV4WeightLoader:
 
         def _pageout_safetensors():
             import ctypes as _ct
+
             torch.cuda.synchronize()
             try:
                 _libc = _ct.CDLL("libc.so.6", use_errno=True)
                 for _ln in open("/proc/self/maps"):
                     if ".safetensors" in _ln:
                         _a, _b = _ln.split()[0].split("-")
-                        _libc.madvise(_ct.c_void_p(int(_a, 16)),
-                                      _ct.c_size_t(int(_b, 16) - int(_a, 16)), 4)  # MADV_DONTNEED
+                        _libc.madvise(
+                            _ct.c_void_p(int(_a, 16)), _ct.c_size_t(int(_b, 16) - int(_a, 16)), 4
+                        )  # MADV_DONTNEED
             except Exception:
                 pass
 
--- a/tensorrt_llm/_torch/models/modeling_deepseekv4.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv4.py
@@ -830,19 +830,13 @@ class DeepseekV4WeightLoader:
         # small-host nodes it is what keeps the load under the cgroup limit.
 
         def _pageout_safetensors():
-            import ctypes as _ct
+            # Local import mirrors base_weight_loader's madvise call site and
+            # avoids importing the MoE load balancer at module import time.
+            from ..modules.fused_moe.moe_load_balancer import \
+                pageout_file_backed_regions
 
             torch.cuda.synchronize()
-            try:
-                _libc = _ct.CDLL("libc.so.6", use_errno=True)
-                for _ln in open("/proc/self/maps"):
-                    if ".safetensors" in _ln:
-                        _a, _b = _ln.split()[0].split("-")
-                        _libc.madvise(
-                            _ct.c_void_p(int(_a, 16)), _ct.c_size_t(int(_b, 16) - int(_a, 16)), 4
-                        )  # MADV_DONTNEED
-            except Exception:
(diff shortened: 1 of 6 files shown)


tensorrt_llm/_torch/modules/fused_moe/mega_moe/mega_moe_deepgemm.py
```diff
@@ -602,7 +619,20 @@ def create_weights(self):
...
+        # run under ``MetaInitMode``: a ``c10d.barrier`` on a meta tensor
+        # raises ``MetaInitException``, which aborts the whole model's
+        # meta-init and forces the slow regular-init fallback — every weight is
+        # then materialized on host RAM, OOM-killing small-host nodes (e.g. cmh
+        # GB300, ~900 GiB/node). When constructed lazily under MetaInitMode we
+        # defer the rendezvous to ``post_load_weights`` (same build-time
+        # lockstep window, after MetaInitMode has exited and before any forward
+        # / CUDA-graph capture). Same deferral idiom as the vision/sound
+        # encoders in ``modeling_nemotron_nano.py``.
+        if not _under_meta_init():
```
yuxianq: How about defer `self._alloc_symm_buffer()` in all cases so that we can remove this part?

tensorrt_llm/_torch/models/modeling_deepseekv4.py
```diff
@@ -819,7 +819,34 @@ def load_flat_hc_weights(module, names: List[str]) -> bool:
...
+        def _pageout_safetensors():
+            import ctypes as _ct
+
+            torch.cuda.synchronize()
+            try:
+                _libc = _ct.CDLL("libc.so.6", use_errno=True)
+                for _ln in open("/proc/self/maps"):
+                    if ".safetensors" in _ln:
+                        _a, _b = _ln.split()[0].split("-")
+                        _libc.madvise(
```
yuxianq: There are several places that wrap the `madvise`, can we extract them as a util function and reuse it? See https://github.com/NVIDIA/TensorRT-LLM/blob/5c1c1e2fdc6a441557d41e57411bd4d742fc8c83/tensorrt_llm/_torch/modules/fused_moe/moe_load_balancer.py#L92 and https://github.com/NVIDIA/TensorRT-LLM/blob/47d7eccdf5ea322ed1d05a11726176042de9e242/tensorrt_llm/runtime/kv_cache_manager_v2/_utils.py#L402

tensorrt_llm/_torch/modules/fused_moe/mega_moe/mega_moe_deepgemm.py
```diff
@@ -602,7 +619,20 @@ def create_weights(self):
...
+        # ``get_symm_buffer_for_mega_moe``) is a real collective and MUST NOT
+        # run under ``MetaInitMode``: a ``c10d.barrier`` on a meta tensor
+        # raises ``MetaInitException``, which aborts the whole model's
+        # meta-init and forces the slow regular-init fallback — every weight is
+        # then materialized on host RAM, OOM-killing small-host nodes (e.g. cmh
+        # GB300, ~900 GiB/node). When constructed lazily under MetaInitMode we
+        # defer the rendezvous to ``post_load_weights`` (same build-time
+        # lockstep window, after MetaInitMode has exited and before any forward
+        # / CUDA-graph capture). Same deferral idiom as the vision/sound
+        # encoders in ``modeling_nemotron_nano.py``.
```
yuxianq: Can you point me to the `modeling_nemotron_nano.py` code that use the same deferral idiom? Can it also defer in all cases?

Barry-Delaney (approved):
Thanks on the fix. LGTM after addressing Yuxian's comments.

tensorrt_llm/_torch/modules/fused_moe/mega_moe/mega_moe_deepgemm.py
qiaoxj07: Done in 6a76bd0781. `_alloc_symm_buffer()` is now invoked only from `post_load_weights`; the `create_weights` conditional and the `_under_meta_init` probe are removed. Verified `post_load_weights` always runs before forward on both paths: the loader iterates every module at `model_loader.py:491`, and the unit test calls it explicitly at `test_moe_backend.py:664`.

tensorrt_llm/_torch/modules/fused_moe/mega_moe/mega_moe_deepgemm.py
qiaoxj07: The idiom is `NemotronH_Nano_VL_V2.load_weights` in `modeling_nemotron_nano.py` (~line 2428): the vision/sound encoders are constructed there unconditionally, i.e. after `MetaInitMode` has exited, rather than in `__init__`. Yes — same unconditional deferral now applies here: as of 6a76bd0781 the SymmBuffer is allocated only in `post_load_weights`, no meta-init probe.

tensorrt_llm/_torch/models/modeling_deepseekv4.py
qiaoxj07: Extracted in 6a76bd0781: `madvise_range(addr, size, mode)` plus `pageout_file_backed_regions(path_substring, mode)` in `moe_load_balancer.py`. `advise_tensor_pageout` now delegates to `madvise_range`, and the DeepSeek-V4 page-out path calls `pageout_file_backed_regions(".safetensors")` (the inlined ctypes loop is gone). I left `kv_cache_manager_v2/_utils.py::_madvise` as-is: that package is mypyc-compiled and deliberately import-isolated (defines its own libc bindings, and *warns* rather than *raises* on failure), so cross-importing the `_torch` wrapper would break its isolation and change its failure semantics.

tensorrt_llm/_torch/modules/fused_moe/mega_moe/mega_moe_deepgemm.py
```diff
@@ -589,20 +590,31 @@ def _get_quant_method(self):
...
+        #    slow regular-init fallback that materializes every weight on host
+        #    RAM, OOM-killing small-host nodes (e.g. GB300, ~900 GiB/node).
+        #
+        # ``post_load_weights`` runs in the same build-time lockstep window on
+        # every EP rank, after MetaInitMode has exited and before any forward /
+        # CUDA-graph capture, so deferring there unconditionally keeps the
+        # rendezvous safety invariant intact while sidestepping both hazards.
+        # Same unconditional-deferral idiom as the vision/sound encoders in
+        # ``modeling_nemotron_nano.py`` (constructed in ``load_weights``,
+        # outside MetaInitMode; see ``NemotronH_Nano_VL_V2.load_weights``).
```
yuxianq: agents like to add comments for removed code, it is meaningless for future reader of main branch, just for code review. Can you avoid this pattern by combine this comment to the comment for `self._alloc_symm_buffer()` in `post_load_weights`?

yuxianq: The mention of `modeling_nemotron_nano.py` is also unnecessary, we don't need to know where does the inspiration come from.

tensorrt_llm/_torch/modules/fused_moe/moe_load_balancer.py
```diff
@@ -20,6 +20,77 @@
 from ...utils import EventType
 from ..multi_stream_utils import do_multi_stream
 
+_MADV_DONTNEED = 4
+_MADV_PAGEOUT = 21
+_MADV_ADVICE_BY_MODE = {"dontneed": _MADV_DONTNEED, "pageout": _MADV_PAGEOUT}
+
+
+def madvise_range(addr: int, size: int, mode: str = "dontneed") -> None:
```
yuxianq: Can we move `madvise_range/pageout_file_backed_regions/advise_tensor_pageout` to a common place instead of moe_load_balancer.py? Some candidate places include tensorrt_llm/_torch/memory_buffer_utils.py, tensorrt_llm/_torch/utils.py or create a new tensorrt_llm/_torch/mmap_utils.py

tensorrt_llm/_torch/modules/fused_moe/mega_moe/mega_moe_deepgemm.py
qiaoxj07: Let me remove them.

qiaoxj07: Done in 90300f4a4b. Dropped the deferral comment from `create_weights` (it now has only its functional lines) and moved the EPLB-sizing + MetaInitMode rationale into the `post_load_weights` comment, right next to `self._alloc_symm_buffer()`.

qiaoxj07: Removed the `modeling_nemotron_nano.py` reference as well (90300f4a4b).

tensorrt_llm/_torch/modules/fused_moe/moe_load_balancer.py
qiaoxj07: Done in de643ae8ff. Created `tensorrt_llm/_torch/mmap_utils.py` and moved `madvise_range`, `pageout_file_backed_regions` and `advise_tensor_pageout` there; repointed the call sites in `moe_load_balancer.py`, `quantization.py`, `base_weight_loader.py` and `modeling_deepseekv4.py`. Went with a new dedicated leaf module rather than `utils.py`/`memory_buffer_utils.py`: it keeps these host-mmap helpers cohesive and, since it only depends on `ctypes`/`mmap`, carries no circular-import risk. (`kv_cache_manager_v2/_utils.py::_madvise` is left as-is — that mypyc-isolated package uses madvise for a different purpose, MADV_HUGEPAGE THP hints, and warns rather than raises.)

yuxianq (approved):

PR #14895, opened 2026-06-03
[None][perf] executor: avoid deepcopy of prompt_token_ids on enqueue

lancelly:
## What
`_enqueue_request` deep-copied `request.prompt_token_ids` on every request. It's a flat list of token ids (immutable ints), so a shallow `list(...)` gives the same isolation far more cheaply — nothing downstream mutates the list in place (the prompt-adapter path rebuilds a new list; `tllm.Request` copies `input_token_ids` into its own C++ buffer). Also drops the now-unused `import copy`.

## Why
For a ~39k-token prompt, `copy.deepcopy` of the token-id list takes ~4.6 ms vs ~0.07 ms for `list(...)` (~70x), and it holds the GIL the entire time. In RPC mode this serialized work runs on an `rpc_server_worker` thread inside `RpcWorker.submit`, so the executor model loop was observed stalling on long "Waiting for the GIL" while a submit handler held the GIL doing the deepcopy.


tensorrt_llm/executor/base_worker.py
```diff
@@ -449,7 +448,7 @@ def _enqueue_request(self,
         else:
             lora_config = None
 
-        prompt_token_ids = copy.deepcopy(request.prompt_token_ids)
```
longlee0622, line 452: Why is the deepcopy needed in the first place? @hyukn

lancelly: I think it could be a typo. I discussed this with @yuxianq

JunyiXu-nv (approved):

PR #14897, opened 2026-06-03
[None][feat] improve iteration stats reporting

yizhang-nv:
Summary:
- Allow -1 as the unbounded sentinel for iter/request stats buffers.
- Preserve C++ semantics where 0 disables iter/request stats, while Python max_stats_len=0 keeps the legacy bounded-buffer trim behavior.
- Avoid overlapping iteration-stats drains in the OpenAI server metrics path.
- Track KV cache peak block stats by cache level and report primary/secondary peak free, used, and evictable blocks.
- Serialize the new peak free/evictable fields alongside the existing peak used fields, with C++ binding and serializer coverage.

Testing:
- python3 scripts/build_wheel.py --use_ccache --skip_building_wheel --linking_install_binary
- pytest -s tests/unittest/bindings/test_executor_bindings.py::test_executor_config
- pytest -s tests/unittest/executor/test_stats_serializer.py
- tests/unittest/llmapi/apps/test_disagg_perf_metrics_collector.py
- tests/unittest/llmapi/apps/test_openai_server_iteration_stats.py


tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache_manager.py
```diff
@@ -191,6 +191,13 @@ class _StorageLevelStats:
     allocated_bytes: int
 
 
+@dataclass(slots=True, frozen=True)
+class _PoolGroupBlockStats:
```
lowsfer: Why not just use StorageStatistics in _storage_manager.py? `used` is a bit confusing IMO, user may wonder if it includes evictable or not. That's why I avoided it and added property unavailable to make it clear.

yizhang-nv: I updated the internal peak stats fields from `free/used` to `available/unavailable/evictable` to match `StorageStatistics` semantics.

I still kept a separate `_PoolGroupPeakBlockStats` instead of reusing `StorageStatistics` because this object tracks interval peaks, not a single storage snapshot. `available`, `unavailable`, and `evictable` peaks are updated independently and may come from different points in time. With `StorageStatistics`, `unavailable` is derived from `total - available`, so it cannot represent an independently tracked peak unavailable value without making `total` misleading.

So the public metric names stay unchanged, but internally the semantics now line up with `StorageStatistics`.

tensorrt_llm/runtime/kv_cache_manager_v2/__init__.pyi
```diff
@@ -447,6 +447,7 @@ class KVCacheManager:
     def get_quota(self, cache_level: CacheLevel) -> int: ...
     def get_committed_stats(self) -> KVCacheStatsDelta: ...
     def get_and_reset_iteration_stats(self) -> dict[LifeCycleId, KVCacheIterationStatsDelta]: ...
+    def get_and_reset_iteration_peak_block_stats(self) -> Sequence[Any]: ...
```
lowsfer: Let's use strong type and avoid Any.

lowsfer (approved):

PR #14922, opened 2026-06-03
Fix PyExecutor FPM iteration timing

tedzhouhk:
## Summary

- Add batch-matched `gpuForwardTimeMS` for PyExecutor iteration stats using CUDA events around the batch `_forward_step`.
- Snapshot scheduler/FPM counters before forward mutates request state, so `inflightBatchingStats` stays matched to the exported iteration.
- Pool FPM-owned CUDA timing events for PP and iter-stats-only modes; when `PerfMetricsManager` provides pooled events, this code reuses those instead.
- Fix the iter-stats regression helper so explicit pre-forward `iter_states` snapshots are actually passed into `_update_iter_stats`.

## Validation

- `python -m py_compile tensorrt_llm/_torch/pyexecutor/py_executor.py tests/unittest/pyexecutor/test_iter_stats_populate.py tensorrt_llm/_torch/pyexecutor/adp_iter_stats.py tensorrt_llm/executor/base_worker.py`
- Downstream Dynamo FPM publisher test: `29 passed in 0.11s`
- Local A6000 load test with Dynamo + TRT-LLM PyTorch backend, `Qwen/Qwen3-0.6B`, FPM enabled:
  - `aiperf` concurrency=50, ISL=5000, OSL=500, requests=200
  - active FPM records after trimming idle margins: 2049
  - idle records inside trimmed window: 0
  - active `gpuForwardTimeMS` wall-time p50=40.8 ms, p95=71.7 ms, max=159.2 ms

FPM metrics vs execution time

TRT-LLM pytest was not runnable in my local checkout because the source tree lacks generated `tensorrt_llm.bindings` and the available TRT-LLM venv does not include `pytest`; the modified files were syntax-checked instead.

## Summary by CodeRabbit

* **New Features**
  * Added GPU forward pass timing metric to iteration statistics, providing granular performance insights alongside existing host and device timing measurements.

* **Bug Fixes**
  * Enhanced iteration-state snapshots to ensure accurate scheduler metrics and timing data capture across distributed execution.

* **Tests**
  * Updated test coverage to validate new timing metric propagation and scheduler state snapshot behavior.


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -150,6 +150,11 @@ class BatchState:
 
     iter_start_time: float = 0
     iter_stats: IterationStats = None
+    iter_states: Dict[str, Union[int, float]] = dataclasses.field(
```
hchings: I find the naming of `iter_states` can be easily confused w/ `iter_stats` at first glance, can we rename it to sth more visually different? 

Also, for `iter_states`, I'd prefer that we use a data class or named tuple, etc, instead of `Dict[str, Union[int, float]] = dataclasses.field(default_factory=dict)` to make it more self explainatory. You can follow the convention of other fields.

hchings:
> TRT-LLM pytest was not runnable in my local checkout because the source tree lacks generated tensorrt_llm.bindings and the available TRT-LLM venv does not include pytest; the modified files were syntax-checked instead.

Hi @tedzhouhk, thanks for the contribution! In the future, it'd still be better to run new tests locally first, as it'll take TRTLLM's CI quite a while to get to the new tests which might delay time-sensitive MRs. For tensorrt_llm.bindings, it'll need to compile the TRTLLM source following https://nvidia.github.io/TensorRT-LLM/installation/build-from-source.html#build-from-source.

tensorrt_llm/_torch/pyexecutor/py_executor.py
tedzhouhk: Done in 23c2f23c. I renamed the batch-carried payload to `scheduled_batch_stats` and replaced the dict with a `ScheduledBatchStats` dataclass. `iter_stats` remains the final `IterationStats` record, while `scheduled_batch_stats` is only the pre-forward scheduled-batch counters carried with `BatchState`.

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1276,6 +1294,88 @@ def _get_init_iter_stats(self, num_new_active_requests,
...
+        if start_event is not None and end_event is not None:
+            self._iter_timing_event_pool.append((start_event, end_event))
+
+    @staticmethod
+    def _compute_gpu_forward_time_ms(start_event, end_event) -> Optional[float]:
+        if start_event is None or end_event is None:
+            return None
+        try:
+            if not end_event.query():
+                end_event.synchronize()
```
achartier: This adds a GPU-host sync inside the inference loop, we should be careful this doesn't cause a perf regression.

tedzhouhk: Good catch. Now we measure FPM forward-pass time with CUDA events recorded around _forward_step. In both sync and async/overlap schedulers, FPM reads the elapsed time via a non-blocking event.query() path and does not add a new GPU-host synchronization point. The only synchronization involved is the existing sampler/update sync that the scheduler already needs before consuming results. When per-request perf metrics are enabled, FPM reuses the existing timing events; otherwise it borrows a lightweight forward-only event pair.

achartier (approved):

hchings (approved):

PR #14927, opened 2026-06-04
[None][perf] Optimize r128 compressor prefill reduction

mingyangHao:
@coderabbitai summary

 ## Summary

  Optimize DeepSeek-V4 compressor prefill for `compress_ratio=128`.

  The main change is inside `prefillReductionKernel`: while writing newly
  seen tokens into the paged compressor state cache, the kernel now also
  accumulates those same tokens into the online-softmax reduction. This
  avoids rereading the fresh-token portion from `kv_score + APE` during the
  reduction phase. The state cache is still fully maintained.

  For `compress_ratio=128`, the reduction loop is also split across 4
  reduction groups and merged in shared memory. `compress_ratio=4` stays on
  the original single-group path.

  ## Performance

  CUPTI average active duration, `head_dim=512`:

  | Case | Before | After | Change |
  |---|---:|---:|---:|
  | `r128 b1 s8k` | 58.195 us | 18.848 us | +67.6% |
  | `r128 b1 s64k` | 172.292 us | 97.177 us | +43.6% |
  | `r128 b8 s8k` | 172.881 us | 99.832 us | +42.2% |
  | `r128 b8 s64k` | 1971.040 us | 1182.502 us | +40.0% |

  `compress_ratio=4` non-regression:

  | Case | Before | After |
  |---|---:|---:|
  | `r4 b1 s8k` | 27.305 us | 27.280 us |
  | `r4 b1 s64k` | 184.273 us | 184.312 us |

## PR Checklist

## GitHub Bot Help


cpp/tensorrt_llm/kernels/compressorKernels/compressorKernels.cu
```diff
@@ -826,10 +824,19 @@ __global__ void prefillReductionKernel(void const* __restrict__ kv_score_raw, fl
...
     static_assert(VEC >= 4, "VEC must be >= 4 for float4 ape loads");
+    static_assert(NUM_RED_WARPS == 1 || NUM_RED_WARPS == 4, "Unsupported NUM_RED_WARPS");
 
-    int const tid = threadIdx.x;
+    constexpr int NTHRD_BASE = HEAD_DIM / VEC;
+    constexpr int HEAD_BLOCKS = (COMPRESS_RATIO == 128 && NTHRD_BASE > 32) ? (NTHRD_BASE / 32) : 1;
+    constexpr int NTHRD_INNER = NTHRD_BASE / HEAD_BLOCKS;
+    constexpr int ELEM_PER_BLOCK = NTHRD_INNER * VEC;
+
+    int const tid = (HEAD_BLOCKS == 1 && NUM_RED_WARPS == 1) ? threadIdx.x : (threadIdx.x % NTHRD_INNER);
```
heyuhhh: Can we remove this conditional expression? I think it's the same with or without it.

heyuhhh:
The changes roughly LGTM.
Just one question: can we use the optimization stratagy on `compress_ratio=4` situations? I think it's still efficitive for r8 even though the improvement is not so huge because of the small token slots.

mingyangHao:
> The changes roughly LGTM. Just one question: can we use the optimization stratagy on `compress_ratio=4` situations? I think it's still efficitive for r8 even though the improvement is not so huge because of the small token slots.

I dont think this gonna work for r=4. This pr break comp tokens into groups of 4, which means one warp handle 32 tokens. For r=4 scenario, it means one warp only handle one token. Another techinique used in this PR is accumlate new tokens when writing state. That also wont work for r=4 since there are two window, that will introduce some atomic add.

cpp/tensorrt_llm/kernels/compressorKernels/compressorKernels.cu
mingyangHao: Nice catch, I’ll simplify it to `threadIdx.x % NTHRD_INNER` for readability.

heyuhhh:
Thanks for the clarify! But i still have a question: how about only split head dims into 4 CTAs? We don't need break tokens if that's unnecessary.

peihu-nv (approved):

PR #14940, opened 2026-06-04
[None][fix] Update DeepGEMM to fix paged MQA metadata OOB

Barry-Delaney:
Update DeepGEMM pin to include fix for an out-of-bounds shared memory access in the paged MQA logits metadata scheduler kernel (smxx_paged_mqa_logits_metadata).

When all context_lens are zero (warmup decode with kv_lens=1 and indexer_compress_ratio=4), the binary search result reaches batch_size, causing prefix_sum[batch_size] to read past the shared memory array. This triggers a deterministic IMA during DSv4-Pro disagg GEN warmup at batch_size=32.

Fix: clamp the binary search result to batch_size - 1.

DeepGEMM PR: https://github.com/deepseek-ai/DeepGEMM/pull/353


Shixiaowei02 (approved):

jiaganc (approved):

peihu-nv (approved):
CI failures are unrelated DS-V4 B200 node issues. SBSA passed; x86 single hit the known DS B200 KV-transfer failure, and x86 multi failed on node SSH setup. Skipping CI and merging PR14940.

PR #14947, opened 2026-06-04
[None][opt] attn kernel epilogue fuse RopeQuant

yunruis:
## 1. Summary

This PR integrates DeepSeek-V4 FMHA epilogue fusion into the TensorRT-LLM PyTorch MLA path.

Before this change, the DeepSeek-V4 attention path materialized the FMHA latent output and then launched a separate inverse-RoPE + FP8 quantization kernel before the DSv4 O projection BGEMM:

```text
FMHA
  -> BF16 latent output
  -> standalone inverse-RoPE + E4M3 quant + FP32 scale
  -> DSv4 O projection BGEMM
```

With epilogue fusion enabled, trtllm-gen FMHA performs inverse RoPE and 1x128 E4M3 quantization in the FMHA correction epilogue, and directly writes the physical layout consumed by the DSv4 O projection BGEMM:

```text
FMHA correction epilogue
  -> inverse RoPE
  -> 1x128 E4M3 quant + FP32 scale
  -> DSv4 O projection BGEMM
```

The optimization removes the standalone rope-quant kernel launch and avoids the intermediate latent write/read between FMHA and quantization. The O projection itself is not fused into FMHA; it still runs as the existing DSv4 FP8 BGEMM, now consuming `(fp8_o, fp32_scale)` produced by FMHA.

## 2. Applicable Scenarios

The feature is intentionally gated to the validated DeepSeek-V4 Pro configuration:

- Model: DeepSeek-V4 Pro MLA / DSA path.
- Parallelism: attention DP / DP-only attention heads, with local `num_heads == num_heads_tp == 128`.
- GPU: SM100f path with matching trtllm-gen FMHA cubins.
- Precision/layout: FP8 E4M3 Q/K/V/O path, FP8 KV cache, DSv4 O projection A in E4M3.
- Attention mode: DeepSeek-V4 sparse MLA with paged KV.
- Batch shape: context-only or generation-only batches. Mixed context + generation batches fall back because the fused output buffers currently do not carry an explicit token-offset contract for two FMHA calls writing one combined output.
- Disabled cases: CP Helix, non-DSv4 models, non-FP8 KV cache, non-target DSv4 dimensions, missing sparse config, and user-disabled fusion.

The runtime kill switch is:

```bash
TRTLLM_DSV4_DISABLE_FMHA_EPILOGUE_FUSION=1
```

By default, the feature is allowed and the runtime gate decides whether the current batch can use the fused path.

## 3. Perf Gain

The kernel-level comparison should be reported as:

```text
non_fused_total_us = standard_fmha_us + standalone_rope_quant_us
ep_fusion_fmha_us = FMHA time with DSv4 epilogue fusion enabled
speedup = non_fused_total_us / ep_fusion_fmha_us
saved_pct = 1 - ep_fusion_fmha_us / non_fused_total_us
```

| sparse mode | phase | B | Q | KV | standard_us | rope_us | baseline total | fused_us | speedup | saved_pct |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| hca | prefill | 1 | 8192 | 8192 | 627.236 | 361.429 | 988.665 | 655.505 | 1.5082x | 33.70% |
| hca | prefill | 4 | 8192 | 8192 | 2512.980 | 1488.042 | 4001.022 | 2540.840 | 1.5747x | 36.50% |
| hca | prefill | 1 | 16384 | 16384 | 1216.700 | 736.190 | 1952.890 | 1278.080 | 1.5280x | 34.55% |
| hca | prefill | 4 | 16384 | 16384 | 5044.020 | 3107.636 | 8151.656 | 5038.630 | 1.6178x | 38.19% |
| csa | prefill | 1 | 8192 | 8192 | 1230.050 | 361.429 | 1591.479 | 1160.070 | 1.3719x | 27.11% |
| csa | prefill | 4 | 8192 | 8192 | 5250.850 | 1488.042 | 6738.892 | 4925.850 | 1.3681x | 26.90% |
| csa | prefill | 1 | 16384 | 16384 | 2815.160 | 736.190 | 3551.350 | 2642.630 | 1.3439x | 25.59% |
| csa | prefill | 4 | 16384 | 16384 | 11646.100 | 3107.636 | 14753.736 | 10855.700 | 1.3591x | 26.42% |
| hca | generation | 32 | 1 | 8192 | 9.340 | 2.690 | 12.030 | 10.528 | 1.1426x | 12.48% |
| hca | generation | 32 | 4 | 8192 | 16.696 | 6.532 | 23.228 | 17.288 | 1.3436x | 25.57% |
| csa | generation | 32 | 1 | 8192 | 19.683 | 2.690 | 22.373 | 17.142 | 1.3051x | 23.38% |
| csa | generation | 32 | 4 | 8192 | 30.577 | 6.532 | 37.109 | 30.863 | 1.2024x | 16.83% |

### 4.1 New DSv4 Epilogue Fusion Inputs and Outputs

New runtime inputs/outputs are added for the fused path:

| Field | Direction | Meaning |
|---|---|---|
| `output` / `oPtr` | FMHA output | E4M3 physical O buffer consumed by DSv4 O projection BGEMM |
| `output_sf` / `oSfPtr` | FMHA output | FP32 1x128 scale buffer for the E4M3 O buffer |
| `dsv4_inv_rope_cos_sin_cache` | FMHA input | FP32 inverse-RoPE cos/sin cache |
| `enable_dsv4_epilogue_fusion` | runtime flag | Enables DSv4 inverse-RoPE + FP8 quant in the FMHA epilogue |
| `scaleBufM` | runtime scalar input | Padded token stride derived from `output_sf.size(2)` |

`scaleBufM` is an input scalar, not an output tensor. It describes the padded token stride of `output_sf` and is propagated to trtllm-gen so the kernel writes FP32 scales with the same layout allocated by TensorRT-LLM.

The fused output layout is:

```text
fp8_o:
  dtype = E4M3
  shape = [num_groups, num_tokens, heads_per_group * v_head_dim]

output_sf:
  dtype = FP32
  shape = [num_groups, heads_per_group * (v_head_dim / 128), scaleBufM]
```

### 4.2 Python Model Forward Changes

The DeepSeek-V4 MLA forward path now has two possible attention outputs:

```text
non-fused path:
  attn_output = BF16 latent
  _deepseek_v4_o_proj(attn_output, position_ids)
    -> standalone inverse-RoPE + FP8 quant
    -> DSv4 O projection BGEMM

fused path:
  dsv4_epilogue_output = (fp8_o, fp32_scale)
  _deepseek_v4_o_proj(dsv4_epilogue_output)
    -> DSv4 O projection BGEMM
```

The O projection call site remains unified in `MLA.forward()`. The fused path changes the input type to `_deepseek_v4_o_proj()` but does not move the O projection into FMHA.

For `register_to_config`, the MLA custom op contract is extended so custom-op execution can receive DSv4 fused buffers explicitly:

- `trtllm::create_mla_outputs` creates the legacy output and, when the runtime gate passes, the DSv4 `(fp8_o, output_sf)` buffers.
- `trtllm::mla_custom_op_inplace` accepts optional `dsv4_output`, `dsv4_output_sf`, and `enable_dsv4_epilogue_fusion`.
- DeepSeek-V4 custom-op execution calls `forward_impl_with_deepseek_v4(..., dsv4_epilogue_output=...)`.

This keeps the custom-op path functionally aligned with the eager Python path and avoids dropping the fused epilogue buffers before the O projection.


tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1887,8 +1891,81 @@ def _get_attn_scale(position_ids: torch.Tensor) -> torch.Tensor:
         q = (q * attn_scale).to(q.dtype)
         return q
 
-    def _deepseek_v4_o_proj(self, attn_out_latent: torch.Tensor,
-                            position_ids: torch.Tensor) -> torch.Tensor:
+    def _should_use_dsv4_epilogue_fusion(self, num_contexts: int,
```
mingyangHao, line 1945: I doubt if we really need that much conditions, i think your fusion is a global win right?

yunruis:
@mingyangHao @heyuhhh  Could you please help review this PR

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -2455,6 +2601,8 @@ def _indexer_branch():
             0] == num_tokens, f"Expect q.shape[0] to be {num_tokens}, but got {q.shape[0]}"
 
         assert output is not None, "output must be provided"
+        enable_dsv4_epilogue_fusion = self._should_use_dsv4_epilogue_fusion(
```
mingyangHao: `_should_use_dsv4_epilogue_fusion` has been called previously at create_mla_outputs_impl, lets unify this flag by passing it instead of separate calling.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -100,6 +104,30 @@ def _(q, attention_mask, layer_idx):
     return create_attn_outputs_impl(q, attention_mask, layer_idx)
 
 
+def create_mla_outputs_impl(hidden_states: torch.Tensor,
+                            layer_idx: str) -> List[torch.Tensor]:
+    metadata, mla_layer = extract_extra_attrs(layer_idx, "mla")
+    attn_output = mla_layer.create_output(hidden_states, metadata.num_contexts)
```
mingyangHao: The fused path still allocates the unused BF16 latent output (create_mla_outputs_impl, forward)
  attn_output = create_output(...) allocates [num_tokens, 128 × 512] BF16 (128 KB/token — over 1 GiB for an 8k×… prefill batch) that is never written or read when
  fusion is active; the FMHA writes the FP8 buffers and o_proj consumes those. That defeats part of the stated benefit ("avoids the intermediate latent write/read") on
  the memory side and inflates peak memory. Consider allocating a 0-size placeholder when fusion is on.

cpp/tensorrt_llm/kernels/trtllmGenKernels/fmha/trtllmGen_fmha_export/trtllm/dev/Fp8Utils.h
```diff
@@ -135,5 +137,142 @@ inline __device__ void convertFp16ToMxE4m3(OutT& out,
...
+                                                   int32_t headDimOffset) {
+  static_assert(NumVals == NumPackedRegs * 4, "One packed register stores four E4M3 values.");
+  static_assert(NumVals == 128, "DSv4 fused epilogue processes one 1x128 quant group.");
+
+  // DSv4 inverse-RoPE applies to the last 64 dimensions, [448, 512), of a 512-dim head. The
+  // helper processes one 128-value quant block, so the RoPE block starts at 384 and the RoPE
+  // values begin at offset 64 inside that block. DSv4 uses non-NeoX interleaved RoPE: adjacent
+  // element pairs share one cos/sin value. Each cos/sin cache row is laid out as
+  // [cos(32), sin(32)].
+  int32_t constexpr ropeStart = 448;
```
mingyangHao, line 228: These constants are duplicated with the ones you defined in `Dsv4Constants.h`, please reuse them.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1887,8 +1933,98 @@ def _get_attn_scale(position_ids: torch.Tensor) -> torch.Tensor:
         q = (q * attn_scale).to(q.dtype)
         return q
 
-    def _deepseek_v4_o_proj(self, attn_out_latent: torch.Tensor,
-                            position_ids: torch.Tensor) -> torch.Tensor:
+    def _should_use_dsv4_epilogue_fusion(self, num_contexts: int,
+                                         num_generations: int) -> bool:
+        if _is_env_truthy("TRTLLM_DSV4_DISABLE_FMHA_EPILOGUE_FUSION"):
```
mingyangHao: `_is_env_truthy` is consulted in `_should_use_dsv4_epilogue_fusion`, which runs twice per layer per forward. Cheap, but this is the host-latency-sensitive decode path; cache it at import/init.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1887,8 +1933,98 @@ def _get_attn_scale(position_ids: torch.Tensor) -> torch.Tensor:
...
+        self,
+        attn_out_latent: Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]],
+        position_ids: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        if isinstance(attn_out_latent, (tuple, list)):
+            attn_fp8, attn_scale = attn_out_latent
+            num_tokens = attn_fp8.shape[1]
+            scale_buf_m = (num_tokens + 3) // 4 * 4
+            if attn_scale.shape[2] != scale_buf_m:
+                attn_scale = attn_scale[:, :, :scale_buf_m]
```
mingyangHao: _deepseek_v4_o_proj's attn_scale[:, :, :scale_buf_m] slice can never trigger given _get_dsv4_epilogue_buffers enforces fp8_o.shape[1] ==
  num_tokens and buffers are always allocated with exactly that padding — and if it did trigger, the sliced view is non-contiguous (stride = original padded M), which
  is only correct if cute_dsl_fp8_bmm_blackwell honors strides. Either delete it or make the check exact. Also a cosmetic leftover in KernelTraits.h: the "data type of
  softmax computation" comment now dangles above nothing.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -2476,7 +2624,11 @@ def _indexer_branch():
                 position_ids=position_ids[:num_ctx_tokens],
                 latent_cache=latent_cache_ctx,
                 topk_indices=topk_indices_ctx,
+                enable_dsv4_epilogue_fusion=enable_dsv4_epilogue_fusion,
+                dsv4_epilogue_output=dsv4_epilogue_output,
             )
+            if enable_dsv4_epilogue_fusion:
```
mingyangHao: forward_impl_with_deepseek_v4: the mid-function return ctx_result / return gen_result pattern is fragile

  This works today, but only because of an invariant that lives far away: _should_use_dsv4_epilogue_fusion returns False for mixed batches, so when fusion is enabled at
  most one of the two if blocks runs and the early return can't skip anything. Three concerns:

  1. Dual contract. With fusion off, this function is in-place (writes output, returns None); with fusion on, it returns the fused buffers and output is never written.
  The return-value semantics flip based on a flag that is re-derived internally, and the caller has to infer which path was taken via is not None.
  2. Silent failure mode. If the gating ever changes to allow mixed batches (or the two _should_use_dsv4_epilogue_fusion evaluations disagree), the early return 
  ctx_result would silently skip the entire generation branch — no error, just missing attention results for part of the batch. The only protection is in another
  function ~60 lines away.
  3. Dead variables on the non-fused path. ctx_result/gen_result are assigned but unused when fusion is off (these were bare calls before).

  Suggestion: assert the invariant locally and converge to a single return point, e.g.

  fused_result = None
  if enable_dsv4_epilogue_fusion:
      assert (num_contexts > 0) != (num_generations > 0), \
          "DSv4 epilogue fusion requires a context-only or generation-only batch"

  if num_contexts > 0:
      ...
      out = self.forward_context_sparse_mla(...)
      if enable_dsv4_epilogue_fusion:
          fused_result = out

  if num_generations > 0:
      ...
      out = self.forward_generation_sparse_mla(...)
      if enable_dsv4_epilogue_fusion:
          fused_result = out

  return fused_result

  Straight-line control flow, a broken invariant fails loudly instead of silently dropping results, and the dead variables go away. Ideally also pass
  enable_dsv4_epilogue_fusion in as a parameter instead of re-deriving it here, so the buffer-providing caller and this function can't disagree.

tensorrt_llm/_torch/modules/attention.py
yunruis: We need to ensure it is dsv4 pro attention DP, fp8, and only ctx or only gen (for memory layout contrains).

cpp/tensorrt_llm/common/attentionOp.cpp
```diff
@@ -2932,6 +2944,10 @@ int AttentionOp::initialize() noexcept
             fmhaParams.dataTypeKv = DATA_TYPE_E4M3;
             fmhaParams.dataTypeOut = DATA_TYPE_BF16;
         }
+        if (mFusesDsv4InvRopeFp8Quant)
+        {
+            fmhaParams.dataTypeOut = DATA_TYPE_E4M3;
```
heyuhhh, line 2949: Should we check it but not set it? IIUC, `mFusesDsv4InvRopeFp8Quant` enable means fp8 output, so we should assert this equal but not transfer this.

tensorrt_llm/_torch/modules/attention.py
yunruis: `create_mla_outputs_impl` is a no conditional function, it is not designed to save `enable_dsv4_epilogue_fusion ` flag. Besides, `_should_use_dsv4_epilogue_fusion` it is a very lightly function.

cpp/tensorrt_llm/common/attentionOp.h
```diff
@@ -519,6 +519,8 @@ class AttentionOp
 
     // Whether to fuse FP4 quant into attention kernel.
     bool mFuseFp4Quant = false;
+    // Whether to fuse DSv4 inverse-RoPE + FP8 output quant into trtllm-gen FMHA.
+    bool mFusesDsv4InvRopeFp8Quant = false;
```
heyuhhh, line 523: Notice the place of this flag, should align with `data()`

cpp/tensorrt_llm/kernels/trtllmGenKernels/fmha/trtllmGen_fmha_export/trtllm/dev/Fp8Utils.h
yunruis: this is export from trtllm-gen, and from SW design side `trtllm/dev/Fp8Utils.h` should not include  and rely on file in `kernels/fmha`. Thus, this is a trade-off design

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -1887,8 +1933,98 @@ def _get_attn_scale(position_ids: torch.Tensor) -> torch.Tensor:
...
-                            position_ids: torch.Tensor) -> torch.Tensor:
+    def _should_use_dsv4_epilogue_fusion(self, num_contexts: int,
+                                         num_generations: int) -> bool:
+        if _is_env_truthy("TRTLLM_DSV4_DISABLE_FMHA_EPILOGUE_FUSION"):
+            return False
+        if not self.is_deepseek_v4:
+            return False
+        if num_contexts == 0 and num_generations == 0:
+            return False
+        if num_contexts > 0 and num_generations > 0:
```
heyuhhh, line 1953: Why this condition? We can explicit split the output tensor to seperate context/generation forward

cpp/tensorrt_llm/common/attentionOp.cpp
yunruis: This is the place where set `datatypeOut, datatypeKv`.

tensorrt_llm/_torch/modules/attention.py
yunruis: As comment

```
# The fused output buffers are produced by a single context or
            # generation FMHA call. Mixed batches would need an explicit token
            # offset contract so the two calls do not overwrite each other.
```

heyuhhh: I mean we don't need this condition. We can split output buffer for context and generation

yunruis: I understand your point, but later output buffer consumer `bgemm` will need a contiguous buffer. That will introduce `concat` memory operation.  For potential later optimization, will need a extra offset params to trtllm-gen fmha kernels.

tensorrt_llm/_torch/modules/attention.py
yunruis: Done. cancel allocate BF16 output when epilogue fusion enabled.

tensorrt_llm/_torch/modules/attention.py
yunruis: Done. Drop the unnecessary slice, and set it to error when `scale_buf_m` do not  match `shape[0]`

tensorrt_llm/_torch/modules/attention.py
yunruis: Done.

cpp/tensorrt_llm/common/attentionOp.h
yunruis: Done, aligned with `data()`

yunruis:
all comments has been resolved from my side. If you agree please mark it as resolved. Thanks  @heyuhhh  @mingyangHao

tensorrt_llm/_torch/modules/attention.py
mingyangHao: I understand your point, but to me it looks like there are many duplicated conditions such as 
`if not self.is_deepseek_v4:`
and 
```
        if self.kv_lora_rank != 448 or self.qk_rope_head_dim != 64:
            return False
        if self.qk_head_dim != 512 or self.v_head_dim != 512:
            return False
```
and
```
        self._disable_dsv4_epilogue_fusion = (
            self.is_deepseek_v4
            and _is_env_truthy("TRTLLM_DSV4_DISABLE_FMHA_EPILOGUE_FUSION"))
```
since those are the only config for deepseekv4 models.

yunruis:
All comments has been resolved.

lfr-0531 (approved):

PR #14954, opened 2026-06-04
[https://nvbugs/6160629][fix] AutoDeploy: Fix manual seed setting for standalone tests

galagam:
## Description

BF16 with only 7 mantissa bits is more numerically sensitive than FP16. Test is passing for some seeds, failing for others with small mismatches for large values. The original atol/rtol = 1e-4/1e-4 threshold is too strict. 
This test was manually seeded with a fixed seed at import, which produced deterministic results on normal runs. However, within the standalone package, it runs with pytest-xdist (dynamic test distribution), rendering the seed at the module ineffective. 
This PR sets the same manual seed in each of the existing test. 
This is not the correct solution from a numeric POV, since we're relying on some lucky seed where the accuracy can pass instead of correctly scoping and guarding the accuracy for any given input.
However, this practice is widely used in the test suite, and this PR aims to fix the standalone verification and not improve accuracy testing overall.

Examples:
  seed=23
  q_hf[(0, 3, 0, 129)]      = -0.0810546875
  q_flash/custom_q[...]     = -0.0805664062
  abs diff                  = 0.0004882812
  relative diff             ~= 0.0060

  seed=41
  k_hf[(1, 2, 0, 189)]      = 0.7148437500
  k_flash/custom_k[...]     = 0.7187500000
  abs diff                  = 0.0039062500
  relative diff             ~= 0.0054

  seed=95
  q_hf[(1, 3, 0, 227)]      = 1.1640625000
  q_flash/custom_q[...]     = 1.1718750000
  abs diff                  = 0.0078125000
  relative diff             ~= 0.0067

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tests/unittest/auto_deploy/singlegpu/custom_ops/rope/test_rope_op_variants.py
+++ b/tests/unittest/auto_deploy/singlegpu/custom_ops/rope/test_rope_op_variants.py
@@ -30,7 +30,8 @@ torch.manual_seed(1234)
 @pytest.mark.parametrize(
     "dtype,atol,rtol",
     [
-        (torch.bfloat16, 1e-4, 1e-4),
+        # FlashInfer and the HF reference can differ by one bfloat16 bin (ULP).
+        (torch.bfloat16, 1e-4, 1e-2),
         (torch.float16, 5e-4, 5e-4),
     ],
     ids=["bfloat16", "float16"],  # q/k must be in half precision
--- a/tests/unittest/auto_deploy/singlegpu/custom_ops/rope/test_rope_op_variants.py
+++ b/tests/unittest/auto_deploy/singlegpu/custom_ops/rope/test_rope_op_variants.py
@@ -23,15 +23,19 @@ from _model_test_utils import (
 
 import tensorrt_llm._torch.auto_deploy  # noqa: F401
 
-torch.manual_seed(1234)
+# NOTE: the existing tests below seed the RNG explicitly (torch.manual_seed) at
+# their start. They use tight (near-ULP) half-precision tolerances that only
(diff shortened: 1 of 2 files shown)


bmarimuthu-nv (approved):
Thanks @galagam !

tests/unittest/auto_deploy/singlegpu/custom_ops/rope/test_rope_op_variants.py
```diff
@@ -30,7 +30,8 @@
 @pytest.mark.parametrize(
     "dtype,atol,rtol",
     [
-        (torch.bfloat16, 1e-4, 1e-4),
+        # FlashInfer and the HF reference can differ by one bfloat16 bin (ULP).
+        (torch.bfloat16, 1e-4, 1e-2),
```
nvchenghaoz: what about having a fixed seed instead of changing the threshold?

galagam: @nvchenghaoz Fixed seeds can hide seed-dependent issues, and are generally not robust enough. Adjusting tolerances better reflects bf16 inherent numeric limitations.
If we want to be extra vigilant, we can allow only a small percentage of elements to be above the strict threshold, and require the rest 1-p elements to be below the lenient threshold. But I think it's an overkill here.

galagam: Ended up using a manual seed, see updated description. I still don't think it's the correct approach, but it's the smallest-footprint fix for the standalone failing.

bmarimuthu-nv (approved):

PR #14956, opened 2026-06-04
[None][refactor] split VisualGen pipeline and model configs

bobboli:
## Summary
- split VisualGen pipeline-level config from per-model DiffusionModelConfig
- add BaseDiffusionModel and pass component-specific configs to transformer modules
- route multi-component pipelines through model_configs for transformer/transformer_2

## Testing
- python -m py_compile on changed Python files
- git diff --check
- pre-commit run --files on changed files

## Release Notes

* **Refactor**
  * Refactored visual generation pipeline configuration system to support multi-component pipelines through new unified configuration and model base classes, improving pipeline management across supported architectures.

### DIFF
--- a/tensorrt_llm/_torch/visual_gen/config.py
+++ b/tensorrt_llm/_torch/visual_gen/config.py
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Internal DiffusionModelConfig and loading helpers."""
+"""Internal VisualGen pipeline and model configuration helpers."""
 
 import json
 from pathlib import Path
@@ -85,24 +85,18 @@ def create_attention_metadata_state() -> Dict[str, Any]:
     return {"metadata": None, "capacity": (0, 0)}
 
 
-# =============================================================================
-# DiffusionModelConfig - Internal configuration (merged/parsed)
-# =============================================================================
+class _VisualGenConfigBase(BaseModel):
+    """Base for internal VisualGen configs that carry runtime objects."""
 
-
-class DiffusionModelConfig(BaseModel):
-    """Internal ModelConfig for diffusion models.
-
-    This is created by PipelineLoader from VisualGenArgs + checkpoint.
-    Contains merged/parsed config from:
-    - pretrained_config: From checkpoint/config.json
-    - quant_config: From checkpoint or user quant config
-    - Sub-configs: From VisualGenArgs (pipeline, attention, teacache)
-    - visual_gen_mapping: Populated by setup_visual_gen_mapping() from ParallelConfig
-    """
-
+    # Pydantic reserves `model_config` for class-level settings. This is not
+    # a VisualGen model config; it lets fields hold objects such as Mapping.
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
+
+class DiffusionModelConfig(_VisualGenConfigBase):
+    """Internal config for one TRT-LLM VisualGen model component."""
+
+    component_name: Optional[str] = None
     pretrained_config: Optional[Any] = None
     mapping: Mapping = PydanticField(default_factory=Mapping)
     skip_create_weights_in_init: bool = False
@@ -110,12 +104,12 @@ class DiffusionModelConfig(BaseModel):
     allreduce_strategy: AllReduceStrategy = PydanticField(default=AllReduceStrategy.NCCL)
     extra_attrs: Dict = PydanticField(default_factory=dict)
 
-    # Unified parallelism mapping (populated by setup_visual_gen_mapping)
+    # Unified parallelism mapping copied from the owning pipeline config.
     visual_gen_mapping: Optional[Any] = None  # VisualGenMapping (lazy import)
 
     dynamic_weight_quant: bool = False
 
-    # Sub-configs from VisualGenArgs (merged during from_pretrained)
+    # Shared runtime configs copied from the owning pipeline config.
     quant_config: QuantConfig = PydanticField(default_factory=QuantConfig)
     # Per-layer quant (from load_diffusion_quant_config layer_quant_config; None until mixed-precision parsing exists)
     quant_config_dict: Optional[Dict[str, QuantConfig]] = None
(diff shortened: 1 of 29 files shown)


tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -161,6 +222,33 @@ def get_quant_config(self, name: Optional[str] = None) -> QuantConfig:
...
+            allreduce_strategy=self.allreduce_strategy,
+            extra_attrs=self.extra_attrs,
+            visual_gen_mapping=self.visual_gen_mapping,
+            dynamic_weight_quant=self.dynamic_weight_quant,
+            quant_config=self.quant_config,
+            quant_config_dict=self.quant_config_dict,
+            compilation=self.compilation,
+            torch_compile=self.torch_compile,
+            cuda_graph=self.cuda_graph,
+            attention=self.attention,
```
chang-l, line 245: **Shared mutable references across components.** `_make_model_config` hands `attention`, `quant_config`, `cache`, and `extra_attrs` to every component **by reference** — each component's `DiffusionModelConfig` and the owning `DiffusionPipelineConfig` end up sharing the *same* objects. Fine today, but the very next step (per-component skip-softmax in #14687) needs to set a per-component `attention.sparse_attention_config`; mutating it in place would alias across all components. Suggest `model_copy()`-ing the per-component sub-configs here (at least `attention`) so per-component divergence is safe.

^^^ @bobboli  can you please check and see if you want to address this in this PR or later PR?

tensorrt_llm/_torch/visual_gen/config.py
```diff
@@ -566,8 +661,29 @@ def from_pretrained(
...
+        for component_name, config_dict in component_config_dicts.items():
+            if component_name == component:
+                component_pretrained_config = pretrained_config
+            else:
+                component_pretrained_config = SimpleNamespace(**config_dict)
+                if not getattr(component_pretrained_config, "_name_or_path", None):
+                    component_pretrained_config._name_or_path = getattr(
+                        pretrained_config, "_name_or_path", ""
+                    )
+            pipeline_config.model_configs[component_name] = pipeline_config._make_model_config(
```
chang-l, line 678: **Opportunity for #14687:** now that each component owns its own `pretrained_config`, the cleanest place to resolve sparse calibration is *here*, per component (from `component_config_dicts[name]`), rather than the module-name-prefix matching (`transformer.` / `transformer_2.`) in `sparse_attention.py`'s `_component_configs`. That would let `model_configs['transformer_2']` carry its own resolved skip-softmax `(a, b)` / `target_sparse_ratio` directly — replacing a stringly-typed prefix hack with structured per-component config. Worth coordinating the #14687 rebase around this. (Note: this PR by itself does not yet apply per-component sparse calibration — `sparse_attention_config` is still resolved once from the primary `transformer` config and the same `AttentionConfig` is shared to all components.)

tests/unittest/_torch/visual_gen/test_model_loader.py
```diff
@@ -50,13 +50,13 @@ def test_meta_init_mode_creates_meta_tensors(checkpoint_exists):
...
     from tensorrt_llm._torch.models.modeling_utils import MetaInitMode
-    from tensorrt_llm._torch.visual_gen.config import DiffusionModelConfig
+    from tensorrt_llm._torch.visual_gen.config import DiffusionPipelineConfig
     from tensorrt_llm._torch.visual_gen.models import AutoPipeline
     from tensorrt_llm.visual_gen.args import VisualGenArgs
 
     # Load config directly
     args = VisualGenArgs(model=CHECKPOINT_PATH)
-    config = DiffusionModelConfig.from_pretrained(
+    config = DiffusionPipelineConfig.from_pretrained(
```
chang-l, line 59: **Missing test for this PR's headline behavior.** The tests here only ever touch `.model_configs['transformer']`. There is no test asserting that a *two-transformer* checkpoint yields `model_configs` containing **both** `transformer` and `transformer_2`, each with a **distinct** `pretrained_config`. Since reading per-component configs is the entire point of this refactor, please add a test on a Wan2.2-style dual-transformer fixture — e.g. assert the two components' `pretrained_config.sparse_attention_config.threshold_scale_factor.prefill.a` differ (1443.5 vs 182.6 in the `wan22-skip-softmax` checkpoint). That's the regression guard for the split.

^^^ @bobboli can you address this?

tensorrt_llm/_torch/visual_gen/config.py
chang-l: ^^^ FYI and just a heads-up for your next PR – feel free to resolve this thread anytime

chang-l (approved):

tensorrt_llm/_torch/visual_gen/config.py
bobboli: Will do in https://github.com/NVIDIA/TensorRT-LLM/pull/14687

zhenhuaw-me: Does it make sense to let the "internal class" such as and PipelineConfig ModelConfig to reference the original `VisualGenArgs` objects for the "general configs" thus we don't need to copy and flatten all the fields down to internal utils?

zhenhuaw-me (approved):
Thank you so much for helping refactoring this settings :)

tburt-nv:
/bot run --stage-list "GB200-4_GPUs-PyTorch-1, GB200-4_GPUs-PyTorch-2, GB200-4_GPUs-PyTorch-3, GB200-4_GPUs-PyTorch-4"

tburt-nv:
If these stages pass, let's use "/bot reuse-pipeline"

tensorrt_llm/_torch/visual_gen/config.py
bobboli: Addressed in commit `136325cb0f` of https://github.com/NVIDIA/TensorRT-LLM/pull/14687: `_make_model_config()` now copies per-component model config values so mutable fields such as `attention` and `quant_config` do not alias across components. `mapping` / `visual_gen_mapping` intentionally use `deep_copy=False` because they carry distributed topology/process-group state.

bobboli: I would avoid having internal configs reference `VisualGenArgs` directly. `VisualGenArgs` is the public input schema; `DiffusionPipelineConfig` / `DiffusionModelConfig` should be normalized internal state with explicit ownership. The current approach keeps per-component values copied by default, while intentionally global objects like `mapping` / `visual_gen_mapping` stay shared. If this duplication grows later, I would prefer adding a narrow internal shared-config object instead of depending on the public args object.

tensorrt_llm/_torch/visual_gen/config.py
bobboli: Leaving this for #14687: per-component sparse calibration resolution will be handled as part of the sparse-attention refactor in that PR, rather than mixed into this config-split follow-up commit.

tests/unittest/_torch/visual_gen/test_model_loader.py
bobboli: Addressed in commit `136325cb0f` of https://github.com/NVIDIA/TensorRT-LLM/pull/14687: added a synthetic Wan-style two-transformer checkpoint test that verifies `transformer` and `transformer_2` each get separate `DiffusionModelConfig` / `pretrained_config` objects and do not share mutable per-model config objects like `attention`.

PR #14998, opened 2026-06-05
[None][feat] Weight trtllm-bench AR/AL averages by output length

zhaoyangwang-nvidia:
When requests produce different output lengths, an equally-weighted mean over per-request acceptance rate (AR) and acceptance length (AL) biases the reported average toward short requests, which run fewer decoding iterations. Detect inconsistent output lengths and, in that case, report an output-length weighted average for the AR/AL .average statistic so longer requests contribute proportionally. When all output lengths match, behavior is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of speculative decoding benchmark metrics reporting when requests produce varying output lengths by using output token count-weighted averaging instead of unweighted calculations.

## Summary
In `trtllm-bench`, the reported average acceptance rate (AR) and acceptance
length (AL) for speculative decoding were computed as an equally-weighted
mean over per-request values. When requests produce different output
lengths, this biases the average toward short requests (which run fewer
decoding iterations).

This PR detects whether all requests share the same output length. When
they do not, the AR/AL `.average` statistic is reported as an
output-length weighted mean — `Σ(output_i · value_i) / Σ(output_i)` — so
longer requests contribute proportionally. When all output lengths match,
behavior is unchanged.

## Changes
- `tensorrt_llm/bench/dataclasses/reporting.py`: compute
  `output_lengths_consistent` and override the AR/AL average with an
  output-length weighted mean when lengths differ.

## Test Coverage
N/A (statistics-only change; min/max/percentiles unchanged).

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/bench/dataclasses/reporting.py
+++ b/tensorrt_llm/bench/dataclasses/reporting.py
@@ -165,6 +165,7 @@ class StatsKeeper:
         num_accepted_draft_tokens = []
         draft_acceptance_rate = []
         acceptance_length = []
+        decoding_iterations = []
 
         for entry in self.requests.values():
             start_time = min(entry.start_timestamp, start_time)
@@ -193,6 +194,7 @@ class StatsKeeper:
                     float(num_draft_tokens[-1]))
                 acceptance_length.append(entry.num_total_output_tokens /
                                          (entry.decode_iteration + 1))
+                decoding_iterations.append(entry.decode_iteration + 1)
 
         global_acceptance_length = sum(
             output_tokens) / total_decoding_iterations
@@ -203,17 +205,21 @@ class StatsKeeper:
         num_accepted_draft_tokens_percentiles = PercentileStats.from_iterable(
             num_accepted_draft_tokens) if num_accepted_draft_tokens else None
         # Weight the per-request acceptance rate (AR) and acceptance length
-        # (AL) averages by output length. An equally-weighted mean would bias
-        # the result toward short requests, which run fewer decoding
-        # iterations; output-length weighting makes the .average a token-level
-        # mean so longer requests contribute proportionally. Percentiles are
-        # unaffected.
+        # (AL) averages by the number of decoding iterations the request ran.
+        # An equally-weighted mean would bias the result toward short requests,
+        # which run fewer decoding iterations; iteration weighting makes the
+        # .average a token-level mean so longer requests contribute
+        # proportionally. This also makes acceptance_length_percentiles.average
+        # equal the globally-computed acceptance_length
+        # (sum(output_tokens) / total_decoding_iterations), since AL_i =
+        # output_tokens_i / iterations_i and the weights cancel the per-request
+        # iterations. Percentiles are unaffected.
         draft_acceptance_rate_percentiles = PercentileStats.from_iterable(
             draft_acceptance_rate,
-            weights=output_tokens) if draft_acceptance_rate else None
+            weights=decoding_iterations) if draft_acceptance_rate else None
         acceptance_length_percentiles = PercentileStats.from_iterable(
             acceptance_length,
-            weights=output_tokens) if acceptance_length else None
+            weights=decoding_iterations) if acceptance_length else None
 
         requests = list(self.requests.values())
         stats = BenchmarkStatistics(
--- a/tensorrt_llm/bench/dataclasses/reporting.py
+++ b/tensorrt_llm/bench/dataclasses/reporting.py
@@ -207,6 +207,28 @@ class StatsKeeper:
         acceptance_length_percentiles = PercentileStats.from_iterable(
             acceptance_length) if acceptance_length else None
 
+        # Determine whether every request produced the same number of output
+        # tokens. When the output lengths differ, an equally-weighted mean over
+        # per-request AR/AL biases the result toward short requests (which run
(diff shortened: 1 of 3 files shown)


zhaoyangwang-nvidia:
Hi @NVIDIA/trtllm-bench-reviewers please help to review this PR, just very small fix, thanks~

tensorrt_llm/bench/dataclasses/reporting.py
```diff
@@ -207,6 +207,28 @@ def generate_statistics_summary(self, max_draft_tokens: int,
...
+            total_output_tokens = sum(output_tokens)
+            if total_output_tokens > 0:
+                if draft_acceptance_rate_percentiles is not None:
+                    draft_acceptance_rate_percentiles.average = sum(
+                        w * v
+                        for w, v in zip(output_tokens,
+                                        draft_acceptance_rate,
+                                        strict=True)) / total_output_tokens
+                if acceptance_length_percentiles is not None:
+                    acceptance_length_percentiles.average = sum(
```
dc3671: I think you should modify `PercentileStats` directly for AL calculation? rather than hack this special condition. Token level rather than request level weighted is reasonble for me.

BTW, can you add some tests for this?

dc3671: Same for AR

zhaoyangwang-nvidia: Thanks! I added an optional weights arg to PercentileStats.from_iterable (weighted mean sum(w*v)/sum(w)), and both AR and AL now pass output_tokens as weights — so it's a token-level average, no special-casing. When all output lengths are equal it reduces to the plain mean, so behavior is unchanged.

Question: I also added unit tests in test_bench_statistics.py. Do you think they're necessary to keep, or should I drop them to keep the PR minimal?

dc3671:
Thanks for adding `test_bench_statistics.py`. One concern on **test coverage**: the new test only exercises `PercentileStats.from_iterable` as a generic weighted-mean helper. It validates the arithmetic (`sum(w·v)/sum(w)`, zero-weight fallback, `strict=` length check) — but none of the AR/AL behavior this PR actually changes.

What's *not* covered is the layer that decides **what the weights are and how the per-request metrics are formed** — i.e. `StatsKeeper.generate_statistics_summary()` in `reporting.py`, where:

- per-request `AL_i = num_total_output_tokens / (decode_iteration + 1)` and `AR_i = num_accepted / num_draft`,
- the PR's key decision `weights=output_tokens` is applied,
- the result coexists with the separately-computed `global_acceptance_length = sum(output_tokens) / total_decoding_iterations`.

A passing `from_iterable` test gives no signal that the *reported AR/AL numbers* are correct, and there is currently no test on `StatsKeeper` at all. I'd suggest adding a test at that level — it's pure-Python / CPU-only, you can drive it directly:

```python
from tensorrt_llm.bench.dataclasses.reporting import StatsKeeper, PerfItemTuple

def _perf_item(rid, num_out, decoding_iteration):
    return PerfItemTuple(
        start_timestamp=0, end_timestamp=1000, request_id=rid,
        num_input_tokens=8, response_is_final=True, error=False,
        tokens=list(range(num_out)),              # num_total_output_tokens = num_out
        decoding_iteration=decoding_iteration,    # -> decode_iteration
        time_on_first_token=100,
    )

def test_acceptance_length_weighting():
    keeper = StatsKeeper()
    # short request high AL, long request low AL
    keeper.register_request_perf_item(_perf_item(0, num_out=10,  decoding_iteration=1))   # AL=5.0
    keeper.register_request_perf_item(_perf_item(1, num_out=300, decoding_iteration=199)) # AL=1.5
    stats = keeper.generate_statistics_summary(max_draft_tokens=3, batch_size=1)
    # documents the chosen weight (= output tokens): long request dominates
    weighted = (10*5.0 + 300*1.5) / (10 + 300)
    assert stats.acceptance_length_percentiles.average == pytest.approx(weighted)
```

This is also the place to encode the design intent. Since the goal is a "token-level" mean, it's worth deciding whether `acceptance_length_percentiles.average` should equal the report's existing `global_acceptance_length` (= `sum(out)/sum(iter)`). That equality holds only when AL is weighted by **iterations** (`decode_iteration+1`) rather than output tokens — with the current output-token weight the average is `Σ(out²/iter)/Σout`, a different number. A single assertion makes the choice explicit and regression-proof:

```python
# passes only under iteration-weighting; fails under output-token weighting
assert stats.acceptance_length_percentiles.average == pytest.approx(stats.acceptance_length)
```

Either way (keep output-token weighting or switch to iteration weighting), a `StatsKeeper`-level test is what actually guards the behavior this PR introduces.

zhaoyangwang-nvidia:
Hi @dc3671 Thanksfor your comments. You're right that the from_iterable test only covered the generic helper, not the actual AR/AL behavior. I've switched the weighting from output tokens to decoding iterations, so acceptance_length_percentiles.average now equals the report's global acceptance_length — the two AL numbers are consistent by construction. Added a StatsKeeper-level test (test_acceptance_length_weighted_by_iterations) that drives register_request_perf_item → generate_statistics_summary and asserts both the iteration-weighted value and average == acceptance_length (which would fail under output-token weighting).

dc3671 (approved):

PR #15002, opened 2026-06-05
[None][fix] tunable_fp4_quantize: rename misnamed kwarg + add real SF-swizzle control

luyiyun1021:
## Release Notes

* **Bug Fixes**
  * Corrected FP4 quantization dispatch to properly align scale factors between different inference backends

* **Enhancements**
  * Added configuration options for scale factor layout format and sizing to provide finer control over quantization behavior

## Description

Fixes a latent bug in `tunable_fp4_quantize` (added in PR #12126) where the Python wrapper's 4th kwarg, named `is_sf_swizzled_layout`, was misforwarded inside `_fp4_quantize_dispatch`. The TRTLLM dispatch branch passed it as the 4th positional argument to the 5-arg C++ `fp4_quantize` op, where the 4th slot is actually `sfUseUE8M0` (the MXFP4 toggle) and the 5th is `isSfSwizzledLayout`. As a result, three things were wrong: (1) the wrapper kwarg name lied about what it controlled; (2) the FlashInfer branch interpreted the same kwarg correctly as `do_shuffle` (swizzled), so the wrapper had divergent semantics across tactics; (3) callers had no way to actually control `isSfSwizzledLayout` — the C++ default `True` was always used.

The bug stayed latent because every existing call site passes positional `False` (production NVFP4 Linear in `tensorrt_llm/_torch/modules/linear.py`, plus the two cases in `tests/unittest/_torch/thop/parallel/test_fp4_quantize_flashinfer.py`), which lands as `sfUseUE8M0=False` (correct for NVFP4) and lets the C++ default `isSfSwizzledLayout=True` produce SWIZZLED output that downstream `nvfp4_gemm` consumes. Trying to flip the swizzled flag by passing `True` instead crashes with `RuntimeError: sfVecSize can only be 32, when sfUseUE8M0 is true` (the UE8M0 + sf_vec_size=16 combination is rejected by the C++ op).

The fix renames the 4th wrapper kwarg to `sf_use_ue8m0` (matching what the TRTLLM dispatch actually controls), adds a real 5th kwarg `is_sf_swizzled_layout: bool = True` (matching the C++ default), and threads both through the dispatch helper, `Fp4QuantKernelRunner`, and the fake registration. The FlashInfer branch now asserts `not sf_use_ue8m0` (FlashInfer has no MXFP4 path) and uses the new `is_sf_swizzled_layout` for `do_shuffle`. All existing call sites continue to pass positional `False`, which now binds to `sf_use_ue8m0=False` while `is_sf_swizzled_layout` falls back to the new default `True` — so each existing caller's effective C++ call is byte-identical to pre-fix.

## Test Coverage

- `tests/unittest/_torch/thop/parallel/test_fp4_quantize_flashinfer.py` (the wrapper's own op-level test, both `test_tunable_fp4_quantize_op` and `test_tunable_fp4_quantize_with_autotune`) — pass.
- LTX-2 transformer block tests (which exercise the production NVFP4 Linear path through this wrapper) — pass.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
+++ b/tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
@@ -2417,11 +2417,22 @@ class Fp4QuantTactic(enum.IntEnum):
 
 
 def _fp4_quantize_dispatch(input: torch.Tensor, input_scale: torch.Tensor,
-                           scaling_vector_size: int,
+                           scaling_vector_size: int, sf_use_ue8m0: bool,
                            is_sf_swizzled_layout: bool,
                            tactic: int) -> Tuple[torch.Tensor, torch.Tensor]:
-    """Dispatch FP4 quantization to TRTLLM or FlashInfer kernel."""
+    """Dispatch FP4 quantization to TRTLLM or FlashInfer kernel.
+
+    The two layout bools map straight through to the C++ ``fp4_quantize`` op
+    arguments of the same name. The earlier wrapper exposed only a single
+    ``is_sf_swizzled_layout`` kwarg that, in the TRTLLM branch, was actually
+    sent to the C++ ``sfUseUE8M0`` slot -- so the param name lied. This
+    dispatch is now explicit on both axes.
+    """
     if tactic == Fp4QuantTactic.FLASHINFER and IS_FLASHINFER_AVAILABLE:
+        # FlashInfer FP4 kernel does not support MXFP4 (UE8M0) scaling.
+        assert not sf_use_ue8m0, (
+            "FlashInfer FP4 tactic does not support sf_use_ue8m0=True; "
+            "force the TRTLLM tactic.")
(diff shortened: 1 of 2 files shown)


chang-l (approved):

tensorrt_llm/_torch/custom_ops/torch_custom_ops.py
chang-l, line 2425: If FI is selected but `sf_use_ue8m0` is somehow passed, would that cause a problem?
can we exclude FI in `get_valid_tactics` when `self.sf_use_ue8m0` is True, so the tuner never selects or profiles it?

luyiyun1021: Excluded FI in get_valid_tactics when self.sf_use_ue8m0

hyukn (approved):

PR #15016, opened 2026-06-05
[TRTLLM-12534][fix] Nemotron Nano - properly account for text prompts in inflight batching with EVS on

moraxu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multimodal video context handling in Nemotron Nano V2VL models.
  * Fixed token-positioning so mixed text-only and multimodal requests in a batch no longer overwrite each other.
  * Ensured multimodal embeddings are placed accurately within flattened token sequences.
  * Removed an overly-restrictive batch validation that rejected certain multimodal batching patterns.

* **Tests**
  * Added regression tests covering interspersed multimodal/text contexts and chunked-prefill scenarios.

## Problem

`merge_evs_mm_embeds` wrote each request's EVS-adjusted token IDs at a fixed offset 0:

```python
input_ids[: context_ids.shape[0]] = context_ids
```

This assumes multimodal context requests are a contiguous prefix of the flattened `input_ids`. With in-flight batching that's false - any text-only context request scheduled ahead of the EVS one gets silently clobbered (independent of chunked prefill). A guard (`_validate_evs_context_batch`) merely *rejected* such batches as a stopgap.

```
batch layout:  [ A text-only ] [ B EVS video ] [ decode ]
write target:  input_ids[0 : len(B)]
result:        B's IDs land on top of A  (A clobbered)
```

## Fix

Record where each request begins in the flattened `input_ids` (known in the model engine as `len(input_ids)` before the buffer is extended) on a new `MultimodalParams.input_ids_start_offset`, and write each request into its own span:

```python
for multimodal_param, context_part in zip(multimodal_params, context_parts, strict=True):
    start = multimodal_param.input_ids_start_offset
    input_ids[start : start + context_part.shape[0]] = context_part
```

Safe because the merge is **length-preserving**: the slot is pre-sized to the post-EVS token count and EVS only redistributes tokens across tubelets, never growing the span. The `_validate_evs_context_batch` guard is removed.

```
batch layout:  [ A text-only ] [ B EVS video ] [ decode ]
write target:  input_ids[len(A) : len(A) + len(B)]
result:        only B's span written  (A and decode untouched)
```

## Changes

- `inputs/multimodal.py` - add `input_ids_start_offset: int = 0` to `MultimodalParams`.
- `_torch/pyexecutor/model_engine.py` - record the offset in both context-prep paths.
- `_torch/models/modeling_nemotron_nano.py` - per-request offset writes; drop `_validate_evs_context_batch`.

## Test Coverage

`test_nemotron_nano_preprocessing.py::TestMergeEvsMMEmbeds` - replaced the rejection test with positive coverage: text-only before/after, interspersed multi-request, and chunked prefill with a preceding text-only request.

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
@@ -2861,13 +2861,13 @@ class NemotronH_Nano_VL_V2(transformers.PreTrainedModel):
         for multimodal_param, context_part in zip(multimodal_params, context_parts, strict=True):
             start = multimodal_param.input_ids_start_offset
             end = start + context_part.shape[0]
-            if end > input_ids.shape[0]:
+            if not (0 <= start <= end <= input_ids.shape[0]):
                 raise ValueError(
-                    f"EVS-adjusted context span ([{start}, {end})) exceeds "
-                    f"input_ids length ({input_ids.shape[0]}). This usually means "
-                    "a chunked multimodal request reached merge_evs_mm_embeds "
-                    "without multimodal_runtime metadata, or input_ids_start_offset "
-                    "was not set for this request."
+                    f"EVS-adjusted context span [{start}, {end}) is invalid for "
+                    f"input_ids of length {input_ids.shape[0]} (expected "
+                    "0 <= start <= end <= length). This usually means an incorrect "
+                    "input_ids_start_offset, or a chunked multimodal request reached "
+                    "merge_evs_mm_embeds without multimodal_runtime metadata."
                 )
             input_ids[start:end] = context_part.to(device=input_ids.device, dtype=input_ids.dtype)
 
--- a/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
+++ b/tensorrt_llm/_torch/models/modeling_nemotron_nano.py
@@ -2776,41 +2776,6 @@ class NemotronH_Nano_VL_V2(transformers.PreTrainedModel):
         runtime.total_embeds_in_request = int(embed_mask_cumsum[-1])
         return context_ids[runtime.past_seen_token_num : runtime.chunk_end_pos]
 
-    @staticmethod
-    def _validate_evs_context_batch(
-        ctx_params: List[MultimodalParams],
-        num_context_requests: int,
-    ) -> None:
-        """Reject EVS video mixed with text-only context requests.
-
-        Args:
-            ctx_params: Multimodal params associated with current context
-                requests. Today this list contains only context requests with
-                multimodal content.
-            num_context_requests: Total number of current context requests in
-                the flattened forward batch, including text-only requests.
-
-        Raises:
-            ValueError: If an EVS video context is batched together with a
-                text-only context request.
-        """
-        has_video = any(
-            param.has_content() and param.multimodal_data.get("modality_type") == "video"
-            for param in ctx_params
-        )
-        has_text_only_context = len(ctx_params) < num_context_requests or any(
-            not param.has_content() for param in ctx_params
-        )
-        if has_video and has_text_only_context:
-            # TODO(TRTLLM-12534): Remove this guard once merge_evs_mm_embeds writes each
-            # multimodal context chunk using its flattened input_ids offset
-            # instead of assuming a contiguous multimodal prefix.
-            raise ValueError(
-                "EVS video requests cannot be inflight-batched with text-only "
-                "context requests yet. merge_evs_mm_embeds currently assumes "
-                "multimodal context chunks form a contiguous input_ids prefix."
-            )
-
     def _check_encoders_exist(self, raw_ctx_params: List[MultimodalParams]) -> None:
         """Check encoders needed by raw inputs exist.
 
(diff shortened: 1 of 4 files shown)


moraxu:
@coderabbitai full review

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -2677,7 +2681,8 @@ def _prepare_tp_inputs(
 
             multimodal_params = MultimodalParams(
                 multimodal_data=request.py_multimodal_data,
-                multimodal_runtime=py_multimodal_runtime)
+                multimodal_runtime=py_multimodal_runtime,
+                input_ids_start_offset=context_start_idx)
```
2ez4bz, line 2739: Sorry, I may be missing something - how is this an offset? Aren't we always counting from zero when we define `context_start_idx = len(input_ids)` at line 2640?

tests/unittest/_torch/modeling/test_nemotron_nano_preprocessing.py
```diff
@@ -1124,6 +1119,142 @@ def test_chunked_prefill_first_chunk_can_be_shorter_than_full_context(self):
...
+                decode.clone(),
+            ]
+        )
+
+        result = NemotronH_Nano_VL_V2.merge_evs_mm_embeds(
+            model, num_tokens_in_videos, [b_param, d_param], input_ids
+        )
+
+        expected_b = torch.tensor([_IMG_START] + [_IMG_CTX_ID] * 3 + [_IMG_END], dtype=torch.long)
+        assert (result[0:3] == a_text).all()
```
2ez4bz, line 1200: Nit: could just construct the entire expected tensor and have a single `assert (x == y).all()?

tensorrt_llm/_torch/pyexecutor/model_engine.py
moraxu: So `input_ids` is initialized once, outside the loop at line 2595 (`input_ids = []`), then this loop accumulates every request's tokens into that same list via `input_ids.extend(prompt_tokens)` above. So `context_start_idx = len(input_ids)` is the running length so far - it's `0` only on the first iteration.

tests/unittest/_torch/modeling/test_nemotron_nano_preprocessing.py
moraxu: I think I will keep it as is - those per-span asserts are verifying which spans get written vs. left untouched (the clobbering regression). We'd have to do more digging which region broke exactly with just `assert (x == y).all()`

tensorrt_llm/_torch/pyexecutor/model_engine.py
2ez4bz: Oh I see. Thanks for explaining.

2ez4bz (approved):

tensorrt_llm/inputs/multimodal.py
```diff
@@ -496,6 +502,7 @@ class MultimodalParams:
     multimodal_input: Optional[MultimodalInput] = None
     multimodal_data: Optional[Dict[str, Any]] = field(default_factory=dict)
     multimodal_runtime: Optional[MultimodalRuntimeData] = None
+    input_ids_start_offset: int = 0
```
achartier, line 505: Is the expectation that the user will always override this value? If so, maybe it should be an Optional and add a check for None?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -2882,24 +2847,29 @@ def merge_evs_mm_embeds(
...
+        # contiguous prefix: text-only context requests carry no multimodal_params
+        # (so they never appear here) yet still occupy input_ids, and they may
+        # precede or sit between multimodal ones. Each write therefore starts at
+        # the request's recorded flattened offset. The merge is length-preserving
+        # (the slot was pre-sized to the post-EVS retained count), so every span
+        # ends exactly where the next request begins and `input_ids` is never
+        # grown. Under chunked prefill, multimodal_runtime narrows each context
+        # part to the current chunk so decode ids after it are preserved.
+        for multimodal_param, context_part in zip(multimodal_params, context_parts, strict=True):
+            start = multimodal_param.input_ids_start_offset
```
achartier, line 2862: Maybe add a check that it's >= 0

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
```diff
@@ -2882,24 +2847,29 @@ def merge_evs_mm_embeds(
...
+            end = start + context_part.shape[0]
```
achartier, line 2863: Should we also check that start is before end?

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Added in https://github.com/NVIDIA/TensorRT-LLM/pull/15016/commits/3728c57df0d4a05c80231290ef607abd4b45589a

tensorrt_llm/_torch/models/modeling_nemotron_nano.py
moraxu: Added in https://github.com/NVIDIA/TensorRT-LLM/pull/15016/commits/3728c57df0d4a05c80231290ef607abd4b45589a

tensorrt_llm/inputs/multimodal.py
moraxu: This is engine-populated and the first request in a batch legitimately starts at offset `0`

achartier (approved):

2ez4bz:
first try

PR #15025, opened 2026-06-05
[None][fix] Disagg: handle ctx-only completion when ctx_request_id is None with terminal finish_reason

fredricz-20070104:
## Description

Fixes a flaky HTTP 500 in disaggregated serving (nvbugs/6245861).

With MTP, the context/prefill worker occasionally finishes a request (EOS, `finish_reason='stop'`) before the KV-cache handoff to the GEN worker is initiated, so `ctx_request_id` is never populated — yet the response is a complete, valid answer. The previous code in `_verify_ctx_response` raised unconditionally in that case (`ValueError`), turning a valid context-only completion into an intermittent HTTP 500 (observed ~1/640–1/16384).

## Fix

- `tensorrt_llm/serve/openai_disagg_service.py`: short-circuit when `ctx_request_id is None` **and** the CTX choice has a terminal `finish_reason`, returning the context-only completion instead of raising (mirrors the existing `_need_gen` logic).
- `tests/unittest/disaggregated/test_openai_disagg_service.py`: regression test for the ctx-only terminal case.

## Test Coverage

Added unit test reproducing the `ctx_request_id=None` + terminal `finish_reason` path.

Regression introduced by 7443f1d05f.

### DIFF
--- a/tensorrt_llm/serve/openai_disagg_service.py
+++ b/tensorrt_llm/serve/openai_disagg_service.py
@@ -394,6 +394,14 @@ class OpenAIDisaggregatedService(OpenAIService):
                     f" finish_reason={choice.finish_reason!r}"
                 )
             if choice.disaggregated_params.ctx_request_id is None:
+                # MTP early-termination edge case: the context phase may finish
+                # the whole request (e.g. EOS) before the KV-cache handoff is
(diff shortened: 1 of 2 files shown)


fredricz-20070104:
For test only, close it.

PR #15046, opened 2026-06-06
[None][feat] Add KV cache block reuse policy

jiaganc:
@coderabbitai summary

## Description

Add a KV cache block reuse policy knob for cache manager v2. The new `block_reuse_policy` defaults to `all_committed`, and allows `per_request` to commit only sliding-window blocks when stop-committing is called. Scratch reuse is controlled by `KvCacheConfig.enable_swa_scratch_reuse` (default true) instead of an environment variable. `block_reuse_policy=all_committed` with scratch reuse enabled saves only non-scratch blocks.

The change wires the policies through LLM args, PyExecutor resource manager setup, generic KV cache manager v2 construction, and KV cache manager v2 config/types.

## Experiment Results

DeepSeek-V4-Pro disaggregated serving, `ctx1x4_gen1x8`, `dep8`, `mtp3`, `bs64`, `c160`. Metrics below are CTX-server only.

Data sources:
- Offload/onboard blocks: CTX server `[KVCacheMetrics]` interval counters from `3_output_CTX_0.log`.
- Hit rate: latest CTX `kvCacheStats` per rank from `metrics_poll_CTX.jsonl`, aggregated as `reusedBlocks / (reusedBlocks + missedBlocks)`.

| metric | per_request | all_committed | delta |
|---|---:|---:|---:|
| requests | 33,956 | 8,547 | +25,409 |
| offload blocks | 74,081 | 81,844 | -7,763 |
| onboard blocks | 33,958 | 3,716 | +30,242 |
| CTX hit rate | 92.92% | 66.35% | +26.58 pp |
| offload blocks / request | 2.18 | 9.58 | -7.39 |
| onboard blocks / request | 1.00 | 0.43 | +0.57 |

Per-rank CTX comparison:

| rank | per_request offload | all_committed offload | delta | per_request onboard | all_committed onboard | delta | per_request hit | all_committed hit | delta |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 0 | 19,632 | 22,237 | -2,605 | 9,329 | 856 | +8,473 | 93.11% | 31.27% | +61.84 pp |
| 1 | 18,752 | 21,940 | -3,188 | 8,649 | 927 | +7,722 | 93.23% | 29.53% | +63.71 pp |
| 2 | 17,784 | 21,307 | -3,523 | 7,932 | 885 | +7,047 | 92.61% | 29.84% | +62.77 pp |
| 3 | 17,913 | 16,360 | +1,553 | 8,048 | 1,048 | +7,000 | 92.70% | 93.97% | -1.27 pp |

Notes:
- `per_request` has much higher CTX cache hit rate overall, mostly because ranks 0-2 improve from roughly 30% to roughly 93%.
- `per_request` logged about 4.0x more requests, so normalized offload/onboard volume is important: it offloaded fewer blocks per request, but onboarded more blocks per request.
- `all_committed` rank 3 is already high-hit-rate, so the per-rank hit-rate improvement is concentrated on ranks 0-2.

## Test Coverage

- Added unit coverage in `tests/unittest/kv_cache_manager_v2_tests/test_kv_cache_manager_v2.py`.
- Added LLM args coverage in `tests/unittest/llmapi/test_llm_args.py`.
- Not run after dropping the stacked metrics commit: local environment has no `pytest` (`pytest`: command not found; `python3 -m pytest`: No module named pytest).

## PR Checklist

- PR description clearly explains what and why. If using CodeRabbit summary, please make sure it makes sense.

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/sparse/deepseek_v4/cache_manager.py
+++ b/tensorrt_llm/_torch/attention_backend/sparse/deepseek_v4/cache_manager.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 from collections import defaultdict
 from typing import Dict, List, Optional, Tuple
 
@@ -62,10 +61,6 @@ from .deepseek_v4 import (
     get_token_bytes,
     is_overlap_compressor,
 )
-
-# Keep the env override so per-layer block tables can still be tested
-# without scratch-page remapping when needed.
-DSV4_ENABLE_SWA_SCRATCH_REUSE_ENV = "TRTLLM_DSV4_ENABLE_SWA_SCRATCH_REUSE"
 
 
 def _estimate_non_sliding_attn_size_per_token(
@@ -243,11 +238,6 @@ def _compute_sliding_block_tables_with_scratch_compiled(
     output[:, :, :scratch_capacity, :].copy_(torch.where(mask, scratch_index, scratch_rows))
 
 
-def _enable_swa_scratch_reuse_from_env() -> bool:
-    value = os.environ.get(DSV4_ENABLE_SWA_SCRATCH_REUSE_ENV, "1")
-    return value.strip() == "1"
-
-
 class DeepseekV4CacheManager(KVCacheManagerV2):
     # This tensor is for compatibility with AttentionOp, it only contains swa attention.
     # kv_cache_pool_pointers contains one virtual attention-op pool per local
@@ -339,7 +329,6 @@ class DeepseekV4CacheManager(KVCacheManagerV2):
             vocab_size=vocab_size,
             mapping=mapping,
             dtype=dtype,
-            enable_swa_scratch_reuse=_enable_swa_scratch_reuse_from_env(),
             **kwargs,
         )
         self.is_vswa = True  # DeepSeek-V4 must has VSWA
--- a/tensorrt_llm/_torch/models/modeling_deepseekv4.py
+++ b/tensorrt_llm/_torch/models/modeling_deepseekv4.py
@@ -2490,7 +2490,13 @@ class DeepseekV4Model(DecoderModel):
 class DeepseekV4ForCausalLM(SpecDecOneEngineForCausalLM[DeepseekV4Model, PretrainedConfig]):
     @classmethod
     def get_model_defaults(cls, llm_args: "TorchLlmArgs") -> dict:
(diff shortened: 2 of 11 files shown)


lowsfer:
This is for multi-round / agentic usage, right? The originally assumed usage pattern for this is that the runtime manages lifetime of suspended requests. These suspended requests always have committing stopped, so blocks never goes to block radix tree. When next round prompt arrives, we resume that request directly and it will only have the necessary blocks because committing was stopped.

This PR's effect is somewhat similar for one round, in that it also only keeps the last SWA blocks. But there are two differences:
1. we need to pay the radix tree match cost
2. when the second round finished, in principle we can drop the last SWA blocks of the first round immediately, but that doesn't happen with this PR.

But I agree our runtime may need more change for that usage pattern. @jiaganc I'm fine with this design to unblock DSV4. Just want to discuss. Do you think we eventually want to use suspend/resume and request-level caching?

tensorrt_llm/runtime/kv_cache_manager_v2/__init__.pyi
```diff
@@ -45,6 +47,10 @@ class PageIndexMode(enum.IntEnum):
     SHARED = 0
     PER_LAYER = 1
 
+class BlockReusePolicy(StrEnum):
```
lowsfer: I think it's more about committing, instead of reuse. Can we refactor it? I think one option is:

class CommitMode:
    FULL
    MIN_SNAPSHOT

I think that would better describe the purpose: save all data, or a minimum snapshot that can be used to recover the state for all the committed tokens.

lowsfer: BTW, can we use int enum instead? that would be better for C++ translation.

jiaganc: Now I implement the feature in runtime and do not touch the cache manager's codes.

jiaganc:
> But I agree our runtime may need more change for that usage pattern. @jiaganc I'm fine with this design to unblock DSV4. Just want to discuss. Do you think we eventually want to use suspend/resume and request-level caching?

After offline discuss with @lowsfer, we agree that we need session/conversation level of caching, but suspend/resume patter in the previous comment is not a good way to implement this feature. Need further discussion and will implement it in another PR.

lowsfer (approved):

PR #15063, opened 2026-06-07
[#14225][perf] AutoDeploy MTP + ADP enablement and MoE all-to-all optimization

MrGeva:
## Description

Optimizes the SuperV3 MTP + attention-DP path, building on the recently merged
MoE all-to-all stateful cache (#13718 / #13723) and the SSM-replay PR.

**MoE all-to-all per-rank token budget (`runtime_max_tokens_per_rank`)**
- Replaces the per-iteration cross-rank-max read (an `int(batch_info_host[14].item())`
  on a pinned-host tensor, fed by a per-forward `tp_allgather`) with a sync-free,
  shape-based budget via `_hybrid_runtime_max_tokens_per_rank`:
  - Under cuda-graph capture/warm-up the budget is `x.shape[0]` — uniform across DP
    ranks because `maybe_pad_for_cuda_graph` pads every rank to a common
    `cg_batch_size` and MTP tokens-per-seq is uniform — gated so the tight budget is
    only taken while the MoE-GEMM row count stays in the fast small-M tactic region.
  - In eager (prefill, or any step the mixed-mode bypass forces eager) it falls back
    to the static `max_num_tokens`, which every rank computes identically.
  - No per-layer host read, and no extra collective.
- Drops the now-dead `batch_info_host` DP-max plumbing: the slot-14 (`max_dp_num_tokens`)
  storage + accessors in `BatchInfo` (`_NUM_ELEMENTS` 15→14), the pre-forward
  `tp_allgather`+update in the AutoDeploy shim, the injection into the MoE op in both
  the dict-based (`sharding.py`) and IR-based (`sharding_ir.py`) sharding paths, and the
  corresponding op signatures in `trtllm_moe.py` / `torch_moe.py`.

**Scheduling / sharding**
- Keeps the draft-EP revert under attention-DP in sharding (replicate the draft model's
  MoE rather than EP-sharding it) to avoid the shared-workspace corruption that hangs the
  all-to-all at concurrency.

Mixed-mode cuda-graph bypass uses the process-wide `BypassCapturedGraphs()` context
manager (`cuda_graph_state.in_bypass()`), keeping all ranks consistent when one is in prefill.

## Perf Impact
see: https://github.com/NVIDIA/TensorRT-LLM/issues/14225#issuecomment-4601016969

## Test Coverage

- `tests/integration/defs/accuracy/test_llm_api_autodeploy.py::TestNemotronSuperV3::test_mtp[*]`
  and `::test_accuracy[*-4-attn_dp_on-trtllm]` (bf16/fp8/nvfp4) — exercise the MTP +
  attention-DP MoE all-to-all path.
- `tests/unittest/auto_deploy/multigpu/compile/test_bypass_captured_graphs.py` — guards the
  mixed-mode captured-graph bypass.
- New perf-sanity config + post-merge enrollment:
  `perf/test_perf_sanity.py::test_e2e[aggr_upload-super_mtp_ad_nvfp4_blackwell-super_mtp_ad_nvfp4_ws4_1k1k]`.
- Local e2e validation (4× GB200, TP4, BF16 SuperV3-MTP, attention-DP): builds end-to-end
  with no NaN / hang / all-to-all timeout; GSM8K 47/50 = 94.0%.

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added low-latency SuperV3 MTP configuration optimized for reduced serving latency.

* **Configuration Updates**
  * Enhanced SuperV3 MTP configuration with batch-size targets and attention-DP support.
  * Updated speculative decoding (Eagle) to properly handle attention-DP mode.

  * Optimized MoE all-to-all dispatch to use actual token counts instead of padded estimates, reducing overhead.
  * Streamlined batch metadata handling for more efficient attention computation.

### DIFF
--- a/examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml
+++ b/examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml
@@ -1,29 +1,32 @@
-# Config for SuperV3 with MTP speculative decoding.
-#
+# Config for SuperV3 with MTP speculative decoding + attention-DP.
 runtime: trtllm
 compile_backend: torch-cudagraph
-max_batch_size: 64
-max_seq_len: 8192 # tunable
-max_num_tokens: 2048
+# For low-concurrency / latency-sensitive serving, lower max_batch_size to 16 for better latency.
+max_batch_size: 128
+max_seq_len: 65536 # tunable
 enable_chunked_prefill: false # TODO: Support chunked prefill w/ spec dec
 attn_backend: trtllm
 model_factory: AutoModelForCausalLM
 skip_loading_weights: false
 cuda_graph_config:
-  max_batch_size: 64
-  enable_padding: true
+  batch_sizes: [1, 2, 4, 8, 16, 24, 32, 64, 128]
 kv_cache_config:
   mamba_ssm_cache_dtype: auto
 speculative_config:
   decoding_type: MTP
   num_nextn_predict_layers: 6
   mtp_eagle_one_model: true
+attention_dp_config:
+  enable_balance: true
+  batching_wait_iters: 10
+  timeout_iters: 16
 transforms:
   compile_model:
     # MTP speculative decoding does not support piecewise CUDA graph capture yet.
     piecewise_enabled: false
   detect_sharding:
     allreduce_strategy: NCCL
+    enable_attention_dp: true
     # NOTE: add 'tp' to sharding dims only for high-throughput runs
     # For low-latency, keep mamba and attention replicated
     sharding_dims: ['ep', 'bmm']
@@ -46,10 +49,9 @@ transforms:
         # MoLE: latent projections: simple shard
         "fc1_latent_proj": "gather"
         "fc2_latent_proj": "gather"
-  # multi_stream_moe disabled - creates NaNs in target's sampled logits. TODO: investigate + fix
   multi_stream_moe:
     stage: compile
-    enabled: false
+    enabled: true
   fuse_mamba_a_log:
     stage: post_load_fusion
     enabled: true
--- a/tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe/trtllm_moe.py
+++ b/tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe/trtllm_moe.py
@@ -23,10 +23,55 @@ from tensorrt_llm._utils import get_sm_version
 from tensorrt_llm.mapping import Mapping
 
 from ..._compat import ActivationType, is_sm_100f
+from ...utils.cuda_graph import cuda_graph_state
 from ...utils.dist_config import DistConfig
 from ..quantization.quant import TRTLLM_NVFP4_SCALING_VECTOR_SIZE
 
 _f32_scale_cache: dict = {}
+
+
+def _hybrid_runtime_max_tokens_per_rank(
+    local_num_tokens: int,
+    ep_size: int,
+    max_num_tokens: int,
+) -> int:
+    """Pick the MoE all-to-all per-rank token budget (dispatch/sanitize/combine/GEMM stride).
+
+    Hybrid of the cross-rank-max (per-layer ``.item()``) and static-``max_num_tokens``
+    strategies that avoids the pitfalls of each. Callers never pad ``x``: the dispatch kernel
+    writes only the real ``local_num_tokens`` rows and ``moe_a2a_sanitize_expert_ids``
+    invalidates the unfilled slots, so the returned value controls only the workspace stride /
+    recv-view / GEMM row count, not how many real rows are moved.
+
+    - **Capture / warm-up path** (decode/extend): use ``local_num_tokens``. Under attention-DP
+      every rank pads to the same ``cg_batch_size``, so the per-rank count already equals the
+      cross-rank max for the captured bucket, and at pure-decode replay it stays consistent.
+      It is a *shape*, not a device scalar, so there is **no per-layer ``.item()`` host read**
+      (the cost that made the cross-rank-max read regress at low concurrency). Mixed
+      prefill/decode — where a captured value would be stale — is routed to eager by
+      ``maybe_pad_for_cuda_graph`` via ``BypassCapturedGraphs()`` (#13718), so the captured
+      path only sees the uniform decode shape. The tight budget is taken only while the
+      resulting MoE-GEMM row count (``budget * ep_size``) stays in the fast small-M
+      ``bmm_E2m1`` tactic region; above the gate the trtllm-gen autotuner picks a
+      launch-latency-bound tactic, so fall back to the big-M
+      ``max_num_tokens`` budget.
+
+    - **Eager path** (prefill, or any step the bypass forced eager): use ``max_num_tokens``.
+      It is a constant every rank computes identically (layout agrees across ranks) and
+      involves no host sync. Prefill is compute-bound, so the fat recv view is amortized.
+    """
+    if torch.cuda.is_current_stream_capturing() or cuda_graph_state.in_warm_up():
+        budget = local_num_tokens
+        # Only take the tight budget while the MoE-GEMM row count (budget * ep_size) stays in
+        # the fast small-M trtllm-gen tactic region; above it the autotuner switches to a
+        # launch-latency-bound tactic.
+        # The threshold is on the FC1 grouped-GEMM's row dimension M, because that's what the
+        # trtllm-gen autotuner selects the kernel tactic from — and that M is budget × ep_size
+        # TODO: remove this WAR https://github.com/NVIDIA/TensorRT-LLM/issues/15167
+        # See: https://nvbugspro.nvidia.com/bug/6247543
+        if budget > 0 and budget * ep_size * 4 <= max_num_tokens:
+            return budget
+    return max_num_tokens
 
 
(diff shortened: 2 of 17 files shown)


tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe/trtllm_moe.py
```diff
@@ -22,12 +22,50 @@
...
+      launch-latency-bound tactic (the c=32 regression), so fall back to the big-M
+      ``max_num_tokens`` budget.
+
+    - **Eager path** (prefill, or any step the bypass forced eager): use ``max_num_tokens``.
+      It is a constant every rank computes identically (layout agrees across ranks) and
+      involves no host sync. Prefill is compute-bound, so the fat recv view is amortized.
+    """
+    if torch.cuda.is_current_stream_capturing() or cuda_graph_state.in_warm_up():
+        budget = local_num_tokens
+        if budget > 0 and budget * ep_size * 4 <= max_num_tokens:
```
MrGeva, line 72: add the bug url

examples/auto_deploy/model_registry/configs/super_v3_mtp_low_latency.yaml
```diff
@@ -0,0 +1,77 @@
...
+#   GPU memory; if you'll never serve c>16, capturing larger sizes is
+#   wasted.
+# - Everything else (ADP, sharding, MoE fusion) matches the throughput
+#   yaml so the MoE alltoall + tight-slice optimization still applies.
+#
+# Pair this with serving at low concurrency (c=1..16). For higher c, use
+# ``super_v3_mtp.yaml`` and accept the higher TPOT.
+runtime: trtllm
+compile_backend: torch-cudagraph
+max_batch_size: 16
```
MrGeva: since this is the only diff, consider removing this file and commneting in the other yaml to switch to it for low latency

tensorrt_llm/_torch/auto_deploy/custom_ops/fused_moe/trtllm_moe.py
```diff
@@ -22,12 +22,50 @@
...
+            return budget
+    return max_num_tokens
```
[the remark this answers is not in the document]
MrGeva: WARM_UP/capturing are capture-time states; BYPASS is a runtime decision in maybe_pad_for_cuda_graph. They're mutually exclusive — they never co-occur. So during a bypass
  step both gate conditions are False → the if is skipped → the function already returns max_num_tokens, which is exactly the desired eager budget.

  The full call matrix confirms correctness:
  - Capture / warm-up: tight local_num_tokens baked into the decode-bucket graph ✓ (intended)
  - Replay (normal decode): the op isn't re-run — the baked value is reused ✓
  - Bypass-forced eager (mixed prefill/decode): not capturing, not warm-up → max_num_tokens ✓

  So the bypass flag is implicitly respected. Adding and not cuda_graph_state.in_bypass() would be a pure no-op (the branch is never reached under bypass) — it adds nothing
  and violates "keep changes minimal." No code change.

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
```diff
@@ -404,6 +404,14 @@ def __init__(
         self.llm_args.print_iter_log = reporting_info.print_log
         self.llm_args.enable_iter_perf_stats = reporting_info.enable_iter_perf_stats
         self.llm_args.enable_iter_req_stats = reporting_info.enable_iter_req_stats
+        self.llm_args.stream_interval = 1
+        # The attention-DP request balancer (PyExecutor._balance_adp_requests) is left
+        # disabled: empirically it is neutral-to-negative on SuperV3-MTP serving throughput
```
govind-ramnarayan: nit: I'd rather not have references to optimizations on particular models  in generic infra code - especially since these seem like they will go stale very quickly. If we want to keep track of this maybe we create a GitHub issue that contains these specific details (e.g. SuperV3-MTP serving input not being great) and is about evaluating the attention-DP request balancer more generally.

tensorrt_llm/_torch/auto_deploy/transform/library/sharding.py
```diff
@@ -1233,6 +1233,27 @@ def _apply(
             if ShardingDim.EP in config.sharding_dims:
                 ad_logger.info("Running autodeploy EP sharding heuristics")
                 info += detect_ep_shard(gm, transform_container)
+                # Under attention-DP, EP-sharding the draft submodel forces its
```
govind-ramnarayan, line 1217: I don't understand completely, but this sounds like it is conveniently working around a correctness issue (draft submodel having MoE all to all) with a performance observation (it's okay if draft models are unsharded from a performance standpoint because they are small). I think the correctness issue deserves a GitHub issue so that we don't forget about it - drafters could get bigger (indeed there are some use cases for using e.g. a GPT-oss 20B as a meaningful drafter for GPT-oss 120B) in which case we may need to revisit this for good performance.

tensorrt_llm/_torch/auto_deploy/llm_args.py
```diff
@@ -441,6 +441,14 @@ def create_factory(self) -> ModelFactory:
             `self.max_seq_len` so that all downstream consumers see the same value.
         """
 
+        # Resolve enable_attention_dp from the same source `init_dist_config`
+        # uses so the factory sees the same value the runtime will see.
+        ash = self.transforms.get("apply_sharding_hints", {})
+        sharding_config = (
+            ash if ash.get("enabled", False) else self.transforms.get("detect_sharding", {})
+        )
+        enable_attention_dp = sharding_config.get("enable_attention_dp", False)
```
govind-ramnarayan, line 507: is this just about exposing the enable_attention_dp flag to the Eagle factory? Is it needed for anything else?

I want to eventually remove the broadcast that is there for Eagle, in which case we may not need to pipe the attention DP flag down like this. I think we're almost at the point where we can experiment and see if we can remove this broadcast, which existed because otherwise I saw instability with our "replicated attention" SuperV3 config, but we have fixed some bugs in MambaCacheManager / Eagle that make me hopeful this instability is no longer there... this issue is about removing that broadcast: http://github.com/orgs/NVIDIA/projects/83/views/2?pane=issue&itemId=176893404&issue=NVIDIA%7CTensorRT-LLM%7C13134)

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
govind-ramnarayan: additionally I'm unclear on what some of these other llm_args setting lines are about. Are they related to the attention DP request balancer or not? if they are maybe a function or method to set them all together would be good. If they are not it would be good to understand what these settings are about.

govind-ramnarayan (commented):
Thanks! Have a few comments mostly about tracking issues and documentation.

jieli-matrix (approved):

tensorrt_llm/_torch/auto_deploy/llm_args.py
MrGeva: @govind-ramnarayan yes, this is the only reason I plumbed it. should I add a TODO in the code to remove it once that bug is fixed?

tensorrt_llm/_torch/auto_deploy/transform/library/sharding.py
MrGeva: I will add a ticket, thanks!

tensorrt_llm/_torch/auto_deploy/shim/ad_executor.py
MrGeva: thanks for catching. I will remove.

tensorrt_llm/_torch/auto_deploy/llm_args.py
MrGeva: added todo with your ticket link

greg-kwasniewski1 (approved):

examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml
```diff
@@ -21,6 +19,7 @@ speculative_config:
 transforms:
   detect_sharding:
     allreduce_strategy: NCCL
+    enable_attention_dp: true
```
galagam, line 29: @MrGeva Please enable attention DP for the non-MTP config as well (in this PR or in a follow up PR)

MrGeva: will do

govind-ramnarayan (approved):
Looks good, thanks!

examples/auto_deploy/model_registry/configs/super_v3_mtp.yaml
galagam: @MrGeva Can we enable now?

MrGeva: yes. I will

tburt-nv:
please rebase or merge origin/main to resolve the GB200-4_GPUs-PyTorch-PerfSanity-* failures with https://github.com/NVIDIA/TensorRT-LLM/commit/91dc1458b5b326c190f0d2a26e9f5bd97af0fb14

fredricz-20070104 (approved):
Approved

fredricz-20070104:
@MrGeva Please resolve the DCO and bolossom CI error, otherwise this one can't be merged.

Issue #14942, opened 2026-06-04
[Bug]: Gemma 3 / Gemma 4 fail in TensorRT-LLM 1.3.0rc17 container on NVIDIA L4: FlashInfer unsupported architecture, kv_layout metadata error, AutoDeploy use_cache mismatch


ylnhari:
## System Info

- CPU architecture:

  ```text
  x86_64
  ```

- CPU / Host memory:

  ```text
                 total        used        free      shared  buff/cache   available
  Mem:            15Gi       872Mi       4.0Gi       0.0Ki        10Gi        14Gi
  Swap:             0B          0B          0B
  ```

- GPU:

  ```text
  NVIDIA L4
  ```

- GPU compute capability:

  ```text
  8.9
  ```

- GPU memory:

  ```text
  TensorRT-LLM logs report:
  GPU 0 has a total capacity of 21.95 GiB
  ```

- GPU clocks:

  ```text
  Driver Version: 550.90.07
  CUDA Version: 12.4

  Graphics Clock: 1020 MHz
  SM Clock:       1020 MHz
  Memory Clock:   6250 MHz
  Video Clock:    945 MHz

  Max Graphics Clock: 2040 MHz
  Max SM Clock:       2040 MHz
  Max Memory Clock:   6251 MHz
  Max Video Clock:    1770 MHz
  ```

- Container image:

  ```text
  nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc17
  ```

- TensorRT-LLM version from logs:

  ```text
  [TensorRT-LLM] TensorRT LLM version: 1.3.0rc17
  ```

- Python:

  ```text
  Python 3.12 inside container
  ```

- PyTorch version from logs:

  ```text
  torch version 2.11.0a0+eb65b36914.nv26.02
  ```

- NVIDIA driver:

  ```text
  550.90.07
  ```

- CUDA reported by nvidia-smi:

  ```text
  CUDA Version: 12.4
  ```

- Transformers version from warning:

  ```text
  transformers version 5.5.4
  ```

- ModelOpt warning from container:

  ```text
  /usr/local/lib/python3.12/dist-packages/modelopt/torch/__init__.py:36:
  UserWarning: transformers version 5.5.4 is incompatible with nvidia-modelopt and may cause issues.
  Please install recommended version with `pip install nvidia-modelopt[hf]` if working with HF models.
  ```

- torchao warning from container:

  ```text
[...]

PR #15064, opened 2026-06-07
[None][fix] Guard Gemma4 backend on unsupported GPUs and fix Gemma3 AutoDeploy use_cache

ssam18:
This picks off two of the Gemma failures reported in #14942 on an NVIDIA L4. Gemma 4 used to default to the FlashInfer trtllm-gen path on every GPU and then crash deep in the kernel with an unhelpful Unsupported architecture message, so it now reads the SM version first and raises a clear error on anything outside the SM100 family. Gemma 3 on the AutoDeploy backend used to die during model construction because use_cache was forwarded into a constructor that does not accept it, so leftover config-only kwargs are now stripped before the model is built. I added unit tests covering both changes and kept the kv layout report and the E4B out of memory case out of scope since they look resolved or expected on current main.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model loading robustness to handle varying HuggingFace model versions gracefully.
  * Added GPU architecture validation for Gemma4 models to provide clear error messages when unsupported hardware is detected.

* **Tests**
  * Added comprehensive test coverage for model construction and hardware compatibility checks.


ssam18:
Quick note on scope, I focused on the two failures that are clear code bugs and skipped the kv layout error since it already looks fixed on current main, along with the E4B out of memory case since it reads like a genuine memory limit on L4. Happy to tweak the error wording or split this into two commits if that is easier to review. I could not run the full suite locally so the new tests are written to run inside the container.

Wanli-Jiang (approved):

PR #15098, opened 2026-06-08
[TRTLLM-13319][fix] fix output distribution correctness for Eagle3 dynamic-tree rejection sampling

zhaoyangwang-nvidia:
## Summary by CodeRabbit

# Release Notes

  * Optimized speculative decoding verification path with simplified computation logic, reducing unnecessary intermediate steps and memory overhead.
  * Added compilation optimization for probability calculations to improve inference throughput.

* **Refactor**
  * Streamlined internal verification interface for speculative decoding, removing intermediate computation stages while maintaining accuracy.

## Summary

This PR fixes the probability distribution used during verification in Eagle3 one-model dynamic tree rejection sampling, and includes a refactor and a new feature.

## Changes

- **Fix**: Correct probability distribution in dynamic tree rejection sampling — previously using incorrect logits slice for verification
- **Refactor**: Pass `draftTokens[num_gens, N-1]` directly to rejection kernel, removing redundant reshaping
- **Feat**: Support `TLLM_SPEC_DECODE_FORCE_NUM_ACCEPTED_TOKENS` for both dynamic tree and linear rejection sampling paths. Previously unsupported because rejection sampling processes tokens sequentially; when the forced length exceeds actual accepted length, unfilled slots contain out-of-vocabulary garbage. Fix: pre-fill those positions with draft tokens when force mode is active.

## Test Coverage

> TRT-LLM numbers reflect batch-full periods only (all concurrency slots occupied), so they appear higher than system-level throughput.

| Model | Framework | greedy tok/s | rejection tok/s | AL (greedy) | AL (rejection) | AR (greedy) | AR (rejection) |
|---|---|---|---|---|---|---|---|
| Qwen3-8B | SGLang | 2060| 1998 | 3.23 | 3.18 | - | - |
| Qwen3-8B | TRT-LLM | 2977.2 | 2806.6 | 3.327 | 3.330 | 0.776 | 0.777 |
| Qwen3-235B-FP8 | SGLang | 977| 1022 | 2.33 | 2.34 | - | - |
| Qwen3-235B-A22B | TRT-LLM | 1344.7 | 1332.3 | 2.713 | 2.709 | 0.571 | 0.570 |
| Llama-3.3-70B | SGLang | 1381| 1354 | 2.33 | 2.31 | - | - |
| Llama-3.3-70B | TRT-LLM | 1168.3 | 1118.6 | 3.747 | 3.708 | 0.458 | 0.451 |

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
+++ b/tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
@@ -1247,63 +1247,10 @@ def _register_fake():
           skip_temperature: bool = False) -> torch.Tensor:
         return logits.new_empty(list(logits.shape), dtype=torch.float32)
 
-    @torch.library.register_fake("trtllm::build_draft_prob_indices_out_op")
-    def _(topkScoreIndices: torch.Tensor, draftProbIndices: torch.Tensor,
-          topK: int, numDraftTokens: int) -> None:
-        return None
-
     @torch.library.register_fake("trtllm::verify_dynamic_tree_rejection_out_op")
-    def _(candidates: torch.Tensor, draftProbs: torch.Tensor,
-          targetProbs: torch.Tensor, targetSupportIndices: torch.Tensor,
-          targetSupportLengths: torch.Tensor, draftProbIndices: torch.Tensor,
+    def _(candidates: torch.Tensor, targetProbs: torch.Tensor,
           retrieveNextToken: torch.Tensor, retrieveNextSibling: torch.Tensor,
           treeValid: torch.Tensor, acceptIndex: torch.Tensor,
           acceptTokenNum: torch.Tensor, acceptToken: torch.Tensor,
           numSpecStep: int, seed: torch.Tensor, offset: torch.Tensor) -> None:
         return None
-
-    @torch.library.register_fake(
-        "trtllm::compute_draft_probs_for_dynamic_tree_rejection_op")
-    def _(draftLogits: torch.Tensor,
-          temperatures: torch.Tensor,
-          numDraftProbRows: int,
-          targetVocabSize: int,
-          top_k: Optional[torch.Tensor] = None,
-          top_p: Optional[torch.Tensor] = None,
-          skip_temperature: bool = False,
-          d2t: Optional[torch.Tensor] = None,
-          top_k_max: int = 0,
-          skip_all_sampling_params: bool = False) -> torch.Tensor:
-        batch_size = temperatures.shape[0]
-        return draftLogits.new_empty(
-            (batch_size, numDraftProbRows, targetVocabSize),
-            dtype=torch.float32)
-
-    @torch.library.register_fake(
-        "trtllm::compute_target_probs_for_dynamic_tree_rejection_op")
-    def _(
-        targetLogits: torch.Tensor,
-        temperatures: torch.Tensor,
-        numDraftTokens: int,
-        top_k: Optional[torch.Tensor] = None,
-        top_p: Optional[torch.Tensor] = None,
-        skip_temperature: bool = False,
-        top_k_max: int = 0,
-        skip_all_sampling_params: bool = False
-    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
-        batch_size = temperatures.shape[0]
-        target_vocab_size = targetLogits.shape[-1]
-        target_probs = targetLogits.new_empty(
-            (batch_size, numDraftTokens, target_vocab_size),
-            dtype=torch.float32)
-        has_filtering = (top_k is not None) or (top_p is not None)
-        if skip_all_sampling_params or not has_filtering:
-            support_indices = targetLogits.new_empty((0, ), dtype=torch.int32)
-            support_lengths = targetLogits.new_empty((0, ), dtype=torch.int32)
-        else:
-            support_dim = top_k_max if top_k_max > 0 else target_vocab_size
-            support_indices = targetLogits.new_empty(
-                (batch_size, numDraftTokens, support_dim), dtype=torch.int32)
-            support_lengths = targetLogits.new_empty(
-                (batch_size, numDraftTokens), dtype=torch.int32)
-        return target_probs, support_indices, support_lengths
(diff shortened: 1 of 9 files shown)


zhaoyangwang-nvidia:
Hi @NVIDIA/trt-llm-torch-graph-compiler @NVIDIA/trt-llm-torch-spec-decoding please help to review this PR, thanks a lot~

tensorrt_llm/_torch/speculative/eagle3_dynamic_tree.py
```diff
@@ -834,115 +820,145 @@ def _sample_and_accept_dynamic_tree(
...
                 )
+                num_accepted_tokens[num_contexts:batch_size] = 1
+                accepted_tokens[num_contexts:batch_size, 0] = gen_root_tokens
+                self._accepted_draft_indices_tensor[num_contexts:batch_size] = -1
                 return accepted_tokens, num_accepted_tokens
 
+            target_logits_tree = logits[-num_gens * N :].reshape(-1, vocab_size)
+            gen_slice = slice(num_contexts, num_contexts + num_gens)
+            temps = spec_metadata.request_temperatures[gen_slice]
+            top_ks_rej = spec_metadata.request_top_ks[gen_slice]
```
mikeiovine, line 914: Do you have to clamp these like on the context path? I think things can go wrong if they're set to 0 (e.g. topk disabled)

tensorrt_llm/_torch/speculative/one_model_sampler.py
```diff
@@ -117,6 +117,7 @@ def sampling_batch_spec_dec_one_model(
     return random_sampled
 
 
+@torch.compile(options={"max-autotune": True})
```
mikeiovine, line 135: Can you confirm that this doesn't cause issue during warmup for the existing linear chain path? torch.compile during stream capture can fail

zhaoyangwang-nvidia: Confirmed empirically on H200 with Qwen3-8B Eagle3 (one-model linear chain, rejection sampling on, CUDA graph on, non-greedy temp/top_k/top_p). Warmup + capture complete cleanly (exit 0, full throughput results); the dedicated advanced sampling capture pass exercises compute_probs_from_logits for all 4 captured batch sizes with no stream is capturing / inductor / dynamo errors. The capture path runs WARMUP_STEPS eager iterations of the exact callable before torch.cuda.graph(), so max-autotune compilation/autotuning happens outside the capture region. No capture failure.

tensorrt_llm/_torch/speculative/eagle3_dynamic_tree.py
zhaoyangwang-nvidia: Good catch — this turned out to be worth digging into. Here's what we changed and why:

1. Moved the clamp into the sampling helpers instead of the call site. The top-k range requirement really belongs to the functions that call flashinfer (compute_probs_from_logits / sampling_batch_spec_dec_one_model), not to each caller. Keeping it at the call site is exactly how this bug slipped in — the context path clamped but the rejection path didn't. Centralizing it in a shared sanitize_top_k() means every current and future caller is protected automatically and we can't forget it again.

2. Upper bound is now vocab_size, not vocab_size - 1. Checking the flashinfer kernel (topk.cuh), top_k_mask_logits only takes the "no masking, just copy" fast path when k >= vocab_size. With vocab_size - 1 it still runs a full radix top-k selection (just to drop the single lowest-logit token), which is wasted work for a request that wants no top-k filtering. Clamping to vocab_size lets the disabled case hit the skip path. I verified k = vocab_size is safe on all three backends (flashinfer mask + fused sampling op, the C++ op, and the PyTorch-native path all keep-all without error).

3. Disabled top-k must map to vocab_size, not 1. Per SamplingParams, top_k = 0 (and the effective default when sampling is on) means "all logits" — i.e. top-k disabled. The problem is our three operators don't agree on a literal 0: the C++ op treats 0 as keep-all, but flashinfer masks the entire row (all-zero probs) and the PyTorch-native path gathers out of bounds. So clamping 0 → 1 would be wrong (that's greedy, the opposite of "all logits"). The new normalization mirrors the C++ op — where(k > 0, k, vocab_size) — so 0 correctly becomes keep-all across all backends.

4. Scope. This fixes the top-k inconsistency for now. I suspect other sampling params (top-p, temperature) may have similar "param definition vs. operator behavior" mismatches, so I'll audit the full set and fix them properly as part of the upcoming sampler refactor rather than patching each one here.

tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
```diff
@@ -1247,63 +1247,10 @@ def _(logits: torch.Tensor,
...
-    @torch.library.register_fake("trtllm::build_draft_prob_indices_out_op")
-    def _(topkScoreIndices: torch.Tensor, draftProbIndices: torch.Tensor,
-          topK: int, numDraftTokens: int) -> None:
-        return None
-
     @torch.library.register_fake("trtllm::verify_dynamic_tree_rejection_out_op")
-    def _(candidates: torch.Tensor, draftProbs: torch.Tensor,
-          targetProbs: torch.Tensor, targetSupportIndices: torch.Tensor,
-          targetSupportLengths: torch.Tensor, draftProbIndices: torch.Tensor,
+    def _(draftTokens: torch.Tensor, targetProbs: torch.Tensor,
```
liji-nv: This fake impl is unnecessary. As it is an inplace op. Better adding to inplace_info() in tensorrt_llm/_torch/compilation/utils.py

liji-nv (approved):

tensorrt_llm/_torch/custom_ops/cpp_custom_ops.py
zhaoyangwang-nvidia: Fixed

zhaoyangwang-nvidia:
Hi @mikeiovine , I’ve addressed all the comments you left earlier and made the corresponding changes. Could you please take another look when you have a chance? Thanks!

mikeiovine (approved):

PR #15106, opened 2026-06-08
[None][feat] Add BaseResourceManager-based KV-cache compression manager framework

Hudayday:
## Description

  Adds a standalone **KV-cache compression** framework: a `BaseResourceManager`
  base class that lets a KV-reduction algorithm (e.g. periodic token eviction /
  KV compaction) run alongside the normal KV cache manager, driven entirely by
  PyExecutor's existing resource-manager loop. **No PyExecutor changes, no
  attention-backend changes**; fully dormant unless a compression config is set.

  **Framework only — no concrete algorithm ships in this PR.**

  ### Design
  KV-cache compression changes *which* KV is stored, not the attention
  computation — so it is a resource manager, kept entirely separate from
  `SparseAttentionConfig` and the attention backend.

  - **`BaseKVCacheCompressionManager(BaseResourceManager)`** (in `resource_manager.py`)
    — four KV-cache lifecycle hooks, all default no-op:
    - `on_request_init` — per-request setup (e.g. allocate scoring buffers)
    - `on_context_step_end` — end of prefill (one-shot prefill-end eviction)
    - `on_generation_step_end` — each decode step (periodic / budget eviction)
    - `on_request_finish` — release per-request state

    Because it *is* a `BaseResourceManager`, PyExecutor's main loop already drives
    it: `prepare_resources` / `update_resources` / `free_resources` translate into
    the four hooks, gated on the same signals the peer resource managers use
    (`is_first_context_chunk`, `context_requests_last_chunk`) — no manager-side
    request bookkeeping. It holds a `KVCacheManagerV2` as a tool and never inherits
    from it (the cache manager owns the physical KV). `__init__` refuses KV-cache
    block reuse (a method that rewrites stored K/V can't share prefix blocks — the
    same guard `RocketKVCacheManager` makes); `get_max/needed_resource_to_completion`
    return 0 (it owns no physical resource, so it never gates the scheduler).
  - **`create_kv_cache_compression_manager(config, kv_cache_manager)`** — factory
    dispatched from LLM init; framework-only, so it warns + returns None for any
    algorithm (concrete methods add a dispatch branch in a follow-up PR).
  - **`KvCacheCompressionConfig`** — a top-level `LlmArgs` field, separate from
    `SparseAttentionConfig`.

### Wiring
  `create_py_executor` (`_util.py`) builds the manager from
  `kv_cache_compression_config` when set and registers it in the resource-manager
  registry (`ResourceManagerType.KV_CACHE_COMPRESSION_MANAGER`) the same way the
  KV cache manager is registered; it runs after the cache manager so it reconciles
  once the cache is resized.

  ### Scope
  - Framework only — no concrete algorithm ships in this PR.
  - Touches **no** attention-backend or sparse-attention code: a compression
    method drives standard attention over the (physically compacted) KV; it is not
    sparse attention.
  - Fully dormant when no compression config is set — zero behavior change for
    existing inference.

  ### Changes per file
  | File | Change |
  |---|---|
  | `_torch/pyexecutor/resource_manager.py` | `BaseKVCacheCompressionManager(BaseResourceManager)` (4 lifecycle hooks + RM-API→hook translation + zero resource counts + block-reuse guard) + `create_kv_cache_compression_manager` factory + `ResourceManagerType.KV_CACHE_COMPRESSION_MANAGER`. |
  | `_torch/pyexecutor/_util.py` | Build + register the compression manager from `kv_cache_compression_config`, before warmup. |
  | `llmapi/llm_args.py` | `KvCacheCompressionConfig` + `LlmArgs.kv_cache_compression_config`. |
  | `tests/unittest/_torch/pyexecutor/test_kv_cache_compression_manager.py` | Unit tests (below). |

  ## Test Coverage
  `tests/unittest/_torch/pyexecutor/test_kv_cache_compression_manager.py`:
  - `BaseResourceManager` inheritance; four hooks default no-op + accept extra
    kwargs; zero resource counts.
  - RM-API → hook translation: `prepare_resources` fires `on_request_init` on the
    first prefill chunk only; `update_resources` fires `on_context_step_end` for
    `context_requests_last_chunk` + exactly one `on_generation_step_end` per
    iteration; `free_resources` fires `on_request_finish`.
  - Factory returns None + warns for an unregistered algorithm.
  - Block-reuse guard raises; canonical names live in `resource_manager`, not in
    the sparse module.

  ## PR Checklist
  - [x] New functionality is covered by unit tests.
  - [x] No change to existing inference paths when no compression method is configured.


tensorrt_llm/_torch/attention_backend/sparse/__init__.py
```diff
@@ -1,9 +1,24 @@
-from .utils import (get_flashinfer_sparse_attn_attention_backend,
+# Concrete manager subclasses are not re-exported at package level — they
```
heyuhhh: Could you please remove these comments? Eespecially the domain-specific comments. We can have some concise words to add here if you want

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
+"""L2 behavior layer for KV cache compression methods.
```
heyuhhh: Could you please refactor the comments? Some refactor advice:
* For the head comments, we can just add breif sentences to describe what this file does. Also, because this is a new created file, we can add copy license here
* For the class we can introduce more. However, please remove the domain-specific words which happened in the design or discussions, many other users can't understant that easily. We should try to let everyone who uses that understands
* Besides, we can add some comments in the methods to balance, to avoid too many comments in the head of class for clean

This is common rule which can be applied for other place.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+if TYPE_CHECKING:
+    from tensorrt_llm._torch.attention_backend.interface import AttentionMetadata
+    from tensorrt_llm._torch.pyexecutor.llm_request import LlmRequest
+    from tensorrt_llm._torch.pyexecutor.resource_manager import KVCacheManagerV2
+    from tensorrt_llm._torch.pyexecutor.scheduler.scheduler import ScheduledRequests
+
+
+# ``(indices, offsets)`` tuple consumed by the attention kernel as an input-
+# side sparse mask; ``None`` falls back to dense attention.
+SparseAttentionIndices = Tuple[torch.Tensor, torch.Tensor]
```
heyuhhh: I don't think we need this. For current stage we don't need to consider sparse x compression

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+    # ------------------------------------------------------------------ #
+    # Class-level metadata                                                #
+    # ------------------------------------------------------------------ #
+
+    # Whether this manager class is compatible with prefix / block KV
+    # cache reuse (radix-tree / APC). Default conservative ``False``;
+    # subclasses opt in explicitly. The LLM init factory may enforce a mutex:
+    # combining a manager with ``supports_kv_cache_reuse=False`` and
+    # ``KvCacheConfig.enable_block_reuse=True`` raises a config error at init.
+    supports_kv_cache_reuse: ClassVar[bool] = False
```
heyuhhh: Do we need this standalone variable? Can we introduce this from `KvCacheConfig` if we need this?

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Understood, thank you

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+
+    # Pattern 3 (V2 subclass) escape hatch: subclass may declare a specific
+    # ``KVCacheManagerV2`` subclass type via this ClassVar. PyExecutor factory
+    # uses this to instantiate the right V2 type. ``None`` (default) means use
+    # plain ``KVCacheManagerV2`` — adequate for Patterns 1 and 2.
+    kv_cache_manager_class: ClassVar[Optional[Type["KVCacheManagerV2"]]] = None
```
heyuhhh: I think we can remove this for simplify

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+                f"the right V2 instance."
+            )
+        self.kv_cache_manager = kv_cache_manager
+
+    # ================================================================== #
+    # Semantic lifecycle hooks (8 total, temporal order).                #
+    # Subclasses override what they need; all default to no-op.          #
+    # ================================================================== #
+
+    def on_request_init(self, request: "LlmRequest") -> None:
```
heyuhhh: Add `**kwargs` for compatiability. Same for others

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+        pass
+
+    def on_context_attention(
+        self,
+        layer_idx: int,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        attn_scores: Optional[torch.Tensor],
+        metadata: "AttentionMetadata",
+    ) -> Optional[SparseAttentionIndices]:
```
heyuhhh: We can remove the return indicator because we don't have a clear return design here. Same for others if it's not sure.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+        layer otherwise. ``attn_scores`` is populated only when the
+        attention kernel instantiation exposes scores (compile-time
+        template flag); ``None`` when scores are not materialized.
+
+        Sparse-mask managers (e.g. RocketKV) return an ``(indices, offsets)``
+        tuple as an input-side sparse mask; physical-evict and storage
+        managers return ``None``. Called directly from
+        ``TrtllmAttention.forward`` via ``metadata.compression_manager``.
+        """
+        return None
```
heyuhhh: Change to `pass`

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+
+        Side-effect only (returns ``None``): the sparse mask, if any, was
+        already applied. Use this when the algorithm conceptually runs *after*
+        attention — e.g. stash per-layer ``q``/``k``/``attn_output`` so a
+        unified eviction can be computed in :meth:`on_context_end`,
+        rather than per-layer during attention.
+        """
+        pass
+
+    def on_context_end(
```
heyuhhh: --> `on_context_step_end`

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+
+    def on_request_init(self, request: "LlmRequest") -> None:
+        """Per-request init hook.
+
+        Override to allocate per-request accumulators (e.g. per-request
+        scoring buffers).
+        """
+        pass
+
+    def on_context_attention(
```
heyuhhh: Actually, we can keep these computation methods, not related to runtime lifecycle, beside of the compression class. Let's keep that in mind for later refactor.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,379 @@
...
+        base_method = getattr(BaseKVCacheCompressionManager, hook_name, None)
+        if base_method is None:
+            return False
+        # In Python 3, accessing a method via class returns the function directly.
+        # MRO lookup means an inherited (non-overridden) hook returns the SAME
+        # function object as the base, so identity check suffices.
+        return own_method is not base_method
+
+
+class SparseAttentionManager(BaseKVCacheCompressionManager):
```
heyuhhh: Seems that we can remove these two class for now

Hudayday: Agree

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -457,6 +457,22 @@ def needs_separate_short_long_cuda_graphs(self) -> bool:
         """
         return False
 
+    @property
+    def is_behavior_layer_method(self) -> bool:
```
heyuhhh: It seems that we don't need this for the compression framework. If one sparse attention algorithm is per-layer computation, we can implement it in attention backend, or we can integrate the compression base class and to implement the `on_context_attention_begin/end` method to register. We have no need to add this flag. If anything i misunderstanding, please correct me.

tensorrt_llm/_torch/pyexecutor/_util.py
heyuhhh, line 2037: Why not register KVCacheCompressionManager at here? It's no need to register from `model_engine`

tensorrt_llm/llmapi/llm_args.py
Hudayday: The design idea of `is_behavior_layer_method` is some methods may not need special Attention core or special KV Cache manager. Therefore, with using this flag, `sparse/utils.py` know that it doesn't need to assign special Cache Manager / Attention to these methods.

By using this flag, lots of behaviors don't need to modify the `sparse/utils.py`.

tensorrt_llm/_torch/attention_backend/sparse/__init__.py
Hudayday: Removed.

tensorrt_llm/llmapi/llm_args.py
heyuhhh: Sorry, i'm still a little confused about here. The compress manager still supports Behavior-layer methods, right? For specific algorithm we just implement specific compress manager and it's okay to cover, so there is no need to add `is_behavior_layer_method`

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Romoved. It was added to support RocketKV x compression.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: I think we need it if some algorithms claim that it supporrt reuse. I modify it to `__init__` following what `rocket.py` does, so it now raises if `enable_block_reuse` from `KvCacheConfig` is on and the method can't support it.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Agree, deleted.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Agree, added for all hooks

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Renamed, thanks

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Done

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Done

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Yes, I will take a look whether the refactor is necessary for future methods.

tensorrt_llm/llmapi/llm_args.py
Hudayday: Yes, your comment make sense. Removed.

tensorrt_llm/_torch/pyexecutor/_util.py
Hudayday: Agree, moved it into the resources dict (with the other managers) instead of mutating `resource_manager.resource_managers` after construction. 

However, the `model_engine.compression_manager = mgr` line isn't a registration, instead it is the reference the attention-metadata builder needs to fire the 4 per-layer hooks (`model_engine` only gets the resource manager as a method arg, so it can't look it up there), so that one stays.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
heyuhhh: So can we just use `enable_block_reuse` to flag? They have totally the same semantics

tensorrt_llm/_torch/attention_backend/sparse/utils.py
```diff
@@ -1,33 +1,76 @@
...
+
+    Returns one :class:`BaseKVCacheCompressionManager`. It inherits
+    :class:`BaseResourceManager`, so PyExecutor's main loop auto-drives its
+    lifecycle hooks once it is added to the resource-manager registry.
+    Multi-method stacking (composing several axes) is future work and
+    intentionally unsupported here.
+    """
+    if sparse_attn_config is None:
+        return None
+    return create_sparse_attention_manager(sparse_attn_config, kv_cache_manager)
```
heyuhhh: We have no need to add this router to avoid introduce some ambiguity

tensorrt_llm/_torch/attention_backend/interface.py
```diff
@@ -70,6 +71,11 @@ class AttentionMetadata:
     kv_cache_manager: Union[KVCacheManager, KVCacheManagerV2, None] = None
     # Draft KV cache manager for one-model speculative decoding with separate KV cache layouts
     draft_kv_cache_manager: Union[KVCacheManager, KVCacheManagerV2, None] = None
+    # When set, ``TrtllmAttention.forward`` dispatches the per-attention
+    # sparse hooks (``on_context_attention`` / ``on_generation_attention``)
+    # to this compression manager. Its other lifecycle hooks fire from
+    # PyExecutor's BaseResourceManager callbacks and do not use this field.
```
heyuhhh: Could you please simplify the comments? The interface is an entrance of attention backend we should keep the code clean

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
```diff
@@ -0,0 +1,273 @@
...
+        layer_idx: int,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        attn_scores: Optional[torch.Tensor],
+        metadata: "AttentionMetadata",
+        **kwargs,
+    ) -> None:
+        """Per-layer hook after every generation-phase attention forward.
+
+        Side-effect only. Called directly from ``TrtllmAttention.forward``
```
heyuhhh: Could you please double check the side-effects of all the hooks to make sure if they are necessary to keep? We should try to avoid the subjective analyse in the code.

tensorrt_llm/_torch/attention_backend/sparse/utils.py
```diff
@@ -1,33 +1,76 @@
...
     if sparse_attn_config.algorithm == "rocket":
         return RocketKVCacheManager
     elif sparse_attn_config.algorithm == "dsa":
         return DSACacheManager
     elif sparse_attn_config.algorithm == "skip_softmax":
         return KVCacheManager
-    else:
-        raise ValueError(
-            f"Unsupported sparse attention algorithm: {sparse_attn_config.algorithm}"
-        )
```
heyuhhh, line 33: Maybe we can add a warning here to notice users. If someone implement a new sparse attention algorithm but forget to register here, a warning may raise attention and help. Same for others.

tensorrt_llm/_torch/attention_backend/trtllm.py
```diff
@@ -1768,6 +1782,19 @@ def forward(
 
         self._run(q, k, v, metadata, forward_args)
 
+        # Post-attention hooks, fired after the output is computed, for a
+        # method that needs the attention result (e.g. to stash per-layer
+        # q/k/output for a later eviction). Side-effect only; both phases fire
+        # and the manager decides what to act on.
```
heyuhhh: How to understand this `side-effect`?

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -90,8 +89,15 @@ def get_kv_cache_manager_cls(
...
-    elif is_hybrid_linear(config):
+        # The legacy methods (rocket / dsa / skip_softmax) own their own
+        # cache-manager class; use it. A compression-framework method returns
+        # None here, so we fall through to the standard manager (its algorithm
+        # runs in the compression manager registered via
+        # create_sparse_attention_manager after PyExecutor instantiation).
+        cls = get_sparse_attn_kv_cache_manager(sparse_attn_config)
+        if cls is not None:
+            return cls
+    if is_hybrid_linear(config):
```
heyuhhh: Why change `elif -> if` ? Seems that we can also remove the comments here

tensorrt_llm/_torch/pyexecutor/_util.py
```diff
@@ -1644,13 +1657,32 @@ def create_py_executor_instance(
...
     # depend on the results of other resource managers.
     if kv_cache_manager is not None:
         resource_manager.resource_managers.move_to_end(
             ResourceManagerType.KV_CACHE_MANAGER, last=True)
+    # The compression manager runs after the cache manager: it reconciles the
+    # request's history length once the cache manager has resized it.
+    if (ResourceManagerType.KV_CACHE_COMPRESSION_MANAGER
+            in resource_manager.resource_managers):
+        resource_manager.resource_managers.move_to_end(
+            ResourceManagerType.KV_CACHE_COMPRESSION_MANAGER, last=True)
```
heyuhhh, line 2063: Seems that move to last makes sense because we can have compression operation after update the resources. However, please pay attention here about the orders. I'm concerned that some methods or operations may not follow this.

tensorrt_llm/_torch/pyexecutor/_util.py
heyuhhh: Why not pass `compressionManager` just like `kvCacheManager`? I think it's a little strange here. We can follow the current passing pattern.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Dropped the `supports_kv_cache_reuse` ClassVar, instead, block-reuse is now disallowed directly via `enable_block_reuse`. A future reuse-safe method can relax this in its own `__init__`.

tensorrt_llm/_torch/pyexecutor/_util.py
Hudayday: Understand, fixed

tensorrt_llm/_torch/attention_backend/sparse/utils.py
Hudayday: Agree, removed.

tensorrt_llm/_torch/attention_backend/interface.py
Hudayday: Fixed.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Oops, the term 'no side-effect' just simply means it returns `None`. Removed and polished all comments.

tensorrt_llm/_torch/attention_backend/sparse/utils.py
Hudayday: Agree, fixed.

tensorrt_llm/_torch/attention_backend/trtllm.py
Hudayday: As mentioned above, it means the function returns `None`, instead of return a KV cache mask for Attention. Fixed.

tensorrt_llm/_torch/pyexecutor/_util.py
Hudayday: In current design, methods like 'TriAttention' don't have a special kv cache manager, so it can defer to a standard kv cache manager. Update the comments to clarify this design.

tensorrt_llm/_torch/pyexecutor/_util.py
Hudayday: Thanks for pointing this out. From the TriAttention work: `KVCacheManagerV2.update_resources` resets each request's length back to its full value every step, so registering the compression manager last, after the cache manager, is what makes the ordering correct here, since it then has the final say on the request length. I'll keep this constraint in mind for other methods

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1629,10 +1629,11 @@ def _update_draft_inference_state_for_warmup(
...
     def _set_up_attn_metadata(
-        self,
-        kv_cache_manager: Union[KVCacheManager, KVCacheManagerV2],
-        draft_kv_cache_manager: Optional[Union[KVCacheManager,
-                                               KVCacheManagerV2]] = None):
+            self,
+            kv_cache_manager: Union[KVCacheManager, KVCacheManagerV2],
+            draft_kv_cache_manager: Optional[Union[KVCacheManager,
+                                                   KVCacheManagerV2]] = None,
+            compression_manager=None):
```
heyuhhh: Please follow `draft_kv_cache_manager`: `Optional[BaseKVCacheCompressionManager] = None`

Hudayday: Understand, fixed.

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
lfr-0531, line 1: Does it make sense to move this `BaseKVCacheCompressionManager` to `resource_manager.py`?

lfr-0531:
Per an offline sync with @Hudayday , there'll be another round of refactoring. Thanks, @Hudayday !

tensorrt_llm/_torch/attention_backend/sparse/kv_cache_compression_manager.py
Hudayday: Refactored this PR.

heyuhhh (approved):
Thanks for the efforts! I have no other concerns about this PR

heyuhhh:
Add this PR: https://github.com/NVIDIA/TensorRT-LLM/pull/14687 here for visibility. Please take a look about this PR's change. Now for sparse attention there is a lowered params used by attention. But for now, the compression manager lives at executer runtime which may not be related about the sparse changes. If the later algorithm integration needs, it would be great to follow this pattern. Just a kindly reminder here.

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -2708,6 +2708,22 @@ def supports_backend(self, backend: str) -> bool:
...
+class KvCacheCompressionConfig(StrictBaseModel):
+    """Config for KV-cache compression: a compression manager runs a KV-reduction
+    algorithm (e.g. periodic token eviction) alongside KVCacheManagerV2.
+
+    Kept separate from SparseAttentionConfig by design -- compression changes
+    which KV is stored, not the attention computation. The manager is registered
+    as a resource manager in create_py_executor (_util.py), like the KV cache
+    manager itself. Concrete algorithms subclass this and add their parameters.
+    """
+    algorithm: str = Field(
```
Superjomn, line 2847: Is there a fixed set of algorithm candidates? If so, a `Literal` is preferred to be more intuitive in both the typing and the auto-generated api reference doc.

Hudayday: Thanks for the suggestion. There isn't a fixed candidate set at the base: this PR ships no algorithms, and concrete methods subclass `KvCacheCompressionConfig` and register their own dispatch branch, so `algorithm` is an open extension point. I think it is fine to keep it as `str` here and have each concrete algorithm subclass narrow it to `Literal["<name>"]` when it lands, which gives the typing and auto-doc benefits you mentioned at that level. Otherwise here is an empry `Literal`

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -2149,6 +2154,148 @@ def _free_blocks(self, block_list: list):
...
+        for req in scheduled_batch.context_requests_last_chunk:
+            self.on_context_step_end(req, attn_metadata)
+        self.on_generation_step_end(scheduled_batch, attn_metadata)
+
+    def free_resources(self, request: "LlmRequest") -> None:
+        """Fire :meth:`on_request_finish`."""
+        self.on_request_finish(request)
+
+
+def create_kv_cache_compression_manager(
```
lfr-0531: Since `_create_kv_cache_manager` is defined in `_util.py`, it makes sense to move `create_kv_cache_compression_manager` there as well for consistency.

Hudayday: Yes, it makes sense. Moved to `_util.py`.

bobboli (approved):

bobboli:
Could you document the design ideas in https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/docs/source/developer-guide/sparse-attention-development-guide.md, after https://github.com/NVIDIA/TensorRT-LLM/pull/14687 is merged?

Hudayday:
> Could you document the design ideas in https://github.com/bobboli/TensorRT-LLM/blob/bli/skip-softmax-decouple-llm-vg/docs/source/developer-guide/sparse-attention-development-guide.md, after #14687 is merged?

This KV-cache compression manager is independent from Sparse attention now. Hence I will prepare a seperate doc for this design.

lfr-0531 (approved):

DomBrown (approved):
LGTM for the API stability part.

thorjohnsen (approved):
LGTM

tensorrt_llm/_torch/pyexecutor/resource_manager.py
```diff
@@ -2244,6 +2247,129 @@ def _free_blocks(self, block_list: list):
...
+
+    Concrete compression methods subclass this directly. All 4 hooks default to
+    no-op; subclasses override what they need. The manager never inherits from
+    any cache manager because this layer decides *how* the physical KV is used,
+    not *what* physical KV exists. Subclasses hold ``KVCacheManagerV2`` as a tool.
+    """
+
+    def __init__(self, kv_cache_manager: "KVCacheManagerV2"):
+        self.kv_cache_manager = kv_cache_manager
+        # Compression evicts/rewrites stored keys and values, so a shared prefix
```
lowsfer, line 2272: Is this because for the tail incomplete compression block, every new token may trigger re-compression that may update the whole compressed block? In that case, I think I can enhance kv cache manager v2 with reuse-constraint support, so user can control granularity of reuse, e.g. reuse only at multiples of 128 for 128-token compression.

lowsfer: This doesn't need to be part of this PR, though. Just want to understand what is needed to enable reuse in the future.

Hudayday: You are right that the concern is general, not just the trailing block: an eviction pass can rewrite stored KV across blocks.

It also depends on the method. TriAttention, the one we are adding, only evicts decode-generated tokens and keeps the prompt prefix intact, so its prefix is reusable in principle. Other methods rewrite the prefix as well (like to RocketKV we already supported), and for those reuse is fundamentally incompatible. So a blanket disable is the right default here: prefix-rewriting methods require it, and for prefill-reusable ones like TriAttention we simply do not need reuse yet.

A reuse-constraint in V2 would be the right place to re-enable it later for the prefix-stable case. Agreed it does not need to be in this PR.

hchings (approved):
LGTM on LLM API part.

PR #15138, opened 2026-06-09
[None][feat] add CuteDSL FP8/FP16 MLA decode attention fmha lib

haow-nv:
## Description

Adds **CuteDSL MLA decode** as an internal **FMHA library** of the `TRTLLM` attention backend (not a separate `attn_backend`). The new `CuteDslMlaFmha` library intercepts the **generation-phase (decode) MLA** portion of a batch and dispatches it to Blackwell CuTe DSL kernels; context-phase requests and anything the library rejects fall through to the next library in the ordered list (`flashinfer_trtllm_gen`, then `fallback`).

### Selection & gating

- Registered as `cute_dsl_mla` in the FMHA library registry. The default order is `cute_dsl_mla,flashinfer_trtllm_gen,fallback`; override with `TLLM_FMHA_LIBS` (exact list, e.g. `TLLM_FMHA_LIBS=flashinfer_trtllm_gen,fallback`, or deltas, e.g. `TLLM_FMHA_LIBS=-cute_dsl_mla`).
- `is_available()` (static): SM100/103 + `nvidia-cutlass-dsl` installed + MLA with `kv_lora_rank=512`, `qk_rope_head_dim=64`, `num_heads <= 128` (DeepSeek geometry).
- Per-forward: the kernel's own `can_implement` check, plus a **perf allowlist** — the library only takes `(num_heads, seq_len_q)` in `{(16, 2), (16, 4), (128, 1)}`, the shapes where CuteDSL measured as an end-to-end win over TRTLLM-Gen in DeepSeek-V3 A/B runs (TP8 + MTP, and attention-DP without MTP). All other shapes fall back to TRTLLM-Gen with a debug log.

### Main pieces

- `attention_backend/fmha/cute_dsl.py`: `CuteDslMlaFmha(PhasedFmha)` — generation-phase hook that prepares latent/rope Q views, paged-KV page table, and per-request cache lengths, then calls the CuteDSL decode op. FP8 KV cache routes to the FP8 kernel; FP16/BF16 activations route to the FP16/BF16 kernel.
- `cute_dsl_kernels/blackwell/attention/mla/`: vendored Blackwell CuTe DSL MLA decode kernels (FP8 and FP16/BF16) + helpers, adapted from the CUTLASS MLA reference and aligned with FlashInfer's `mla_decode` (persistent and split-KV modes, `seq_len_q > 1` causal masking with fold-sq for MTP).
- `custom_ops/cute_dsl_custom_ops.py`: registers `trtllm::cute_dsl_mla_decode_{fp8,fp16}_blackwell` and the kernel runner (JIT compile/cache). `split_kv` and `is_persistent` are **AutoTuner tactic elements** chosen per shape (power-of-two split-KV candidate ladder, batch-bucketed tuning config); `o`/`lse`/`workspace` are mutated in place, and the workspace is sized batch-independently so CUDA-graph capture across batch sizes stays valid.
- `pyexecutor/model_engine.py`: adds one mixed context+generation warmup request so TRTLLM-Gen kernels are JIT-warmed for the mixed batches that `cute_dsl_mla` does not take (it is decode-only).
- `modules/fused_moe/moe_scheduler.py`: attention-DP empty-chunk substitution fix — the substitution now updates every rank's slot in the collective sizes (emptiness is deterministic from the cross-rank-consistent `all_rank_num_tokens_list`), keeping the variable-size all-gather consistent across ranks. Previously each rank only patched its own slot, which deadlocked warmup with imbalanced attention-DP batches.
- `modules/ATTENTION_DEVELOPER_GUIDE.md` and lint-exclusion lists (`.pre-commit-config.yaml`, `legacy-files.txt`, `pyproject.toml`, `ruff-legacy.toml`) updated for the new library and vendored kernel files.

MTP (`seq_len_q > 1`) and CUDA graphs are supported on the CuteDSL path.

## Test Coverage

Validated locally on **B200 (SM100)**; both suites are gated on `get_sm_version() in (100, 103)` and `IS_CUTLASS_DSL_AVAILABLE`.

- `tests/unittest/_torch/attention/test_attention_mla.py` (existing): full MLA module dispatch passes with `cute_dsl_mla` in the default library order.

End-to-end: DeepSeek-V3 FP8 on 8x B200 (TP8/EP8, CUDA graphs on) runs green with and without attention-DP and MTP — no fallbacks, correct outputs, MTP acceptance rate identical to the TRTLLM-Gen baseline; throughput on allowlisted shapes is at parity or better, which is what the perf allowlist encodes.

| # | ADP | seq_len_q | baseline (tok/s) | cutedsl (tok/s) | speedup |
  |---|-----|-----------|------------------|-----------------|---------|
  | 1 | OFF | 1 | 3081.9 | 3082.0 | +0.00% |
  | 2 | OFF | 2 | 4304.5 | 4358.6 | +1.26% |
  | 3 | OFF | 4 | 4771.4 | 4911.6 | +2.94% | 
  | 4 | ON  | 1 | 2676.1 | 2719.8 | +1.63% |
  | 5 | ON  | 2 | 3918.0 | 3919.7 | +0.04% | 
  | 6 | ON  | 4 | 4403.5 | 4458.4 | +1.25% |

## PR Checklist

## GitHub Bot Help

## Dev Engineer Review

- Added a Blackwell-only FMHA backend (`CuteDslMlaFmha`) under the `cute_dsl_mla` registry key and exported it from the FMHA package for dispatch of eligible DeepSeek MLA *generation-phase* decode requests to CuTe DSL kernels (FP8 and FP16/BF16).
- Implemented strict runtime gating/rejection in `tensorrt_llm/_torch/attention_backend/fmha/cute_dsl.py`, including CuTe/SM100/103 availability and tight geometry/mode constraints (decode-only, specific lora/head-dim expectations, head-count limits, and rejection of unsupported speculative/beam/tree/custom mask modes).
- Added CuTe DSL MLA decode custom ops plus an autotuned CUDA kernel runner (`CuteDSLNVMlaDecodeBlackwellRunner`) in `tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py`:
  - enumerates feasible MMA tilers via `can_implement`,
  - autotunes split-K and persistence,
  - validates and slices workspace for LSE/output and optional split-K intermediates,
  - constructs Cute tensors from input page-table/cache/paged KV layouts,
  - includes CUDA-graph-safe caching for FP8 decode scales.
- Added new Blackwell CuTe DSL MLA kernel modules under `tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/`:
  - `mla_decode_fp8.py`, `mla_decode_fp16.py`,
  - `mla_helpers.py` (static tile scheduling helpers for persistent/non-persistent modes),
  - and updated package `__init__.py` exports.
- Updated FMHA documentation to include `cute_dsl_mla` in the default/ordering guidance (`ATTENTION_DEVELOPER_GUIDE.md`).
- Fixed attention-DP empty-chunk collective sizing in `ExternalCommMoEScheduler._forward_multiple_chunks` (`tensorrt_llm/_torch/modules/fused_moe/moe_scheduler.py`) by ensuring all ranks use identical `sizes` vectors when `split_chunk` yields 0-token chunks on some ranks.
- Adjusted TRTLLM-Gen FMHA JIT warmup configuration in `tensorrt_llm/_torch/pyexecutor/model_engine.py` to enable a mixed context+generation warmup only under the specified compatibility conditions.
- Updated lint/config allowlists so the newly added CuTe DSL MLA Python modules are included in pre-commit and Ruff scopes (e.g., `.pre-commit-config.yaml`, `legacy-files.txt`, `pyproject.toml`, `ruff-legacy.toml`).

**Review notes / risk areas**
- Correctness/performance risks concentrate in eligibility/rejection gating, FP8 vs FP16/BF16 dispatch feasibility (`can_implement` + tiler selection), workspace sizing/slicing (including split-K paths), and persistent vs non-persistent static tile scheduling.
- Given the described CI flakiness patterns, failures should be revalidated with the requested multi-GPU CI invocations, paying special attention to CUDA-graph capture/warmup behavior and allowlisted shape routing.

## QA Engineer Review

No test changes.

### DIFF
--- a/tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
+++ b/tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
@@ -238,8 +238,6 @@ class CuteDslMlaFmha(PhasedFmha):
             (16, 8): 16,
             (128, 1): 64,
             (128, 2): 32,
-            (128, 4): 32,
-            (128, 8): 16,
         }
         min_batch = _PERF_MIN_BATCH_FP8.get((num_heads, seq_len_q))
         if min_batch is None:


tensorrt_llm/_torch/pyexecutor/model_engine.py
haow-nv, line 1: Hi @hyukn , I add the gen request in the autotuner warmup here. Could you help to review?

xxi-nv (approved):

tensorrt_llm/_torch/modules/fused_moe/moe_scheduler.py
```diff
@@ -629,15 +629,17 @@ def _forward_multiple_chunks(
             router_logits_list = list(router_logits_list)
             input_ids_list = list(input_ids_list)
             for idx_chunk in range(num_chunks):
-                _x = x_list[idx_chunk]
-                if _x.numel() == 0:
+                # Keep the collective `sizes` consistent across ranks: substitute
```
haow-nv: cc @xxi-nv

tensorrt_llm/_torch/pyexecutor/model_engine.py
```diff
@@ -1475,18 +1486,26 @@ def _run_autotuner_warmup(self, resource_manager: ResourceManager):
             token_num_upper_bound=token_num_upper_bound,
             max_num_draft_tokens=self.original_max_draft_len)
 
+        warmup_configs = [(curr_max_num_tokens, 0)]
+        if (not self.is_draft_model and self.guided_decoder is None
+                and not self.mapping.has_pp()):
+            # Add generation request to warmup the autotuner cache.
+            warmup_configs.append((1 + self.max_total_draft_tokens, 1))
```
yuxianq, line 1585: The context-only case uses `(curr_max_num_tokens, 0)`, to avoid violating `curr_max_num_tokens`, we can use `(curr_max_num_tokens, 1)` here.

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
yuxianq, line 1: make the file name match with registered name, how about using `cute_dsl_mla.py`?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_helpers.py
```diff
@@ -0,0 +1,310 @@
...
+) -> MLAStaticTileScheduler:
+    return MLAStaticTileScheduler(params, blk_coord[0], blk_coord, grid_shape)
+
+
+LOG2_E = 1.4426950408889634074
+# avoid register indexing on array.
+MAX_SPLITS = 256
+
+
+def ceil_div(a: int, b: int) -> int:
```
yuxianq, line 309: Please use the `ceil_div` in tensorrt_llm/math_utils.py instead of re-refining it.

.pre-commit-config.yaml
yuxianq, line 1: For these 5 new files, we can apply new ruff fules for them. We can revert all changes in `.pre-commit-config.yaml`, `legacy-files.txt`, `pyproject.toml` and `ruff-legacy.toml`.

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
```diff
@@ -0,0 +1,572 @@
...
+        if attn.kv_lora_rank is None or attn.kv_lora_rank <= 0:
+            logger.debug("CuTe DSL MLA FMHA is unavailable: kv_lora_rank must be positive.")
+            return False
+        if attn.qk_rope_head_dim is None or attn.qk_rope_head_dim <= 0:
+            logger.debug("CuTe DSL MLA FMHA is unavailable: qk_rope_head_dim must be positive.")
+            return False
+        if attn.qk_nope_head_dim is None or attn.qk_nope_head_dim <= 0:
+            logger.debug("CuTe DSL MLA FMHA is unavailable: qk_nope_head_dim must be positive.")
+            return False
+        if attn.kv_lora_rank != 512 or attn.qk_rope_head_dim != 64:
```
yuxianq, line 78: This is stricter condition, don't need to check `attn.kv_lora_rank <= 0` and `attn.qk_rope_head_dim <= 0` before

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
haow-nv: We need to integrate cutedsl fmha/mla prefill kernel latter. Shall we update the fmha lib name from "CuteDslMlaFmha" to "CuteDslFmha" here?

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
```diff
@@ -0,0 +1,572 @@
...
+    def _is_supported_with_reason(
+        self,
+        q: torch.Tensor,
+        attn: "TrtllmAttention",
+        meta: "TrtllmAttentionMetadata",
+        fwd: AttentionForwardArgs,
+    ) -> tuple[bool, str]:
+        if fwd.attention_input_type != AttentionInputType.generation_only:
+            return False, "CuTe DSL MLA FMHA only supports generation-only attention."
+        if meta.num_contexts != 0 or meta.num_generations <= 0:
```
yuxianq, line 269: This condition seems to be equivalent to `fwd.attention_input_type != AttentionInputType.generation_only`, can we remove it?

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl.py
```diff
@@ -0,0 +1,572 @@
...
+        sparse_kv_indices = fwd.sparse_prediction.sparse_kv_indices
+        sparse_attn_indices = fwd.sparse_prediction.sparse_attn_indices
+        if (
+            (sparse_kv_indices is not None and sparse_kv_indices.numel() > 0)
+            or (sparse_attn_indices is not None and sparse_attn_indices.numel() > 0)
+            or (fwd.topk_indices is not None and fwd.topk_indices.numel() > 0)
+            or meta.num_sparse_topk > 0
+            or attn.sparse_params is not None
+        ):
+            return False, "CuTe DSL MLA FMHA does not support sparse attention."
```
yuxianq: We have already checked `sparse_params` in `is_available`, is it necessary to check sparse attention again here?

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl.py
```diff
@@ -0,0 +1,572 @@
...
+        # an explicit packed mask the kernel cannot express, and an explicit
+        # CUSTOM mask is likewise unsupported -> fall back to TRTLLM for those.
+        if (
+            fwd.attention_mask == CustomAttentionMask.CUSTOM
+            or fwd.attention_mask_data is not None
+            or getattr(meta, "is_spec_dec_tree", False)
+            or getattr(meta, "is_spec_dec_dynamic_tree", False)
+        ):
+            return False, "CuTe DSL MLA FMHA does not support custom/tree speculative masks."
+        if q.shape[0] % meta.num_generations != 0:
```
yuxianq: We have checked gen-only batch before, this condition is true for gen-only batch, should we check it in runtime again?

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl.py
```diff
@@ -0,0 +1,572 @@
...
+            return (
+                False,
+                f"num_tokens ({q.shape[0]}) must be divisible by "
+                f"num_generations ({meta.num_generations}).",
+            )
+        seq_len_q = q.shape[0] // meta.num_generations
+        if seq_len_q < 1:
```
yuxianq: `is_supported` is not for runtime assertion, it is for quick dispatch, we should avoid adding too much obvious conditions here. This condition must be false for correct q shape in gen-only case, can we remove it?

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
yuxianq: If we want to use a single fmha lib to integrate both mla and non-mla kernels, we can use `CuteDslFmha` and `cute_dsl` as registration name.

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl.py
```diff
@@ -0,0 +1,572 @@
...
+        if not supported:
+            logger.debug(f"CuTe DSL MLA FMHA does not support request: {reason}")
+        return supported
+
+    @staticmethod
+    def _is_perf_favorable(
+        num_heads: int,
+        batch_size: int,
+        seq_len_q: int,
+        predicted_tokens_per_seq: int,
```
yuxianq: `predicted_tokens_per_seq` is unused.

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
```diff
@@ -0,0 +1,572 @@
...
+                batch_size,
+                seq_len_q,
+                attn.predicted_tokens_per_seq,
+                self._get_kernel_dtype(attn, q),
+            )
+            if not favorable:
+                return False, reason
+        if meta.kv_cache_block_offsets is None:
+            return False, "Paged KV block offsets are required."
+        if meta.kv_cache_manager is None:
```
yuxianq, line 301: This condition is enough to exclude no-cache case, `kv_cache_block_offsets/pool_mapping` checks are unnecessary.

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl.py
```diff
@@ -0,0 +1,572 @@
...
+            if not favorable:
+                return False, reason
+        if meta.kv_cache_block_offsets is None:
+            return False, "Paged KV block offsets are required."
+        if meta.kv_cache_manager is None:
+            return False, "KV cache manager is required."
+        pool_mapping = meta.host_kv_cache_pool_mapping
+        if pool_mapping is None:
+            return False, "KV cache pool mapping is required."
+        if fwd.latent_cache is None:
```
yuxianq: `latent_cache` is not None for MLA case, which is checked, can we remove this condition?

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl.py
```diff
@@ -0,0 +1,572 @@
...
+            return False, "latent_cache is required."
+        if fwd.output is None:
+            return False, "output is required."
+
+        tokens_per_block = meta.tokens_per_block
+        if tokens_per_block is None:
+            tokens_per_block = getattr(meta.kv_cache_manager, "tokens_per_block", 0)
+        if tokens_per_block <= 1:
```
yuxianq: Don't need to check the correctness of `tokens_per_block` here

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl.py
```diff
@@ -0,0 +1,572 @@
...
+        # The kernel type is the input dtype
+        kernel_dtype = self._get_kernel_dtype(attn, q)
+        if kernel_dtype is None:
+            return (
+                False,
+                f"Unsupported dtype combination: q={q.dtype}, "
+                f"has_fp8_kv_cache={getattr(attn, 'has_fp8_kv_cache', False)}.",
+            )
+        if kernel_dtype == torch.float8_e4m3fn and (
+            fwd.quant_q_buffer is None or fwd.mla_bmm1_scale is None or fwd.mla_bmm2_scale is None
```
yuxianq: MLA module has ensured it, don't need to check here.

.pre-commit-config.yaml
haow-nv: Thanks. Will revert all changes.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp16.py
```diff
@@ -0,0 +1,4499 @@
...
+        L: int,
+        R: int,
+        in_dtype: Type[cutlass.Numeric],
+        out_dtype: Type[cutlass.Numeric],
+        acc_dtype: Type[cutlass.Numeric],
+        lse_dtype: Type[cutlass.Numeric],
+        mma_qk_tiler_mn: Tuple[int, int],
+        mma_pv_tiler_mn: Tuple[int, int],
+        split_kv: int,
+        is_persistent: bool,
```
yuxianq, line 3477: `split_kv/is_persistent` are unused, can we remove them?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp8.py
```diff
@@ -0,0 +1,4456 @@
...
+        L: int,
+        R: int,
+        in_dtype: Type[cutlass.Numeric],
+        out_dtype: Type[cutlass.Numeric],
+        acc_dtype: Type[cutlass.Numeric],
+        lse_dtype: Type[cutlass.Numeric],
+        mma_qk_tiler_mn: Tuple[int, int],
+        mma_pv_tiler_mn: Tuple[int, int],
+        split_kv: int,
+        is_persistent: bool,
```
yuxianq, line 3423: `split_kv/is_persistent` are unused, can we remove them?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp16.py
```diff
@@ -0,0 +1,4499 @@
...
+                            common_params.blk_coord[1] * self.fold_sq_ratio +
+                            (tTR_tS[i][0] + common_params.blk_coord[0] *
+                             cta_m_rows) // self.num_heads)
+                    else:
+                        q_tok = common_params.blk_coord[1]
+                    k_bound = common_params.K - (self.seq_len_q - 1) + q_tok
+                    tTR_rAcc[i] = (tTR_rAcc[i] if cute.elem_less(
+                        tTR_tS[i][1] + self.mma_qk_tiler[1] * k_index,
+                        k_bound,
+                    ) else self.acc_dtype(-1.0e6))
```
yuxianq, line 2604: Is `-1.0e6` small enough that it won't be row max?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp8.py
haow-nv: Sure, we could remove those here. But, I want to confirm the strategy to maintain kernel source file from CuteDsl. I thought we shall not modify the source file unless necessary functional/perf fixing.

tensorrt_llm/_torch/modules/ATTENTION_DEVELOPER_GUIDE.md
```diff
@@ -247,21 +247,23 @@ The main differences across backends:
...
+`FlashInferTrtllmGenFmha` integrates trtllm-gen kernels from FlashInfer into
+the `TRTLLM` backend, and `FallbackFmha` calls the regular `thop.attention`
+runtime path. These are not separate attention backends.
 
 `TLLM_FMHA_LIBS` controls the ordered list. Unset means
-`flashinfer_trtllm_gen,fallback`; use `TLLM_FMHA_LIBS=fallback` or
-`TLLM_FMHA_LIBS=-flashinfer_trtllm_gen` to force the fallback path. Each FMHA
-library exposes `is_available()` for module/static environment checks and
-`is_supported()` for per-forward request checks.
+`cute_dsl_mla,flashinfer_trtllm_gen,fallback`; use `TLLM_FMHA_LIBS=fallback`
```
yuxianq: Seems that we forget to add msa_sparse_gqa in ATTENTION_DEVELOPER_GUIDE.md, can you also add it?

haow-nv: Sure. Will add it.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp16.py
```diff
@@ -0,0 +1,4499 @@
...
+        :type split_kv: int
+        :param acc_dtype: The data type of the output tensor C
+        :type acc_dtype: Type[cutlass.Numeric]
+
+        :return: The workspace size for the MLA kernel
+        :rtype: int
+        """
+        if split_kv == 1:
+            return 0
+        return B * H * S * split_kv * (D + 1) * acc_dtype.width // 8
```
yuxianq, line 3401: **Pad the split-KV workspace to the 128-row MMA store**

Both kernel variants accept every `H <= 128`, but size and lay out the split-KV partials with logical `H`. For the production `H=16, S_q=1, split_kv>1` path, the physical MMA store writes 128 rows: padded head 16 starts immediately after the H-sized `acc_o`, so the store aliases `acc_lse` and can continue beyond the raw workspace. The host-side maximum-workspace formula also uses `H`, so this can corrupt reduction inputs or trigger an illegal memory access.

**Suggested fix:** use `workspace_H = max(H, 128)` in both host sizing paths and the equivalent `cutlass.max` in both `initialize_workspace()` implementations, then derive the `acc_o`/`acc_lse` shapes and strides from `workspace_H`. Until that is fixed, reject split-KV when the post-fold head count is below 128.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp8.py
```diff
@@ -0,0 +1,4456 @@
...
+        gAccO = mAccO[blk_coord[0], None, None, blk_coord[1], blk_coord[2]]
+        rAccO = cute.make_rmem_tensor(cute.make_layout(elements_per_thread),
+                                      self.acc_dtype)
+        rO = cute.make_rmem_tensor(cute.make_layout(elements_per_thread),
+                                   self.o_dtype)
+        rAccO.fill(0.0)
+        for i in range(local_split_kv):
+            for j in cutlass.range_constexpr(elements_per_thread):
+                element_idx = tidx + j * self.threads_per_warp * self.num_compute_warps
+                rAccO[j] += gAccO[i, element_idx] * smem_lse_scale[i]
```
yuxianq, line 1438: **Bound split-KV before indexing the fixed-size reduction table**

`smem_lse_scale` has exactly `MAX_SPLITS` entries, but this loop runs to runtime `local_split_kv`; neither kernel's `can_implement()` bounds `split_kv`. For example, `seq_len_k=32896, split_kv=257` produces 257 K tiles with the default 128-column tile, and iteration 256 reads `smem_lse_scale[256]` out of bounds. The FP16 reduction has the same issue.

**Suggested fix:** reject `split_kv < 1` and `split_kv > MAX_SPLITS` in both `can_implement()` methods. For variable split-KV, also validate every `block_split_kvs[b]` against `min(split_kv, MAX_SPLITS)`, or consume scales in bounded chunks.

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
```diff
@@ -0,0 +1,572 @@
...
+        return True, ""
+
+    def _is_supported_with_reason(
+        self,
+        q: torch.Tensor,
+        attn: "TrtllmAttention",
+        meta: "TrtllmAttentionMetadata",
+        fwd: AttentionForwardArgs,
+    ) -> tuple[bool, str]:
+        if fwd.attention_input_type != AttentionInputType.generation_only:
```
yuxianq, line 265: **Reject Helix until the CuTe path exports softmax statistics**

This support gate does not reject Helix. `MLA._attn_forward_gen()` allocates an uninitialized `softmax_stats` tensor and later consumes it in `_helix_post_process()`, while the CuTe path ignores `softmax_stats_tensor` and keeps its LSE private in its workspace. Helix therefore combines partial outputs with uninitialized statistics and can silently return incorrect tokens.

**Suggested fix:** mirror the sibling FlashInfer guard:

```python
if meta.helix_position_offsets is not None:
    return False, "CuTe DSL MLA FMHA does not support Helix parallelism."
```

Supporting Helix instead requires exporting the correct per-token statistics and honoring inactive-rank semantics.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp16.py
Observer007: It is oob value to avoid -inf - inf => nan, not the initial value for max comparison.

yuxianq: Got it, thanks~

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp8.py
yuxianq: Decide to keep them after offline discussion

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp16.py
yuxianq: Decide to keep them after offline discussion.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
```diff
@@ -8209,3 +8209,776 @@ def _(
...
+            "trtllm::cute_dsl_mla_decode_fp16_blackwell",
+            [runner],
+            runner.get_tuning_config(),
+            inputs,
+        )
+        # ``forward`` requires a 4-tuple tactic; if the tuner returned its -1
+        # fallback (cache not warmed), supply the default 4-tuple so split_kv +
+        # is_persistent still come from a length-4 tactic.
+        if not (isinstance(best_tactic, tuple) and len(best_tactic) == 4):
+            best_tactic = runner.default_tactic(int(q_latent.shape[-1]))
```
hyukn, line 9812: I think we do not have default coverage for -1 during warmup as the default (needs confirmation). Then the fallback may cause another runtime cute.compile. This should not be a big deal because normally we do not expect fallback to happen.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
```diff
@@ -8209,3 +8209,776 @@ def _(
...
+            candidate_tiler_tactics = [
+                ((128, 128), (128, 256)),
+            ]
+            max_active_blocks = self._get_max_active_blocks()
+            split_candidates = self.get_split_kv_candidates(
+                batch_size, seq_len_q, max_active_blocks)
+            persistent_candidates = self.get_is_persistent_candidates()
+
+            valid = []
+            for mma_qk_tiler_mn, mma_pv_tiler_mn in candidate_tiler_tactics:
```
hyukn: nit: Using itertool.product requires only one indent level, which may be simpler.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
```diff
@@ -8209,3 +8209,776 @@ def _(
...
+                stream,
+            )
+            return o
+
+    @torch.library.custom_op(
+        "trtllm::cute_dsl_mla_decode_fp8_blackwell",
+        mutates_args=("o", "workspace"),
+        device_types="cuda",
+    )
+    def cute_dsl_mla_decode_fp8_blackwell(
```
hyukn, line 9669: So we can directly rely on the current attention unit tests to cover the new custom ops. Can the new tuning path be covered correctly?

hyukn (approved):
Overall LGTM.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
haow-nv: The current attention unit test could cover the new custom ops, but doesn't cover autotune.

yuxianq: @yihwang-nv is adding DSA unit tests for attention backends in https://github.com/NVIDIA/TensorRT-LLM/pull/16309, if we add `unittest/_torch/attention/test_attention_backends.py` to `tests/integration/test_lists/test-db/l0_rtx_pro_6000.yml`, will it cover autotune?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_decode_fp8.py
haow-nv: Discussed offline. The functionality is not affected even  split_kv > MAX_SPLITS. But will add checking to make sure splitkv is in [1,32] for performance.

.pre-commit-config.yaml
haow-nv: Discussed offline. We don't change the sync files (mla_decode.* + mla_helper.py) unless necessary fix. 

So still need to add the three cutedsl kernel files into  legacy-files.txt. Then call `python3 scripts/legacy_utils.py check-configs` to update `.pre-commit-config.yaml`, `pyproject.toml `and `ruff-legacy.toml`

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/attention/mla/mla_helpers.py
haow-nv: Discussed offline. We don't change the sync files (mla_decode.* + mla_helper.py) unless necessary fix.

tensorrt_llm/_torch/pyexecutor/model_engine.py
haow-nv: Discussed offline. We use `(1 + self.max_total_draft_tokens, 1)` to create generation-only request.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
haow-nv: The unit test without autotuner will get -1 .

yuxianq (approved):

yiqingy0 (approved):

Shixiaowei02 (approved):

tensorrt_llm/_torch/attention_backend/fmha/registry.py
```diff
@@ -33,6 +34,7 @@ def init_fmha_libs() -> dict[str, "FmhaCls"]:
     from .msa_sparse_gqa import MsaSparseGqaFmha
 
     return {
+        "cute_dsl_mla": CuteDslMlaFmha,
```
Shixiaowei02, line 37: Since a disaggregated generation server only ever runs decode-only batches, this decode-only library ends up taking essentially every forward there, versus a fraction of them in aggregated serving — yet the only off switch is `TLLM_FMHA_LIBS`, which no checked-in disagg script sets, and there is no disagg test covering it. Would you consider adding one disagg smoke (rather than a dis-agg e2e test) case?

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
brnguyen2: Revisiting this one: the earlier reply covered the autotuner-off unit test path, not serving. The serving-path story is:

1. With the autotuner on, warmup tunes each decode variant over a batch bucket ladder fixed by the engine's max batch size (the op passes `max_num_requests`), so any runtime batch maps to a tuned bucket. The `-1` fallback is not reached, and the tuned kernels were compiled during the profiling pass itself, so no runtime `cute.compile` occurs.
2. `-1` can still be reached when the engine skips the generation warmup (e.g. guided decoding, pipeline parallelism, draft engines). In that case the fallback derived `split_kv` from the raw batch size, which is a kernel variant tuning never compiled, so it would `cute.compile` inside the serving loop.

Addressed in #17266 (mirror of this PR with the review fixes applied): `default_tactic` now rounds the batch down to its tuning bucket first, so a fallback on a tuned runner reuses an already-compiled kernel, and an untuned runner compiles at most one variant per bucket (each compile is logged). A new unit test asserts a post-tuning pass triggers no new `cute.compile`. Leaving this unresolved so @haow-nv can cherry-pick from #17266 as needed.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
brnguyen2: Update on coverage: #16309 hasn't merged yet, and the in-tree `test_attention_backends.py` currently has no autotune or cute_dsl coverage. Also note `l0_rtx_pro_6000.yml` is SM120 hardware, where this lib never engages (it gates on SM100/103), so wiring that list wouldn't cover it.

Instead, #17266 adds a focused test in `test_attention_mla.py` (`test_attention_mla_cute_dsl_autotune`): a tuning-mode pass asserts the op is tuned and both tactic elements the tuner owns (`split_kv`, `is_persistent`) are profiled and compiled, then a serving-mode pass asserts the tuned kernels are reused with no new runtime `cute.compile`, with outputs matching the reference. `l0_b200.yml` already collects `unittest/_torch/attention` as a directory, so it runs in pre-merge B200 CI with no test-list change. Verified green on B200 (SM100). Leaving unresolved for cherry-pick from #17266.

tensorrt_llm/_torch/attention_backend/fmha/registry.py
brnguyen2: Added in #17266: `test_disaggregated_deepseek_v3_lite_bf16_gentp2_cute_dsl_mla_smoke`, a minimal ctxTP1+genTP2 DeepSeek-V3-Lite bf16 disagg cluster. Gen TP2 yields the 16 heads/rank the bf16 path admits at any batch size, so the decode-only generation server routes every forward through this lib (that is also why the smoke needs 3 GPUs rather than 2: at TP1 the 32-head geometry is not admitted by either dtype path). The test asserts client output and additionally requires the lib's kernel-compile log marker in a generation-worker log, so a silent fallback to the next FMHA lib fails the test rather than passing quietly.

On the off switch: `TLLM_FMHA_LIBS` accepts exact lists or +/- deltas, so `TLLM_FMHA_LIBS=-cute_dsl` on the generation server disables just this lib. No checked-in disagg config sets it, so the lib is on by default there, which this smoke now covers. Verified green on B200 (SM100). Leaving unresolved for cherry-pick from #17266.

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
brnguyen2: Drafted the rename in #17266 as the last commit so it's easy to drop either way: `CuteDslFmha` with registry key `cute_dsl` (so `TLLM_FMHA_LIBS` uses `cute_dsl`), and the file back to `fmha/cute_dsl.py` to keep matching the registered name. The kernel-specific op names (`trtllm::cute_dsl_mla_decode_*`) are unchanged and dispatch behavior is identical, so future non-MLA CuTe DSL kernels can land in this same lib.

@haow-nv this is your and @yuxianq's call: if you'd rather keep the MLA-specific name until the prefill integration actually lands, just drop that commit when cherry-picking from #17266.

Shixiaowei02:
Default-on and decode-only means this takes nearly every attention call on a disaggregated generation server, and no stage on that hardware exercises it. Can we get coverage in before it's the default?

haow-nv:
> Default-on and decode-only means this takes nearly every attention call on a disaggregated generation server, and no stage on that hardware exercises it. Can we get coverage in before it's the default?

Hi @Shixiaowei02 , I contact you through slack. It seems that I don't find the correct people. Could you ping me? Do we have testcases for disaggregated generation server? I think the cutedsl mla is naturally covered by those cases?

tensorrt_llm/_torch/attention_backend/fmha/cute_dsl_mla.py
brnguyen2: Will address this in a follow-up PR; see TRTLLM-15030.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
brnguyen2: Will address this in a follow-up PR; see TRTLLM-15030.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
brnguyen2: Will address this in a follow-up PR; see TRTLLM-15030.

tensorrt_llm/_torch/attention_backend/fmha/registry.py
brnguyen2: Will address this in a follow-up PR; see TRTLLM-15030.

PR #15163, opened 2026-06-09
[None][feat] skip-softmax on SM120: TMA-load + sync-MMA warp-specialized context FMHA for sm_120/sm_121

dcampora:
## Description

  Adds **skip_softmax**, a warp-specialized context (prefill) FMHA for the
  sm_120 / sm_121 (consumer Blackwell) family. It is the sm_120 / sm_121 attention
  path that implements the per-warp skip-softmax optimization (hence the name).

  Only half of the Hopper warp-specialization recipe ports to consumer Blackwell:
  the TMA-driven async loads survive, but async MMA does not (no `wgmma.async`),
  so a single producer warp issues `cp.async.bulk.tensor` loads for Q/K/V into
  granular shared-memory buffers via driver-API `CUtensorMap` descriptors, while
  the consumer warps run BMM1 + softmax (+ per-warp skip-softmax) + BMM2 on
  `mma.sync`, synchronized by an mbarrier producer/consumer handshake.

  It is opt-in and activates when a `SkipSoftmaxAttentionConfig` is attached to
  the attention layer (`TrtllmAttention.use_skip_softmax_fmha`). The C++ runner
  restricts dispatch to the supported shapes — BF16 in/out,
  `head_dim == head_dim_v` in {128, 256}, causal mask, packed QKV — and ignores
  the flag everywhere else, so it is a no-op on other hardware and configs.

  What's included:
  - **Kernel** (`cpp/kernels/fmha_v2/src/fmha/warpspec_sm120/`): `kernel_traits.h`,
    `dma_sync_mma.h` (producer / TMA), `compute_sync_mma.h` (consumer), plus the
    entry header `fused_multihead_flash_attention_kernel_ws_sm120.h` and a design
    doc `README.md`. Compiled sm_120-only via the `_context_attention_kernels_120`
    CMake target, guarded by `TLLM_ENABLE_SKIP_SOFTMAX_SM120`.
  - **In-engine bridges + dispatch**: `use_skip_softmax_fmha` is plumbed from the
    `thop.attention` op → `AttentionOp` → `MHARunnerParams` → `Launch_params` →
    `FusedMultiHeadAttentionXMMAKernelV2::run`, which routes the matching config to
    the `run_skip_softmax_*` bridges. Python property in
    `tensorrt_llm/_torch/attention_backend/trtllm.py`.
  - **Batched-prefill correctness**: the producer adds the per-request cumulative
    token offset (`cu_q_seqlens[bidb]`) to the Q/K/V TMA sequence coordinates;
    without it, multi-request prefill batches all returned request 0's result.

  Verified on an RTX PRO 6000 Blackwell (sm_120).
  
  ## Test Coverage
  
  - `tests/unittest/_torch/attention/test_skip_softmax_sm120.py` (new): forces the
    skip_softmax kernel via a `SkipSoftmaxAttentionConfig` with a tiny
    `threshold_scale_factor` (no tiles skipped ⇒ full softmax) and checks single-
    and multi-request causal prefill against both an fp32 reference and the default
    (non-skip_softmax) TRTLLM context kernel. Skips on non-sm_120/121 GPUs; the
    multi-request cases are the regression guard for the per-request TMA offset.
  - `tests/unittest/_torch/attention_backend/test_attention_op_sync.py` (existing):
    guards the `use_skip_softmax_fmha` kwarg plumbing across the `thop.attention`
    boundary.

  ## PR Checklist
  
  ## GitHub Bot Help
  
### DIFF
--- a/cpp/kernels/fmha_v2/src/fmha/warpspec_sm120/kernel_traits.h
+++ b/cpp/kernels/fmha_v2/src/fmha/warpspec_sm120/kernel_traits.h
@@ -17,7 +17,7 @@
 
 #pragma once
 
-// Kernel_traits for the halfspec sm_120 / sm_121 warp-specialized FMHA.
+// Kernel_traits for the skip_softmax sm_120 / sm_121 warp-specialized FMHA.
 //
 // Wraps the existing fmha::Kernel_traits_ template (which already provides
 // LDGSTS-friendly Smem_tile_a/b/v with ldmatrix swizzle and the right
@@ -82,10 +82,10 @@ template <
     bool ENABLE_SKIP_SOFTMAX_ = false,
     // Producer warp count -- single 32-thread warp by default.
     int NUM_PRODUCER_WARPS_ = 1>
-struct Kernel_traits_halfspec_sm120
+struct Kernel_traits_skip_softmax_sm120
 {
 
-    // Compose the existing PACKED_QKV tiled kernel traits with our halfspec
+    // Compose the existing PACKED_QKV tiled kernel traits with our skip_softmax
     // overrides. Kernel_traits_v2 (in fmha/kernel_traits.h) bakes in
     // fmha::v2::Gmem_tile_qkv for Q / K / V, which is the right gmem tile
     // for the PACKED_QKV input layout TRT-LLM passes for Qwen3.5 prefill.
@@ -93,7 +93,7 @@ struct Kernel_traits_halfspec_sm120
     // FLAGS: bit 0x1   = USE_LDGSTS_Q
     //        bit 0x2   = USE_LDGSTS_K
     //        bit 0x4   = USE_LDGSTS_V
-    //                    Halfspec leaves all three LDGSTS bits OFF; the
+    //                    Skip_softmax leaves all three LDGSTS bits OFF; the
     //                    producer warp issues TMA, not LDGSTS. The smem
     //                    layout is governed by BYTES_PER_LDG (=16,
     //                    independent of USE_LDGSTS) and the buffer count
@@ -132,7 +132,7 @@ struct Kernel_traits_halfspec_sm120
     using Mma_tile_p = typename Base::Mma_tile_p;
     using Mma_tile_o = typename Base::Mma_tile_o;
 
-    // Halfspec uses a RING_DEPTH-buffer ring for K, V (and Q -- though Q only
+    // Skip_softmax uses a RING_DEPTH-buffer ring for K, V (and Q -- though Q only
     // loads once per CTA, we still go through the same ring API so the
     // barrier handshake is uniform).
     enum
@@ -294,12 +294,12 @@ struct Kernel_traits_halfspec_sm120
     //
     // The consumer reuses the existing LDGSTS Smem_tile_q/k/v which, with
     // USE_GRANULAR_TILING, stream the contraction dim in chunks through a
-    // BUFFERS_PER_TILE-deep ping-pong (== GRANULAR_DEPTH below). The halfspec
+    // BUFFERS_PER_TILE-deep ping-pong (== GRANULAR_DEPTH below). The skip_softmax
     // TMA producer fills those exact granular buffers chunk by chunk -- it does
     // NOT add a ring on top. So the barrier depth == the granular buffer count.
     static constexpr int GRANULAR_DEPTH = Smem_tile_k::BUFFERS_PER_TILE;
     static_assert(GRANULAR_DEPTH == Smem_tile_q::BUFFERS_PER_TILE && GRANULAR_DEPTH == Smem_tile_v::BUFFERS_PER_TILE,
-        "halfspec assumes Q/K/V share the same granular buffer depth.");
+        "skip_softmax assumes Q/K/V share the same granular buffer depth.");
 
     // Bytes of one granular buffer (one chunk) for Q / K / V.
     static constexpr int BYTES_PER_BUFFER_Q = Smem_tile_q::BYTES_PER_BUFFER;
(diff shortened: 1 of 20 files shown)


cpp/tensorrt_llm/kernels/contextFusedMultiHeadAttention/fused_multihead_attention_v2.cpp
```diff
@@ -261,9 +261,44 @@ uint64_t FusedMultiHeadAttentionXMMAKernelV2::hashID(KernelMeta const& kernelMet
...
+
 void FusedMultiHeadAttentionXMMAKernelV2::run(
     Fused_multihead_attention_params_v2& params, Launch_params& launch_params, cudaStream_t stream) const
 {
+#if defined(TLLM_ENABLE_SKIP_SOFTMAX_SM120)
+    // Skip_softmax dispatch: route the matching config (sm_120 / sm_121, BF16
+    // in/out, head_dim == head_dim_v in {128, 256}, causal, PACKED_QKV) to the
+    // TMA-load + sync-MMA warp-specialized kernel when the caller opted in.
+    // Falls through to the normal cubin/launcher path otherwise.
+    if (launch_params.useSkip_softmaxFmha && (mSM == kSM_120 || mSM == kSM_121) && mInputDataType == DATA_TYPE_BF16
```
yuxianq: Why skip softmax sm120 is opt-in if we don't have existed kernels to support this case? Can we remove useSkip_softmaxFmha or even TLLM_ENABLE_SKIP_SOFTMAX_SM120 to always enable it?

cpp/tensorrt_llm/kernels/contextFusedMultiHeadAttention/fused_multihead_attention_common.h
```diff
@@ -340,6 +340,11 @@ struct MHARunnerParams
     uint32_t* skipSoftmaxTotalBlocks;
     uint32_t* skipSoftmaxSkippedBlocks;
 #endif
+
+    // Opt into the skip_softmax (TMA-load + sync-MMA warp-specialized) FMHA on
+    // sm_120 / sm_121. Only honored for the supported shapes (BF16, causal,
+    // head_dim == head_dim_v in {128, 256}, packed QKV); ignored otherwise.
+    bool useSkip_softmaxFmha = false;
```
yuxianq: Use `useSkipSoftmaxFmha` instead.

cpp/tensorrt_llm/common/attentionOp.h
```diff
@@ -542,6 +542,9 @@ class AttentionOp
     int mSageAttnNumEltsPerBlkK = 0;
     int mSageAttnNumEltsPerBlkV = 0;
     bool mSageAttnQkInt8 = false;
+    // Opt into the skip_softmax (TMA-load + sync-MMA warp-specialized) context FMHA
+    // on sm_120 / sm_121 for supported shapes.
+    bool mUseSkip_softmaxFmha = false;
```
yuxianq: Use `mUseSkipSoftmaxFmha` instead.

cpp/kernels/fmha_v2/src/fmha/warpspec_sm120/kernel_traits.h
```diff
@@ -0,0 +1,394 @@
...
+    using Cta_tile_o = typename Base::Cta_tile_o;
+    using Mma_tile_p = typename Base::Mma_tile_p;
+    using Mma_tile_o = typename Base::Mma_tile_o;
+
+    // Skip_softmax uses a RING_DEPTH-buffer ring for K, V (and Q -- though Q only
+    // loads once per CTA, we still go through the same ring API so the
+    // barrier handshake is uniform).
+    enum
+    {
+        RING_DEPTH = RING_DEPTH_
```
yuxianq: It seems that `RING_DEPTH` is never used

cpp/tensorrt_llm/kernels/contextFusedMultiHeadAttention/fused_multihead_attention_v2.cpp
dcampora: Do you mean that we should always have it enabled, and set a low threshold by default?

yuxianq: skip softmax feature is only enabled when we set skip_softmax_threshold_scale_factor_prefill/skip_softmax_threshold_scale_factor_decode, right? So we don't need another flag to determine whether we enable skip softmax or not. Once skip_softmax_threshold_scale_factor_prefill is set and we can find a kernel to run it, we enable it.

dcampora: Sounds good, it's now enabled when the factor is set.

cpp/tensorrt_llm/common/attentionOp.h
dcampora: Removed these as now there is a central skip-softmax threshold `params.skip_softmax_threshold_scale_factor` that controls execution of the algorithm.

cpp/kernels/fmha_v2/src/fmha/warpspec_sm120/kernel_traits.h
dcampora: Removed.

cpp/tensorrt_llm/kernels/contextFusedMultiHeadAttention/fused_multihead_attention_common.h
dcampora: Removed by using `params.skip_softmax_threshold_scale_factor` instead.

yuxianq (approved):

PR #15164, opened 2026-06-09
[TRTLLM-5516][feat] Qwen3.5/3.6 MoE engine-path support + naive Gated DeltaNet plugin

trappedintesseract:
## Summary

Adds **classic TensorRT engine-path** support for the Qwen3.5 / Qwen3.6 hybrid MoE family (HF architectures `Qwen3_5MoeForConditionalGeneration` / `Qwen3NextForCausalLM`), together with a naive (correctness-first) TensorRT plugin for the Gated DeltaNet (GDN) linear-attention recurrence.

The model is a Qwen3-Next-style hybrid built on the RecurrentGemma hybrid template: a 3:1 interleave of **Gated DeltaNet linear-attention** layers and **full-attention** layers, each followed by a **256-expert MoE + shared expert**. The GDN layers reuse the engine's paged conv/RNN-state path (as for Mamba / RecurrentGemma); the full-attention layers use the KV cache and add an `attn_output_gate`.

> **Note on scope vs. the PyTorch backend:** this targets the *classic TensorRT engine* path (`tensorrt_llm/models/qwen3_next/`, the `convert → trtllm-build → serialized engine` flow). It is independent of the existing PyTorch-backend Qwen3-Next implementation (`tensorrt_llm/_torch/...`) and the FlashInfer GDN kernels — disjoint code paths, no overlap.

## Changes

- **C++ — naive Gated DeltaNet plugin** (`cpp/tensorrt_llm/plugins/gatedDeltaNetPlugin/`): correctness-first kernel that walks each sequence token-by-token with an fp32 recurrent state `[D_k, D_v]` per (request, v-head). Optimization (chunking, tensor cores, vectorized loads, state tiling) is intentionally out of scope; the goal is numerical parity with the PyTorch reference. Only the dense (non-paged, non-remove-padding) path is wired up; the paged_state / remove_input_padding signatures are parsed and reserved for a follow-up. Registered in `tllmPlugin.cpp` + plugin CMake list.
- **C++ — runtime recognition** (`modelConfig.h`, `gptJsonConfig.cpp`): adds a `kQwen3Next` model variant, marks it RNN-based so the recurrent conv/RNN-state path is reused for the GDN layers, maps both architectures, and parses the recurrent-state config fields.
- **Python op binding** (`functional.gated_delta_rule`): binds the plugin; consumes the post-conv, head-split q/k/v and per-head decay `g` / update-gate `beta`, returns recurrence output + present state.
- **`layers/attention.py` — `attn_output_gate`**: optional sigmoid output gate (`context * sigmoid(gate)`) applied before `o_proj`. The gate shards over heads like Q (`gather_output=False`), so `o_proj` still owns the single TP all-reduce. Handles the FP8 attention-context case (dequantize with `dense`'s activation scale before the gate multiply, then re-quantize).
- **Model** (`tensorrt_llm/models/qwen3_next/`): config parser (nested HF text/vision/quant config → flat engine fields + recurrent-state self-description for the C++ runtime), checkpoint converter (mixed-precision FP8 attention + NVFP4 MoE/lm_head ingestion, gate-doubled `q_proj` de-interleave, Gemma-style RMSNorm `+1` fold), and the model (GDN mixer, full attention with output gate, paged conv/RNN-state wiring). Registers both architectures in `models/__init__.py`.

## Scope / not included

This is an initial engine-path drop covering the **text-generation path**. Out of scope here:

- The MTP (multi-token-prediction) head.
- The Qwen3-VL vision tower.
- Interleaved MRoPE for multimodal positions (text-only positions collapse to standard RoPE, so text generation is unaffected).
- The paged_state / remove_input_padding kernel paths in the GDN plugin (signatures reserved).

## Testing

The engine-path numerical validation was done with a standalone correctness harness (GDN recurrence vs. a PyTorch reference, plus token-exact comparisons against the HF/vLLM reference) that lives outside the library source tree, so it is **not** part of this PR. Adding in-tree unit tests for the GDN plugin and the model construction/conversion is a planned follow-up. Reviewers' guidance on the preferred test location/shape is welcome.

## Notes for reviewers

- Per CONTRIBUTING, this is filed against `main`.
- Pre-commit formatting/lint hooks (isort, yapf, clang-format, cmake-format, ruff, ruff-legacy, autoflake, codespell) pass on the changed files.

### DIFF
--- a/cpp/include/tensorrt_llm/runtime/modelConfig.h
+++ b/cpp/include/tensorrt_llm/runtime/modelConfig.h
@@ -47,6 +47,7 @@ public:
         kMamba = 3,          // https://github.com/state-spaces/mamba
         kRecurrentGemma = 4, // https://github.com/google-deepmind/recurrentgemma
         kEncDec = 5,
+        kQwen3Next = 6,      // Qwen3.5/3.6 hybrid: Gated DeltaNet (linear attention) + full attention
     };
 
     struct RnnConfig
@@ -782,7 +783,8 @@ public:
 
     [[nodiscard]] bool constexpr isRnnBased() const noexcept
     {
-        return mModelVariant == ModelVariant::kMamba || mModelVariant == ModelVariant::kRecurrentGemma;
+        return mModelVariant == ModelVariant::kMamba || mModelVariant == ModelVariant::kRecurrentGemma
+            || mModelVariant == ModelVariant::kQwen3Next;
     }
 
     [[nodiscard]] std::vector<LayerType> const& getLayerTypes() const noexcept
--- a/cpp/tensorrt_llm/plugins/CMakeLists.txt
+++ b/cpp/tensorrt_llm/plugins/CMakeLists.txt
@@ -72,6 +72,7 @@ set(PLUGIN_LISTS
     mixtureOfExperts
     selectiveScanPlugin
     mambaConv1dPlugin
+    gatedDeltaNetPlugin
     lruPlugin
     cumsumLastDimPlugin
     topkLastDimPlugin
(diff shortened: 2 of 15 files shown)


karljang:
@trappedintesseract ,
Thanks for your contribution. 
However, this PR targets the legacy TensorRT engine path (models/, TRT plugins, trtllm-build). We are deprecating that backend in favor of PyTorch and won't accept new feature work on the engine path.  Please route Qwen3.5/3.6 support through _torch/ instead.

PR #15171, opened 2026-06-09
[None][perf] Serialize GenerationRequest token-id lists as int32 bytes

hyukn:
## What

`GenerationRequest.prompt_token_ids` / `query_token_ids` are stored as Python `list[int]`. When the request is pickled on the RPC submit path (proxy → worker), each token emits a separate PyLong/BININT pickle frame. For long-context requests (tens of thousands of prompt tokens) this per-token framing dominates host-side serialization cost.

This adds `__getstate__` / `__setstate__` to `GenerationRequest` that encode the flat token-id lists as a raw int32 byte buffer (single numpy C pass) and decode them back to `list[int]` on the receiving side. Non-list values (`None`, `list[list[int]]`, ndarray) pass through untouched, so all downstream consumers are unchanged. The encoding is symmetric and only consumed within one build (proxy ↔ worker), so there is no cross-version wire-format concern.

This is the Python analog of the C++ `tllm.Request` `__getstate__`/`__setstate__` change; it targets the object actually pickled on the submit path.

## Effect

On DSv4 CTX-only (c256, ~41K-token ISL), the host-side `_fetch_new_requests` NVTX window drops ~22% in profiles where the IPC thread is the contention source. The effect is host/GIL-side; end-to-end throughput at GPU-bound concurrencies is unchanged within noise. Complementary to the C++ `tllm.Request` serialization change (which covers the worker-side DP broadcast path).

### DIFF
--- a/tensorrt_llm/executor/request.py
+++ b/tensorrt_llm/executor/request.py
@@ -146,6 +146,45 @@ class GenerationRequest:
         self.id = id
         return self
 
+    # --- int32-byte serialization of token-id lists (kills per-token pickle PyLong) ---
+    _I32 = "\x00i32be"
+
+    @staticmethod
+    def _enc_tokens(v):
+        # flat list[int] -> (_I32, int32 bytes); leave anything else (None,
+        # list[list[int]], ndarray) untouched.
+        if type(v) is list and (len(v) == 0 or type(v[0]) is int):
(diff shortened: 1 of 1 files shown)


Tabrizian (approved):

longlee0622:

Issue #14672, opened 2026-05-28
[Bug][AutoDeploy]: model-coverage entries fail because config rejects use_cache


tcherckez-nvidia:
### System Info

- CPU architecture: x86_64
- GPU properties: not captured from CI artifact
- Libraries
  - TensorRT-LLM branch or tag: main (pipeline context)
  - TensorRT-LLM commit: not captured here
  - Container used: AutoDeploy model-coverage CI container
- OS: CI environment
- Additional context: terminal pipeline 52785950, job 327395444

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

Pipeline `52785720` resolved to terminal model-coverage pipeline `52785950`.

Affected registry entries in `examples/auto_deploy/model_registry/models.yaml`:
- `apple/OpenELM-270M-Instruct`
- `apple/OpenELM-1_1B-Instruct`
- `apple/OpenELM-3B-Instruct`

Representative evidence:
1. Open job `327395444`.
2. Inspect per-test stdout logs for the three OpenELM tests.
3. Each traceback ends with:

```text
TypeError: OpenELMConfig.post_init() got an unexpected keyword argument 'use_cache'
```

Relevant stack tail:
```text
File "/opt/tensorrt-llm/tensorrt_llm/_torch/pyexecutor/config_utils.py", line 479, in load_pretrained_config
  model_config = transformers.AutoConfig.from_pretrained(
...
File "/usr/local/lib/python3.12/dist-packages/huggingface_hub/dataclasses.py", line 190, in init
  self.post_init(**additional_kwargs)
TypeError: OpenELMConfig.post_init() got an unexpected keyword argument 'use_cache'
```

### Expected behavior

OpenELM model-coverage entries should load HF config successfully through the current AutoDeploy/llmapi config-loading path.

### actual behavior

TensorRT-LLM fails during HF config loading before model execution starts because `OpenELMConfig.post_init()` rejects the `use_cache` keyword argument.

### additional notes

This reproduces consistently for all three OpenELM instruct entries. The current local registry has those entries disabled with this reason. Likely fix areas are the TensorRT-LLM config-loading path or OpenELM integration path so unsupported kwargs are filtered or normalized before config instantiation.

### Before submitting a new issue...

[...]

PR #15175, opened 2026-06-09
[#14672][fix] AutoDeploy: Vendor OpenELMConfig locally to fix OpenELM config loading

plapagesse:
## Release Notes

* **New Features**
  * Added support for three Apple OpenELM models (270M, 1.1B, and 3B Instruct variants) with integrated local configuration for improved reliability and deployment flexibility.

  ## Description

  Part of #14672 — fixes the OpenELM failures reported there; does **not** close the issue (see scope note at
  the end).

  `apple/OpenELM-*` ships a `trust_remote_code` config written for transformers 4.x. Its private `__post_init__(self)` collides with transformers 5.x's strict-dataclass `PreTrainedConfig`, which forwards unrecognized kwargs (e.g. `use_cache`) via
  `self.__post_init__(**kwargs)`, raising `TypeError: OpenELMConfig.__post_init__() got an unexpected keyword argument 'use_cache'` before any model code runs. Apple no longer maintains the upstream repo, so this cannot be fixed at the source.

  This PR bundles the config locally as `OpenELMConfig` (`model_type="openelm"`) inside the existing custom modeling file and registers it via `AutoConfig.register`, mirroring the pattern already used by other AutoDeploy custom models (EXAONE,
  GLM4-MoE-Lite, Kimi-K2). Because the registered class lives outside the `transformers` package, transformers treats it as `explicit_local_code` and uses it even when `trust_remote_code=True` and the checkpoint declares an `auto_map` — Apple's
  frozen remote code never executes. The per-layer head/FFN derivation is inlined into `__init__` (no `__post_init__`), producing values identical to Apple's published per-layer lists for both the 270M and 3B size points.

  Also re-enables the three OpenELM entries in the AutoDeploy model registry (`examples/auto_deploy/model_registry/models.yaml`) so they are picked up by model-coverage runs again.

  #14672 also reports `TypeError: Qwen3_5ForCausalLM.__init__() got an unexpected keyword argument 
  'use_cache'` for the Qwen3.5 dense models. That is the same root cause (transformers 5.x forwarding
  `use_cache`) but at a **different stage**: it fires at model construction (`from_config` → `__init__`),
  after the config has already loaded — so this config-layer fix cannot reach it, and no single in-repo
  mechanism cleanly covers both stages. The generic kwarg-dropping helper proposed in #15064 wraps exactly
  that construction call and should resolve the reported Qwen TypeError as well

  ## Test Coverage

  - `tests/unittest/auto_deploy/singlegpu/models/test_openelm_modeling.py` — full suite passes (16/16) on an L40S with transformers 5.5.4. New tests added by this PR:
    - `test_openelm_config_derivation_matches_apple` / `..._3b` — per-layer head/FFN derivation matches Apple's published 270M and 3B presets
    - `test_openelm_config_tolerates_unknown_kwargs` — vendored config accepts forwarded kwargs (the transformers 5.x failure mode)
    - `test_openelm_config_registered_as_local_class` — registered in `CONFIG_MAPPING` outside `transformers.`
    - `test_openelm_autoconfig_prefers_local_over_remote` — local class wins even with `trust_remote_code=True` + checkpoint `auto_map` (proves remote code never runs)
  - Manual before/after validation (L40S, NGC `release:1.3.0rc17` container, transformers 5.5.4):
    - **Without this fix:** `AutoConfig.from_pretrained("apple/OpenELM-270M-Instruct", trust_remote_code=True)` and `examples/auto_deploy/build_and_run_ad.py --use-registry --registry-config-id openelm` both fail with the `use_cache` `TypeError`
  from #14672 (executor worker init failure).
    - **With this fix:** the same end-to-end command loads, compiles (flashinfer + torch-cudagraph per registry config), and generates correctly for both `apple/OpenELM-270M-Instruct` and `apple/OpenELM-1_1B-Instruct`.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/auto_deploy/models/custom/modeling_openelm.py
```diff
@@ -60,6 +70,148 @@ def _make_divisible(v, divisor=8, min_value=None):
...
+        self.ffn_multipliers = ffn_multipliers
+        self.ffn_with_glu = ffn_with_glu
+        self.ffn_dim_divisor = ffn_dim_divisor
+        self.activation_fn_name = activation_fn_name
+        self.normalization_layer_name = normalization_layer_name
+        self.normalize_qk_projections = normalize_qk_projections
+        self.share_input_output_layers = share_input_output_layers
+        self.rope_freq_constant = rope_freq_constant
+        self.rope_max_length = rope_max_length
+        self.num_query_heads = (
```
govind-ramnarayan: nit: AI flagged that self.num_query_heads is rewritten below before it is ever read. Please remove.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_openelm.py
```diff
@@ -395,5 +547,18 @@ def forward(
         return OpenELMCausalLMOutput(logits=logits)
 
 
-# Register with AutoModelForCausalLMFactory
+# Register the vendored config with AutoConfig so AutoConfig.from_pretrained resolves
```
govind-ramnarayan, line 550: nit: can you check if this try-except pattern is really needed? We have custom models e.g. here that seems to be running fine without it with just a clean AutoConfig.register(). See:

https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_gemma4.py#L443

tests/unittest/auto_deploy/singlegpu/models/test_openelm_modeling.py
```diff
@@ -470,3 +470,133 @@ def test_openelm_model_can_be_exported():
...
+    assert config.num_transformer_layers == 36
+    assert config.num_query_heads == _EXPECTED_3B_NUM_QUERY_HEADS
+    assert config.num_kv_heads == _EXPECTED_3B_NUM_KV_HEADS
+    # ffn multipliers expand to a length-36 linspace(0.5, 4.0)
+    assert len(config.ffn_multipliers) == 36
+    assert config.ffn_multipliers[0] == 0.5 and config.ffn_multipliers[-1] == 4.0
+    assert all(q % k == 0 for q, k in zip(config.num_query_heads, config.num_kv_heads))
+
+
+def test_openelm_config_tolerates_unknown_kwargs():
```
govind-ramnarayan: why is this test meaningful (do we expect to be initializing it with unknown args? Is this some kind of defensive check against future modeling / transformers updates)?

govind-ramnarayan (commented):
Thanks! Added some nits and comments but overall looks good. I'll wait for your responses before accepting.

tensorrt_llm/_torch/auto_deploy/models/custom/modeling_openelm.py
plapagesse: ah, yes I had followed the pattern at [`modeling_exaone.py`](https://github.com/NVIDIA/TensorRT-LLM/blob/a8c4007284f7e452e71e751c1c12988d651e1cbb/tensorrt_llm/_torch/auto_deploy/models/custom/modeling_exaone.py#L120-L127), where it tolerates transformers versions predating `exist_ok`; should be good to have the bare one-liner like the mentioned models, though. Updated

plapagesse:
thanks for the review @govind-ramnarayan! Addressed the points

tests/unittest/auto_deploy/singlegpu/models/test_openelm_modeling.py
```diff
@@ -470,3 +470,165 @@ def test_openelm_model_can_be_exported():
...
+]
+_EXPECTED_3B_NUM_KV_HEADS = [
+    3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4,
+    5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 6, 6, 6, 6, 6, 6,
+]
+# fmt: on
+
+
+def test_openelm_config_derivation_matches_apple():
+    from tensorrt_llm._torch.auto_deploy.models.custom.modeling_openelm import OpenELMConfig
```
govind-ramnarayan: One more comment. Here we are importing the OpenELMConfig individually for each test. But I think now that we are creating a custom config in production code, we should modify this whole test file accordingly. We should change `_create_small_config()` to not create a dummy test config, but rather use the proper OpenELMConfig that we import once at the top of the file. Every test should continue to work

tests/unittest/auto_deploy/singlegpu/models/test_openelm_modeling.py
```diff
@@ -470,3 +470,165 @@ def test_openelm_model_can_be_exported():
...
+    assert not hasattr(config, "use_cache")
+    # Structural tripwire: nobody reintroduces a __post_init__ on this class
+    # (the Apple pattern that caused issue 14672).
+    assert "__post_init__" not in OpenELMConfig.__dict__
+    # Construction stayed healthy alongside the surplus kwargs.
+    assert len(config.num_query_heads) == 16
+
+
+def test_openelm_config_registered_as_local_class():
+    from transformers.models.auto.configuration_auto import CONFIG_MAPPING
```
govind-ramnarayan: nit: let's move other imports like this up to the top of the file if we can too.

tests/unittest/auto_deploy/singlegpu/models/test_openelm_modeling.py
```diff
@@ -470,3 +470,165 @@ def test_openelm_model_can_be_exported():
...
+def test_openelm_autoconfig_prefers_local_over_remote(tmp_path):
+    """AutoConfig must prefer the local class over remote code.
+
+    Even with trust_remote_code=True and an auto_map present — and with no
+    configuration_openelm.py on disk, so a remote-code path would fail — the
+    local class is used. This proves Apple's remote code never runs.
+    """
+    import json
+
+    from transformers import AutoConfig
```
govind-ramnarayan: ditto

tests/unittest/auto_deploy/singlegpu/models/test_openelm_modeling.py
govind-ramnarayan: See e.g. https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/auto_deploy/singlegpu/models/test_glm4_moe_lite_modeling.py#L42 as an example

plapagesse:
responded inline to the comment above earlier, but commenting down here for better viz 🤙. Modified the tests to use the custom class and moved the imports to the top. All still passing. Thanks again for the review!

govind-ramnarayan (approved):
Thank you for the contribution! Tested that `build_and_run_ad.py --use-registry` works with these models now.

PR #15197, opened 2026-06-10
[None][fix] Support V2 KV cache beam search

yizhang-nv:
## Release Notes

* **New Features**
  * KVCacheManagerV2 now supports beam search with configurable beam widths at generation time, removing the previous beam_width=1 restriction for broader compatibility.

* **Tests**
  * Added integration and unit tests validating beam search functionality with the v2 KV cache manager, including scenarios with block reuse and partial token matching.

## Description

Enable KV cache manager v2 to support beam-search expansion during generation. The change materializes page-index buffers for expanded beams, copies the live partial-tail/SSM page state from beam 0 for new beams, and keeps historical full blocks unmapped for expanded beams. It also disables partial reuse when V2 is configured with multi-beam capacity.

This keeps V2 selected for `max_beam_width > 1` instead of falling back to the V1 KV cache manager. The runtime `commit()` path still only accepts `beam_search_indices is None`, but it no longer rejects a cache whose layout has already been expanded for beam search by overlap scheduling. Published blocks are treated as beam0-canonical for reuse, and non-default beam pages are cleared when a block is committed.

This adds targeted KV cache manager v2 unit coverage and LLM API accuracy coverage for V2 beam search.

## Test Coverage

- Syntax/diff checks passed.
- Targeted V2 beam unit tests passed.
- V2 beam LLM API and V1/V2 integration checks passed.

## PR Checklist

## GitHub Bot Help


jieli-matrix (approved):

fredricz-20070104:
Thanks for picking this up — V2 multi-beam has a lot of edges. Three correctness issues I think need to be addressed before merge:

1. **`_commit_block` still asserts single-beam.** `commit()` no longer rejects `max_beam_width > 1`, but the downstream `_commit_block` in `_kv_cache.py` still has `assert typed_len(seq_block.pages) == 1, "Must have 1 beam only"`. After beam expansion, the first full-block commit will hit this assert. If the intent (per the PR description) is for multi-beam caches to flow through `commit()`, this assertion needs to be relaxed and the non-beam-0 pages need to be released — consistent with the "publish beam 0 as canonical, drop the rest" policy already in `commit()`. Otherwise we'd need a guarantee that `commit()` always runs before beam expansion, and I don't see that guarantee in the current code.

2. **`_append_beams` skips historical blocks without padding `block.pages`.** When `_is_block_before_history_cursor(ordinal)` is true the loop `continue`s without appending a placeholder. After `self._beam_width` is bumped to `new_bw`, those historical blocks still have `len(block.pages) == old_bw`, so any later access via `b.pages[beam_idx]` (including `_block(ordinal, beam_idx)`) will `IndexError`. `_check_sanity` only tolerates `holder=None`, not missing rows. Two options: (a) append a placeholder `_BeamBlock(holder=None)` to historical blocks so `pages` stays rectangular, or (b) make every `pages[beam_idx]` access tolerate sparsity. (a) is the smaller, safer change.

3. **`_ensure_generation_beam_width` is not rolled back on caller failure.** In `try_allocate_generation` / `_prepare_draft_resources`, beam width is bumped and pages are cloned *before* the caller's `kv_cache.resize(...)`. If `resize` fails (OOM, etc.), the cache is left in a "already expanded, holding cloned pages" state and scheduling will keep retrying / failing. Either roll back the beam-width bump and cloned pages on caller resize failure, or move the resize inside `_ensure_generation_beam_width` so the whole expansion is transactional.

fredricz-20070104:
Thanks for taking this on — V2 multi-beam has a lot of subtle edges. I see three correctness blockers (matching the CodeRabbit Critical/Major findings) that should be resolved before merge:

**1. (Critical) `_commit_block` still hard-asserts single beam.**
`commit()` no longer rejects `max_beam_width > 1`, but `_kv_cache.py:1116` still has:
```python
assert typed_len(seq_block.pages) == 1, "Must have 1 beam only"
```
So the first full-block commit after a beam-width expansion will fail. If the design intent (per the PR description) is to allow a multi-beam cache to flow through `commit()`, this assertion needs to be relaxed and non-default beam pages need to be released (treating beam 0 as canonical, consistent with what `commit()` already does for the `beam_search_indices` argument). Otherwise the invariant "commit() always runs before beam expansion" needs to be made explicit — I don't see that guarantee in the current code.

**2. (Major) `_append_beams` leaves historical blocks' `pages` array shorter than the new beam width.**
At `_kv_cache.py:1397-1401`, when `_is_block_before_history_cursor(ordinal)` is true the loop `continue`s without appending a placeholder to `block.pages`. After `self._beam_width` is promoted to `new_beam_width`, those historical blocks still have `len(block.pages) == old_beam_width`. Any subsequent access via `block.pages[beam_idx]` (e.g. `_block(ordinal, beam_idx)`) will `IndexError`, and `_check_sanity` does not currently allow a `pages` row to be entirely missing — it only allows `holder=None`.

Two ways to fix this:
- (a) Append a placeholder `_BeamBlock` (holder=None) for historical blocks too, so `pages` length stays consistent.
- (b) Make every `pages[beam_idx]` access tolerant of a sparse array.

(a) is smaller and safer.

**3. (Major) `_ensure_generation_beam_width` does not roll back if a later `resize` fails.**
In `resource_manager.py:2896-2902` (and similarly in `_prepare_draft_resources` / the draft path), `_ensure_generation_beam_width` is called first — it promotes `beam_width` and clones pages — and then `kv_cache.resize(...)` is called. If the resize fails (e.g. OOM), the cache is left in a half-expanded state with cloned pages held, and the scheduler will fail the same request indefinitely. Either roll back `beam_width` and the cloned pages when the caller's `resize` fails, or move the `resize` inside `_ensure_generation_beam_width` and make the whole thing transactional.

Once 1–3 are addressed I think this is ready to merge. Non-blocker side note for future cleanup: `enable_partial_reuse = config.enable_partial_reuse and max_beam_width == 1` is gated on the configured `max_beam_width`, so a single-beam request running under a multi-beam configuration cannot benefit from partial reuse. V1 gates this on the cache's current beam width. Worth converging at some point but not a merge blocker.

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
```diff
@@ -869,24 +889,20 @@ def history_length(self, history_length: int) -> None:
...
     # (num_committed_tokens <= history_length). Note that history_length increase may trigger out-of-window
     # block eviction/dropping for SWA layers.
-    # beam_search_indices: indices indicating which candidate to choose for each token. A block with all
-    # tokens committed will be unified to one memory page and the other memory pages are dropped. Only for
-    # beam search.
+    # beam_search_indices is reserved for future use and must be None.
     def commit(
         self,
         accepted_input_tokens: Sequence[TokenIdExt],
         beam_search_indices: Sequence[int] | None = None,
```
lowsfer: Let's just remove this parameter. It was a mistake.

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
```diff
@@ -1526,6 +1552,137 @@ def _recover_excess_scratch_slots(
...
+            ):
+                raise NotImplementedError(
+                    "Changing beam_width on a suspended non-empty KV cache is not supported"
+                )
+
+        storage = self._storage
+        num_life_cycles = storage.num_life_cycles
+        slot_counts = filled_list(0, num_life_cycles)
+        num_new_beams = int(new_beam_width - old_beam_width)
+        for ordinal, block in typed_enumerate(self._blocks):
```
lowsfer: I think we can start from current `history_length // tokens_per_block` here. beam_width resize should only impact future tokens. history tokens won't be impacted, right?

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
```diff
@@ -1526,6 +1552,137 @@ def _recover_excess_scratch_slots(
...
+        if any(
+            type(indices) is memoryview
+            for beam_indices in self._base_page_indices
+            for indices in beam_indices
+        ):
+            pending_ready_events = [
+                event for event in set(new_slot_ready_events) if not event.query_complete()
+            ]
+            for event in pending_ready_events:
+                event.synchronize()
```
lowsfer: why trigger sync?

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
```diff
@@ -1526,6 +1552,137 @@ def _recover_excess_scratch_slots(
...
+                raise NotImplementedError(
+                    "Changing beam_width on a suspended non-empty KV cache is not supported"
+                )
+
+        storage = self._storage
+        num_life_cycles = storage.num_life_cycles
+        slot_counts = filled_list(0, num_life_cycles)
+        num_new_beams = int(new_beam_width - old_beam_width)
+        for ordinal, block in typed_enumerate(self._blocks):
+            if self._is_block_shared_across_beams(ordinal, block):
```
lowsfer: Can we simply start from `history_length // tokens_per_block`?

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
```diff
@@ -1526,6 +1552,137 @@ def _recover_excess_scratch_slots(
...
+            )
+
+        new_slot_ready_events = tuple(slot.ready_event for slot in chain.from_iterable(new_slots))
+        if any(
+            type(indices) is memoryview
+            for beam_indices in self._base_page_indices
+            for indices in beam_indices
+        ):
+            pending_ready_events = [
+                event for event in set(new_slot_ready_events) if not event.query_complete()
```
lowsfer: stream_wait_events already does deduplication. And no need to call query_complete(), as query is not much cheaper than cudaStreamWaitEvent anyway. query_complete() is mainly used for scrubbing to avoid accumulating too many events.

tensorrt_llm/runtime/kv_cache_manager_v2/_core/_kv_cache.py
yizhang-nv: For overlap scheduler, we expand the beam width ahead of commit/history len increase, so we can only use this to make sure we won't expand the context prompt.

tensorrt_llm/runtime/kv_cache_manager_v2/__init__.pyi
```diff
@@ -465,6 +463,8 @@ class KVCacheManager:
     def allow_seq_rebasing(self) -> bool: ...
     @property
     def enable_partial_match(self) -> bool: ...
+    @property
```
litaotju, line 466: I dont understand this. Shouldn't this file be generated/updated automatically? 

Do we need to check the .pyi files to the git repo?

litaotju (approved):

PR #14413, opened 2026-05-21
[None][perf] DSv4 indexer Top-K: align kFTarget=kK, uniform kSeqSmall=4096, diagnostic env knobs

longcheng-nv:
## Summary

Three correlated tuning changes for the V4 indexer Top-K Heuristic dispatch on
Blackwell GPUs, all exercising the same `invokeIndexerTopKDecode` /
`invokeIndexerTopKDecodeDtype` code path and sharing a single B200/B300
swe-bench-synth benchmark anchor (V3.2-Q19c protocol). Ships as one PR.

### 1. `kFTarget = kK` alignment for K=512/1024 (`heuristic_topk.cuh`)

| Specialisation | Old `kFTarget` | New `kFTarget` |
|---|---:|---:|
| `<{fp32,bf16,fp16}, 512>` | 384 | **512** (= kK) |
| `<{fp32,bf16,fp16}, 1024>` | 2560 | **1024** (= kK) |
| `<*, 2048>` | 3072 fp32 / 4096 bf16,fp16 | unchanged |

Aligns the secant solver's soft-target with the band's lower edge. Eliminates
upper-clamp saturation on tight-σ high-A2 layers (K=512) and corrects a
mistune for K=1024 originally calibrated against M=512 (Flash inheritance)
instead of production M=K=1024.

**Full GVR `GvrParams<InputT, TopK>` table (post-PR, `heuristic_topk.cuh:224-313`).**
`f = kNumBins` (P4 histogram bin count) · `C = kC` (candidate buffer cap) ·
`f_target = kFTarget` (P2-secant soft target for early termination):

| K \\ dtype | **fp32** | **bf16** | **fp16** |
|---|---|---|---|
| **512** | f=1024, C=5120, f_target=**512** | f=512, C=5120, f_target=**512** | f=512, C=5120, f_target=**512** |
| **1024** | f=1024, C=5120, f_target=**1024** | f=512, C=5120, f_target=**1024** | f=1024, C=5120, f_target=**1024** |
| **2048** | f=2048, C=**6144**, f_target=**3072** ★ | f=2048, C=5120, f_target=**4096** | f=2048, C=5120, f_target=**4096** |

★ `(fp32, K=2048)` is **V2e SASS byte-identity locked** — `kFTarget` and
`kC` cannot be moved without altering ptxas output for the V3.2 production
hot path; this PR preserves the legacy `(3072, 6144)` tuple for that
single cell. All other 8 cells use `kC=5120` (sized for 4–5 CTA/SM
[...]

PR #15198, opened 2026-06-10
[TRTLLM-35882][feat] cute dsl gvr-top multi-cta optimization

limin2021:
## Summary by CodeRabbit
Add cute dsl gvr-topk multi-cta optimization, particularly used for smarll BS and Large seq_len cases.

Parallelism: Only parallel for phase 2 and phase 3. Phase1&4 is only done for one-cta. Therefore, the scalability of this multi-cta impl is not vary well.

### performance

cute dsl gvr-topk multi-cta version VS single-cta version
image

cute dsl gvr-topk tuned version (combined by single-cta and multi-cta) VS cuda gvr-topk
image

Note: next_n=1 (MTP=0)

## Test Coverage

```
# ut running in CI, used for production.
python -m pytest tests/unittest/_torch/attention/sparse/test_cute_dsl_gvr_topk_decode.py
# standard alone ut, which is not dependent on trtllm env. It don't run in CI.
python tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
python -m pytest tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
```

## PR Checklist

## GitHub Bot Help


hyukn (approved):
LGTM

tests/unittest/_torch/attention/sparse/test_cute_dsl_gvr_topk_decode.py
```diff
@@ -190,27 +197,46 @@ def _tie_aware_check(
...
+        (torch.float32, 2048),
+    ],
+)
 @pytest.mark.parametrize("N", [4096, 65536])
+@pytest.mark.parametrize("varlen", [False, True])
 @pytest.mark.parametrize("next_n", [1, 2])
 @pytest.mark.parametrize("batch_size", [1, 32])
 @pytest.mark.parametrize("compress_ratio", [1, 4])
 @pytest.mark.parametrize("preidx_hit_rate", [0.0, 0.5])
+@pytest.mark.parametrize("cluster_size", [1, 4])
```
mingyangHao, line 219: This `cluster_size` is sweeping within [1,2,4,8], should we add full set into test cases?

tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
```diff
@@ -120,6 +121,7 @@ def _compile(
         enable_warp_parallel_reduce=enable_warp_parallel_reduce,
         compress_ratio=compress_ratio,
         return_output_values=return_output_values,
+        cluster_size=cluster_size,
```
mingyangHao, line 124: `enable_smem_cache` is not passing into the function, are we deliberately want to disable that feature?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
```diff
@@ -1591,60 +1963,146 @@ def gvr_topk_kernel(
...
+                        local_cnt_self = s_iscalars[5]
+                        local_iscalars_ptr = s_iscalars.iterator + cutlass.Int32(5)
+                        smem_keys_iter = smem_keys.iterator
+                        smem_vals_iter = smem_vals.iterator
+                        base_offset = local_cnt_self
+                        for peer in cutlass.range_constexpr(1, cluster_size):
+                            peer_iscalars_addr = mapa_shared_cluster(
+                                local_iscalars_ptr, cutlass.Int32(peer)
+                            )
+                            peer_cnt = ld_shared_cluster_i32(peer_iscalars_addr)
```
longcheng-nv, line 2094: The gather bounds the copy by the peer's **uncapped** slice-local count (`s_iscalars[5]`), but each CTA's `smem_keys` is sized `kC` and Phase-3's stream-write caps writes at `kCC (== kC)`. Only the *write* below is guarded (`if dst < self.kC`), not this *read* — so if a peer's slice-local count ever exceeds `kC`, the loop DSMEM-reads `peer.smem_keys[i_gather]` past the peer's allocation.

This requires the cluster-total `cand_count > kC`, i.e. the `done==2` bracket-exhaustion path. I checked captured V4 data (op-bench set validated against this kernel: cr=4, deep bucket, N≤256K, Flash K=512 + Pro K=1024) — `cand_count` never reaches `kC` and `done==2` is never hit, so this is **not** triggered in production. But it's a cheap defense-in-depth gap vs the single-CTA path (immune via `cand_count_p4 = min(s_iscalars[0], kC)`). Suggest:
```python
peer_cnt = ld_shared_cluster_i32(peer_iscalars_addr)
peer_cnt = min(peer_cnt, cutlass.Int32(self.kC))  # peer wrote at most kCC; reading more is OOB
```
(keeping `base_offset` accumulation consistent).

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
```diff
@@ -141,17 +227,37 @@ def __init__(
...
+            )
+        self.cluster_size = cluster_size
+        # SMEM slice cache (optional): pre-stage each CTA's slice into SMEM
+        # once between Phase 1 and Phase 2, so Phase 2/3's GE-count scans
+        # read LDS instead of re-streaming GMEM. Requires
+        # ``slice_len <= smem_cache_elems`` at runtime (wrapper auto-disables
+        # otherwise); ``smem_cache_elems`` also sets the JIT-time alloc size.
+        if enable_smem_cache and smem_cache_elems <= 0:
+            raise ValueError("smem_cache_elems must be > 0 when enable_smem_cache")
+        self.enable_smem_cache = enable_smem_cache
```
longcheng-nv, line 249: `enable_smem_cache` (and `load_slice_to_smem` + all the `smem_input is not None` branches threaded through `block_count_ge`/`phase3_collect_candidates`, ~150 lines) is never enabled: grepping `tensorrt_llm/` and `tests/` it is set in **zero** call sites — the custom op `forward` doesn't pass it and neither test does. So the path is uncompiled in CI and unused in production.

Could we either (a) add one test that sets `enable_smem_cache=True` so it's exercised, or (b) split it into a follow-up PR? As-is it's complexity + silent-rot risk with no coverage, and it's a separate concern from the multi-CTA feature.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
```diff
@@ -478,50 +651,72 @@ def phase1_preidx_stats(
...
+        slice_end,  # int32: index in input_row where this CTA's slice ends
         threshold,  # cutlass.Float32 scalar
         smem_ptcnt,  # cute.Tensor [BLOCK_SIZE] int32 (P3 cache)
         smem_wcnt,  # cute.Tensor [NUM_WARPS] int32 (block reduce scratch)
         s_iscalars,  # cute.Tensor [5] int32 (writes [0] = cand_count)
+        s_cluster_partial,  # cute.Tensor [1] int32 (per-CTA partial scratch for DSMEM)
         tidx,
         warp_id,
         lane,
+        do_cluster_aggregation: cutlass.Constexpr[bool] = True,
```
longcheng-nv: `do_cluster_aggregation` looks dead — every call site uses the default `True`, so the `False` branch (and its "leader-only path / cluster sync against absent peers would deadlock" rationale) is never reached. Either drop the parameter or wire up the leader-only path the comment describes. Minor.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
```diff
@@ -1591,60 +1963,146 @@ def gvr_topk_kernel(
...
-                # =============================================================
-                # Phase 3 — Ballot-free candidate collect
-                # =============================================================
+                # Cluster handoff #1 (end of Phase 2): after Phase 2 every
+                # CTA agrees on s_iscalars[0] (cluster cand_count) and
+                # s_thr (deterministic secant), and each CTA's smem_ptcnt
+                # holds its slice-local per-thread counts — the basis for
+                # its own Phase 3 prefix sum.
+                if cutlass.const_expr(cluster_size > 1):
+                    cute.arch.cluster_arrive_relaxed()
```
longcheng-nv, line 2040: The DSMEM producer→consumer handoffs rely on `cluster_wait`'s acquire pairing with a *relaxed* arrive (no explicit `fence.cluster`). This matches the merged `single_pass_multi_cta_radix_topk_cluster.py` idiom, so treating it as validated precedent — just flagging since a wrong ordering assumption here would surface as an *intermittent* (not deterministic) failure. No action if the radix path is known-correct.

tests/unittest/_torch/attention/sparse/test_cute_dsl_gvr_topk_decode.py
```diff
@@ -190,27 +197,46 @@ def _tie_aware_check(
 
 
 @skip_not_sm100
-@pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16, torch.float16])
-@pytest.mark.parametrize("top_k", [512, 1024, 2048])
+@pytest.mark.parametrize(
+    "dtype,top_k",
+    [
+        (torch.bfloat16, 512),
```
longcheng-nv, line 207: The dtype/K parametrization changed from the full cross-product `{fp32,bf16,fp16} x {512,1024,2048}` to a zipped 3-tuple `{(bf16,512),(bf16,1024),(fp32,2048)}`. Reasonable to offset the `cluster_size`/`varlen` blow-up, but the net effect is **fp16 is no longer exercised at all** (and `fp32@{512,1024}`, `bf16@2048` dropped). Given the dtype-specific paths (256-bit load, cvt-to-fp32 scalar tail, fragment-reg footprint), could we keep at least one fp16 cell? Please confirm fp16 GVR is covered elsewhere if intentional. (Same applies to `tests/scripts/.../run_gvr_topk.py`.)

longcheng-nv (commented):
Nice work — the `cluster_size` auto-dispatch and the `cluster_size=1` const-expr degeneration are clean, and reusing the merged radix `mapa.shared::cluster` / `ld.shared::cluster` idiom keeps the DSMEM path consistent with existing precedent. I verified the supporting symbols (`_query_max_cluster_size`, `_get_num_sms`, `logger.warning_once(key=)`) and the JIT cache-key plumbing (`cluster_size` added to both key tuples) — all correct, and the new `Optional[int]=None` arg is added symmetrically to the `custom_op` impl and `register_fake` so existing callers are unaffected.

Findings below: one latent-OOB question (cluster gather), one scope/coverage item (`enable_smem_cache` dead code), one test-matrix concern (fp16 dropped), plus a few minors. **None are merge-blockers** given the data check described in the gather thread.

**On the cluster-gather OOB — I validated it against captured data, not just reasoned about it.** Using the op-bench `cand_count`/`p2_converged` set that is validated 108/108 against *this* cuteDSL kernel (cr=4 current params per PR #14413, deep beta bucket, N up to 256K, all dtypes, K=512 Flash + K=1024 Pro): across 2880 cells `p2_converged==1` (zero `done==2`, max P2 iters = 5) and `cand_count` never reaches `kC` (Pro K=1024 max 5115 < 5120; even the deep bucket maxes 5082). So per-slice ≤ row-total `cand_count` < `kC`, and the gather never reads past a peer's `smem_keys` on realistic distributions. The clamp suggestion is therefore **defense-in-depth, not a correctness blocker**.

Minor (no clean line to anchor): in the degenerate path (`N <= top_k`) all `cluster_size` CTAs execute the copy and write the full row output to the same global addresses — benign (identical values) but a redundant multi-writer; consider gating on `is_leader` when `cluster_size > 1`.

tests/unittest/_torch/attention/sparse/test_cute_dsl_gvr_topk_decode.py
limin2021: 2,4,8 runs the same code branch. To reduce the UT time, I only use 4 to cover this code branch.

tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
limin2021: Yes. Now I don't tune good/stable perf improvement with enable_smem_cache=true. Will expore it latter.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
```diff
@@ -5931,6 +5987,7 @@ def cute_dsl_gvr_topk_decode(
         next_n: int = 1,
         compress_ratio: int = 1,
         max_seq_len: Optional[int] = None,
+        cluster_size: Optional[int] = None,
```
yuxianq, line 5990: Also update the Args in docstring

lfr-0531 (approved):

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
limin2021: done.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
limin2021: done.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
limin2021: The kernel has the support for enable_smem_cache. I don't enable it in op interface because I don't get good/stable perf improvement with enable_smem_cache=true simplily. Considering the tune time, I don't do it now. Maybe I will expore it latter. We could keep it in kernel file. So it could be reused in the near feature.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
limin2021: Good catch. Done.

tests/unittest/_torch/attention/sparse/test_cute_dsl_gvr_topk_decode.py
limin2021: done.

tensorrt_llm/_torch/custom_ops/cute_dsl_custom_ops.py
limin2021: done.

yuxianq (approved):

mingyangHao (approved):
LGTM

longcheng-nv (approved):
Re-reviewed at `8fb7e58` — all findings from my prior pass are resolved, verified against the code (not just the "done" replies):

- **Cluster-gather OOB (gather, leader path)** — fixed. `peer_cnt = min(peer_cnt, kC)` (line 2105) now bounds the DSMEM `ld.shared::cluster` read loop, matching the `dst < kC` write guard and the single-CTA `cand_count_p4 = min(s_iscalars[0], kC)` downstream. Read and write are now symmetrically clamped to the peer's `smem_keys`/`smem_vals` allocation; the `done==2` bracket-exhaustion path can no longer over-read. Good comment documenting why.
- **DSMEM producer→consumer ordering** — fixed correctly, and with the right granularity. The two *release-critical* handoffs (cross-CTA partial-sum publish in `block_count_ge`, line 852; leader gather pre-read, line 2073) now use non-relaxed `cluster_arrive()` so the release pairs with the peer `cluster_wait` acquire. The remaining two `cluster_arrive_relaxed()` are correctly left as-is: handoff #1 (line 2040) is followed by own-slice-only Phase 3 (no cross-CTA read), and the final barrier (line 2155) is a pure liveness/SMEM-lifetime guard with no DSMEM read after it. The distinction is exactly right.
- **`do_cluster_aggregation` dead param** — removed (grep-clean).
- **fp16 dropped from test matrix** — restored: `(torch.float16, 1024)` cell added with a comment, keeping the fp16 cvt-to-fp32 tail path under CI even though K=1024 fp16 isn't a deployed cell.
- **`cluster_size` docstring** — added to the custom-op `forward`, accurately describing `None` → auto-dispatch vs. pinned int.

**Residual (non-blocking, author's call):** `enable_smem_cache` and its ~150 lines remain unwired (zero call sites). The author is keeping it in the kernel for near-future reuse since it doesn't yet give a stable perf win — acceptable, but please track it so it doesn't silently rot uncompiled. A follow-up that either lands a test enabling it or removes it would be cleaner.

LGTM. The OOB clamp and the corrected release/acquire pairing close the only two correctness-adjacent items. Nice work on the single/multi-CTA unification.

PR #15201, opened 2026-06-10
[None][perf] CuTeDSL MegaMoE: eliminate per-launch workspace memset o…

Barry-Delaney:
…verhead

Two optimizations that reduce CuTeDSL MegaMoE forward latency by up to 2.7x at large token counts (DSv4 shapes, B200 8-GPU EP=8):

1. Targeted workspace zeroing: the per-launch `_zero_local_workspace_preserving_phase` was zeroing the entire ~1.2 GB local workspace (minus the 4-byte nvlink_barrier_counter). Only ~0.57 MB of atomic counters and small safety-critical data regions actually need resetting.  Atomic counters (l1_arrival_count, expert_send_count, grid_sync_counter, fc1_done_counter, fc2_done_counter, load_balance_counter) must start at zero; l1_topk_weights_buffer and token_src_metadata must be zero for MMA padding rows to prevent stale data from corrupting FC2 → combine scatter.  The bulk data regions (l1_token_buffer, l1_sf_buffer, fc1_output, fc1_output_sf) are overwritten by dispatch/FC1 before being read.  This eliminates ~3 ms/fwd of memset at 32768T.

2. Fuse bf16->fp32 cast into sum: replace `combine_output.to(fp32).sum(dim=1)` with `combine_output.sum(dim=1, dtype=fp32)`, eliminating a full-size bf16->fp32 materialization kernel (~300 us/layer).

3. Cache `get_regions()` return value to avoid re-creating symmetric memory tensor views on every forward call, which is incompatible with CUDA-graph capture (cudaErrorStreamCaptureUnsupported).

Measured (min latency, CG, balanced routing, 8×B200):
  32768T: 5.78 ms -> 2.15 ms (2.69x)
  16384T: 2.98 ms -> 1.16 ms (2.57x)
   8192T: 1.56 ms -> 0.65 ms (2.40x)
   4096T: 0.85 ms -> 0.40 ms (2.10x)

### DIFF
--- a/tensorrt_llm/_torch/custom_ops/cute_dsl_megamoe_custom_op.py
+++ b/tensorrt_llm/_torch/custom_ops/cute_dsl_megamoe_custom_op.py
@@ -374,27 +374,29 @@ if IS_MEGAMOE_OP_AVAILABLE:
         _MEGAMOE_LOCAL_WORKSPACE_CACHE[cache_key] = local_workspace
         return local_workspace
 
-    def _zero_local_workspace_preserving_phase(local_workspace, kernel) -> None:
-        """Per-launch zero of the local workspace that PRESERVES the
-        self-priming ``nvlink_barrier_counter`` region (multi-rank EP path).
+    # Counters + small data regions that must be zeroed every launch.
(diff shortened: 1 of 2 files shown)


Barry-Delaney:
Will not merge after discussion. Closing.

PR #15213, opened 2026-06-10
[None][chore] Fix lock_infra_error

yufeiwu-nv:
Introduced a new helper function to identify lock-infrastructure errors, improving the robustness of the config_file_lock context manager. This change allows for better handling of temporary directory fallbacks during lock acquisition failures, ensuring that exceptions are properly propagated and logged.

Signed-off-by: yufeiwu-nv <230315618+yufeiwu-nv@users.noreply.github.com>

@coderabbitai summary

## Problem

`config_file_lock()` re-raises `filelock.Timeout` instead of using its tempdir fallback. The errno-narrowing added in #11960 — `isinstance(e, OSError) and e.errno not in {EACCES, EPERM, ENOLCK, ESTALE}` — was meant to let non-lock `OSError`s propagate. But `filelock.Timeout` is an `OSError` subclass with `errno=None`, so it satisfies that condition and gets re-raised, defeating the very lock-acquisition-timeout fallback the function is supposed to provide.

## Impact

When multiple ranks load a `trust_remote_code` model concurrently (tp/ep > 1), they contend on the single global `_remote_code.lock`. The ranks that time out crash during executor init and trigger `MPI_ABORT` — observed on the `deepseek_r1_0528_fp4 ... ep:4-tp:4` perf test.

## Fix

Refactor `config_file_lock` into a single-yield context manager that guards only the `acquire()` call. The `yield` is moved into `else` + `finally release`, so exceptions raised by the caller body (e.g. HF `RepositoryNotFoundError`, also an `OSError` subclass) propagate cleanly without a second yield. Fallback-eligible failures are now selected via `isinstance` — matching `filelock.Timeout` and `PermissionError` explicitly, plus NFS errnos `ENOLCK`/`ESTALE`.

## Verification

Reproduced with a real-`filelock` multi-process contention test (1 holder + 3 workers): the shipped logic makes all waiting workers raise `Timeout`, while the fix lets them all fall back and succeed.

## PR Checklist

## GitHub Bot Help


yufeiwu-nv:
hi @juney-nvidia , asked multiple times but still no response from @NVIDIA/trt-llm-torch-devs .Could you please help it? Thanks

tensorrt_llm/_torch/model_config.py
```diff
@@ -92,6 +92,19 @@ def _unified_kv_pool_includes_mamba(
     return not (use_split_pool or use_spec)
 
 
+def _is_lock_infra_error(exc: BaseException) -> bool:
+    """Whether exc indicates broken lock infrastructure (not mere contention)."""
+    # filelock.Timeout is contention, not broken infra, so it is not eligible.
+    if isinstance(exc, filelock.Timeout):
```
yuxianq: We only catch PermissionError and OSError when calling this function, it seems that checking filelock.Timeout is unnecessary.

yuxianq (approved):

tensorrt_llm/_torch/model_config.py
yufeiwu-nv: thanks, I will remove dead codes

Issue #15178, opened 2026-06-09
[Bug] Hybrid Mamba KV-cache estimation clamps max_tokens to 0 on integrated/unified-memory GPUs (regression in 1.3.0rc15, commit 091ad7b0)


janbernloehr:
### System Info

- TensorRT-LLM 1.3.0rc15 through 1.3.0rc17 (fails); 1.3.0rc14 is the last good version.
- GPU: integrated / unified-memory device, compute capability 12.1 (GB10-class, e.g. DGX Spark), arch linux/arm64. CUDA 13.1, driver 580.126.x, PyTorch 2.11.0a0, container nvcr.io/nvidia/tensorrt-llm/release:1.3.0rc17.
- Contrast device that passes: discrete RTX Pro 6000 Blackwell (dedicated VRAM), same TRT-LLM 1.3.0rc17.

### Who can help?

Maintainers of the PyTorch-flow KV-cache memory estimator (tensorrt_llm/_torch/pyexecutor/_util.py: KvCacheCreator._get_token_num_for_estimation / CacheCost).

### Information

- [x] The official example scripts (standard trtllm PyTorch-flow serving)
- [x] My own modified scripts

### Tasks

- [x] My own task: serve a hybrid Mamba+attention model on an integrated GPU.

### Reproduction

1. On an integrated/unified-memory GPU (compute capability 12.x, GB10/DGX Spark), serve a ~31B hybrid Mamba+attention model (attention window 3000 + mamba layers) in bf16, TP=1, PyTorch flow, with kv_cache_config.enable_block_reuse=false.
2. The ~58.8 GB bf16 weights are mmap'd from safetensors. On integrated GPUs TRT-LLM keeps them in shared host memory and logs: "Skipping device transfer from cpu to cuda on integrated GPU to conserve shared memory."
3. KV-cache estimation runs.

### Expected behavior

On a discrete GPU with the same rc17 binary and identical config, the estimation dry run yields {3000: (512, 0), -2147483647: (512, 0)} and the server starts successfully (exit 0).

### actual behavior

[...]

PR #12301, opened 2026-03-18
[#15793][perf] Optimize KV cache for unified memory systems (DGX Spark)

mihai-chiorean:
## Summary

Fixes #15793

On unified-memory systems like DGX Spark / GB10, CPU and GPU share the same physical LPDDR memory pool. The current KV cache sizing path can still treat GPU memory and host memory as separate tiers when `host_cache_size` is nonzero, but there is no separate host DRAM tier to offload to on these systems. `cudaMemGetInfo` already reports the unified pool, so the secondary host-cache budget should be folded away instead of tracked as a separate pool.

This patch adds integrated-device detection and folds the host KV cache tier on unified-memory systems:

1. **Detect true unified-memory devices**: Add `isUnifiedMemorySystem()` in C++ using `cudaDevAttrIntegrated`. This intentionally does not use `cudaDevAttrPageableMemoryAccess`, which can also be true on HMM-enabled discrete GPUs.

2. **Fold Python `host_cache_size` defaults**: Add `unified_memory_detected` to `KvCacheConfig`, auto-detect via `is_device_integrated()`, and force `host_cache_size` to `0` when unified memory is detected.

3. **Defensive C++ budget guard**: In `calculateFreeMemBytes()`, force `freeSecondaryMemBytes = 0` on unified-memory systems if C++ executor users bypass the Python validator and still provide a nonzero `host_cache_size`.

All changes are gated on integrated-device detection and preserve existing behavior on discrete GPU systems.

## Test Coverage

- Verified `is_device_integrated()` returns true on DGX Spark GB10 / SM121 with about 121.7 GB visible memory.
- Verified `KvCacheConfig` auto-detects the unified-memory device path.
- Pre-commit checks pass.
- Existing KV cache manager unit tests require `LLM_MODELS_ROOT` and full model data, so they are deferred to NVIDIA CI.

## Test plan

- [x] Verify integrated-device detection on DGX Spark / GB10 / SM121.
- [x] Verify `KvCacheConfig` auto-detects the unified-memory device path.
- [x] Pre-commit checks pass (clang-format, codespell, DCO).
- [x] Build and run existing KV cache manager unit tests on discrete GPU — deferred to NVIDIA CI because model data is not available locally.
- [x] Run inference workload on DGX Spark — deferred to NVIDIA CI because full model weights are not available locally.

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
[...]

PR #15215, opened 2026-06-10
[#15178][fix] Fix unified-memory Mamba KV estimation

peter941221:
## Description

Refs #15178.

On integrated GPUs, the estimation dry run can start from a `mem_get_info()` budget that is already depressed by mmap-backed weights sharing the same physical memory pool.

When hybrid Mamba models use an affine `CacheCost`, `_get_token_num_for_estimation()` subtracts the recurrent-state intercept from that reduced budget and can clamp the provisional token cap to zero. That is enough to trip `assert max_blocks_per_seq > 0` in the attention-window path even though the later affine sizing still succeeds.

This change keeps the final affine sizing unchanged. It only relaxes the provisional estimation cap on integrated GPUs by dropping the affine intercept from the dry-run budget calculation.

The regression test covers the zero-clamp case by mocking an affine `CacheCost`, a small `mem_get_info()` budget, and an integrated device.

## Test Coverage

Validated in the matching TensorRT-LLM 1.3.0rc18 CUDA 13 / PyTorch 2.10 runtime with:

- `python -m pytest tests/unittest/_torch/executor/test_kv_cache_estimation.py -k integrated_gpu_estimation_ignores_affine_intercept -q`
- `python -m pytest tests/unittest/_torch/executor/test_kv_cache_estimation.py -q`

## PR Checklist

## GitHub Bot Help

## Dev Engineer Review

- Integrated-GPU KV-cache estimation ignores the fixed `CacheCost` intercept only during provisional token-cap calculation.
- Final affine KV sizing remains unchanged.
- No public API signatures changed.
- Discrete-GPU behavior remains unchanged.
- The fix addresses the zero-token clamp caused by temporarily reduced unified-memory availability from mmap-backed weights.
- The change does not address the separate `MambaHybridCacheManagerV2` quota error.

## QA Engineer Review

- Modified `tests/unittest/_torch/executor/kv_cache/test_kv_cache_estimation.py`.
- Added `test_integrated_gpu_estimation_ignores_affine_intercept()`.
- The test covers limited memory, an integrated device, and an affine `CacheCost` intercept that exceeds the available budget.
- No matching `test-db/` or `qa/` entry was found.
- Reported targeted tests, full tests, and DGX Spark hardware testing completed. Internal CI failures were also reported, so CI status needs follow-up.
- Coverage verdict: needs follow-up.

## Per-File QA Perspective

- `tensorrt_llm/_torch/pyexecutor/_util.py`: Verify that only the provisional integrated-GPU calculation omits the intercept. Verify that final sizing and discrete-GPU behavior remain unchanged.
- `tests/unittest/_torch/executor/kv_cache/test_kv_cache_estimation.py`: Covers the integrated-GPU zero-token-clamp regression. No matching `test-db/` or `qa/` list entry was found.


janbernloehr:
Confirmed on hardware — this fixes #15178 for us. We applied this exact change on top of `1.3.0rc18` and re-ran the workload that was failing: a ~31B hybrid Mamba+attention model (bf16, TP=1) served on a **DGX Spark (GB10, unified memory)**.

- **Before (stock rc15–rc18):** engine init aborts with `AssertionError: Impossible to fit in any sequence in kvCache` — the KV-cache estimation dry-run clamps `max_tokens` to 0 on the integrated device, so the attention window gets 0 primary blocks.
- **After (rc18 + this PR):** the dry-run no longer clamps to 0, the server starts, and the full benchmark runs to completion (`exit 0`). The `is_device_integrated()` gate fires as intended and the final affine KV sizing is unchanged, so this only relaxes the provisional dry-run cap.

Discrete-GPU (RTX Pro 6000) runs were never affected, consistent with the integrated-memory root cause. Thanks for the fix @peter941221 — would be great to land this. Happy to test a candidate build on DGX Spark if useful. cc @joyang-nv

peter941221:
Thanks for testing this on DGX Spark and confirming that it fixes the rc18 unified-memory KV-cache failure while leaving the final affine sizing unchanged.

tongyuantongyu:
@VALLIS-NERIA this seems to related to your #12896. Could you take a look?

VALLIS-NERIA (approved):

VALLIS-NERIA:
Hi @peter941221 , thank you for the fix. It looks good to me but I'm not quite clear about the cause. My understanding is:
The weights are mmap-ed to CPU memory space, which is unified memory in this case, and causes `mem_get_info` to get a much smaller number. After the estimation dry run, the weights are unloaded and we get a normal free memory number, right?

peter941221:
> Hi @peter941221 , thank you for the fix. It looks good to me but I'm not quite clear about the cause. My understanding is: The weights are mmap-ed to CPU memory space, which is unified memory in this case, and causes `mem_get_info` to get a much smaller number. After the estimation dry run, the weights are unloaded and we get a normal free memory number, right?

@VALLIS-NERIA Yes — on integrated/unified-memory devices, the dry-run sees a temporarily depressed `mem_get_info()` budget from mmap-backed weights, and subtracting the affine intercept can clamp the provisional token cap to 0. The final affine sizing is unchanged.

peter941221:
The public GitHub checks are green on the current head. The remaining failure is the internal `blossom-ci` run on `fc3b3fb` after `main` was merged into this branch, and I can't see the internal failure details from my side. Could you share the failing job or log snippet so I can check whether it comes from this patch or from the merged `main` changes?

mihai-chiorean:
@peter941221 nice catch on the affine-intercept clamp. fyi #12301 is sitting in the same UMA-detection territory — different files (cpp kvCacheManager + llm_args.py) so no merge conflict, but same symptom (zero/negative caps on UMA) from a different angle. Forces host_cache_size=0 via cudaDevAttrIntegrated. Probably complementary with yours, worth looking at together if a reviewer picks one up.

peter941221:
@mihai-chiorean 
Thanks, that makes sense. This PR is scoped to the Python-side dry-run path where the affine intercept can collapse the provisional token cap to 0 on integrated / unified-memory systems. #12301 looks complementary from the KV-cache-manager / host-cache-budget side, so I agree they are related but fixing different layers. I’d prefer to keep this patch narrowly scoped unless a reviewer wants them aligned more explicitly.

janbernloehr:
@peter941221 can you maybe merge main / rebase so that we can try to get this finally merged. THX

peter941221:
Merged the latest main into this branch in commit fe52a0bd6. Resolved the _util.py import conflict by preserving both the current main imports and the integrated-device fix, and moved the regression coverage to the current 	ests/unittest/_torch/executor/kv_cache/test_kv_cache_estimation.py path. The diff against current main remains limited to the intended two files. The merge commit is GPG-signed and DCO-signed.

janbernloehr:
Thanks @peter941221 for refreshing this PR! One follow-up for #15178: our testing on newer releases still encounters `ValueError: The V2 Mamba GPU cache quota is too small`.

This PR addresses the provisional token-cap calculation in the V1 estimation path. With NemotronH using `MambaHybridCacheManagerV2` from rc24 onward (via #16598), the V2 quota check needs separate attention; landing the V1 guard alone would not resolve that remaining failure.

Could a KV-cache owner advise whether to cover V2 here or track it in a linked follow-up? It would be helpful to make that scope explicit while moving this fix toward merge. Happy to help validate a candidate on Spark. Thanks!

---
_Prepared with assistance from Codex (GPT-6)._

janbernloehr:
@peter941221 can you address the pre-commit ci failures? THX

peter941221:
Applied the pre-commit fix in d2f6f56c27. The only failing hook on the previous head was ruff-format; the formatted test file now passes the full pre-commit suite locally. Please trigger CI again on the new head.

PR #15227, opened 2026-06-10
[TRTLLM-1234][feat] initial profiling of worker wall-clock latency on grid of isl/batch size conditions (api-compatible)

sachalmalick:
## Summary by CodeRabbit

* **New Features**
  * Added a self-benchmark mode for serving, with configurable warmup, prefill, decode, and output settings.
  * Requests can now carry self-benchmark metadata so benchmark traffic is tracked consistently across the system.
  * Benchmark results are written out as a structured report after completion.

* **Bug Fixes**
  * Improved handling so benchmark requests are separated from normal traffic and measured correctly.
  * Added validation to prevent unsupported runtime configurations when self-benchmarking is enabled.

## Description

Adds startup self-benchmarking for the PyTorch executor path. The benchmark runs synthetic prefill and/or decode points inside the scheduler loop, writes the collected iteration FWP metrics to a JSON file.
The goal of this PR is to let the engine profile itself on startup, so that consumers of this - namely Dynamo planner can use this to initialize the perf prediction regression model and make smarter online scaling decisions.  See this equivalent vLLM PR for more details [here](https://github.com/ai-dynamo/dynamo/pull/7779) 

Self Benchmarking is gated by the `SelfBenchmarkConfig`, `trtllm-serve` self-benchmark flags.

## Motivation

In pursuit of intelligent LLM inference autoscaling that can maximize goodput for some SLAs such as max `TTFT`, `TPOT` - Autoscalers like Dynamo Planner aim to make predictions about how a LLM inference server would perform under different deployment configurations and traffic conditions.  

By emitting `ForwardPassMetrics` one can observe the actual wall clock latency for the model worker forward at every iteration of the scheduler.  

The goal of Self-benchmarking is to collect `ForwardPassMetrics` on server start - this will allow downstream consumers like Dynamo Planner to bootstrap performance prediction models with real data and ultimately make more accurate predictions and therefore better scaling decisions.

What makes this feature valuable when implemented in the inference engines themselves, is that we can gather more data a given benchmarking time budget, by skipping actual prefill during decode benchmarking.  We can still measure how decode performs for a given ISL *by using synthetic kv cache block placement*.  Based on our tests, this still gives us an accurate estimate of what decode would look like if prefill had actually run (falling within 3% of FPMs observed with real prefill).

**Callout**: While this is can be considered an 'invasive' change as we mutate some scheduler state, this is gated by the self-benchmarking flags which default to off.

## More on how it works

The benchmark sweeps a configurable grid of operating points:

- Prefill: input sequence lengths / ISLs.
- Decode: context lengths and batch sizes.

### TRT-LLM Self-Benchmark PR Results

Common setup:
- Runtime image: `public.ecr.aws/m4g4v5k2/sachalm-test/d-test:trt-selfbenchmarking`
- Model: `Qwen/Qwen3-0.6B`
- Offline HF cache: `HF_HOME=/model-store`, model loaded from `/model-store/hub`
- Attention backend: `FLASHINFER`
- CUDA graphs: enabled via the qwen3 TRT-LLM config
- Benchmark grid: `prefill=2`, `kv-read=2`, `decode-length=2`, `decode-batch=2`, `warmup=0`

| Experiment | Topology | Result | Artifact |
|---|---:|---|---|
| `01-agg-tp1` | Aggregated, TP1 | Complete, valid, 6 points: 2 prefill + 4 decode | `experiments/self-benchmarking/trtllm/results/01-agg-tp1-trtllm_self_benchmark.json` |
| `02-agg-tp2` | Aggregated, TP2 | Complete, valid, 6 points: 2 prefill + 4 decode | `experiments/self-benchmarking/trtllm/results/02-agg-tp2-trtllm_self_benchmark.json` |
| `07-disagg-tp4` | Disaggregated, TP4 prefill + TP4 decode | Complete, valid. Decode emitted 4 decode points; prefill emitted 2 prefill points | `experiments/self-benchmarking/trtllm/results/07-disagg-tp4-decode-trtllm_self_benchmark.json`, `experiments/self-benchmarking/trtllm/results/07-disagg-tp4-prefill-trtllm_self_benchmark.json` |

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -3770,15 +3770,19 @@ class PyExecutor:
                             timeout))
 
             if benchmark_active:
+                # During self-benchmarking, defer real user requests
+                # (accumulate them) and forward only control/special items.
+                # Synthetic benchmark PREFILL requests are NOT injected here:
+                # they are built per-rank in _fetch_and_activate_new_requests so
+                # every TP rank deterministically produces the same forward
+                # batch. Injecting them into rank 0's broadcast queue would
+                # double-inject on non-rank-0 ranks (once via broadcast, once
+                # locally) and desync the per-rank state machine.
                 for req_item in queued_requests:
                     if req_item.is_normal_request:
                         self.request_accumulated.append(req_item)
                     else:
                         new_requests.append(req_item)
-                if len(new_requests) == 0:
-                    new_requests.extend(
-                        self.self_benchmark.make_prefill_queue_items(
-                            self.active_requests, waiting_queue))
             else:
                 new_requests.extend(queued_requests)
 
@@ -3972,8 +3976,19 @@ class PyExecutor:
 
         self.active_requests.extend(validated_requests)
         if self.self_benchmark is not None and self.self_benchmark.active:
-            benchmark_requests = self.self_benchmark.make_decode_requests(
+            # Self-benchmark injection runs identically on EVERY TP rank here
+            # (this method is invoked per-rank by the main loops). Both prefill
+            # and decode requests are constructed locally and deterministically
+            # from the grid point index, so each rank advances the same state
+            # machine and produces the same forward batch in lockstep -- no
+            # per-iteration broadcast/collective is needed. Prefill is attempted
+            # first; make_prefill_requests returns [] when the next point is a
+            # decode point (and vice versa), so at most one path injects.
+            benchmark_requests = self.self_benchmark.make_prefill_requests(
                 self.active_requests, self.waiting_queue)
+            benchmark_requests.extend(
+                self.self_benchmark.make_decode_requests(
+                    self.active_requests, self.waiting_queue))
             self.active_requests.extend(benchmark_requests)
             validated_requests.extend(benchmark_requests)
         return validated_requests
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -1,6 +1,7 @@
 import dataclasses
 import datetime
 import functools
+import json
 import os
 import threading
 import time
@@ -84,6 +85,7 @@ from .scheduler import (RequestScheduler, ScheduledRequests,
                         SerializableSchedulerOutput, WaitingQueue,
                         create_waiting_queue)
 from .scheduler.adp_router import ADPRouter
+from .self_benchmark import SelfBenchmark
 
 if TYPE_CHECKING:
     from ray.actor import ActorHandle
@@ -623,6 +625,8 @@ class PyExecutor:
         # Waiting queue for requests that have been fetched but not yet scheduled
         self.waiting_queue: WaitingQueue = create_waiting_queue(
             waiting_queue_policy)
+        self.self_benchmark = SelfBenchmark(
+            self) if self.llm_args.self_benchmark_config is not None else None
 
         self.control_request_barrier = threading.Event()
         self.control_action_done = threading.Event()
@@ -1336,7 +1340,8 @@ class PyExecutor:
 
     @staticmethod
     def _is_stats_dummy_request(req) -> bool:
-        return bool(getattr(req, "is_dummy", False))
+        return bool(getattr(req, "is_dummy", False)) and not bool(
+            getattr(req, "is_self_benchmark_request", False))
 
     def _collect_scheduled_batch_stats(
             self, scheduled_batch: ScheduledRequests) -> ScheduledBatchStats:
@@ -1904,6 +1909,19 @@ class PyExecutor:
                                         batch_state.scheduled_requests,
                                         micro_batch_id,
                                         batch_state.scheduled_batch_stats)
+        if self.self_benchmark is not None and self.self_benchmark.active:
+            stats_dict = json.loads(stats.to_json_str())
+            if host_step_time_ms is not None:
+                stats_dict["hostStepTimeMS"] = host_step_time_ms
+            if prev_device_step_time_ms is not None:
+                stats_dict["prevDeviceStepTimeMS"] = prev_device_step_time_ms
+            is_overlap_like = (not self.disable_overlap_scheduler
+                               or self.dist.pp_size > 1)
+            stats_dict["schedulerMode"] = (
+                "overlap" if is_overlap_like else "non_overlap")
+            if self.self_benchmark.observe_iteration(
+                    batch_state.scheduled_requests, stats_dict):
+                return
         if self.enable_attention_dp:
             self._adp_iter_stats.queue(
                 stats,
@@ -3725,18 +3743,24 @@ class PyExecutor:
         else:
             timeout = datetime.timedelta(0)
 
-        # Fetch requests from rank 0
         new_requests = []
-        if self.dist.rank == 0:
-            # Process accumulated requests that were queued during control request handling.
-            if len(self.request_accumulated) != 0:
-                new_requests.extend(self.request_accumulated)
-                self.request_accumulated.clear()
-                # Reset timeout to 0 to avoid hanging when no new requests are available
-                timeout = datetime.timedelta(0)
-            with self.hang_detector.pause():
+        if self.self_benchmark is not None and self.self_benchmark.active:
+            if self.dist.rank == 0:
                 new_requests.extend(
-                    self.executor_request_queue.get_from_request_queue(timeout))
+                    self.self_benchmark.make_prefill_queue_items(
+                        self.active_requests, waiting_queue))
+        else:
+            # Fetch requests from rank 0
+            if self.dist.rank == 0:
+                # Process accumulated requests that were queued during control request handling.
+                if len(self.request_accumulated) != 0:
+                    new_requests.extend(self.request_accumulated)
+                    self.request_accumulated.clear()
+                    # Reset timeout to 0 to avoid hanging when no new requests are available
+                    timeout = datetime.timedelta(0)
+                with self.hang_detector.pause():
+                    new_requests.extend(
+                        self.executor_request_queue.get_from_request_queue(timeout))
 
         # Broadcast requests and handle Python objects
         new_requests, py_request_objects = self.request_broadcaster.broadcast(
@@ -3927,6 +3951,11 @@ class PyExecutor:
         ]
 
         self.active_requests.extend(validated_requests)
+        if self.self_benchmark is not None and self.self_benchmark.active:
+            benchmark_requests = self.self_benchmark.make_decode_requests(
+                self.active_requests, self.waiting_queue)
+            self.active_requests.extend(benchmark_requests)
+            validated_requests.extend(benchmark_requests)
         return validated_requests
 
     def _add_kv_cache_events(self):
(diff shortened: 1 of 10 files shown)


tensorrt_llm/_torch/pyexecutor/py_executor.py
venkywonka, line 1893: is there a particular reason why `gpu_forward_time_ms` is never captured in `stats_dict`?

tensorrt_llm/llmapi/llm_args.py
```diff
@@ -2340,6 +2340,55 @@ def validate_histogram_buckets(cls, v: Optional[List[float]],
         return v
 
 
+class SelfBenchmarkConfig(StrictBaseModel):
```
hchings, line 2343: nit: I find the whole self-benchmark naming confusing, though I understand what it does after reading MR descr and vllm changes. Would prefer sth more informative.

tensorrt_llm/commands/serve.py
```diff
@@ -858,6 +893,54 @@ def convert(self, value: Any, param: Optional["click.Parameter"],
                   "Applied by Efficient Video Sampling (EVS). "
                   "None disables EVS, values in [0, 1) enable pruning.",
                   "prototype"))
+@click.option(
```
hchings, line 896: Can we have better UX here, such as a single flag to read a config yaml rather than flatten all knobs in SelfBenchmarkConfig and add many new flags to serve?

tests/unittest/_torch/executor/test_self_benchmark.py
```diff
@@ -0,0 +1,480 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
```
hchings, line 4: The new test file needs a yaml entry to actually run. Please check whether `tests/unittest/_torch/pyexecutor/` is a more suitable location.

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
```diff
@@ -929,32 +929,49 @@ def drafting_loop_wrapper(model):
...
-        )
+        executor_llm_args = llm_args
+        original_model_engine_llm_args = None
+        if estimating_kv_cache and llm_args.self_benchmark_config is not None:
+            # The temporary KV-estimation executor disables iter stats, so
+            # self-benchmark points cannot observe completion there.
+            logger.debug(
+                "Disabling self_benchmark_config for temporary KV-cache "
+                "estimation executor")
+            executor_llm_args = copy.copy(llm_args)
```
hchings, line 940: The copy.copy(llm_args) + model_engine.llm_args = ... + original_model_engine_llm_args restore pattern is fragile. Can you explore a implementation where model_engine.llm_args is not reassigned mid-flow?

hchings (commented):
I'd like to have someone familiar with `trtllm-bench` cc @FrankD412 & `trtllm-serve` to review as well.

tensorrt_llm/_torch/pyexecutor/py_executor.py
sachalmalick: addressed!

tensorrt_llm/llmapi/llm_args.py
sachalmalick: hmm, what about `initial-profiling`?

tensorrt_llm/commands/serve.py
sachalmalick: addressed!

tests/unittest/_torch/executor/test_self_benchmark.py
sachalmalick: addressed!

tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
sachalmalick: addressed!

tensorrt_llm/_torch/pyexecutor/self_benchmark.py
FrankD412, line 58: I'd need someone who owns the PyExecutor module to chime in, but I have reservations about integrating a self-benchmark directly into the executor/scheduler itself. I'm just coming back from bereavement leave so I apologize if I don't have the full context or am misunderstanding something. Here are my thoughts:

1. The Dynamo/vLLM version of this self-benchmark seems to utilize a derived version of the scheduler directly (the `InstrumentedScheduler`) which returns the forward pass information (via ZMQ if I understand correctly). This structure lets the benchmarking loop/flow ownership sit with Dynamo rather than vLLM.

2. I don't think this workflow should live directly in the executor. In my mind, TRT-LLM should provide the configurability to enable forward pass statistics to be returned and then a workflow like this should be in a separate utility. I'm okay with that being in `trtllm-bench` under its own subcommand if you would like to provide the ability native to TRT-LLM. But I definitely think that the biggest thing is to allow a controller higher in the stack to control this flow.

tensorrt_llm/commands/serve.py
FrankD412, line 330: Please correct me if I'm mistaken, but this benchmark mode simply launches serve and runs ahead of accepting requests. It isn't changing any settings in the start of the server. So essentially you start `trtllm-serve` with this config, get the JSON data out, and kill the server?

Augmenting `trtllm-serve` in this way, while I understand why, kind of breaks this entrypoint being production/customer facing since we're introducing a benchmark flow here.

FrankD412:
Generally, I understand the intent. My concern with this implementation is that it couples the self-benchmark too deeply into the executor itself whereas the vLLM implementation makes a distinct interface that Dynamo can cleanly use.

~~I'm not sure how this would work with the KV cache events -- though that looks like it relies on a benchmarking flag within a request to work. While I understand the intent, to me it's not the most clear what the benchmark request lifecycle looks like because there are paths in the request utilities that explicitly check for a benchmark flag, etc. So I would say that this feature also needs documentation because having a more explicit understanding of this flow will help others contribute/understand it when they work with it.~~

Edit: I'm mistaken on the request flagging itself as a benchmark request.

Some fundamental questions that could be included in documentation:
1. What is this flow, what are its goals?
2. How does this flow work methodology-wise and functionally? (this might help clarify why it's coupled to the executor)
3. How do you run this benchmark? (if we keep the `trtllm-serve` flow, documentation on how to use it)

FrankD412:
> Generally, I understand the intent. My concern with this implementation is that it couples the self-benchmark too deeply into the executor itself whereas the vLLM implementation makes a distinct interface that Dynamo can cleanly use.
> 
> ~I'm not sure how this would work with the KV cache events -- though that looks like it relies on a benchmarking flag within a request to work. While I understand the intent, to me it's not the most clear what the benchmark request lifecycle looks like because there are paths in the request utilities that explicitly check for a benchmark flag, etc. So I would say that this feature also needs documentation because having a more explicit understanding of this flow will help others contribute/understand it when they work with it.~
> 
> Edit: I'm mistaken on the request flagging itself as a benchmark request.
> 
> Some fundamental questions that could be included in documentation:
> 
>     1. What is this flow, what are its goals?
> 
>     2. How does this flow work methodology-wise and functionally? (this might help clarify why it's coupled to the executor)
> 
>     3. How do you run this benchmark? (if we keep the `trtllm-serve` flow, documentation on how to use it)

Adding onto this, I had Claude provide a more complete summary:

## Benchmark Request Lifecycle

### Phase 0 — Startup: grid built, output invalidated

When `PyExecutor` initializes, it creates `SelfBenchmark(self)` (`py_executor.py:625`). The constructor immediately builds the full sweep grid (`_build_grid`) and writes a `{"valid": false, "status": "running"}` sentinel to the output path. This prevents a stale file from a prior crashed run from being read as valid by Dynamo.

---

### Phase 1 — Request injection (per scheduler iteration)

Each iteration, `_fetch_and_activate_new_requests` is called on every TP rank. Two things happen here simultaneously:

**Real requests are deferred.** During benchmarking, rank 0 drains its request queue with `get_nowait()` (non-blocking) instead of the normal blocking `get_from_request_queue()`. Any real user requests that arrive are pushed into `request_accumulated` — held aside until the sweep finishes (`py_executor.py:3760–3775`).

**Synthetic requests are injected.** After the normal request path, `make_prefill_requests` or `make_decode_requests` is called directly on every TP rank — bypassing the rank-0 queue and broadcast entirely. This is the key TP correctness mechanism: every rank constructs the same `LlmRequest` deterministically from the grid point index (same token ids `[1]*isl`, same request id via pure arithmetic, same `cache_salt`). The `is_self_benchmark_request = True` and `py_self_benchmark_point_id = point.index` flags are set on the request here.

**Prefill vs. decode injection differ significantly:**

- **Prefill:** builds a `Request(input_token_ids=[1]*isl, max_tokens=1)` and converts it through `executor_request_to_llm_request` normally. The KV cache fills organically as a real prefill would.
- **Decode:** calls `kv_cache_manager.add_dummy_requests(...)` with `is_gen=True`. This directly allocates KV blocks without running prefill — synthetic block placement that simulates the memory footprint of `context_length` tokens having already been processed. Real decode attention access patterns run over the allocated blocks, but with zero prefill compute cost.

---

### Phase 2 — Forward pass

The synthetic request goes through the normal scheduler → model forward path. Nothing in the forward pass knows these are benchmark requests — they are shaped like real requests (correct beam width, correct KV block footprint). For prefill-with-KV-reads (`prefill_seed` + `prefill` pairs), the `cache_salt` on the seed request is reused on the actual prefill point, so the prefill request finds the seed's blocks already cached — measuring the real KV read cost.

---

### Phase 3 — Stats observation

After the forward pass, `_process_iter_stats` fires. If benchmarking is active, the iteration stats dict is augmented with `hostStepTimeMS`, `prevDeviceStepTimeMS`, and `schedulerMode`, then passed to `self_benchmark.observe_iteration()` (`py_executor.py:1909–1920`).

`observe_iteration` validates the batch is benchmark-only (no mixed real and synthetic batches), records cache hit validation for prefill points, and appends the stats to `_current.stats`. A point is marked complete when the stats show `numContextRequests > 0` (for prefill) or `numGenRequests > 0` (for decode). Benchmark iterations return early — they never reach the normal `enable_iter_perf_stats` path or the public stats buffer.

---

### Phase 4 — Request termination

In `_handle_responses`, benchmark requests are intercepted before the normal response path (`py_executor.py:5361–5367`): `perf_manager.append_step_metrics` is called to capture per-request timing, then the request is pushed to `requests_to_terminate` — it is never returned to any external caller.

---

### Phase 5 — Sweep completion

When the grid is exhausted (`_peek_next_point` returns `None`), `_finish()` is called. This calls `write_output()`, which atomically writes the final JSON (via `tempfile.mkstemp` + `os.replace`) with `{"valid": true, "status": "complete"}`. Real requests accumulated in `request_accumulated` are then drained normally and the server begins serving.

---

### Summary

```text
PyExecutor init
  └─ SelfBenchmark.__init__: build grid, write {valid:false} sentinel

Per scheduler iteration (benchmark active):
  ├─ _fetch_and_enqueue_requests:  drain real requests → request_accumulated (deferred)
  ├─ _fetch_and_activate_new_requests:
  │    ├─ make_prefill_requests  → LlmRequest([1]*isl, is_self_benchmark=True)
  │    └─ make_decode_requests   → add_dummy_requests (synthetic KV blocks, no prefill)
  ├─ Scheduler + model forward   (benchmark requests look like real ones)
  ├─ _process_iter_stats:
  │    └─ observe_iteration → record stats, validate cache hits, mark point complete
  └─ _handle_responses:
       └─ benchmark request → append_step_metrics → terminate (never returned to caller)

Grid exhausted:
  └─ write_output: atomic write {valid:true, status:complete}
  └─ request_accumulated drained → normal serving resumes

schetlur-nv:
Superceded by https://github.com/NVIDIA/TensorRT-LLM/pull/16191

PR #15229, opened 2026-06-10
[None][feat] VisualGen: async mp4 encode + fixed noise latent via env vars

wu6u3tw:
## Summary

Video generation responses currently block the async server coroutine on
`output.save()` (libx264 mp4 encode, ~1.7 s/video) before the next request
can start. This adds an opt-in to offload the encode to a thread-pool executor,
so the server can begin GPU generation for the next request while the previous
video is being encoded.

**Change:** `TRTLLM_VIDEO_ASYNC_ENCODE=1` env var switches `output.save()` to
run via `asyncio.get_running_loop().run_in_executor(None, ...)`, making the
await non-blocking to the event loop.

## Motivation

On GB300, each mp4 encode takes ~1.7 s (CPU libx264; no hardware encoder on
datacenter Blackwell). With the default synchronous path, this sits on the
critical cycle path — generation is idle during encode. With async overlap and
≥2 concurrent requests in flight, encode is fully hidden: measured +0.5%
end-to-end QPS at 4 GPUs in MLPerf Wan2.2-T2V-A14B offline scenario.

The flag is off by default (`"0"`), so existing behavior is unchanged.

## Testing

- Ran MLPerf Wan2.2-T2V-A14B offline benchmark (DP4, GB300):
  - Without flag: 0.0354 QPS
  - With `TRTLLM_VIDEO_ASYNC_ENCODE=1`: 0.0368 QPS (+0.5%)
- Flag=0 path unchanged; no regression on existing serve behavior.

## Release Notes

* **Performance**
  * Video generation endpoints now support optional asynchronous encoding to reduce blocking operations. When enabled via environment configuration, encoding runs in a background thread, improving endpoint responsiveness while maintaining backward compatibility.


wu6u3tw:
@chang-l for review as well.

wu6u3tw:
@zhenhuaw-me for fixed_latent from env variable change.

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
```diff
@@ -108,6 +109,20 @@ def __init__(self, pipeline_config):
...
 
+        # Fixed latent for reproducible benchmarking (e.g. MLPerf).
+        # Set TRTLLM_VIDEO_FIXED_LATENT_PATH to a .pt file containing a pre-sampled
+        # noise tensor; it will be used in place of freshly sampled random latents for
+        # all T2V requests.  Loaded once at server startup, reused across requests.
+        self._fixed_latent: Optional[torch.Tensor] = None
+        _fixed_latent_path = os.environ.get("TRTLLM_VIDEO_FIXED_LATENT_PATH")
+        if _fixed_latent_path:
+            self._fixed_latent = torch.load(_fixed_latent_path, weights_only=True)
+            logger.info(
```
zhenhuaw-me:
```suggestion
            logger.warn(
```

(not a blocker) Does it make sense to use a higher logging level?

tensorrt_llm/serve/openai_video_routes.py
```diff
@@ -160,11 +160,20 @@ async def openai_video_generation_sync(self, raw_request: Request) -> Response:
             resolved_fmt = resolved_encoder_fmt
             batch_size = output.video.shape[0] if output.video.dim() == 5 else 1
             paths_in = [self.media_storage_path / f"{video_id}_{i}" for i in range(batch_size)]
-            saved_paths = output.save(
-                paths_in,
+            _save_kwargs = dict(
                 format=resolved_fmt,
                 frame_rate=output.frame_rate or request.frame_rate or params.frame_rate,
             )
+            if os.environ.get("TRTLLM_VIDEO_ASYNC_ENCODE", "0") == "1":
```
JunyiXu-nv: Why not use `openai_video_generation_async` for asynchronous video gen?

zhenhuaw-me: Are we trying to bypass the median encoding time? If the client doesn't need media encoding it can either:
1. query the `generation` perf metric
2. query the format `safetensor` or `pt` (enabled by https://github.com/NVIDIA/TensorRT-LLM/pull/14733)

wu6u3tw: As discussed offline, we need mp4 in accuracy eval.

tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
wu6u3tw: Modified.

tensorrt_llm/serve/openai_video_routes.py
wu6u3tw: Using openai_video_generation_async wouldn't help because:                                                                                                                                                 
  1. MLPerf's harness hits the sync endpoint. It would change the benchmark behavior entirely.                                                                                             
  2. The async endpoint's background task runs encode in the same event loop anyway; it doesn't overlap with the next request's GPU work unless the encode is also offloaded to a thread.  
  We need the overlap with the next req.

wu6u3tw: Not bypassing encoding — we need mp4 for MLPerf accuracy evaluation (FVD metric computed on decoded frames). The goal is to maximize QPS by overlapping the CPU encode of request N with GPU generation of request N+1, so the encode time doesn't sit on the critical path. `safetensor`/`pt` would work for perf-only runs but accuracy mode requires mp4 output.

wu6u3tw: @JunyiXu-nv to check again.

wu6u3tw: @JunyiXu-nv please let me know your preference. @chang-l suggested that async video enc to set as defualt.

JunyiXu-nv: Reasonable. It's ok to make it default.

JunyiXu-nv (approved):
LGTM

tensorrt_llm/serve/openai_video_routes.py
zhenhuaw-me: Like @chang-l mentioned, if no side effect, we may consider to "always async".

Moving forward we may consider queue and scheduler support, but we need to understand the MLPerf benchmarking requirements.

zhenhuaw-me (approved):

tensorrt_llm/serve/openai_video_routes.py
wu6u3tw: Modified

wu6u3tw:
@chang-l can you help me merge this PR. I don't have permission. Thanks

PR #15237, opened 2026-06-10
[TRTLLMINF-81][feat] Avoid failed runners on infra retry

dpitman-nvda:
## Summary by CodeRabbit

**Chores**
* Enhanced CI pipeline retry logic to intelligently track and exclude infrastructure nodes that previously failed during test execution.
* Applies to both SLURM and Kubernetes deployments, preventing retry attempts from reusing the same compute resources that previously caused failures.

## Description

In order to increase the likelihood of a successful retry in the event of infra failure (and equally to prevent retries from overloading an already suspect machine), we modify the job submissions to avoid runners that failed on previous runs.

## Test Coverage

N/A, this is a CI change

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/jenkins/L0_Test.groovy
+++ b/jenkins/L0_Test.groovy
@@ -1092,6 +1092,7 @@ def runLLMTestlistWithSbatch(pipeline, platform, testList, config=VANILLA_CONFIG
     // Create a unique suffix for the job name
     String customSuffix = "${env.BUILD_TAG}-${UUID.randomUUID().toString().replaceAll("-", "").substring(0, 6)}".toLowerCase()
     def jobUID = "${cluster.host}-multi_node_test-${customSuffix}"
+    def jobWorkspace = "/home/svc_tensorrt/bloom/scripts/${jobUID}"
     def disaggMultiNodeMode = stageName.contains("Disagg-PerfSanity")
     def aggMultiNodeMode = !disaggMultiNodeMode && nodeCount > 1 && stageName.contains("PerfSanity")
 
@@ -1106,7 +1107,6 @@ def runLLMTestlistWithSbatch(pipeline, platform, testList, config=VANILLA_CONFIG
             def tarName = BUILD_CONFIGS[config][TARNAME]
             def llmTarfile = "https://urm.nvidia.com/artifactory/${ARTIFACT_PATH}/${tarName}"
             def llmPath = sh (script: "realpath .", returnStdout: true).trim()
-            def jobWorkspace = "/home/svc_tensorrt/bloom/scripts/${jobUID}"
             def resourcePathNode = "/tmp"
             def llmSrcNode = "${resourcePathNode}/TensorRT-LLM/src"
             def llmSrcLocal = "${llmPath}/TensorRT-LLM/src"
--- a/jenkins/L0_Test.groovy
+++ b/jenkins/L0_Test.groovy
@@ -2004,10 +2004,12 @@ def readSlurmWorkspaceFile(def pipeline, Map remote, String path, String stageNa
     }
 }
 
+// State flows through the mutated placementContext map (read by the retry
+// loop); this records into it and returns nothing.
 def recordSlurmPlacementContext(Map placementContext, String slurmJobID, def nodeList, String stageName)
 {
     if (placementContext == null) {
-        return null
+        return
     }
 
     if (slurmJobID) {
@@ -2020,7 +2022,6 @@ def recordSlurmPlacementContext(Map placementContext, String slurmJobID, def nod
         def jobLabel = slurmJobID ? "job ${slurmJobID}" : "job"
         echo "[INFRA-RETRY] ${stageName}: SLURM ${jobLabel} ran on node list(s): ${placementContext.lastSlurmNodeList}"
     }
-    return placementContext.lastSlurmNodeList
 }
 
 def captureSlurmWorkspaceMetadata(def pipeline, Map remote, String jobWorkspace, Map placementContext, String stageName)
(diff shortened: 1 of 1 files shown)


jenkins/L0_Test.groovy
```diff
@@ -1800,6 +1847,151 @@ boolean retryContextAllowsRetry(def pipeline, Map retryContext, Throwable error,
...
+        }
+    }
+    if (!hostNodes.isEmpty()) {
+        echo "[INFRA-RETRY] ${stageName}: recorded Kubernetes host node(s) for retry avoidance: ${hostNodes.join(', ')}"
+    }
+}
+
+def rememberAvoidedSlurmNodeLists(List avoidedNodeLists, def nodes, String stageName)
+{
+    def nodeLists = trtllm_utils.normalizeSlurmNodeLists(nodes)
```
tburt-nv, line 1975: What does "normalize" mean here?

jenkins/L0_Test.groovy
```diff
@@ -1800,6 +1847,151 @@ boolean retryContextAllowsRetry(def pipeline, Map retryContext, Throwable error,
...
+            def metadata = captureSlurmWorkspaceMetadata(pipeline, remote, jobWorkspace, placementContext, stageName)
+            capturedJobID = capturedJobID ?: metadata.slurmJobId
+            nodeList = metadata.nodeList
+            if (!capturedJobID && jobWorkspace) {
+                capturedJobID = readSlurmWorkspaceFile(pipeline, remote, "${jobWorkspace}/slurm_job_id.txt", stageName, 3)
+                recordSlurmPlacementContext(placementContext, capturedJobID, null, stageName)
+            }
+            if (!nodeList && jobWorkspace) {
+                nodeList = readSlurmWorkspaceFile(pipeline, remote, "${jobWorkspace}/slurm_node_list.txt", stageName, 3)
+            }
```
tburt-nv, line 2066: Shouldn't we return here? Why run sacct below if we already have the node list from `slurm_node_list.txt`?

jenkins/L0_Test.groovy
```diff
@@ -1800,6 +1847,151 @@ boolean retryContextAllowsRetry(def pipeline, Map retryContext, Throwable error,
...
+        placementContext.slurmJobId = slurmJobID
+    }
+
+    def normalized = trtllm_utils.normalizeSlurmNodeLists(nodeList)
+    if (!normalized.isEmpty()) {
+        placementContext.lastSlurmNodeList = normalized.join(' ')
+        def jobLabel = slurmJobID ? "job ${slurmJobID}" : "job"
+        echo "[INFRA-RETRY] ${stageName}: SLURM ${jobLabel} ran on node list(s): ${placementContext.lastSlurmNodeList}"
+    }
+    return placementContext.lastSlurmNodeList
```
tburt-nv: Why do we return this value? It doesn't seem to be referenced anywhere.

jenkins/L0_Test.groovy
```diff
@@ -3877,11 +4069,16 @@ def runKubernetesPodWithInfraRetry(Map opts = [:], pipeline, podSpec, containerN
     // failure caps the budget at 1 retry (attempt 2 is final), so the next
     // attempt must not suppress synthetic stage-fail XML / junit().
     def lastSeverity = null
+    def avoidedKubernetesHostNodes = []
     while (true) {
         attempt++
+        Map attemptPlacementContext = [:]
```
tburt-nv, line 4186: Why do we need both the placementContext and avoidedSlurmNodeLists?

jenkins/L0_Test.groovy
dpitman-nvda: SLURM can a mishmash of node types: bracket ranges, comma-separated, or plain hostnames. This function deduplicates and "canonicalizes" the node list for our purposes.

jenkins/L0_Test.groovy
dpitman-nvda: The return is a few lines below on L1960. That ensures we only run `sacct` if there's no node list file.

jenkins/L0_Test.groovy
dpitman-nvda: Good catch, will fix.

jenkins/L0_Test.groovy
dpitman-nvda: `avoided...` is an accumulated list of failed nodes across retries. `attempt...` is the per-run context: it tracks both the nodes to exclude (from `avoided`) and the node where the attempt actually ran. On failure, that "last attempt" is appended to the `avoided` list for future attempts.

tburt-nv: Couldn't we just push to `avoidedKubernetesHostNodes` inside `launchKubernetesPodWithPlacement` instead of mutating `attemptPlacementContext`?  The only time `avoidedKubernetesHostNodes` will be read again is if we need to retry, right? If we always push, I think we could rename `avoidedKubernetesHostNodes` to `attemptedKubernetesHostNodes`, and we simply avoid any nodes we've attempted before.

I think the same concept applies to SLURM. Could we just have an `attemptedSlurmNodeLists` that we append to at the end of the job, and use that in any retries? I suppose this might not work if the list is shared between stages, unless we can determine if the job failed before appending to `attemptedSlurmNodeLists`.

I don't think this is a blocker, but I hope we can simplify this in the future.

tburt-nv (approved):

jenkins/L0_Test.groovy
dpitman-nvda: +1, definitely can refactor this in a follow-up.

PR #14177, opened 2026-05-15
[https://nvbugs/6156492][fix] Fix disaggregated usage propagation

reasonsolo:
## Summary by CodeRabbit

* **New Features**
  * Added context response usage tracking and propagation across disaggregated serving phases.
  * Token usage now includes cached token accounting alongside prompt and completion tokens.
  * Context usage information is now carried through to generation response outputs.
  * Supports both streaming and non-streaming chat and completion endpoints.

* **Tests**
  * Updated existing tests to verify token usage data structure and propagation.
  * Added tests validating usage rewriting for generation postprocessors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/TensorRT-LLM/pull/14177)

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- If PR introduces API changes, an appropriate PR label is added - either `api-compatible` or `api-breaking`. For `api-breaking`, include `BREAKING` in the PR title.
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

To see a list of available CI bot commands, please comment `/bot help`.

PR #15246, opened 2026-06-11
[https://nvbugs/6223556][fix] Propagate gen-first ctx usage via aux buffer to postproc

reasonsolo:
## Summary

**Root cause:** In gen-first disaggregated scheduling the context and generation requests run concurrently, so the context worker's usage accounting cannot be injected into the request as it is in the context-first path. It is instead delivered to the generation worker through the KV-transfer aux buffer (`RxSession.unpack_aux`), which sets `py_disaggregated_params.ctx_usage`. However, that value was never surfaced onto the response, so the postprocessor fell back to the generation worker's local accounting — which treats the entire transferred prompt as cached and over-reports `cached_tokens` (the full prompt length 17 instead of the reused `prompt_tokens-1 = 16`). This caused `test_disaggregated_overlap_gen_first` to fail its cache-reuse usage check.

**Fix:** Complete the existing aux-buffer transmission pipeline (introduced in #14177) instead of rewriting usage at the orchestrator:
- Carry `ctx_usage` on `LlmResult`.
- Set it on the response alongside `cached_tokens` in `py_executor` (all three response-creation sites: generation, fast-transfer, and streaming emit).
- Surface it onto the output's `disaggregated_params` in `result.py` so the existing postprocessor path (`_ctx_usage_from_outputs` -> `rewrite_usage_info_from_ctx`) adopts the context-side accounting.

This keeps the disaggregated orchestrator a thin proxy (consistent with #14177) and fixes **both** streaming and non-streaming gen-first usage, since the postprocessor already handles both uniformly.

The `overlap_gen_first` test configs are updated to enable block/partial reuse so the cache-reuse usage check is meaningful, and the corresponding waives are removed.

## Test plan
- [x] `test_disaggregated_overlap_gen_first[ctx_pp1-TinyLlama-1.1B-Chat-v1.0]` passes (B200). Before fix: every request reported `cached_tokens == prompt_tokens` (17/17). After fix: first (cold) request reports the context-side count (e.g. `cached_tokens: 3`), second (reuse) request reports `prompt_tokens - 1` (16 for completions, 30 for chat).
- [x] `ctx_pp4` variant — verified by CI (locally blocked by GPU availability; logic is identical and independent of context-side pipeline parallelism).

## Links
- Bug: https://nvbugs/6223556

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed disaggregated scheduling test cases for TinyLlama-1.1B-Chat model that were previously waived.

* **Improvements**
  * Enhanced context-side usage tracking and accounting in disaggregated execution scheduling modes.
  * Enabled KV-cache block reuse optimization in test configurations for better memory efficiency.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/llm_request.py
+++ b/tensorrt_llm/_torch/pyexecutor/llm_request.py
@@ -578,6 +578,12 @@ class LlmResult:
         self._py_result = py_result
         self.is_final = is_final
         self.cached_tokens = 0
+        # Context-worker usage propagated to the generation worker via the
+        # KV-transfer aux buffer in gen-first disaggregated scheduling. Carried
+        # on the response so the postprocessor can adopt the context-side
+        # prompt/cached token accounting (the gen worker treats the whole
+        # transferred prompt as cached and would otherwise over-report).
+        self.ctx_usage = None
         # Time breakdown metrics for performance analysis
         # Contains: step_metrics (list), ctx_gpu_forward_time (float), ctx_gpu_sample_time (float)
         self.time_breakdown_metrics = time_breakdown_metrics
--- a/tensorrt_llm/_torch/pyexecutor/llm_request.py
+++ b/tensorrt_llm/_torch/pyexecutor/llm_request.py
@@ -578,11 +578,8 @@ class LlmResult:
         self._py_result = py_result
         self.is_final = is_final
         self.cached_tokens = 0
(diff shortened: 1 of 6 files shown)


Shixiaowei02 (approved):

pcastonguay (approved):

fredricz-20070104 (approved):
Approved

lancelly (approved):
LGTM

PR #14816, opened 2026-06-01
[None][fix] Add KV cache V2 recompute pause path

yizhang-nv:
@coderabbitai summary

## Description

This PR separates destructive recompute pause from ordinary KV cache V2 suspension in the Python scheduler/executor path. It adds a scheduler output field for recompute-paused requests with a default empty value so V1 scheduler construction remains unchanged, gates suspended-victim promotion on host-tier availability, and has PyExecutor terminate/replay recompute-paused requests separately from ordinary suspended requests.

## Test Coverage

- `python3 -m pytest -s tests/unittest/_torch/executor/test_kv_cache_v2_scheduler.py tests/unittest/_torch/executor/test_scheduler_serializable_output.py`
- `python3 -m pytest -s tests/unittest/pyexecutor/test_iter_stats_populate.py tests/unittest/_torch/executor/test_py_scheduler.py -k "paused or full_pipeline_output_structure"`
- `python3 -m pytest -s tests/unittest/_torch/executor/test_py_scheduler.py -k full_pipeline_output_structure`
- `git diff --check`
- pre-commit hooks on commit

## PR Checklist

Please review the following before submitting your PR:
- PR description clearly explains what and why. If using CodeRabbit's summary, please make sure it makes sense.
- PR Follows [TRT-LLM CODING GUIDELINES](https://github.com/NVIDIA/TensorRT-LLM/blob/main/CODING_GUIDELINES.md) to the best of your knowledge.
- Test cases are provided for new code paths (see [test instructions](https://github.com/NVIDIA/TensorRT-LLM/tree/main/tests#1-how-does-the-ci-work))
- Any new dependencies have been scanned for license and vulnerabilities
- [CODEOWNERS](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/CODEOWNERS) updated if ownership changes
- Documentation updated as needed
- Update [tava architecture diagram](https://github.com/NVIDIA/TensorRT-LLM/blob/main/.github/tava_architecture_diagram.md) if there is a significant design change in PR.
- The reviewers assigned automatically/manually are appropriate for the PR.

- [x] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

To see a list of available CI bot commands, please comment `/bot help`.

PR #15252, opened 2026-06-11
[None][fix] Add KV cache V2 recompute pause path

yizhang-nv:
## Dev Engineer Review

- Adds KV cache V2 support for recompute-paused requests.
- Tracks recompute-paused requests in scheduling, executor loops, pipeline-parallel handling, serialization, and iteration statistics.
- Adds host-tier detection and GPU-only fallback when host-tier initialization fails.
- Refactors `LlmRequest` state initialization for recompute resets.
- Preserves legacy `SchedulerOutput` construction with an empty default list.
- Review should verify deferred termination, resource cleanup, request-state reset, and host-tier fallback.
- No configuration or test-list files changed.

## QA Engineer Review

- Added or updated:
  - `test_recompute_pause_does_not_apply_executor_max_input_len`
  - `test_recompute_pause_defers_reset_until_pp_consensus`
  - `test_terminal_request_does_not_recompute_after_pp_consensus`
  - `test_paused_decode_requests`
  - `test_serializable_scheduler_output_round_trip`
  - KV cache V2 scheduler allocation and recompute-pause tests
  - `test_full_pipeline_output_structure`
- No files under `tests/integration/test_lists/` were changed.
- Coverage in `test-db/` and `qa/` is unavailable.
- Verdict: needs follow-up.

## Description

Cherry-pick/adapt of #14816 onto `main`.

This adds KV cache V2 handling for recompute-paused requests in the PyTorch executor path:
- account recompute-paused requests in scheduled batch stats and iteration stats
- terminate and pause recompute-paused requests separately from the non-V2 pause path
- preserve current `main` disk cache tier logic while adding host cache tier detection and GPU-only fallback when host tier setup fails

Opened as draft so the main-branch adaptation can be reviewed before marking ready.

## Test Coverage

- [x] `python3 -m pytest -s tests/unittest/_torch/executor/test_kv_cache_v2_scheduler.py tests/unittest/_torch/executor/test_scheduler_serializable_output.py` (`168 passed`)
- [x] `python3 -m pytest -s tests/unittest/pyexecutor/test_iter_stats_populate.py tests/unittest/_torch/executor/test_py_scheduler.py -k "paused or full_pipeline_output_structure"` (`3 passed, 150 deselected`)
- [x] `python3 -m pytest -s tests/unittest/pyexecutor/test_iter_stats_populate.py tests/unittest/pyexecutor/test_recompute_pause.py` (`25 passed`)

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor.py
@@ -8234,7 +8234,12 @@ class PyExecutor:
         if not requests:
             return
         for req in requests:
-            if req.py_request_id not in self._pending_recompute_pause_ids:
+            should_recompute = (
+                req.py_request_id not in self._pending_recompute_pause_ids
+                and any(active_request is req
+                        for active_request in self.active_requests)
+                and not req.is_finished)
+            if should_recompute:
                 self._pause_recompute_request(req)
 
     def _add_inflight_ids(self, scheduled_requests: ScheduledRequests):
--- a/tensorrt_llm/_torch/pyexecutor/scheduler/scheduler_v2.py
+++ b/tensorrt_llm/_torch/pyexecutor/scheduler/scheduler_v2.py
@@ -1071,6 +1071,10 @@ class KVCacheV2Scheduler(RequestScheduler):
     ) -> bool:
         if req.request_id in inflight_request_ids:
             return False
+        # is_generation_in_progress_state also includes GENERATION_TO_COMPLETE,
+        # which is outside the schedulable range and may still be finalizing.
+        if req.state_value == self._gen_to_complete_state_value:
+            return False
         # Completed multimodal prefill deliberately releases the inputs and
         # embedding needed for replay, leaving an empty or MRoPE-only dict as
         # the durable marker. Partial-context requests still retain replay data.
--- a/tests/unittest/_torch/executor/test_kv_cache_v2_scheduler.py
+++ b/tests/unittest/_torch/executor/test_kv_cache_v2_scheduler.py
@@ -587,6 +587,24 @@ class TestKVCacheFailuresGen:
         assert out.recompute_paused_requests == []
         mgr.free_resources.assert_not_called()
 
+    def test_recompute_pause_skips_generation_to_complete_victim(self):
+        """A finalizing request cannot be replayed by destructive pause."""
+        mgr = make_kv_cache_manager(try_allocate_generation_fn=lambda req: False)
+        sched = make_scheduler(mgr, max_num_tokens=100)
+        current = make_gen_request(0)
+        current.is_generation_to_complete_state = False
+        victim = make_gen_request(1)
+        victim.state_value = GEN_TO_COMPLETE
+        victim.is_generation_to_complete_state = True
+        mgr.kv_cache_map[victim.py_request_id].is_active = False
+
+        out = sched.schedule_request([current, victim], set())
+
+        assert out.generation_requests == []
+        assert ids(out.paused_requests) == [0]
+        assert out.recompute_paused_requests == []
+        mgr.free_resources.assert_not_called()
+
     def test_recompute_frontier_survives_ordinary_frontier_shrink(self):
         """Recompute still sees a suspended tail after ordinary eviction skips it."""
         attempts = {}
--- a/tests/unittest/_torch/executor/test_recompute_pause.py
+++ b/tests/unittest/_torch/executor/test_recompute_pause.py
@@ -45,6 +45,21 @@ def test_recompute_pause_does_not_apply_executor_max_input_len() -> None:
     request.reset_for_recompute.assert_called_once_with(_UNBOUNDED_PAUSE_MAX_INPUT_LEN)
 
 
+def test_recompute_pause_skips_request_finished_during_overlap() -> None:
+    executor = _make_executor(None)
+    request = _StubRequest()
+    executor.active_requests = [request]
+    scheduled_batch = ScheduledRequests()
+    scheduled_batch.recompute_paused_requests = [request]
+
+    executor._terminate_recompute_paused_requests(scheduled_batch)
+    request.is_finished = True
+    executor._pause_recompute_paused_requests(scheduled_batch)
+
+    executor.resource_manager.free_resources.assert_called_once_with(request)
+    request.reset_for_recompute.assert_not_called()
+
+
 def test_recompute_pause_defers_reset_until_pp_consensus() -> None:
     handler = Mock()
     executor = _make_executor(handler)


tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -1337,7 +1342,9 @@ def _collect_scheduled_batch_stats(
                 pass
 
         num_paused_requests = 0
-        for req in scheduled_batch.paused_requests:
+        paused_requests = (scheduled_batch.paused_requests +
```
lancelly, line 2026: should this be a property of ScheduledResults?

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -5180,6 +5201,44 @@ def _pause_requests(self, requests_to_pause):
         for req in requests_to_pause:
             req.pause(self.max_input_len)
 
+    def _pause_recompute_request(self, req):
```
lancelly: This duplicates request-init state and will silently drift the next time someone adds a py_* field to LlmRequest.  Perhaps we can consider centralizing this in a single LlmRequest.reset_for_recompute() (or reusing the existing init path) so the canonical field list lives in one place?

tensorrt_llm/_torch/pyexecutor/kv_cache_manager_v2.py
```diff
@@ -638,7 +646,30 @@ def append_to_kv_heads_per_layer(
 
         self.kv_cache_manager_py_config = config
 
-        self.impl = KVCacheManagerPy(config)
+        try:
```
lancelly, line 1068: Why don't we need this try catch section before this PR?

yizhang-nv: This is ported from deepseek v4 feature branch. We had encountered such issue before but only fixed it on deepkseek v4

tensorrt_llm/_torch/pyexecutor/py_executor.py
yizhang-nv: We need to handle it separately. For paused req, we don't need to update the input prompt while we need to do it for recompute path.

nvpohanh:
[by Codex] @lowsfer Could you review this PR? Thanks!

tensorrt_llm/_torch/pyexecutor/llm_request.py
```diff
@@ -842,6 +825,45 @@ def cached_tokens(self, value: int):
...
+        else:
+            self.py_draft_tokens = ([] if self.draft_tokens is None else
+                                    self.draft_tokens)
+        self.py_last_context_chunk = (None, None)
+        self.py_last_draft_tokens = None
+        self.py_num_accepted_draft_tokens = 0
+        self.py_num_accepted_draft_tokens_indices = []
+        self.py_rewind_draft_token_separate_adjustment = 0
+        self.py_decoding_iter = 0
+        self.py_ctx_pre_resize_cap = None
```
eopXD, line 1168: Is this `py_ctx_pre_resize_cap` ever used?

nvpohanh:
[by Codex] @lowsfer Friendly reminder to review this PR as the primary KV-cache-manager reviewer when you have a chance. Thanks!

nvpohanh:
[by Codex] @lowsfer Could you please review this PR? Thank you!

nvpohanh:
[by Codex] @lowsfer Friendly review reminder: this PR is awaiting your review. Thanks!

tensorrt_llm/_torch/pyexecutor/py_executor.py
```diff
@@ -2583,6 +2592,10 @@ def _executor_loop_pp(self):
                             local_disagg_candidates,
                             fitting_disagg_gen_init_requests)
 
+                if self._is_kv_manager_v2:
+                    self._terminate_recompute_paused_requests(scheduled_batch)
+                    self._pause_recompute_paused_requests(scheduled_batch)
```
Shixiaowei02, line 2762: Nice cleanup of the V2 pause path. One concern about this call landing in the PP loop specifically: this is the only loop where `_disagg_pp_termination_handler` is live (constructed at :961 for `pp_size > 1` + disagg), and there `_terminate_request` (:6711) only stashes the request — the actual free waits for a ring consensus. The pause on the next line runs immediately, and `pause()` ends in `mSeqSlot.reset()`, so the victim becomes re-schedulable while its old slot mapping is still present and `add_slot` (`resource_manager.py:2333`) asserts; if the consensus lands first, it frees a request that has meanwhile re-acquired a slot and KV. Could the recompute victims bypass the deferred handler here, or the pause wait until termination has actually landed?

yizhang-nv: Added a buffer list to defer the termination of the requests

nvpohanh:
[by Codex] @lowsfer Could you please review PR #15252 for the KV-cache manager changes? Thanks!

nvpohanh:
[by Codex] @lowsfer Could you review this PR? Thanks!

yizhang-nv:
> The pause candidate filter keys on forward-inflight requests, not on KV-transfer state, and `_recompute_pause_request` calls `free_resources`. So a generation request whose KV just arrived over the wire can be torn down and re-prefilled locally.

Added a gate so that we will not enable this feature under disagg serving.

tensorrt_llm/_torch/pyexecutor/kv_cache_manager_v2.py
```diff
@@ -1045,6 +1045,10 @@ def append_to_kv_heads_per_layer(
                 f"KV cache manager v2 disk cache quota set to {disk_cache_size / (1 << 30):.2f}GiB at {disk_cache_path}"
             )
 
+        self.has_host_cache_tier = any(
```
lowsfer: Is there a reason why it has to be host tier? we may have a secondary gpu tier (for compressed data) for offloading. Maybe we should use has_secondary_cache_tier and just set it to len(cache_tiers) > 1?

yizhang-nv: Currently the scheduler only supports evicting the request to host. Maybe we can change it until the scheduler can properly handle other eviction policy.

eopXD (changes requested):
Recompute-pause victim selection can tear down a request that has already finished.

(1) GENERATION_TO_COMPLETE requests are eligible victims. The victim scan at scheduler_v2.py:1155 admits any request passing _is_recompute_pause_candidate, which bottoms out in _is_started_request → is_generation_in_progress_state — and in C++ that predicate includes kGENERATION_TO_COMPLETE (llmRequest.h:1639-1642). Note this same file already uses the correct bound at :465 (and not r.is_generation_to_complete_state); the victim filter just omits it. Could we mirror :465 in _is_recompute_pause_candidate?

(2) Under the overlap scheduler a GENERATION_IN_PROGRESS victim may already be finished. py_executor.py:4816 schedules while :4972 applies the previous batch's GPU results, so at selection time the last sampled token is computed but unapplied. If it is EOS or a stop word, _handle_stop_criteria → finish_by (llm_request.py:1283) sets GENERATION_COMPLETE at :4972 — note this path skips GENERATION_TO_COMPLETE entirely, since will_complete_next_iteration() only predicts LENGTH-based completion. The reset then runs unconditionally at :5006 and pause() rewinds mState to kCONTEXT_INIT.

The consequence is user-visible: createResult returns nullopt for a non-finished request, so the completed response is never emitted; request_done is never set, and the request is re-added to active_requests at :8129 to re-prefill. On replay the EOS is inside the prompt and the stop-word window has moved past it, so it can run on to max_new_tokens.

This one can't be filtered at selection — the information doesn't exist yet at :4816. It needs a re-check at _pause_recompute_paused_requests, which is exactly what _on_disagg_pp_termination already does at :7798-7800 (and not request.is_finished). Could we apply the same guard on this path?

yizhang-nv:
@eopXD Addressed in `54cf104409`:

1. Excluded `GENERATION_TO_COMPLETE` requests from recompute-pause victim selection.
2. Added a final active/finished re-check before `reset_for_recompute()` so an overlap request completed by the previous sample result is not rewound.

Added focused regression coverage for both cases. All pre-commit hooks pass. I am triggering CI on the new head now. Thanks for catching these races.

eopXD (approved):
The pull request looks good to me. Thank you for addressing the comments.

Shixiaowei02 (approved):

tensorrt_llm/_torch/pyexecutor/kv_cache_manager_v2.py
```diff
@@ -1060,31 +1059,36 @@ def append_to_kv_heads_per_layer(
...
+            isinstance(tier, HostCacheTierConfig) for tier in config.cache_tiers
+        )
 
         self.kv_cache_manager_py_config = config
 
         try:
             self.impl = KVCacheManagerPy(config, event_manager=self.event_manager)
         except (CuError, KVCacheOutOfMemoryError):
-            if len(cache_tiers) > 1:
+            if has_host_cache_tier:
```
QiJune, line 1071: Could this fallback be synchronized across the model-parallel ranks that share scheduling/KV state? `cuMemHostRegister` or `KVCacheOutOfMemoryError` may occur on only a subset of ranks. In that case, some ranks would retain the host tier while others rebuild without it, causing `cache_tiers` and `can_evict` to diverge across ranks.

If any rank fails host-tier initialization, should all relevant ranks tear down the initially created manager and rebuild without the host tier? Alternatively, is there an invariant in `KVCacheManagerPy` that guarantees initialization succeeds or fails symmetrically across ranks?

QiJune (approved):

tensorrt_llm/_torch/pyexecutor/kv_cache_manager_v2.py
yizhang-nv: Will add the synchronization in a separate pr since this feature is unrelated to this pr.

PR #14812, opened 2026-06-01
[#10710][feat] Make explicit CLI flags take precedence over --config / --extra_llm_api_options YAML

marinayanov:
## Summary by CodeRabbit

* **Documentation**
  * Updated CLI/YAML override behavior documentation: explicit CLI flags now take precedence over YAML configuration values. Unset CLI flags fall back to YAML settings, then to defaults for `trtllm-serve`, `trtllm-eval`, and `trtllm-bench` commands.
  * Updated TensorRT-LLM Release 1.2 release notes to reflect the revised precedence rules.

## Description

Title
[#10710][feat] Make explicit CLI flags take precedence over --config / --extra_llm_api_options YAML
Body

@coderabbitai summary
## Description
Today, `--config` / `--extra_llm_api_options` YAML values silently override
explicit CLI flags in `trtllm-serve`, `trtllm-eval`, and `trtllm-bench`.
For example, `trtllm-serve <model> --tensor_parallel_size 4 --config foo.yaml`
with `tensor_parallel_size: 8` in `foo.yaml` runs at TP=8. This is
unconventional and inconsistent with AutoDeploy's documented
`CLI > YAML > defaults`. See #10710.
This PR flips the precedence to **explicit CLI > YAML > defaults**, using
`click.Context.get_parameter_source()` to detect which flags the user
typed. Un-set CLI flags still fall back to YAML.
Backward-compatible: programmatic callers of
`update_llm_args_with_extra_dict` (Triton backend, LLM API, third-party)
that don't pass the new `explicit_cli_keys` argument keep today's
behavior.

Why this is the right direction, briefly:
- **In-repo precedent.** AutoDeploy's config system already documents
  exactly this order — see `docs/source/features/auto_deploy/advanced/expert_configurations.md`:
  *"CLI Arguments (highest priority) > YAML Configs > Default Settings (lowest priority)."*
  This PR brings the rest of the CLI surface in line.
- **Conventional behavior.** kubectl, Helm, Hydra, Pydantic Settings,
  and ~every argparse-based Python CLI implement `CLI > config-file > defaults`.
  The current "YAML > CLI" is the surprise.
- **PR #11035 acknowledged the gap.** That PR added a value-based mitigation in `trtllm-serve`'s `is_non_default_or_required`, and the
  reviewers explicitly flagged the remaining limitation:
  *"CLI params can make this a bit tricky, since you need to
  distinguish between a CLI param being explicitly set vs. defaulting
  to a value."* This PR closes that gap with
[...]

Issue #10710, opened 2026-01-15
[Bug]: [AutoDeploy] The CLI-specified max_batch_size only takes effect if max_num_tokens is also set in the CLI


MrGeva:
### System Info

The CLI-specified max_batch_size only takes effect if max_num_tokens is also set in the CLI

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

`python3 benchmarks/cpp/prepare_dataset.py --stdout --tokenizer nvidia/Llama-3.1-8B-Instruct-FP8 token-norm-dist --input-mean 1000 --output-mean 2000 --input-stdev 0 --output-stdev 0 --num-requests 64 > llama_8b_1k_2k_64.inp`

`trtllm-bench --model nvidia/Llama-3.1-8B-Instruct-FP8 throughput --dataset /home/scratch.egeva_coreai/TensorRT-LLM/llama_8b_1k_2k.inp --backend _autodeploy --extra_llm_api_options /home/scratch.egeva_coreai/TensorRT-LLM/examples/auto_deploy/llama_8b.yaml`

use this yaml:
```
attn_backend: flashinfer
compile_backend: torch-cudagraph
enable_chunked_prefill: true
free_mem_ratio: 0.88
max_batch_size: 1
max_seq_len: 65536
model_factory: AutoModelForCausalLM
runtime: trtllm
skip_loading_weights: false
```

### Expected behavior

the CLI max_batch_size should win the one defined in the yaml

### actual behavior

na

### additional notes

na

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and checked the [documentation](https://nvidia.github.io/TensorRT-LLM/) and [examples](https://github.com/NVIDIA/TensorRT-LLM/tree/main/examples) for answers to frequently asked questions.

marinayanov:
Issue was Yaml args taking precedence over explicitly passed CLI args from the user.
merged PR: https://github.com/NVIDIA/TensorRT-LLM/pull/14812
follow-up open PR to fix logging: https://github.com/NVIDIA/TensorRT-LLM/pull/15254

PR #15254, opened 2026-06-11
[#10710][fix] clarify and align trtllm-bench runtime logging

marinayanov:
## Description

Follow-up to #14812. That PR made explicit CLI flags win over YAML config; this
PR fixes the *logging and reporting* of `max_batch_size` / `max_num_tokens` in
`trtllm-bench` so what's printed matches what the engine actually runs with.

## Problem

After #14812, the values reported by `trtllm-bench` could diverge from the
values the engine actually used:

1. **Single CLI flag silently ignored.** `get_settings()` only honored
   `--max_batch_size` / `--max_num_tokens` when *both* were passed. Providing
   just one fell back to the heuristic for *both*, discarding the user's value
   with no clear log.

2. **Report showed pre-merge values.** `settings_config` was populated before
   the `extra_llm_api_options` YAML merge in `get_llm_args()`, but the final
   report reads from `settings_config`. So when YAML supplied a value the user
   didn't set on the CLI, the report (e.g. `Max Runtime Tokens`) showed the
   stale pre-merge number while the engine ran the merged one.

3. **Ambiguous logging.** The settings log didn't make clear which values were
   user-provided vs. heuristic, and why the final report might differ.

## Changes

- `utils/general.py`: honor a *single* CLI flag — use the user value for the
  one provided and the heuristic for the other. Unified all four cases into a
  consistent `Initial settings: max_batch_size=... (user-provided|heuristic),
  max_num_tokens=... (user-provided|heuristic).` line so each value's source is
  explicit and the "initial/pre-merge" nature is clear.
- `throughput.py` / `low_latency.py`: sync `settings_config.max_batch_size`
  and `max_num_tokens` to the post-merge `kwargs` (after `get_llm_args()`,
  before `get_llm()`), so the final report reflects the values actually passed
  to the engine.

## PR Checklist

- [ v] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced runtime configuration synchronization to ensure batch size and token limit overrides are properly applied across benchmarks.
  * Improved granular logging for batch size and token limit configuration scenarios during benchmark execution.


Superjomn (approved):
LGTM

tensorrt_llm/bench/benchmark/low_latency.py
```diff
@@ -315,6 +315,9 @@ def latency_command(
     if bench_env.telemetry_config is not None:
         kwargs["telemetry_config"] = bench_env.telemetry_config
 
+    runtime_config.settings_config.max_batch_size = kwargs["max_batch_size"]
+    runtime_config.settings_config.max_num_tokens = kwargs["max_num_tokens"]
```
galagam: Is it guaranteed that these keys exist in kwargs? Using `kwargs.get(key, default)` is a safe way to avoid a KeyError if they are missing.

tensorrt_llm/bench/benchmark/throughput.py
```diff
@@ -452,6 +452,9 @@ def throughput_command(
         if bench_env.telemetry_config is not None:
             kwargs["telemetry_config"] = bench_env.telemetry_config
 
+        runtime_config.settings_config.max_batch_size = kwargs["max_batch_size"]
+        runtime_config.settings_config.max_num_tokens = kwargs["max_num_tokens"]
```
galagam: See comment above - can you guarantee that these keys exist in kwargs?

galagam (approved):
Overall looks good, see comments

tensorrt_llm/bench/benchmark/low_latency.py
marinayanov: It is guaranteed, these come from runtime_config.get_llm_args() (merged into kwargs at line 313), which always emits them from the required settings_config fields. That said, I'll switch to kwargs.get(...) as you suggested so we're not coupled to that behavior if get_llm_args() ever changes

PR #15274, opened 2026-06-11
[None][infra] PLC nightly pipeline update

yuanjingx87:
## Summary by CodeRabbit

* **New Features**
  * Added configurable pipeline parameters to selectively enable or disable source code scanning, container scanning, and SonarQube analysis stages.

* **Improvements**
  * Enhanced token retrieval with retry logic for increased reliability.
  * Updated license scanning to consider component types for more accurate preapproval matching.
  * Added automatic archiving of scan reports.

## Description

1. add some retry
2. using package name + package type to check if the dependency is pre-approved
3. Can choose to run the scan type when launching the pipeline

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/jenkins/TensorRT_LLM_PLC.groovy
+++ b/jenkins/TensorRT_LLM_PLC.groovy
@@ -366,7 +366,12 @@ def processScanResults(ref) {
                     if (result.dashboard_url) {
                         echo "Dashboard: ${result.dashboard_url}"
                     }
-                    currentBuild.result = 'UNSTABLE'
+                    // In pre_merge mode, we intentionally skip marking the build as UNSTABLE.
+                    // PLC scanning accuracy needs further improvement before we can reliably
+                    // gate pre-merge builds on scan results without causing false failures.
+                    if (params.scanMode != "pre_merge") {
+                        currentBuild.result = 'UNSTABLE'
+                    }
                 } else {
                     echo "No new risks detected."
                 }
@@ -385,7 +390,7 @@ pipeline {
         string(name: 'forkOwner', defaultValue: '', description: 'Name of the fork owner, need to select \"github_fork\" for repoUrlKey, otherwise it will be ignored')
         string(name: 'postMergePipelineName', defaultValue: '', description: 'Optional: post-merge pipeline job name to associate with this scan')
         string(name: 'postMergeBuildNumber', defaultValue: '', description: 'Optional: post-merge pipeline build number to associate with this scan')
-        choice(name: 'scanMode', choices: ['monitor','release'], description: "When set to monitor, only report newly introduced dependencies. When set to release, will report all detected risks")
+        choice(name: 'scanMode', choices: ['monitor','release','pre_merge'], description: "When set to monitor, only report newly introduced dependencies. When set to release, will report all detected risks")
         booleanParam(name: 'runSourceCodeScanning', defaultValue: true, description: 'Run Source Code OSS Scanning (lock file generation + Pulse OSS scan)')
         booleanParam(name: 'runContainerScanning', defaultValue: true, description: 'Run Container Scanning (Pulse container scan)')
         booleanParam(name: 'runSonarQube', defaultValue: true, description: 'Run SonarQube Code Analysis')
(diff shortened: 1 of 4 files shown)


jenkins/scripts/pulse_in_pipeline_scanning/submit_report.py
```diff
@@ -156,9 +154,11 @@ def submit_source_code_licenses(
                 "s_component_type": component.get("type"),
                 "b_is_new": is_new,
             }
-            if (is_new or not only_report_new_risk) and (package_name not in map_preapproved):
+            if (is_new or not only_report_new_risk) and not is_preapproved(
+                map_preapproved, package_name, (component.get("type") or "").lower()
```
tburt-nv: The difference between `""` and `None` might be confusing, could we use `unknown` here to explicitly distinguish this from the wildcard case?

Is the process to update "pre-approve" entries still manual, or is there somewhere else we would need to convert None->"unknown"?

```suggestion
                map_preapproved, package_name, (component.get("type") or "unknown").lower()
```

jenkins/scripts/pulse_in_pipeline_scanning/submit_report.py
```diff
@@ -290,9 +290,11 @@ def submit_container_licenses(
             "s_license_ids": ",".join(license_ids),
             "b_is_new": is_new,
         }
-        if (is_new or not only_report_new_risk) and (package_name not in map_preapproved):
+        if (is_new or not only_report_new_risk) and not is_preapproved(
+            map_preapproved, package_name, (v.get("type") or "").lower()
```
tburt-nv:
```suggestion
            map_preapproved, package_name, (v.get("type") or "unknown").lower()
```

jenkins/TensorRT_LLM_PLC.groovy
```diff
@@ -327,7 +366,12 @@ def processScanResults(ref) {
                     if (result.dashboard_url) {
                         echo "Dashboard: ${result.dashboard_url}"
                     }
-                    currentBuild.result = 'UNSTABLE'
+                    // In pre_merge mode, we intentionally skip marking the build as UNSTABLE.
+                    // PLC scanning accuracy needs further improvement before we can reliably
+                    // gate pre-merge builds on scan results without causing false failures.
+                    if (params.scanMode != "pre_merge") {
+                        currentBuild.result = 'UNSTABLE'
+                    }
```
tburt-nv: Instead of marking this job as passing, could we change https://github.com/NVIDIA/TensorRT-LLM/pull/15552 to wrap the trigger in a `catchError` block or try/catch so that the failures are more obvious? I think we can even mark the `[OSS-Compliance-Check] Run` stage as unstable without marking the overall pipeline as unstable with:

```
catchError(message: message, stageResult: 'UNSTABLE') {
    // job trigger
}
```

tburt-nv (approved):

jenkins/scripts/pulse_in_pipeline_scanning/submit_report.py
yuanjingx87: Yeah, that is still manual, since pre-approved dependencies should be added when newly filed OSRB bugs are approved by the OSRB team, in order to automate that we might need to talk to NVBUG. Potentially something can be done by the PLC agent in the future I guess..

To answer your question. Actually ` or ""` part should never been accessed, since it is trying to get type from SBOM / container license result file, so the type should always there. I have already added type for the previously OSRB approved dependencies in opensearch. and Moving forward all the dependencies should come with a type.

yuanjingx87: But change it to unknown sounds good to me, only thing we need to keep in mind is when we add pre-approved dependencies to open-search, we need to make sure if type is not there, we need to set as unknown.

jenkins/TensorRT_LLM_PLC.groovy
yuanjingx87: I just updated both PR to address your feedback, will try out once the CI is fixed.

yuanjingx87:
/bot skip --comment "no need to run CI"

PR #15287, opened 2026-06-12
[#15289][fix] AutoDeploy: Remove host sync on mixed batches with speculative decoding

### LINKED ISSUE #15289: [Feature]: AutoDeploy: Remove host sync for MTP + overlap scheduler + mixed prefill/decode batches

There was a host sync in the above setting, caused by the fact the overlap scheduler was trying to make adjustments to the input_pos of prefill sequences of unknown length. This can be fixed by simply noticing that prefill sequences do not have prior-iteration MTP acceptance to factor in, so they don't need any adjustment in the overlap scheduler and can be skipped.

### Alternatives

_No response_

### Additional context

_No response_

govind-ramnarayan:
fixes: #15289 

Remove a host sync that breaks cuda graph capture for overlap scheduler + mixed prefill/{extend/decode} sequences + MTP

Enable chunked prefill for SuperV3 + MTP in the config.

## Release Notes

* **New Features**
  * Enabled chunked prefill support for speculative decoding configurations.

* **Improvements**
  * Enhanced speculative decoding offset computation for mixed prefill/extend/decode scenarios to improve efficiency.

* **Tests**
  * Added unit test validating speculative decoding offset behavior with overlapping requests.
  * Updated speculative decoding smoke test with optimized configuration.

## PR Checklist

## GitHub Bot Help


tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1723,9 +1731,35 @@ def offset_pos_and_cache_(
...
+            extend_seq_end = extend_seq_start + num_extend
+            decode_seq_start = extend_seq_end
+            decode_seq_end = decode_seq_start + num_decode
+
+            offset_for_pos_ids[extend_token_start:extend_token_end] = offset[
+                extend_seq_start:extend_seq_end
+            ].repeat_interleave(tokens_per_extend)
+            offset_for_pos_ids[decode_token_start:decode_token_end] = offset[
+                decode_seq_start:decode_seq_end
+            ]
```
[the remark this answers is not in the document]
hnover-nv: Any reason not to do this?  or at least some sort of check, if it's cheap?  My understanding is that this branch is currently for overlap scheduler w/out cuda graph capture, so we should be able to inspect things?

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1723,9 +1731,35 @@ def offset_pos_and_cache_(
             _, num_extend_tokens, _ = self.batch_info.get_num_tokens()
             tokens_per_seq = num_extend_tokens // num_extend
             offset_for_pos_ids = offset[:num_extend].repeat_interleave(tokens_per_seq)
+        elif no_prefill_offsets:
```
hnover-nv: Does this generalize the decode and draft situations above?  In is_extend you get the token length from num_extend_tokens // num_extend.  Here you get it from max draft length + 1. Can they differ, given we don't support the varying draft length?
I'm wondering if we could just drop the earlier two cases and just do this or the remaining fallback dpeending on whether we're pure encode, pure decode, or whatever the prefill case is whre this applies but we don't have a pure situation. (I need to understand that still)

Basically, this seems like it generalizes the case where the layout is "static", prefill that doesn't matter here, uniform draft length for all drafts, and then decodes of course are uniform.  So I'd think everything could fall back to this?

tests/unittest/auto_deploy/singlegpu/smoke/test_ad_speculative_decoding.py
```diff
@@ -66,6 +66,17 @@ def test_super_mtp_smoke():
         mtp_eagle_one_model=True,
         speculative_model=model_path,
     )
+    experiment_config["args"]["model_kwargs"].update(
```
hnover-nv: is overlap scheduler on?  that's the motivation, right?  in case there's some corruption or something using host data, we'd want to see that case?  Or is this just a compile and run w/out error test?  what can we see?  I see the params below so I'm guessing there are no weights here?  do we do any math checks at all?

tests/unittest/auto_deploy/singlegpu/smoke/test_ad_speculative_decoding.py
```diff
@@ -66,6 +66,17 @@ def test_super_mtp_smoke():
...
+        {
+            "hidden_size": 256,
+            "intermediate_size": 256,
+            "mamba_head_dim": 64,
+            "ssm_state_size": 64,
```
hnover-nv: any particlar reason our smoke is nemotron-like, out of curiosity?

hnover-nv (commented):
i slacked you some questions I haven't had a chacne to read the replies, yet, but I'll send this now in case it's helpful.  feel free to wait until I look at your slack stuff, maybe you already covered it :)

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
govind-ramnarayan: Yes this is true - it could just all be this case and we don't need the special cases for is-extend-only and is-generation-only; in the case where we ignore the prefill. So it could be like, 

if no_prefill_offsets or is_extend_only or is_generation_only:
- Do this code
else:
Generic code to handle prefills as well

good catch!

tests/unittest/auto_deploy/singlegpu/smoke/test_ad_speculative_decoding.py
govind-ramnarayan: No math checks - the smoke test is can we run a small version of the model that is architecturally similar through our pipeline without error. As you said there are no weights. This can be useful for catching some issues like host syncs during cuda graph capture, whether we hit some sort of assertion or error. I think it makes sense to switch this to trtllm attention + cudagraph enabled + overlap scheduler on, to make sure that works, but it's not a test of this feature - the only test of this feature is the unit test that was added to make sure we are not relying on `seq_len` in the mixed prefill/decode case.

tests/unittest/auto_deploy/singlegpu/smoke/test_ad_speculative_decoding.py
govind-ramnarayan: Was useful with catching errors in Nemotron + MTP during development. Sometimes (when smoke tests are not waived) it's caught some errors like piecewise cudagraph getting changed to default etc. I do think we should probably have a simple Llama + Eagle3 smoke test as well, or maybe as our "main" smoke test here, at this point it's probably equally useful.

tests/unittest/auto_deploy/singlegpu/shim/test_cached_sequence_interface.py
```diff
@@ -71,6 +72,29 @@ def allocate(self, sequence_info: SequenceInfo) -> torch.Tensor:
         return torch.empty(1, device=sequence_info.device)
 
 
+def _make_mixed_sequence_info(device: str | torch.device | None = None) -> SequenceInfo:
```
govind-ramnarayan: This type declaration is a bit excessive; skip it

tests/unittest/auto_deploy/singlegpu/shim/test_cached_sequence_interface.py
```diff
@@ -1104,6 +1128,130 @@ def test_sequence_info_set_capture_batch_max_draft_len_gt_0_builds_extend_only_b
...
+    real_repeat_interleave = torch.repeat_interleave
+    repeat_interleave_output_sizes = []
+
+    def checked_repeat_interleave(input_tensor, repeats, *args, **kwargs):
+        repeat_interleave_output_sizes.append(kwargs.get("output_size"))
+        return real_repeat_interleave(input_tensor, repeats, *args, **kwargs)
+
+    # Check the repeat_interleave contract directly; CUDA sync-debug mode is a useful
+    # manual probe, but it is prototype-only and too environment-sensitive for CI.
+    monkeypatch.setattr(torch, "repeat_interleave", checked_repeat_interleave)
```
govind-ramnarayan: This seems pointless; better to just simplify this to a correctness for our offset_pos_and_cache_(), and explicitly document that it is purely correctness, not checking no host sync? No need for monkey-patch

tests/unittest/auto_deploy/singlegpu/shim/test_cached_sequence_interface.py
```diff
@@ -1104,6 +1128,130 @@ def test_sequence_info_set_capture_batch_max_draft_len_gt_0_builds_extend_only_b
...
+    expected = position_ids_before.clone()
+    expected[:3] += 1
+    expected[3:5] += 2
+    expected[5:9] += 3
+    expected[9:] += 4
+
+    assert torch.equal(position_ids_after, expected)
+
+
+def test_sequence_info_overlap_mixed_mtp_offset_updates_non_prefill_requests():
```
govind-ramnarayan: It looks like this is a correctness test for offset_with_new_lens_(). Can you check if this is actually needed? We might already have some tests for this + the new test for offset_pos_and_cache_(). If it is not covered, it's a good test, but make it clear that it's a correctness test for this function and add comments to make it clear what it is doing.

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1645,7 +1645,9 @@ def run_host_prepare_for_attention_forward(self) -> None:
 
     @nvtx_range("ad_offset_pos_and_cache_")
     def offset_pos_and_cache_(
-        self, offset: torch.Tensor, active_args_override: Optional[Set[str]] = None
```
govind-ramnarayan, line 1648: I think you just added a comma and then reformatted? Revert if possible

tensorrt_llm/_torch/auto_deploy/custom_ops/attention_interface.py
```diff
@@ -1777,6 +1770,9 @@ def offset_with_new_lens_(self, new_lens_ungathered: torch.Tensor) -> None:
         increment[num_prefill : num_prefill + num_overlap] = (
             new_lens_ungathered[gather_slot_idx] - 1
         )
+        # This adjustment accounts for requests that went through speculative verification in the
```
govind-ramnarayan: Stale comment?

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
```diff
@@ -120,6 +121,142 @@ def _has_active_host_arg(si: SequenceInfo):
...
+
+    assert torch.equal(position_ids_after, expected)
+    assert seq_info.get_arg("input_pos", truncate=True).tolist() == [11, 22, 33, 44]
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+@pytest.mark.filterwarnings("ignore:Synchronization debug mode is a prototype feature:UserWarning")
+def test_sequence_info_mixed_offset_has_no_cuda_sync():
+    """Mixed-batch position-id offset expansion must not synchronize the CUDA stream."""
+    assert_no_cuda_sync = pytest.importorskip("utils.util").assert_no_cuda_sync
```
govind-ramnarayan: I guess it's fine to put this on top if we think it gets package in the "standalone" tests. But it seems to me like it relies on a TRTLLM utility. I would rather have this imported if it gets packaged properly in standalone

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
```diff
@@ -120,6 +121,142 @@ def _has_active_host_arg(si: SequenceInfo):
...
+    expected = position_ids_before.clone()
+    expected[:3] += 1
+    expected[3:5] += 2
+    expected[5:9] += 3
+    expected[9:] += 4
+
+    assert torch.equal(position_ids_after, expected)
+
+
+def test_sequence_info_offset_with_new_lens_updates_overlap_requests():
```
govind-ramnarayan: How does this differ from the first test that is a "correctness" test? Testing offset_with_new_lens()? Just make sure that it is very clear - rename it to be simple and redo the comment. Basically I think this is a basic correctness for offset_with_new_lens(), in addition to the offset_pos_and_cache_() one we have above. Nothign more, nothing less.

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
govind-ramnarayan: Done. I moved the no-sync helper into the standalone-packaged `_torch_test_utils` module and import it normally from the test now, so the test no longer depends on the TRTLLM `utils.util` helper. Re-ran the standalone copied-package test and it passes (`1803 passed, 99 skipped, 45 xfailed, 8 xpassed` inside the package run).

tests/unittest/auto_deploy/singlegpu/custom_ops/test_switch_to_generate_inplace.py
govind-ramnarayan: Done. I renamed this to `test_offset_with_new_lens_correctness` and simplified the docstring to make it explicit that this is only basic correctness coverage for `SequenceInfo.offset_with_new_lens_`, complementing the direct `offset_pos_and_cache_` correctness/no-sync tests above.

hnover-nv (approved):

PR #15303, opened 2026-06-12
[None][fix] LTX-2: re-enable Ulysses for v2a cross-attention

luyiyun1021:
## Description

#14044 added Ulysses sequence-parallelism for the **v2a** (video→audio) cross-attention; #13978 (async-Ulysses refactor) silently disabled it — cross-attn was excluded from the Ulysses path at both the base-class `use_ulysses` gate and the LTX-2 dual-attn gate, and `use_ulysses=True` was dropped from the v2a construction. As a result v2a fell back to all-gathering the full video K/V every block, so its per-rank communication grows with `ulysses_size` instead of scaling.

This restores the v2a Ulysses path under **pure Ulysses (`cp_size == 1`)**, while preserving #13978's async-Ulysses (self-attn only) and keeping the combined ring/attn2d + Ulysses (`cp_size > 1`) all-gather fallback unchanged:
- `modules/attention.py`: head-shard the `SEPARATE_QKV` cross-attn inner backend when `cp_size == 1`.
- `transformer_ltx2.py`: build the dual-attn pair for cross-attn under pure Ulysses, and pass `use_ulysses=True` to `video_to_audio_attn`.

Output is numerically unchanged; this is a communication/perf fix (v2a uses the Ulysses 4-collective a2a instead of the full-video all-gather).

## Test Coverage

Existing multi-GPU parity tests (`multi_gpu/test_ltx2_ulysses.py`, `multi_gpu/test_ltx2_async_ulysses.py`) cover correctness/parity for the v2a + async paths.

## GitHub Bot Help

## Release Notes

* **Improvements**
  * Enhanced attention mechanism configuration in visual generation models for improved performance.
  * Optimized cross-attention setup for audio-visual task processing.

### DIFF
--- a/tensorrt_llm/_torch/visual_gen/modules/attention.py
+++ b/tensorrt_llm/_torch/visual_gen/modules/attention.py
@@ -4,6 +4,7 @@ from typing import Optional, Tuple
 import torch
 import torch.nn as nn
 
+from tensorrt_llm.logger import logger
 from tensorrt_llm.visual_gen.sparse_attention import SkipSoftmaxAttentionConfig
 
 from ...modules.linear import Linear, TensorParallelMode, WeightMode, WeightsLoadingConfig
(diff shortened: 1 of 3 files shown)


tensorrt_llm/_torch/visual_gen/modules/attention.py
```diff
@@ -180,10 +180,11 @@ def __init__(
         # The async-ulysses path uses SEPARATE_QKV for stream-pipelined
         # V/Q/K projections AND still needs the head-sharding wrap — opt in
         # via async_ulysses=True.
+        cp_size = vgm.cp_size if vgm else 1
         use_ulysses = (
             ulysses_size > 1
             and enable_sequence_parallel
-            and (self.qkv_mode != QKVMode.SEPARATE_QKV or async_ulysses)
+            and (self.qkv_mode != QKVMode.SEPARATE_QKV or async_ulysses or cp_size == 1)
```
NVShreyas, line 190: could you add a log to tell whether ulysses was disabled due to these reasons?
Also the `wrap_parallel_attention` method doesnt take use_ulysses argument right now so it should be passed to that method to disable ulysses if needed

NVShreyas (approved):
LGTM after requested change

tensorrt_llm/_torch/visual_gen/modules/attention.py
luyiyun1021: Addressed in 42cc7f14dc:

1. `wrap_parallel_attention` now takes `use_ulysses` and gates the `UlyssesAttention` wrap on it (forwarded from `Attention.__init__`), so the outer wrap is driven by the same flag as the inner backend head-count — a full-head backend can no longer be wrapped in Ulysses.
2. Added a `logger.debug` when Ulysses is disabled despite `ulysses_size > 1` and SP on (a `SEPARATE_QKV` cross-attn that is neither async nor `cp_size == 1`), so the all-gather fallback is observable.

litaotju (approved):

PR #15304, opened 2026-06-12
[TRTLLM-35882][feat] cute dsl gvr-topk load-balance optimization

limin2021:
## Summary by CodeRabbit

* **New Features**
  * Added clustering support and load-balancing capabilities for Top-K decode operations
  * Introduced `cluster_size` parameter for performance tuning
  * Added request-level dispatch ordering for variable sequence length optimization
  * New load-balancing strategy that partitions requests into optimized execution paths

* **Tests**
  * Expanded test coverage for clustering and load-balancing scenarios
  * Added variable sequence length handling in test utilities

#### seqlen varlen distribution:

image

image

#### Perf

1. sort version perf:

image

2. load-balance version perf:
image

TODO:
1. enable cute_dsl gvr-topk in model level, e.g., in dsa.py
2. test perf on model.

## Test Coverage

```
# ut running in CI, used for production.
python -m pytest tests/unittest/_torch/attention/sparse/test_cute_dsl_gvr_topk_decode.py
# standard alone ut, which is not dependent on trtllm env. It don't run in CI.
python tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
python -m pytest tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
```

## PR Checklist

## GitHub Bot Help


tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
```diff
@@ -247,22 +291,248 @@ def gvr_topk_decode(
...
+    """
+    assert seq_lens.is_cuda and seq_lens.dtype == torch.int32
+    batch_size = seq_lens.shape[0]
+    if order_row is None:
+        order_row = torch.full((max_batch_size,), -1, dtype=torch.int32, device=seq_lens.device)
+    if counters is None:
+        counters = torch.zeros(2, dtype=torch.int32, device=seq_lens.device)
+
+    compiled = _compile_lb_prepare(max_batch_size, batch_size, long_threshold)
+    compiled(seq_lens, order_row, counters, cutlass.Int32(batch_size))
```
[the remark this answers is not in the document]
limin2021: done.

tests/unittest/_torch/attention/sparse/test_cute_dsl_gvr_topk_decode.py
```diff
@@ -234,7 +305,213 @@ def test_cute_dsl_gvr_topk_decode(
...
+def test_lb_prepare_partition(B, ratio):
+    """Prepare kernel: counters + order_row match a numpy partition reference.
+
+    Builds synthetic seq_lens with a controlled long/short ratio, shuffles
+    them, then verifies the kernel partitions the request_ids into
+    [long...][short...] correctly.
+    """
+    long_threshold = 64 * 1024
+    torch.manual_seed(B * 1000 + int(ratio * 100))
+    n_long_expect = int(round(B * ratio))
```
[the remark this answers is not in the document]
limin2021: done.

tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
```diff
@@ -180,6 +211,19 @@ def gvr_topk_decode(
...
+        assert (
+            order_row is not None
+            and order_row.dtype == torch.int32
+            and order_row.is_cuda
+            and order_row.shape == seq_lens.shape
+        ), (
+            "seqlen_sorted=True requires order_row: int32[batch_size] on CUDA"
+            f" (expected shape {tuple(seq_lens.shape)}, got "
+            f"{tuple(order_row.shape) if order_row is not None else None})"
+        )
```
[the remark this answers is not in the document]
limin2021: No need to add this.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
```diff
@@ -72,14 +160,12 @@ class GvrParams:
...
-        — V4's kFTarget=kK alignment eliminates upper-clamp saturation on
-        tight-σ + high-A2 layers; cross-prompt swe-bench shows 1.5-2.2×
-        P2-iter reduction vs V3.2's kFTarget=384/2560. K=2048 is identical
-        across cr (V4 doesn't natively use K=2048; values reused for safety).
+        """Per-(dtype, K, cr) tuning constants, mirroring CUDA's
+        ``GvrParams<T, K>`` template specialization. For K ∈ {512, 1024}
+        cr=1 (DSv3.2) and cr=4 (DSv4, PR #14413) use different kFTarget —
+        V4 aligns kFTarget with kK to avoid upper-clamp saturation on
+        tight-σ layers (1.5-2.2× fewer P2 iters on swe-bench). K=2048 is
+        identical across cr (V4 doesn't natively use it).
```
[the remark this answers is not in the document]
limin2021: done.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
```diff
@@ -1591,60 +2039,158 @@ def gvr_topk_kernel(
...
+                            # upper bound). Phase 3's stream-write guards
+                            # writes at ``dst < kC`` so peer.smem_keys /
+                            # smem_vals are only valid in [0, kC), but
+                            # s_iscalars[5] is uncapped and could exceed
+                            # kC in the done==2 bracket-exhaustion path
+                            # (not reached in V4 production data, but a
+                            # cheap guard vs. an OOB DSMEM read). Mirrors
+                            # the single-CTA path's cand_count_p4 = min(
+                            # s_iscalars[0], kC) downstream.
+                            peer_cnt = min(peer_cnt, cutlass.Int32(self.kC))
```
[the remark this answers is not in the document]
limin2021: Thanks. The cap doesn't actually save peer candidates here:

- Normal path (>99.9% of production): local_cnt_self ≈ kC/cs ≈ 2560, far below kC=5120. Cap is a no-op.
- Overflow path (done==2, V4 production never reaches): peer writes use dst = base_offset + i_gather with a dst < kC guard. Whether base_offset is the uncapped 6000 or the capped 5120, peer writes land at dst ≥ kC and are all dropped either way — the gather is a forward concat, not reverse-pack.

To actually preserve peer candidates we'd need a real cluster-wide top-K merge (drop leader's lowest to make room), not a one-line cap. The existing comment already documents this as a known overflow path that V4 production doesn't reach; I'd prefer to leave the code as-is so that comment stays the single source of truth.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
```diff
@@ -1535,12 +1950,45 @@ def gvr_topk_kernel(
...
+        # vec_w-wide LDG→STS→LDS pipeline works for fp32/bf16/fp16. The
+        # wrapper auto-disables this when slice_len > smem_cache_elems
+        # (e.g. cs=2 + N>64K, cs=4 + N>128K).
+        if cutlass.const_expr(self.enable_smem_cache):
+            smem_input = smem.allocate_tensor(
+                element_type=self.dtype,
+                layout=cute.make_ordered_layout((self.smem_cache_elems,), order=(0,)),
+                byte_alignment=128,
+            )
+        else:
```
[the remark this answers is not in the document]
limin2021: Good catch. Added a launch-time assert in __call__ that backstops the SMEM-cache path against oversized slices — checks worst-case ceil(N / cluster_size) <= smem_cache_elems before launch. Guarded under cutlass.const_expr(self.enable_smem_cache) so the legacy path is unchanged. Error message points the caller to the three valid mitigations (raise the budget, raise cluster_size, or disable the cache).

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode_load_balance.py
```diff
@@ -0,0 +1,432 @@
...
+        # Cluster size for the long branch. cs=4 is the default
+        # break-even pick on B200 (T_row > 3.2us recoups DSMEM sync);
+        # cs=2 trades less parallelism for cheaper sync, cs=8 is GPC-
+        # constrained and only profitable on very long rows. The same
+        # value is used as the hardware cluster shape at launch (and
+        # is exposed via ``self.cluster_size`` to the main kernel).
+        cluster_size: int = 4,
+        # Cluster cs=4 break-even ≈ T_row > 3.2us ≈ 64K elements on B200.
+        # See ``gvr-topk-load-balance/design/idea_c_design.md`` §3.
+        long_threshold: int = 64 * 1024,
```
yuxianq: `long_threshold` is never used, can we remove it?

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode_load_balance.py
```diff
@@ -0,0 +1,432 @@
...
+        # Step 1: classify request `tidx`. Threads with tidx >= batch_size
+        # contribute 0 (their seq_lens slot is unset / out of range).
+        is_long_val = cutlass.Int32(0)
+        if tidx < batch_size:
+            seq = seq_lens[tidx]
+            if seq > cutlass.Int32(long_threshold):
+                is_long_val = cutlass.Int32(1)
+
+        # Step 2: block prefix sum over is_long_val -> position within long group.
+        # ``need_total_sum=True`` makes warp_sums[num_warps-1] hold n_long_total
```
yuxianq: This comment says `need_total_sum=True`, but the code uses `need_total_sum=False`, please make them consistent.

limin2021: done.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode_load_balance.py
limin2021: done.

yuxianq (approved):

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
```diff
@@ -497,7 +508,7 @@ def phase1_preidx_stats(
         smem_wsum_f32,  # cute.Tensor [NUM_WARPS] float32
         smem_wcnt_i32,  # cute.Tensor [NUM_WARPS] int32
         s_thr,  # cute.Tensor [3] float32: [threshold, val_lo, val_hi]
-        s_iscalars,  # cute.Tensor [5] int32: [cand_count, done, cnt_lo, cnt_hi, out_count]
+        s_iscalars,  # cute.Tensor [6] int32: [cand_count, done, cnt_lo, cnt_hi, out_count, local_cand_count]
```
hyukn, line 511: How do we know that the tensor shape has already been updated? Is it aligned?

hyukn (approved):
How can we detect the data distribution automatically? Shall we do calibration based on the latest input data? This will be a TODO, right?

mingyangHao (approved):
LGTM

tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
```diff
@@ -251,22 +292,261 @@ def gvr_topk_decode(
...
         return out_values, out_indices
     else:
         return None, out_indices
 
 
+def gvr_topk_sort_prepare(seq_lens: torch.Tensor) -> torch.Tensor:
+    """Build the LJF dispatch order for :func:`gvr_topk_decode`.
+
+    Returns ``int32[num_rows]`` whose i-th entry is the original-batch
+    index of the i-th longest row. Run once per decode step; the same
```
longcheng-nv: This returns `int32[batch_size]` (request-level — `argsort(seq_lens)` has length `seq_lens.shape[0] == num_rows // next_n`), not `int32[num_rows]`, and the entries are request indices, not row indices. Given how carefully the rest of the PR distinguishes request-level vs row-level, suggest:

```
Returns ``int32[batch_size]`` whose i-th entry is the original-batch
index of the i-th longest request.
```

longcheng-nv (commented):
**Review: kernel + UT landing looks solid; approve with comments (one doc fix needed before merge).**

Thanks for this — the load-balance design is sound and the change is opt-in / zero-regression for the existing path. I verified the core correctness paths and they hold up:

- **LJF reorder is logical, not physical**: the sort decision (`argsort` / long-short partition) happens outside the GVR kernel and is reused across layers; the kernel only does a `bidx -> row_idx` index indirection. No logits data is moved, and outputs are written back at the original `row_idx`, so result layout is unchanged. The `const_expr` elision means `order_row=None` keeps the legacy SASS effectively unchanged. 👍
- **LB kernel = one launch, two homogeneous branches**: long clusters cooperate (cs=4) on one row, short clusters pack 4 distinct rows (1 CTA/row, cs=1). Branch choice is uniform within a hardware cluster (`cluster_id = bidx // cluster_size`), so cluster-sync can't deadlock and partial short-tail clusters are safe.
- **Grid never overflows**: `total_clusters = n_long_req*next_n + ceil(n_short_req*next_n/cs) <= batch*next_n <= max_batch_size*next_n`.
- **prepare kernel is race-free**: `block_prefix_sum_kernel` ends Step 3 with a `barrier(barrier_id=1)`, so thread 0 reading `warp_sums[num_warps-1]` afterward is safe (the "distinct barrier IDs let the hardware pipeline" comment is a weak justification, but the ordering itself is correct).

**Issues / suggestions**

1. **(should fix before merge — doc bug)** `gvr_topk_sort_prepare` docstring says it returns `int32[num_rows]` / "i-th longest **row**", but it actually returns `int32[batch_size]` (request-level — `argsort(seq_lens)`). The whole PR is careful about the request-level vs row-level distinction, so this one is worth correcting to avoid misleading callers. (inline comment below)

2. **(scope clarity)** The LB kernel is only reachable via the standalone wrappers in `tests/scripts/` — there is **no `custom_op` wrapper**, so it can't be CUDA-graph captured / used in production yet. The single-CTA `seqlen_sorted` path *is* wired through the op. This matches the PR's own TODO ("enable in dsa.py / test perf on model"). Suggest stating explicitly in the description that this is "kernel + UT only, LB not yet production-wired", and listing the follow-up custom-op wrapping.

3. **(perf data gap)** The LB grid is fixed at `max_batch_size * next_n * cluster_size` for a single CUDA-graph shape. For small batch / all-short workloads it degenerates to the baseline mapping but still pays prepare + dead-cluster cost (commented as "<1us" but no data). The perf plots only cover the large-BS varlen case. A small-batch / all-short microbench (or per-batch-bucket compile) would make the production gate (`num_rows > num_sms` AND `max_seq_len >= 64K`) concrete.

4. **(nit)** `long_threshold = 64*1024` is a hardcoded magic constant tied to a B200 break-even, and it's compared against raw (non-compressed) kv-len — worth noting both in the docstring, and longer-term making it tunable/adaptive.

5. **(nit)** The test's `sys.path` injection (6 nested `dirname` + `sys.path.insert(0, ...)`) is brittle to file moves and risks shadowing; consider a conftest/fixture.

Nothing here blocks the kernel landing except item 1. Items 2-4 can be follow-ups.

tensorrt_llm/_torch/cute_dsl_kernels/blackwell/top_k/gvr_topk_decode.py
limin2021: This is a fix for the comment. This tensor is with 6 elements as defined as followings:
```
 The tensor shape is confirmed at the single allocation site on line 1963:
▎
▎ s_iscalars = smem.allocate_tensor(
▎     element_type=cutlass.Int32,
▎     layout=cute.make_ordered_layout((6,), order=(0,)),  # ← size 6
▎     byte_alignment=16,
▎ )
```
▎

tests/scripts/cute_dsl_kernels/top_k/run_gvr_topk.py
limin2021: Good catch — fixed. The function returns the same tensor; the doc was wrong about its shape/level.

limin2021:
> How can we detect the data distribution automatically? Shall we do calibration based on the latest input data? This will be a TODO, right?
Yes. It's a TODO. Long Cheng is studying the distribution, will enable in the next MR.

lfr-0531 (approved):
LGTM~

longcheng-nv (approved):
**Approve — LB kernel + ops + UT are good to land.**

Re-reviewed at head `fdd4ad3`. Both concerns from my earlier review are now resolved:

- `gvr_topk_sort_prepare` docstring is fixed (request-level `int32[batch_size]`, not row-level).
- The LB path is now wired through real custom ops (`trtllm::cute_dsl_gvr_topk_lb_prepare` + the `counters` branch of `cute_dsl_gvr_topk_decode`), with a fixed grid (`max_batch_size*next_n*cluster_size`) so it is CUDA-graph-capturable.

Correctness re-verified on the current diff and holds up:
- Three-way dispatch on `(counters, order_row)` is clean; `counters` without `order_row` is rejected.
- Branch choice is uniform within a hardware cluster (`cluster_id = bidx // cluster_size`), so the long-branch cluster-sync can't deadlock and partial short-tail clusters are safe.
- Grid never overflows: `total_clusters = n_long_req*next_n + ceil(n_short_req*next_n/cs) <= batch*next_n <= max_batch_size*next_n`.
- The request->row expansion (`order_row[req]*next_n + nn`) is consistent across both LB branches and the sort path, and matches `block_prefix_sum_kernel`'s inclusive-scan contract (`excl_pos_long = pos_long - is_long_val`, total read from `warp_sums[num_warps-1]`).

**Remaining items worth addressing (non-blocking):**

1. **Perf-data provenance.** The standalone bench in `tests/scripts/.../run_gvr_topk.py::_compile_lb` constructs `GvrTopKLBKernel(...)` *without* the tuning knobs, so it uses the ctor defaults (`enable_phase3_unroll=False`, `use_256bit_load=True`, `min_blocks_per_mp=3`). The production op path spreads `_pick_tuning(...)`, which gives `enable_phase3_unroll=True` and `use_256bit_load=False` for bf16/fp16. So the bf16 perf plots in this PR were produced with a 256-bit-load config the heuristic deliberately rejects. The plotted numbers may not reflect what production actually runs — please align the bench with `_pick_tuning` (or at least note the knob delta).

2. **Tightened runtime contract.** `forward()` now hard-asserts `seq_lens.shape[0] == num_rows // next_n`. Please confirm no existing in-tree caller (the dsa.py path) passes a padded `seq_lens` that would now trip this.

**Follow-up suggestion:** per the PR's own TODO (enable in `dsa.py` + test perf on model), I'd recommend an on-model end-to-end measurement after this lands — both to validate the production tuning path (item 1) and to confirm the LB/sort dispatch actually helps the real large-BS varlen decode workload, rather than relying on the standalone microbench numbers alone.

PR #14851, opened 2026-06-02
[https://nvbugs/6185446][fix] Add warmup for trtllm-gen fmha JIT kernels

pengbowang-nv:
Revert https://github.com/NVIDIA/TensorRT-LLM/pull/13505/changes/c37992cacc1e501d22e83465506172663b7a7077 to remove padding.

Added warmup with a grid that should be dense enough to catch almost every possible JIT kernels. Note that this PR only enables warmup and have no shape binning at runtime. I'd like to see how it performs and if it can warmup most kernels then we may not need further works.

Added a message when a possible JIT happened, to give a clear clue why performance dropped.

Some warmup data collected on our unit tests.

The kernel compilation times are similar, generally falling within the 7-9s range. The increase on warmup time can be estimated as $t_{kernel Compile} \cdot (N_{WarmupKernels} - N_{OriginalWarmupKernel})$

| Model and Configuration | Warmup Enabled | Number of Kernels in FMHA Warmup Phase | Number of JIT Compilations in Original Warmup Phase | Number of JIT Compilations During Runtime |
|---|---|---:|---:|---:|
| dsv3-bf16-mtp - single GPU - without cudagraph | On | 16 | 0 | 0 |
| dsv3-bf16-mtp - single GPU - without cudagraph | Off | 0 | 4 | 8 |
| dsv3-bf16-mtp - single GPU - with cudagraph | On | 16 | 0 | 0 |
| dsv3-bf16-mtp - single GPU - with cudagraph | Off | 0 | 13 | 1 |
| dsr1-fp4-mtp - 4 GPUs - with cudagraph | On | 12 | 0 | 0 |
| dsr1-fp4-mtp - 4 GPUs - with cudagraph | Off | 0 | 4 | 8 |
| gptoss-eagle3 - 4 GPUs - with cudagraph | On | 16 | 0 | 0 |
| gptoss-eagle3 - 4 GPUs - with cudagraph | Off | 0 | 15 | 1 |
| Qwen3.5-mtp - single GPU - with cudagraph | On | 9 | 0 | 0 |
| Qwen3.5-mtp - single GPU - with cudagraph | Off | 0 | 6 | 0 |
| Qwen3.5 - single GPU - with cudagraph | On | 3 | 0 | 0 |
| Qwen3.5 - single GPU - with cudagraph | Off | 0 | 2 | 0 |
| dsv3.2 - 4 GPUs - with cudagraph | On | 0 | 0 | 0 |
[...]

PR #15305, opened 2026-06-12
[https://nvbugs/6248837][fix] Densify trtllm-gen fmha warmup grid to catch missing kernels

pengbowang-nv:
### Background and Analysis

After previous warmup modification PR at #14851 , most JIT in the middle of execution issue is resolved. However we met a problem on some particular DSR1 case, where a kernel is not covered in the warmup grid and cause performance issue afterwards. In this special case the kernel would be selected only if seqlen between a very narrow area (8193-9316).

This PR densified the grid based on learning from autotuner, and should protect this case and other possible narrow cases without introducing much overhead.

### Why the soultion in this PR should work (even for future cases)

Given other parameters, the autotuner would select kernel based on `batchSize`, `seqLenQ`, `seqLenKv`. For decode case, `seqLenQ` is given. As attention is done in tiles, these parameters only affect autotuner by affecting number of total tile and tile per sequence. 

When tile number is in the range of 1 - 24, the autotuner logic would be rather delicate so that we require dense cover of this area. Outside this area the autotuner logic will be much less delicate and we can just cover some of them. 

Each tile can be of size 128, 256 and 512. The tile number per sequence is capped at `multiprocessorCount` and the total tile number also mostly work by comparing with `multiprocessorCount` multiplied by a factor.

So for `seqLenKv` we use such range: `[*dense_tile_num_range, *sprase_tile_num_range] multiplied by [128, 256, 512]`. `dense_tile_num_range = [1,2, ..., 24]` covers full 1-24 tile num range, and `sparese_tile_num_range = [26, 28, 32, 40, 48, 64, 96, 128, 192, 256]` would loosely cover from 24 - 256 (upper cap of multiprocessor count)

For batchsize we want total number of tiles cover 1 - 24 and loosely cover other cases. Remember we have full 1-24 tile for `seqLenKv` range, we would like to cover most cases for each `seqLenKv` (effectively tile number per sequence). For `tileNumPerSeq = 1`, we want `batchsize = 1, ..., 24`; for `tileNumPerSeq = 2`, we want `batchsize = 1, ..., 12`, etc. So the final coverage would be similar to tile range of  `seqLenKv`, but I added more batchsizes just to be sure about real cases.

### Overhead

As the warmup happens on C++ fmhaRunner level, the overhead would be much easier to handle. `Overhead = Tuning Overhead + Kernel Compilation Overhead`. For a given model config, as the warmup should catch every posssible kernel in the run, this can be seen as a constant. Before and after this PR this should not change much (it will increase as we did catch a previously missing kernel), likely an increase of 8s. And the `Tuning Overhead = mesh point * tuning time * model layer number * number of execution`, tuning time is typically less than 20us, interpreting to less than 6s for most models. In actual test environment the tuning overhead is measured less than 1s. I believe this is due to the fact we are overestimating the tuning time and mesh point.

### Follow up tasks

1. Change to tune only on a new layer, so we can remove model layer number factor from tuning overhead and allow for even more warmup.
2. Reduce kernel compilation overhead.

## PR Checklist

## GitHub Bot Help

## Summary by CodeRabbit

* **Refactor**
  * Updated generation kernel warmup configuration to enhance compilation efficiency and provide optimized performance across various batch and sequence length scenarios.

### DIFF
--- a/cpp/tensorrt_llm/kernels/trtllmGenKernels/fmha/fmhaKernels.h
+++ b/cpp/tensorrt_llm/kernels/trtllmGenKernels/fmha/fmhaKernels.h
@@ -301,12 +301,29 @@ public:
     }
 
 private:
+    // Warmup up grid selection should be able to cover most possible kernel cases.
+    // Given a model, autotuner decision is relative to number of numCtas/numCtasPerSeqKv,
+    // which translate to batch size, seqLenQ and seqLenKv. Autotuner is sensitive to
+    // numCtas/numCtasPerSeqKv in the range of 1-24, but became much insensitive later.
+
+    // Selection of batch size: numCtas = numCtasPerSeqKv * batchsize
+    // So we need batch size = 1-24 to cover sensitive area when numCtasPerSeqKv = 1
+    // And then gradually increase gap between batch sizes.
     inline static std::vector<int> const kDefaultWarmupBatchSizeCandidates
-        = {1, 2, 3, 4, 5, 6, 7, 8, 10, 12, 14, 16, 20, 24, 28, 32, 40, 48, 56, 64, 80, 96, 128, 256, 512, 1024};
-    inline static std::vector<int> const kDefaultWarmupPrefillBatchSizeCandidates
-        = {1, 2, 3, 4, 6, 8, 12, 16, 24, 32, 48, 64, 128, 256};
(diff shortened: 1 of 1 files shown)


pengbowang-nv:
/bot run --stage-list "DGX_B200-8_GPUs-PyTorch-PerfSanity-Post-Merge-3"

yuxianq (approved):

PR #15317, opened 2026-06-12
[feat] visual_gen: emulate eager precision casts under torch.compile

karljang:
> _Developed with an AI coding agent (Claude); the change, the A/B measurements, and an independent cold review were all verified manually on B200._

## Description

When `torch.compile` is enabled for a VisualGen pipeline, Inductor is free to fuse the BF16 block math (norms, modulation, activations, residuals) in a way that does **not** reproduce eager's intermediate precision casts, so compiled output drifts numerically from eager. PyTorch exposes `torch._inductor.config.emulate_precision_casts` to make Inductor emulate those eager-mode casts (e.g. bf16 intermediate rounding).

This PR:
- adds `TorchCompileConfig.emulate_precision_casts` (default **`True`**) to `tensorrt_llm/visual_gen/args.py`;
- sets `torch._inductor.config.emulate_precision_casts` from it in `BasePipeline.torch_compile()` **before** the compile loop (Inductor reads the flag at trace time, i.e. the first compiled forward during warmup).

Net diff is +23 lines across two files.

## Why (measured impact)

WAN 2.2 T2V A14B, B200, 50 steps, 720p/81f. **LPIPS measured against the eager (no-`torch.compile`) output of the same precision** — lower = closer to eager. `off` = current default (flag unset), `on` = this PR's default. No sparse attention involved.

| precision | LPIPS vs eager: off → on | Δ fidelity | gen time: off → on | Δ perf |
|---|---|---|---|---|
| **BF16** | 0.2185 → **0.1524** | **−30.3%** | 528.2s → 528.4s | +0.1s (noise) |
| **FP8 (block)** | 0.3965 → **0.3840** | −3.2% | 492.0s → 490.1s | −1.9s (noise) |
| **NVFP4** | 0.5096 → 0.5134 | +0.8% | 468.3s → 467.1s | −1.2s (noise) |

- **BF16 (the default precision) gets a large fidelity gain** — compiled output ~30% closer to eager — because the fused BF16 math dominates the numerics there.
- The effect tapers as quantization error grows: small for FP8, neutral for NVFP4 (within run-to-run noise).
- **Perf is unchanged** across all three (every delta is within noise on ~470–530s runs). The fidelity gain is effectively free.

## Notes for reviewers

- **Process-global flag, intentionally left set.** `torch._inductor.config.emulate_precision_casts` is process-global. It is set once in `torch_compile()` and not restored, because recompiles at new shapes during inference must still observe it. This matches the existing precedent in `tensorrt_llm/_torch/compilation/backend.py`, which also mutates `torch._inductor.config` unconditionally. In a VisualGen worker only the diffusion pipeline runs, so there is no cross-model leakage in practice.
- **Default flip changes compiled numerics.** Any golden/accuracy test that exercises a **compiled** VisualGen path (e.g. LPIPS goldens in `tests/integration/defs/examples/test_visual_gen.py`) was captured with the flag effectively off. Those goldens/thresholds may need refreshing to the (more eager-faithful) compiled output. Happy to regenerate if pointed at the reference set.
- Toggleable: set `TorchCompileConfig(emulate_precision_casts=False)` to restore prior behavior.

## Test Coverage

- Manual A/B above (eager reference vs compiled emu-off vs compiled emu-on) across BF16 / FP8 / NVFP4, 4 prompts.
- Pipeline loads and generates end-to-end with the flag on; `logger.info` confirms the resolved value at compile time.

### DIFF
--- a/tensorrt_llm/_torch/visual_gen/pipeline.py
+++ b/tensorrt_llm/_torch/visual_gen/pipeline.py
@@ -4,6 +4,7 @@ import time
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Set, Tuple, Type
 
 import torch
+import torch._inductor.config as inductor_config
 import torch.distributed as dist
 import torch.nn as nn
 from pydantic import Field
@@ -528,6 +529,15 @@ class BasePipeline(nn.Module):
         For non-transformer components, compiles the entire module.
         """
         tc_config = self.pipeline_config.torch_compile
+
+        # Inductor reads emulate_precision_casts at trace time (first compiled
+        # forward / warmup), so set it before compiling. Emulating eager-mode
+        # precision casts keeps compiled output numerically close to eager
+        # (diffusion fidelity); without it bf16 intermediate rounding diverges.
+        # Guarded for torch builds that predate the flag.
+        if hasattr(inductor_config, "emulate_precision_casts"):
+            inductor_config.emulate_precision_casts = True
+            logger.info("torch.compile: emulate_precision_casts=True")
 
         # Using default as max-autotune mode takes more initialization time and
         # does not improve performance a lot.


karljang:
The scope of the project was unnecessarily broad, so closed. :(

PR #15318, opened 2026-06-12
[None][fix] visual_gen/wan: pass precision-cast compile option

karljang:
## Description

The WAN transformer block performs several explicit dtype casts, especially around BF16 modulation and normalization intermediates. When `torch.compile` is enabled, Inductor may drop the eager-mode rounding of those casts while fusing, so compiled WAN output can drift numerically from eager. PyTorch supports `emulate_precision_casts` through `torch.compile(options=...)` to make Inductor emulate those eager casts.

This PR adds a VisualGen `TorchCompileConfig.options` dictionary and forwards it to `torch.compile(options=...)`. WAN uses that hook to enable `emulate_precision_casts` by default when the installed PyTorch exposes the option, while explicit user-provided options still take precedence. The shared WAN YAML configs now show the option explicitly:

```yaml
torch_compile_config:
  options:
    emulate_precision_casts: true
```

One implementation detail: PyTorch rejects passing `mode` and `options` together, so the compile path passes `mode="default"` only when no `options` are supplied.

## Why (measured impact)

WAN 2.2 T2V A14B, B200, 50 steps, 720p/81f. **LPIPS measured against the eager (no-`torch.compile`) output of the same precision**; lower = closer to eager. `off` = compiled without `emulate_precision_casts`; `on` = compiled with `options={"emulate_precision_casts": True}`. No sparse attention involved.

| precision | LPIPS vs eager: off -> on | Δ fidelity | gen time: off -> on | Δ perf |
|---|---|---|---|---|
| **BF16** | 0.2185 -> **0.1524** | **-30.3%** | 528.2s -> 528.4s | +0.1s (noise) |
| **FP8 (block)** | 0.3965 -> **0.3840** | -3.2% | 492.0s -> 490.1s | -1.9s (noise) |
| **NVFP4** | 0.5096 -> 0.5134 | +0.8% | 468.3s -> 467.1s | -1.2s (noise) |

- **BF16 (the default precision) gets a large fidelity gain**: compiled output is about 30% closer to eager because fused BF16 cast-heavy block math dominates the numerics there.
- The effect tapers as quantization error grows: small for FP8, neutral for NVFP4 within run-to-run noise.
- **Perf is unchanged** across all three precisions. Every delta is within noise on roughly 470-530s runs.

## Notes for reviewers

- The option is now part of the VisualGen config surface instead of a process-global `torch._inductor.config` mutation.
- WAN keeps the previous default behavior through its pipeline-specific compile-options hook, but users can override it with `torch_compile_config.options.emulate_precision_casts: false`.
- YAML users can pass arbitrary `torch.compile(options=...)` entries through `torch_compile_config.options`.
- This changes WAN compiled numerics. Any golden/accuracy test that exercises a compiled WAN path may need refreshing to the more eager-faithful compiled output.

## Test Coverage

- `pre-commit run --files` on all modified files.
- `pytest -q tests/unittest/_torch/visual_gen/test_visual_gen_args.py` passed before rebasing onto latest `main` (`52 passed`).
- Manual A/B above: eager reference vs compiled without the option vs compiled with the option across BF16 / FP8 / NVFP4, 4 prompts.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced Wan model compilation handling for improved precision casting behavior during PyTorch compilation.

### DIFF
--- a/examples/visual_gen/configs/wan2.2-i2v-fp4-1gpu.yaml
+++ b/examples/visual_gen/configs/wan2.2-i2v-fp4-1gpu.yaml
@@ -23,5 +23,8 @@ attention_config:
 parallel_config:
   cfg_size: 1
   ulysses_size: 1
+torch_compile_config:
+  options:
+    emulate_precision_casts: true
 cuda_graph_config:
   enable: false
--- a/examples/visual_gen/configs/wan2.2-t2v-fp4-1gpu.yaml
+++ b/examples/visual_gen/configs/wan2.2-t2v-fp4-1gpu.yaml
@@ -23,5 +23,8 @@ attention_config:
 parallel_config:
   cfg_size: 1
   ulysses_size: 1
+torch_compile_config:
+  options:
+    emulate_precision_casts: true
 cuda_graph_config:
   enable: false
--- a/examples/visual_gen/configs/wan2.2-t2v-fp4-4gpu.yaml
+++ b/examples/visual_gen/configs/wan2.2-t2v-fp4-4gpu.yaml
@@ -25,5 +25,8 @@ parallel_config:
   ulysses_size: 2
   async_ulysses: true
   parallel_vae_size: 4
+torch_compile_config:
+  options:
+    emulate_precision_casts: true
 cuda_graph_config:
   enable: false
--- a/examples/visual_gen/configs/wan2.2-t2v-fp8-8gpu.yaml
+++ b/examples/visual_gen/configs/wan2.2-t2v-fp8-8gpu.yaml
@@ -24,5 +24,8 @@ parallel_config:
   cfg_size: 2
   ulysses_size: 4
   parallel_vae_size: 8
+torch_compile_config:
+  options:
+    emulate_precision_casts: true
 cuda_graph_config:
   enable: false
--- a/examples/visual_gen/serve/configs/wan21.yml
+++ b/examples/visual_gen/serve/configs/wan21.yml
@@ -4,3 +4,6 @@ cache_config:
 parallel_config:
   cfg_size: 1
   ulysses_size: 1
+torch_compile_config:
+  options:
+    emulate_precision_casts: true
--- a/examples/visual_gen/serve/configs/wan22.yml
+++ b/examples/visual_gen/serve/configs/wan22.yml
@@ -3,3 +3,6 @@ cache_config:
 parallel_config:
   cfg_size: 1
   ulysses_size: 1
+torch_compile_config:
+  options:
+    emulate_precision_casts: true
--- a/tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
+++ b/tensorrt_llm/_torch/visual_gen/models/wan/pipeline_wan.py
@@ -5,6 +5,7 @@ from typing import List, Optional, Union
 import diffusers
 import PIL.Image
 import torch
+import torch._inductor.config as inductor_config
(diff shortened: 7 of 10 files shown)


chang-l (commented):
I’m a bit concerned about the performance impact if we always turn on this option. See: https://github.com/pytorch/pytorch/blob/eea34e6d9be147680f8afa66ae2e0e3d6445e9f5/torch/_inductor/config.py#L3018-L3025
It seems this may block some kernel fusion chains. Can we make it optional?

karljang:
> I’m a bit concerned about the performance impact if we always turn on this option. See: https://github.com/pytorch/pytorch/blob/eea34e6d9be147680f8afa66ae2e0e3d6445e9f5/torch/_inductor/config.py#L3018-L3025 It seems this may block some kernel fusion chains. Can we make it optional?

Yeah, I had the same concern, so I checked the end-to-end time difference and confirmed that it’s negligible.
Additionally, we can make it optional by using an environment variable instead of a user-facing option. Is that what you meant?  And turn the setting on by default?

chang-l:
Two suggestions — both aim to keep the fix but make it config-driven and scoped, instead of a WAN-specific hardcode + process-global write.

**1. Pass it via `torch.compile(options=...)` rather than mutating `torch._inductor.config` globally.** `emulate_precision_casts` is a valid inductor option, and `options=` applies it as a per-compile `config_patches` (`_TorchCompileInductorWrapper.__call__` → `compile_fx(..., config_patches=...)`), scoped to the compiled blocks only. This removes the "process-global flag, intentionally left set / not restored" wart — and the stated reason for going global ("recompiles at new shapes must still observe it") still holds with `options=`, because the backend instance carries the patches across dynamic-shape recompiles. Note `mode=` and `options=` are mutually exclusive, so fold a non-`default` mode in via `torch._inductor.list_mode_options(mode)`.

**2. Make it a config field instead of WAN-specific code.** Add `emulate_precision_casts: bool = False` to `TorchCompileConfig` (`tensorrt_llm/visual_gen/args.py`) and apply it in the **base** `BasePipeline.torch_compile()` (`pipeline.py:522`). Then the `WanPipeline.torch_compile()` override goes away, and WAN opts in through its yaml — the `torch_compile_config:` section is already wired (`VisualGenArgs.from_yaml` → `cls(**config_dict)`):

```yaml
torch_compile_config:
  emulate_precision_casts: true
```

Default `False` means zero behavior change for every other model/golden — only configs that set it true get the new (more eager-faithful) numerics, which is exactly where you measured the +30% BF16 fidelity win. It also sets a cleaner precedent than the LLM-path `TorchCompileConfig`, which has no such field yet.

karljang:
@chang-l,
Thanks for your suggestions~, I'll give a try to the first one first~

@zhenhuaw-me,
Could you share your opinion on the 2nd suggestion of exposing an argument to control emulate_precision_casts?
The WAN transformer blocks require emulate_precision_casts=True because it has LayerNorms followed by cast operation. But still the option might be a good control knob for future models as well.

chang-l:
Alternatively, we can also enable the dict passthrough of `torch.compile(options=...)` via the VisualGen API, since it directly maps to the Pydantic field name, e.g.,
```
torch_compile_config:
  options:
    emulate_precision_casts: true
``` 
= `torch.compile(options={emulate_precision_casts: True})`

karljang:
@chang-l , thanks for your suggestions! could you review the PR again?

karljang:
I'm closing this PR in favor of https://github.com/NVIDIA/TensorRT-LLM/pull/15762

Issue #15344, opened 2026-06-13
[Bug]: [pyexecutor] SM121 appears to be unintentionally excluded from MLA block-reuse and chunked-prefill allowlists


DhineshPonnarasan:
### System Info

* CPU architecture: N/A (code inspection issue)
* GPU: N/A (issue identified through source analysis)
* TensorRT-LLM branch: main
* TensorRT-LLM commit: current main branch at time of investigation
* OS: N/A
* Additional information:

  * This issue was identified through source-code inspection and review of the MLA capability gating logic.
  * No specific hardware was required to observe the behavior.
  * The report concerns the SM allowlists used by the MLA block-reuse and chunked-prefill feature gates.

### Who can help?

@kaiyux 

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

## Summary

While reviewing the MLA capability gating logic in `tensorrt_llm/_torch/pyexecutor/py_executor_creator.py`, I noticed that both MLA KV-cache reuse and MLA chunked prefill are gated by the following SM allowlist:

```python
[90, 100, 103, 120]
```

SM121 is excluded from both checks.

Relevant code:

```python
if kv_cache_config.enable_block_reuse and sm_version not in [
    90, 100, 103, 120
]:
    ...
```

```python
if enable_chunked_context and sm_version not in [
    90, 100, 103, 120
]:
    ...
```

I could not find any code, comments, tests, documentation, or commit history indicating that SM121 is intentionally unsupported for MLA block reuse or MLA chunked prefill.

At the same time, multiple other locations in the repository treat SM120 and SM121 as the same Blackwell family.

Examples include:

* `fused_moe_cute_dsl_b12x.py`
* `deep_ep_low_latency.py`
* `eagle3_dynamic_tree.py`
* several integration tests using `(120, 121)` checks

Additionally, the MLA XQA JIT path contains:

```cpp
// SM121 uses the same cubin target as SM120 (sm_120f) for compatibility.
```

in:

```cpp
cpp/tensorrt_llm/kernels/decoderMaskedMultiheadAttention/decoderXQAImplJIT/nvrtcWrapper/src/nvrtcWrapper.cpp
```

## Steps to reproduce the behavior

1. Review the MLA feature gating logic in:
[...]

PR #15347, opened 2026-06-14
[feat] Enable MLA chunked prefill and KV cache reuse on SM121

CodersAcademy006:
I updated the SM version check logic in `py_executor_creator.py` to allow MLA chunked prefill and KV cache block reuse on SM121 (Blackwell) architectures. Specifically, I added SM121 (`121`) to the validation lists for both `enable_block_reuse` and `enable_chunked_context` checks, preventing the executor from automatically disabling these features on Blackwell GPUs.

This resolves issue #15344, where these optimization features were being disabled on SM121 devices because the Python-side validator was missing SM121 in its allowlist. The underlying C++ kernels already support SM121, so this change enables full compatibility for MLA optimizations on this hardware.

### DIFF
--- a/tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
+++ b/tensorrt_llm/_torch/pyexecutor/py_executor_creator.py
@@ -673,10 +673,10 @@ def create_py_executor(
 
         sm_version = get_sm_version()
         if kv_cache_config.enable_block_reuse and sm_version not in [
-                90, 100, 103, 120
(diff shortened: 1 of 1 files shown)


karljang:
@CodersAcademy006 ,
Thanks for putting this together.
However, we're going to land #15434 instead. Nothing wrong with the approach here; we just want one PR with tests rather than two near-identical ones. Appreciate you flagging the SM121 gap!
I'm closing this in favor of #15434

Issue #4442, opened 2025-05-19
Scaffolding support A2A protocal


WeiHaocheng:
Scaffolding should be easy to support A2A protocal after support MCP. We can develop an example to verify this matter.

wu1du2:
I'm interested in A2A. Can this task be assigned to me?

Achyuthan-S:
Hi @WeiHaocheng — I'd like to work on #4442.
Proposed plan:
1. Scaffolding-as-a-server wrapper (similar to trtllm-serve)
2. A2A protocol layer via `a2a-sdk`, following the existing MCP contrib pattern
3. Example under `examples/scaffolding/contrib/a2a/` using OpenaiWorker for local dev
4. Unit tests for the A2A handler + controller wiring

Does this scope look right?

Achyuthan-S:
Update: opened a draft implementation as [#15350](https://github.com/NVIDIA/TensorRT-LLM/pull/15350).

**Scope for this PR (client-only):**
- `tensorrt_llm/scaffolding/contrib/a2a/` — `A2AWorker`, `A2AController`, tasks, lazy `a2a-sdk` wrapper
- Example under `examples/scaffolding/contrib/a2a/` + offline unit tests
- Mirrors the MCP contrib pattern (#4410)

**Out of scope for this PR:** exposing Scaffolding as an A2A server (can follow up if desired).

@WeiHaocheng — could you confirm this matches your intent for #4442 and assign if appropriate?

PR #4410, opened 2025-05-17
Scaffoldingllm supports MCP

wu1du2:
# Support MCP in TensorRT-LLM
Scaffolding support MCP #3335 https://github.com/NVIDIA/TensorRT-LLM/issues/3335

## Description
MCP provides a standard tool-use ability to TensorRT-LLM, to utilize LLM function-call.

## examples
run a mcp server first. TRT-LLM connects to it, then processing function-call automatically.
see examples\scaffolding\contrib\mcp\README.md

PR #15350, opened 2026-06-14
[#4442][feat] Add client-side A2A (Agent2Agent) protocol support to Scaffolding

Achyuthan-S:
Description:
### What this does

Follow-up to the MCP work (#4410). Issue #4442 asked for A2A protocol support in
Scaffolding now that MCP tool-calling is in place, plus an example to prove it
works end to end. This adds a client-side A2A integration under
`scaffolding/contrib/a2a/`, built the same way as the MCP contrib so it should
feel familiar to anyone who's looked at that code.

The idea: the generation model sees each reachable remote agent as a callable
tool, decides whether to delegate, and the `A2AWorker` forwards the message over
A2A and hands the reply back to the model to fold into a final answer.

This is intentionally the client side only (Scaffolding consuming other agents).
Exposing a Scaffolding pipeline *as* an A2A server is a natural next step but I
kept it out of scope here to keep the PR focused — happy to do that as a
follow-up if there's interest.

### What's included

- `A2AWorker` + `A2ASendTask` / `A2AListTask` — discover remote agents and send
  them messages (mirrors `MCPWorker` / `MCPCallTask` / `MCPListTask`).
- `A2AController` — discovers agents, lets the LLM pick one, dispatches the call,
  and summarizes the result. Same `process()` shape as `MCPController`.
- `A2AAgentConnection` — thin wrapper over `a2a-sdk`. The SDK is imported lazily
  so importing the package (and the unit tests) don't need it installed.
- An example under `examples/scaffolding/contrib/a2a/` with a runnable
  orchestrator and a small reference agent server, plus a README.
- Offline unit tests in `tests/unittest/scaffolding/test_a2a_worker.py` that
  inject fake connections, so they need neither `a2a-sdk` nor network access.

### A couple of decisions worth calling out

- I reuse the MCP contrib's `ChatTask` / `chat_handler` (imported from the
  submodules directly) rather than duplicating them. `contrib/README.md`
  explicitly allows cross-project reuse, and the MCP chat handler is the one that
  works against generic OpenAI-compatible endpoints. Easy to inline if you'd
  prefer the package fully self-contained.
- I did **not** add `a2a-sdk` to `requirements.txt` — it's an optional dependency
  pulled in lazily with a clear install message, so it doesn't bloat the base
  install. `mcp` is in requirements today, so if you'd rather have parity I can
  add it.

- [ ] `pytest tests/unittest/scaffolding/test_a2a_worker.py` (offline, runs in CI — no `a2a-sdk` or network required)
- [ ] Manual smoke test (local, needs `a2a-sdk` + an OpenAI-compatible endpoint): run `weather_agent_server.py`, then `a2a_run.py` against it

Note: `a2a-sdk`'s server/client API has shifted across versions; the example
scripts target the SDK's published "helloworld" pattern.

cc @WeiHaocheng — please confirm scope / assign if this looks right.

cc @chang-l  @Shixiaowei02 

Addresses #4442.

## Dev Engineer Review

- Added client-side A2A support for agent discovery, tool selection, message dispatch, and response summarization.
- Added `A2AWorker`, `A2ASendTask`, `A2AListTask`, `A2AController`, and lazy-loaded `A2AAgentConnection`.
- Added runnable orchestration and weather-agent examples with documentation.
- No configuration or test-list changes were identified.

## QA Engineer Review

- Added five async tests:
  - `test_a2a_worker_list_agents`
  - `test_a2a_worker_send_routes_to_named_agent`
  - `test_a2a_worker_send_unknown_agent`
  - `test_a2a_worker_async_shutdown`
  - `test_scaffolding_with_a2a_controller`
- No matching `test-db/` or `qa/` entries were identified.
- Verdict: needs follow-up.


Achyuthan-S:
cc @WeiHaocheng --- could you please check
cc @chang-l  @Shixiaowei02

Achyuthan-S:
cc @WeiHaocheng cc @chang-l @Shixiaowei02

PR #15362, opened 2026-06-15
[None][perf] avoid full input_token_ids copy in ADP router token count

lancelly:
## Summary
The ADP request router only needs the **input length** to load-balance new
requests across attention-DP ranks, but it read `ExecutorRequest.input_token_ids`
and called `len()` on it. `Request::getInputTokenIds()` returns `VecTokens`
**by value**, so each access materializes a fresh ~ISL-sized Python list (one
PyObject per token) under the GIL — redundantly on **every rank** (routing is
symmetric), for **every new request**.

On DeepSeek-V4 disaggregated generation at high concurrency (avg ISL ~39k), the
per-iteration `_fetch_new_requests` host path is a known rank-0 GIL hotspot.

## Change
- Add an **O(1)** `Request::getNumInputTokens()` that reads `mInputTokenIds.size()`
  directly — **no `VecTokens` copy and no Python list** — and bind it as the
  `num_input_tokens` nanobind property.
- Route the `DefaultADPRouter` / `KVCacheAwareADPRouter` token-count reads
  through a small `_num_input_tokens()` helper backed by that accessor.

## Test
- `TestNumInputTokens` covers the helper (None request, prefers the accessor
  without touching `input_token_ids`).
- Router unit tests use a `_MockRequest` whose `num_input_tokens` mirrors
  `len(input_token_ids)`.

## Notes
- The count read uses the new accessor and requires the C++ binding to be built
  (no Python fallback).
- Same change as #15199, targeting `feat/deepseek_v4_bench`.

### DIFF
--- a/cpp/include/tensorrt_llm/executor/executor.h
+++ b/cpp/include/tensorrt_llm/executor/executor.h
@@ -745,6 +745,7 @@ public:
     ~Request();
 
     [[nodiscard]] VecTokens getInputTokenIds() const;
+    [[nodiscard]] SizeType32 getNumInputTokens() const;
     [[nodiscard]] SizeType32 getMaxTokens() const;
     [[nodiscard]] bool getStreaming() const;
     [[nodiscard]] SamplingConfig getSamplingConfig() const;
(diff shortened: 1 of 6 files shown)


longlee0622:

PR #15363, opened 2026-06-15
[https://nvbugs/6179661][fix] Fix disagg generation-side KV transfer timeout deadlocks and teardown crashes

nv-xtf:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown and teardown reliability for distributed cache management operations.
  * Enhanced timeout detection and error handling for generation transfers in distributed settings.
  * Added safeguards against invalid iterator access during transfer response processing.

* **Chores**
  * Updated environment configuration for distributed communication setup.
  * Refactored transfer status polling with improved multi-rank synchronization and budget management.

## Description

In disaggregated serving, when a generation-side KV cache transfer does not complete in time (e.g. under heavy transport load), the PyTorch executor/cache-transceiver path could deadlock or crash the engine. This PR fixes the following issues:

1. **Unbounded blocking in `checkGenTransferStatus` (C++).** To satisfy `atLeastRequestNum`, the generation-side status check force-selected not-yet-ready requests and then blocked in `future::get()` with no bound. A transfer that never completes wedged that rank inside `get()` while sibling ranks advanced to the next collective, deadlocking the whole sync group. It is now a bounded poll: only futures ready on every rank of the sync group are completed, the wait for `atLeastRequestNum` is capped, and incomplete transfers are left in flight for later checks. All loop-exit decisions are derived from rank-identical data so every rank runs the same number of collectives.

2. **Generation-side transfer timeouts were not acted upon consistently (Python).** The timeout flag was set but its only consumer was unreachable behind the blocking call above, and it was rank-local. `_cancel_timed_out_gen_transfers` now runs every iteration on every rank and reaches a TP-wide consensus before changing request state: an OR-gate allreduce to skip when nothing timed out, an allgather union of the timed-out ids (so wall-clock skew between ranks does not diverge the set), and a second allgather on the per-rank cancel result so a request is only failed once every rank tracking it has cancelled its share.

3. **Collective gated on rank-local error state (Python).** `_check_cache_transfer_errors` only entered `_handle_errors` (`→ _enqueue_responses → tp_gather`) when the local rank had errored requests. Under attention DP a cancel succeeds on a subset of ranks, so some ranks entered the gather while others skipped it, deadlocking the TP group. The generation-side path now flushes errors on every rank unconditionally.

4. **Use-after-free on teardown (C++).** `~CacheTransceiver` relied on member destruction order, but `ConnectionManager` is declared after — and thus destroyed before — `CacheSender`/`CacheReceiver`, whose background threads still dereferenced it, causing a segfault during shutdown. The destructor now resets the sender and receiver first, while the manager is still alive.

5. **Two shutdown-path crashes in `CacheSender::response()` (C++).** (a) `recvRequestInfo` returning a null connection was only handled when the manager had already stopped; otherwise an empty `RequestInfo` flowed into processing and threw `bad_optional_access`. (b) Terminating while waiting in the inner loop left the iterator at `end()`, which was then dereferenced in `sendResponse`. Both now bail out cleanly when terminating.

A bounded-poll metric (waited calls / total poll iterations / budget-exhaustion count) is logged on budget exhaustion to keep the throughput impact of the non-blocking rewrite observable.

## Test Coverage

Validated with the disaggregated perf-sanity case `perf/test_perf_sanity.py::test_e2e[disagg_upload-e2e-gb200_kimi-k25-thinking-fp4_8k1k_con4096_ctx1_dep4_gen1_dep16_eplb0_mtp0_ccb-NIXL]` on GB200 (gen TP16 / attention-DP, ctx TP4, NIXL)

## PR Checklist

## GitHub Bot Help

### DIFF
--- a/cpp/include/tensorrt_llm/batch_manager/cacheTransceiver.h
+++ b/cpp/include/tensorrt_llm/batch_manager/cacheTransceiver.h
@@ -285,7 +285,6 @@ private:
     // request while a C++ status check still dereferences it.
     std::vector<std::pair<std::shared_ptr<LlmRequest>, std::future<void>>> mSenderFutures;
     std::vector<std::pair<std::shared_ptr<LlmRequest>, std::future<void>>> mRequesterFutures;
-    // Dedup sets so observe-only timeout WARN logs fire at most once per stuck request.
     std::unordered_set<LlmRequest::RequestIdType> mTimedOutSenderIds;
     std::unordered_set<LlmRequest::RequestIdType> mTimedOutRequesterIds;
     std::unordered_set<LlmRequest::RequestIdType> mCompletedSenderRequestIds;
@@ -294,6 +293,10 @@ private:
     std::unordered_set<LlmRequest::RequestIdType> mCompletedRequesterRequestIds;
     std::unordered_set<LlmRequest::RequestIdType> mFailedRequesterRequestIds;
     std::unordered_map<LlmRequest::RequestIdType, std::shared_ptr<LlmRequest>> mRequesterRequestsAwaitingConsensus;
+    // checkGenTransferStatus bounded-poll metrics, logged on budget exhaustion.
+    size_t mGenPollWaitedCalls{0};
+    size_t mGenPollWaitedIterationsTotal{0};
+    size_t mGenPollBudgetExhaustedCount{0};
     mpi::MpiComm const* mMpiWorldComm{nullptr};
 
     std::shared_ptr<CacheTransceiverComm> mGroupComm;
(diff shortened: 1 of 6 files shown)


Shixiaowei02 (approved):

chuangz0 (approved):

chienchunhung (commented):
Following up from the related #15356 thread: thanks for putting this together, and for calling out the ADP timeout/cancellation issue clearly there.

After reviewing this PR against the current cancellation PR chain, I plan to split the work across multiple PRs for better scoping and deduced risks:
- #15422 carries the teardown/lifetime hardening from this PR. That keeps shutdown safety independent and lets it land before #15238 if possible.
- #15238 should own the generation-side timeout/cancellation semantics, including rank-consistent agreement on the timeout/cancellable request set before `_handle_errors()` / `_enqueue_responses()` can enter `tp_gather`.
- #15356 -> #15238 remains the bounded-polling / request-cancellation chain. #15356 itself should stay focused on V2/Python transceiver bounded context polling, not cancellation semantics.

**tl;dr:** the ADP `tp_gather` participant mismatch is a real gap, intentionally not addressed in #15356, and should be handled in #15238 through rank-consistent cancellation. The teardown hardening is independent from the dependency graph and is split into #15422. I’ll port/reimplement the relevant cancellation-consensus logic from this PR into #15238.

PR #15376, opened 2026-06-15
[None][perf] disagg: CTX-side incremental conversation tokenization

lancelly:
@coderabbitai summary

## Description

In disaggregated serving with **conversation-sticky routing**, the router
(`ConversationRouter`, PR #14957) routes by the session table and **skips its
own tokenize / block-hash compute**, so it does not set `prompt_token_ids`. The
**context server** therefore re-renders and re-tokenizes the *full* conversation
prompt — tens of thousands of tokens for long multi-turn / agentic
conversations — on **every** turn. Host-overhead profiling of a DeepSeek-V4-Pro
6×CTX+1×GEN deployment at high concurrency showed the CTX serving process
saturated at 100–135% CPU, almost entirely in the HF fast-tokenizer, on the TTFT
critical path.

This PR caches, per session, the previous turn's rendered prompt and its token
ids on the context path, and re-encodes only the suffix that extends it — the
same approach as the merged `KvCacheAwareRouter._encode_with_prefix_cache`
(PR #15040), applied where the work actually lands in a conversation-sticky
disagg deployment.

- **Opt-in / zero default impact** — enabled only when `TLLM_CTX_INCR_TOKENIZE=1`;
  otherwise the cache is never constructed and the path is inert.
- **Context path only** — runs in `openai_chat` after `apply_chat_template`; the
  generation path takes `prompt_token_ids` and never reaches it.
- **Session-correct key** — keyed by the hash of the first messages, which carry
  the per-session cache-bust id, so concurrent sessions that share a routing
  `conversation_id` do not clobber each other (verified on real traffic: two
  concurrent sessions of one conversation → disjoint keys, no clobber).
- **Exact, fallback-safe** — the delta encode is bit-exact because chat renders
  end on a special token (an atomic added-token boundary that BPE never merges
  across; confirmed `<think>`/`</think>`/`<｜Assistant｜>` are dedicated token
  ids). Any prompt that does not extend the cached one falls back to a full
  encode; the cache is a bounded LRU (1024 entries).

Measured on the 6p1d DeepSeek-V4-Pro deployment with the feature on: CTX serving
CPU ~135% → ~27%, steady-state prefix reuse ~96–97%, and at the SLO operating
point a TTFT p50/p95 reduction of roughly 15–25% with throughput +6–11% (the CTX
tokenize was the binding host stage there). The win is concentrated where the
CTX stage is the bottleneck; it is neutral once GEN becomes the limiter at higher
concurrency.

## Test Coverage

- Gated behind `TLLM_CTX_INCR_TOKENIZE` (default off) — no change to existing
  default code paths; current `serve` / `openai_server` tests are unaffected.
- Correctness rests on the special-token boundary property above (the same
  property the merged PR #15040 relies on for the router). The incremental result
  was validated bit-identical to a full `encode` on real DeepSeek-V4-Pro
  disaggregated traffic (no divergence observed over tens of thousands of CTX
  tokenize calls).
- Follow-up (can add to this PR on request): a unit test mirroring #15040's —
  assert the incremental ids equal a full `tokenizer.encode` over a sequence of
  growing renders, plus a non-extending case that exercises the full-encode
  fallback.

## PR Checklist

- [ ] Please check this after reviewing the above items as appropriate for this PR.

## GitHub Bot Help

### DIFF
--- a/tensorrt_llm/serve/openai_server.py
+++ b/tensorrt_llm/serve/openai_server.py
@@ -23,7 +23,7 @@ import socket
 import time
 import traceback
 import uuid
-from collections import deque
+from collections import OrderedDict, deque
 from contextlib import asynccontextmanager
 from datetime import datetime
 from http import HTTPStatus
@@ -360,6 +360,15 @@ class OpenAIServer(_VideoRoutesMixin):
 
     def _init_llm(self, chat_template: Optional[str] = None):
         self.tokenizer = self.generator.tokenizer
+        # CTX-side incremental tokenizer (opt-in via TLLM_CTX_INCR_TOKENIZE=1).
+        # Re-encodes only the suffix that extends the previous turn's prompt,
+        # avoiding a full re-tokenize of the (often tens-of-thousands-token)
+        # prompt on every turn. See _encode_with_prefix_cache.
+        self._tok_prefix_cache = (
+            OrderedDict()
+            if os.environ.get("TLLM_CTX_INCR_TOKENIZE") == "1" else None)
+        if self._tok_prefix_cache is not None:
+            logger.info("CTX incremental tokenization enabled")
         hf_tokenizer_path = self.generator._hf_model_dir
         if not hf_tokenizer_path:
             hf_tokenizer_path = getattr(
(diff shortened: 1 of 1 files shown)


longlee0622:

PR #15381, opened 2026-06-15
[None][feat] DSv4 prep: IndexerTopK and TopK primitives

lfr-0531:
## Release Notes

* **New Features**
  * Added compression ratio support for optimized KV cache handling during decoding operations
  * Introduced auxiliary buffer management for improved decoding performance across different hardware configurations

* **Bug Fixes**
  * Fixed edge cases in decoding block scheduling and hardware saturation scenarios

* **Tests**
  * Expanded test coverage for compressed KV operations and decoding path transitions

## Description

This is PR 3 from the staged split of #14751. It lands the IndexerTopK and heuristic TopK primitives as an independent, reviewable change while keeping DSv4 sparse attention dispatch, compressor/mHC, MoE routing, and DSv4 model/tokenizer integration in later PRs.

Changes included here:
- Update the IndexerTopK CUDA kernel and thop binding.
- Update heuristic TopK decode kernel entry points and policy behavior.
- Register the updated fake custom-op schema for `indexer_topk_decode`, including `compress_ratio`, `radix_aux_indices`, and `radix_aux_logits`.
- Expand focused IndexerTopK/heuristic TopK unit coverage in `tests/unittest/_torch/thop/parallel/test_indexer_topk.py`.

Deferred by design:
- `test_deepseek_v4_fp4_indexer.py` remains deferred to the later sparse attention/model PR because it imports DSv4 sparse config/backend pieces.
- `RoutingKernelTopK.cuh` and MoE routing changes remain deferred to the MoE/routing PR.
- Sparse MLA backend dispatch and attention op plumbing remain in separate PRs.

No new dependency is introduced.

## Test Coverage

- Built TensorRT-LLM from this branch with:
  `python3 ./scripts/build_wheel.py --trt_root /usr/local/tensorrt --benchmarks --use_ccache --cuda_architectures "90-real;100-real" --configure_cmake`
- Confirmed the build compiled `indexerTopK.cu`, `heuristicTopKDecode.cu`, and `IndexerTopKOp.cpp`; the wheel was produced successfully as `build/tensorrt_llm-1.3.0rc18-cp312-cp312-linux_x86_64.whl`.
- Installed the built wheel into `.venv-3.12` and ran an import smoke test from this worktree. `tensorrt_llm.__file__` and `cpp_custom_ops.__file__` both resolved under this branch, and `torch.ops.trtllm.indexer_topk_decode` / `indexer_topk_prefill` were present.
- Checked GPU state before testing; selected idle GPU 0.
- Ran:
  `CUDA_VISIBLE_DEVICES=0 timeout 1800 .venv-3.12/bin/python -m pytest -q --tb=short -ra tests/unittest/_torch/thop/parallel/test_indexer_topk.py`
  Result: `2871 passed, 2 warnings in 926.27s (0:15:26)`.
- Ran:
  `pre-commit run --files $(git diff --name-only HEAD)`
  Result: passed.
- Ran scope leakage check. The diff contains only the intended IndexerTopK/heuristic TopK files, and forbidden grep for sparse MLA backend, compressor/mHC, `RoutingKernelTopK`, MoE, DSv4 model/tokenizer, attention op plumbing, and DSA relocation tests returned no output.

Build note: the attribution generation step emitted `ninja -t inputs returned no results for wheel targets`, but the wheel build completed with exit code 0. This is the same non-fatal attribution warning observed in the preceding split PR.

## PR Checklist

## GitHub Bot Help


longcheng-nv (approved):
LGTM — approving.

Reviewed the full kernel refactor end-to-end (including the unchanged multi-block/merge addressing in the head source) and cross-checked the prefill path against `feat/deepseek_v4`. This is a clean, well-documented stage-3 split of #14751.

**Strengths**
- The two correctness fixes in `topKPerRowJob` are real and minimal: (1) the `step==1` reset of `smemFoundTopKValues[0]`/`smemFinalDstIdx[0]` (replacing the old `step0WillContinue` suppression) correctly kills the 2x-duplicated-index defect in the multi-block merge path when `-FLT_MAX` padding dominates the threshold bin; (2) `finalIndices[ii] = -1` pairing with the `-FLT_MAX` sentinel so unused radix-sort slots survive the sort as `-1` instead of garbage.
- The caller-owned `radix_aux_{indices,logits}` design is the correct fix for the G4 CUDA-Graph stale-pointer hazard, and mirrors the existing `heuristic_scratch` convention. I verified the part-1 write (`rowIdx*gridDim.y*topK + blockIdx.y*topK`) and part-2 read (`rowIdx*numBlocksToMerge*topK`) use a self-consistent flat stride, so the over-sized `(max_gen_tokens, 10, topK)` allocation is correct and the cpp `needed_elts` check matches exactly.
- `compress_ratio` threading (N = actual_kv_len/cr, `preIdxOffset=0` for cr!=1) is consistent across fp32/bf16/fp16 entries and the heuristic launcher.
- Good test discipline: the split+merge path is validated vs `torch.topk` via the SM-saturation / launch-policy tests (which reach bp up to ~9), plus allocation-independence + CUDA-graph-replay equivalence and malformed-buffer validation.

**Non-blocking note (no change required to approve)**
- The prefill compile-time `useRadixSort` split (and the decode single-block choice) intentionally replaces the runtime per-row `finalCount`-based selection that #14268 landed on main; the prefill logic here is byte-identical to `feat/deepseek_v4`. Correctness is unaffected (both paths run the identical histogram refinement and differ only in the bounded final-sort). Worth a one-line note in the PR/commit message that this is a deliberate replacement of #14268's runtime-adaptive path, and optionally one >=200k-column equivalence case to cover the forced `blocksPerRow=10` branch (the removed multi-pass-radix tests previously exercised 131k-524k columns; the new ground-truth tests top out near 32k).

Nice work.

hyukn (approved):

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
```diff
@@ -94,13 +94,21 @@ def warmup_heuristic_topk_decode(top_k: int = 2048,
...
+                                   device=device)
     torch.ops.trtllm.indexer_topk_decode(logits,
                                          seq_lens,
                                          indices,
                                          1,
                                          top_k,
                                          pre_idx=pre_idx,
-                                         heuristic_scratch=scratch)
+                                         heuristic_scratch=scratch,
+                                         radix_aux_indices=radix_aux_indices,
```
yuxianq, line 110: Do we need to add compress_ratio here?

yuxianq (approved):

tensorrt_llm/_torch/attention_backend/sparse/dsa.py
longcheng-nv: No — `compress_ratio` isn't needed here.

`warmup_heuristic_topk_decode`'s only job is to trigger the one-time host-side `cudaDeviceGetAttribute(MultiProcessorCount / L2CacheSize)` queries before any CUDA Graph capture begins. Those values are cached via `std::call_once` inside `getSchemeXBounds(numColumns, bytesPerElem)` (reached through `computeIndexerTopKDecodeBlocksPerRow(numRows, numColumns, splitWorkThreshold)`), and **neither of those signatures takes `compress_ratio`** — the cached state depends only on `numColumns` + device attributes.

The default `compress_ratio=1` still passes the `compressRatioOk = (cr==1 || cr==4)` gate with a non-null `pre_idx`, so the warmup call routes through the same heuristic dispatcher entry that performs (and caches) the attribute query. `compress_ratio` only affects per-call numerics (`N = kv_len/cr`, `preIdxOffset`, the heuristic eligibility gate) — none of which feed the cached static state, and the warmup discards its (zeroed) output anyway.

For reference, `feat/deepseek_v4` runs `compress_ratio=4` in the real decode path while still calling this warmup with the default `cr=1`, so leaving it out here matches the established behavior. Happy to thread a `compress_ratio` param through the warmup purely for signature parity if you'd prefer, but it's functionally a no-op.

PR #15383, opened 2026-06-15
[None][feat] Handle empty-block ranks with Helix CP

brb-nv:
## Description

With Helix context parallelism (CP), a sequence's KV cache is sharded across CP ranks. When a sequence has fewer KV blocks than the CP size, the highest CP rank(s) own zero blocks for that sequence ("empty" ranks). This shows up most easily with short prompts on cp_size > 1.

Empty Helix CP ranks aren't supported on main. This MR adds support for the same.

## Test Coverage

```
$ pytest tests/integration/defs/disaggregated/test_disaggregated.py::test_disaggregated_deepseek_v3_lite_bf16_tllm_gen_helix -s -v
$ pytest tests/unittest/_torch/thop/parallel_hw_agnostic/test_helix_postprocess.py::TestHelixPostProcess::test_helix_postprocess_empty_rank_noop -s -v
$ pytest tests/unittest/_torch/executor/test_request_utils.py::test_merge_helix_requests_empty_ranks -s -v
```

## PR Checklist

## GitHub Bot Help

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved edge case handling for context parallelism scenarios where some ranks own zero KV cache blocks, preventing potential undefined behavior and failed transfers.

* **New Features**
  * Added support for asymmetric KV cache block distribution across context parallel ranks in Helix mode.

* **Tests**
  * Added test coverage for context parallelism with empty ranks and asymmetric block distributions.

### DIFF
--- a/tensorrt_llm/_torch/modules/attention.py
+++ b/tensorrt_llm/_torch/modules/attention.py
@@ -2,7 +2,7 @@ import functools
 import math
 import os
 import weakref
-from typing import List, Optional, Union, cast
+from typing import List, Optional, Tuple, Union, cast
 
 import torch
 from torch import nn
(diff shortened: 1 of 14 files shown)


SimengLiu-nv (approved):
The kvcache related changes look good to me.

tensorrt_llm/_torch/pyexecutor/request_utils.py
```diff
@@ -280,10 +286,14 @@ def partition_context_for_helix(
     input_id_blocks = list(all_input_ids.split(tokens_per_block, dim=-1))
     position_id_blocks = list(all_position_ids.split(tokens_per_block, dim=-1))
 
-    input_ids_this_rank = torch.cat(input_id_blocks[cp_rank::cp_size], dim=-1).flatten().tolist()
-    position_ids_this_rank = (
-        torch.cat(position_id_blocks[cp_rank::cp_size], dim=-1).flatten().tolist()
-    )
+    my_input_blocks = input_id_blocks[cp_rank::cp_size]
```
Tabrizian: nit:
```suggestion
    curank_input_blocks = input_id_blocks[cp_rank::cp_size]
```

Tabrizian (approved):
small nit otherwise looks good to me.

LarryXFly (approved):

fredricz-20070104 (approved):
Approved.

fredricz-20070104:
Overall this LGTM — the core invariant (rank 0 always owns global block 0, so the attention-combine denominator is never zero) is sound, the empty-block send/receive skipping is symmetric across `cacheFormatter.cpp` / `mlaCacheFormatter.cpp` / `dataTransceiver.cpp`, the C++ boundary guards are correct, and test coverage is good. Before merging I'd just like to confirm two changes that are broader than the "handle empty-block ranks" title suggests:

1. **`use_nccl_for_alltoall` default flips `True -> False`** (`_mnnvl_utils.py:431` and the `_helix_post_process` path in `attention.py`). This changes the default all-to-all path for *all* Helix CP scenarios when the key isn't explicitly set (defaulting to MNNVL instead of NCCL), which is unrelated to empty-block handling. Both call sites are changed consistently so it looks intentional — could you confirm this is the intended new default and call it out in the PR description?

2. **`requestImpl.h::validate()` downgrades `TLLM_CHECK(!mInputTokenIds.empty())` to a WARNING.** This relaxes a *global* request-validation check, not just the Helix empty-rank path. In principle it lets empty-input requests on other code paths turn a hard failure into a warning, which could mask non-Helix bugs. It's needed functionally here — would it be feasible to gate it more narrowly (e.g. only in the Helix CP context) so the global invariant is preserved elsewhere?

Neither is blocking from my side if both are intentional and the scope impact is acceptable — just want them on the record before approval.

tensorrt_llm/_torch/modules/attention.py
```diff
@@ -147,11 +219,18 @@ def _helix_post_process(
...
+    zero_kv_mask marks tokens for which this CP rank owns no KV blocks; those rows
+    are forced to a no-op contribution before the exchange (see
+    _helix_sanitize_empty_kv).
+
     When *aux_stream* and *ln_events* are provided the two
     ``.contiguous()`` calls in the FIFO-v1 path are overlapped on
     separate CUDA streams for better performance.
     """
-    if mapping.cp_config.get("use_nccl_for_alltoall", True):
+    partial_o, softmax_stats = _helix_sanitize_empty_kv(partial_o,
```
PerkzZheng, line 243: are those NANs generated by previous attention kernels because of invalid pages that contain NANs ? we have met similar issues somewhere else because the reused pages (evicted from other requests) contain NANs.

brb-nv: Hi Perkz, the NaNs are not due to kernel issues but because of a context parallel rank owning 0 KV blocks. So, the gen attention attends over an empty key set and normalizes by a zero softmax sum. So, we mask-and-sanitize such rows to (-inf, 0) before the all-to-all.

brb-nv:
> Overall this LGTM — the core invariant (rank 0 always owns global block 0, so the attention-combine denominator is never zero) is sound, the empty-block send/receive skipping is symmetric across `cacheFormatter.cpp` / `mlaCacheFormatter.cpp` / `dataTransceiver.cpp`, the C++ boundary guards are correct, and test coverage is good. Before merging I'd just like to confirm two changes that are broader than the "handle empty-block ranks" title suggests:
> 
> 1. **`use_nccl_for_alltoall` default flips `True -> False`** (`_mnnvl_utils.py:431` and the `_helix_post_process` path in `attention.py`). This changes the default all-to-all path for _all_ Helix CP scenarios when the key isn't explicitly set (defaulting to MNNVL instead of NCCL), which is unrelated to empty-block handling. Both call sites are changed consistently so it looks intentional — could you confirm this is the intended new default and call it out in the PR description?
> 2. **`requestImpl.h::validate()` downgrades `TLLM_CHECK(!mInputTokenIds.empty())` to a WARNING.** This relaxes a _global_ request-validation check, not just the Helix empty-rank path. In principle it lets empty-input requests on other code paths turn a hard failure into a warning, which could mask non-Helix bugs. It's needed functionally here — would it be feasible to gate it more narrowly (e.g. only in the Helix CP context) so the global invariant is preserved elsewhere?
> 
> Neither is blocking from my side if both are intentional and the scope impact is acceptable — just want them on the record before approval.

Reverted the default switch of `use_nccl_for_alltoall`. The change to `validate()` is required as some context-parallel ranks may have no KV blocks (and hence no corresponding tokens) to begin with.

tensorrt_llm/_torch/modules/attention.py
PerkzZheng: okay, it makes sense. thanks.

PerkzZheng (approved):

